PR #3, opened 2015-12-19
Moved to mostly using only functional components.

n1k0:
After reading https://facebook.github.io/react/blog/2015/12/18/react-components-elements-and-instances.html, I couldn't resist the urge to moving all stateless components to pure functional ones.

r=? @leplatrem


n1k0:
I'm just gonna land this with r=me and keep moving forward.

PR #4, opened 2015-12-21
Fields & widgets

n1k0:
This will prepare introducing uiSchema for specifying the kind of widget we want for a given JSONSchema data type.

- A field is a container for a label and (a) widget(s)
- A widget is basically the html input form tag (`input`, `select`, etc.)


src/components/widgets/RadioWidget.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+
+import Field from "./../fields/Field";
+
+
+export default function RadioField({schema, formData, options, onChange}) {
+  // Generating a unique field name to identify this set of radio buttons
+  const name = Math.random().toString();
```
leplatrem: Do we have this for the other widgets ?

n1k0: Noipe as this is really specific to radio buttons.

n1k0:
Landing this with r=me.

PR #15, opened 2015-12-25
Fixes #14: Update form state when new formData prop is passed.

### LINKED ISSUE #14: Allow setting formData after the Form component is mounted
This would be useful for the gh-pages live form builder, so one can alter the formData object and observe live form updates. That should be implemented through `componentWillReceiveProps`.

n1k0:
refs #14


n1k0:
r=me

PR #18, opened 2015-12-29
Fixes #9: Default object field property labels.

### LINKED ISSUE #9: Default field label should be its parent object key property
```js
{
  type: "object",
  properties: {
    firstName: {type: "string"}
  }
}
```

The firstName field should have its label automatically set to `firstName` by default.

n1k0:
r=? @Natim


Natim:
r+

PR #19, opened 2015-12-29
Added UpDown and RangeWidget.

n1k0:
r=? @Natim ([demo](http://mozilla-services.github.io/react-jsonschema-form/) has been updated with this patch applied, look at the Numbers section)

Note that I've also removed a bunch of obsolete fixtures in this patch.


n1k0:
r=me

PR #26, opened 2015-12-31
Fixes #13: Added support for custom widget components.

### LINKED ISSUE #13: Add support for custom widget components
```js
function MyCustomWidget(props) {
  const {placeholder, value, defaultValue, required, onChange} = props;
  return (
    <div className="my-custom-widget">
      <em>I am a custom widget with value {value}.</em>
      <input type="text"
        value={value}
        defaultValue={defaultValue}
        placeholder={placeholder}
        required={required}
        onChange={(event) => onChange(event.target.value)} />
    </div>
  );
}

const uiSchema = {
  foo: {widget: MyCustomWidget}
}
```

n1k0:
refs #13

For now it only supports single input widgets, next step will be to provide object field ones; but let's iterate on this.

r=? @leplatrem

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,47 @@ Will result in:
 </div>
 ```
 
+## Custom widgets
+
+You can provide your own custom widgets to a uiSchema for the following json data types:
+
+- `string`
+- `number`
+- `integer`
+- `boolean`
+- `date-time`
+
+```js
+const schema = {
+  type: "string"
+};
+
+const uiSchema = {
+  widget: (props) => {
+    return (
+      <input type="text"
+        className="custom"
+        value={props.value}
+        defaultValue={props.defaultValue}
+        required={props.required}
+        onChange={(event) => props.onChange(event.target.value)} />
+    );
+  }
+};
+
+render(<Form schema={schema} uiSchema={uiSchema} />);
+```
+
+The following props are passed to the widget component:
+
+- `schema`: The JSONSchema subschema object for this field;
+- `value`: The current value for this field;
+- `defaultValue`: The default value for this field;
+- `required`: The required status of this field;
+- `onChange`: The value change event handler; call it with the new value everytime it changes;
+- `placeholder`: The placeholder value, if any;
+- `options`: The list of options for `enum` fields;
+
 ## Development server
 
 ```
--- a/playground/app.js
+++ b/playground/app.js
@@ -119,24 +119,24 @@ class App extends Component {
           <h1>react-jsonschema-form</h1>
           <Selector onSelected={this.load.bind(this)} />
         </div>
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -138,6 +138,47 @@ Will result in:
...
+
+The following props are passed to the widget component:
+
+- `schema`: The JSONSchema subschema object for this field;
+- `value`: The current value for this field;
+- `defaultValue`: The default value for this field;
+- `required`: The required status of this field;
+- `onChange`: The value change event handler; call it with the new value everytime it changes;
+- `placeholder`: The placeholder value, if any;
+- `options`: The list of options for `enum` fields;
```
leplatrem: nit: `enum` is not listed among supported data types

README.md
```diff
@@ -138,6 +138,47 @@ Will result in:
...
+- `boolean`
+- `date-time`
+
+```js
+const schema = {
+  type: "string"
+};
+
+const uiSchema = {
+  widget: (props) => {
```
leplatrem: what does `widget` mean here?

README.md
```diff
@@ -138,6 +138,47 @@ Will result in:
...
+The following props are passed to the widget component:
+
+- `schema`: The JSONSchema subschema object for this field;
+- `value`: The current value for this field;
+- `defaultValue`: The default value for this field;
+- `required`: The required status of this field;
+- `onChange`: The value change event handler; call it with the new value everytime it changes;
+- `placeholder`: The placeholder value, if any;
+- `options`: The list of options for `enum` fields;
+
```
leplatrem: Shouldn't we mention the `getAlternativeWidget()` function ?

```js
const uiSchema = {
     "comments": getAlternativeWidget("string", "textarea")
};

render(<Form schema={schema} uiSchema={uiSchema} />);
```

leplatrem:
r+ (nicer with a bit of improvements in readme  IMO)

README.md
n1k0: Hmm, `enum` is not a jsonschema datatype, it's a property specifying the field restricts the possible values for its type. How would you alter the docs to convey that?

README.md
n1k0: That function isn't intented to be part of the public API; you can simply describe the type of alt widget using its string identifier in the `widget` uiSchema property.

README.md
n1k0: Well, that's the way you set a widget in uiSchema. Thought it was clear enough, but oh well :)

n1k0: Isn't it clearer if you read the whole section instead of just the diff as a normal user would do? https://github.com/mozilla-services/react-jsonschema-form/blob/custom-widgets/README.md#alternative-widgets

PR #29, opened 2016-01-09
Added object properties ordering to uiSchema.

n1k0:
## Object fields ordering

The `uiSchema` object spec also allows you to define in which order a given object field properties should be rendered using the `order` property:

```jsx
const schema = {
  type: "object",
  properties: {
    foo: {type: "string"},
    bar: {type: "string"}
  }
};

const uiSchema = {
  order: ["bar", "foo"]
};

render((
  <Form schema={schema}
        uiSchema={uiSchema} />
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
 
 ## Object fields ordering
 
-The `uiSchema` object spec also allows you to define in which order a given object field properties should be rendered using the `order` property:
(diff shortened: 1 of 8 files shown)


n1k0:
Bah, I realize you then can't use an object field named "order". I'll come with another name, eg. "__order".

n1k0:
Much better. Also, I'll have to rename `ui:widget` to `ui:widget` as well for the same reason. I'll do that in a next PR.

almet:
Cool :')

PR #36, opened 2016-02-19
fix:Form use old schema to validate new formData when new props have new schema and new formData

fangj:
`Form` uses old schema to validate new `formData` when new props have a new schema and new `formData`. For example:

```js
old_props={schema:{type: "string"}, formData: "yo"}
new_props={schema:{type: "number"}, formData: "42"}
```

After switching from `old_props` to `new_props`, the `Form` uses schema's `old_props` to validate `formData` from `new_props`. And give error said *42 is not  string*.

It is because `validate(formData)` uses `this.props.schema` to validate data, and `validate()` is called from `getStateFromProps()`, which has not applied the new props to `this.props`. so I use `Object.assign({}, this.props, newProps)` to merge the new props.


src/components/Form.js
```diff
@@ -25,13 +25,14 @@ export default class Form extends Component {
       status: "initial",
       formData,
       edit,
-      errors: edit ? this.validate(formData) : []
+      errors: edit ? this.validate(formData,props) : []
```
almet: nit: add a space before the variable.

src/components/Form.js
```diff
@@ -25,13 +25,14 @@ export default class Form extends Component {
...
     };
   }
 
-  validate(formData) {
+  validate(formData,newProps) {
```
almet: ditto

src/components/Form.js
```diff
@@ -25,13 +25,14 @@ export default class Form extends Component {
...
     const validator = new Validator();
-    return validator.validate(formData, this.props.schema).errors;
+    var props=!newProps?this.props:Object.assign({},this.props,newProps);
```
almet: ditto :)

almet: I believe it would be easier to read if you do the conditionel on `newProps` directly, rather than negating it.

src/components/Form.js
```diff
@@ -25,13 +25,17 @@ export default class Form extends Component {
...
+      errors: edit ? this.validate(formData, props) : []
     };
   }
 
-  validate(formData) {
+  validate(formData, newProps) {
     const validator = new Validator();
-    return validator.validate(formData, this.props.schema).errors;
+    var schema=this.props.schema;
+    if(newProps&&newProps.schema){
```
almet: still here some spaces missing here :)

src/components/Form.js
```diff
@@ -25,13 +25,17 @@ export default class Form extends Component {
...
+      schema=newProps.schema;
```
almet: ditto

test/index_test.js
```diff
@@ -1351,5 +1351,21 @@ describe("Form", () => {
...
+      it("should update form state from new formData and new schema prop value", () => {
+        const schema = {type: "string"};
+        const {comp} = createComponent({schema});
+
+        comp.componentWillReceiveProps({formData: "yo"});
+
+        expect(comp.state.formData).eql("yo");
+
+        const schema2 = {type: "number"};
+        comp.componentWillReceiveProps({schema:schema2,formData: 2});
```
almet: ditto

test/index_test.js
```diff
@@ -1351,5 +1351,21 @@ describe("Form", () => {
...
+        expect(comp.state.errors).to.have.length.of(0);
+
```
almet: you can remove this one!

test/index_test.js
```diff
@@ -541,14 +541,6 @@ describe("Form", () => {
         .eql("my object");
     });
 
-    it("should render a customized title", () => {
```
almet: Hmm, I think this wasn't merged properly!

almet:
this looks great, but I think you mis-merged the tests file (removing some new test rather than adding your new one)

n1k0:
This is strange, the current global patch diff doesn't have tests anymore. I'm probably take over the issue and land the fix separately.

magopian:
I believe this issue should now be fixed by 1e3796e56040cad763ba1ed492d6d8fd8c43fafc and could be closed

magopian:
#53 is now merged, so we can close this one ;)

Thanks @fangj for filing this issue!

PR #38, opened 2016-02-22
Allow to customize form titles.

almet:
Use a registry rather than passing yet another prop to children.

r? @n1k0

### DIFF
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -39,7 +39,8 @@ class ObjectField extends Component {
   render() {
     const {schema, uiSchema, name} = this.props;
     const title = schema.title || name;
-    const SchemaField = this.props.SchemaField;
+    const SchemaField = this.props.registry.SchemaField;
+    const TitleField = this.props.registry.TitleField;
     try {
       var orderedProperties = orderProperties(
         Object.keys(schema.properties), uiSchema["ui:order"]);
@@ -53,7 +54,7 @@ class ObjectField extends Component {
     }
     return (
       <fieldset>
-        {title ? <legend>{title}</legend> : null}
+        {title ? <TitleField title={title}/> : null}
         {schema.description ?
           <div className="field-description">{schema.description}</div> : null}
         {
@@ -66,7 +67,7 @@ class ObjectField extends Component {
               uiSchema={uiSchema[name]}
               formData={this.state[name]}
               onChange={this.onChange.bind(this, name)}
-              SchemaField={SchemaField}/>
+              registry={this.props.registry} />
             );
         })
       }</fieldset>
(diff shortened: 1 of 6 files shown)


test/index_test.js
```diff
@@ -541,6 +541,16 @@ describe("Form", () => {
         .eql("my object");
     });
 
+    it("should render a customized title", () => {
+      const CustomTitleField = function(props) {
+        return (<div id="custom">{ props.title }</div>);
+      };
```
n1k0: Nit: 

```js
const CustomTitleField = ({title}) => <div id="custom">{title}</div>;
```

src/components/fields/ObjectField.js
```diff
@@ -81,7 +82,7 @@ if (process.env.NODE_ENV !== "production") {
     onChange: PropTypes.func.isRequired,
     formData: PropTypes.object,
     required: PropTypes.bool,
-    SchemaField: PropTypes.func.isRequired,
+    registry: PropTypes.object,
```
n1k0: Would be great if we could describe a little more what's expected as a data structure here, eg. expected keys and their type.

src/components/fields/ObjectField.js
```diff
@@ -39,7 +39,8 @@ class ObjectField extends Component {
   render() {
     const {schema, uiSchema, name} = this.props;
     const title = schema.title || name;
-    const SchemaField = this.props.SchemaField;
+    const SchemaField = this.props.registry.SchemaField;
+    const TitleField = this.props.registry.TitleField;
```
n1k0: Nit:

```js
const {SchemaField, TitleField} = this.props.registry;
```

src/components/fields/TitleField.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+function TitleField(props) {
+  return <legend>{props.title}</legend>;
+}
```
n1k0: Would be great to have propTypes defined as for other components.

n1k0:
Note: This will break BC for consumers of the current API, because you now have to pass a registry object instead of a SchemaField. We need to document this change in the release notes with a big red warning.

almet:
> Note: This will break BC for consumers of the current API, because you now have to pass a registry object instead of a SchemaField. We need to document this change in the release notes with a big red warning.
This won't, you still pass `SchemaField`, the registry is only exposed internally.

n1k0:
Ah right, cool. So we only need to update the docs and we're good to go with nits addressed.

almet:
Docs updated.

n1k0:
LGTM, r+

n1k0:
Could you take care of releasing v0.11.0?

almet:
Did this:

```bash
npm run dist
vim package.json
npm publish
git commit -am "Bump v0.11.0"
git tag v0.11.0
git push
```

n1k0:
No `npm run build`? Did you test installing the package after the release?

almet:
In the `package.json` file, I see `"dist": "npm run build:lib && npm run build:dist && npm run build:css",`.

Despite that, I still need to run `npm run build`?

n1k0:
Ah no it's fine then :)

Issue #44, opened 2016-03-03
Object without `properties` throws


chmac:
Reproduction from the example:

``` js
const schema = {
  title: "Todo Tasks",
  type: "object",
  required: ["title"],
  properties: {
    title: {type: "object", title: "Title"},
    done: {type: "boolean", title: "Done?", default: false}
  }
};
```

The `title` field has been changed to `type: "object"`. This causes the package to throw unless there is a `properties` key on the `title` field. In my (admittedly very light) testing, the schema is valid without the `properties` keyword, but I could be entirely wrong about that.

The error I see is `Uncaught TypeError: Cannot convert undefined or null to object`. Adding `properties: {}` solves the issue for me.

Not sure if this is really a bug or not. Although I guess there would be no harm in checking for the existence of `properties` rather than assuming it exists. I'll see if I can find time to come up with a PR on that front if it would be helpful.

n1k0:
> the schema is valid without the properties keyword, but I could be entirely wrong about that.

Confirmed, an object field doesn't required a properties parameter.

> I'll see if I can find time to come up with a PR on that front if it would be helpful.

That'd be just plain great, thank you :)

chmac:
Done, without any testing at all...

chmac:
@n1k0 Trying to get the tests running locally like `npm install --dev` but having some challenges. First time my dev VM ran out of memory. I killed some other big memory hogs and started again. Now it's been running for a while, maybe 15 minutes, and is complaining about missing repositories, and permissions errors in `~/.npm`. Any suggestions?

n1k0:
I suggest to use [nvm](https://github.com/creationix/nvm) to install a separated nodejs environment and deal with different versions much more easily.

After installing it, you'll have node versions it managed installed in your homedir, which usually prevents permission errors. I suggest you `nvm install 4 && nvm use 4` to start hacking on this specific project. Don't forget to `rm -rf ./node_modules` and run `npm install` again when you're done.

This usually works everytime :)

n1k0:
I'm realizing we can't really support `object` fields without any properties described, as that would mean an empty form or fieldset. We should probably raise instead, or at least trigger a warning in the console to notify the developer about that.

chmac:
If the field is optional, and has `additionalProperties: true` then it could simply be skipped from the form. But I think the more useful strategy would be to allow a custom widget for fields of type object. Then we could implement file upload widgets and other whacky stuff which produces objects.

[...]

PR #46, opened 2016-03-03
Don't assume `schema.properties` exists. #44

chmac:
This is **completely untested**.

I'm not able to easily test this on our system, but as best I can tell without testing, I guess it'll work. :-)

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -67,10 +67,14 @@ export function getDefaultFormState(schema) {
     return schema.default;
   }
   if (schema.type === "object") {
-    return Object.keys(schema.properties).reduce((acc, key) => {
-      acc[key] = getDefaultFormState(schema.properties[key]);
-      return acc;
-    }, {});
+    if (schema.properties) {
+      return Object.keys(schema.properties).reduce((acc, key) => {
+        acc[key] = getDefaultFormState(schema.properties[key]);
+        return acc;
+      }, {});
+    } else {
+      return {}
+    }
   }
   return defaultTypeValue(schema.type);
 }


src/utils.js
```diff
@@ -67,10 +67,14 @@ export function getDefaultFormState(schema) {
...
-      return acc;
-    }, {});
+    if (schema.properties) {
+      return Object.keys(schema.properties).reduce((acc, key) => {
+        acc[key] = getDefaultFormState(schema.properties[key]);
+        return acc;
+      }, {});
+    } else {
+      return {}
+    }
```
n1k0: We could actually simply default the passed `schema.properties` value in `Object.keys`:

```js
return Object.keys(schema.properties || {}).reduce((acc, key) => {
```

n1k0:
Thank you!

I'm expecting a unit test for this change; if you struggle with writing one I can take care of it :)

src/utils.js
```diff
@@ -67,10 +67,14 @@ export function getDefaultFormState(schema) {
...
-      acc[key] = getDefaultFormState(schema.properties[key]);
-      return acc;
-    }, {});
+    if (schema.properties) {
+      return Object.keys(schema.properties).reduce((acc, key) => {
+        acc[key] = getDefaultFormState(schema.properties[key]);
+        return acc;
+      }, {});
+    } else {
+      return {}
```
n1k0: Nit: Missing semicolon (hence build failure). You can run `npm run lint` before pushing to ensure it doesn't break :)

src/utils.js
chmac: Smart, that's a much better approach. I'll kill this PR and create a new one with only that change, keep the git history neater.

chmac:
I'll have a go at a unit test also! :-)

Issue #43, opened 2016-03-03
Option to hide a field?


chmac:
Is there an option we can pass to the ui schema that would result in a field being hidden?

almet:
that's currently not implemented, but if you're willing to add one, don't hesitate!

n1k0:
We should definitely add support for hidden fields (with static value). @chmac, would you want to start working on this?

chmac:
@n1k0 :-) If you can point me in the right direction I'm happy to have a look, but we're using this package inside meteor which makes developing on it a little tricky. I'm not exactly sure where my `node_modules/` folder is to start messing with the source while the code is running...

n1k0:
You should definitely clone this repo and work using the existing unit and DOM integration test suites available.

Basically the idea here would be to create a new kind of widget, eg. `HiddenWidget` which would simply hold the associated formData value it's attached to. Also we should add support for using it to override default widget used to render a given field using the `uiSchema` objecy and a `hidden` widget value for that field.

If that sounds super complicated no bif deal, you can either try or we'll take care of it. But if you want to work on this I'll provide guidance so you're not alone :)

PR #50, opened 2016-03-04
#43 hide a field / added HiddenWidget

HelloLudger:
Added HiddenWidget and hide HiddenFields by adding `display: none` to Wrapper within SchemaFields.js.

Sorry, no tests. Couldn't get the tests to run, in addition to my 0-knowledge of testing.

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -83,7 +83,7 @@ function SchemaField(props) {
   return (
     <Wrapper
       label={schema.title || name}
-      hidden={(uiSchema["ui:widget"] === "hidden" ? true : false)}
+      hidden={uiSchema["ui:widget"] === "hidden"}
       required={required}
       type={schema.type}
       displayLabel={displayLabel}
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -44,9 +44,11 @@ function getContent({type, label, required, children, displayLabel}) {
 }
 
 function Wrapper(props) {
+  const style = props.hidden ? {"display": "none"} : {};
   const {type, classNames} = props;
+
   return (
-    <div className={`field field-${type} ${classNames}`}>
+    <div style={style} className={`field field-${type} ${classNames}`}>
       {getContent(props)}
     </div>
   );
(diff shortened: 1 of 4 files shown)


src/components/fields/SchemaField.js
```diff
@@ -81,6 +83,7 @@ function SchemaField(props) {
   return (
     <Wrapper
       label={schema.title || name}
+      hidden={(uiSchema["ui:widget"] === "hidden" ? true : false)}
```
n1k0: Nit: `uiSchema["ui:widget"] === "hidden"` is good enough

src/components/widgets/HiddenWidget.js
```diff
@@ -0,0 +1,39 @@
...
+  required,
+  onChange
+}) {
+  return (
+    <input type="hidden"
+      value={value}
+      defaultValue={defaultValue}
+      placeholder={placeholder}
+      required={required}
+      onChange={(event) => onChange(event.target.value)} />
```
n1k0: Is this input value supposed to change ever?

n1k0:
We definitely want tests and documentation updated here. What prevents you running them? What error are you getting?

src/components/widgets/HiddenWidget.js
HelloLudger: good question... I thought about it. Maybe in combination with [_Add computed fields ? #45_](https://github.com/mozilla-services/react-jsonschema-form/issues/45).
But I think it's safe to say that the placeholder won't be needed.

HelloLudger:
Maybe the problem is that I use Windows. I don't know. When I start `npm test` the error says that the command "NODEv_ENV" is not a viable command.

```
C:\Projekte\react-jsonschema-form>npm test

> react-jsonschema-form@0.12.0 test C:\Projekte\react-jsonschema-form
> NODE_ENV=test mocha --compilers js:babel/register --recursive --require ./test/setup-jsdom.js $(find test -name '*_test.js')

Der Befehl "NODE_ENV" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
npm ERR! Test failed.  See above for more details.
```

n1k0:
Superseded by #111.

PR #51, opened 2016-03-04
Force the hiding of the label

HelloLudger:
Allows to hide the creation of the label by adding `ui:hideLabel: true` to the corresponding uiSchema.

E.g. (from the simple example):
```json
{
  "age": {
    "ui:widget": "updown",
    "ui:hideLabel": true
  },
  "bio": {
    "ui:widget": "textarea"
  },
  "password": {
    "ui:widget": "password"
  }
}
```

This is useful if you want to create a custom widget to be styled by an external CSS framework.

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -45,6 +45,7 @@ function getContent({type, label, required, children, displayLabel}) {
 
 function Wrapper(props) {
   const {type, classNames} = props;
+
   return (
     <div className={`field field-${type} ${classNames}`}>
       {getContent(props)}
@@ -66,7 +67,7 @@ if (process.env.NODE_ENV !== "production") {
 Wrapper.defaultProps = {
   classNames: ""
 };
-
+ 
 function SchemaField(props) {
   const {schema, uiSchema, name, required} = props;
   const FieldComponent = COMPONENT_TYPES[schema.type] || UnsupportedField;
@@ -77,6 +78,8 @@ function SchemaField(props) {
   if (schema.type === "object") {
     displayLabel = false;
   }
+
+  if (props.uiSchema["ui:hideLabel"] ===  true) displayLabel=false;
 
   return (
     <Wrapper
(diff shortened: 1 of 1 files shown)


n1k0:
Couldn't this be achieved using just CSS? Using the `className` uiSchema option, you could easily target and hide the label I think, eg. `.myfield > label { display: none }`.

What do you think?

HelloLudger:
If a label is provided, the function _getContent()_ will wrap label string and the form element in a `<label></label>` tag, so your code would also hide the form element. If you only target the span containing the label string, the form element would still be wrapped by an label what could result in conflicts with external CSS frameworks.

E.g. if you try to replicate [Semantic UI's Checkbox](http://semantic-ui.com/collections/form.html#checkbox), you would end up with two label fields.

almet:
Hi, thanks for taking the time to create a discussion around this.

This pull request needs to have some tests added which reflect the new feature you're adding here, thanks!

leplatrem:
>  you would end up with two label fields.

Maybe that's something we can fix: checkboxes should not have any label ?
capture d ecran de 2016-04-04 15-28-58

n1k0:
For checkboxes it makes a lot of sense to hide the field label as the input one is the same, indeed. I'd tend to do that automatically and not through an option though.

leplatrem:
Since this was addressed in #119. I suggest that we close this approach. Don't hesitate to reopen if you think it's still relevant! Thanks for your feedback!

mikedizon:
This would actually be great since I'm using material-ui. The TextFields provide hintText that's pretty slick

mplis-jetsetter:
@n1k0 I have a use case different than the one addressed in #119 where I simply want to hide the label for a particular field. I'm aware that the uiSchema accepts a `classNames` parameter, but I don't know if it's possible to specifically target the label through the uiSchema object. If this isn't something that's supported, that's fine, I can deal with it.

n1k0:
Yeah we might eventually support this use case, stay tuned.

zalmoxisus:
It would still be rather useful to have this merged. I often need to use forms without labels, and it would be better than hiding labels with css. Also it's good for consistency, since a checkbox can be replaced with inline radio group (and the label is not necessary as well).

PR #69, opened 2016-03-17
Fixes #11: Added support for referenced schema definitions.

### LINKED ISSUE #11: Add support for canonical and inline referencing
Spec: http://json-schema.org/latest/json-schema-core.html#anchor25

I suggest we start by supporting local definitions only:

```json
{
  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": { "type": "string" },
        "city":           { "type": "string" },
        "state":          { "type": "string" }
      },
      "required": ["street_address", "city", "state"]
    }
  },

  "type": "object",

  "properties": {
    "billing_address": { "$ref": "#/definitions/address" },
    "shipping_address": { "$ref": "#/definitions/address" }
  }
}
```

(Sample JSONSchema source extracted from http://spacetelescope.github.io/understanding-json-schema/structuring.html)

n1k0:
Refs #11.

## Schema definitions & references

This library partially supports [inline schema definition dereferencing]( http://json-schema.org/latest/json-schema-core.html#rfc.section.7.2.3), which is Barbarian for *avoiding to copy and paste commonly used field schemas*:

```json
{
  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": { "type": "string" },
        "city":           { "type": "string" },
        "state":          { "type": "string" }
      },
      "required": ["street_address", "city", "state"]
    }
  },
  "type": "object",
  "properties": {
    "billing_address": { "$ref": "#/definitions/address" },
    "shipping_address": { "$ref": "#/definitions/address" }
  }
}
```

*(Sample schema courtesy of the [Space Telescope Science Institute](http://spacetelescope.github.io/understanding-json-schema/structuring.html))*

Note that it only supports local definition referencing, we do not plan on fetching foreign schemas over HTTP anytime soon. Basically, you can only reference a definition from the very schema object defining it.

Feedback=? @magopian @leplatrem

### DIFF
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -66,6 +66,146 @@ describe("Form", () => {
     });
   });
 
+  describe("Schema definitions", () => {
+    it("should use a single schema definition reference", () => {
+      const schema = {
+        definitions: {
+          testdef: {type: "string"}
+        },
+        $ref: "#/definitions/testdef"
+      };
+
+      const {node} = createComponent({schema});
+
+      expect(node.querySelectorAll("input[type=text]"))
+        .to.have.length.of(1);
+    });
+
+    it("should handle multiple schema definition references", () => {
+      const schema = {
+        definitions: {
+          testdef: {type: "string"}
+        },
+        type: "object",
+        properties: {
+          foo: {$ref: "#/definitions/testdef"},
+          bar: {$ref: "#/definitions/testdef"},
+        }
+      };
+
+      const {node} = createComponent({schema});
+
+      expect(node.querySelectorAll("input[type=text]"))
+        .to.have.length.of(2);
+    });
+
+    it("should handle deeply referenced schema definitions", () => {
+      const schema = {
+        definitions: {
+          testdef: {type: "string"}
+        },
+        type: "object",
+        properties: {
+          foo: {
+            type: "object",
+            properties: {
+              bar: {$ref: "#/definitions/testdef"},
+            }
+          }
+        }
+      };
+
+      const {node} = createComponent({schema});
+
+      expect(node.querySelectorAll("input[type=text]"))
+        .to.have.length.of(1);
+    });
+
+    it("should handle referenced definitions for array items", () => {
+      const schema = {
+        definitions: {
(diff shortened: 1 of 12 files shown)


leplatrem:
nice!

r+

n1k0:
Thanks, this was indeed suprisingly (suspiciously?) easy :)

I'll add more tests before landing this though.

n1k0:
@magopian brace yourself, I think we'll need to support referenced schema defaults propagation here :D

magopian:
errr, wat, can you explain?

n1k0:
> errr, wat, can you explain?

We need to deeply propagate schema defaults including when a referenced subschema is used for a given field. That means resolving the node where the definition is, checking for a default value and use it if any. Shouldn't be too hard tbh.

n1k0:
> We need to deeply propagate schema defaults including when a subschema is referenced at some given node in the tree

Addressed in 4ad85cdfba67d57504a10840a4097adc82a966f9

n1k0:
Now need to ensure referenced schema definitions can be ordered using uiSchema's `ui:order` directive.

test/Form_test.js
```diff
@@ -66,6 +66,146 @@ describe("Form", () => {
...
+            properties: {
+              bar: {$ref: "#/definitions/testdef"},
+            }
+          }
+        }
+      };
+
+      const {node} = createComponent({schema});
+
+      expect(node.querySelectorAll("input[type=text]"))
```
n1k0: Maybe assert on a more accurate selector, eg. `.field-object .field-object input[type=text]`?

test/Form_test.js
```diff
@@ -66,6 +66,146 @@ describe("Form", () => {
...
+            items: {$ref: "#/definitions/testdef"}
+          }
+        }
+      };
+
+      const {node} = createComponent({schema, formData: {
+        foo: ["blah"]
+      }});
+
+      expect(node.querySelectorAll("input[type=text]"))
```
n1k0: Maybe check the value to be equal to `blah`.

PR #70, opened 2016-03-17
Added shallow comparison checks to stateful components.

### LINKED ISSUE #54: Performance optimizations
Loving this component so far! However, in my own usage, and even in the demo site, there's a noticeable lag when attempting to enter text into fields at a reasonable typing speed. Enabling the "trace react updates" option in Chrome [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi), it becomes apparent that the entire form is re-rendered any time an input changes.

render

I haven't dug into the code, but it would seem reasonable to implement [`shouldComponentUpdate`](https://facebook.github.io/react/docs/component-specs.html#updating-shouldcomponentupdate) (perhaps simply with [`pure-render-mixin`](https://facebook.github.io/react/docs/pure-render-mixin.html)) to prevent rerendering of components not affected by the change in form values.

n1k0:
Fixes #54.

@aduth, @magopian: could you please check this branch and see if it solves (or improves) the performance issue you're encountering? Thanks!


n1k0:
First tests don't make me confident this is enough :/

n1k0:
**BAH.** https://github.com/facebook/react/issues/5677#issuecomment-165430821

n1k0:
Well, even if hackish as hell, last commit improves the situation a lot when filling large forms on the playground. Please guys provide feedback /cc @aduth @magopian

aduth:
Apologies that I've not had the time to take a close look at this. Just took it for a spin, and I can confirm that it feels much more performant for reasonably fast form input. I must say, I'm a bit concerned about the impact that including Immutable.js will have on the bundle size, as it looks as though it may add upwards of 70kb minified prior to gzipping.

n1k0:
I share your concerns. Though from my research, I couldn't find any better ways of performing nested object shallow comparison... If you have any idea how to implement `shouldRender` using vanilla js in an efficient fashion, I take it. Maybe just using `deep-equal`, `deeper` or equivalent may be just as fast but much less bloated?

n1k0:
Side note, would be great to have some benchmark suite as a referencen I'll setup something.

n1k0:
Achieving nearly the same perf boost using deeper, much better. Immutable would make sense if we wanted to use its structures everywhere, but it was definitely overkill here.

PR #73, opened 2016-03-18
Use first enum item default when no default is provided

n1k0:
Feedback=? @magopian


magopian:
awesome, flawless, perfect :star2: 

r+ :ship: :shipit: :rocket:

Issue #77, opened 2016-03-21
Support custom object and array fields through `ui:widget` 


n1k0:
Reference discussion: https://github.com/mozilla-services/react-jsonschema-form/issues/44#issuecomment-199160583

PR #78, opened 2016-03-21
Ref #77: Custom widget support for all types added to ui:widget.

n1k0:
### Custom field components

You can provide your own field components to a uiSchema for basically any json schema data type, by specifying a `ui:field` property.

For example, let's create and register a dumb `geo` component handling a *latitude* and a *longitude*:

```jsx
const schema = {
  type: "object",
  required: ["lat", "lon"],
  properties: {
    lat: {type: "number"},
    lon: {type: "number"}
  }
};

// Define a custom component for handling the root position object
class GeoPosition extends React.Component {
  constructor(props) {
    super(props);
    this.state = {...props.formData};
  }

  onChange(name) {
    return (event) => {
      this.setState({
        [name]: parseFloat(event.target.value)
      }, () => this.props.onChange(this.state));
    };
  }

  render() {
    const {lat, lon} = this.state;
    return (
      <div>
        <input type="number" value={lat} onChange={this.onChange("lat")} />
        <input type="number" value={lon} onChange={this.onChange("lon")} />
      </div>
    );
  }
}

// Define the custom field component to use for the root object
const uiSchema = {"ui:field": "geo"};

// Define the custom field components to register; here our "geo"
// custom field component
const fields = {geo: GeoPosition};

// Render the form with all the properties we just defined passed
// as props
render(<Form
  schema={schema}
  uiSchema={uiSchema}
  fields={fields}/>);
```

Note: Registered fields can be reused accross the entire schema.

Feedback? @chmac
r=?@magopian

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ A [live demo](https://mozilla-services.github.io/react-jsonschema-form/) is host
      - [Multiple choices list](#multiple-choices-list)
      - [Form action buttons](#form-action-buttons)
   - [Advanced customization](#advanced-customization)
-     - [Custom widgets](#custom-widgets)
+     - [Custom widget components](#custom-widget-components)
+     - [Custom field components](#custom-field-components)
      - [Custom SchemaField](#custom-schemafield)
      - [Custom titles](#custom-titles)
   - [Schema definitions and references](#schema-definitions-and-references)
@@ -271,7 +272,12 @@ render(
 
 ## Advanced customization
 
-### Custom widgets
+The API allows to specify your own custom *widgets* and *fields* components:
+
+- A *widget* represents a HTML tag for the user to enter data, eg. `input`, `select`, etc.
+- A *field* usually wraps one or more widgets and most often handles internal field state; think of a field as a form row, including the labels.
+
+### Custom widget components
 
 You can provide your own custom widgets to a uiSchema for the following json data types:
 
@@ -341,6 +347,65 @@ The following props are passed to the widget component:
 - `onChange`: The value change event handler; call it with the new value everytime it changes;
 - `placeholder`: The placeholder value, if any;
 - `options`: The list of options for `enum` fields;
+
+### Custom field components
+
+You can provide your own field components to a uiSchema for basically any json schema data type, by specifying a `ui:field` property.
+
+For example, let's create and register a dumb `geo` component handling a *latitude* and a *longitude*:
+
+```jsx
+const schema = {
+  type: "object",
+  required: ["lat", "lon"],
+  properties: {
+    lat: {type: "number"},
+    lon: {type: "number"}
+  }
+};
+
+// Define a custom component for handling the root position object
+class GeoPosition extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {...props.formData};
+  }
+
+  onChange(name) {
+    return (event) => {
+      this.setState({
+        [name]: parseFloat(event.target.value)
+      }, () => this.props.onChange(this.state));
+    };
+  }
+
+  render() {
+    const {lat, lon} = this.state;
+    return (
+      <div>
+        <input type="number" value={lat} onChange={this.onChange("lat")} />
+        <input type="number" value={lon} onChange={this.onChange("lon")} />
+      </div>
+    );
+  }
+}
+
+// Define the custom field component to use for the root object
+const uiSchema = {"ui:field": "geo"};
+
+// Define the custom field components to register; here our "geo"
+// custom field component
+const fields = {geo: GeoPosition};
+
+// Render the form with all the properties we just defined passed
+// as props
+render(<Form
+  schema={schema}
+  uiSchema={uiSchema}
+  fields={fields}/>);
+```
+
+Note: Registered fields can be reused accross the entire schema.
 
 ### Custom SchemaField
 
--- a/playground/app.js
+++ b/playground/app.js
@@ -28,6 +28,41 @@ const cmOptions = {
   tabSize: 2,
 };
 
+class GeoPosition extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {...props.formData};
+  }
+
+  onChange(name) {
+    return (event) => {
+      this.setState({
+        [name]: parseFloat(event.target.value)
+      }, () => this.props.onChange(this.state));
+    };
+  }
+
+  render() {
+    const {lat, lon} = this.state;
+    return (
+      <div>
+        <p>
+          <label>Latitude
+            <input type="number" value={lat} step="0.00001"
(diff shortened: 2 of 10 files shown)


n1k0:
Hmm this patch is wrong as we need to distinguish between fields and widgets; providing a replacement for an object field is definitely not the same as providing an input (widget) for eg. a string.

I might introduce a new `ui:field` to disambiguate the two distinct use cases.

n1k0:
This is now ready for early review. r=? @magopian

almet:
How do you do the correlation between "geo" and the component used here?

README.md
```diff
@@ -271,7 +272,12 @@ render(
 
 ## Advanced customization
 
-### Custom widgets
+The API allows to specify your own custom *widgets* and *fields* components:
+
+- A *widget* represents a HTML tag for the user to enter data, eg. `input`, `select`, etc.
```
almet: nit: should this be "an HTML tag" ?

README.md
```diff
@@ -271,7 +272,12 @@ render(
...
+- A *field* usually wraps one or more widgets and most often handle internal field state; think of a field as a form row, including the labels.
```
almet: nit: handles

README.md
```diff
@@ -342,6 +348,57 @@ The following props are passed to the widget component:
...
+
+class GeoPosition extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {...props.formData};
+  }
+
+  onChange(name) {
+    return (event) => {
+      this.setState({
```
almet: It might be a react idiom, but I don't understand this line. It might be useful to add some comments explaining why we need two argumens here and why we're passing only a partial state to `setState`.

playground/app.js
```diff
@@ -28,6 +28,41 @@ const cmOptions = {
...
+  constructor(props) {
+    super(props);
+    this.state = {...props.formData};
+  }
+
+  onChange(name) {
+    return (event) => {
+      this.setState({
+        [name]: parseFloat(event.target.value)
+      }, () => this.props.onChange(this.state));
```
almet: I don't understand why we need this?

almet: I don't understand why we need this?

playground/app.js
```diff
@@ -176,6 +211,7 @@ class App extends Component {
               formData={this.state.formData}
               onChange={data => this.setState({formData: data.formData})}
               onSubmit={data => this.setState({formData: data.formData})}
+              fields={{geo: GeoPosition}}
```
almet: So that's where the connection is made. The documentation should reflect that to be easier to grasp.

src/components/Form.js
```diff
@@ -119,12 +125,11 @@ if (process.env.NODE_ENV !== "production") {
     schema: PropTypes.object.isRequired,
     uiSchema: PropTypes.object,
     formData: PropTypes.any,
+    fields: PropTypes.object,
```
almet: Good move :)

src/components/fields/SchemaField.js
```diff
@@ -105,7 +115,10 @@ if (process.env.NODE_ENV !== "production") {
   SchemaField.propTypes = {
     schema: PropTypes.object.isRequired,
     uiSchema: PropTypes.object,
-    registry: PropTypes.object,
+    registry: PropTypes.shape({
+      widgets: PropTypes.object.isRequired,
+      fields: PropTypes.object.isRequired,
```
almet: Should we also check the shape of the `fields` object?

almet:
This is great! thanks for doing this, I think it will be a really useful addition to the library. Everything looks good to my non-react expert eyes.

n1k0:
> How do you do the correlation between "geo" and the component used here?

With this line from the docs sample:

```js
const uiSchema = {"ui:field": "geo"};
```

README.md
n1k0: Two lines below we're calling `this.onChange("lat")`, which returns a function, which is our event handler for this specific field. I wonder what kind of comment is needed here.

playground/app.js
n1k0: Why we need what?

playground/app.js
n1k0: Well, that's part of the example from the doc, and furthermore the registration is approached in the section before... Well, I'll see if I can improve it then.

src/components/fields/SchemaField.js
n1k0: No because it's not predictable (or at least PropTypes don't expose any API for ensuring a map is composed from a given typed data structure).

playground/app.js
n1k0: I'm realizing the README wasn't updated accordingly to this. Will fix.

src/components/fields/SchemaField.js
n1k0: Of course, I lied:

>  // An object with property values of a certain type
>  optionalObjectOf: React.PropTypes.objectOf(React.PropTypes.number),

n1k0: Addressed in 856a1b47fa15b404d10fd197ccf0f5207301a99a

README.md
almet: Previous comment was about the `this.setState` call, which is taking two arguments: I don't understand what the second argument is for.

playground/app.js
almet: Oh, ok, I think I got it. We first set the state and then we call `props.onChange`, probably to propagate the action.

I was mistaken by:

- The `setState` which isn't promise-based.
- The fact that we're calling `onChange` on the props, probably to cascade the changes.

playground/app.js
almet: Thanks!

README.md
n1k0: Oh, that callback is to ensure the next operation is performed *after* the new state has been taken into account and the call to render() being executed. That's important so we avoid races where the state is declared to be a and we propagate b.

playground/app.js
n1k0: :)

chmac:
@n1k0 Wow, I missed all the excitement. The change looks **very** cool. Don't have sufficient react / es6 experience to talk as to the code quality, so can only comment on the API.

On first pass, the `uiSchema` -> `fields` was not immediately obvious to me. Now I've read it again, I see the logic, and I like it. I can define a field once, and use it multiple times. I could even define the `fields` map globally and pass it to all my forms, even if all the field components are not necessarily used in every form I guess.

It would be great to add a list of the props to the docs. For example, the `this.state = {...props.formData}` line doesn't make a lot of sense unless you know what `formData` is. Something similar to the list in the widgets component would be ideal.

Bottom line, thanks a lot @n1k0, this change is a huge benefit for us. I think this may now be the react form package I've been dreaming of... :-)

Issue #80, opened 2016-03-22
Option to specify boolean labels?


chmac:
In some cases it would be nice to be able to provide labels instead of the strings `true` and `false`. The simplest example being `Yes` and `No` in English. I wonder if it would be practical to add this to the `uiSchema` definition? As far as I know, JSON schema doesn't provide this support.

n1k0:
There's the `enumNames` jsonschema property which allows defining labels for enums, though for booleans that'd mean having this property without an explicit `enum` defined (as it's implicit as `[true, false]`):

```js
const schema = {type: "boolean", enumNames: ["Yes", "No"]};
const uiSchema = {"ui:widget": "radio"};
```

Odd, and unlikely to be used in existing JSON schemas out there...

Having a new `ui:labels` uiSchema property seems to be the simplest way to go:

```js
const schema = {type: "boolean"};
const uiSchema = {
  "ui:widget": "radio", 
  "ui:labels": ["Yes", "No"]
};
```

We could also add support for options to widgets:

```js
const schema = {type: "boolean"};
const uiSchema = {
  "ui:widget": {
    name: "radio", 
    labels: ["Yes", "No"]
  }
};

The latter would require a much larger internal refactoring and would be probably slightly more cumbersome to write, though the feature might be useful to other widgets as well (including custom ones).

Thoughts?
```

n1k0:
Thinking of it a little more, I could probably live with support provided by using enumNames right from the schema. As a bonus, the patch is a three liner.

```js
const schema = {type: "boolean", enumNames: ["Yes", "No"]};
const uiSchema = {"ui:widget": "radio"};
```

n1k0:
@chmac quickly put a patch in #81 so we can stare at code. Would this be enough to you?

chmac:
@n1k0 Certainly looks like it would do exactly what we want. Being able to pass in the values, like Yes / No would cover most cases where we'd have thought about building our own `ui:widget`.

n1k0:
Fixed by #81.

PR #81, opened 2016-03-22
[Experiment] Ref #80: Add support for enumNames to boolean widgets.

n1k0:
WiP, experimental, do not merge. Refs #80.

Feedback=? @chmac

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,8 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
   * `select`: a select box with `true` and `false` as options;
   * by default, a checkbox is used
 
+> Note: To set the labels for a boolean field, instead of using `true` and `false` you can set `enumNames` in your schema. Note that `enumNames` belongs in your `schema`, not the `uiSchema`, and the order is always `[true, false]`.
+
 #### For `string` fields
 
   * `textarea`: a `textarea` element;
--- a/src/components/fields/BooleanField.js
+++ b/src/components/fields/BooleanField.js
@@ -3,6 +3,14 @@ import React, { PropTypes } from "react";
 import { defaultFieldValue, getAlternativeWidget, optionsList } from "../../utils";
 import CheckboxWidget from "./../widgets/CheckboxWidget";
 
+
+function buildOptions(schema) {
+  return optionsList(Object.assign({
+    enumNames: ["true", "false"],
+    enum: [true, false]
+  }, {enumNames: schema.enumNames}));
+}
+
 function BooleanField(props) {
   const {schema, name, uiSchema, formData, widgets, required, onChange} = props;
   const {title, description} = schema;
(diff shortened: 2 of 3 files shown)


chmac:
Can't really comment on the code, looks good to me...

The docs might be clearer with an example. Something like this?

> To set the labels for a boolean field, instead of using `true` and `false` you can set `enumNames` in your schema. Note that `enumNames` belongs in your `schema`, not the `ui:schema` and the order is `[true, false]`.

One minor thought, will the `enumNames` keyword be valid on fields of type `boolean`? My JSON schema fu is not strong enough to answer that definitively, and I couldn't find anything in the `enumNames` proposal to suggest either way...

n1k0:
> One minor thought, will the enumNames keyword be valid on fields of type boolean?

It's quite surprisingly "valid" when tested on http://jsonschemalint.com/draft4/, though I think we're probably bending the spec a little here.

I may be spend a little while figuring out how hard would be to add support for proper uiSchema widget options, which would sound definitely much cleaner to me.

chmac:
@n1k0 Could it be that arbitrary keys are allowed, but once v5 gets finalised the key might be only allowed in some situations? Anyway, if it works now, great, no harm, no foul!

n1k0:
Yeah, and my main concern is that uiSchema introduces a nice way of separating the data model from its presentation, and moving the labels definition there seems most definitely the right thing to do.

I'll probably implement the uiSchema widget option object thing.

n1k0:
Okay, I just had a look at the required changes to properly support widget options, it's too much work for what we're trying to achieve here. I think I'm gonna land this for now, and we'll revisit the way we handle widget options when we'll need them.

chmac:
Awesome, thanks a lot @n1k0. At this rate, I'll soon be your biggest fan! :-)

If you're ever passing by Berlin, beers are on @superlumen. :-)

n1k0:
Haha, noted! Cheers.

hiitskiran:
I know this is a very old post but, along with enumNames, can you also add support for enum for Boolean so we can specify in our schema what are the saved values? without enum support we have to transform the data back and forth from form so the values are saved as Y/N instead of true/false.

              "myCheckbox": {
                "type": "boolean",
                "enum" : ["Y","N"],
                "enumNames": ["Yes","No"],
                "title": "Enabled"
              },

epicfaace:
@hiitskiran in that case, you should use a string type:

```
          "myCheckbox": {
            "type": "string",
            "enum" : ["Y","N"],
            "enumNames": ["Yes","No"],
            "title": "Enabled"
          },
```

hiitskiran:
> @hiitskiran in that case, you should use a string type:
> 
> ```
>           "myCheckbox": {
>             "type": "string",
>             "enum" : ["Y","N"],
>             "enumNames": ["Yes","No"],
>             "title": "Enabled"
>           },
> ```

but we want to present in UI as a check box and checked value to store it as "Y" and unchecked "N"

epicfaace:
It might be best to make a custom widget in that case. otherwise, #1651 might be related to something you would need.

hiitskiran:
thanks Ashwin, but looking at BooleanField.js, it does look like taking "enum" from schema, but wonder why my enums are not taking into account and the values are still stored as true/false instead of enum values Y/N

    enumOptions = (0, _utils.optionsList)({
      "enum": schema["enum"] || [true, false],
      enumNames: schema.enumNames || (schema["enum"] && schema["enum"][0] === false ? ["no", "yes"] : ["yes", "no"])
    });

PR #83, opened 2016-03-22
Adds explicit spans for labels

alexeygolev:
React 15.0 doesn't add spans automatically which makes the current solution for labels a bit inconvenient


n1k0:
Looks good. Could you please quash your commit so I can land this?

n1k0:
Since #100 we now have the label string directly wrapped in a single `<label>` tag.

PR #84, opened 2016-03-23
FieldSet component w/ tests

zbyte64:
New FieldSet component which acts like Form but:
* no onSubmit
* no onError
* no default submit button
* renders as fieldset tag

Should this also be exposed as part of index.js?

### DIFF
--- a/src/components/FieldSet.js
+++ b/src/components/FieldSet.js
@@ -1,0 +1,66 @@
+import React, { Component, PropTypes } from "react";
+
+import SchemaField from "./fields/SchemaField";
+import TitleField from "./fields/TitleField";
+import {
+  toIdSchema
+} from "../utils";
+
+
+export default class FieldSet extends Component {
+  static defaultProps = {
+    uiSchema: {},
+    errorSchema: {}
+  }
+
+  getRegistry() {
+    // For BC, accept passed SchemaField and TitleField props and pass them to
+    // the "fields" registry one.
+    const _SchemaField = this.props.SchemaField || SchemaField;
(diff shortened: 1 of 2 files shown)


n1k0:
I'm surprised the `Form` component doesn't leverage this new `Fieldset` component at all... I'm not comfortable duplicating the shared logic here.

As an illustration, we recently landed support for error contextualization at the field level in #85 and your patch doesn't conflict at all, while it should definitely use the new feature. Another obvious illustration is the duplicated tests.

Could you find a way to share the common logic bits between the two components?

zbyte64:
Ended up making `Form` inherit from `FieldSet`. A couple things I am uncertain about:
* Should `FieldSet` include `renderErrors`; FieldSet does not call it
* Maybe both `FieldSet` and `Form` should extend a common class instead of `Form` extending `FieldSet`; not sure what it would be called

n1k0:
> Ended up making Form inherit from FieldSet. 

I was more expecting composition here. How about something like:

```js
class Form extends Component {
  constructor(props) {
    ...
  }

  render() {
    return (
      <form onSubmit={...}>
        <FieldSet {...props} />
      </form>
    );
  }
}
```

zbyte64:
Would that mean `Form` maintains its own `formData` state or would it just use ref (ie this.refs.fieldset.state) when someone triggers `onSubmit`?

n1k0:
I never worked with refs, so I don't know, but if it works like you suggest it does, why not! Otherwise, replicating state is probably unavoidable.

PR #90, opened 2016-03-24
Exclude Buffer from dist bundles.

n1k0:
This patch excludes -- or more precisely *mocks* -- the node `Buffer` object, so it doesn't bloat the generated dist files. `Buffer` is not used at all by the code (except in a check by the `deeper` dependency, hence the mock) so this operation should be safe.


n1k0:
Pushed the updated demo version to gh-pages, appears to be fully functional. I'm landing this.

Issue #99, opened 2016-03-25
Investigate and choose a CSS framework


magopian:
Following our decision to not support theming (see #98 and [rationale](https://github.com/mozilla-services/react-jsonschema-form/issues/91#issuecomment-201236773)), we also decided to use a CSS framework to have better styling.

Some options:
- bootstrap
- material-ui (see #39)
- zurb foundation

Once a solution has been chosen, use it, and document the fact that we're using it (and that it's thus very simple to re-use themes/css/js to change the look and feel, layout, behaviour...).

n1k0:
Bootsrtap has been selected.

Issue #87, opened 2016-03-23
Improve styling of errors by default


n1k0:

n1k0:
Fixed by #100.

Issue #57, opened 2016-03-11
Add a unique ID to widget props


chmac:
I'm trying to use a stateless widget like so:

```coffeescript
'ui:widget': (props) ->
  <input type="checkbox" id="ID"/><label htmlFor="ID">I agree</label>
```

The issue is generating the ID. I could obviously use a real component and do it inside the component. I could hope the component only loaded once (ouch). I could also probably hack the onChange into the onClick of the label. But all of those options aside, would it make sense for the package to generate a unique ID for each field and pass it in? `_.uniqueId()` maybe?

chmac:
OK, I see the error of my ways here. The widget doesn't render the `<label>`, it only renders the input. The label is rendered as well as whatever I supply for `ui:widget`. I guess the `id` question remains, even though my example no longer makes sense. My example does raise a different question, how I could replace a whole component, including the label, but that's for a separate issue...

n1k0:
I'm surprised your code snippet works; JSX requires to have a single root element, so you should use something like this:

```coffeescript
'ui:widget': (props) ->
  <div><input type="checkbox" id="ID"/><label htmlFor="ID">I agree</label></div>
```

For unique id generation, the issue is probably no persist them along the entire lifecycle of the component, so that means persisting this in comp state, or keeping a whole tree of nested ids in root component state.

To be noted, maybe #45 would help achieving what you're after here. 

Also, we don't have identified use cases for this in our current projects using this lib; so would you be willing working on implementing this?

n1k0:
>  I'm trying to use a stateless widget (...) I could hope the component only loaded once (ouch)

Stateless components are indeed fully rendered every time they receive props. If you want to "initialize" a component, eg with something generated once attached to it, you need state, hence you can't use stateless components.

n1k0:
We now have generated unique ids for forms ([docs](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#autogenerated-widget-ids)).

chmac:
@n1k0 Yet more awesomeness from y'all. Thanks.

PR #100, opened 2016-03-27
Switched to use Bootstrap semantics

n1k0:
Refs #99, #87.

/cc @magopian


n1k0:
Note that the new `ui:rootFieldId` should address #57.

n1k0:
I'm landing this with r=me.

PR #103, opened 2016-03-29
Fixes #102: Make live form validation an option.

### LINKED ISSUE #102: Make live form validation an option
And this option should be set to false by default, as it has a very large impact on performances with no real added value for end users.

n1k0:
## Live form data validation

By default, form data are only validated when the form is submitted or when a new `formData` prop is passed to the `Form` component.

You can enable live form data validation by passing a `liveValidate` prop to the `Form` component, and set it to `true`. Then, everytime a value changes within the form data tree (eg. the user entering a character in a field), a validation operation is performed, and the validation results are reflected into the form state.

Be warned that this is an expensive strategy, with possibly strong impact on performances.

r=? @magopian

### DIFF
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -500,29 +500,54 @@ describe("Form", () => {
         minLength: 8
       };
 
-      it("should update the errorSchema when the formData changes", () => {
-        const {comp, node} = createFormComponent({schema});
+      describe("Lazy validation", () => {
+        it("should not update the errorSchema when the formData changes", () => {
+          const {comp, node} = createFormComponent({schema});
 
-        Simulate.change(node.querySelector("input[type=text]"), {
-          target: {value: "short"}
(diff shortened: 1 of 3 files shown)


test/Form_test.js
```diff
@@ -500,29 +500,54 @@ describe("Form", () => {
...
+          Simulate.change(node.querySelector("input[type=text]"), {
+            target: {value: "short"}
+          });
+
+          expect(comp.state.errorSchema).eql({});
         });
 
-        expect(comp.state.errorSchema).eql({
-          errors: ["does not meet minimum length of 8"]
+        it("should not denote an error in the field", () => {
```
magopian: s/denote/list/ ?

magopian:
It's so cool that the patch is so small, thanks to some previous clever thinking ;)
You rock!

r+

PR #104, opened 2016-03-30
adds example for cdn users

danprince:
Found a [confused Stack Overflow user](http://stackoverflow.com/questions/36302120/using-react-component-from-js-source-maps/36303960#36303960) trying to access `Form` on the global namespace, figured this should clear up future ambiguities.


n1k0:
Thanks you!

PR #106, opened 2016-03-31
fixed bug of nested array: "add" button doesn't work for outer array

rhgb:
example schema:

```javascript
{
  "type": "object",
  "properties": {
    "listOfArrays": {
      "type": "array",
      "title": "A list of arrays",
      "items": {
        "type": "array",
        "title": "A list of numbers",
        "items": {
          "type": "number"
}}}}}
```

When clicking **Add** button of **listOfArrays**, nothing happens. That's because _default state_ of `type: "array"` is `[]` and `items.concat([])` has no effect. adding a pair of extra `[]` fixed this.

Also add `.idea` to `.gitignore` for convenience for WebStorm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 dist
 lib
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ node_modules
 build
 dist
 lib
-.idea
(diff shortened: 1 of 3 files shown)


n1k0:
Sounds reasonable. Could you please add a test?

.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 build
 dist
 lib
+.idea
```
n1k0: This should be added by people in their own `~/.gitignore`, please remove this from here.

rhgb:
ok, I'll remove the .gitignore modification and add a test.

rhgb:
I've added two tests for correctly rendering nested lists and adding inner list

n1k0:
Perfect! Thanks for your contribution, much appreciated.

PR #111, opened 2016-04-03
Add support for hidden widgets.

n1k0:
Supersedes #50, refs #43.

#### Hidden widgets

It's possible to use an hidden widget for a given field by setting the `ui:widget` uiSchema directive to `hidden` for this field:

```js
const schema = {
  type: "object",
  properties: {
    foo: {type: "boolean"}
  }
};

const uiSchema = {
  foo: {"ui:widget": "hidden"}
};
```

Note that hiding widgets is only supported for `boolean`, `string`, `number` and `integer` schema types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
         - [For boolean fields](#for-boolean-fields)
         - [For string fields](#for-string-fields)
         - [For number and integer fields](#for-number-and-integer-fields)
+        - [Hidden widgets](#hidden-widgets)
      - [Object fields ordering](#object-fields-ordering)
      - [Custom CSS class names](#custom-css-class-names)
      - [Custom labels for enum fields](#custom-labels-for-enum-fields)
@@ -188,6 +189,28 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
   * by default, a regular `input[type=text]` element is used.
 
 > Note: for numbers, `min`, `max` and `step` input attributes values will be handled according to JSONSchema's `minimum`, `maximium` and `multipleOf` values when they're defined.
+
+#### Hidden widgets
+
+It's possible to use an hidden widget for a given field by setting the `ui:widget` uiSchema directive to `hidden` for this field:
+
+```js
+const schema = {
+  type: "object",
+  properties: {
+    foo: {type: "boolean"}
+  }
+};
+
+const uiSchema = {
+  foo: {"ui:widget": "hidden"}
+};
+```
+
+> Notes
+>
+> - Hiding widgets is only supported for `boolean`, `string`, `number`, `integer` and `date-time` schema types;
+> - An hidden widget takes its value from the `formData` prop.
 
 ### Object fields ordering
 
--- a/src/components/widgets/HiddenWidget.js
+++ b/src/components/widgets/HiddenWidget.js
@@ -1,0 +1,31 @@
+import React, { PropTypes } from "react";
+
+
+function HiddenWidget({
+  id,
+  value,
+  defaultValue
+}) {
+  return (
+    <input type="hidden"
+      id={id}
+      value={value}
+      defaultValue={defaultValue} />
+  );
+}
+
+if (process.env.NODE_ENV !== "production") {
+  HiddenWidget.propTypes = {
+    id: PropTypes.string.isRequired,
+    value: PropTypes.oneOfType([
+      React.PropTypes.string,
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -189,6 +190,28 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
...
+};
+
+const uiSchema = {
+  foo: {"ui:widget": "hidden"}
+};
+```
+
+> Notes
+>
+> - Hiding widgets is only supported for `boolean`, `string`, `number` and `integer` schema types;
```
leplatrem: typo: are (?)

README.md
```diff
@@ -189,6 +190,28 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
...
+> - As hiding a field obviously prevents the user from entering data, the only way to set its value is to pass the form component a `formData` prop containing data for this field.
```
leplatrem: I think this is a bit too obvious. You could just note *it takes its value from the `formData` prop...*

BTW this reminds me about an hidden auto-datefield, whose value could set from `onSubmit` :)

test/uiSchema_test.js
```diff
@@ -1,8 +1,10 @@
+/*eslint no-unused-vars: [2, { "varsIgnorePattern": "^d$" }]*/
+
 import { expect } from "chai";
 import React from "react";
 import { Simulate } from "react-addons-test-utils";
 
-import { createFormComponent } from "./test_utils";
+import { createFormComponent, d } from "./test_utils";
```
leplatrem: Might deserve a comment no? Why is that with above eslint ?

test/uiSchema_test.js
```diff
@@ -135,6 +137,38 @@ describe("uiSchema", () => {
...
+        expect(node.querySelector("[type=hidden]").value)
+          .eql("a");
+      });
+
+      it("should map widget value to a typed state one", () => {
+        const {comp} = createFormComponent({schema, uiSchema, formData: {
+          foo: "a"
+        }});
+
+        expect(comp.state.formData.foo).eql("a");
```
leplatrem: Is there any specific code for that in `HiddenWidget`?

src/components/widgets/HiddenWidget.js
```diff
@@ -0,0 +1,39 @@
...
+  value,
+  defaultValue,
+  required
+}) {
+  return (
+    <input type="hidden"
+      id={id}
+      value={value}
+      defaultValue={defaultValue}
+      required={required} />
```
n1k0: This should be removed. Maybe we could add a test proving that validation works an hidden field is marked as required in the schema definition.

test/uiSchema_test.js
```diff
@@ -266,6 +300,38 @@ describe("uiSchema", () => {
...
+      it("should accept a uiSchema object", () => {
+        const {node} = createFormComponent({schema, uiSchema});
+
+        expect(node.querySelectorAll("[type=hidden]"))
+          .to.have.length.of(1);
+      });
+
+      it("should support formData", () => {
+        const {node} = createFormComponent({schema, uiSchema, formData: {
+          foo: "42"
```
n1k0: This should be defined as an actual number.

test/uiSchema_test.js
```diff
@@ -349,6 +415,38 @@ describe("uiSchema", () => {
...
```
n1k0: Ditto.

src/utils.js
```diff
@@ -14,20 +15,24 @@ const altWidgetMap = {
...
     updown: UpDownWidget,
     range: RangeWidget,
+    hidden: HiddenWidget,
   },
   integer: {
     updown: UpDownWidget,
     range: RangeWidget,
+    hidden: HiddenWidget,
   }
 };
```
n1k0: We should add the "date-time" jsonschema type in here so we can support hiding related widget as well.

README.md
n1k0: "Hiding" being a verb here, nope.

test/uiSchema_test.js
n1k0: `d` is a debugging helper logging the indented html version of a rendered node. We've been using this for months in the codebase.

test/uiSchema_test.js
n1k0: Nope but there's some in `SchemaField` which is what we're testing.

test/uiSchema_test.js
n1k0: Added the eslint rules to tests .eslintrc.

Issue #113, opened 2016-04-04
Add support for more jsonschema builtin string formats


n1k0:
From http://spacetelescope.github.io/understanding-json-schema/reference/string.html

The following is the list of formats specified in the JSON Schema specification.

-    "date-time": Date representation, as defined by RFC 3339, section 5.6.
-    "email": Internet email address, see RFC 5322, section 3.4.1.
-    "hostname": Internet host name, see RFC 1034, section 3.1.
-    "ipv4": IPv4 address, according to dotted-quad ABNF syntax as defined in RFC 2673, section 3.2.
-    "ipv6": IPv6 address, as defined in RFC 2373, section 2.2.
-    "uri": A universal resource identifier (URI), according to RFC3986.

n1k0:
`date-time` string format support has landed in #116.

n1k0:
Fixed by #113.

PR #116, opened 2016-04-04
Fixes #114: Added a widget for datetime string format.

### LINKED ISSUE #114: Add a DateTimeWidget component
UI could be simply 6 select inputs for setting year, month, day, hour, minutes and seconds values. Or alternatively a single text input with some supplementary bidirectional format transformation.

Note that the legit format received from and exposed to `formData` should always be the JSON compatible date format (eg. `new Date().toJSON()`).

n1k0:
WiP. Refs #114 and #113. I've realized that the `string` JSONSchema type allows specifying a `format` property, which can be `date-time`. Previously, we were erroneously considering this value as a *type*, so this patch removes support for it as a type and moves its support as a `format`.

It also introduces a very basic [`input[type=datetime]`](https://www.w3.org/TR/html-markup/input.datetime.html) widget which offers a specific UI on some platforms; by default, browsers not supporting it will render a simple text input. The good thing being the validator works in every situation when validating the value, which must follow the [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt), section 5.6 format. 

- [x] Implement
- [x] Test
- [x] Document
- [x] Add sample to the playground

Note: next steps will be to add support for more string formats (email, url, etc.) through #113.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,8 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
 
 The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following formats:
 
-- `date-time`: An `input[type=datetime]` will be rendered.
+- `date-time`: An `input[type=datetime-local]` will be rendered.
+- More formats will be supported in a near future, feel free to help us going faster!
 
 #### For `number` and `integer` fields
 
--- a/README.md
+++ b/README.md
@@ -182,6 +182,10 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
   * `password`: an `input[type=password]` element;
   * by default, a regular `input[type=text]` element is used.
 
+The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following formats:
+
+- `date-time`: An `input[type=datetime]` will be rendered.
+
 #### For `number` and `integer` fields
 
   * `updown`: an `input[type=number]` updown selector;
@@ -209,7 +213,7 @@ const uiSchema = {
 
 > Notes
 >
-> - Hiding widgets is only supported for `boolean`, `string`, `number`, `integer` and `date-time` schema types;
+> - Hiding widgets is only supported for `boolean`, `string`, `number` and `integer` schema types;
 > - An hidden widget takes its value from the `formData` prop.
 
 ### Object fields ordering
@@ -330,7 +334,6 @@ You can provide your own custom widgets to a uiSchema for the following json dat
 - `number`
 - `integer`
 - `boolean`
-- `date-time`
 
 ```jsx
 const schema = {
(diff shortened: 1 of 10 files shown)


n1k0:
r=? @magopian

src/components/widgets/DateTimeWidget.js
```diff
@@ -0,0 +1,37 @@
...
+  schema,
+  id,
+  placeholder,
+  value,
+  defaultValue,
+  required,
+  onChange
+}) {
+  return (
+    <input type="datetime"
```
leplatrem: It looks like the right type is `datetime-local`.
See https://developer.mozilla.org/en/docs/Web/HTML/Element/Input#Attributes and https://github.com/whatwg/html/issues/336

src/utils.js
```diff
@@ -33,17 +35,22 @@ const altWidgetMap = {
...
-  "date-time": {
-    hidden: HiddenWidget,
   }
 };
 
+const stringFormatWidgets = {
+  "date-time": DateTimeWidget,
+  "email": TextWidget, // XXX: to customize appropriately
+  "hostname": TextWidget,
+  "ipv4": TextWidget,
```
leplatrem: Did you have the `pattern` attribute in mind for those?

README.md
```diff
@@ -182,6 +182,10 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
   * `password`: an `input[type=password]` element;
   * by default, a regular `input[type=text]` element is used.
 
+The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following formats:
+
+- `date-time`: An `input[type=datetime]` will be rendered.
```
leplatrem: Should we remove the dash : `datetime` ?

leplatrem: We could mention something like ``` Other formats like email, password, color or url would be trivial to contribute ```

src/utils.js
```diff
@@ -33,17 +35,22 @@ const altWidgetMap = {
...
+  "ipv6": TextWidget,
+  "uri": TextWidget, // XXX: to customize appropriately
```
leplatrem: From the specs, interesting input types would be: color, password, search, tel, time

leplatrem:
r+wc :)

src/components/widgets/DateTimeWidget.js
n1k0: Indeed, I'll update that.

src/utils.js
n1k0: I don't know, as I don't really want to write patterns for these (which can be tedious), and as maintaining the json schema should be fully done by the user of the lib.

Also, I think we're pretty much covered by our *jsonschema* lib dependency, which validates these natively.

README.md
n1k0: > Should we remove the dash : `datetime` ?

Well, `date-time` is what's expected by the JSONSchema spec, so nope.

> Other formats like email, password, color or url would be trivial to contribute

Thi is actually my next followup PR ;)

src/utils.js
n1k0: Indeed!

PR #118, opened 2016-04-04
Refs #113: Added support for email and uri string formats.

n1k0:
Refs #113. This patch introduces support for JSONSchema `email` and `uri` string formats through dedicated widgets.

Note that  `hostname`, `ipv4` and `ipv6` formats are already handled by regular `TextWidget` widgets, and that their validation is already provided by our *jsonschema* dependency.

r=? @magopian @leplatrem


n1k0:
Landed because I could.

PR #121, opened 2016-04-04
Fixes #115: Added ui:help uiSchema directive.

### LINKED ISSUE #115: Add support for "ui:help" uiSchema directive
So we can render that help text next to the widget.

n1k0:
Refs #115.

### Help texts

Sometimes it's convenient to add some text next to a field to guide the end user filling it; this is the purpose of the `ui:help` uiSchema directive:

```js
const schema = {type: "string"};
const uiSchema = {
  "ui:widget": "password",
  "ui:help": "Hint: Make it strong!"
};
```

Help texts work for any kind of field at any level, and will always be rendered immediately below the field component widget(s), but after contextualized errors, if any.


n1k0:
I'm landing this.

PR #127, opened 2016-04-06
Added support for custom widget registry in nested schemas

minettiandrea:
I've noticed that the documented feature of adding external widgets was not working if there was a SchemaField or ObjectField.
I just added a test and propagated the `widgets` property through those objects.

Cheers

Andrea

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -110,7 +110,7 @@ export default class Form extends Component {
   }
 
   render() {
-    const {children, schema, uiSchema} = this.props;
+    const {children, schema, uiSchema, widgets} = this.props;
     const {formData, errorSchema, idSchema} = this.state;
     const registry = this.getRegistry();
     const _SchemaField = registry.fields.SchemaField;
@@ -124,6 +124,7 @@ export default class Form extends Component {
           idSchema={idSchema}
           formData={formData}
           onChange={this.onChange}
+          widgets={widgets}
           registry={registry}/>
         { children ? children :
           <p>
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -25,7 +25,7 @@ class ObjectField extends Component {
   }
 
   getStateFromProps(props) {
-    const {schema, formData, registry} = props;
+    const {schema, formData, registry, widgets} = props;
     return getDefaultFormState(schema, formData, registry.definitions) || {};
   }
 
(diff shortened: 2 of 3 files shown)


n1k0:
Custom widgets, fields are stored in `Form` *registry*, which is passed to all child components down the tree. That means `SchemaField` (including a custom one) will always receive that `registry` prop.

I'm drafting a patch to improve the docs on the matter, here's an excerpt:

A custom `SchemaField` will always be passed the following props:

 - `schema`: The JSON schema for this schema field;
 - `uiSchema`: The [uiSchema](#the-uischema-object) for this schema field;
 - `idSchema`: The tree of unique ids for every field in the schema;
 - `formData`: The form data for this schema field;
 - `errorSchema`: The tree of errors for every field in the schema;
 - `registry`: An object containing the registered custom fields and widgets as well as root schema definitions;
     - `widgets`: The [custom registered widgets](#custom-widget-components), if any;
     - `fields`: The [custom registered fields](#custom-field-components), if any;
     - `definitions`: The root schema definitions, if any.

Does this clarify usage a little?

minettiandrea:
Thanks for your quick answer.
The improved documentation doesn't alter FormComponent API so the problem persists, the test I've added is simply using the library as stated in https://github.com/mozilla-services/react-jsonschema-form#custom-widget-components with a slightly more complex schema and it fails

```
describe("Custom widgets", () => {
    
    const schema = {
      "type": "object",
      "properties": {
        "field": {
          "type": "string"
        }
      }
    }
    
    const uiSchema = {
      "field": {
        "ui:widget": "customWidget"
      }
    }
    
    const CustomWidget = (props) => {
      return (
        <input type="text"
          className="customWidget"
          value={props.value}
          defaultValue={props.defaultValue}
          required={props.required}
          onChange={(event) => props.onChange(event.target.value)} />
      );
    };

    const widgets: {
          customWidget: CustomWidget
    };
    
    it("should be rendered", () => {
      const comp = renderIntoDocument(
        <Form schema={ schema } uiSchema={ uiSchema } widgets={ widgets } />
      );
      const node = findDOMNode(comp);
      const input = node.querySelectorAll(".customWidget");
      expect(input.length).eql(1)

    });
  });
```

n1k0:
I've read your code too quickly, and you're most definitely right, we were relying on the asumption of having a `widgets` property always available to field props.

I've created #129 to fix this situation, don't hesitate to give it a look.

PR #129, opened 2016-04-06
Leverage registry for widgets.

n1k0:
Refs https://github.com/mozilla-services/react-jsonschema-form/pull/127#issuecomment-206434248

Note: While this was previously poorly documented, it's possible this breaks existing usage of the `widgets` props in custom field components. As we're in a 0.x version and we didn't see much adoption yet, this is probably not that much of a big deal if we properly document the change in the release notes.   

Feedback=? @minettiandrea
r=? @almet @magopian @leplatrem

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
   - [Advanced customization](#advanced-customization)
      - [Custom widget components](#custom-widget-components)
      - [Custom field components](#custom-field-components)
+        - [Field props](#field-props)
+        - [The registry object](#the-registry-object)
      - [Custom SchemaField](#custom-schemafield)
      - [Custom titles](#custom-titles)
   - [Live form data validation](#live-form-data-validation)
@@ -428,7 +430,7 @@ const uiSchema = {
 render(<Form
   schema={schema}
   uiSchema={uiSchema}
-  widgets={widgets}/>);
+  widgets={widgets} />);
 ```
 
 This is useful if you expose the `uiSchema` as pure JSON, which can't carry functions.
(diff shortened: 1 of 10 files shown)


src/components/fields/BooleanField.js
```diff
@@ -49,11 +55,17 @@ if (process.env.NODE_ENV !== "production") {
     onChange: PropTypes.func.isRequired,
     formData: PropTypes.bool,
     required: PropTypes.bool,
+    registry: PropTypes.shape({
+      widgets: PropTypes.objectOf(PropTypes.func).isRequired,
+      fields: PropTypes.objectOf(PropTypes.func).isRequired,
+      definitions: PropTypes.object.isRequired,
+    })
```
almet: Could we possibly define this somewhere and reuse it?

README.md
```diff
@@ -497,18 +499,39 @@ const fields = {geo: GeoPosition};
...
+
+#### The `registry` object
+
+The `registry` is an object containing the registered custom fields and widgets as well as root schema definitions.
+
+ - `fields`: The [custom registered fields](#custom-field-components). By default this object contains the standard `SchemaField` and `TitleField` components;
+ - `widgets`: The [custom registered widgets](#custom-widget-components), if any;
+ - `definitions`: The root schema [definitions](#schema-definitions-and-references), if any.
+
+The registry is passed down the component tree, so you can access it from your custom field and `SchemaField` components.
```
almet: Thanks for doing this :)

src/components/fields/BooleanField.js
n1k0: That'd be possible but we wouldn't benefit from the exclusion at build time (these are defined in an if statement removed when the env is `production`), as we would need to import these def and so so they would be part of the resulting bundle anyway.

It's no big deal having these copied, and it's convenient to have the exact proptypes at hand in each component file anyway.

leplatrem:
Didn't see any horror, but didn't get the point of this either, so r+! Better if an approval comes from someone else too :)

PR #132, opened 2016-04-11
Handle fixed items arrays

### LINKED ISSUE #76: Handle fixed items arrays
From the [spec](http://json-schema.org/latest/json-schema-validation.html#anchor37):

> The value of "items" MUST be either an object **or an array**. If it is an object, this object MUST be a valid JSON Schema. **If it is an array, items of this array MUST be objects, and each of these objects MUST be a valid JSON Schema.** 

```json
{
  "type": "array",
  "items": [
    {"type": "integer"},
    {"type": "boolean"},
    {"type": "string"}
  ]
}
```

Sample record:

```json
[
  -64586948,
  true,
  "dolor"
]
```

Note: This is a rather strange use case, and should be treated as P2.

rhgb:
fixed #76 
handle fixed items arrays:
```json
{
  "type": "array",
  "title": "List of fixed items",
  "items": [
    {
      "type": "string",
      "title": "Some text"
    },
    {
      "type": "number",
      "title": "A number"
    }
  ]
}
```
NOTE: According to the [spec](http://json-schema.org/latest/json-schema-validation.html#anchor131), it should be **valid** that form data has less items in fixed array than the schema defined, but it seems hard to find an appropriate UI for this. I'm glad to implement it if there's any reasonable suggestion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,6 +636,13 @@ This library partially supports [inline schema definition dereferencing]( http:/
 
 Note that it only supports local definition referencing, we do not plan on fetching foreign schemas over HTTP anytime soon. Basically, you can only reference a definition from the very schema object defining it.
 
+## JSON Schema supporting status
+
+This component follows [JSON Schema](http://json-schema.org/documentation.html) specs. Due to the limitation of form widgets, there are some exceptions as follows:
+
+* `additionalItems` keyword for arrays
+    This keyword works when `items` is an array. `additionalItems: true` is not supported because there's no widget to represent a item of any type. In this case it will be treated as no additional items allowed. `additionalItems` being a valid schema is supported.
+
 ## Troubleshooting
 
 ### Build error wrt missing "buffertools" module
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
   - [Live form data validation](#live-form-data-validation)
   - [Styling your forms](#styling-your-forms)
   - [Schema definitions and references](#schema-definitions-and-references)
+  - [JSON Schema supporting status](#json-schema-supporting-status)
   - [Troubleshooting](#troubleshooting)
      - [Build error wrt missing &quot;buffertools&quot; module](#build-error-wrt-missing-buffertools-module)
   - [Contributing](#contributing)
@@ -641,7 +642,7 @@ Note that it only supports local definition referencing, we do not plan on fetch
 This component follows [JSON Schema](http://json-schema.org/documentation.html) specs. Due to the limitation of form widgets, there are some exceptions as follows:
 
 * `additionalItems` keyword for arrays
-    This keyword works when `items` is an array. `additionalItems: true` is not supported because there's no widget to represent a item of any type. In this case it will be treated as no additional items allowed. `additionalItems` being a valid schema is supported.
+    This keyword works when `items` is an array. `additionalItems: true` is not supported because there's no widget to represent an item of any type. In this case it will be treated as no additional items allowed. `additionalItems` being a valid schema is supported.
 
 ## Troubleshooting
 
--- a/playground/samples/arrays.js
+++ b/playground/samples/arrays.js
@@ -24,15 +24,19 @@ module.exports = {
         title: "A list of fixed items",
         items: [
           {
-            title: "Item 1",
+            title: "A string value",
             type: "string",
             default: "lorem ipsum"
           },
           {
-            title: "Item 2",
-            type: "integer"
+            title: "a boolean value",
+            type: "boolean"
           }
-        ]
+        ],
+        additionalItems: {
+          title: "Additional item",
+          type: "number"
+        }
       },
       nestedList: {
         type: "array",
@@ -52,7 +56,7 @@ module.exports = {
   formData: {
     listOfStrings: ["foo", "bar"],
     multipleChoicesList: ["foo", "bar"],
-    fixedItemsList: ["Some text", 123],
+    fixedItemsList: ["Some text", true, 123],
     nestedList: [["lorem", "ipsum"], ["dolor"]]
   }
 };
--- a/playground/samples/arrays.js
+++ b/playground/samples/arrays.js
@@ -18,12 +18,41 @@ module.exports = {
           enum: ["foo", "bar", "fuzz"],
         },
         uniqueItems: true
+      },
+      fixedItemsList: {
+        type: "array",
+        title: "A list of fixed items",
+        items: [
+          {
+            title: "Item 1",
+            type: "string",
+            default: "lorem ipsum"
+          },
+          {
+            title: "Item 2",
+            type: "integer"
+          }
+        ]
+      },
+      nestedList: {
+        type: "array",
+        title: "Nested list",
+        items: {
+          type: "array",
+          title: "Inner list",
+          items: {
+            type: "string",
+            default: "lorem ipsum"
+          }
+        }
       }
     }
   },
   uiSchema: {},
   formData: {
     listOfStrings: ["foo", "bar"],
-    multipleChoicesList: ["foo", "bar"]
+    multipleChoicesList: ["foo", "bar"],
+    fixedItemsList: ["Some text", 123],
+    nestedList: [["lorem", "ipsum"], ["dolor"]]
   }
 };
--- a/playground/samples/arrays.js
+++ b/playground/samples/arrays.js
@@ -52,7 +52,17 @@ module.exports = {
       }
     }
   },
-  uiSchema: {},
+  uiSchema: {
+    fixedItemsList: {
+      items: [
+        {"ui:widget": "textarea"},
+        {"ui:widget": "select"}
+      ],
+      additionalItems: {
+        "ui:widget": "updown"
+      }
+    }
+  },
   formData: {
     listOfStrings: ["foo", "bar"],
     multipleChoicesList: ["foo", "bar"],
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -95,87 +95,54 @@ class ArrayField extends Component {
   };
 
   render() {
+    const {schema} = this.props;
+    if (isFixedItems(schema)) {
+      return this.renderFixedArray();
+    }
+    if (isMultiSelect(schema)) {
+      return this.renderMultiSelect();
+    }
+    return this.renderNormalArray();
+  }
+
+  renderNormalArray() {
     const {schema, uiSchema, errorSchema, idSchema, name} = this.props;
     const title = schema.title || name;
     const {items} = this.state;
-    const {fields, definitions} = this.props.registry;
-    const {SchemaField} = fields;
-    if (isFixedItems(schema)) {
-      const itemSchemas = schema.items.map(item => retrieveSchema(item, definitions));
-      const additionalSchema = allowAdditionalItems(schema) ? retrieveSchema(schema.additionalItems, definitions) : null;
-      return (
-          <fieldset
-              className={`field field-array field-array-fixed-items`}>
-            {title ? <legend>{title}</legend> : null}
-            {schema.description ?
-                <div className="field-description">{schema.description}</div> : null}
-            <div className="row array-item-list">{
-              itemSchemas.map((itemSchema, index) => {
-                const itemErrorSchema = errorSchema ? errorSchema[index] : undefined;
-                const itemIdPrefix = idSchema.id + "_" + index;
-                const itemIdSchema = toIdSchema(itemSchema, itemIdPrefix, definitions);
-                return (
-                    <div key={index}>
-                      <div className="col-xs-12">
-                        <SchemaField
-                            schema={itemSchema}
-                            uiSchema={uiSchema.items}
-                            formData={items[index]}
-                            errorSchema={itemErrorSchema}
-                            idSchema={itemIdSchema}
-                            required={this.isItemRequired(itemSchema)}
-                            onChange={this.onChangeForIndex(index)}
-                            registry={this.props.registry}/>
-                      </div>
-                    </div>
-                );
-              })
-            }{
-              additionalSchema && items && items.length <= itemSchemas.length ? null :
-                items.slice(itemSchemas.length).map((item, index) => {
-                  index = index + itemSchemas.length;
-                  const itemErrorSchema = errorSchema ? errorSchema[index] : undefined;
-                  const itemIdPrefix = idSchema.id + "_" + index;
-                  const itemIdSchema = toIdSchema(additionalSchema, itemIdPrefix, definitions);
-                  return (
-                      <div key={index}>
-                        <div className="col-xs-10">
-                          <SchemaField
-                              schema={additionalSchema}
-                              uiSchema={uiSchema.items}
-                              formData={items[index]}
-                              errorSchema={itemErrorSchema}
-                              idSchema={itemIdSchema}
-                              required={this.isItemRequired(additionalSchema)}
-                              onChange={this.onChangeForIndex(index)}
-                              registry={this.props.registry}/>
-                        </div>
-                        <div className="col-xs-2 array-item-remove text-right">
-                          <button type="button" className="btn btn-danger col-xs-12"
-                                  tabIndex="-1"
-                                  onClick={this.onDropIndexClick(index)}>Delete</button>
-                        </div>
-                      </div>
-                  );
-                })
-            }</div>
-            {
-              additionalSchema ? (
-                <div className="row">
-                  <p className="col-xs-2 col-xs-offset-10 array-item-add text-right">
-                    <button type="button" className="btn btn-info col-xs-12"
-                            tabIndex="-1" onClick={this.onAddClick}>Add</button>
-                  </p>
-                </div>
-              ) : null
-            }
-          </fieldset>
-      );
-    }
+    const {definitions} = this.props.registry;
     const itemsSchema = retrieveSchema(schema.items, definitions);
-    if (isMultiSelect(schema)) {
-      return (
-        <SelectWidget
+
+    return (
+      <fieldset
+          className={`field field-array field-array-of-${itemsSchema.type}`}>
+        {title ? <legend>{title}</legend> : null}
+        {schema.description ?
+            <div className="field-description">{schema.description}</div> : null}
+        <div className="row array-item-list">{
+          items.map((item, index) => {
+            const itemErrorSchema = errorSchema ? errorSchema[index] : undefined;
+            const itemIdPrefix = idSchema.id + "_" + index;
+            const itemIdSchema = toIdSchema(itemsSchema, itemIdPrefix, definitions);
+            return this.renderArrayFieldItem({
+              index, itemSchema: itemsSchema, itemIdSchema, itemErrorSchema,
+              itemData: items[index],
+              itemUiSchema: uiSchema.items
+            });
+          })
+        }</div>
+        <AddButton onClick={this.onAddClick}/>
+      </fieldset>
+    );
+  }
+
+  renderMultiSelect() {
+    const {schema, idSchema, name} = this.props;
+    const title = schema.title || name;
+    const {items} = this.state;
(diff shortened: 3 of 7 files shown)


src/utils.js
```diff
@@ -213,6 +216,14 @@ export function isMultiSelect(schema) {
   return Array.isArray(schema.items.enum) && schema.uniqueItems;
 }
 
+export function isFixedItems(schema) {
+  return Array.isArray(schema.items) && schema.items.length > 0 && schema.items.every(item => isObject(item));
+}
+
+export function allowAdditionalItems(schema) {
+  return isObject(schema.additionalItems);
+}
```
n1k0: As spec says that ["additionalItems" MUST be either a boolean or an object](http://json-schema.org/latest/json-schema-validation.html#anchor37).

We should therefore fallback on the the default `items` schema if `allowAdditionalItems` its `true`.

src/components/fields/ArrayField.js
```diff
@@ -93,6 +100,78 @@ class ArrayField extends Component {
...
+                            required={this.isItemRequired(itemSchema)}
+                            onChange={this.onChangeForIndex(index)}
+                            registry={this.props.registry}/>
+                      </div>
+                    </div>
+                );
+              })
+            }{
+              additionalSchema && items && items.length <= itemSchemas.length ? null :
+                items.slice(itemSchemas.length).map((item, index) => {
```
n1k0: Nit: for legibility, it would be nice to store the sliced items list into some `fixedItems` intermediary const.

src/utils.js
```diff
@@ -75,6 +75,7 @@ export function defaultTypeValue(schema) {
   case "array":     return [];
   case "boolean":   return false;
   case "number":    return 0;
+  case "integer":   return 0;
```
n1k0: Nit: please keep the alphabetical order here (and thanks for the fix ;))

test/ArrayField_test.js
```diff
@@ -203,4 +203,116 @@ describe("ArrayField", () => {
...
+          type: "string",
+          title: "Some text"
+        },
+        {
+          type: "number",
+          title: "A number"
+        }
+      ]
+    };
+    const schemaAdditional = {
```
n1k0: Nit: newline

test/ArrayField_test.js
```diff
@@ -203,4 +203,116 @@ describe("ArrayField", () => {
...
+          type: "number",
+          title: "Another number"
+        }
+      ],
+      additionalItems: {
+        type: "string",
+        title: "Additional item"
+      }
+    };
+    it("should render a fieldset", () => {
```
n1k0: Nit: newline

test/ArrayField_test.js
```diff
@@ -203,4 +203,116 @@ describe("ArrayField", () => {
...
+      Simulate.click(dropBtns[0]);
+      expect(node.querySelectorAll('.field-string')).to.have.length.of(1);
+      expect(comp.state.formData).eql([1, 2, "baz"]);
+
+      dropBtns = node.querySelectorAll(".array-item-remove button");
+      Simulate.click(dropBtns[0]);
+      expect(node.querySelectorAll('.field-string')).to.be.empty;
+      expect(comp.state.formData).eql([1, 2]);
+    });
+
```
n1k0: Nit: remove empty line

src/components/fields/ArrayField.js
```diff
@@ -93,6 +100,78 @@ class ArrayField extends Component {
     const {items} = this.state;
     const {fields, definitions} = this.props.registry;
     const {SchemaField} = fields;
+    if (isFixedItems(schema)) {
```
n1k0: This render method is getting huge. Could you please extract this into its own component? You can define it in the same file, as a non exported symbol.

src/utils.js
```diff
@@ -131,6 +132,8 @@ function computeDefaults(schema, parentDefaults, definitions={}) {
     // Use referenced schema defaults for this node.
     const refSchema = findSchemaDefinition(schema.$ref, definitions);
     defaults = computeDefaults(refSchema, defaults, definitions);
+  } else if (isFixedItems(schema)) {
+    defaults = schema.items.map(itemSchema => computeDefaults(itemSchema, undefined, definitions));
```
n1k0: This needs a test to be added in the `getDefaultFormState` test suite in `test/utils_test.js` (`getDefaultFormState` uses this function).

test/ArrayField_test.js
```diff
@@ -203,4 +203,116 @@ describe("ArrayField", () => {
...
+      });
+      Simulate.change(numInput, {
+        target: { value: "101" }
+      });
+
+      expect(comp.state.formData).eql(["bar", 101]);
+    });
+
+    it("should generate additional fields and fill data", () => {
+      const {node} = createFormComponent({schema: schemaAdditional, formData: [1, 2, "bar"]});
```
n1k0: Nit: please try to keep lines <= 80 columns.

test/ArrayField_test.js
```diff
@@ -203,4 +203,116 @@ describe("ArrayField", () => {
...
+
+      let dropBtns = node.querySelectorAll(".array-item-remove button");
+      Simulate.click(dropBtns[0]);
+      expect(node.querySelectorAll('.field-string')).to.have.length.of(1);
+      expect(comp.state.formData).eql([1, 2, "baz"]);
+
+      dropBtns = node.querySelectorAll(".array-item-remove button");
+      Simulate.click(dropBtns[0]);
+      expect(node.querySelectorAll('.field-string')).to.be.empty;
+      expect(comp.state.formData).eql([1, 2]);
```
n1k0: I usually prefer large tests like this to be split into many `it`, in a common `describe` section with setup done in a `beforeEach` statement. 

The idea with keeping a few assertions per test case is to ease identifying the failing assertion when a given test is failing.

n1k0:
This looks great. A few comment to address. Thanks a lot for contributing.

n1k0:
BTW I landed a fix for #135 so the playground now works and allow to run your example. You should probably rebase this PR on top of master to benefit from it.

rhgb:
Thanks for the comment: )

src/utils.js
rhgb: In the spec it says:
> 8.2.3.1.  If items is a schema, then the child instance must be valid against this schema, regardless of its index, and regardless of the value of "additionalItems".

Therefore it is only valid when `items` is an array, and when `additionalItems` is `true` it implies **any value** of additional items is valid. IMO falling back to `items` or `items[0]` are not best choices. (Maybe we need a input widget for "any" or "raw"?)

n1k0: > Maybe we need a input widget for "any" or "raw"?

Well, that's quite impossible, even if we decide to render a textarea it's unlikely to be able to reflect *any* JSON data structure, so we're stuck here.

Maybe we should simply raise a warning in such a case, and document this limitation in an appropriate section of the documentation? What do you think?

rhgb: You are right, maybe that's the best way currently. For the detail should we display a message in the error list, or just showing a warning in the console?

src/components/fields/ArrayField.js
rhgb: I also realize that FixedArrayField and normal ArrayField has many lines in common. I think I'm going to split `render()` into several stateless components, `ArrayFieldItem` `NormalArrayField` `FixedArrayField` etc.

src/utils.js
n1k0: > For the detail should we display a message in the error list, or just showing a warning in the console?

A warning in the console sounds good enough to me. I'd like to keep error reporting for strict jsonschema validation stuff, not lib limitations.

src/components/fields/ArrayField.js
n1k0: Sounds good :+1:

rhgb:
Fixed issues above. For the case `additionalItems: true`, a warning message will be printed to console, and some explanations are added to README.md

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
+        {title ? <legend>{title}</legend> : null}
+        {schema.description ?
+            <div className="field-description">{schema.description}</div> : null}
+        <div className="row array-item-list">{
+          items.map((item, index) => {
+            const itemErrorSchema = errorSchema ? errorSchema[index] : undefined;
+            const itemIdPrefix = idSchema.id + "_" + index;
+            const itemIdSchema = toIdSchema(itemsSchema, itemIdPrefix, definitions);
+            return this.renderArrayFieldItem({
+              index, itemSchema: itemsSchema, itemIdSchema, itemErrorSchema,
```
n1k0: Nit: Each key assignment on its own line please.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
+  }
 
+  renderFixedArray() {
+    const {schema, uiSchema, errorSchema, idSchema, name} = this.props;
+    const title = schema.title || name;
+    const {items} = this.state;
+    const {definitions} = this.props.registry;
+    const itemSchemas = schema.items.map(item => retrieveSchema(item, definitions));
+    const additionalSchema = allowAdditionalItems(schema) ? retrieveSchema(schema.additionalItems, definitions) : null;
+    const additionalItems = additionalSchema && items && (items.length > itemSchemas.length) && items.slice(itemSchemas.length);
```
n1k0: Nit: Please try to keep lines <= 80 columns. Create intermediary consts if that helps.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
-        <div className="row">
-          <p className="col-xs-2 col-xs-offset-10 array-item-add text-right">
-            <button type="button" className="btn btn-info col-xs-12"
-              tabIndex="-1" onClick={this.onAddClick}>Add</button>
-          </p>
-        </div>
+        {
+          additionalSchema ? (
+            <AddButton onClick={this.onAddClick}/>
+          ) : null
```
n1k0: Nit: This can probably fit on a single line.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
+        {
+          removable ?
+            <div className="col-xs-2 array-item-remove text-right">
+              <button type="button" className="btn btn-danger col-xs-12"
+                      tabIndex="-1"
+                      onClick={this.onDropIndexClick(index)}>Delete</button>
+            </div>
+          : null
+        }
+      </div>
```
n1k0: I'd really like to see these extracted and moved to their own stateless components.

n1k0:
Looks great. Do you think it could be possible to extract rendered JSX from comp methods into their own separate stateless components?

rhgb:
I'll have a try. But it seems these components will have 7 or 8 props to be passed in. I'm not sure whether this pattern is good or not. What's your suggestion?

n1k0:
> But it seems these components will have 7 or 8 props to be passed in

I see what you mean, this could easily turn verbose and messy. Plus component methods are certainly convenient as they have natural immediate access to the prototype.

I tend to agree with the current approach, even if it bloats a little the whole `ArrayField` component; though that's expected as we're now covering a lot of use cases.

I gonna do a second pass over the diff, but I have a feeling we are very close from having our final patch.

README.md
```diff
@@ -636,6 +636,13 @@ This library partially supports [inline schema definition dereferencing]( http:/
 
 Note that it only supports local definition referencing, we do not plan on fetching foreign schemas over HTTP anytime soon. Basically, you can only reference a definition from the very schema object defining it.
 
+## JSON Schema supporting status
```
n1k0: You should run the `npm run build:readme` so this entry will be added to the TOC.

README.md
```diff
@@ -636,6 +636,13 @@ This library partially supports [inline schema definition dereferencing]( http:/
...
+
+This component follows [JSON Schema](http://json-schema.org/documentation.html) specs. Due to the limitation of form widgets, there are some exceptions as follows:
+
+* `additionalItems` keyword for arrays
+    This keyword works when `items` is an array. `additionalItems: true` is not supported because there's no widget to represent a item of any type. In this case it will be treated as no additional items allowed. `additionalItems` being a valid schema is supported.
```
n1k0: Nit: *an* item.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
+    const title = schema.title || name;
+    const {items} = this.state;
+    const {definitions} = this.props.registry;
+    const itemSchemas = schema.items.map(item => retrieveSchema(item, definitions));
+    const additionalSchema = allowAdditionalItems(schema) ? retrieveSchema(schema.additionalItems, definitions) : null;
+    const additionalItems = additionalSchema && items && (items.length > itemSchemas.length) && items.slice(itemSchemas.length);
     return (
       <fieldset
-        className={`field field-array field-array-of-${itemsSchema.type}`}>
+          className="field field-array field-array-fixed-items">
```
n1k0: Nit: unnecessary indent.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
         {title ? <legend>{title}</legend> : null}
         {schema.description ?
-          <div className="field-description">{schema.description}</div> : null}
+            <div className="field-description">{schema.description}</div> : null}
```
n1k0: Ditto.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
-                </div>
-                <div className="col-xs-2 array-item-remove text-right">
-                  <button type="button" className="btn btn-danger col-xs-12"
-                    tabIndex="-1"
-                    onClick={this.onDropIndexClick(index)}>Delete</button>
-                </div>
-              </div>
-            );
+            const itemIdSchema = toIdSchema(itemSchema, itemIdPrefix, definitions);
+            const itemUiSchema = (uiSchema.items instanceof Array) ? uiSchema.items[index] : uiSchema.items || {};
```
n1k0: Nit: `Array.isArray`

n1k0: Side note regarding uiSchema handling; were you able to define a custom widget for additional items? I didn't. This is probably something we'll want to address, either in this patch or as a followup.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
+            return this.renderArrayFieldItem({
+              index, itemSchema, itemIdSchema, itemErrorSchema,
```
n1k0: Nit: one property per line please.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,157 @@ class ArrayField extends Component {
...
+          : null
+        }
+      </div>
+    );
+  }
+}
+
+function AddButton({onClick}) {
+  return (
+      <div className="row">
```
n1k0: Nit: 2 spaces indentation please.

src/utils.js
```diff
@@ -213,6 +216,17 @@ export function isMultiSelect(schema) {
   return Array.isArray(schema.items.enum) && schema.uniqueItems;
 }
 
+export function isFixedItems(schema) {
+  return Array.isArray(schema.items) && schema.items.length > 0 && schema.items.every(item => isObject(item));
```
n1k0: Please try to keep line length under 80c.

n1k0:
All this a really good, thanks. r+ with remaining nits addressed.

rhgb:
Done fixing, besides I've rewritten `renderFixedArray()` so it generates all fields in one pass.
For the `uiSchema` I've added an sample, please see the diff for detail.

playground/samples/arrays.js
```diff
@@ -18,12 +18,55 @@ module.exports = {
...
+  uiSchema: {
+    fixedItemsList: {
+      items: [
+        {"ui:widget": "textarea"},
+        {"ui:widget": "select"}
+      ],
+      additionalItems: {
+        "ui:widget": "updown"
       }
     }
```
rhgb: Here's example of defining ui for fixed items and additional items

n1k0: Woah, super nice.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,165 @@ class ArrayField extends Component {
...
-          title={title}
-          value={items}
-        />
-      );
-    }
 
     return (
       <fieldset
-        className={`field field-array field-array-of-${itemsSchema.type}`}>
+          className={`field field-array field-array-of-${itemsSchema.type}`}>
```
n1k0: 2 spaces indent.

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,165 @@ class ArrayField extends Component {
...
         {title ? <legend>{title}</legend> : null}
         {schema.description ?
-          <div className="field-description">{schema.description}</div> : null}
+            <div className="field-description">{schema.description}</div> : null}
```
n1k0: 2 spaces indent.

src/components/fields/ArrayField.js
rhgb: missed that:P

src/components/fields/ArrayField.js
```diff
@@ -88,68 +94,165 @@ class ArrayField extends Component {
...
+    let {items} = this.state;
+    const {definitions} = this.props.registry;
+    const itemSchemas = schema.items.map(item =>
+      retrieveSchema(item, definitions));
+    const additionalSchema = allowAdditionalItems(schema) ?
+      retrieveSchema(schema.additionalItems, definitions) : null;
+
+    if (!items || items.length < itemSchemas.length) {
+      items = items || [];
+      items = items.concat(new Array(itemSchemas.length - items.length));
```
n1k0: Smart :)

n1k0:
From my testing, fixed array field error contextualization doesn't work, eg. with this schema:

```
    "nestedList": {
      "type": "array",
      "title": "Nested list",
      "items": {
        "type": "array",
        "title": "Inner list",
        "items": {
          "type": "string",
          "default": "lorem ipsum",
          "maxLength": 3
        }
      }
```

Errors are only listed at the top of the form, meaning we're probably not properly passing the errorSchema object for these fields.

rhgb:
I'll have a look on the errorSchema issue.

rhgb:
Seems it's a bug of error string parsing.
I've made some fix, but this whole approach is vulnerable to carefully picked names e.g.
 `{ "someName[0]": { /* schema */ } }`

n1k0:
> this whole approach is vulnerable to carefully picked names

Personally I can live with it, but I wonder if you had a better parsing strategy in mind?

Also, it seems we should add some unit tests to cover the new cases handled by this last change.

rhgb:
I'll add some test for that.

In my opinion maybe we should let components maintain their own error status. I'm not very experienced in react so I'd like to hear your suggestion. Also there seems to be a lot of work, maybe we discuss it later in another thread?

n1k0:
> In my opinion maybe we should let components maintain their own error status

My initial idea was to prevent field components to deal with state at all; of course it's not possible for arrays and objects, as they both carry one while the user is "building" them -- which kinda defeats the purpose admittedly. For errors, validating at the field level may get harder are you often need the parent schema eg. for required or unique fields, but that's just plain theory from me.

Feel free to file an issue if you think there's a better way :)

n1k0:
@rhgb Sorry I had to land #137 and issue a v0.24.0 asap, so your patch is now conflicting. Just tell me if you prefer me to take over the branch if you don't have the time to rebase its current state on top of latest master, I'll happily do that.

rhgb:
It's fine, I'll rebase it myself and add tests we've discussed before. Just have too much other work to do yesterday, didn't finish those tests yet.

rhgb:
I've done rebasing master and modified some test to fit new changes of #139. Tests for error string parsing also added.

n1k0:
Fantastic work, thanks a lot for your work!

PR #138, opened 2016-04-13
Fixes #130: Live validation not fully disabled by default.

### LINKED ISSUE #75: Avoid reformatting entered JSON in the playground
This "feature" makes it really hard to tweak the JSON data in the editors.

### LINKED ISSUE #130: Add support for disabling all validation until form submit
### Prerequisites

### Description

If you pass down formData from a parent component to the form, props can update when the user is editing the form and, [according to the docs](https://github.com/mozilla-services/react-jsonschema-form#user-content-live-form-data-validation) (and real life experience), make liveValidate=false futile since;

> By default, form data are only validated when the form is submitted **or when a new formData prop is passed to the Form component**.

There should be a way to only validate the form onSubmit, so maybe add propValidate=true (or a similarly but better named prop) if I actually want the form to validate automatically when props change.

### Version

`"react-jsonschema-form": "^0.23.0"`

n1k0:
Refs #130 for context. r=? @leplatrem /cc @matiaslarsson

This also nicely fixes #75.

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -324,6 +324,7 @@ class App extends Component {
         <div className="col-sm-5">
           {!this.state.form ? null :
             <Form
+              liveValidate
               schema={schema}
               uiSchema={uiSchema}
               formData={formData}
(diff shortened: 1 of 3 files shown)


leplatrem:
r+ :)

PR #139, opened 2016-04-13
Fixes #137: Prevent non-required fields to fill state with defaults.

### LINKED ISSUE #137: non required number type field starts off with a value of 0
### Description

When a field has a type number and is not required, the starting value for the field is 0. Since 0 is possibly a legitimate value, shouldn't this field be empty by default instead?

Here is the output for JSON.stringify(props):
```
{"name":"temperature",
"required":false,
"schema":{"description":"Temperature of the asset in CELSIUS.","type":"number"},
"idSchema":{"id":"root_temperature"},
"formData":0,
"registry":{"fields":{},"widgets":{},"definitions":{}}}
```

#### Expected behavior

The field should start off as an empty field.

#### Actual behavior

The field starts off with a value of 0.

### Version

"react-jsonschema-form": "^0.21.0"

n1k0:
Refs #137, /cc @alexmnguyen.

We should carefully review this one at is has possible strong impact on forms behavior, though couldn't spot any problem in tests and in the playground.

r=? @leplatrem @magopian

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -66,28 +66,8 @@ export function getDefaultRegistry() {
   };
 }
 
-export function defaultTypeValue(schema) {
-  if (typeof schema.default !== "undefined") {
-    return schema.default;
-  }
-  const {type} = schema;
-  switch (type) {
-  case "array":     return [];
-  case "boolean":   return false;
-  case "number":    return 0;
-  case "object":    return {};
-  case "string":    {
-    if (schema.format === "date-time") {
-      return new Date().toJSON();
-    }
-    return "";
-  }
-  default:        return undefined;
-  }
-}
-
 export function defaultFieldValue(formData, schema) {
-  return typeof formData === "undefined" ? defaultTypeValue(schema) : formData;
+  return typeof formData === "undefined" ? schema.default : formData;
 }
 
 export function getAlternativeWidget(schema, widget, registeredWidgets={}) {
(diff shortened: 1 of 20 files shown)


n1k0:
I'm gonna move the default definition for datetime into the widget itself

n1k0:
Deployed the changes to gh-pages http://mozilla-services.github.io/react-jsonschema-form/

src/utils.js
```diff
@@ -66,28 +66,8 @@ export function getDefaultRegistry() {
...
-  }
-  const {type} = schema;
-  switch (type) {
-  case "array":     return [];
-  case "boolean":   return false;
-  case "number":    return 0;
-  case "object":    return {};
-  case "string":    {
-    if (schema.format === "date-time") {
-      return new Date().toJSON();
```
leplatrem: That particular one is a breaking change, no ?

How do we get date-time fields to default to current date ? It should take the current datetime, not the one when the app/schema was instantiated.

leplatrem:
Sorry I hadn't seen your comment about `parseDateString`.

leplatrem:
it's a pity that we're obliged to repeat the ``sandbox.stub(console, "error", (error)`` in every test suite.

r+ otherwise :)

(as long as the default datetime is still «now»)

n1k0:
> it's a pity that we're obliged to repeat the sandbox.stub(console, "error", (error) in every test suite.

Factored this.

I'm gonna land the patch after #132 is landed, as it probably conflicts heavily.

n1k0:
I need to land this now and make a release as this fixes a bunch of issues we need on other projects.

Issue #142, opened 2016-04-17
Seperating Date and DateTime widgets.


plasmashadow:
Instead of having DateTime widget. Can you guys separate out date and date-time widgets. The reason i am saying this is because if i want to get the date of birth at that moment date-time widgets seems to be not meaningful. 

Or is there any way i can get only date widget?

n1k0:
Makes sense, will do.

plasmashadow:
+1

plasmashadow:
It seems when used {format: "date"} it causing some problems. I am getting
``` Uncaught Error: No alternative widget "date" for type string/date```. Should i use format name to some other than "date".

PR #143, opened 2016-04-18
Add a DateWidget.

n1k0:
refs #142, r=? @leplatrem

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
 The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following formats:
 
 - `date-time`: Six `select` elements are used to select the year, the month, the day, the hour, the minute and the second;
+    * If you don't want to deal with time, a `"ui:widget": "date"` uiSchema widget is alternatively available, exposing three selects for year, month and day only;
 - `email`: An `input[type=email]` element is used;
 - `uri`: An `input[type=url]` element is used;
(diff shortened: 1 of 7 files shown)


src/utils.js
```diff
@@ -312,7 +314,7 @@ export function toIdSchema(schema, id, definitions) {
   return idSchema;
 }
 
-export function parseDateString(dateString) {
+export function parseDateString(dateString, includeTime = true) {
```
leplatrem: Nit: wouldn't it better to have `parseDateString()` and `parseDatetimeString()` instead ?

README.md
```diff
@@ -217,7 +217,7 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
 
 The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following formats:
 
-- `date-time`: Six `select` elements are used to select the year, the month, the day, the hour, the minute and the second;
+- `date-time`: Six `select` elements are used to select the year, the month, the day, the hour, the minute and the second; If you don't want to deal with time, a *date* `ui:widget` is also available;
```
leplatrem: nit: `date` instead of *date* ?

leplatrem:
Looks good to me :)

src/utils.js
n1k0: I tried but it was weird; we're really *always* parsing datetimes here (the JSON date format always exposes hours, minutes and seconds), and it's just an export option I'm adding here.

PR #148, opened 2016-04-19
Do not render current date by default.

n1k0:
Atm we render date widgets with current date by default, but this rendered value isn't populated to state; so it needs user interaction for the current date to be taken into account. We need to fix that.

We should allow the user to leave the field blank when a date field isn't required. Basically right now rendered widgets are decorellated from actual form data state, which is confusing and likely to introduce mistakes.

So I think we should add new blank `option`s to every date widget selects, so by default we:

- allow users to leave these blank, so no value at all is assigned to the formData associated with the field (useful when the field is *not* required);
- allow users to explicitly start entering the actual value they want, and reflect this value to state.

We might also consider adding "Now" links to allow easily populate these fields to the current date/time with a single explicit click, as well as a "Clear" button to allow resetting the widget to an undefined value.

Later on, probably when we'll get widget options, we shall add an option to allow setting the value to the current date by default. Alternatively we could add new `AutoDateWidget` and `AutoDateTimeWidget` with appropriate aliases to use for `ui:widget`, eg. `auto-date` and `auto-datetime`.

For now one will have to click the newly introduced "Now" button.

### DIFF
--- a/playground/samples/widgets.js
+++ b/playground/samples/widgets.js
@@ -3,6 +3,28 @@ module.exports = {
     title: "Widgets",
     type: "object",
     properties: {
+      stringFormats: {
+        type: "object",
+        title: "String formats",
+        properties: {
+          email: {
+            type: "string",
+            format: "email"
+          },
+          uri: {
+            type: "string",
+            format: "uri"
+          },
+          datetime: {
+            type: "string",
+            format: "date-time"
+          },
+          date: {
+            type: "string",
+            format: "date-time"
+          }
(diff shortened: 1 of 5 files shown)


n1k0:
This is now ready for review; r=@leplatrem

kaihendry:
Be good to see some screenshots

n1k0:
> Be good to see some screenshots

n1k0:
I'm gonna move forward and land this with r=me.

PR #149, opened 2016-04-19
Add autocomplete and combocomplete - add doc

cifren:
Hi, I added some modification for an autocomplete widget, might be you'd like it !

I added some dependencies, grunt / lodash / awesomplete, it should be ready to go.

There is no test, I am sorry... Don't really have the courage to start learning how to do test on Javascript. I will try to do them next :)

Let me know if you need some help or explanation. Those modification come from my project https://github.com/cifren/sound-admin, it is working on it, I will try display those fields on the demo.

I choose awesome because it seems to be something light and easy to understand with great feature, not as good as jQuery-ui on some point, but lighter for sure and easier to integrate. The API http://leaverou.github.io/awesomplete/

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,16 @@
+module.exports = function (grunt) {
+    grunt.initConfig({
+        "pkg": grunt.file.readJSON('package.json'),
+        copy: {
+          main: {
+            files: [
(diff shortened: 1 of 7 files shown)


n1k0:
I think these shouldn't be part of this lib core, but rather part of their own repository and provided as optional installable custom widgets.

cifren:
I agree that is more a plugin widget than a core widget :)

I proposed it just in case.

Thanks the answer.

PR #152, opened 2016-04-20
Attempt at providing memoization for stateless components.

n1k0:
This is an experimental attempt at addressing #147 by memoizing stateless component widgets, hopefully for faster rerenderings.

Issue #147, opened 2016-04-19
Slow performance 


kaihendry:
### Description

Love the promise of your project. We tried our schema of the data we need to collect for our Web application and we were disappointed with the performance when filling in text fields. We have a form with several long enums, think countries, states and currencies that we think trigger the issue. Perhaps there is a better approach we missed?

### Steps to Reproduce

You can see here performance gets progressively worse: 
http://s3.jptdev.com/2016-04-19/react-json-test-5x50.html
http://s3.jptdev.com/2016-04-19/react-json-test-15x50.html
http://s3.jptdev.com/2016-04-19/react-json-test-30x50.html

(We are not using live validation)

#### Expected behavior

A fast and nimble UX

#### Actual behavior

Slow to type in the text area.

http://s.natalian.org/2016-04-19/TimelineRawData-20160419T155117.json
http://s.natalian.org/2016-04-19/1461052305_2558x1404.png

### Version

Er your https://npmcdn.com/react-jsonschema-form@0.25.0/dist/react-jsonschema-form.js

n1k0:
I've played a little with your last example [in jsfiddle](https://jsfiddle.net/n1k0/8uL37fet/), and removing console logging has improved the situation a little.

I agree there's still room for improvement, while we have already applied some performance optimization strategies... I'll investigate this later, and would warmly welcome any expertise on the matter.

n1k0:
@kaihendry could you please bring an update on the performance when disabling console logging? I've been toying with large forms lately and performance is rather acceptable when livValidate is disabled and onChange listener not leveraged at all.

n1k0:
Side note, as we're mostly using stateless components in this lib and they're [not memoized just yet](https://github.com/facebook/react/issues/5677), we may suffer from it pretty badly here.

kaihendry:
<img src=http://s.natalian.org/2016-04-20/1461138658_2558x1404.png>

Still takes seconds for me to update a text field:
http://s3.jptdev.com/2016-04-20/147.html

Sorry, I'm a bit of ReactJS newbie atm... how or where do I try shouldComponentUpdate ? Do you have an example?

n1k0:
> Sorry, I'm a bit of ReactJS newbie atm... how or where do I try shouldComponentUpdate ? Do you have an example?

That's not something you have to do, rather it's something we should investigate as an implementation solution.

n1k0:
[...]

PR #153, opened 2016-04-20
Improved async rendering strategy

n1k0:
This supersedes #152 trying to address #147.

This patch updates our approach for dealing with asynchronous state propagation from field components to their parent; previously we were waiting for setState issued renderings to be fully processed before starting propagating changes upstream. This new approach brings a perceivable performance boost as we're now propagating nearly immediately making the UI much more responsive; definitely an improvement while we're probably not entirely done in the performances improvement area.

I had to update our tests accordingly, hence introducing a new `SimulateAsync` test helper for ensuring that we wait an appropriate amount of time before checking for async rendes to be fully rendered after DOM events triggering them have been fired.

/cc @kaihendry

r=? @leplatrem

### DIFF
--- a/playground/samples/index.js
+++ b/playground/samples/index.js
@@ -7,6 +7,7 @@ import ordering from "./ordering";
 import references from "./references";
 import custom from "./custom";
 import errors from "./errors";
+import large from "./large";
 
 export const samples = {
   Simple: simple,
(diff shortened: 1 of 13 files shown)


n1k0:
As always I'm expecting feedback from users here /poke @kaihendry :)

n1k0:
Note: I've pushed this patch to the demo at http://mozilla-services.github.io/react-jsonschema-form/, where you can find a new *Large* section involving some bulky schema.

n1k0:
I'm gonna move forward and land this with r=me.

PR #158, opened 2016-04-22
Fixes #150: Native date time widgets.

### LINKED ISSUE #150: "ui:widget": "date" to native input type=date ?
### Description

The date widget imo seems like a steps backwards in usability.

Not sure how to use the native https://html.spec.whatwg.org/multipage/forms.html#date-state-(type=date) instead.

### Steps to Reproduce

1. http://s3.jptdev.com/2016-04-20/date.html
2. Marvel at dropdown selectors to choose a date
3. Wonder if it's possible to use type=date instead

#### Expected behavior

Use the native browser date picker.
http://caniuse.com/#feat=input-datetime

#### Actual behavior

Drop down select boxes

### Version

https://npmcdn.com/react-jsonschema-form@0.25.0/dist/react-jsonschema-form.js

n1k0:
WiP, refs #150 /cc @kaihendry 

This adds new native date and datetime widgets which are now used by default, and demotes previous default select-based date widgets as alternative ones.

#### Todo

- [x] Update docs
- [x] Write tests ~~(hard as jsdom doesn't have a clue what these inputs are)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
      - [Alternative widgets](#alternative-widgets)
         - [For boolean fields](#for-boolean-fields)
         - [For string fields](#for-string-fields)
+           - [String formats](#string-formats)
         - [For number and integer fields](#for-number-and-integer-fields)
         - [Hidden widgets](#hidden-widgets)
      - [Object fields ordering](#object-fields-ordering)
@@ -215,13 +216,22 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
   * `password`: an `input[type=password]` element is used;
   * by default, a regular `input[type=text]` element is used.
 
-The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following formats:
+##### String formats
 
-- `date-time`: Six `select` elements are used to select the year, the month, the day, the hour, the minute and the second;
-    * If you don't want to deal with time, a `"ui:widget": "date"` uiSchema widget is alternatively available, exposing three selects for year, month and day only;
+The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following string formats:
+
 - `email`: An `input[type=email]` element is used;
 - `uri`: An `input[type=url]` element is used;
-- More formats could be supported in a near future, feel free to help us going faster!
+- `date-time`: By default, an `input[type=datetime-local]` element is used; if you solely want to rely on a date, a `date` uiSchema alternative widget is available:
+
+![](http://i.imgur.com/xqu6Lcp.png)
+
+Please note that while standardized, `datetime-local` and `date` input elements are not yet supported by Firefox and IE. If you plan on targetting these platforms, two alternative widgets are available:
+
+- `alt-datetime`: Six `select` elements are used to select the year, the month, the day, the hour, the minute and the second;
+- `alt-date`: Three `select` elements are used to select the year, month and the day.
+
+![](http://i.imgur.com/VF5tY60.png)
 
 #### For `number` and `integer` fields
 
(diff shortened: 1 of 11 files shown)


n1k0:
We should probably keep the select-based date widgets as default ones, as support isn't great across platforms just yet.

playground/samples/date.js
```diff
@@ -0,0 +1,54 @@
+module.exports = {
+  schema: {
+    title: "Date and time widgets",
+    type: "object",
+    properties: {
+      native: {
+        title: "Native",
+        description: "Note that a few browsers support these.",
```
leplatrem: nit: `May not work on some browsers` could be enough ?

test/StringField_test.js
```diff
@@ -175,12 +175,14 @@ describe("StringField", () => {
     });
   });
 
-  describe("DateTimeWidget", () => {
+  describe("AltDateTimeWidget", () => {
```
leplatrem: You haven't any tests for the native ones right ?

n1k0: Nope, see notes & todo in the bug description.

leplatrem:
FYI it works like a charm on Firefox mobile:

photo_2016-04-22_19-26-18
photo_2016-04-22_19-26-27

almet:
> after discussing about that with @leplatrem, we decided to keep it this way.

Can we know what the turning point was?

As the default HTML5 date input doesn't work on almost all non-mobile platforms. I would suggest having something that works for all platforms by default instead.

n1k0:
> Can we know what the turning point was?

Pushing and promoting Web standards.

n1k0:
Note: For Firefox, [bug 825294](https://bugzilla.mozilla.org/show_bug.cgi?id=825294) should land anytime soon. MSDN strangely declare it [supported](https://msdn.microsoft.com/en-us/library/mt584200%28v=vs.85%29.aspx), so maybe it's just matter of time for it to land on a stable version?

almet:
Mouarf, pushing and promoting web standards is great, but we should make our library users aware that this will not work for most of their users, and probably propose an alternative solution.

n1k0:
> we should make our library users aware that this will not work for most of their users, and probably propose an alternative solution

That's probably why I've added a specific section in the documentation along providing alternative widgets compatible with those other platforms.

almet:
Perfect! :)

README.md
```diff
@@ -215,13 +216,22 @@ Here's a list of supported alternative widgets for different JSONSchema data typ
...
+The built-in string field also supports the JSONSchema `format` property, and will render an appropriate widget by default for the following string formats:
 
-- `date-time`: Six `select` elements are used to select the year, the month, the day, the hour, the minute and the second;
-    * If you don't want to deal with time, a `"ui:widget": "date"` uiSchema widget is alternatively available, exposing three selects for year, month and day only;
 - `email`: An `input[type=email]` element is used;
 - `uri`: An `input[type=url]` element is used;
-- More formats could be supported in a near future, feel free to help us going faster!
+- `date-time`: By default, an `input[type=datetime-local]` element is used; if you solely want to rely on a date, a `date` uiSchema alternative widget is available:
+
+![](http://i.imgur.com/xqu6Lcp.png)
```
almet: I believe imgur isn't a good place to host these images, as the pictures could disapear any time. Should we host them in this very repository instead?

n1k0: There's no easy way to make github serving images hosted in the repo itself. Also, if these are ever gone it's no big deal to redo some captures.

n1k0:
I'm gonna land this with r=me.

README.md
almet: I believe it's possible to do it like that:

```
![alt tag](https://raw.githubusercontent.com/mozilla-services/react-jsonschema-form/master/path/to/img.png)
```

Issue #155, opened 2016-04-21
Allow the insertion of errors from outside JSONSchema validation


nuclearspike:
It's great to be able to define a regex to validate an input, however, that regex should really not be shown to the user. I'd like to be able to define an error string "Your username must only contain letters, numbers and underscores" rather than having it display the regex it isn't matching... no average user can understand what it's looking for from displaying a regex.

nuclearspike:
Oops. errorSchema appears to cover this.

nuclearspike:
Nope, I don't think it does. I thought that was a way to provide error messages.

n1k0:
You're right, we must find a way to override default error messages. Unfortunately, there doesn't seem to be any documented way in [jsonschema](https://www.npmjs.com/package/jsonschema), which is what we use for JSONSchema validation :/

nuclearspike:
The same way this component has the uiSchema, it'd be great to provide a way to define error messages where it's something outside of the scope of jsonschema. I've had to just set the .panel.errors to display:none because of all of the "instance" stuff that is not end-user-ready and that it is not using the "title" for a field. Something like "instance.first_name" is not ok to display to a user. So my solution is I just let it put the errors under the fields and hide the error list and avoid regex validation.

I should also be able to include my own error messages. For example, I created component that wraps this one that is designed to work with standard REST endpoints and Rails error messages. I'd love to be able to tell the component that "email" is already in use and have that error message appear under the email field just as it would for jsonschema validation errors. The error back from Rails has the field name that (in this common situation) matches the schema field name.

I've seen other comments about wanting to be able to perform validations yourself, that are outside of what jsonschema validations supports (passwords must match, etc). I'd like to see an a prop that allows you to provide an externally derived errorSchema that gets blended with what the component came up with on it's own. Devs could either have it touch the server and generate an errorSchema to display or just have additional client-side validations that get mixed in... this is probably a separate issue. But this would let me work around the issues I am having.

n1k0:
We definitely want to add support for custom validation with associated relevant custom error messages. "Blending" these things with what our jsonschema validation lib exposes as data and API can get tricky though.

Todo:

- Accept an `errorSchema` prop to merge with the form internal one so you can inform it some fields are in an error state according to some external validation rule; the issue is how could we possibly know when such fields are not invalid anymore?
- Accept some `validationSchema` object to define external validation rules for fields (as suggested in #145);

[...]

Issue #145, opened 2016-04-19
Custom field validation


n1k0:
Refs #144 and possibly other issues.

We should be able to define custom validation logic for fields:

```js
const schema = {
  type: "object",
  properties: {
    pass1: {type: "string", minLength: 8},
    pass2: {type: "string", minLength: 8},
  }
};

function validate({pass1, pass2}, error) {
  if (pass1 !== pass2) {
    return error("Passwords don't match");
  }
}

const uiSchema = {
  "ui:validate": validate,
  pass1: {"ui:widget": "password"},
  pass2: {"ui:widget": "password"},
};

render(<Form schema={schema} uiSchema={uiSchema} />);
```

We should also be able to registering these, as we do for custom fields and widgets. The `error` function should handle field error contextualization, so it might be bound appropriately with the field path. 

Not entirely sure these validators should be part of uiSchema or have their own container:

```js
const validatorSchema = {
  root: validate
};

render(<Form schema={schema} validatorSchema={validatorSchema} />);
```

Or simply 

```js
function validate({pass1, pass2}, error) {
  if (pass1 !== pass2) {
    return error("root.pass2", "Passwords don't match");
  }
}

render(<Form schema={schema} customValidate={validate} />);
```

Where the function will always be passed the whole `formData`, and the developer will have to specify the location of the error by hand.

Last, it appears more and more that validation and error handling could be made at the field level, as suggested by @rhgb in https://github.com/mozilla-services/react-jsonschema-form/pull/132#issuecomment-209976438, and so we would propagate errors a little like we propagate `formData` up the tree. That would obviously require some heavy refactoring, but quite possibly for the better. Also, that would allow defining custom validation logic and error propagation right from custom fields, which would be definitely convenient.

Decisions, decisions. Thoughts?

n1k0:
We should probably examine https://www.npmjs.com/package/jsonschema#custom-formats for custom validation.

n1k0:
I've started a tiny experiment in #163 about that, feedback welcome.

n1k0:
The fix has landed as part of [v0.30.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.30.0).

revolunet:
About [jsonschema custom formats](https://www.npmjs.com/package/jsonschema#custom-formats), being able to define custom formats could be useful if we dont want to use the custom form validate function.

[...]

Issue #144, opened 2016-04-18
Validation password fields


Chachaproper:
Fields: password and repeat-password. How to validate this? And how to return errors in form from backend?

n1k0:
> Password and repeat-password. How to validate this? 

This is likely to need #79 to be addressed.

> And how to return errors in form from backend?

This lib is intended to use client-side validation only. We don't have use cases for propagating errors from server side validation, so if you want this to happen that's likely to require contribution :)

n1k0:
Side note: Of course you can always add two password fields in your schema and perform validation when the form is submitted, but you then admittedly have no way to render error messages contextually to the form (meaning notifying the user outside of the form itself).

Chachaproper:
> We don't have use cases for propagating errors from server side validation.
Hmm. Example: user registers and enters a username already contained in the database, it is logical to return an error in form.

n1k0:
I know there are legitimate use cases for password validation in the real world, just here we're dealing with a client side librariy which has no plan on supporting server side rendering nor validation anytime sound. I'm just suggesting that if you want to see this happening, you should probably start to get involved by participating in the reflexion process how should it be done, eg by providing feedback on the issue I've just filed for your specific need in #145?

Chachaproper:
Thanks for answers!

n1k0:
Also filed #146 to initiate a conversation about async external validation.

n1k0:
@Chachaproper I've started an experiment in #163 about that, feedback welcome.

n1k0:
The fix has landed as part of [v0.30.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.30.0).

PR #163, opened 2016-04-24
Experiment: external form validation.

### LINKED ISSUE #146: Allow passing an errorSchema to the form component instance?
Refs #144.

If some validation can only be performed server side, do we want to offer the capability of informing the client-side form component that something has failed externally? That would mean exposing an `errorSchema` prop and being able to augment the current errorSchema with supplementary information retrieved from *elsewhere*.

The main problem being that we'll have to keep the state of these information during the lifecycle of the component, which is likely to be hard to achieve (meaning merging external with internal state, notably).

Another approach would be to only rely on custom validation at the field level as suggested in #145, but then we should ensure async validation (eg. with server side communication involved) is allowed.

Thoughts on this?

n1k0:
Refs #144, #145.

This introduces a new `validate` Form prop which accepts a function receiving `formData` and an error handler object exposing an API to add errors to the targeted field nodes in the tree. 

Here's the pitch:

```js
function validate({pass1, pass2}, errors) {
  if (pass1 !== pass2) {
    errors.pass2.addError("Passwords don't match");
  }
  return errors;
}

const schema = {
  type: "object",
  properties: {
    pass1: {type: "string", minLength: 3},
    pass2: {type: "string", minLength: 3},
  }
};

render(<Form schema={schema} validate={validate} />);
```

Note that this doesn't solve #155, which should be addressed in a separate issue. This is most likely to close #146 though.

### TODO

- [x] Support for custom validator
- [ ] Async support
- [ ] Documentation


n1k0:
Ideally I'd like to offer a really convenient API for denoting fields in error. Something like this:

```js
function validate({pass1, pass2}) {
  if (pass1.value !== pass2.value) {
    pass2.error("Passwords don't match");
  }
}
```

Chachaproper:
Perfect. I tested in the context of my project. 
The second option seems preferable.

n1k0:
> Perfect. I tested in the context of my project.

Thanks a lot for your feedback.

> The second option seems preferable.

Yeah, I agree. I'll work on this asap.

MoOx:
I am very new to this package, but sounds like a good start, and far better than nothing :)

n1k0:
I've progressed in my attempt at improving the validation API in 5a6984e34ebbf1280b26e4aad1b6a58ad40466bd, but the code is rather convoluted for what it's trying to achieve.

I'll spend more time to try to simplify things as much as I can here.

n1k0:
Now this needs to handle async validation through promises, which is our next implementation step.

n1k0:
Defenitely not very happy with that custom validation API, no mention of its current implementation... Would love some feedback from you guys if you have any other approach in mind. /cc @glasserc

n1k0:
Trying to directly use the return value from the custom validate function in https://github.com/mozilla-services/react-jsonschema-form/pull/163/commits/8d047dec379e39b2b4530a05eedaf8676a8f2149, doesn't feel much better :(

```js
function validate(formData) {
  const {pass1, pass2} = formData;
  if (pass1.getValue() !== pass2.getValue()) {
    pass2.addError("Passwords don't match");
  }
  return formData;
}
```

n1k0:
Now I think about it, I wonder if this wouldn't be much better:

```js
function validate({pass1, pass2}, {getValue, addError}) {
  if (getValue(pass1) !== getValue(pass2)) {
    addError(pass2, "Passwords don't match");
  }
}
```

I'm not 100% sure it's doable but it most definitely looks better. Will keep investigating.

glasserc:
Maybe you already considered this and discarded it, but how about something like one of these?

```javascript
function validate({pass1, pass2}) {
    let errors = {};
    if (pass1 !== pass2) {
        errors.pass2 = "passwords don't match";  // and also accept arrays like ["passwords don't match"]
    }
    return errors;
}
```

or

```javascript
function validate({pass1, pass2}) {
    const errors = new FormValidateResult();
    if (pass1 !== pass2) {
        errors.addError("pass2", "passwords don't match");
    }
    return errors;    // no return value means no errors
}
```

(You could also pass the FormValidateResult as another argument to the function, in which case the user might not have to return it.)

src/components/Form.js
```diff
@@ -51,8 +53,13 @@ export default class Form extends Component {
...
 
   validate(formData, schema) {
-    const validator = new Validator();
-    return validator.validate(formData, schema || this.props.schema).errors;
+    const {validate} = this.props;
+    const {errors} = jsonValidate(formData, schema || this.props.schema);
+    const errorSchema = toErrorSchema(errors);
+    if (typeof validate === "function") {
+      return userValidate(validate, formData, errorSchema);
+    }
```
glasserc: I might be missing something obvious here, but it seems like if you go down this code path, you lose `errors`. Is that OK? Is it only necessary to keep `errorSchema`?

n1k0:
> errors.addError("pass2", "passwords don't match");

The thing is you need to know which node matches this `pass2` identifier, which might get tricky with deeply nested structures.

src/components/Form.js
n1k0: userValidate will return an `errors` property, recomputed against the internally augmented `errorSchema`. 

I know :(

glasserc:
Oops, I didn't consider nested structures. I guess you could take dotted paths like `"obj.user.auth.pass2"` or arrays like `["obj", "user", "auth", "pass2"]`. 

How about:

```javascript
function validate({user: {auth: {pass1, pass2}}}, errors) {
    if (pass1 !== pass2) {
        errors.user.auth.pass2.addError("passwords don't match");
    }
}
```

or

```javascript
function validate(formData) {
    const errors = new FormValidatorResult(formData);
    const {pass1, pass2} = formData.user.auth;
    if (pass1 !== pass2) {
        errors.user.auth.pass2.addError("passwords don't match");
    }
    return errors;
}
```

I don't like the need to use `getValue` to access the value of the fields. I think this is part of what you were saying feels too "object oriented". I'm OK with mutating a given object and expecting the caller to notice that it got mutated; this frees you from accidentally not returning something. I'm also OK with making the validate function not mutate anything and just "purely" return some kind of object that describes the errors we found. In this case I think we should provide some kind of helper to make it easier to accumulate errors, and maybe we should enforce that the user always returns *something* so that we can help them catch the situation where they forgot to put a return statement.

n1k0:
I like the idea of passing two distinct object; the first one would be the initial, untouched `formData` object and the second one being an object exposing an API to append errors appropriately, its structure being predefined. Thank you, that's very helpful :)

n1k0:
@glasserc implemented your first suggestion in https://github.com/mozilla-services/react-jsonschema-form/pull/163/commits/e1d42a5cb3efd854e311b02c45b2056ef778bae7, would love to hear your feedback :)

There's room for improvement and many more tests to write, but I feel we're at least on the right track. Thoughts?

glasserc:
I agree! 

- I don't think we should expect the user to return something. What could they possibly return besides the thing we gave them, which they've already mutated? If they do return something, we'll try to unwrap it, which probably won't work it.
- Can we initialize the error "handler" from the data in the `errorSchema`, so that we don't have to do a merge later?

This question might be outside the scope of this PR, but why do we store both `errors` and `errorSchema`? It seems like the only place we use it is in the toplevel `Form`. Can we just keep the `errorSchema` around an generate `errors` once when we need it there?

src/validate.js
```diff
@@ -0,0 +1,36 @@
+import { isObject, toErrorList, mergeObjects } from "./utils";
+
+
+function createErrorHandler(formData) {
+  const handler = {
+    __errors: [],
+    addError(message) {
+      this.__errors = [...this.__errors, message];
```
glasserc: Why not just use `push`?

src/validate.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+  return handler;
+}
+
+function unwrapErrorHandler(errorHandler) {
+  return Object.keys(errorHandler).reduce((acc, key) => {
+    if (key === "addError") {
+      return acc;
+    } else if (key === "__errors") {
+      return {...acc, [key]: errorHandler[key]};
```
glasserc: I don't think this is right...

n1k0: Why? If the key is named `__errors` we want the accumulator to have it as well. Maybe that's just poor naming that confused you here? Or am I missing something else?

src/validate.js
n1k0: Because why not not? :D I must confess I like this syntax more, but push works too. I'm not too attached to it though, so... your call.

n1k0:
> - I don't think we should expect the user to return something. What could they possibly return besides the thing we gave them, which they've already mutated? If they do return something, we'll try to unwrap it, which probably won't work it.

The thing is, we'll have at some point to accept their function to return a `Promise` so we can support asynchronous validation.

> - Can we initialize the error "handler" from the data in the `errorSchema`, so that we don't have to do a merge later?

We could probably do that indeed.

> This question might be outside the scope of this PR, but why do we store both `errors` and `errorSchema`? 

The `errors` list is what our jsonschema lib gives us when validating data against a JSON schema. It's useful to list error in a flat fashion, like what we do for the error list at the top of the form.

The `errorSchema` map is the representation of that list as a tree, which is more convenient to propagate down our component tree so they can reflect errors contextually to their node in that tree.

> It seems like the only place we use it is in the toplevel `Form`. Can we just keep the `errorSchema` around an generate `errors` once when we need it there?

So as 

1. the `errors` list is what we get from the jsonschema validatio lib, and
2. it's used to render the flat list of errors at the top of the form

I'm afreaid we'll need to keep it around.

If you have a good plan for factorizing all this mess efficiently, I'm all ears :)

Thanks again for taking the time to provide feedback on this tricky topic!

glasserc:
Oh, yeah, promises, I didn't even consider that. Thanks.

src/validate.js
glasserc: I had assumed the output of unwrapping ``{a: {b: {__errors: ["an error on a.b"]}}}`` should be ``{a: {b: ["an error on a.b"]}}``. But thinking about it now, maybe you're trying to permit someone to add an error to an "internal" node like a user as well as "leaf" nodes like password fields.

src/validate.js
glasserc: This, and the definition of `reduce` to return new objects every iteration, feel a little scary to me because they won't be as efficient as mutating objects in-place, and I'm not sure the JS RTS is smart enough to optimize the update into an in-place mutation. I know this is JS and we probably don't care about the overhead of copying objects, but it just feels wasteful to me and not very idiomatic. But JS isn't my best language so I'm willing to accept whatever here.

src/validate.js
n1k0: > maybe you're trying to permit someone to add an error to an "internal" node

Yes.

src/validate.js
n1k0: I won't go the microbenchmark route, but I have a hunch the impact on perfs would be rather small here. Anyway I can switch back to `push` anyway if that makes you more comfortable with the change.

test/utils_test.js
```diff
@@ -603,4 +627,23 @@ describe("utils", () => {
...
+          b: {
+            __errors: ["err1", "err2"]
+          }
+        },
+        c: {
+          __errors: ["err3"]
+        }
+      })).eql([
+        {stack: "b err1"},
+        {stack: "b err2"},
```
glasserc: Shouldn't these be `a.b` ?

n1k0: In theory yes, but that would mean generating these paths. I may work on this, but I'm wondering about the added value. There's some for sure, but I'm pretty sure things would stay useable without full field path info. Need to think about it.

glasserc: Well, this output format no longer corresponds to the `jsonschema` return type. Specifically, in `jsonschema` it would be `instance.a.b`, and there would be properties called `property`, `name`, `instance`, etc. I think we should either match that output or document the format we're producing instead. I don't have a very clear sense of what the use case is for the error list, so I don't know which is better. It looks like you intend for `stack` to be usable directly as an error message, so maybe we should call it `errorMsg`. I think we will probably find that having the path available is going to be useful eventually even if it doesn't show up in the error message (that's why `jsonschema` provides it).

n1k0: > Well, this output format no longer corresponds to the `jsonschema` return type.

I don't want to stick to the format exposed by *jsonschema* too much, mostly because I'm not finding it super convenient as it uses custom error class instances, and because I'd like to decouple our implementation from their API as much as I can. For example, *jsonschema* doesn't allow defining our very own error messages instead of their builtins (eg. see #155), and we may want to investigate other libraries at some point.

> Specifically, in `jsonschema` it would be `instance.a.b`, and there would be properties called `property`, `name`, `instance`, etc. 

... and `stack`, indeed, which is the *jsonschema* property name they use for storing the complete error message. Hence why I'm complying here.

> I think we should either match that output or document the format we're producing instead. 

We must define our API and associated format before documenting it; be sure I'll update docs as soon as the ongoing implementation is validated :)

> I don't have a very clear sense of what the use case is for the error list, so I don't know which is better. 

The use case is for displaying the list of all errors at the top of the form, in a flat fashion:



This capture is made against [the hosted demo](http://mozilla-services.github.io/react-jsonschema-form/) (see the *Errors* tab).

As you can see, iterating over the error list is pretty much straigthforward to render this. Basically we really do have the two distinct use cases of

1. rendering a flat list of errors at the top of the form (our `errors` list);
2. rendering errors contextually to their field level, which needs a tree (our `errorSchema` map).

> It looks like you intend for `stack` to be usable directly as an error message, so maybe we should call it `errorMsg`.

No because as mentioned previously, I need to comply with the *jsonschema* format which uses a `stack` property to store the full error description string.

> I think we will probably find that having the path available is going to be useful eventually even if it doesn't show up in the error message (that's why `jsonschema` provides it).

Yeah that's why I still need to think about it. My main concern is we're already processing a lot of this stuff from one format to another and vice versa. We could invent some intermediary format, which is likely to look a lot like the `{errors, errorSchema}` one we already have (as, again, we actually need both data structures). Just need to ensure we're not wasting too many CPU cycles going back and forth here.

leplatrem:
Ethan, we decided to move forward and land this. We spawned some issues, if there is anything that you'd like to see addressed, please open an issue or chase us ;)

Issue #162, opened 2016-04-24
Intermittent failure on DateWidget tests


n1k0:
![](http://i.imgur.com/MZimf2C.png)

PR #164, opened 2016-04-24
Added required symbol to Array and Object fieldset captions.

apkoponen:
Refactored Arrays to use TitleField, added required prop to TitleField. Updated the README to reflect the changes.

Referencing: https://github.com/mozilla-services/react-jsonschema-form/issues/31

### DIFF
--- a/test/ObjectField_test.js
+++ b/test/ObjectField_test.js
@@ -50,7 +50,10 @@ describe("ObjectField", () => {
     });
 
     it("should render a customized title", () => {
-      const CustomTitleField = ({title}) => <div id="custom">{title}</div>;
+      const CustomTitleField = ({title, required}) => {
+        const legend = required ? title + '*' : title;
+        return <div id="custom">{legend}</div>;
+      };
 
       const {node} = createFormComponent({schema, TitleField: CustomTitleField});
       expect(node.querySelector("fieldset > #custom").textContent)
--- a/test/ObjectField_test.js
+++ b/test/ObjectField_test.js
@@ -50,10 +50,7 @@ describe("ObjectField", () => {
     });
 
     it("should render a customized title", () => {
-      const CustomTitleField = ({title, required}) => {
-        const legend = required ? title + "*" : title;
-        return <div id="custom">{legend}</div>;
-      };
+      const CustomTitleField = ({title}) => <div id="custom">{title}</div>;
 
       const {node} = createFormComponent({schema, TitleField: CustomTitleField});
       expect(node.querySelector("fieldset > #custom").textContent)
--- a/test/ObjectField_test.js
+++ b/test/ObjectField_test.js
@@ -51,7 +51,7 @@ describe("ObjectField", () => {
 
     it("should render a customized title", () => {
       const CustomTitleField = ({title, required}) => {
-        const legend = required ? title + '*' : title;
+        const legend = required ? title + "*" : title;
         return <div id="custom">{legend}</div>;
       };
 
--- a/test/TitleField_test.js
+++ b/test/TitleField_test.js
@@ -1,6 +1,4 @@
 import { expect } from "chai";
-
-import React from "react";
 import { renderIntoDocument } from "react-addons-test-utils";
 import { findDOMNode } from "react-dom";
 
@@ -26,7 +24,7 @@ describe("TitleField", () => {
     const comp = renderIntoDocument(TitleField(props));
     const node = findDOMNode(comp);
 
-    expect(node.tagName).to.equal('LEGEND');
+    expect(node.tagName).to.equal("LEGEND");
   });
 
   it("should include only title, when field is not required", () => {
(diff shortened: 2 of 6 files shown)


apkoponen:
FYI, the test fail is not related to my changes: https://travis-ci.org/mozilla-services/react-jsonschema-form/jobs/125439098

n1k0:
Yes, failing test is related to #162.

n1k0:
The patch looks pretty good, thanks. Could you please add some tests?

apkoponen:
I added a few tests and updated the ObjectField test.

test/ObjectField_test.js
```diff
@@ -50,7 +50,10 @@ describe("ObjectField", () => {
     });
 
     it("should render a customized title", () => {
-      const CustomTitleField = ({title}) => <div id="custom">{title}</div>;
+      const CustomTitleField = ({title, required}) => {
+        const legend = required ? title + '*' : title;
+        return <div id="custom">{legend}</div>;
+      };
```
n1k0: Is this really necessary as we don't have any related assertion here? Should we add one?

test/TitleField_test.js
```diff
@@ -0,0 +1,55 @@
...
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("should return a legend", () => {
+    const props = {
+      title: "Field title",
+      required: true
+    };
+    const comp = renderIntoDocument(TitleField(props));
```
n1k0: You could import and use `createComponent` from test_utils.js here.

test/TitleField_test.js
```diff
@@ -0,0 +1,55 @@
+import { expect } from "chai";
+
+import React from "react";
```
n1k0: This is not required as you're not using JSX at all in this file.

test/TitleField_test.js
```diff
@@ -0,0 +1,55 @@
...
+    const props = {
+      title: "Field title",
+      required: true
+    };
+    const comp = renderIntoDocument(TitleField(props));
+    const node = findDOMNode(comp);
+
+    expect(node.textContent).to.equal(props.title + "*");
+  });
+
```
n1k0: Nit: unnecessary blank line.

n1k0:
Thank you. Looks good with a few nits to be addressed.

test/TitleField_test.js
apkoponen: I tried this at first, but it seemed that because TitleField is a function component, not a class, it won't work. It return just null.

n1k0: I see. No big deal, let's keep it that way.

test/TitleField_test.js
```diff
@@ -0,0 +1,55 @@
...
+      title: "Field title",
+      required: false
+    };
+    const comp = renderIntoDocument(TitleField(props));
+    const node = findDOMNode(comp);
+
+    expect(node.textContent).to.equal(props.title);
+  });
+
+
```
n1k0: Nit: one blank line too many.

test/TitleField_test.js
```diff
@@ -0,0 +1,55 @@
+import { expect } from "chai";
+
+import React from "react";
+import { renderIntoDocument } from "react-addons-test-utils";
+import { findDOMNode } from "react-dom";
+
+import TitleField from "../src/components/fields/TitleField";
+import { createSandbox } from "./test_utils";
+
```
n1k0: Nit: two blank lines after imports (I know, I'm picky)

test/ObjectField_test.js
apkoponen: I think tests as the spec, so I wanted to update the example here to reflect the changes in the TitleField, even though the change does not affect the assertion.

`expect(node.querySelector("fieldset > #custom").textContent)`

n1k0: It's probably worth adding one to ensure textContent of a required label contains the star, even if admittedly redundant with the other unit tests you've added. Think of it as an integration test ;)

apkoponen: I can do that, however, isn't CustomTitleField something that is supposed to be overridden? Meaning, the implementation and handling of the required field is the responsibility of whomever implements CustomTitleField. Or is your idea that we should test my reference implementation of the CustomTitleField?

n1k0: You're right. So my gut would be to revert to the old code for `CustomTitleField`, as the reader may really wonder why required label handling is implemented here, as I initially did :)

apkoponen: OK, should we still leave the required to the props so that people reading the tests know that they can expect it or just leave it out, because it is not used by the function? Should I revert the README example too (same code)?

n1k0: Keep the example in there. No need mention the `required` prop at all in this file.

n1k0:
Landed in c73e039515cf783607831466b954f0be626d30bb.

PR #175, opened 2016-04-28
Fixes #174: Fixed playground didn't update editors.

### LINKED ISSUE #174: Playground doesn't update the JSONSchema field when navigating tabs
It seems due to state updates and render in `Codemirror#componentWillReceiveProps` being always executed asynchronously while our `Editor` component needs then to run in sync for consistent rendering according to local state.

n1k0:
> It seems due to state updates (and triggered render) in Codemirror#componentWillReceiveProps being always executed asynchronously while our Editor component needs then to run in sync for consistent rendering according to local state.

Also filed https://github.com/JedWatson/react-codemirror/issues/47 about the source issue.

Published the playground with this patch, works fine: https://mozilla-services.github.io/react-jsonschema-form/.

r=? @leplatrem

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -18,6 +18,23 @@ import "codemirror/theme/solarized.css";
 import "codemirror/theme/monokai.css";
 import "codemirror/theme/eclipse.css";
 
+// Patching CodeMirror#componentWillReceiveProps so it's executed synchronously
+// Ref https://github.com/mozilla-services/react-jsonschema-form/issues/174
+Codemirror.prototype.componentWillReceiveProps = function (nextProps) {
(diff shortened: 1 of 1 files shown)


leplatrem:
It works and I can notice that the rendering is costly (keep a key pressed in the editor while rendering the *large* forms)

I would r+ since you seem sure about it :)

n1k0:
I've just applied the setImmediate trick to a bunch of other components and it seems to improve the performances; could you recheck the playground at https://mozilla-services.github.io/react-jsonschema-form/ to see if it's actually better for you? Don't forget to force-refresh the page.

PR #179, opened 2016-04-30
Fixes #178: Local schema props precedence over foreign definitions.

### LINKED ISSUE #178: Is there a way to set different titles for fields that use $ref?
I have a schema that defines a $ref, for let's say "address" and I want both a shipping and billing address. Currently, I'm not able to figure a way to set different titles. If you set the title on the node that references the $ref definition, it's ignored. Even if you make a custom component, the schema that's passed in is only the definition, not the instance of that definition. How can I set it? 

The example of this problem is on the https://mozilla-services.github.io/react-jsonschema-form/ on the "References" page. I can set the title of the definition, but that doesn't help at all since both then end up with the same heading. There's no way to not have the underscored version of billing_address/shipping_address, is there?

n1k0:
Refs #178. r=? @leplatrem @glasserc (Monday)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "chai": "^3.3.0",
     "codemirror": "^5.13.2",
     "css-loader": "^0.23.1",
-    "eslint": "^2.6.0",
+    "eslint": "^2.9.0",
     "eslint-plugin-react": "^4.2.3",
     "estraverse": "^4.2.0",
     "estraverse-fb": "^1.3.1",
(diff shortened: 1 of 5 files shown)


leplatrem:
r+

I first thought the tests were redundant. But having a high level test for the final feature is always reassuring.

n1k0:
Yeah, there are basically a unit and an integration test, both are significant at their own level. Merging.

PR #182, opened 2016-05-03
Fixes #180: Added support for the date string format.

### LINKED ISSUE #180: date format validation does not play with type=date
### Description

type=date format validation is not working

### Steps to Reproduce

1. Navigate to https://jsfiddle.net/kaihendry/08d3reth/ in Chrome or some other browser that implements type=date http://caniuse.com/#feat=input-datetime
2. Choose a date
3. See a validation error http://s.natalian.org/2016-05-03/1462255048_2558x1404.png

#### Expected behavior

Should be considered a valid date

#### Actual behavior

    does not conform to the "date" format

### Version

https://npmcdn.com/react-jsonschema-form@0.28.1/dist/react-jsonschema-form.js

n1k0:
This adds support for the `"date"` json schema string format, which is supported by the jsonschema lib and provide a much more convenient way to describe dates comparatively to datetimes.

This also updates the `AltDateWidget` behavior accordingly. 

/cc @kaihendry
r=? @natim @glasserc @leplatrem

### DIFF
--- a/src/components/widgets/DateWidget.js
+++ b/src/components/widgets/DateWidget.js
@@ -1,16 +1,6 @@
 import React, { PropTypes } from "react";
 
 
-function fromJSONDate(jsonDate) {
-  return jsonDate ? jsonDate.slice(0, 10) : "";
-}
-
-function toJSONDate(dateString) {
-  if (dateString) {
-    return new Date(dateString).toJSON();
-  }
-}
-
 function DateWidget({
   schema,
   id,
@@ -22,9 +12,9 @@ function DateWidget({
     <input type="date"
       id={id}
       className="form-control"
-      value={fromJSONDate(value)}
+      value={typeof value === "undefined" ? "" : value}
       required={required}
-      onChange={(event) => onChange(toJSONDate(event.target.value))} />
+      onChange={(event) => onChange(event.target.value)} />
   );
 }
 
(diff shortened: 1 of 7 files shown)


test/StringField_test.js
```diff
@@ -309,17 +309,31 @@ describe("StringField", () => {
...
         type: "string",
-        format: "date-time",
+        format: "date",
       }, uiSchema});
 
       expect(node.querySelector("[type=date]").id)
         .eql("root");
     });
 
+    it("should accept a valid entered datetime", () => {
```
glasserc: nit: date

src/components/widgets/DateWidget.js
```diff
@@ -22,9 +12,9 @@ function DateWidget({
     <input type="date"
       id={id}
       className="form-control"
-      value={fromJSONDate(value)}
+      value={typeof value === "undefined" ? "" : value}
       required={required}
-      onChange={(event) => onChange(toJSONDate(event.target.value))} />
+      onChange={(event) => onChange(event.target.value)} />
```
glasserc: Could you explain why we no longer need `toJSONDate` and `fromJSONDate`? It seems like we have the same types of events and they should have the same kinds of values. Are we changing our contract or am I missing something fundamental?

n1k0: Because native HTML5 `input[type=date]` no more outputs JSON datetimes but string formatted as `YYYY-MM-DD`, which is exactly what we want (we doesn't want datetimes for dates anymore).

n1k0:
I'm landing this.

PR #183, opened 2016-05-03
[WiP] Fixes #126: Added support for disabled and readonly fields

### LINKED ISSUE #126: Add support for disabled and readonly fields
The idea is to mark a given field as either disabled or read-only. This should be handled in the uiSchema, eg. through `ui:disabled` and `ui:readonly` directives.

This can get tricky for `array` and `object` fields, where we'll have to process all child fields and widgets.

n1k0:
Feedback=? @glasserc


n1k0:
This is now ready for final review. r=? @glasserc

n1k0:
I'm gonna land this with r=me.

Issue #192, opened 2016-05-06
Add a file widget


n1k0:
Refs #191. 

We should add a default `FileWidget` component, capable of mapping the file contents to their dataURL string representation. Other use cases should be dealt with custom field/widget components I think.

Nice to have:

- Single or multiple files (`string` or `array` of `string`s schema)
- Preview for pictures
- Multiple files handling (mapping to an array of dataURLs)

Sample fiddle: https://jsfiddle.net/f9vcb6pL/1/

nbasu02:
Having this added would be really helpful on my end.  Will this be included in the next release?

n1k0:
We're gonna need this for some Mozilla projects very soon, so be assured it will eventually land.

mohsen1:
In other tools `format: file` is usually used to indicate a file schema:

```json
{
  "type": "string",
  "format": "file"
}
```

n1k0:
> In other tools format: file is usually used to indicate a file schema

Which other tools?

PR #193, opened 2016-05-06
Introducing FileWidget

n1k0:
This is an ongoing experimentation around #192.


Natim:
Note that kinto-attachment doesn't supports multiple files per records just yet.

n1k0:
This lib should remain fully agnostic from Kinto anyway.

Natim:
> This lib should remain fully agnostic from Kinto anyway.

Yes sure.

Natim:
This is now ready :dancer:

PR #198, opened 2016-05-09
[PARKED] Fixes #197: Support for asynchronous validation.

### LINKED ISSUE #197: Custom asynchronous validation
Allow the `validate()` function to return a promise so that custom validation can be asynchronous.

The whole process of validation (and display of errors in state) should now be asynchronous.

n1k0:
WiP refs #197.

This attempts at bringing support for asynchronous validation. The most challenging part is to offer decent performances.

Feedback? @leplatrem @glasserc

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -108,22 +108,26 @@ function unwrapErrorHandler(errorHandler) {
  * This function processes the formData with a user `validate` contributed
  * function, which receives the form data and an `errorHandler` object that
  * will be used to add custom validation errors for each field.
+ *
+ * @return {Promise}
  */
 export default function validateFormData(formData, schema, customValidate) {
   const {errors} = jsonValidate(formData, schema);
   const errorSchema = toErrorSchema(errors);
 
   if (typeof customValidate !== "function") {
-    return {errors, errorSchema};
+    return Promise.resolve({errors, errorSchema});
   }
 
-  const errorHandler = customValidate(formData, createErrorHandler(formData));
-  const userErrorSchema = unwrapErrorHandler(errorHandler);
-  const newErrorSchema = mergeObjects(errorSchema, userErrorSchema, true);
-  // XXX: The errors list produced is not fully compliant with the format
-  // exposed by the jsonschema lib, which contains full field paths and other
-  // properties.
-  const newErrors = toErrorList(newErrorSchema);
-
-  return {errors: newErrors, errorSchema: newErrorSchema};
+  const errorHandler = createErrorHandler(formData);
+  return Promise.resolve(customValidate(formData, errorHandler))
+    .then(UserErrorHandler => {
+      const userErrorSchema = unwrapErrorHandler(UserErrorHandler);
+      const newErrorSchema = mergeObjects(errorSchema, userErrorSchema, true);
+      // XXX: The errors list produced is not fully compliant with the format
+      // exposed by the jsonschema lib, which contains full field paths and other
+      // properties.
+      const newErrors = toErrorList(newErrorSchema);
+      return {errors: newErrors, errorSchema: newErrorSchema};
+    });
 }
--- a/src/validate.js
+++ b/src/validate.js
@@ -53,7 +53,7 @@ function toErrorSchema(errors) {
     if (Array.isArray(parent.__errors)) {
       // We store the list of errors for this node in a property named __errors
       // to avoid name collision with a possible sub schema field named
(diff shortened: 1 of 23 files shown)


n1k0:
I still need to use generator functions in a few `beforeEach` statements it seems. Will do later.

src/validate.js
```diff
@@ -108,22 +108,26 @@ function unwrapErrorHandler(errorHandler) {
...
-  // XXX: The errors list produced is not fully compliant with the format
-  // exposed by the jsonschema lib, which contains full field paths and other
-  // properties.
-  const newErrors = toErrorList(newErrorSchema);
-
-  return {errors: newErrors, errorSchema: newErrorSchema};
+  const errorHandler = createErrorHandler(formData);
+  return Promise.resolve(customValidate(formData, errorHandler))
+    .then(UserErrorHandler => {
+      const userErrorSchema = unwrapErrorHandler(UserErrorHandler);
```
glasserc: Why this capitalization?

n1k0: Hmm, mistake; will fix.

test/ArrayField_test.js
```diff
@@ -38,21 +38,21 @@ describe("ArrayField", () => {
       expect(legend.id).eql("root__title");
     });
 
-    it("should contain no field in the list by default", () => {
+    it("should contain no field in the list by default", function*() {
```
glasserc: Could you elaborate a little bit on the conversion to generator functions? To take one example at random, I don't understand why this test needed to be converted. It's not yielding anything, nor even (so far as I can tell) causing any asynchronous functions to be invoked.

glasserc:
Otherwise looks fine to me.

test/ArrayField_test.js
n1k0: As stated in the description, I wish I could scientifically elaborate, but I can't. I *suspect* the trick is that the babel polyfill for generators uses `setTimeout` (or `setImmediate`) behind the hood and *maybe* that ensures us having at least a tick passed for each contained statement, which addresses our test use cases just fine.

Also, I just blindly migrated all `it` statements to leverage this *feature* to keep things consistent, and also because before this change we simply add 158 test failures (and zero after the migration).

If you're interested in investigating this, it's probably worth inspecting the babel polyfill code for generators, to see what they're doing precisely, and figure out how it actually helps us writing synchronous test code compatible with the asynchronous nature of our current expectations (which is inherently highly surprising anyway, to say the least).

n1k0: Side notes:

- this may be highly fragile as it probably heavily depends on the current Babel polyfill implementation details;
- we should also look into `mocha` code to see if they're doing anything specific when generator functions are passed to `it()`.

n1k0:
Sigh. So it appears using generators actually avoid executing assertion statements **at all**, hence why we don't get failures. Realized that while trying to introduce erroneous assertions... 

Now I feel dumb, and is probably the best time for going to bed.

(In case this wasn't clear, *do not merge this*)

n1k0:
Last note, I quickly played with [co-mocha](https://github.com/blakeembrey/co-mocha) and generator functions, and it allows actually using generator functions, though at the cost of yielding explicit event loop ticking:



Yeah that's pretty ugly. I'm gonna sleep on it.

n1k0:
I've finally updated the patch to actually leverage generators and ensured components renders were checked asynchronously everywhere applicable.

I've still 3 failing tests to fix I've no clue how.

More worryingly the perfs when using the patch are horrible: the fact we trigger validation on each keystroke is definitely something to revisit.

I'm starting burning out on this one, so as it's not P1, I'll park this here; if anybody feels like taking over, go.

n1k0:
> More worryingly the perfs when using the patch are horrible: the fact we trigger validation on each keystroke is definitely something to revisit.

One solution would be to not support async validation when live validation is enabled. Another alternative would be to decouple validation from rendering (so local state updates), but I can't see how.

n1k0:
Idea: debouncing onChange propagation at small interval to release pressure on the constant rerendering flow. Starting emitting regular events on input blur/focus, stopping when the focus is lost.

glasserc:
That sounds good to me!

nuclearspike:
What's the status of this PR?

n1k0:
The status of this PR is stale. Would you want to work on this?

n1k0:
No activity since last year, I'm gonna close this as a mighty act of gardening. Please anyone feel free to take over.

PR #203, opened 2016-05-12
Id & className props for form

Chachaproper:
Need a custom IDs and classes for `<form>`.
I added props from Form component:    `id` and `className`;

```jsx
<Form
     id="edit-form"
     className="form form-wide"
     liveValidate={liveValidate}
     schema={schema}
     uiSchema={uiSchema}
     formData={formData}
     onChange={this.onFormDataChange}
     fields={{geo: GeoPosition}}
     validate={validate}
     onError={log("errors")} />
```


src/components/Form.js
```diff
@@ -113,12 +113,14 @@ export default class Form extends Component {
...
 
   render() {
-    const {children, schema, uiSchema, safeRenderCompletion} = this.props;
+    const {children, schema, uiSchema, safeRenderCompletion, id, className} = this.props;
     const {formData, errorSchema, idSchema} = this.state;
     const registry = this.getRegistry();
     const _SchemaField = registry.fields.SchemaField;
     return (
-      <form className="rjsf" onSubmit={this.onSubmit}>
+      <form className={className ? className : 'rjsf'}
```
n1k0: nit: we use double quotes

src/components/Form.js
```diff
@@ -113,12 +113,14 @@ export default class Form extends Component {
...
+            id={id}
```
n1k0: nit: 2 spaces indent

n1k0:
Would you mind adding a test? :)

n1k0:
We may also want to propagate other supported attributes for the `form` tag? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#Attributes

Chachaproper:
>Would you mind adding a test? :)

Yes.

>We may also want to propagate other supported attributes for the form tag?

Yes. It is better to be able to use the attributes, than not to have)

>Mixed feelings about novalidate which should probably disable validation entirely?

As far as I understand, now there is no way to disable validation. It would be nice to have the opportunity.
It seems to me that there must be a explicit prop `validate` disables validation, and attribute `novalidate` is an alias.

Chachaproper:
I added more arributes (except `novalidate`), but the problem in the tests `autocomplete`. It works but test not passed. I  tested `autocomplete` and `autoComplete`. 
Where did I go wrong?

glasserc:
I'd also like to know the answer to this. It's almost like the HTML lib only supports a predefined list of attributes, and `autocomplete` isn't one of them. It seems like jsdom is built using this set of attributes: https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/nodes/HTMLFormElement.idl. It seems like autocomplete falls into the category of "unimplemented IDL", although I don't know why: https://github.com/tmpvar/jsdom/commit/6d54df2404192257dfa9a4993621d93a226ebc80#diff-5d22a8b5a55cef0ad3d29216b466ea79.

Chachaproper:
Interesting.
Postpone the implementation of `autocomplete` in <Form />?

n1k0:
In a node repl, same jsdom version:

```
> require("jsdom").jsdom("<form autocomplete=off>").querySelector("form").attributes[0].value
'off'
```

Could it be React swallowing the attribute for some reason?

src/components/Form.js
```diff
@@ -113,12 +113,50 @@ export default class Form extends Component {
...
+      autocomplete,
+      enctype,
+      acceptcharset
+    } = this.props;
+
     const {formData, errorSchema, idSchema} = this.state;
     const registry = this.getRegistry();
     const _SchemaField = registry.fields.SchemaField;
+
+    const formJSX = (
```
n1k0: What is this used for?

src/components/Form.js
```diff
@@ -113,12 +113,50 @@ export default class Form extends Component {
...
+        autoComplete={autocomplete}
+        encType={enctype}
+        acceptCharset={acceptcharset}>
+      </form>
+    );
+    
     return (
-      <form className="rjsf" onSubmit={this.onSubmit}>
+      <form className={className ? className : "rjsf"}
+            id={id}
```
n1k0: We still want 2 spaces indent

test/Form_test.js
```diff
@@ -925,4 +925,57 @@ describe("Form", () => {
...
+    it("should set attr target of form", () => {
+      expect(node.target).eql(formProps.target);
+    });
+
+    it("should set attr action of form", () => {
+      expect(node.action).eql(formProps.action);
+    });
+
+    it("should set attr autoComplete of form", () => {
+      expect(node.autocomplete).eql(formProps.autocomplete);
```
n1k0: I suspect React actually resolves the attribute value as either `true` or `false`, and not the passed string value; could you please check about it and update the test to check for `false` instead?

Chachaproper: `undefined`. it is the specificity of React, i think.
```
const {node} = createFormComponent(formProps);
console.log(node.querySelectorAll("form")); //empty
```
I think need to try this:

```
const comp = renderIntoDocument(
  <div>
    <Form {...props} />
  </div>
);
const node = findDOMNode(comp);
node.querySelectorAll("form");
```
I do not have the possibility to test. May be tomorrow.

src/components/Form.js
Chachaproper: Sorry. What do you mean?

Chachaproper:
Fix tests.

src/components/Form.js
n1k0: This constant is never used.

Chachaproper: Yes. Sorry. Its for debug.

test/Form_test.js
```diff
@@ -925,4 +925,63 @@ describe("Form", () => {
...
+      acceptcharset: "ISO-8859-1"
+    };
+
+    const comp = renderIntoDocument(
+      <div>
+        <Form {...formProps} />
+      </div>
+    );
+
+    const node = findDOMNode(comp).querySelector("form");
```
n1k0: You need to put this in a `beforeEach` statement so it's evaluated. Also, please use the `createFormComponent` helper.

Chachaproper: @n1k0 
> Also, please use the createFormComponent helper.

If i use `createFormComponent`, `node.querySelector("form")` return `null`. I need wrapper `<div> ... </div>`.

Chachaproper: Yes! I'm going to bed)

Chachaproper: @n1k0 
> You need to put this in a `beforeEach` statement so it's evaluated before the dependent tests access it.

you mean this:

```
describe("Attributes", () => {
    ...

    let node;

    beforeEach(() => {
      node = createFormComponent(formProps).node;
    });

    ...
});
```
?

n1k0: Yes!

playground/app.js
```diff
@@ -352,6 +352,14 @@ class App extends Component {
...
           {!this.state.form ? null :
             <Form
+              id="test-id"
+              className="testclass big"
+              name="test-name"
+              method="put"
+              target="_blank"
+              autocomplete="off"
+              enctype="multipart/form-data"
+              acceptcharset="ISO-8859-1"
```
n1k0: Not sure these should be in the playground. On the other hand I think these attributes should be documented :)

Would you mind adding a section to the docs?

Chachaproper: Sorry, i'm not sure that my English allows you to do this(

Example:

## Form attributes

Form component supports the following html attributes:

```
<Form
  id="edit-form"
  className="form form-wide"
  name: "testName"
  method: "post"
  target: "_blank"
  action: "/users/list"
  autocomplete: "off"
  enctype: "multipart/form-data"
  acceptcharset: "ISO-8859-1"
  schema={} />
```

It's ok?

n1k0: Yes! Just ensure to use the `jsx` directive to get proper syntax coloration for the code block.

n1k0: One last bit: could you please remove the use of custom attributes from the playground? I don't really want to limit the accepted charset for instance :) I think we shouldn't add any.

n1k0:
This is a nice and useful patch! Thanks for contributing.

PR #213, opened 2016-05-24
(RFC) Specify displayLabel in the uiWidget component.

almet:
I would like to validate the idea with @n1k0 first, and then this needs to have some tests written for it.

### DIFF
--- a/test/SchemaField_test.js
+++ b/test/SchemaField_test.js
@@ -34,6 +34,23 @@ describe("SchemaField", () => {
     });
   });
 
+  describe("ui:widget displayLabel support", () => {
+    class MyCustomWidget extends React.Component {
+      render() {
+        return <div>my-widget</div>;
+      }
+    }
+
+    MyCustomWidget.displayLabel = false;
+
+    const uiSchema = {"ui:widget": MyCustomWidget};
+
+    const {node} = createFormComponent({schema, uiSchema});
+
+    expect(node.querySelectorAll("#custom"))
+      .to.have.length.of(1);
+  });
+
   describe("ui:field support", () => {
     class MyObject extends React.Component {
       constructor(props) {
--- a/playground/samples/simple.js
+++ b/playground/samples/simple.js
@@ -4,6 +4,10 @@ module.exports = {
     type: "object",
     required: ["firstName", "lastName"],
     properties: {
+      description: {
+        type: "string",
+        description: "This is a form where you can define some data."
+      },
       firstName: {
         type: "string",
         title: "First name",
@@ -28,6 +32,9 @@ module.exports = {
     }
   },
   uiSchema: {
+    description: {
+      "ui:widget": "paragraph"
+    },
     age: {
       "ui:widget": "updown"
     },
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -7,6 +7,8 @@ import NumberField from "./NumberField";
 import ObjectField from "./ObjectField";
 import StringField from "./StringField";
 import UnsupportedField from "./UnsupportedField";
+import {getAlternativeWidget} from "../../utils";
+
 
 const REQUIRED_FIELD_SYMBOL = "*";
 const COMPONENT_TYPES = {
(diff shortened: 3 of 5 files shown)


test/SchemaField_test.js
```diff
@@ -34,6 +34,23 @@ describe("SchemaField", () => {
...
+      }
+    }
+
+    MyCustomWidget.displayLabel = false;
+
+    const uiSchema = {"ui:widget": MyCustomWidget};
+
+    const {node} = createFormComponent({schema, uiSchema});
+
+    expect(node.querySelectorAll("#custom"))
```
almet: This test isn't finished, I want some feedback on the approach first :)

n1k0:
Mostly thinking out loud quickly here.

I'm a little uncomfortable with supporting this specific use case in this lib, as its scope was initially targetted at building forms, not advanced DOM compositions. Okay, that highly depends on one's definition of what a *form* precisely is, but I keep feeling this may be slightly out of scope.

For example, this whole library relies on the assumption that json schema fields eventually map to state (`formData` as we call it). What should we do with paragraphs defined as part of the json schema? Should they be submitted along the other regular fields? Blacklisted? Are these part of the data model, or are they cosmetic pieces attached to the presentation of the model data? I tend to think the latter, but that may be just me.

Also, if we start supporting static paragraphs, should we consider supporting illustrative images? youtube videos? HTML tables? Raw HTML contents? We'd need to draw a line, and maybe "only paragraphs" would be that line, though I'd expect more demands coming our way if we send the signal that this lib actually wants to become an HTML authoring tool :D

If it was me, I'd probably start thinking of defining the static cosmetic logic in the *uiSchema* instead, but right now I can't imagine any convenient format for doing that easily.

Natim:
For me the question is more what is the best way to store some text between fields with JSON Schema.

Apparently there are none. Which means we are starting to build JSONSchema that needs a uiSchema to be rendered correctly. They will also probably needs validators to be validated correctly.

> I'd probably start thinking of defining the static cosmetic logic in the uiSchema instead

We could use that with ``ui:order`` to actually be able to define widget and then put them in between fields using ``ui:order``

```
{"ui:order": ["name", "ui:widget:gender_description", "gender", "age"],
 "ui:widgets": {"gender_description": {"type": "paragraph", "title": "foo", "description": "bar"}}}

almet:
In general, I tend to believe that we need a way to add some text to a form in order to be usable: I need to explain what the form actually does, but I understand what you mean.

> Are these part of the data model, or are they cosmetic pieces attached to the presentation of the model data?

I believe they're cosmetic, and not part of the data model: everything in the data model will actually generate some inputs, and we don't want any input to be generated out of these paragraphs of text.

>  Should they be submitted along the other regular fields? Validated? Blacklisted?

Nothing should be generated out of them: I don't know how you're gathering the data from the form, but here you don't have any way to add some data, so it shouldn't generate any data.

I'm actually perfectly fine having the uiComponents defined outside the scope of this lib (I can for instance keep them in the formbuilder). I wanted to have some input on the idea to display or not the description depending on what's specified in the component. What do you think about that?

almet:
Comenting out after @Natim comment: currently, we drive everything by the `schema`, but we cannot add new fields in there because of the json-schema spec.

Should we start driving everything via another way? how?

n1k0:
> I tend to believe that we need a way to add some text to a form in order to be usable

There are also several three ways to provide contextual text next to a field or set of fields:

- The title schema property; don't laugh, it's useful, and it's text :)
- The description; For arrays and objects, it's rendered at the top of the fieldset, where you can explain things;
- The `"ui:help"` uiSchema property; It's rendered below a given field, and it even accepts a react component:



Here's the [code to render this](https://github.com/Kinto/kinto-admin/blob/d385e7454710bfe33a1b161b8bfdc80733733713/scripts/components/AuthForm.js#L70-L81).

That leaves us with large custom chunks of HTML being not supported, but I'm having a hard time to consider it as a very string limitation...

almet:
Yes, that might be a solution to our needs, but as it currently generates a field (json schema) for each component we add, it doesn't map very well to this, but I will try alternative ways :)

n1k0:
@almet do you still think we need this or are you okay to close?

PR #222, opened 2016-06-01
Add multiple checkboxes support.

almet:
As discussed with @n1k0 earlier today, I'm pushing this as an unfinished thing. Feel free to take over, I'll have a look next week otherwise.

- [x] Implement the feature
- [ ] Add some tests
- [ ] Update the documentation

### DIFF
--- a/src/components/widgets/CheckboxesWidget.js
+++ b/src/components/widgets/CheckboxesWidget.js
@@ -1,0 +1,68 @@
+import React, { Component, PropTypes } from "react";
+
+class CheckboxesWidget extends Component {
+  constructor(props) {
+    super(props);
+    this.state = this.getStateFromProps(props);
+  }
+
+  getStateFromProps(props) {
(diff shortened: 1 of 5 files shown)


src/components/widgets/CheckboxesWidget.js
```diff
@@ -0,0 +1,68 @@
+import React, { Component, PropTypes } from "react";
+
+class CheckboxesWidget extends Component {
+  constructor(props) {
+    super(props);
+    this.state = this.getStateFromProps(props);
+  }
+
+  getStateFromProps(props) {
```
almet: After thinking about it, I believe we don't need to have any state here.

n1k0: Indeed, this should be functional component. I'm gonna take over the patch.

n1k0:
Superseded by #224.

PR #224, opened 2016-06-02
Add support for Multiple Checkboxes

n1k0:
Supersedes #222.

### Multiple choices list

The default behavior for array fields is a list of text inputs with add/remove buttons. Though there are two alternative simpler widgets for common situations like picking elements against a list of choices; typically this maps to a schema having:

- an `enum` list for the `items` property of an `array` field
- with the `uniqueItems` property set to `true`

Example:

```js
const schema = {
  type: "array",
  title: "A multiple choices list",
  items: {
    type: "string",
    enum: ["foo", "bar", "fuzz", "qux"],
  },
  uniqueItems: true
};
```

By default, this will automatically render a multiple select box. If you prefer a list of checkboxes, just set the uiSchema `ui:widget` directive to `"checkboxes"` for that field:

```js
const uiSchema = {
  "ui:widget": "checkboxes"
};
```

See the "Arrays" section of the playground for cool demos.

### Todo

- [x] Implementation
- [x] Tests
- [x] Docs


n1k0:
This is now ready for review. r=? @almet @Natim @leplatrem

src/components/widgets/CheckboxesWidget.js
```diff
@@ -0,0 +1,59 @@
...
+            <label>
+              <input type="checkbox"
+                id={`${id}_${index}`}
+                checked={checked}
+                disabled={disabled}
+                onChange={(event) => {
+                  const all = options.map(({value}) => value);
+                  if (event.target.checked) {
+                    onChange(selectValue(option.value, value, all));
+                  } else {
```
almet: Very instructive to have done it once and now seeing a better solution,  thanks

almet:
SHIP IT!

n1k0:
Released in [v0.33.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.33.0).

PR #225, opened 2016-06-02
Add widget capability to Array and Object fields

kdmcclel:
I had an array of roles for a user which I wanted to present as switches.  I wanted to be able to do this inline without having to declare all my fields manually.  Adding the capability for a widget to be used on ArrayField and ObjectField gives more flexibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,6 +504,8 @@ You can provide your own custom widgets to a uiSchema for the following json dat
 - `number`
 - `integer`
 - `boolean`
+- `array`
+- `object`
 
 ```jsx
 const schema = {
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -2,6 +2,7 @@ import React, { Component, PropTypes } from "react";
 
 import {
   getDefaultFormState,
+  getAlternativeWidget,
   isMultiSelect,
   isFilesArray,
   isFixedItems,
@@ -121,6 +122,10 @@ class ArrayField extends Component {
 
   render() {
     const {schema, uiSchema} = this.props;
+    const widget = uiSchema["ui:widget"] || schema.format;
+    if (widget && widget !== 'checkboxes') {
+      return this.renderWidget();
+    }
     if (isFilesArray(schema, uiSchema)) {
       return this.renderFiles();
     }
@@ -131,6 +136,25 @@ class ArrayField extends Component {
       return this.renderMultiSelect();
     }
     return this.renderNormalArray();
+  }
+
+  renderWidget() {
+    const {schema, idSchema, uiSchema, name, required} = this.props;
+    const {title, description} = schema;
+    const {items} = this.state;
+    const {widgets} = this.props.registry;
+
+    const widget = uiSchema["ui:widget"] || schema.format;
+    const Widget = getAlternativeWidget(schema, widget, widgets);
+    return <Widget
+        id={idSchema && idSchema.id}
+        label={title || name}
+        placeholder={description}
+        onChange={this.onSelectChange}
+        schema={schema}
+        value={items}
+        required={required}
+      />;
   }
 
   renderNormalArray() {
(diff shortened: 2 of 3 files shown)


kdmcclel:
Didn't see the checkbox widget on multiselect.  I'll fix these errors and squash.

n1k0:
This is an interesting extension of widgets scope. Would you mind adding some tests to help me figuring out the implication of such a change? Thanks.

n1k0:
@kdmcclel do you plan on keeping working on this or should I take over? Thanks

n1k0:
Okay, I'm gonna challenge this need; what would this bring over [custom fields](https://github.com/mozilla-services/react-jsonschema-form#custom-field-components)?

- Custom field can render any free form HTML for any JSON schema types; including array and object fields;
- Custom fields allow rendering a full form row, meaning you can render labels, requirements, errors list, help texts, etc. the way you like.

Do we actually want this? If so, why?

amarnus:
It is true that a custom field can be used to extend default behaviour. There are some field-level features though. Eg. _Ordered properties_ for objects. Custom fields based off the `ObjectField` will have to replicate for no reason.

Also, your question can be extended to: Why do we need widgets when we can create custom fields? Even for primitive fields. Especially because the signature for both, is the same: `f(Value, Schema, ...) -> Html`

n1k0:
> There are some field-level features though. Eg. _Ordered properties_ for objects. Custom fields based off the `ObjectField` will have to replicate this for no reason.

As far as I can tell, having *custom object widgets* wouldn't solve this, would it? What API would you provide in order for the custom widget component to know about the order of fields? 

> Also, your question can be extended to: Why do we need widgets when we can create custom fields? Even for primitive field types. Especially because the signature for both, is roughly the same: `f(Value, Schema, ...) -> Virtual DOM`

There's actually a [documented](https://github.com/mozilla-services/react-jsonschema-form#advanced-customization) distinction between fields and widgets:

> The API allows to specify your own custom *widgets* and *fields* components:
>
>- A *widget* represents a HTML tag for the user to enter data, eg. `input`, `select`, etc.
>- A *field* usually wraps one or more widgets and most often handles internal field state; think of a field as a form row, including the labels.

That's especially why I've always be reluctant at transforming *widgets* into more than single input components.

kdmcclel:
Sorry I've been meaning to write those tests just been very busy.

I understand the documented distinction but they seem to behave very similarly.  

I initially used custom fields for my roles example (dynamic options list switches).  The problem was with a custom field I was forced to specify all fields in the form.  I wanted to let one form page handle all objects based on schema.

Perhaps a better modification would have been to add ui:field tag to ui json?

amarnus:
You are right. *Ordered properties* was not a very good example.

>handles internal field state

If we take this *alone* as the definition of a field, then I think object/array widgets still make sense as the logic of handling the internal data is the same irrespective of whether you want to render an object sub-schema in a regular fieldset or within a set of tabs.

>Also custom array and object widgets would mean ideally accessing the schema and uiSchema, possibly idSchema and errorSchema as well, which would imho broaden the concept of widget far too much.

If this is the main concern and we can make an assumption that a widget will work only on its immediate children, we can provide a `renderChild` function as a prop to the `Widget` that instantiates and renders `SchemaField` - the widget is responsible for calling this function at the right place with the right child.

n1k0:
@kdmcclel 

> Perhaps a better modification would have been to add ui:field tag to ui json?

I don't understand; there's already a [`ui:field` uiSchema directive](https://github.com/mozilla-services/react-jsonschema-form#custom-field-components). What do you mean?

kdmcclel:
@n1k0 

Ah sorry its been a while since I looked at this.

So you define your custom field in the uiSchema and then link it to a react class in the fields prop.  The problem here is that now the form isn't dynamic.  I would have to add every other field into this prop.

My current setup handles everything via api schema.  There is only one edit page which uses routing data to pass the appropriate portion of the full schema into the form.  Due to this I want to avoid defining all the fields at form startup.  Using the widget instead was an easy solution as the setup is the same but the widgets prop doesn't override the dynamic behavior for form creation.

The other solution may be a customField prop which overwrites standard fields but doesn't define the entire fieldset of a form.

Issue #230, opened 2016-06-08
"title" is auto-added to custom components when nested but not top-level


nuclearspike:
I have a custom field editor for a multi-select dropdown (react-select) and with a somewhat recent change to react-jsonschema-form lib, the titles are no longer showing up on their own for properties at the top-most level. However, when nested fields are using the same editor, the titles *are* added. My solution originally was to add the label to the custom component, which fixes the top-most level, but then the nested ones now have double titles.

n1k0:
Would you mind trying to provide a reproducible example, ideally a [jsfiddle](https://github.com/mozilla-services/react-jsonschema-form/issues/181)?

nuclearspike:
Everything works great! thanks for the speedy resolution!

PR #242, opened 2016-06-21
Ensure use of property specific ui:field suppresses the default label

garethhunt:
My understanding from the documentation is that a custom field will suppress the default label.  That is not the case when a field is set on a specific schema property.

Also, I think this could be a fix for #230

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -149,6 +149,9 @@ function SchemaField(props) {
   if (schema.type === "boolean" && !uiSchema["ui:widget"]) {
     displayLabel = false;
   }
+  if (uiSchema["ui:field"]) {
+    displayLabel = false;
+  }
 
   return (
     <Wrapper
--- a/test/SchemaField_test.js
+++ b/test/SchemaField_test.js
@@ -62,6 +62,24 @@ describe("SchemaField", () => {
 
       expect(node.querySelectorAll("#custom"))
         .to.have.length.of(1);
+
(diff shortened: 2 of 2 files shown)


n1k0:
Code looks good, use case is valid. 

@nuclearspike, can you please confirm if this fixes #230 as well? Thanks!

nuclearspike:
I'll wait to do full testing until after it's released but looking at the commit, it seems that it will cover the issue I'm having. Thanks for mentioning me here. I'll close my issue once I've tested.

n1k0:
I'm landing this, thanks.

n1k0:
Released in [v0.34.1](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.34.1).

mplis-jetsetter:
I know this is a little late, but I'm just now running into the issue where my custom field's `title` isn't being shown as a label above it anymore. Based on this commit, it doesn't look like there's a way for me to set `displayLabel` and get the default label to show, so I have to basically duplicate the code from [here](https://github.com/mozilla-services/react-jsonschema-form/blob/16ec5d5d7b6e6e3086c0a4a5795129e31df54c42/src/components/fields/SchemaField.js#L43) if I want the same label as the rest of the fields.

I understand that this commit fixed an issue where you couldn't specify your own label for a custom field, but it would have been nice if the this PR was written in a way that didn't simply change the default behavior to suit one particular use case so that it didn't effect those of us that were perfectly content with the default behavior.

Issue #217, opened 2016-05-27
Allow re-ordering of array items


mohsen1:
Maybe it makes sense to have a totally different widget for this.

n1k0:
Just encountered the issue myself, we definitely want this.

dapetcu21:
I need this for a project pretty urgently, so I'll do a PR.

n1k0:
Fixed by #247.

PR #247, opened 2016-06-24
Implement array reordering

dapetcu21:
The implementation is pretty quick and dirty, but it works.

I'm not sure what to do about the UI. I currently just crammed 3 buttons with icons instead of the delete button. Probably a better solution would be using [react-dnd](https://github.com/gaearon/react-dnd).

An elegant way to keep track of the `key` prop on array items after items are deleted or reordered would also be nice.


n1k0:
Thanks, I like it a lot. I'd probably move the delete button to the right though. Otherwide the implementation is good. Can we expect to have tests for this?

dapetcu21:
I'd write some tests, but right now I'm quite in a hurry. Maybe later into the weekend?
Also, tagging #217 here.

n1k0:
For consistency, could you please replace the *Add* caption with a plus sign too?

n1k0:
> I'd write some tests, but right now I'm quite in a hurry. Maybe later into the weekend?

Sure we can wait, take your time. And thanks for contributing.

src/components/fields/ArrayField.js
```diff
@@ -105,6 +105,16 @@ class ArrayField extends Component {
     };
   };
 
+  onReorderClick = (index, newIndex) => {
+    return (event) => {
+      event.preventDefault();
+      const items = this.state.items.slice(0);
+      const item = items.splice(index, 1)[0];
+      items.splice(newIndex, 0, item);
```
n1k0: We may want to move this as a helper in the `utils` module. Also you may want to factor the code [here](https://github.com/mozilla-services/react-jsonschema-form/blob/15c3f4b81edfd01a5241750166ba3d15d57f99ed/src/components/widgets/CheckboxesWidget.js#L6-L9) in that helper and use it in this implementation too.

n1k0:
Poke?

dapetcu21:
I didn't forget :D I was just a little busy this week.

dapetcu21:
Done 😄

n1k0:
This is a great feature and a great patch. Thanks a lot for your contribution, much appreciated.

n1k0:
Released in [v0.36.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.36.0).

matiaslarsson:
Awesome work!

I'm wondering if there perhaps should be a way for people to opt-in or opt-out of array field reordering (without having to use fixed item arrays) but keep the add and remove buttons.

It sure makes sense to be able to reorder array items in some cases, but not all.

n1k0:
@matiaslarsson Sure, we could add a uiSchema widget option for that, eg. `"orderable": false` or similar. Would you like to work on this?

matiaslarsson:
@n1k0 Unfortunately I wouldn't be able to prioritize it right now, but if nobody else has fixed it when I have time, I'll let you know!

swetapatnaik:
Hi,
Is the suggestion given by @matiaslarsson above like "orderable": false is implemented ?

epicfaace:
@swetapatnaik yes -- see https://react-jsonschema-form.readthedocs.io/en/latest/form-customization/#orderable-option

PR #252, opened 2016-06-30
Fixes #250: Add support for custom widget options.

### LINKED ISSUE #250: Add support for custom widget options
We should add support for options to custom widgets, so these could access them as props.

```js
const CustomWidget(props) {
  const {value, onChange, options} = props;
  return <div className={options.className} />;
}

const uiSchema = {
  "ui:widget": {
    widget: CustomWidget, 
    options: {
      className: "toto"
    }
  }
};
```

n1k0:
Refs #250, WiP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
      - [Form attributes](#form-attributes)
   - [Advanced customization](#advanced-customization)
      - [Custom widget components](#custom-widget-components)
+        - [Custom component registration](#custom-component-registration)
+        - [Custom widget options](#custom-widget-options)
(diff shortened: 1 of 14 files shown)


n1k0:
When suddenly you read this in the custom widget docs:

> - `options`: The list of options for `enum` fields;

We need to find a better prop name; `config`? `widgetOptions`? `context`? `params`?

olzraiti:
We have used 'ui:options' as options prop name for custom schema fields' uiSchemas.

n1k0:
> We have used 'ui:options' as options prop name for custom schema fields' uiSchemas.

For custom *fields* that's a good way to go, but this patch is about provising options to *custom widgets* :)

n1k0:
Paired with @magopian.

Issue #255, opened 2016-07-01
Inefficiencies with shouldRender


mplis-jetsetter:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme).

### Description

I believe the `shouldRender` function in `utils.js` is returning `true` in too many instances, particularly for nested `ArrayField`s or `ObjectField`s, because it's unnecessarily requiring function equality in props and state.

### Steps to Reproduce

1. In your clone of this repository, add this line to `shouldRender`: `console.log(comp.props, nextProps, !deeper(comp.props, nextProps) || !deeper(comp.state, nextState));`
2. Start the development server with `npm start`
3. Go to `localhost:8080` and click on the "Nested" tab
4. Start typing in the "Task list title" field
5. Look in the console logs for something like `Object {name: "tasks"...`. This is where `shouldRender` was called for the "tasks" `ArrayField`
6. Observe that the objects logged on that line are structurally equivalent, but that `shouldRender` is still returning `true`. I believe that this is happening only because the `onChange` functions in those objects are referentially non-equal.

#### Expected behavior

I would expect that `shouldRender` for the "tasks" `ArrayField` would return `false` and that the component does not re-render, because none of the data that affects how the component renders has changed. Re-rendering all `ArrayField`s or `ObjectField`s when none of the data in that component has changed, particularly in large forms, is a performance drain.

#### Actual behavior

`ArrayField` re-renders when making a change to another part of the form.

### Version

0.34.1

### Proposed Solution

I think that a function like the one below should be added to `utils.js` and called on `comp.props` and `nextProps` before they're passed to `deeper`

```
function _removeFunctionsFromObj(obj) {
  var newObj = {};
  for (var key in obj) {
    if (typeof obj[key] !== 'function') {
      newObj[key] = obj[key];
    }
  }
  return newObj;
}
```

I'd be happy to submit a PR, but I thought it might be better to start a discussion first in case there's another way to solve this problem.

n1k0:
This is exactly why I used instance properties for many methods in the components API; the root of the issue is binding:

```js
> function foo(){}
undefined
> foo === foo
true
> foo.bind(null) === foo
false
```

Instance function properties being pre-bound, they're always passed down to children components as a static reference, so they are comparable.

In theory we shouldn't have so many render calls, maybe we forgot to use instance properties in some place in the code?

[...]

PR #256, opened 2016-07-01
Remove functions from props during shouldComponentUpdate

mplis-jetsetter:


Natim:
Could you give a little bit of a rationate about that PR?

Natim:
Is it related to #255 discussions?

n1k0:
> Is it related to #255 discussions?

Yes

n1k0:
I've just deployed your patch to [gh-pages](https://github.com/mozilla-services/react-jsonschema-form). Sadly I couldn't spot any specific perf improvements... is there any specific to inspect and test manually? I've use Chrome CPU recording and timeline tabs but I'm not a spcialist, so hints are welcome.

mplis-jetsetter:
You're sure it's deployed to [gh-pages]()? When I run this branch locally and am on the Nested tab, I can turn on "Trace React updates" when using the React Developer Tools in Chrome and see that the items in the Tasks array are not being rerendered, [but I'm not seeing the same results when I do the same thing in the playground](http://imgur.com/a/bKfif).

n1k0:
> You're sure it's deployed to gh-pages?

It's not deployed there anymore as I've released v0.35.1 in the meanwhile, and had to deploy the updated version of the playground. I'll redeploy the patch.

n1k0:
> I can turn on "Trace React updates" when using the React Developer Tools in Chrome and see that the items in the Tasks array are not being rerendered

That means we have a mutated function reference passed down to child components there. I'm now pretty convinced it's [`onChangeForIndex`](https://github.com/mozilla-services/react-jsonschema-form/blob/37eb321873729d43d4d0cbbd51144048edf15d4e/src/components/fields/ArrayField.js#L108-L116), which is an higher order function (a function returning a function). Because:

```js
> function foo(){return () => {}}
undefined
> foo() === foo()
false
```

mplis-jetsetter:
> I'm now pretty convinced it's onChangeForIndex, onPropertyChange and similarily shaped functions

Oh interesting. It looks like changing the structure of those functions does prevent some of the unnecessary rerendering I was seeing before. I've been unsuccessfully fighting with `npm link` to try to test out those changes in my app (any tips?), but it does at least seem promising.

n1k0:
> It looks like changing the structure of those functions does prevent some of the unnecessary rerendering I was seeing before

What did you do? I'm pulling my hair out trying to figure out how to pass down the right comparable function for these with no luck so far... Maybe your approach is the way to go after all.

>  I've been unsuccessfully fighting with npm link to try to test out those changes in my app (any tips?)

I never managed to achieve that, but I'm interested :)

n1k0:
Care to elaborate why closing this? :)

mplis-jetsetter:
Yeah sorry that was my bad. I was messing with my fork and accidentally closed this. I can resubmit it if you want.

> What did you do? I'm pulling my hair out trying to figure out how to pass down the right comparable functions for these with no luck so far.

I didn't do anything fancy. I literally just pulled the returned function out and put it on the class like this:

```
_onChangeForIndex = (value) => {
  this.asyncSetState({
    items: this.state.items.map((item, i) => {
      return index === i ? value : item;
    })
  });
}

onChangeForIndex = (index) => {
  return this._onChangeForIndex;
};
```

I was just playing around and wanted to see if it made any difference, [which it did](http://imgur.com/YWTSRy4), though there's still some unnecessary rerendering.

n1k0:
> I was just playing around and wanted to see if it made any difference, which it did

Well maybe there's some perf improvement but this code doesn't work for actually updating the state :) you need to pass `index` to `_onChangeForIndex`, and that's precisely where it gets tricky...

mplis-jetsetter:
Ha wow, I obviously wasn't paying close enough attention when I made that change. I wonder why I didn't get a bunch of `ReferenceError`s saying that `index is not defined`.

I had assumed that my approach was going to be a little too hacky, but if you're thinking it might be the way to go, should I resubmit the PR? I might actually have to submit a new one instead of reopening this one.

n1k0:
Yeah please resubmit it. I've seen other people removing functions from objects before comparing them in `shouldComponentUpdate`, and after thinking about it for a while now renders should never be conditional to the function references we pass as props in this project. Also, I like that it's a simple and fast approach :)

mplis-jetsetter:
Resubmitted the PR in #261

PR #261, opened 2016-07-05
Remove functions from props before checking if component should render

mplis-jetsetter:
Resubmission of #256 since I messed up my fork


n1k0:
I've deployed the patched playground again to gh-pages. Couldn't spot any misbehavior, but I'll spend some time doing more testing. In the near future I may just rewrite a local equivalent of *deeper*, as we're not leveraging it enough to justify depending on it anymore.

n1k0:
I've done just that in #264. Closing this one, sorry for the added trouble :x

PR #262, opened 2016-07-06
Fix issue because of which the idSchema gets corrupted if the schema has a property by the name "id"

amarnus:


src/components/fields/ArrayField.js
```diff
@@ -22,15 +22,15 @@ function ArrayFieldTitle({TitleField, idSchema, title, required}) {
   if (!title) {
     return null;
   }
-  const id = `${idSchema.id}__title`;
+  const id = `${idSchema["$id"]}__title`;
```
n1k0: You can write `idSchema.$id`

test/utils_test.js
```diff
@@ -511,20 +511,59 @@ describe("utils", () => {
...
+              age: {
+                type: "integer"
+              }
+            },
+            required: [ "id", "age", "name" ]
+          }
+        }
+      };
+      expect(toIdSchema(schema)).eql({
+        "$id": "root",
```
n1k0: `$id: "root"` works

test/utils_test.js
```diff
@@ -511,20 +511,59 @@ describe("utils", () => {
...
+                type: "string",
+                enum: [ "M", "F", "Other" ]
+              },
+              age: {
+                type: "integer"
+              }
+            },
+            required: [ "id", "age", "name" ]
+          }
+        }
```
n1k0: I don't think we need such a large sample schema; a simple schema like

```js
const schema = {
  type: "object",
  properties: {
    id: {type: "string"}
  }
};
```

is good enough.

n1k0:
Good catch, thanks.

amarnus:
Noted comments. Will update the request with these changes.

amarnus:
Added recommended changes, squashed commits into one and updated request.

n1k0:
Looks good, thanks for contributing.

n1k0:
Released in [v0.36.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.36.0).

PR #264, opened 2016-07-06
Fix #255: Replace deeper dependency with local deepEquals helper.

### LINKED ISSUE #255: Inefficiencies with shouldRender

n1k0:
This removes our dependency to [deeper](https://github.com/othiym23/node-deeper), introduces a new `deepEquals` util leveraging part of its implementation and tweaks the comparison alg. to whitelist functions, so that solves #255 as well.

We had troubles with *deeper* in the past, and the project doesn't seem to move forward anymore (see [this PR](https://github.com/othiym23/node-deeper/pull/4) which has not even received a comment in 3 months), so I think now is a good time to move to something else more immediately suited to our needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -878,34 +878,6 @@ This component follows [JSON Schema](http://json-schema.org/documentation.html) 
 * `additionalItems` keyword for arrays
     This keyword works when `items` is an array. `additionalItems: true` is not supported because there's no widget to represent an item of any type. In this case it will be treated as no additional items allowed. `additionalItems` being a valid schema is supported.
 
-## Troubleshooting
-
-### Build error wrt missing "buffertools" module
-
-> Build error: Cannot find module 'buffertools' from  .../node_modules/react-jsonschema-form/node_modules/deeper
-
-This is [an issue](https://github.com/othiym23/node-deeper/issues/3#issuecomment-206232232) in the way `deeper` handles the optional `buffertools` dependency (by using try/catch around a require('buffertools')).
-
-If you use **webpack** you can do the following to ignore this import:
-
-```javascript
-module.exports = {
-  entry: "app",
-  output: { ... },
-  plugins: [
-    new webpack.IgnorePlugin(/^(buffertools)$/), // unwanted "deeper" dependency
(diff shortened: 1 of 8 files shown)


n1k0:
Landed in [v0.36.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.36.0).

PR #267, opened 2016-07-07
Fix #266: Introduce the ui:placeholder uiSchema directive.

### LINKED ISSUE #266: Introduce a "ui:placeholder" uiSchema directive to replace our current use of schema description for this
Now we're rendering the schema "description" property above each field, it shouldn't be used anymore as a placeholder for the field as it's redundant and doesn't carry the same meaning.

We should rather introduce a new `ui:placeholder` uiSchema directive which would convey a much more explicit usage.

n1k0:
Refs #266. r=? @magopian @leplatrem @Natim 

Note: this is a backward incompatible change as we're no more relying on the schema description property to fill placeholders. This will be documented in the release changelog.

### Placeholders

Text fields can benefit from placeholders by using the `ui:placeholder` uiSchema directive:

```jsx
const schema = {type: "string", format: "uri"};
const uiSchema = {
  "ui:placeholder": "http://"
};
```


n1k0:
I need to move on and release this asap.

n1k0:
Released in [v0.37.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.37.0).

jared83:
Ok, bu how do I use `ui:placeholder` with a $ref that is itself an object with fields that I want to set placeholder for?
This doesn't work:
JSON schema:
```
{
definitions: {
    ObjA: {
        type: "object",
        propA: {
            type: "string"
        }
    }
}
type: "object",
properties: {
    name: {
      type: "string",
      title: "Name"
    },
    Arr: {
      "type": "array",
      "title": "Arrrrr!!!!",
      "minItems": 1,
      "items": {
        "$ref": "#/definitions/ObjA"
      }
    },
}
}
```

UI schema:
```
{
// this does nothing. How do I accomplish this?
    definitions: {
        ObjA: { propA: { "ui:placeholder": "Property A", label: false } }
    }
}
```

epicfaace:
You can't use `definitions` in uiSchema as of now.

jared83:
@epicfaace Thank you, understood. But then, how do I accomplish what I'm aiming at in this scenario?
I want every prop defined in the ObjA to have no label and a placeholder instead.
ObjA will be used only as a $ref in an array, as shown in the example schemas.
Thank you for your time.

epicfaace:
@jared83 as of now, you would have to add the uiSchema attribute in the location where the definition is actually used, not where it is defined. 

So in your case, your uiSchema should be
```json
{
   "Arr": {
      "items": { "propA": { "ui:placeholder": "Property A", "label": false } }
   }
}
```

PR #276, opened 2016-07-12
Fix #275: Denote optional field more explicitely.

### LINKED ISSUE #275: Optional/required label formatting
### Prerequisites

### Description
Per https://uxdesign.cc/design-better-forms-96fadca0f49c#.mmbenef8x it'd look a heck of a lot nicer if optional params had -Optional labels rather than required ones having a star.

Alternatively we might be able to move that into CSS so the asterisk or optional text could be controlled there.

I get that some people likely prefer the current behavior so I'd love to see ideas on where to control this otherwise.

Happy to investigate here if a pull request of some kind would be welcomed

n1k0:
Refs #275, WiP, cc @kav; patch deployed to [gh-pages](https://mozilla-services.github.io/react-jsonschema-form).

Denoting optional fields instead of required ones, as suggested in [these UX guidelines for forms](https://uxdesign.cc/design-better-forms-96fadca0f49c):

uxforms

Edit: revamped the patch with better styling for optional fields. it now prevents the Form root schema field to be considered as optional:

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -791,8 +791,8 @@ Simply pass a `fields` object having a `TitleField` property to your `Form` comp
 
 ```jsx
 
-const CustomTitleField = ({title, required}) => {
-  const legend = required ? title + '*' : title;
+const CustomTitleField = ({title, required, root}) => {
+  const legend = root || required ? title + '*' : title;
   return <div id="custom">{legend}</div>;
 };
 
@@ -807,6 +807,8 @@ render((
         fields={fields} />
 ), document.getElementById("app"));
 ```
+
+> Note: the Boolean `root` prop denotes if the associated field is the root Form schema field itself; usually, you want to consider it to be always required.
 
 ### Custom descriptions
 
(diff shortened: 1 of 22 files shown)


n1k0:
As if doing this wasn't hard enough already, I asked my twitter followers about that; mixed feelings https://twitter.com/n1k0/status/752908241529483265

n1k0:
Damn, this doesn't help



Maybe we should make it an option (no pun intended)?

mplis-jetsetter:
In a form with both required and optional fields, I think both strategies work fine. You can argue over which looks better, but they both get the job done. The issues come into play when you have an all-required or an all-optional form.

I think there's definitely potential for confusion with a large all-required form if you're only denoting optional fields. I think the user is likely to not realize that all fields are required, attempt to submit an incomplete form, and be confused/frustrated by the error messages thinking "I didn't know that field was required".

In general, it just makes more sense to me to call out required fields. You're calling the user's attention to fields that require their attention instead of calling their attention to fields that don't require their attention.

kav:
I think it will vary by use case and in some cases even form to form. See my pull for a bit more flexible alternative that doesn't make a value judgement here

n1k0:
Now #304 has landed, one can easily override how labels are rendered. Discarding this.

PR #277, opened 2016-07-12
fix HiddenWidget error when value is undefined

olzraiti:
This commit makes sure that HiddenWidget is always a 'controller' component. If HiddenWidget's value was undefined, and later on it threw this error when giving it some value: 

>"Warning: HiddenWidget is changing an uncontrolled input of type hidden to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components".

### DIFF
--- a/src/components/widgets/HiddenWidget.js
+++ b/src/components/widgets/HiddenWidget.js
@@ -3,7 +3,7 @@ import React, { PropTypes } from "react";
 
 function HiddenWidget({id, value}) {
   return (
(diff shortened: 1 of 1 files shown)


n1k0:
Thanks!

n1k0:
Released in [v0.38.1](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.38.1).

Issue #221, opened 2016-06-01
Adding an item to an ArrayField triggers form validation even if liveValidate is false


n1k0:
Pressing the _Add_ button to add a new entry in an ArrayField force-triggers the validation of the whole form., even if we're not supposed to (eg. by default when the `liveValidate` option is set to false).

To reproduce, you need to first add an item, then delete it, then readd.

n1k0:
This seems to have been introduced by #195.

mplis-jetsetter:
I was looking into this a little bit and had a question. What's the purpose of [`options.validate` in `Form.onChange()`](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/Form.js#L80)?

It seems like this bug is occurring because in the next line, `mustValidate` is being set to true even if `liveValidate` is false because `options.validate` is being set to true [here](https://github.com/mozilla-services/react-jsonschema-form/blob/402a3e766f565f0d3639019343009d9d03e45ca0/src/components/fields/ArrayField.js#L93). I'm tempted to just remove `options.validate` but I'm not sure what the implications of that would be since I don't fully understand what that value is used for.

Simply removing `options.validate` seems to provide the behavior we want based on some quick testing in my local playground and the only test that fails is [this one](https://github.com/mozilla-services/react-jsonschema-form/blob/d6fa4f4dca73bc97cf9645e8a2f88a61f3c043d9/test/ArrayField_test.js#L171), which seems to allow this bug to pass through testing.

mplis-jetsetter:
PR submitted here #279. Feel free to play around with it and see if ended up breaking anything. It seems like my PR undoes all of #199, so I'm curious why the bug it was supposed to fix isn't showing up when testing with my PR.

Issue #195, opened 2016-05-09
With liveValidate set to false, deleting an array item preserves error at the previous index 


n1k0:
Steps to reproduce:

1. Head to the [playground](http://mozilla-services.github.io/react-jsonschema-form/)
2. Disable live validation 
3. Go to the *Errors* tab
4. Submit the form
5. Delete one of the second item of the *skills* array

Result:

The second listed field is now marked in error, while the value it holds is valid.

nuclearspike:
+1 I've been having this issue ever since the latest npm update.

n1k0:
The fix has landed as part of [v0.30.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.30.0).

PR #279, opened 2016-07-13
Remove options from Form.onchange

mplis-jetsetter:
Validation is now dependent only on noValidate and liveValidate props

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -74,9 +74,9 @@ class ArrayField extends Component {
     return itemsSchema.type === "string" && itemsSchema.minLength > 0;
   }
 
-  asyncSetState(state, options={validate: false}) {
+  asyncSetState(state) {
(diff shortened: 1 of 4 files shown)


mplis-jetsetter:
Attempts to fix #221

src/components/fields/ArrayField.js
```diff
@@ -101,7 +101,7 @@ class ArrayField extends Component {
       event.preventDefault();
       this.asyncSetState({
         items: this.state.items.filter((_, i) => i !== index)
-      }, {validate: true}); // refs #195
```
n1k0: As you can see here, the forced validation option is used to address #195. Your patch reintroduces this bug.

mplis-jetsetter: Hmm sorry about that, I wasn't seeing that bug when I was testing yesterday, but now it's there. I must have missed a step.

PR #280, opened 2016-07-14
An example of moving required styling to classes so can be controlled with css

kav:
No docs update, just want to show the code changes for the jsfiddle I listed in #275

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -14,7 +14,7 @@ import StringField from "./StringField";
 import UnsupportedField from "./UnsupportedField";
 import DescriptionField from "./DescriptionField";
 
-const REQUIRED_FIELD_SYMBOL = "*";
+// const REQUIRED_FIELD_SYMBOL = "*";
 const COMPONENT_TYPES = {
   "array":     ArrayField,
   "boolean":   BooleanField,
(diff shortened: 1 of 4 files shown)


mplis-jetsetter:
I'd definitely like to see some way to allow library users to style this kind of stuff through CSS. I know this is just an example so maybe you've considered this, but I think there should still be some default styling. This PR would force every user of this library to add some CSS to get any kind of required/optional styling. Also, should there also be a class like `control-label-optional` so you can target those elements? Otherwise, it seems like you'd have to use some kind of negative selector if you wanted to just denote the optional elements.

kav:
100% agree

n1k0:
Accessibility considerations apart (which are probably a concern here though), we'll eventually move towards a field template system like suggested in https://github.com/mozilla-services/react-jsonschema-form/issues/275#issuecomment-231948348; that way people will have a way to provide their very own semantics (like what this patch suggests), while we provide accessible defaults.

kav:
Seems reasonable though I'd argue some more detailed classing and hinting as above are far more accessible than * at the end of the label so I'd still encourage for something more like that as the default. Happy to address @mplis-jetsetter's concerns if you'd like.

n1k0:
Templates have landed in #304 and are the recommended way to alter the HTML field semantics.

Issue #272, opened 2016-07-11
Default widget for each field instance


amarnus:
### Description

I thought that it will be helpful to have a root-level prop `defaultWidgets` that stores the default widget type for each field  (We should have sensible defaults, of course). This could be especially useful after #225 is merged in.

**Example**:

The key will be the field type and the value will be a component registered as a widget.

``` javascript
{
  "object": "CollapsibleFieldset",
  "array": "fieldset"
}
```

n1k0:
To be noted, there's an ongoing effort to support this in #283.

LucianBuzzo:
Resolved by https://github.com/mozilla-services/react-jsonschema-form/pull/371

PR #283, opened 2016-07-16
Allows the modification of default fields

kikonejacob:
This addition in `getFieldComponent` function allows us to override default type fields components by customized fields without needing to specify `ui:field` in the uiSchema. This is useful if you want to render the schema in another context than a form. 

I have added an exception for `object` type because of the complexity of the component.

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -32,6 +32,9 @@ function getFieldComponent(schema, uiSchema, fields) {
   if (typeof field === "string" && field in fields) {
     return fields[field];
   }
+  if (schema.type != "object" && schema.type in fields) {
+    return fields[schema.type];
+  }
   return COMPONENT_TYPES[schema.type] || UnsupportedField;
 }
 


n1k0:
Could you please provide a little more context on what you're trying to achieve here, typically with some sample usage code?

Kikonejacob:
Yes off course . Take a look at this example http://www.webpackbin.com/VJkEhrtwW. Use chrome browser to open the link

n1k0:
Sorry for the late feedback. I see what you're after, and there's an open issue about it in #272.

Your approach is smart :) Though having the default widgets mapped to types in the `fields` prop seems a little brutal to me, how about moving them to something like a `defaults` property?

So instead of `fields = {string: Caption}` we'd have `fields = {defaults: {string: Caption}}`. What do you think? If you agree, could you please update the code, and add tests & docs? Thanks!

olzraiti:
This pull request is solving pretty much the same problem I was trying to explain complicatedly in #310.

I like how lightweight this approach is. Could we also allow changing the default "object" field? The one liner commit is currently explicitly preventing that. This would be useful for example implementing a bootstrap grid layout for all object fields.

n1k0:
Ping @kikonejacob, are you still working on this, or should someone else take over?

n1k0:
Fixed by #371.

PR #285, opened 2016-07-20
Context property

### LINKED ISSUE #278: Allow passing arbitrary data to fields/widgets through root registry
We'd like to pass some data to all our custom fields/widgets from the Form root to make some fields/widgets context aware. I think the `registry` object could be utilised for this.

This would make it possible to implement any context aware fields/widgets. This would be useful for example for passing translations (at least for now, since there is no support for translations yet).

Now we pass registry fields and widgets like so:

```
<Form
  fields={fields}
  widgets={widgets}
/>
```

What we'd like to have is this:

```
<Form
  registry={
      fields: fields,
      widgets: widgets,
      someArbitraryObject: someArbitraryObject}
/>
```

olzraiti:
Closes #278


n1k0:
@olzraiti I'm off for 3 weeks, I'll take a look when I'm back.

glasserc:
This looks OK to me, although I don't know the code as well as @n1k0 does.

glasserc:
Maybe a unit test would be good?

n1k0:
While the `context` name for this prop seems to make a lot of sense, it may induce confusion for seasoned React developers as contexts are a [core feature of the lib](https://facebook.github.io/react/docs/context.html).

It may be safe to use a more distinct name like `formContext`. What do you thing?

olzraiti:
That's fine by me. Just a minute.

n1k0:
Thanks! Also as @glasserc mentioned, could we have a few tests added as well?

olzraiti:
Well, I added few tests. I'm not sure if we should also test if updating formContext propagates the changes to all widgets etc. I tried doing that, but it wasn't very simple and none of the other properties are tested for this.

test/FormContext_test.js
```diff
@@ -0,0 +1,61 @@
...
+        schema: {type:  type},
+        uiSchema: {"ui:widget": type},
+        widgets,
+        formContext
+      });
+
+      expect(node.querySelectorAll("#" + formContext.foo))
+        .to.have.length.of(1);
+    });
+  });
```
n1k0: I'd prefer avoid using a `for ` loop generating tests so we can still use mocha's `.only` feature.

src/components/fields/SchemaField.js
```diff
@@ -87,13 +87,14 @@ function DefaultTemplate(props) {
     hidden,
     required,
     displayLabel,
+    formContext,
```
n1k0: Please update field template docs to reflect this prop is now passed to custom template components as well.

n1k0:
Thinking out loud here, what would prevent us to carry the `formContext` in the registry object? I feel like it would avoid passing a new prop around, leveraging this one we already have and which sounds pretty much adequate for what we're trying to do here. Would be matter of destructuring the `formContext` object from the registry before creating the field/widget components we want to know about this specific prop.

Thoughts?

olzraiti:
Yes, that would simplify things a bit. I'll have a look in a few hours.

olzraiti:
One thing to notice is that the formContext can be now of any variable type.

src/components/fields/SchemaField.js
```diff
@@ -214,6 +217,7 @@ if (process.env.NODE_ENV !== "production") {
       fields: PropTypes.objectOf(PropTypes.func).isRequired,
       definitions: PropTypes.object.isRequired,
       FieldTemplate: PropTypes.func,
+      formContext: PropTypes.object.isRequired
```
n1k0: Please leave trailing commas as this is the general style for this project (here and across the patch).

src/components/fields/SchemaField.js
```diff
@@ -175,7 +177,7 @@ function SchemaField(props) {
   ].join(" ").trim();
 
   const fieldProps = {
-    description: <DescriptionField id={id + "__description"} description={description} />,
+    description: <DescriptionField id={id + "__description"} description={description} formContext={formContext} />,
```
n1k0: Nit: maybe worth spreading this one on multiple lines for legibility.

src/components/fields/ObjectField.js
```diff
@@ -117,11 +117,13 @@ class ObjectField extends Component {
...
                    title={title}
-                   required={required} /> : null}
+                   required={required}
+                   formContext={formContext}/> : null}
         {schema.description ?
           <DescriptionField
             id={`${idSchema.$id}__description`}
             description={schema.description}
+            formContext={formContext}
           /> : null}
```
n1k0: We should try to stay consistent, it seems everywhere else we append the self-closing tag bit on the same line.

n1k0:
This looks excellent. A few last nits to address and I'll land this. Thank you!

n1k0:
Landing! Thanks again for contributing.

n1k0:
Released in [v0.40.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.40.0).

Issue #295, opened 2016-08-01
Choice Array, Numbers Parsed To String 


ClockerZadq:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme).

### Description

I have an Array field (choices widget) with an enum set to [1,2,3], with enumNames set to ["fred", "wilma", "barney"]. When selecting a value from the list, I get an error "0: is not of a type(s) integer" "0: is not one of enum values: 1,2,3"

### Steps to Reproduce

Replicated this in the playground with this schema:

*SCHEMA*

```
"multipleChoicesList": {
      "type": "array",
      "title": "A multiple choices list",
      "items": {
        "type": "number",
        "enum": [
          1,
          2,
          3,
          4
        ]
      },
      "uniqueItems": true
    }
```

#### Expected behavior

The value that gets selected is returned in its original form, in this case, number.

#### Actual behavior

The value that gets selected is parsed to a string.

### Version

v0.33.3

mplis-jetsetter:
Is `choice` a custom widget that you've built?

ClockerZadq:
@mplis-jetsetter no i wrote it to indicate the default widget. I'll remove it to prevent confusion

oreon:
I need this behavior as well we have integer Ids and its hard to parse back and forth between strings

mplis-jetsetter:
Do you need the ability to select multiple numbers for your list? If not, I believe you can get the behavior you want by looking at the "Number enum" field on the Numbers tab in the [playground](https://mozilla-services.github.io/react-jsonschema-form/). If you do need that behavior, it does look like there's a bug with [this](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/widgets/SelectWidget.js#L10) function inside SelectWidget.js.

oreon:
Thanks jetsetter for pointing me to the right function

/**
 * This is a silly limitation in the DOM where option change event values are
 * always retrieved as strings.
 */
function processValue(type, value) {
  if (type === "boolean") {
    return value === "true";
  } else if (type === "number") {
    return asNumber(value);
  }
  return value;
}


Here is my schema , the problem is with "categorys" which is a many to many field .


   this.productSchema = {
            title: "Product",
            type: "object",
            required: ['name', 'price'
            ],
            properties: {
                name: { type: "string", title: "Name", },
                price: { type: "number", title: "Price", },

                categorys: {
[...]

PR #298, opened 2016-08-05
fix ArrayWidget so it stores numbers to formData when needed

mplis-jetsetter:


mplis-jetsetter:
Attempts to fix #295

src/components/widgets/SelectWidget.js
```diff
@@ -7,13 +7,23 @@ import { asNumber } from "../../utils";
...
  * always retrieved as strings.
  */
-function processValue(type, value) {
-  if (type === "boolean") {
-    return value === "true";
-  } else if (type === "number") {
-    return asNumber(value);
+function processValue(schema, value) {
+  if (schema.type === "array") {
+    let valueType = schema.items.type;
```
n1k0: Let's use `const`

src/components/widgets/SelectWidget.js
```diff
@@ -7,13 +7,23 @@ import { asNumber } from "../../utils";
...
+    if (valueType === "number" || valueType === "integer") {
+      return value.map(v => asNumber(v));
+    }
+    return value;
+  }
+  else {
```
n1k0: Nit: unnecessary `else` clause as we've returned early.

test/ArrayField_test.js
```diff
@@ -599,4 +599,31 @@ describe("ArrayField", () => {
...
+      Simulate.change(node.querySelector(".field select"), {
+        target: {options: [
+          {selected: true, value: "1"},
+          {selected: true, value: "2"},
+          {selected: false, value: "3"},
+        ]}
+      });
+
+      expect(comp.state.formData).eql([1, 2]);
+
```
n1k0: Nit: please remove this blank line

n1k0:
Great patch, thank you! r+ with comments addressed

src/components/widgets/SelectWidget.js
```diff
@@ -9,21 +9,19 @@ import { asNumber } from "../../utils";
...
-      return asNumber(value);
-    }
-    return value;
+  const valueType = schema.type;
+  if (valueType === "boolean") {
+    return value === "true";
+  } else if (valueType === "number") {
+    return asNumber(value);
   }
+  return value;
```
n1k0: Actually I suspect we could rewrite the whole function in a more compact fashion, like:

```js
function processValue({type, items}, value) {
  if (type === "array" && items && ["number", "integer"].includes(items.type)) {
    return value.map(asNumber);
  } else if (type === "boolean") {
    return value === "true";
  } else if (type === "number") {
    return asNumber(value);
  }
  return value;
}
```

n1k0:
Sorry for the late bit of review ;)

mplis-jetsetter:
No worries, yours is much better.

Side note: is there a way to run a specific test? The full test suite takes several minutes to run and sometimes I just want to check if a change I made causes a specific test to fail.

n1k0:
> is there a way to run a specific test? 

Yeah, you can use `it.only()` to only run a single test. Combined with `npm run tdd` which will rerun the test everytime the source file changes, that's really useful to work on some implementation.

> The full test suite takes several minutes to run

Wow, several minutes? Doesn't sound good :/

n1k0:
I'm landing this, great work, thanks!

n1k0:
Released in [v0.38.1](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.38.1).

PR #302, opened 2016-08-15
Prototype component demonstrating support for OneOf

mplis-jetsetter:


n1k0:
Sorry to be so late to give a look at this. I've just checked the branch out, it's awesome :+1: 

I'm not so sure about the UX to select one of the possible schemas, but couldn't find anything better than the dropdown you came with.

Maybe we should rework the wording? *Add widget* doesn't sound explicit enough, especially for non-technical people. Any input welcome here.

Also, do you think you could work on moving the custom widget logic to the library core, and add tests and docs? That would be just great :)

johanjvrens:
Any progress on this?

jprogermphasis:
Hello, it's a bit difficult for me as I am starting in this technology...
I have not been able to find a solution yet, and my project is asking me to progress on other tasks...

Natim:
@johanjvrens Do you want to give it a try?

spacebaboon:
We're also doing some work on supporting oneOf, as we have a back-end which is auto generating JSON schema which we have to render in a form. The BE (Java) JSON Schema library creates oneOf elements without a type, which is perfectly valid according to http://jsonschemalint.com/#/version/draft-04/markup/json, but blows up in RJSF. Or it also might generate a oneOf as type 'object', but with no properties, and that also blows up in RJSF, but is also valid according to the linter site.
My solution is a CustomSchemaField that adds a dummy type: 'object' and properties: {} to keep RJSF happy, then a custom ui:widget that can render the oneOf. It feels a bit of a hack, but I don't want to fork RJSF if I don't have to.
I can share this work when (if?) we get it working, and if anyone here is interested.
And I'd love to hear if anyone else is doing any work in this area.

jManji:
Good to hear people are looking at this! Not that it's necessarily the right way of doing things, or particularly great code, but if you guys need some ideas, you can have a look at #417, my pull request which adds support for anyOf. It makes use of the `ArrayField` features, as anyOf is actually an array of various types. The initial idea was that oneOf is then going to be just a anyOf array of one item. But I found out the hard way that this is wrong, as oneOf is not an item of an array, but the property itself (if that makes sense).

So I have also looked at oneOf a bit, and one of the challenges faced was where to put the logic. I was thinking it needs to go either in `SchemaField` as this is where the type is decided, or in `ObjectField`, or create a separate component. (`ConditionalComponent` or something). Or we could somehow `hack` the array field and use it's add/remove buttons and new `SelectWidget` feature, but this does not sound right to me. @spacebaboon feel free to share any progress, if/when you've made any.

spacebaboon:
@jManji We're using OneOf as a way to express conditional forms, i.e. depending on a radio button value, show a different set of form elements below. I was trying to do this with customised components to render the OneOf, but felt that I was painting myself into a corner, and was still facing some big problems to which I could only see hacky solutions, so I threw it all away and am using a schema transforming approach instead, which feels cleaner, if still not exactly beautiful, and doesn't require any changes to RJSF. I believe there's an example of schema transformation for conditional forms in here somewhere.

jManji:
@spacebaboon If I understand this right, you are changing the schema input "on-the-fly", depending on the value of the drop-down. That's probably the quickest and less intrusive (as you said) way of doing this. I might be interested in using this, if you can share the code in any way.

b1r3k:
Is anyone working on it currently? @mplis-jetsetter @n1k0

n1k0:
I don't. I also agree that the schema-transformation approach addresses the need quite nicely, so I'm not even sure we'll ever want to add support this jsonschema rule in this lib.

b1r3k:
@spacebaboon would you be so kind to share your approach (schema transforming approach)? Do you use `ArrayFieldTemplate` in order to change `onAddClick` behavior (i.e. changing underlaying schema)?

danrot:
@n1k0 According to your last comment you said that you are not sure if you want to support that `oneOf` rule, which I would find a pity. It is part of the json schema standard this project builds on, and IMHO it should also support these kind of rules then.

Now my question is if you would accept a PR which implements that. I am working on a project, where it looks to me like this is the last missing part, and I think we would be able to contribute it (would have to clarify that first, but I want to make sure you would accept a PR in the first place).

sidthekid134:
@n1k0 @danrot Has anyone worked on getting the ```oneOf``` functionality working? My team is looking to implement this into RJSF as it is not supported as of now. Wanted to reach out to see if anyone has made any progress on this before to get some ideas. I think this PR is much different from the master branch as I am not able to get this prototype ```oneOf``` example working. I took a look at the other PR's but there isn't a central place for this support.

glasserc:
Hi, at the present time RJSF doesn't support `oneOf`. This status is documented in more detail in the project FAQ (in the README).

danrot:
@sidthekid134 No, I haven't got any response, and since this apart other minor problems was critical for us, we decided to implement somehing on our own instead of using this library.

epicfaace:
Closing this PR as oneOf support was added in #1133.

PR #304, opened 2016-08-18
Fix #284: Introduce a field template API.

### LINKED ISSUE #284: Expose a field template API
Context: #275, #268, #205, #196

As suggested in https://github.com/mozilla-services/react-jsonschema-form/issues/275#issuecomment-231948348, we should expose an API for developers to provide a template for field rendering:

```jsx
// dumb example, more props would be exposed
function MyFieldTemplate({id, label, required, errors, children}) {
  return (
    <div className="my-field">
      <label htmlFor={id}>{label}{required ? "*" : ""}</label>
      {children}
      {errors}
    </div>
  );
}

render(<Form fieldTemplate={MyFieldTemplate} />);
```

n1k0:
Refs #284, WiP. Early feedback appreciated :)

### Field template

To take control over the inner organization of each field (each form row), you can define a *field template* for your form.

A field template is basically a React stateless component being passed field-related props so you can structure your form row as you like:

```jsx
function CustomFieldTemplate(props) {
  const {id, classNames, label, help, required, description, errors, children} = props;
  return (
    <div className={classNames}>
      <label htmlFor={id}>{label}{required ? "*" : null}</label>
      {description}
      {children}
      {errors}
      {help}
    </div>
  );
}

render((
  <Form schema={schema}
        FieldTemplate={CustomFieldTemplate} />,
```

The following props are passed to a custom field template component:

- `id`: The id of the field in the hierarchy. You can use it to render a label targetting the wrapped widget;
- `classNames`: A string containing the base bootstrap CSS classes merged with any [custom ones](#custom-css-class-names) defined in your uiSchema;
- `label`: The computed label for this field, as a string;
- `description`: A component instance rendering the field description, if any defined (this will use any [custom `DescriptionField`](#custom-descriptions) defined);
- `children`: The field or widget component instance for this field row;
- `errors`: A component instance listing any encountered errors for this field;
- `help`: A component instance rendering any `ui:help` uiSchema directive defined;
- `hidden`: A boolean value stating if the field should be hidden;
- `required`: A boolean value stating if the field is required;
- `readonly`: A boolean value stating if the field is read-only;
- `displayLabel`: A boolean value stating if the label should be rendered or not. This is useful for nested fields in arrays where you don't want to clutter the UI.

> Note: you can only define a single field template for a form. If you need many, it's probably time to look for [custom fields](#custom-field-components) instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
      - [Placeholders](#placeholders)
      - [Form attributes](#form-attributes)
   - [Advanced customization](#advanced-customization)
+     - [Field template](#field-template)
+     - [Custom widgets and fields](#custom-widgets-and-fields)
      - [Custom widget components](#custom-widget-components)
         - [Custom component registration](#custom-component-registration)
         - [Custom widget options](#custom-widget-options)
--- a/README.md
+++ b/README.md
@@ -553,9 +553,9 @@ Form component supports the following html attributes:
 
 ### Field templates
 
-To take control over the inner organization of each field (each form row), you can use a generic *field template*.
+To take control over the inner organization of each field (each form row), you can define a custom *field template*.
 
-A custom field template is a React stateless component being passed properties so you can structure your form row as you like:
+A field template is basically a React stateless component being passed field-related props so you can structure your form row as you like:
 
 ```jsx
 function CustomFieldTemplate(props) {
--- a/README.md
+++ b/README.md
@@ -580,16 +580,16 @@ render((
 The following props are passed to a custom field template component:
 
 - `id`: The id of the field in the hierarchy. You can use it to render a label targetting the wrapped widget;
-- `classNames`: A string containing the base bootstrap CSS classes merged with any custom ones defined in your uiSchema;
-- `label`: The label for this field;
-- `children`: The field component instance itself;
-- `errors`: A component instance listing any encountered errors;
+- `classNames`: A string containing the base bootstrap CSS classes merged with any [custom ones](#custom-css-class-names) defined in your uiSchema;
+- `label`: The computed label for this field, as a string;
+- `description`: A component instance rendering the field description, if any defined (this will use any [custom `DescriptionField`](#custom-descriptions) defined);
+- `children`: The field or widget component instance for this field row;
+- `errors`: A component instance listing any encountered errors for this field;
 - `help`: A component instance rendering any `ui:help` uiSchema directive defined;
-- `description`: A component instance rendering the field description, if any defined;
 - `hidden`: A boolean value stating if the field should be hidden;
 - `required`: A boolean value stating if the field is required;
 - `readonly`: A boolean value stating if the field is read-only;
-- `displayLabel`: A boolean value stating if the label should be rendered or not. This is useful for nested fields in array where you don't want to clutter the UI.
+- `displayLabel`: A boolean value stating if the label should be rendered or not. This is useful for nested fields in arrays where you don't want to clutter the UI.
 
 > Note: you can only define a single field template for a form. If you need many, it's probably time to look for [custom fields](#custom-field-components) instead.
 
--- a/README.md
+++ b/README.md
@@ -551,7 +551,49 @@ Form component supports the following html attributes:
 
 ## Advanced customization
 
-The API allows to specify your own custom *widgets* and *fields* components:
+### Field templates
+
+To take control over the inner organization of each field (each form row), you can use a generic *field template*.
+
+A custom field template is a React stateless component being passed properties so you can structure your form row as you like:
+
+```jsx
+function CustomFieldTemplate(props) {
+  const {id, classNames, label, help, required, description, errors, children} = props;
+  return (
+    <div className={classNames}>
+      <label htmlFor={id}>{label}{required ? "*" : null}</label>
+      {description}
+      {children}
+      {errors}
+      {help}
+    </div>
+  );
+}
+
+render((
+  <Form schema={schema}
+        FieldTemplate={CustomFieldTemplate} />,
+), document.getElementById("app"));
+```
+
+The following props are passed to a custom field template component:
+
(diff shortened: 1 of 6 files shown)


n1k0:
I'd really appreciate feedback from @mplis-jetsetter @nuclearspike @Natim @magopian @amarnus @MoOx or basically anybody interested in this feature.

MoOx:
lgtm. This is really a nice addition.

It would be nice to have "style" in addition to "classNames", but this comment probably apply for the entire project :D (Nice when you write CSS in JS - which is the future, you already know that /trollOrNotThatSTheQuestion ^^)

n1k0:
tbh I precisely asked to gather this sort of feedback :) we can definitely add more props, I was wondering about passing the schema, uiSchema and friends, but I fear that would be opening the door to a lot of possible abuse ^^

MoOx:
> I was wondering about passing the schema, uiSchema and friends, but I fear that would be opening the door to a lot of possible abuse

With great powers blah blah blah. If that can offer flexibility, maybe that's a good idea?

n1k0:
I don't know, I like the idea of having a template not dealing with business logic; I feel like people will see this as a way to implement logic in there while they should be doing this in custom comps. Sure I could document this heavily with big warnings and so on but in my experience:

- people don't read them
- even if they read them, they'll keep doing the wrong thing
- and after that they'll file issue
- and that usually tend to burn me out

That may just me being paranoid here though :p

MoOx:
> That may just be me being paranoid here though :p 

No, you are concerned by the project you are shipping. Totally legit.

Just bikeshedding: if people want to add logic here, it's because the "U/DX" won't be good enough so they will use that because it's seems more simple for them. This might mean the "normal" way to do is less accessible or more complicated :)

n1k0:
I can always land this limited at first and we can add more props in the feature if anyone with good legit use cases for them.

MoOx:
Definitely!

n1k0:
Okay people, I'm landing this.

n1k0:
Released in [v0.39.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.39.0).

n1k0:
> I can always land this limited at first and we can add more props in the future if anyone comes with good legit use cases for them.

It didn't take long to identify a first good use case :p https://github.com/mozilla-services/react-jsonschema-form/issues/196#issuecomment-240987027

nuclearspike:
Finally got a chance to take this for a spin, and definitely appreciate this. I use it to do automatic title prettification so I can leave out a lot of titles now. "first_name" => "First Name" or in some strange situations where I want no titles/labels (uses placeholders only), no indication that it's required but I still want the required behavior (popup, no posting until it's filled in)... I have run into an issue though. When the I have a field template set up and a schema with nested properties, the titles go all wonky. Seems like it may make sense to not use the field template when displaying the TitleField. 

Additionally (maybe best as a new ticket) it would be nice to be able to override the error field to handle that however I want. One thing I want to do is to just do a find and replace on the word "property" in the error messages. No user should ever see that (I realize that's the json validator library and not this component's code).

matiaslarsson:
Excellent work! Now I can remove the labeling from the field template andhave Material-UI render the title/label.

Unfortunately, the field template only gets me half way since objects like this;

```javascript
"watchType": {
	"type": "object",
	"properties": {
		"id": {
			"title": "Watch Type",
			"type": "number",
			"enum": [1, 2, 3],
			"enumNames": ["Apple", "Big Ben", "Cuckoo"]
		}
	}
},
```
...will be wrapped in a `<fieldset>` with the `<label>` set to *watchType*, like this;
screen shot 2016-09-13 at 18 51 20
...so what I would need is either 

1. A way to turn of the automatic labelling with something like `autolabels: false` which only uses the title from the json-schema if it's available without falling back to the property name
2. Access to a fieldset template, like @nuclearspike suggested in [#284](https://github.com/mozilla-services/react-jsonschema-form/issues/284#issuecomment-234099609)
3. Some very hard coded `display: none` CSS rules.

Or have I missed the perfect solution behind door number four?

prevuelta:
Quick question, I'm guessing the ArrayFieldTemplate (or FieldTemplate) would need to be a full component to allow collapsing? eg. collapsed state & toggle method.. At the moment I can't pass in a component class or object, it has to be a stateless functional component.. or am I missing something?

n1k0:
> At the moment I can't pass in a component class or object, it has to be a stateless functional component

Why can't you pass a stateful component? What's the error?

prevuelta:
If I pass a class I get `TypeError: Cannot set property 'props' of undefined` on this line 
```      // Check if a custom render function was passed in
      var renderFunction = ArrayFieldTemplate || DefaultNormalArrayFieldTemplate;
      return renderFunction(arrayProps);```

It's expecting a render function but gets the class constructor

n1k0:
That's indeed an unfortunate limitation #519 is gonna fix.

prevuelta:
Cheers :)

EvertLagerberg:
I am trying to figuring out how to make a part of my schema collapsed from the start as to improve UX on a large form. According to https://github.com/mozilla-services/react-jsonschema-form/issues/598, this merge was merged for my use case. However, I cannot figure out how I am supposed to use the field template API to make a part of the form collapsible. Can anyone hint me in the right direction?

janpauldahlke:
Will use this to bump. Or can someone be so kind to point me to a example, where only specific parts of the formschema are collapsible?

MatinF:
Hi again,

I've tried to achieve the removal of the "*" for required fields using the Field Template, but I'm finding it to be rather difficult to achieve without breaking something else.

As an example, I'm currently also using an ArrayFieldTemplate similar to the expanded example in the docs. But when I use a FieldTemplate in conjunction, it seems like my ArrayFieldTemplate gets overruled. 

What I would like is really simple: Not having the "*" next to any required fields. It seems like a very clean addition to the source code and something many would like. I realize that it's possible to achieve via custom fields, but for more complex forms, this quickly gets very difficult to manage.

Any thoughts on potentially making this possible?

PR #305, opened 2016-08-18
Always assign a value to defaults in computeDefaults

maclover7:
This was causing an issue for me where `computeDefaults` was returning `undefined`, which led to errors being thrown when I was trying to pass default values for `formData`.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -173,6 +173,11 @@ function computeDefaults(schema, parentDefaults, definitions={}) {
       return acc;
     }, {});
   }
+  
+  if (typeof(defaults) === "undefined") {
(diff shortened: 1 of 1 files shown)


n1k0:
As you can see this breaks a lot of tests. Could you please elaborate a little more on the problems you've encountered in the first place? Providing sample misbehaving code would greatly help.

leplatrem:
Thanks for your intent! I'm closing this PR for lack of response. Feel free to reopen with tests and more details about relevance of use case.

> [About Closing Pull Requests](https://github.com/blog/2124-kindly-closing-pull-requests)

Issue #223, opened 2016-06-02
Allow adding a custom validation error at a given array index


n1k0:
There's currently no way to declare a custom validation error at a given position in an array value. Would be nice to expose an API for that, eg.:

``` js
function validate({myArray}, errors) {
  if (myArray[1] !== "foo") {
    errors.myArray[1].addError("Nope.");
  }
  return errors;
}
```

n1k0:
Released in [v0.45.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.45.0).

PR #308, opened 2016-08-22
Allow adding a custom validation error at a given array index

knilink:
For #223

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -96,9 +96,9 @@ function createErrorHandler(formData) {
     }, handler);
   }
   if(Array.isArray(formData)){
-    return formData.reduce((acc, item, index) => {
+    return Object.assign([],formData.reduce((acc, item, index) => {
       return {...acc, [index]: createErrorHandler(item)};
-    }, handler);
+    }, handler));
   }
   return handler;
 }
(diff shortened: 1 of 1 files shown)


knilink:
Hum, it looks like the test case needs to be updated as well since the errorSchema is changed.
But I'm not sure whether it's appropriate for me to update the script to test my own code.

src/validate.js
```diff
@@ -95,6 +95,11 @@ function createErrorHandler(formData) {
       return {...acc, [key]: createErrorHandler(formData[key])};
     }, handler);
   }
+  if(Array.isArray(formData)){
+    return Object.assign([],formData.reduce((acc, item, index) => {
```
n1k0: I'm not sure `Object.assign` works with arrays; the `reduce` result should suffice in theory.

n1k0:
Thanks for the attempt at tackling this! Tests should be updated in `validate_test.js`, where new ones should be added for this functionality too :)

src/validate.js
n1k0: Ohh, I see what you did here, smart; we need keys as object properties. That should actually work.

leplatrem:
Thanks for your intent! I'm closing this PR for lack of response. Feel free to reopen with tests and more details about relevance of use case.

> [About Closing Pull Requests](https://github.com/blog/2124-kindly-closing-pull-requests)

PR #311, opened 2016-08-30
Pass empty title to TitleField instead of name

olzraiti:
Currently if you put `""` as a title in the schema, it won't be used because `"" == false`.

This pull request makes it possible to give an empty title to a jsonschema field without it being overwritten with the field's name.


n1k0:
Could you please add a test? Thanks.

olzraiti:
Yes, hopefully next week.

olzraiti:
Found the time for this today after all. :-)

n1k0:
r+, thanks a lot for your contribution.

n1k0:

awakentwice:
What if I don't want use label at field?

PR #312, opened 2016-08-31
Support old versions of React

rom10:


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "react-jsonschema-form",
-  "version": "0.40.0",
+  "name": "book-md-react-jsonschema-form",
```
n1k0: hmm.

n1k0:
Which version of React introduced this incompatibility?

package.json
rom10: Reverted

rom10:
@n1k0 https://github.com/facebook/react/blob/master/CHANGELOG.md#1500-april-7-2016

> Rendering null now uses comment nodes. Previously null would render to <noscript> elements. We now use comment nodes. This may cause issues if making use of :nth-child CSS selectors. While we consider this rendering behavior an implementation detail of React, it's worth noting the potential problem. (@spicyj in #5451)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-jsonschema-form",
-  "version": "0.40.0",
+  "version": "0.40.2",
```
n1k0: This should be removed too, I'll take care of the version ;)

src/components/fields/DescriptionField.js
```diff
@@ -3,7 +3,7 @@ import React, {PropTypes} from "react";
 function DescriptionField(props) {
   const {id, description} = props;
   if (!description) {
-    return null;
+    return <div />;
```
n1k0: Could you please add a comment before each empty div returned referencing this PR, so the reader can know about why we're doing this? Something like `// See #312: Ensure compatibility with old version of React.`

package.json
rom10: 👍

src/components/fields/DescriptionField.js
rom10: Done

n1k0:
Thank you :+1:

n1k0:

PR #324, opened 2016-09-16
Support for hiding array type widgets

tiemevanveen:
Pull request for https://github.com/mozilla-services/react-jsonschema-form/issues/322

> The limitation is due to the `<input type="hidden" ... />` input field that is used for hidden fields. Probably to support forms that don't submit with React? 

In this PR I've tried to see what happens if we just keep this approach and use it for arrays as well. This is not ideal but IMO better then not supporting hidden fields on arrays at all. 

I'll add some comments to the code for more explanation.

### DIFF
--- a/playground/samples/hidden.js
+++ b/playground/samples/hidden.js
@@ -1,0 +1,44 @@
+module.exports = {
+  schema: {
+    type: "object",
+    properties: {
+      visible: {
+        type: "string",
+        title: "Inspect elements below",
+      },
+      hiddenString: {
+        type: "string",
+        title: "First name",
+      },
+      hiddenNumber: {
+        title: "Number",
+        type: "number"
+      },
+      hiddenArray: {
+        type: "array",
+        title: "A hidden list",
+        items: {
+          type: "string",
+          enum: ["foo", "bar", "baz"],
+        }
+      },
+    }
+  },
+  uiSchema: {
+    hiddenString: {
+      "ui:widget": "hidden"
+    },
+    hiddenNumber: {
+      "ui:widget": "hidden"
+    },
+    hiddenArray: {
+      "ui:widget": "hidden"
+    }    
+  },
+  formData: {
+    visible: 'To see some hidden items...',
+    hiddenString: 'John Doe',
+    hiddenNumber: 42,
+    hiddenArray: ["foo", "bar"]
+  }
+};
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -5,6 +5,7 @@ import {
   isMultiSelect,
   isFilesArray,
   isFixedItems,
+  isHidden,
   allowAdditionalItems,
   optionsList,
   retrieveSchema,
@@ -16,6 +17,7 @@ import {
 import SelectWidget from "./../widgets/SelectWidget";
 import FileWidget from "./../widgets/FileWidget";
 import CheckboxesWidget from "./../widgets/CheckboxesWidget";
+import HiddenWidget from "../widgets/HiddenWidget";
 
 
 function ArrayFieldTitle({TitleField, idSchema, title, required}) {
@@ -140,6 +142,9 @@ class ArrayField extends Component {
 
   render() {
     const {schema, uiSchema} = this.props;
+    if (isHidden(uiSchema)) {
+      return this.renderHidden();
+    }
     if (isFilesArray(schema, uiSchema)) {
       return this.renderFiles();
     }
@@ -376,6 +381,17 @@ class ArrayField extends Component {
           : null
         }
       </div>
+    );
+  }
+
+  renderHidden() {
+    const {idSchema} = this.props;
+    const {items} = this.state;
+    return (
+      <HiddenWidget
+        id={idSchema && idSchema.$id}
+        value={items}
+      />
     );
   }
 }
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -2,6 +2,7 @@ import React, { PropTypes } from "react";
 
 import {
   isMultiSelect,
+  isHidden,
   retrieveSchema,
   getDefaultRegistry,
   isFilesArray
@@ -167,7 +168,7 @@ function SchemaField(props) {
   const description = props.schema.description || schema.description;
   const errors = errorSchema.__errors;
   const help = uiSchema["ui:help"];
-  const hidden = uiSchema["ui:widget"] === "hidden";
+  const hidden = isHidden(uiSchema);
   const classNames = [
     "form-group",
     "field",
--- a/src/utils.js
+++ b/src/utils.js
@@ -53,6 +53,7 @@ const altWidgetMap = {
   },
   array: {
     checkboxes: CheckboxesWidget,
+    hidden: HiddenWidget
   }
 };
 
@@ -257,6 +258,10 @@ export function isFixedItems(schema) {
   return Array.isArray(schema.items) &&
          schema.items.length > 0 &&
          schema.items.every(item => isObject(item));
+}
+
+export function isHidden(uiSchema) {
+  return uiSchema["ui:widget"] === "hidden";
 }
 
 export function allowAdditionalItems(schema) {
--- a/test/uiSchema_test.js
+++ b/test/uiSchema_test.js
@@ -944,8 +944,10 @@ describe("uiSchema", () => {
           foo: ["foo", "bar"]
         }});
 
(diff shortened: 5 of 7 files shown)


playground/samples/hidden.js
```diff
@@ -0,0 +1,44 @@
...
+      "ui:widget": "hidden"
+    }    
+  },
+  formData: {
+    visible: 'To see some hidden items...',
+    hiddenString: 'John Doe',
+    hiddenNumber: 42,
+    hiddenArray: ["foo", "bar"]
+  }
+};
```
tiemevanveen: Added a "Hidden" section to the samples to demonstrate things

src/utils.js
```diff
@@ -53,6 +53,7 @@ const altWidgetMap = {
   },
   array: {
     checkboxes: CheckboxesWidget,
+    hidden: HiddenWidget
```
tiemevanveen: I'm not sure this was needed here... but might be if my implementation is not correct

src/components/fields/ArrayField.js
```diff
@@ -378,6 +383,17 @@ class ArrayField extends Component {
...
+  renderHidden() {
+    const {idSchema} = this.props;
+    const {items} = this.state;
+    return (
+      <HiddenWidget
+        id={idSchema && idSchema.$id}
+        value={items}
+      />
+    );
+  }
```
tiemevanveen: This was the easiest way to render the HiddenWidget instead of the array one. But it's done a bit differently than the other widgets (which are declared in the `altWidgetMap` in utils right?).

src/components/fields/SchemaField.js
```diff
@@ -167,7 +168,7 @@ function SchemaField(props) {
   const description = props.schema.description || schema.description;
   const errors = errorSchema.__errors;
   const help = uiSchema["ui:help"];
-  const hidden = uiSchema["ui:widget"] === "hidden";
+  const hidden = isHidden(uiSchema);
```
tiemevanveen: Not really necessary but this keeps the checking code in one place.

test/uiSchema_test.js
```diff
@@ -909,6 +909,55 @@ describe("uiSchema", () => {
...
+          .to.have.length.of(1);
+      });
+
+      it("should support formData", () => {
+        const {node} = createFormComponent({schema, uiSchema, formData: {
+          foo: ["foo", "bar"]
+        }});
+
+        expect(node.querySelector("[type=hidden]").value)
+          .eql(["foo", "bar"]);
```
tiemevanveen: This test will fail: 

```
     AssertionError: expected 'foo,bar' to deeply equal [ 'foo', 'bar' ]
```

tiemevanveen:
So as you can see the tests still fail. This is caused by the value of the hidden input becoming `foo,bar` instead of `['foo', 'bar']. I'm not sure if there's a way around that.

An alternative would be not using the HiddenWidget but just hiding the array list. Would that work with css `visibility:hidden` for example?

Or maybe adding multiple hidden input's to the DOM? 

```
function HiddenWidget({id, value}) {
  if (Array.isArray(value)) {
    return (
      <div id={id}>
        {value.map(val => <input type="hidden" name={`${id}[]`} value={val} />)}
      </div>
    );
  }

  return (
    <input type="hidden" id={id} value={typeof value === "undefined" ? "" : value} />
  );
}
```

Something like this might work but would need more work and would require the HiddenWidget to become a more complex component.. 

Any idea's ?

n1k0:
So we've spend some time toying around with a simple idea, which is to simply avoid rendering anything as a field when a field is marked as hidden. Here's the diff against latest master:

```diff
diff --git a/src/components/fields/SchemaField.js b/src/components/fields/SchemaField.js
index b5b0ba1..617e041 100644
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -32,6 +32,9 @@ function getFieldComponent(schema, uiSchema, fields) {
   if (typeof field === "string" && field in fields) {
     return fields[field];
   }
+  if (uiSchema["ui:widget"] === "hidden") {
+    return () => <div/>;
+  }
   return COMPONENT_TYPES[schema.type] || UnsupportedField;
 }
```

It seems to address the issue quite nicely, but could you please keep test it and check if this solves the issue for you as well?

If it works just fine, could you update your PR leveraging this approach? 

Thanks!

tiemevanveen:
I think that would work. But removing the input element from the DOM would mean that submitting the form to the server **without** react would cause the hidden inputs not to be submitted right? Isn't that a problem?

n1k0:
> submitting the form to the server without react would cause the hidden inputs not to be submitted

I'm confused, which server are we talking about? This is a client-side only library, we don't actually post form data to any server URL, instead we catch the submit event and call a callback so the UI knows about the submitted data.

If you're suggesting we should support server side rendering, I'm afraid to say, that's not at all one of our goal here :)

But maybe I've just misunderstood something?

JeffHeard:
I think what the poster's saying is that if the data isn't captured in the
DOM as "<input type='hidden' />", then isn't there more risk of using the
form in such a way as to lose data?

I think since 'formData' is what a person should be submitting (if you're
submitting anything), this really shouldn't be an issue.  That would still
contain all the same data, after all, correct?  Even if it's not reflected
in the DOM?

Also, removing the elements from the DOM entirely would have the nice
effect of being more secure. Otherwise there's a good chance a user could
simply open up their browser's debugging tools and edit values.

tiemevanveen:
> then isn't there more risk of using the form in such a way as to lose data?

Yeah well I thought after going through the tests that you also want to support the scenario where the form is just submitted via a POST or GET. So what if no `onSubmit` is provided, should the form still work? 

A good case for this is indeed server side rendering. Is that currently not supported? An example would be a site that's rendered by the server and browsers without javascript. The form would still submit fine (if the backend supports the call). But with removing the elements from the DOM those hidden values would not be submitted.. 

If you don't consider this a problem (**and I would agree this is indeed not a problem at all!**) the pr as I submitted would work fine and can be merged. The only thing is that the test needs to be adjusted. Please let me know if you want me to do that. 

Regarding the security of the form. I don't think the backend should ever rely on the submitted data not being tempered with so I don't think that will ever be an issue. I mean users can also set breakpoints on the form submit or xhr requests and insert malicious data at that point right?

tiemevanveen:
btw, the only change needed to fix the tests is change this new test:

```diff
diff --git a/test/uiSchema_test.js b/test/uiSchema_test.js
index 8f766c6..d0ff148 100644
--- a/test/uiSchema_test.js
+++ b/test/uiSchema_test.js
@@ -944,8 +944,10 @@ describe("uiSchema", () => {
           foo: ["foo", "bar"]
         }});

+        // array values will be stored as comma seperated list in the DOM,
+        // it's the best we can do (pull request #324).
         expect(node.querySelector("[type=hidden]").value)
-          .eql(["foo", "bar"]);
+          .eql("foo,bar");
       });

       it("should map widget value to a typed state one", () => {
```

tiemevanveen:
Fixed the test. Let me know if you want something else.

n1k0:
So if `"foo,bar"` is the string representation of `[1, 2]` I'm afraid the one for `[{x: 1}, {x: 2}]` will be `["[Object object]", "[Object object]"]`.

Could you please check that? Thanks

tiemevanveen:
But in which case would the value of the array be that?

n1k0:
Maybe I'm missing something here, but jsonschema arrays can contain other stuff than numbers and strings.

```js
const schema = {
  type: "array",
  items: {
    type: "object",
      properties: {
        x: {
          type: "number"
        }
      }
    }
  }
}
```

As an example, what happens if you try to hide the `tasks` array in the `Nested` example in the playground?

I'm genuinely asking, I hadn't a lot of CPU cycles dedicated to this project lately, so bear with me :)

tiemevanveen:
> I hadn't a lot of CPU cycles dedicated to this project lately

Sorry, me neither, PR is from a month ago ;) 

Yeah I think you're right, I think that will cause `["[Object object]", "[Object object]"]` in the DOM. You could consider serializing it first.. But it depends a bit on what the goal is. See my comment here https://github.com/mozilla-services/react-jsonschema-form/pull/324#issuecomment-252003463

leplatrem:
I tried to catch up on this PR, but I'm a bit lost. Could you summarise what's left to be decided/done to land it ?
Thanks!

n1k0:
Let's just say this hasn't seen much demand nor triggered much involvement from the community since last October, so I'm gonna consider this not really needed and close this PR. Feel free to reopen if you feel bad about this decision.

PR #325, opened 2016-09-16
Radio Support for integers and numbers

tiemevanveen:
Added an example to the numbers examples as well so you can check it out. 

This works only if user provided an enum for that integer or number. 

This could optionally be added to this pr. Something like this in `getAlternativeWidget` could check that: 

```js
if(
    (schema.type === 'integer' || schema.type === 'integer')
  && widget === 'radio' && !widgetOptions.enumOptions
) {
  throw new Error(`No radio possible without enumOptions`);
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,7 @@ Please note that while standardized, `datetime-local` and `date` input elements 
 
   * `updown`: an `input[type=number]` updown selector;
   * `range`: an `input[type=range]` slider;
+  * `radio`: a radio button group with enum values. **can only be used when `enum` values are specified for this input**
   * by default, a regular `input[type=text]` element is used.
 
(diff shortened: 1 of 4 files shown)


n1k0:
A very useful addition, thank you. I've deployed the patch to gh-pages, works just fine; test and code look good 👍 

I'm torn adding specific checks as you suggest in the description, though if they're extracted to a dedicated, separated function it's probably a very worthy addition.

tiemevanveen:
> I'm torn adding specific checks as you suggest in the description, though if they're extracted to a dedicated, separated function it's probably a very worthy addition.

Yeah that needs to go in a seperate function. I can work on it

leplatrem:
Do not hesitate to open a new PR/issue with the validation/check that you suggested.

tiemevanveen:
Great, thanks. Will do, some day..

n1k0:

tiemevanveen:
\0/

PR #334, opened 2016-10-06
Properly handle numbers that require trailing zeroes

israelshirk:
## Situation
We ran into this issue entering dollar amounts where the user intends to enter a monetary value with several trailing digits to match locale - such as "12.90" or "12.00".

## Current behavior
Current behavior is that upon entering the trailing zero, it is immediately stripped and shows the user "12.9" or "12", respectively.

## Desired behavior
Show the user what they entered after validating - "12.90" or "12.00", respectively.


israelshirk:
(Update 8:40AM US/Denver - resolved an issue in which the test was in the wrong section)

n1k0:
Great patch, thanks a lot for your contribution.

n1k0:

Issue #339, opened 2016-10-12
Loses characters on Internet Explorer 11


israelshirk:
### Description

Letters of input show up and then rapidly disappear at random when typing quickly on IE 11.
### Steps to Reproduce

When using Internet explorer 11, locally using modern.ie VM and interacting with the first text box on https://mozilla-services.github.io/react-jsonschema-form/ .  A client initially ran into the issue running IE 11 on her desktop.

Quickly typed in: `This is steven thompson`
#### Expected behavior

Should have gotten `This is steven thompson` in the text box.
#### Actual behavior

<img width="670" alt="screen shot 2016-10-12 at 4 17 14 pm" src="https://cloud.githubusercontent.com/assets/1100234/19329715/80e8b30a-9097-11e6-9a64-5f95c74886ec.png">
### Version

Current

israelshirk:
Have confirmed that it's a race condition between receiving the next keypress and the response from onChange() on upstream components propagating back down as formData.  Going to take a stab at unidirectional data flow by propagating the events upstream to the validator and rejecting at the validation stage while keeping the values local inside the components and still using formData to populate initial state.

israelshirk:
Verified that it's a race condition getting data through the onChange propagators out to <Form />; then propagating everything back up through the chain.  Profiling shows it's creating around 15k new React elements in my case - seems a little much.  Going to move those out of the render functions and re-render as needed - holler if you'd like a different approach.

knilink:
Same issue here in meteor and browsing with chrome.
I just realized that the entry point of the package was "lib/index.js" which was not optimized.
Using `import Form from 'react-jsonschema-form/dist/react-jsonschemaform';` instead solve my issue.

n1k0:
> which was not optimized.

What do you mean, not optimized?

n1k0:
@israelshirk is the problem occuring with `liveValidate` set to false (which should be dthe default)?

israelshirk:
AFK, I'll check this afternoon.  Thanks, Israel

knilink:
There were 2 build script according to package.json

`"build:lib": "rimraf lib && NODE_ENV=production babel -d lib/ src/"`

`"build:dist": "rimraf dist && NODE_ENV=production webpack --config webpack.config.dist.js --optimize-minimize"`

If I understand correctlly, the 'lib' one was just build ES6 into plain javascript and it's still human readable while the 'dist' one will bundle and optimize the code which brought a better performance.

[...]

PR #340, opened 2016-10-13
Cache object fields to improve performance S.T. IE 11 works decently

israelshirk:
Ref #339


israelshirk:
Skirts the issue.  Closing.

PR #341, opened 2016-10-13
Pass data to FieldTemplate as strings instead of as React components

olzraiti:
This adds flexibility.

An example use case would be when you want the help block to be a tooltip of the label. Then you don't want the help object to be a React component, but a string instead so you can render a tooltip for the label with the help string as the content.

I think this is also more consistent, since we are rendering the label already in TemplateField instead of passing it as a React component.

I exposed `Label`, `Help` and `ErrorList` components from SchemaField with `export`, because they are probably useful for writing a custom TemplateField - should we put these to a separate file?

### DIFF
--- a/README.md
+++ b/README.md
@@ -575,7 +575,7 @@ A field template is basically a React stateless component being passed field-rel
 
 ```jsx
 function CustomFieldTemplate(props) {
-  const {id, classNames, label, help, required, description, errors, children, fields} = props;
+  const {id, classNames, label, help, required, description, errors, children} = props;
   return (
     <div className={classNames}>
       <label htmlFor={id}>{label}{required ? "*" : null}</label>
@@ -614,7 +614,7 @@ The following props are passed to a custom field template component:
 - `fields`: An array containing all Form's fields including your [custom fields](#custom-field-components) and the built-in fields.
 - `formContext`: The `formContext` object that you passed to Form.
 
-> Note: you can only define a single field template for a form. If you need many, it's probably time to look for [custom fields](#custom-field-components) instead.
+> Note: you can only define a single field template for a form. If you need many, it's probably time to look at [custom fields](#custom-field-components) instead.
 
 ### Custom widgets and fields
 
--- a/README.md
+++ b/README.md
@@ -574,15 +574,21 @@ To take control over the inner organization of each field (each form row), you c
 A field template is basically a React stateless component being passed field-related props so you can structure your form row as you like:
 
 ```jsx
+
+import {Help, ErrorList} from "react-jsonschema-form/lib/components/fields/SchemaField"; // You can import also the default 'Label' from SchemaField. Below we use a custom label.
+
 function CustomFieldTemplate(props) {
   const {id, classNames, label, help, required, description, errors, children} = props;
+
+  const {DescriptionField} = fields;
+
   return (
     <div className={classNames}>
       <label htmlFor={id}>{label}{required ? "*" : null}</label>
-      {description}
+      <DescriptionField id={id + "__description"} description={description} />
       {children}
-      {errors}
-      {help}
+      <ErrorList errors={errors} />
+      <Help help={help} />
     </div>
   );
 }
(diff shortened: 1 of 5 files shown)


n1k0:
This breaks backward compatibility, could you please add a check to test if the passed value is a string or a component so we can keep previous API usage working?

olzraiti:
Good point! Will do.

olzraiti:
Hmm... The default field template is working fine (with backward compatibility) for users who don't use a custom FieldTemplate. However, if a user passes a custom FieldTemplate to the Form, we can't know if he wants the `help`, `description` or `errors` parameters as strings or as React components. There's no way to check this AFAIK.

To keep backward compatibility we could add a parameter to the Form which would make the FieldTemplate receive JS objects (string for `description` & `help`, array of string to `errors`) instead of React components. By default the FieldTemplate would receive React components.

What do you think?

n1k0:
> often it is a good idea to make breaking API changes as soon as possible if changes are needed

I tend to agree, especially as we've not tagged a 1.0 just yet :D

Okay, could you now please update the Field Templates section of the documentation to reflect the changes? Thanks!

README.md
```diff
@@ -574,15 +574,21 @@ To take control over the inner organization of each field (each form row), you c
 A field template is basically a React stateless component being passed field-related props so you can structure your form row as you like:
 
 ```jsx
+
+import {Help, ErrorList} from "react-jsonschema-form/lib/components/fields/SchemaField"; // You can import also the default 'Label' from SchemaField. Below we use a custom label.
```
n1k0: Hmm. People loading the lib optimized bundle from a script tag won't have `Help` and `ErrorList` exposed; that's a problem.

README.md
```diff
@@ -574,15 +574,21 @@ To take control over the inner organization of each field (each form row), you c
...
+
 function CustomFieldTemplate(props) {
   const {id, classNames, label, help, required, description, errors, children} = props;
+
+  const {DescriptionField} = fields;
```
n1k0: Where does `fields` come from?

README.md
olzraiti: How about putting `Help` and `ErrorList` to their own files? There exists a `ErrorList` already in src/components, so this list could be named `InlineErrorList`.

n1k0: That would solve the issue with people loading the lib through from a script tag:

```js
 <script src="https://unpkg.com/react-jsonschema-form/dist/react-jsonschema-form.js"></script>
``` 

There you have no access to individual modules, and I'm uncomfortable attaching them to the root exposed object or anything like that.

olzraiti: I'm working on exposing the raw values along with the React components - it's a good idea and implementation is easy. So maybe we'll just leave the `Help` and `ErrorList` unimportable since the TemplateField will provide them.

test/Form_test.js
```diff
@@ -72,15 +69,29 @@ describe("Form", () => {
...
+        errors,
+        rawErrors,
+        children,
+      } = props;
       return (
         <div className={"my-template " + classNames}>
           <label htmlFor={id}>{label}{required ? "*" : null}</label>
-          <DescriptionField id={`${id}__description`} description={description} />
+          {description}
+          <span id={`${id}__raw_description`}>{`${rawDescription} rendered from the raw format`}</span>
```
n1k0: I think we should keep the `${id}__description` format here, e.g. so users can style both using a same CSS rule.

test/Form_test.js
```diff
@@ -70,14 +69,29 @@ describe("Form", () => {
...
+        children,
+      } = props;
       return (
         <div className={"my-template " + classNames}>
           <label htmlFor={id}>{label}{required ? "*" : null}</label>
           {description}
+          <span id={`${id}__raw_description`}>{`${rawDescription} rendered from the raw format`}</span>
           {children}
           {errors}
+          {rawErrors ? <ul>{rawErrors.map((error, i) => <li key={i} className="raw-error">{error}</li>)}</ul> : null}
```
n1k0: Nit: Would you mind spreading this line onto multiple ones to improve legibility?

test/Form_test.js
```diff
@@ -70,14 +69,29 @@ describe("Form", () => {
...
+        rawDescription,
+        errors,
+        rawErrors,
+        children,
+      } = props;
       return (
         <div className={"my-template " + classNames}>
           <label htmlFor={id}>{label}{required ? "*" : null}</label>
           {description}
+          <span id={`${id}__raw_description`}>{`${rawDescription} rendered from the raw format`}</span>
```
n1k0: Nit: We should stick to a single id format here I think, `${id}__raw_description` is fine.

test/Form_test.js
```diff
@@ -70,14 +69,29 @@ describe("Form", () => {
...
           {children}
           {errors}
+          {rawErrors ? <ul>{rawErrors.map((error, i) => <li key={i} className="raw-error">{error}</li>)}</ul> : null}
           {help}
+          <span id={`${id}__raw_help`}>{`${rawHelp} rendered from the raw format`}</span>
```
n1k0: Nit: Same remark, we should stick to a single id format here I think.

n1k0:
Thank you!

n1k0:

pabloen:
Hi, I'm finding propTypes validations errors after updating to v0.41.0 and I think it's related to this issue.

My use case is the following, I sometimes need to insert simple HTML in the description for specific fields (sometimes for simple formatting, adding icons and minor things like that).

Example:

```javascript
import React from 'react';

const schema = {
  type: 'string',
  description: (
    <div>
      Beware of something <strong>really important</strong> about this field.
    </div>
  )
};
```

As you can see, I'm passing a React element (not a React Component) as the description.

But now, although it's all working just fine I'm getting this all over the place:

```Warning: Failed prop type: Invalid prop `rawDescription` of type `object` supplied to `DefaultTemplate`, expected `string`.```

A simple approach to solve this would be to relax the propTypes and accept React elements as valid values for `rawDescription` prop. I am wondering if is it really okay to asume or even enforce the description to be a simple string. I understand the example that originated this change, but I think that example would work too if we accept React Elements as valid description values.

Am I missing something?

n1k0:
@pabloen No, and I'm using this too. React elements as descriptions are especially useful when you want to add links to texts. I'll update the proptypes to accept elements too.

pabloen:
Great, thank you!

n1k0:
Just released v0.41.1 :)

revolunet:
Hi,

I'm wondering what is the correct to pass custom `ErrorList` component to the fields if we need customisation ?

n1k0:
@revolunet you can use a [custom field template](https://github.com/mozilla-services/react-jsonschema-form/#field-template). Sample jsfiddle https://jsfiddle.net/hdp1kgn6/1/

revolunet:
thanks 👍

PR #343, opened 2016-10-14
fix undefined value in formData for fileInput that cause incorrect error display

hengchenvoon:
I noticed that in the function `validateFormData()` in `validate.js` will receive `formData` objects that contains an undefined value, which will occur if the form consists of FileInput field that have no file selected.

`jsonValidate` will not map the error to the field instance correctly if the value is `undefined`, result in it mapping the error to the form itself.

My changes should change all `undefined` value to `null` value in the formData, allowing `jsonValidate` to do its job correctly.

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -115,6 +115,8 @@ function unwrapErrorHandler(errorHandler) {
  * will be used to add custom validation errors for each field.
  */
 export default function validateFormData(formData, schema, customValidate) {
+  formData = undefinedToNullRecursively(formData);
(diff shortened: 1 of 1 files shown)


leplatrem:
I don't think that's the way to go. Why not fix the `FileWidget` instead?

leplatrem:
I'm closing this!

Do not hesitate to reopen or give another try with a different approach, we'd be glad to help you ship a fix for that! Thanks again!

PR #355, opened 2016-10-20
Added ref to FileWidget.

dehli:

With this change, I've added a reference to the `input` element in `FileWidget`. The reason I did this is because I've created a custom widget which is composed of a hidden `FileWidget` and a nice looking button (that way you don't need to have the browser styled file button).

By exposing `inputRef`, I can progammatiacally call the `input`'s choose file by calling: `fileWidgetRef.inputRef.click()` whenever the fancy button is clicked.

I can add documentation and a playground if you'd like.

Also this library is great, thanks! 😄 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,7 +380,7 @@ const schema = {
 
 ##### File widget input ref
 
-The included `FileWidget` exposes its `<input type="file"></input>` via `inputRef`.
+The included `FileWidget` exposes a reference to the `<input type="file" />` element node as an `inputRef` component property.
 
 This allows you to programmatically trigger the browser's file selector which can be used in a custom file widget.
 
--- a/README.md
+++ b/README.md
@@ -378,6 +378,12 @@ const schema = {
 
 > Note that storing large dataURIs into form state might slow rendering.
 
+##### File widget input ref
+
+The included `FileWidget` exposes its `<input type="file"></input>` via `inputRef`.
+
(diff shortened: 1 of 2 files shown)


leplatrem:
I wouldn't be against if this it really serves this use-case. Indeed, browse buttons are ugly. 

@n1k0 what do you think?

n1k0:
> @n1k0 what do you think?

I don't have a string opinion about it, though I've read bad things about react refs in the past, while I've never used them myself. 

The patch looks small and trivial enough to land it it solves people's problems I think :)

n1k0:
Amybe adding a line or two of documenting this would be great.

dehli:
Thanks! @n1k0 @leplatrem I've added some documentation to the `README.md`. Let me know what you think. :)

README.md
```diff
@@ -378,6 +378,12 @@ const schema = {
 
 > Note that storing large dataURIs into form state might slow rendering.
 
+##### File widget input ref
+
+The included `FileWidget` exposes its `<input type="file"></input>` via `inputRef`.
```
n1k0: Nit: *The included `FileWidget` exposes a reference to the `<input type="file" />` element node as a `inputRef component property.*

n1k0:
Thank you! This will be released anytime soon next week :)

n1k0:

jariwalabhavesh:
@dehli:
Can you please provide example how can i access refs in form that we provide in file widget.?

PR #358, opened 2016-10-24
Add react-bootstrap support

### LINKED ISSUE #356: Render react-bootstrap instead of Bootstrap markup
### Prerequisites
- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme)
### Description

I would like to render all fields with [react-bootstrap](http://react-bootstrap.github.io/) instead of Bootstrap markup since I don't want to let jQuery and Bootstrap.js deal with DOM changes next to React. So, I plan to rework all pure components in order to support react-bootstrap.
I'm aksing myself if you would like this kind of changes too or if I fork this project? 
### Version

0.40.0

frassinier:

I've added [react-bootstrap](https://react-bootstrap.github.io) support, 
The main goal is to not have to manage any Bootstrap markup. The switch, between Bootstrap v3 to v4, will be easier to deal with. 
The second one is to not have to include jQuery if we add some specific JS components (like Tooltip or Dropdown) 

Will closes #356 
### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -1,4 +1,4 @@
-import React, { Component, PropTypes } from "react";
+import React, {Component, PropTypes} from "react";
 
 import Button from "react-bootstrap/lib/Button";
 
@@ -40,9 +40,9 @@ export default class Form extends Component {
     const edit = typeof props.formData !== "undefined";
     const liveValidate = props.liveValidate || this.props.liveValidate;
     const mustValidate = edit && !props.noValidate && liveValidate;
-    const { definitions } = schema;
+    const {definitions} = schema;
     const formData = getDefaultFormState(schema, props.formData, definitions);
-    const { errors, errorSchema } = mustValidate ?
+    const {errors, errorSchema} = mustValidate ?
       this.validate(formData, schema) : {
         errors: state.errors || [],
         errorSchema: state.errorSchema || {}
@@ -65,13 +65,13 @@ export default class Form extends Component {
   }
 
   validate(formData, schema) {
-    const { validate } = this.props;
+    const {validate} = this.props;
     return validateFormData(formData, schema || this.props.schema, validate);
   }
 
   renderErrors() {
-    const { status, errors } = this.state;
-    const { showErrorList } = this.props;
+    const {status, errors} = this.state;
+    const {showErrorList} = this.props;
 
     if (status !== "editing" && errors.length && showErrorList != false) {
       return <ErrorList errors={errors}/>;
@@ -79,12 +79,12 @@ export default class Form extends Component {
     return null;
   }
 
-  onChange = (formData, options = { validate: false }) => {
+  onChange = (formData, options={validate: false}) => {
     const mustValidate = !this.props.noValidate && (this.props.liveValidate || options.validate);
-    let state = { status: "editing", formData };
+    let state = {status: "editing", formData};
     if (mustValidate) {
-      const { errors, errorSchema } = this.validate(formData);
-      state = { ...state, errors, errorSchema };
+      const {errors, errorSchema} = this.validate(formData);
+      state = {...state, errors, errorSchema};
     }
     setState(this, state, () => {
       if (this.props.onChange) {
@@ -95,12 +95,12 @@ export default class Form extends Component {
 
   onSubmit = (event) => {
     event.preventDefault();
-    this.setState({ status: "submitted" });
+    this.setState({status: "submitted"});
 
     if (!this.props.noValidate) {
-      const { errors, errorSchema } = this.validate(this.state.formData);
+      const {errors, errorSchema} = this.validate(this.state.formData);
       if (Object.keys(errors).length > 0) {
-        setState(this, { errors, errorSchema }, () => {
+        setState(this, {errors, errorSchema}, () => {
           if (this.props.onError) {
             this.props.onError(errors);
           } else {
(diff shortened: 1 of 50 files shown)


src/components/Form.js
```diff
@@ -38,9 +40,9 @@ export default class Form extends Component {
     const edit = typeof props.formData !== "undefined";
     const liveValidate = props.liveValidate || this.props.liveValidate;
     const mustValidate = edit && !props.noValidate && liveValidate;
-    const {definitions} = schema;
+    const { definitions } = schema;
```
leplatrem: Thanks for this! 

Couldn't we add it to eslint config to make sure it remains consistent with time?

src/components/widgets/CheckboxWidget.js
```diff
@@ -12,18 +13,19 @@ function CheckboxWidget({
...
+    <FormGroup>
+      <Checkbox
+        type="checkbox"
+        id={id}
+        checked={typeof value === "undefined" ? false : value}
+        required={required}
+        disabled={disabled}
+        autoFocus={autofocus}
+        onChange={(event) => onChange(event.target.checked)}
+      >
```
leplatrem: nitpick: I think the closing `>` should go at the end of the previous line

src/components/widgets/CheckboxesWidget.js
```diff
@@ -14,34 +14,34 @@ function deselectValue(value, selected) {
...
-                }} />
-              <strong>{option.label}</strong>
-            </label>
-          </div>
+          <Checkbox
+            key={index}
+            id={`${id}_${index}`}
+            checked={checked}
+            disabled={disabled}
+            inline={inline}
```
leplatrem: For consistency, shouldn't it be like this ?


```
  <Checkbox key={index}
            id={`${id}_${index}`}
            checked={checked}
            disabled={disabled}
            inline={inline}
```

src/components/widgets/CheckboxesWidget.js
```diff
@@ -14,34 +14,34 @@ function deselectValue(value, selected) {
...
+            autoFocus={autofocus && index === 0}
+            onChange={(event) => {
+              const all = enumOptions.map(({ value }) => value);
+              if (event.target.checked) {
+                onChange(selectValue(option.value, value, all));
+              } else {
+                onChange(deselectValue(option.value, value));
+              }
+            }}
+          >
```
leplatrem: ditto for closing `>`

src/components/widgets/FileWidget.js
```diff
@@ -94,22 +96,21 @@ class FileWidget extends Component {
...
-            id={id}
-            type="file"
-            disabled={readonly || disabled}
-            onChange={this.onChange}
-            defaultValue=""
-            autoFocus={autofocus}
-            multiple={multiple} />
-        </p>
-        <FilesInfo filesInfo={filesInfo} />
+        <Input
```
leplatrem: nit: ditto for attribute indentation

n1k0:
> As for the introduction of react-bootstrap, I'd let @n1k0 decide

I'm undecided. I don't clearly see what this brings to the table. Also, I fear this will bloat bundle filesizes by quite a bunch; could we get metrics about this?

Also, I'm not in favor of changing the coding style like this.

frassinier:
> I'm undecided. I don't clearly see what this brings to the table.

I just proprose a way to not dealing with Bootstrap markup, to not dealing with ARIA attributes and not using `<strong/>` inside form labels, for example, which is not really semantically correct.

Ok for fixing code style by editing `.eslintrc` and `.editorconfig` to fit to your default?

n1k0:
These are good points, thank you for clarifying.

> Also, I fear this will bloat bundle filesizes by quite a bunch; could we get metrics about this?

This question still stands though.

n1k0:
> Ok for fixing code style by editing .eslintrc and .editorconfig to fit to your defaults?

That would indeed be a great contribution, but I'd rather see it done as a separate PR please.

frassinier:
I will produce a before/after to measure the size of the final bundle. :ok_hand:

frassinier:
@n1k0 After [dealing with imports](http://react-bootstrap.github.io/getting-started.html), the final bundle size is ~300ko when without this dependency it was ~100ko.

BTW, I'm asking myself why there is a NPM script `dist` which transpiles ES6 to lib/ and then a Webpack task which transpiles ES6 to dist/? 
Perhaps, it could be better to do: 
```javascript
var webpack = require("webpack");

module.exports = {
  cache: true,
  context: __dirname + "/lib",
  entry: "./index.js",
  output: {
    path: "./dist",
    publicPath: "/dist/",
    filename: "react-jsonschema-form.js",
    library: "JSONSchemaForm",
    libraryTarget: "umd"
  },
  plugins: [
    new webpack.DefinePlugin({
      "process.env": {
        NODE_ENV: JSON.stringify("production")
      }
    })
  ],
  devtool: "source-map",
  externals: {
    react: {
      root: "React",
      commonjs: "react",
      commonjs2: "react",
      amd: "react"
    }
  }
};
```

israelshirk:
Commentary from a different perspective: I've done the same thing, but from the perspective of adding both bootstrap and react-native support (both for iOS and Android, on two different frameworks).

It'd be much better to add this as a pluggable SchemaField than to put it into the core - I believe that's a large part of the intent behind SchemaField.

frassinier:
Do you have any link?

israelshirk:
@frassinier it's all proprietary :/ - I'm releasing the core mods as I go though

n1k0:
> After dealing with imports, the final bundle size is ~300ko when without this dependency it was ~100ko.

That sounds like a nogo to me.

> BTW, I'm asking myself why there is a NPM script dist which transpiles ES6 to lib/ and then a Webpack task which transpiles ES6 to dist/? 

To allow leveraging the lib API from a npm packaged installation, which won't transpile es6 code by itself (nodejs still needs transpiled code, at least for module import compat).

frassinier:
I close to continue with #366

PR #359, opened 2016-10-24
Improved documentation on form initialization

phidias:

Through the docs I got the impression that `formData` prop is used for initialization only, which led to a hard-to-detect bug that was happening when the parent element was re-rendering. So I decided to update the documentation so it's more clear. Please feel free to modify the texts that I wrote so they fit well with the rest of the documentation

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ render((
 ), document.getElementById("app"));
 ```
 
-WARNING: The `formData` field is **not** used only for initialization but it sets the data of the form if your wrapping component rerenders. Therefore make sure you listen to the `onChange` event and update the data you pass to the `formData` field.
+Therefore, if you have situations where your parent component can re-render, make sure you listen to the `onChange` event and update the data you pass to the `formData` field.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -150,6 +150,8 @@ render((
 ), document.getElementById("app"));
 ```
 
+WARNING: The `formData` field is **not** used only for initialization but it sets the data of the form if your wrapping component rerenders. Therefore make sure you listen to the `onChange` event and update the data you pass to the `formData` field.
```
leplatrem: I would say: *Therefore, if you have situations where your parent component can re-render, make sure you listen ...* 

Because in our use-cases we never had to do that, and it can be confusing to read that at first. What do you think?

phidias: Sounds good. Should it be added before the example along with the text on the top or leave it here under the example?

phidias: Ok done

leplatrem:
Merged in cc11737b1399748

PR #363, opened 2016-10-26
Migrate SchemaField to being an extensible React component

israelshirk:
### Motivation

Currently it's necessary to either completely rewrite or clone SchemaField to make a custom object type.  This allows one to add custom property types or override the default components in a much less heavy-handed way.  

Ideally, for the react-bootstrap example this could be implemented while maintaining compatibility with upstream updates.  Adding a new property would consist of something like this:

``` javascript
import SchemaField from 'react-jsonschema-form';
import specialField from 'components/specialField';

class myCustomSchemaField extends SchemaField {
   constructor() {
      super();
      this.OBJECT_FIELD_COMPONENTS['specialField'] = specialField;
   }
}

...

let registry = { 'SchemaField': myCustomSchemaField }
<Form registry={registry} ... />
```
### Note

This is just code I've ported over from an internal project and from the looks of #358 it may be useful to others to improve pluggability/replacement of the default types.
### Checklist

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -134,6 +134,39 @@ class SchemaField extends Component {
     return this.COMPONENT_TYPES[schema.type] || UnsupportedField;
   }
 
+  shouldDisplayLabel(schema, uiSchema) {
+    let displayLabel = true;
+    if (schema.type === "array") {
+      displayLabel = isMultiSelect(schema) || isFilesArray(schema, uiSchema);
+    }
+    if (schema.type === "object") {
+      displayLabel = false;
+    }
+    if (schema.type === "boolean" && !uiSchema["ui:widget"]) {
+      displayLabel = false;
+    }
+    if (uiSchema["ui:field"]) {
+      displayLabel = false;
+    }
+
+    return displayLabel;
+  }
+
+  buildFieldComponent(FieldComponent, props, schema, disabled, readonly, autofocus, formContext) {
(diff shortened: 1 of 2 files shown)


israelshirk:
Going to tack on some other OO inheritance goodies to make this work better for me...  One moment...

israelshirk:
Ok, I think that's all I would use.  We do some things that are based on an element's context within the form hierarchy (ie, multi-page forms with references in between) so it simplifies things quite a bit to have flexible access in there.

leplatrem:
Thanks for this! It's super important to make sure the lib is pluggable enough to remain small, and this seems to serve some various use cases :)

Some documentation about uses cases and a few snippets to get started would help though!

As well as some tests (at least eventually) :)

I would let @n1k0 give his opinion on your approach, especially the OO `buildFieldComponent()` and `buildFieldTemplate()` thing. Not sure, but maybe passing the sub-components via some props with default values would be more idiomatic no?

israelshirk:
Agreed on the buildFieldComponent and buildFieldTemplate, it feels awkward at the moment - if the approach seems worthy to you guys I'm happy to tweak it for a merge and include tests/etc.

frassinier:
@israelshirk I tried your snippet but I don't succeed :( by the way, I like the principle of registry!

n1k0:
Since v0.41.0 you can override all default fields and widgets, including object fields. I'm closing this.

Issue #354, opened 2016-10-20
Refactoring ideas regarding the template field, fields and widgets


olzraiti:
I have some refactoring ideas below. I could have written issues for each idea, but I think the issues are interconnected and it is easier to communicate the main goal in a single post.

The `fields` registry would contain only single functions for representing a single json schema [primitive type](http://json-schema.org/latest/json-schema-core.html#anchor8). The `fields` registry would then have a well defined purpose. I'm guessing this was the original purpose for the `fields` registry, but then DescriptionField and TitleField were added so users could control how they are rendered. Currently the DescriptionField and TitleField are not representing primivite schema types but data inside a schema. 

The TemplateField system could be extended so that it would handle also titles and descriptions and let the `fields` registry handle just the primitive types. After #341 has landed, the TemplateField can handle descriptions. Once it has landed, I suggest that we deprecate the DescriptionField (after #341, the DescriptionField will be useful only if you wan't to create a custom description but not a custom template field - this serves only little use because writing a template field isn't much more difficult or time consuming than writing a custom description field. The benefits are small compared to the complexity it adds).

To get rid of the TitleField also, we could add a template field system to the ArrayField and ObjectField. I'm looking at the code of ObjectField and ArrayField, and ObjectField looks like it could be pretty easily refactored to use a template field. ArrayField looks pretty complex, but it should be possible. I think a template field could maybe simplify ArrayField quite a lot (we could for example drop the reordering functionality and pass the `onChange()` function to the template field, and users who need the ordering could implement it themselves). The ArrayField contains now some hard coded bootstrap elements - the template field system would allow users to choose how they want to render array rows and buttons.

Another way to let users rewrite ArrayField easily would be to make it possible to override it through the `fields` property. I think a template field would be more usable for users, since most likely they want to just override how the component renders, not the well written and tested logic behind it.

Though allowing to override all fields and widgets would make sense in any case. There is discussion about that in #283 and #310. I suggested #310, and I still think it would be a good way to go - if a user wants to replace the CheckboxWidget, he could just provide a custom CheckboxWidget to the `widgets` property and the default BooleanField would render the provided CheckboxWidget - the user would again benefit from the well written and already tested logic.

These changes would make this library a flexible backbone with sane defaults that are overridable to the core in a way that let's you override only the relevant stuff (usually the React components rendering) if the logic is otherwise working OK.

n1k0:
[...]

Issue #310, opened 2016-08-30
Allow replacing hard coded fields & widgets through the registry


olzraiti:
I'd like to change every checkbox to some switch component. To achieve this, I'd have to provide a custom SchemaField to the form, which would be identical to the default SchemaField with the addition, that I'd have to import StringField from a custom StringField. That StringField would be identical the default StringField with the addition, that I'd have to import BooleanField from a custom BooleanField. That BooleanField would be identical to the default BooleanField with the addition, that I'd have to import CheckboxWidget from a custom CheckboxWidget. And finally, I could replace the checkbox with a switch component.

The other option would be to create a custom widget, and point every checkbox in uiSchema to use that custom component. But that wouldn't be a global solution.

What if I could just give the Form a custom CheckboxWidget in Form's `widgets`-property, and BooleanField would import the CheckboxWidget from the registry? It would be way simpler and I could drop like 1000 lines of copy pasted code. 

SchemaField, DescriptionField and TitleField can already be overridden through the registry. Maybe we should allow overriding every field/widget and always import from the registry?

**tl;dr** All hard coded field/widget components should import field/widget components from the registry. All fields/widgets should be possible to override through Form's `widgets` and `fields` properties.

olzraiti:
I have an implementation for this in my personal fork. I could create a pull request if approved.

(I have also suggested #205 and mentioned that I could implement it, but I'd personally give this one higher priority.)

AMongeEntytle:
+1 to this. I though this was something that should be really easy to implement (not only for widgets but also for fields), much more taking into account how amazing and configurable the library is ... so please let us know if your request is accepted !

olzraiti:
Closed by #371

PR #366, opened 2016-10-28
Be able to customize default widgets

frassinier:

Provide a registry service in order to let customizing default widgets (or whatever).
It needs some polish but it works.

That pull request is related to #358 and #363.
And because I don't want to deal with all logic by manipulating fields like with #283.  
### Demo

Let's try to implement `material-ui` to change our project playground aspect,
(Start by editing _app.js_)

``` javascript
// Example of default Checkbox widget override
import WidgetRegistry from "../src/registry/WidgetRegistry";
import MuiThemeProvider from "material-ui/styles/MuiThemeProvider";
import MaterialDesignCheckboxWidget from "./themes/material-design/CheckboxWidget";

WidgetRegistry.addToDefaultWidgetMap("boolean", MaterialDesignCheckboxWidget);
```

Redefine pure Checkbox widget (_themes/material-design/CheckboxWidget.js_) 

``` javascript
import React from "react";
import Checkbox from "material-ui/Checkbox";

function MDCheckboxWidget({
    id,
    value,
    disabled,
    label,
    onChange,
}) {
    return (
        <Checkbox
            id={id}
            checked={!!value}
            disabled={disabled}
            onCheck={(event) => onChange(event.target.checked)}
            label={label}
        />
    );
}

export default MDCheckboxWidget;
```

Because `material-ui`, we need to wrap with MuiThemeProvider (_app.js_)

``` javascript
<MuiThemeProvider>
    <Form schema={schema}
        uiSchema={uiSchema}
        formData={formData}/>
</MuiThemeProvider>}
```

That's all folks! :)

checkbox md

Could you imagine how it will be easy to create themes?
### Checklist


jmfrancois:
The main idea is to keep the lib as it is and let anyone who want to change all default widgets to do so without forking the lib.

We have tried different ways and this one is the best because it let the lib handle the behavior and just change the defaults;

It has a real benefit for us (we want material design in our forms).
Thanks @frassinier for this great job.

PS: hello @leplatrem as an old colleague from makina ;)

olzraiti:
Related also closely to #283, #310 and #354.

src/registry/Registry.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * The registry that will have the singleton
+ * - getRegistry() -> the registry which is a simple key/value POJO
+ * @type {Object}
+ */
+const Registry = {
```
n1k0: So if this becomes a singleton, how about using multiple forms using different registries in a same page?

frassinier: It's a widget registry, just a hook to redefine all pure components (TextWidget, CheckboxWidget, etc.). 
I can't imagine that you would start using `react-bootstrap` for one form and `material-ui` for another one :/

frassinier:
> I don't like the global registry singleton approach; could we imagine having a single one per form? eg. reusing shouldn't be an issue as we could pass a constructed one to multiple form instances

No, I think it's library related. "As a developer, I would set `material-ui` theme for all forms in my app" 

> This adds a lot of code, getter and setters, which hasn't been the general style for this project; I understand that's the style you prefer, but could we conceived trying to adapt it a little more to the existing one?

If this approach is kept, I will simplify the code.

n1k0:
> No, I think it's library related. "As a developer, I would set material-ui theme for all forms in my app"

I disagree; how about we let developers deciding about that? Instantiating a registry object instead of relying on some stateful global sounds like a much saner approach to me, while being as powerful (also, just think about testing: resetting the singleton state before each test, really?).

frassinier:
Ok, maybe it's only my need, so what's about adding a prop `theme` to Form which can edit the default widget mapping from _utils.js_?

n1k0:
That sounds like a very convenient API indeed :)

jmfrancois:
@n1k0 I think this is a different scope.

* be able to change the defaults widgets
* be able to change the default widgets per form

AFAIK you already can do a change for one form by using the schema field.

This is a really common need to set the look & feel of all forms in your app.
I would find this really boring for each form to add *theme* just to change the default.

The registry with getter and setter has nothing related with code style.
Doing import of a widget in a field to say this is the default widget for that field is hardcoded default. I don't know any other way to let integrator changing defaults except by a registry.

n1k0:
```js
const registry = new Registry({...));
render(<div><Form registry={registry} /><Form registry={registry} /></div>)
```

vs

```js
render(<div><Form /><Form /></div>)
```

I'm sorry, are we really arguing over this? Please excuse my bluntness, but I don't want global singletons, period.

frassinier:
You definitely right, even if I propose to set `theme` rather than `registry` prop.

Your third snippet is what we will do as high order function to fit to our needs and I understand it's not what all users want.

I plan to edit widget mapping from _utils.js_ and if a `theme` prop i provided then lib will merge mappings. Looks good to you?

jmfrancois:
> I'm sorry, are we really arguing over this?

No we just try to understand your review.
I had not seen the fact that a registry imply a kind of state, which is bad it's true.
Our goal was to change the default so having an API to do that.

So let's move from a global registry for default to simple props.

n1k0:
> I plan to edit widget mapping from utils.js and if a theme prop i provided then lib will merge mappings. Looks good to you?

Most definitely! Thanks

frassinier:
It's up to date @n1k0

n1k0:
Thanks, I'll give this a proper review tomorrow.

frassinier:
[Here](https://github.com/frassinier/rjsf-material-design) is an example of a possible theme.

```javascript
import Form from "react-jsonschema-form";

import customizedWidgets from "rjsf-material-design";

<Form schema={schema}
      uiSchema={uiSchema}
      formData={formData}
      theme={customDefaultWidgets}/>
```

src/components/fields/SchemaField.js
```diff
@@ -147,6 +148,10 @@ function SchemaField(props) {
     return <div/>;
   }
 
+  if (theme) {
+    mergeToDefaultWidgets(theme);
```
n1k0: That's a massive side effect. I'd rather at least have a new object retrieved from the call to this function:

```js
const themedWidgets = mergeToDefaultWidgets(theme);
```

src/utils.js
```diff
@@ -93,6 +113,36 @@ export function defaultFieldValue(formData, schema) {
...
+  };
+  return widget;
+}
+
+export function getDefaultWidget(
+  schema,
+  widgetType,
+) {
+  const {type} = schema;
+  if(!widgetType && Array.isArray(schema.enum)) {
```
n1k0: Nit: space after `if`

frassinier:
To move forward, that feature will be added with #371

PR #368, opened 2016-11-01
Add ui:order wildcard feature and improve error messages

maartenth:

It can be tedious to maintain a `ui:order` list for large schemas if you only want to change the order of two or three fields. (I know that, strictly, the order of object properties is undefined, but in practice it always yields a constant order.) This feature lets you insert a wildcard (`"*"`) for all fields that you don't reference explicitly in `ui:order`, like this:

```js
const uiSchema = {
  "ui:order": ["bar", "*"]
};
```

This also helps during development so you don't have to always update your `uiSchema` when you update your `schema`. And it is useful when you dynamically insert properties in a schema and you want those dynamic properties to appear in a specific position.

While implementing this feature, I was able to improve the error messages for invalid `ui:order` values, specifying the exact field that causes the error. This really helps during debugging.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ This allows you to programmatically trigger the browser's file selector which ca
 
 ### Object fields ordering
 
-The `uiSchema` object spec also allows you to define in which order a given object field properties should be rendered using the `ui:order` property:
+Since the order of object properties in Javascript and JSON is not guaranteed, the `uiSchema` object spec allows you to define the order in which properties are rendered using the `ui:order` property:
 
 ```jsx
 const schema = {
@@ -409,7 +409,7 @@ render((
 ), document.getElementById("app"));
 ```
 
-For fields for which the order is OK already or a fixed order is not important, you can insert a wildcard `"*"` item in your `ui:oder` definition. All unreferenced fields will be rendered at that point:
+If a guarenteed fixed order is only important for some fields, you can insert a wildcard `"*"` item in your `ui:order` definition. All fields that are not referenced explicitly anywhere in the list will be rendered at that point:
 
 ```js
 const uiSchema = {
--- a/README.md
+++ b/README.md
@@ -408,6 +408,15 @@ render((
         uiSchema={uiSchema} />
 ), document.getElementById("app"));
 ```
+
+For fields for which the order is OK already or a fixed order is not important, you can insert a wildcard `"*"` item in your `ui:oder` definition. All unreferenced fields will be rendered at that point:
+
+```js
+const uiSchema = {
+  "ui:order": ["bar", "*"]
+};
+```
+
 ### Array items ordering
 
 Array items are orderable by default, and react-jsonschema-form renders move up/down buttons alongside them. The `uiSchema` object spec allows you to disable ordering:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -408,6 +408,15 @@ render((
         uiSchema={uiSchema} />
 ), document.getElementById("app"));
 ```
+
+For fields for which the order is OK already or a fixed order is not important, you can insert a wildcard `"*"` item in your `ui:oder` definition. All unreferenced fields will be rendered at that point:
```
glasserc: `ui:oder` should be `ui:order`. Also, maybe we should add a note here saying that the other fields will be rendered in the default order, but I see that we don't actually say what the default order is..

You later add an example where `*` shows up in the middle of fields. Is it maybe a good idea to emphasize that "unreferenced" fields means "unreferenced throughout the list" and not just "unreferenced up to that point"?

src/utils.js
```diff
@@ -249,16 +249,32 @@ export function orderProperties(properties, order) {
...
+  const orderHash = arrayToHash(order);
+  const extraneous = order.find(prop => prop !== '*' && !propertyHash[prop]);
+  if (extraneous) {
+    throw new Error(`uiSchema order list contains extraneous property "${extraneous}"`);
+  }
+  const rest = properties.filter(prop => !orderHash[prop]);
+  const restIndex = order.indexOf('*');
+  if (restIndex === -1) {
+    if (rest.length) {
+      throw new Error(`uiSchema order list does not contain property "${rest[0]}"`);
```
glasserc: Maybe show all the missing properties?

README.md
n1k0: Btw we may also want to add a short note about js/JSON object keys order being not consistently guaranteed, hence why explicitly defining the complete order is important if you care about UI consistency.

src/utils.js
```diff
@@ -249,16 +249,32 @@ export function orderProperties(properties, order) {
   if (!Array.isArray(order)) {
     return properties;
   }
-  if (order.length !== properties.length) {
-    throw new Error(
-      "uiSchema order list length should match object properties length");
+
+  const arrayToHash = arr => arr.reduce((prev, curr) => {
+    prev[curr] = true;
+    return prev;
```
n1k0: Nit: `return {...prev, [curr]: true};`

maartenth:
Thank you for your comments @glasserc and @n1k0, I will update my pull request with your suggestions tomorrow.

README.md
maartenth: @glasserc I updated both the typo and the explanation
@n1k0 also added this

src/utils.js
maartenth: Changed. I wonder, though, if this has the same performance, since it copies the entire object every time now...

n1k0: I must admit, I have no idea... I suspect optimizations are made at the js engine level (the spread operator uses `Object.assign` under the hood), but you're never sure until you produce some benchmarks... *shudder*

n1k0: Maybe switch back to your initial implementation so we're possibly safer than with my suggestion.

maartenth: fixed

n1k0:
Thank you!

n1k0:

PR #369, opened 2016-11-01
asNumber: return undefined when value is empty string

sjhewitt:
It is currently impossible to clear the value of a number field that is not required because `asNumber` converts the empty string to `0`. This change adds a special case that returns `undefined`, so that the input can be cleared

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -503,7 +503,7 @@ export function dataURItoBlob(dataURI) {
   // Built the Uint8Array Blob parameter from the base64 string.
   const binary = atob(splitted[1]);
   const array = [];
-  for(let i = 0; i < binary.length; i++) {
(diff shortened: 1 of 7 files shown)


src/utils.js
```diff
@@ -223,6 +223,9 @@ export function mergeObjects(obj1, obj2, concatArrays = false) {
 }
 
 export function asNumber(value) {
+  if(value === "") {
```
n1k0: Please add a space after `if`. Would be nice to add an eslint rule for this if you have the time btw :)

sjhewitt:
@n1k0 - done, and I added the linter rule and fixed any errors

n1k0:
@sjhewitt Fantastic, thank you so much :+1:

n1k0:

PR #371, opened 2016-11-02
Allow overriding the default fields/widgets

olzraiti:

This pull request is conflicting with #363 and #366. There is some heated discussion about the technical details in the comments of those pull requests. I'm offering this pull request because I had the code changes lying around anyway (as I have mentioned in #310). My intention is not to step on anyones toes - I'm merely offering a solution or at least some insight to the issue.

The reason why I have sitted on this for so long time is because I was waiting for input on the issues I have raised (#354, #310).

This approach has the benefit that we aren't introducing any new techniques or new API designs to solve the issue. All fields and widgets are simply overridable through the `fields` and `widgets` props.

I didn't add any tests or update the documentation since this pull request is conflicting.

### Checklist


src/components/Form.js
```diff
@@ -118,19 +115,11 @@ export default class Form extends Component {
...
     // the "fields" registry one.
-    const _SchemaField = this.props.SchemaField || SchemaField;
-    const _TitleField = this.props.TitleField || TitleField;
-    const _DescriptionField = this.props.DescriptionField || DescriptionField;
-
-    const fields = Object.assign({
-      SchemaField: _SchemaField,
-      TitleField: _TitleField,
-      DescriptionField: _DescriptionField,
-    }, this.props.fields);
```
n1k0: This were kept for backward compatibility purpose, but I'm glad seeing them going away. The change should be documented though (in the next release changelog).

src/utils.js
```diff
@@ -80,10 +78,34 @@ export function getDefaultRegistry() {
...
       SchemaField: require("./components/fields/SchemaField").default,
-      TitleField,
-      DescriptionField,
+      ArrayField: require("./components/fields/ArrayField").default,
+      BooleanField: require("./components/fields/BooleanField").default,
+      ObjectField: require("./components/fields/ObjectField").default,
+      StringField: require("./components/fields/StringField").default,
+      NumberField: require("./components/fields/NumberField").default,
+      TitleField: require("./components/fields/TitleField").default,
+      DescriptionField: require("./components/fields/DescriptionField").default,
```
n1k0: For legibility, we could create some `load` helper, eg. 

```js
const load = comp => require(`./components/${comp}`).default;
const fields = {
  ArrayField: load("fields/ArrayField"),
  // ...
};
```

frassinier:
@olzraiti Could you provide sample of customization please?

olzraiti:
@frassinier for example if you would like to add a material theme, you would just override the specific widgets like so:

## Example 1: Themes ##
```
<Form
    widgets={{
        CheckboxWidget: CustomMaterialWidget // This causes the BooleanField to use CustomMaterialWidget as the widget for checkboxes instead of the default CheckboxWidget
    }}
/>
```

If you need different themes, you could easily wrap your form like so:

```

function ThemedForm(props) {
    const themes = {
        material: {
            widgets: {
                CheckboxWidget: CustomMaterialCheckbox
            }
        },
        someOtherTheme: {
            ...what ever you would like to override
        }
    }

    <Form
        {...themes[props.theme]} 
    />
}
```
Now you have a form component which you can render with different themes:

```
<ThemedForm theme="material" />
```

## Example 2: non orderable arrays as default ##

```
import ArrayField from 'react-jsonschema-form/lib/components/fields/ArrayField';

function NonOrderableArrayField(props) {
    return (
        <ArrayField 
                   {...props} 
                   uiSchema={
                       ...props.uiSchema, 
                       "ui:options": {...props.uiSchema["ui:options"], orderable: false} 
        />
    );  
}

<Form fields={{
  ArrayField: NonOrderableArrayField
}} />
```

frassinier:
Love it, I will close my PR and help you by providing `material-ui` theme if you want?

olzraiti:
I'm glad we came to an agreement so fast, thanks for being so flexible. :)

It's up to @n1k0 if this library should provide themes. I can think of two ways this library could go: 

1. Provide no themes at all (not even the current bootstrap themes). Let the users handle the themes themselves.
2. Provide pluggable themes (I'd like to see the bootstrap theme also as a plugin)

frassinier:
No, IMHO, I think we should let developers apply external themes but not change this lib and its bootstrap markup. 
Developers who already use this lib should will not have to deal with Bootstrap theme, that could be a breaking change.
And I'm pretty sure that nobody wants to launch the playground with no style at all.

olzraiti:
Oh, that's ok. I got you wrong - I though you wanted to add `material-ui` theme to the library.

frassinier:
No, in fact, I just proposed to make [this repo](https://github.com/frassinier/rjsf-material-design) to be compatible with that PR :)

frassinier:
> I didn't add any tests or update the documentation since this pull request is conflicting.

Do you need some help to move forward?

olzraiti:
I'm writing some tests atm. If you want to help, you could for example update the documentation. I can do it myself, if you don't have the time :)

frassinier:
Thanks a lot, I have some stuff to do but ping me if you need any help!

olzraiti:
I didn't add many new tests or extend the documentation, since we are mostly removing a special case instead of adding new features (SOME of the fields/widgets have been overridable before this pull request, but not all. Now all of them are overridable). Maybe the documentation for a custom `SchemaField` could be merged to the added documentation chapter? Though having a specific chapter for `SchemaField` makes sense, since it's a major building block.

frassinier:
Good job @olzraiti ! I will work on a first theme impl ASAP :)

playground/app.js
```diff
@@ -358,6 +358,7 @@ class App extends Component {
               formData={formData}
               onChange={this.onFormDataChange}
               fields={{geo: GeoPosition}}
+              widgets={{AltDateTimeWidget: () => <div id="#custom" >moi</div>}}
```
frassinier: Is this really usefull? :)

frassinier:
Hey @olzraiti, it works well! Here is an [example](https://frassinier.github.io/rjsf-material-design/). cc @n1k0

README.md
```diff
@@ -897,6 +898,32 @@ If you're curious how this could ever be useful, have a look at the [Kinto formb
 
 Props passed to a custom SchemaField are the same as [the ones passed to a custom field](#field-props).
 
+### Customizing any fields or widgets
```
n1k0: How about *Customizing default fields and widgets*?

README.md
```diff
@@ -897,6 +898,32 @@ If you're curious how this could ever be useful, have a look at the [Kinto formb
...
+const widgets = {
+  CheckboxWidget: CustomCheckbox
+};
+
+render((
+  <Form schema={schema}
+        uiSchema={uiSchema}
+        formData={formData}
+        widgets={widgets} />
+), document.getElementById("app"));
```
n1k0: It may look super obvious to seasonned react developers, but maybe we should augment this section with an example of a reusable customized form class, eg.

```jsx
const fields = {StringField: CustomString};
const widgets = {CheckboxWidget: CustomCheckbox};

function MyForm(props) {
  return <Form fields={customFields} widgets={customWidgets} {...props} />;
}

render((
  <MyForm schema={schema}
    uiSchema={uiSchema}
    formData={formData} />
```

src/components/fields/SchemaField.js
```diff
@@ -31,7 +26,7 @@ function getFieldComponent(schema, uiSchema, fields) {
   if (typeof field === "string" && field in fields) {
     return fields[field];
   }
-  return COMPONENT_TYPES[schema.type] || UnsupportedField;
+  return (COMPONENT_TYPES[schema.type] ? fields[COMPONENT_TYPES[schema.type]] : UnsupportedField) || UnsupportedField;
```
n1k0: How about:

```js
const componentName = COMPONENT_TYPES[schema.type];
return componentName in fields ? fields[componentName] : UnsupportedField;
```

README.md
frassinier: Typo

```javascript
const customFields = {StringField: CustomString};
const customWidgets = {CheckboxWidget: CustomCheckbox};

function MyForm(props) {
  return <Form fields={customFields} widgets={customWidgets} {...props} />;
}
```

olzraiti:
Closes #272, closes #310, closes #363.

That's a neat example @frassinier, thanks for showing the potential this pull request brings!

n1k0:
Thanks for this great collaborative effort!

n1k0:

AMongeEntytle:
Great job guys ! Thank you for all the effort !

herzaso:
I might have missed something but how do I change the default "number" input to an "updown" widget?
Obviously, I need the TextWidget for text inputs...

Also, why is the default widget for numbers a TextWidget and not something more numbers oriented?

Natim:
> why is the default widget for numbers a TextWidget and not something more numbers oriented?

I guess because the TextWidget is the best compromise when you have no idea what kind of number one is talking about. For instance up and down button for a timestamp doesn't make any sense.

Natim:
@herzaso The documentation for custom widget is here: https://github.com/mozilla-services/react-jsonschema-form#custom-widget-components

There apparently are some buildin widget for numbers: https://github.com/mozilla-services/react-jsonschema-form#for-number-and-integer-fields

```
const uiSchema = {
  quantity: {"ui:widget": "updown"}
};
```

herzaso:
@Natim , thanks for your comments.

IMO, a better compromise would be to use a field that relates to a number and not to a string since I don't think there would be many real-world situations in which a user would want to have a text input for numbers (I'm not sure that showing a timestamp as is, and not as a date qualifies as a real-world situation). Either way, if the user has the ability to change the default widget, he could do so.


Regarding my main question, I'm looking for a way to change the **default** widget since my form would be dynamic and I wouldn't want to loop over it. Since your answer refers to a specific field (`quantity`), I'm not sure that will yield the required result. In my scenario, all number fields should be rendered as a number input field.

In that respect, the first link you provided (the uiSchema function) looks interesting but I do have 2 concerns:
1. I'll need a way to return the default widget (for all non-numeric fields)
2. I prefer not to implement the entire input myself as it wouldn't be future-proof. I just want to change the input `type` to `number`

Natim:
I guess this pull-request allow you to do just that :) https://github.com/mozilla-services/react-jsonschema-form#customizing-the-default-fields-and-widgets

herzaso:
Actually, I'm not sure it does as it rely on the default widget to replace and in my situation it will be TextWidget (and I don't want to replace the TextWidget).
I need some mapping between a property type (e.g. "number" / "integer") and a widget (`UpDownWidget`).
Ideally, that would be a mapping object (e.g. `{"number": UpDownWidget}`) and not a function replacing the whole implementation...

annjawn:
Hey everyone, I am really struggling to find out how to override widgets globally. We have multiple json schema templates and customizing them individually is going to be painful. Just like MaterialUI we would like to use AntD components but I am not sure how to globally override Text inputs, checkboxes, radio buttons, select's etc. any pointer here would be appreciate.

epicfaace:
@annjawn you'd have to customize each component individually. Eventually if people are interested they could contribute an antd theme (like for material ui) which would take a lot of work to initially make, but would then be easy for others to use.

PR #373, opened 2016-11-03
Removable and addable options for array items

maartenth:

There are cases where you don't want an add button or a remove button for array items, even though the schema allows it. Just like with the recent `orderable` option. The array items in my current case come from another resource and I fill the array in the background on initial form load; the user should not be able to reorder, add or remove the items via the form. I think this is a common use case.

I refactored the code a little bit to align all variable names (`orderable`, `addable` and `removable` for `uiSchema` settings, and `canMoveUp`, `canMoveDown`, `canDelete` for application logic).

options

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,10 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
            - [Multiple files](#multiple-files)
            - [File widget input ref](#file-widget-input-ref)
      - [Object fields ordering](#object-fields-ordering)
-     - [Array items ordering](#array-items-ordering)
+     - [Array item options](#array-items-options)
+        - [Orderable option](#orderable-option)
+        - [Addable option](#addable-option)
+        - [Removable option](#deletable-option)
      - [Custom CSS class names](#custom-css-class-names)
      - [Custom labels for enum fields](#custom-labels-for-enum-fields)
      - [Multiple choices list](#multiple-choices-list)
@@ -419,7 +422,9 @@ const uiSchema = {
 };
 ```
 
-### Array items ordering
+### Array item options
+
+#### `orderable` option
 
 Array items are orderable by default, and react-jsonschema-form renders move up/down buttons alongside them. The `uiSchema` object spec allows you to disable ordering:
 
@@ -434,6 +439,32 @@ const schema = {
 const uiSchema = {
   "ui:options":  {
     orderable: false
+  }
+};
+```
+
+#### `addable` option
+
+If either `items` or `additionalItems` contains a schema object, an add button for new items is shown by default. You can turn this off with the `addable` option:
+
+```jsx
+const schema = {
+const uiSchema = {
+  "ui:options":  {
+    addable: false
+  }
+};
+```
+
+#### `removable` option
+
+A remove button is shown by default for an item if `items` contains a schema object, or the item is an `additionalItems` instance. You can turn this off with the `removable` option:
+
+```jsx
+const schema = {
+const uiSchema = {
+  "ui:options":  {
+    removable: false
   }
 };
 ```
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
      - [Array item options](#array-item-options)
         - [Orderable option](#orderable-option)
         - [Addable option](#addable-option)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -427,6 +433,30 @@ const uiSchema = {
 };
 ```
 
+#### `addable` option
+
+if either `items` or `additionalItems` contains a schema object, an add button for new items is shown by default. You can turn this off with the `addable` option in `uiSchema`:
```
n1k0: I feel like this is trying to fix a misused json schema; If you want a fixed size array, your schema should reflect that intent, and the lib should render the form appropriately. Have a look at the `fixedItemsList` sample in the playground.

I understand that this would add more flexibility, but on the other hand I feel like providing features to help people fixing their schemas is an endless rabbit hole we're really wanting to start exploring... Thoughts? Did you know about support for fixed size arrays?

maartenth: Thank you for your reply. I knew about fixed arrays, I even added a fixed sized array example with 2 unremovable additionalItems in the playground. :)

The use case I have, is that the items _are_ addable, reorderable and removable, but not via _this_ form. Another user may add items in another form, but this user may not add items in this form. So the items in the array are really part of the data, not part of the schema. If I'm not mistaken, a fixed array in the schema would not be a right representation of the data. Then every instance of this particular data type should have their own schema, which would undermine the purpose of json schema.

In angular-schema-form, which appears to be maintained by the same organisation as JSON Schema, the [api for this](https://github.com/json-schema-form/angular-schema-form/blob/development/docs/index.md#array) is as follows (a `uiSchema` is called a `form` in angular-schema-form):

> To suppress add and remove buttons set add to null and remove to null.

```js
function FormCtrl($scope) {
  $scope.form = [
    {
      key: "subforms",
      add: null,
      remove: null,
      style: {
        add: "btn-success"
      },
      items: [
        "subforms[].nick",
        "subforms[].name",
        "subforms[].emails",
      ],
    }
  ];
}
```

n1k0: This makes a lot of sense, I get the point. Also I realize that people may want to alter the UI according to some user ACLs, and your patch allows dealing with this in a very convenient fashion.

Scratch my concerns, this is valid.

src/components/fields/ArrayField.js
```diff
@@ -203,8 +207,9 @@ class ArrayField extends Component {
             });
           })
         }</div>
-        <AddButton
-          onClick={this.onAddClick} disabled={disabled || readonly}/>
+        {addable ? <AddButton
+                   onClick={this.onAddClick}
+                   disabled={disabled || readonly}/> : null}
```
n1k0: Nit: indent props with 2 spaces

src/components/fields/ArrayField.js
```diff
@@ -271,6 +276,11 @@ class ArrayField extends Component {
       retrieveSchema(item, definitions));
     const additionalSchema = allowAdditionalItems(schema) ?
       retrieveSchema(schema.additionalItems, definitions) : null;
+    const {addable} = {
+      addable: true,
+      ...uiSchema["ui:options"]
+    };
+    const hasAdd = addable && additionalSchema;
```
n1k0: how about `canAdd` for consistency?

src/components/fields/ArrayField.js
```diff
@@ -337,19 +347,22 @@ class ArrayField extends Component {
...
+      orderable: true,
+      removable: true,
+      ...uiSchema["ui:options"]
+    };
+    const has = {
+      moveUp: orderable && canMoveUp,
+      moveDown: orderable && canMoveDown,
+      remove: removable && canRemove
+    };
+    has.toolbar = Object.keys(has).map(key => has[key]).includes(true);
```
```js
Object.keys(has).some(key => has[key]);
```

maartenth:
Updated those 3 things and rebased. :)

n1k0:
Woops, I've landed #375 meawhile which makes your patch conflicting again. I can take over the next rebase if you're bored ;)

maartenth:
Re-rebased. :)

n1k0:
Perfect, thank you so much!

pabloen:
Thanks, this is a great addition!
I think the documentation is visible in the README but the feature is not released yet. Am I wrong?

n1k0:
You're not, we'll release a new version very soon now.

n1k0:

PR #376, opened 2016-11-04
readme: Add a note about formData

revolunet:

Hi, not sure this is a feature or a bug, so i try to document the current behaviour.

In the case fo a single-field form, looks like we must pass `formData` as the field value instead of an object containing that field value.

I find this counter-intuitive at first sight, so i could dig deeper and add some tests if you want to change this in the future and make formData shape consistent whatever the destination form. (always as an object).

Thanks for this great tool 👍 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ render((
 ), document.getElementById("app"));
 ```
 
+NOTE: If your form have a single field, pass a single value to `formData`. ex: `formData='Charlie'`
+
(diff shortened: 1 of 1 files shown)


n1k0:
Well it makes sense as soon as you know what are the available matching jsonschema primitives and how form data are matched against these under the hood :)

You can't map a single string form field to a js object, as it would match a jsonschema `object` value.

I'm happy to land your patch, it helps.

revolunet:
Thanks for the clarification ;)

PR #378, opened 2016-11-07
ui:options everywhere (fix #370)

### LINKED ISSUE #370: There should be a single way to provide widget options in uiSchema
We currently have two distinct ways to provide custom widget options in uiSchema:

- Through the object format for the `ui:widget` directive,
- Through the `ui:options` one.

I'm in favor of using the latter, and deprecating support for the former. Thoughts?

maartenth:

See #370.

Issues:

- (Error) One test does not work any more: "referenced descriptor". This is inherent to the change as proposed by me in #370, but maybe we can still make it working? It is undocumented behaviour though, and I doubt if the functionality is really useful. I think a referenced descriptor with `{component, options}` as widget definition is always easily replaceable with a custom widget with some `defaultProps`.

- (Question) I don't like the fact that `ui:options` from `uiSchema` and processed `enum` and `enumNames` data from `schema` are merged into one property for widgets in every field that accepts `enum`. (E.g. in `StringField`: `options={{...options, enumOptions}}`.) These 2 things don't 'naturally' belong together in one property. Could we split out the enum information to a different property called `choices`? (I took 'choices' from the corresponding Django widget but another name would also do.)

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -179,7 +179,7 @@ class ArrayField extends Component {
     const {definitions, fields} = this.props.registry;
     const {TitleField, DescriptionField} = fields;
     const itemsSchema = retrieveSchema(schema.items, definitions);
-    const {addable} = {addable: true, ...getUiOptions(uiSchema)};
+    const {addable=true} = getUiOptions(uiSchema);
 
     return (
       <fieldset
@@ -225,7 +225,7 @@ class ArrayField extends Component {
     const {widgets, definitions} = this.props.registry;
     const itemsSchema = retrieveSchema(schema.items, definitions);
     const enumOptions = optionsList(itemsSchema);
-    const {widget, ...options} = {widget: "select", ...getUiOptions(uiSchema), enumOptions};
+    const {widget="select", ...options} = {...getUiOptions(uiSchema), enumOptions};
     const Widget = getWidget(schema, widget, widgets);
     return (
       <Widget
@@ -246,7 +246,7 @@ class ArrayField extends Component {
     const title = schema.title || name;
     const {items} = this.state;
     const {widgets} = this.props.registry;
-    const {widget, ...options} = {widget: "files", ...getUiOptions(uiSchema)};
+    const {widget="files", ...options} = getUiOptions(uiSchema);
     const Widget = getWidget(schema, widget, widgets);
     return (
       <Widget
@@ -283,7 +283,7 @@ class ArrayField extends Component {
       retrieveSchema(item, definitions));
     const additionalSchema = allowAdditionalItems(schema) ?
       retrieveSchema(schema.additionalItems, definitions) : null;
-    const {addable} = {addable: true, ...getUiOptions(uiSchema)};
+    const {addable=true} = getUiOptions(uiSchema);
     const canAdd = addable && additionalSchema;
 
     if (!items || items.length < itemSchemas.length) {
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -1,7 +1,7 @@
 import React, {Component, PropTypes} from "react";
 
 import {
-  getAlternativeWidget,
+  getWidget,
   getDefaultFormState,
   getUiOptions,
   isMultiSelect,
@@ -226,7 +226,7 @@ class ArrayField extends Component {
     const itemsSchema = retrieveSchema(schema.items, definitions);
     const enumOptions = optionsList(itemsSchema);
     const {widget, ...options} = {widget: "select", ...getUiOptions(uiSchema), enumOptions};
-    const Widget = getAlternativeWidget(schema, widget, widgets);
+    const Widget = getWidget(schema, widget, widgets);
     return (
       <Widget
         id={idSchema && idSchema.$id}
@@ -242,12 +242,15 @@ class ArrayField extends Component {
   }
 
   renderFiles() {
-    const {schema, idSchema, name, disabled, readonly, autofocus} = this.props;
+    const {schema, uiSchema, idSchema, name, disabled, readonly, autofocus} = this.props;
     const title = schema.title || name;
     const {items} = this.state;
-    const {FileWidget} = this.props.registry.widgets;
+    const {widgets} = this.props.registry;
+    const {widget, ...options} = {widget: "files", ...getUiOptions(uiSchema)};
+    const Widget = getWidget(schema, widget, widgets);
     return (
-      <FileWidget
+      <Widget
+        options={options}
         id={idSchema && idSchema.$id}
         multiple
         onChange={this.onSelectChange}
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -3,6 +3,7 @@ import React, {Component, PropTypes} from "react";
 import {
   getAlternativeWidget,
   getDefaultFormState,
+  getUiOptions,
   isMultiSelect,
   isFilesArray,
   isFixedItems,
@@ -178,10 +179,7 @@ class ArrayField extends Component {
     const {definitions, fields} = this.props.registry;
     const {TitleField, DescriptionField} = fields;
     const itemsSchema = retrieveSchema(schema.items, definitions);
-    const {addable} = {
-      addable: true,
-      ...uiSchema["ui:options"]
-    };
+    const {addable} = {addable: true, ...getUiOptions(uiSchema)};
 
     return (
       <fieldset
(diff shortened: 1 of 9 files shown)


src/utils.js
```diff
@@ -99,34 +101,21 @@ export function defaultFieldValue(formData, schema) {
...
-export function getAlternativeWidget(
-  schema,
-  widget,
-  registeredWidgets={},
-  widgetOptions={}
-) {
+export function getAlternativeWidget(schema, widget, registeredWidgets={}) {
   const {type, format} = schema;
 
-  function setDefaultOptions(widget) {
```
maartenth: Extending defaultProps was persistent across all instances of a widget, which was a bug. I fixed that by creating the wrapper element. Merging top level defaultProps is not needed there, since React by default merges those. Only options should be merged here. I documented that behaviour, because merging variables is not the React default behaviour.

src/utils.js
```diff
@@ -99,34 +101,21 @@ export function defaultFieldValue(formData, schema) {
...
-        ...defaultProps.options,
-        ...widgetOptions
-      }
-    };
-    return widget;
+  function mergeOptions(Widget) {
+    // cache return value as property of widget for proper react reconciliation
+    if (!widget.mergedOptions) {
+      const defaultOptions = Widget.defaultProps && Widget.defaultProps.options || {};
+      Widget.mergedOptions = ({options={}, ...props}) =>
```
n1k0: I'd rather name this property `MergedWidget` or equivalent

src/components/fields/BooleanField.js
```diff
@@ -3,26 +3,11 @@ import React, {PropTypes} from "react";
...
-  // Note: uiWidget can be undefined, a string or an object; we only deal with
-  // the inline option when we're provided a definition object.
-  return {
-    inline: isObject(uiWidget) &&
-            isObject(uiWidget.options) &&
-            uiWidget.options.inline,
-    enumOptions: optionsList(Object.assign({
-      enumNames: ["true", "false"],
-      enum: [true, false]
-    }, {enumNames: schema.enumNames}))
```
n1k0: Yeah I agree, yes/no is better.

src/utils.js
```diff
@@ -208,6 +197,31 @@ export function getDefaultFormState(_schema, formData, definitions={}) {
   return formData || defaults;
 }
 
+export function getUiOptions(uiSchema) {
+  // get all passed options from ui:widget, ui:options, and ui:<optionName>
+  return Object.keys(uiSchema).filter(key => /^ui:/.test(key)).reduce((options, key) => {
```
n1k0: Nit: `key.startsWith("ui:")` is probably more performant

n1k0:
Also,

- (Error) One test does not work any more: "referenced descriptor". T(...) I doubt if the functionality is really useful

Yes, let's drop support for it. We'll have to document this breaking change in the next release changelog though.

- (Question) I don't like the fact that `ui:options` from `uiSchema` and processed `enum` and `enumNames` data from `schema` are merged into one property for widgets in every field that accepts `enum`. (E.g. in `StringField`: `options={{...options, enumOptions}}`.) These 2 things don't 'naturally' belong together in one property. Could we split out the enum information to a different property called `choices`? (I took 'choices' from the corresponding Django widget but another name would also do.)

I didn't understand this point; could you please elaborate a little more? What would be the new API?

src/utils.js
maartenth: MDN says that `String.prototype.startsWith()` is IE11- incomplient. Or is that polyfilled?

src/utils.js
```diff
@@ -208,6 +197,31 @@ export function getDefaultFormState(_schema, formData, definitions={}) {
...
+    } else if (key === "ui:options" && isObject(value)) {
+      setOptions(options, value);
+    } else {
+      options[key.substring(3)] = value;
+    }
+
+    return options;
+  }, {});
+
+  function setOptions(dest, src) {
```
maartenth: Check

src/utils.js
n1k0: We don't necessarily care about ie but I can understand that could become a concern. `key.indexOf("ui:") === 0` should work exactly the same and be more efficient than using a regex.

src/components/fields/ArrayField.js
```diff
@@ -226,17 +224,15 @@ class ArrayField extends Component {
     const {items} = this.state;
     const {widgets, definitions} = this.props.registry;
     const itemsSchema = retrieveSchema(schema.items, definitions);
-
-    const Widget = getAlternativeWidget(schema, uiSchema["ui:widget"] || "select", widgets);
+    const enumOptions = optionsList(itemsSchema);
+    const {widget, ...options} = {widget: "select", ...getUiOptions(uiSchema), enumOptions};
```
```js
const {widget="select", ...options} = {...getUiOptions(uiSchema), enumOptions};
```

src/components/fields/ArrayField.js
```diff
@@ -284,10 +283,7 @@ class ArrayField extends Component {
       retrieveSchema(item, definitions));
     const additionalSchema = allowAdditionalItems(schema) ?
       retrieveSchema(schema.additionalItems, definitions) : null;
-    const {addable} = {
-      addable: true,
-      ...uiSchema["ui:options"]
-    };
+    const {addable} = {addable: true, ...getUiOptions(uiSchema)};
```
n1k0: This should just work:

```js
const {addable=true} = getUiOptions(uiSchema);
```

src/components/fields/BooleanField.js
```diff
@@ -39,26 +24,25 @@ function BooleanField(props) {
...
-    formContext,
-    autofocus,
-  };
-  if (widget) {
-    const Widget = getAlternativeWidget(schema, widget, widgets);
-    return <Widget options={buildOptions(schema, uiSchema["ui:widget"])} {...commonProps}/>;
-  }
-  const {CheckboxWidget} = registry.widgets;
-  return <CheckboxWidget {...commonProps}/>;
+  const {widget, ...options} = {widget: "checkbox", ...getUiOptions(uiSchema)};
```
n1k0:
```js
const {widget="checkbox", ...options} = getUiOptions(uiSchema);
```

src/components/fields/StringField.js
```diff
@@ -15,46 +16,37 @@ function StringField(props) {
...
-    readonly,
-    formContext,
-    autofocus,
-    registry,
+  const enumOptions = Array.isArray(schema.enum) && optionsList(schema);
+  const {widget, placeholder, ...options} = {
+    widget: format || (enumOptions ? "select" : "text"),
+    placeholder: "",
+    ...getUiOptions(uiSchema)
   };
```
n1k0:
```js
const defaultWidget = format || (enumOptions ? "select" : "text");
const {widget=defaultWidget, placeholder="", ...options} = getUiOptions(uiSchema);
```

src/utils.js
```diff
@@ -208,6 +194,29 @@ export function getDefaultFormState(_schema, formData, definitions={}) {
   return formData || defaults;
 }
 
+export function getUiOptions(uiSchema) {
+  let options = {};
+
+  // get all passed options from ui:widget, ui:options, and ui:<optionName>
+  Object.keys(uiSchema).filter(key => /^ui:/.test(key)).forEach(key => {
```
n1k0: I think you can use `reduce` and avoid using this `options` variable

src/components/fields/ArrayField.js
maartenth: Yes! I'll change it everywhere.

src/components/fields/BooleanField.js
maartenth: done

maartenth:
@n1k0 as you can see, I started refactoring StringField and BooleanField a little bit to get rid of the if statement jungle. Everything (also the default widget) is handled via `getAlternativeWidget` now, so I just renamed it to `getWidget`. If you (or others) think that this is too rigorous, I can also submit those changes as a separate PR.

n1k0:
> I started refactoring StringField and BooleanField a little bit to get rid of the if statement jungle. Everything (also the default widget) is handled via `getAlternativeWidget` now, so I just renamed it to `getWidget`

No it's okay it made total sense when I reviewed them, great work.

src/utils.js
maartenth: Ah yes I missed that... Updated it to reduce again, Now it's shorter then ever.

maartenth:
> I didn't understand this point; could you please elaborate a little more? What would be the new API?

I'll come back on that later on in a new PR or issue.

src/components/fields/StringField.js
maartenth: Check

n1k0:
I think we're good, I'm landing this. Thank you :+1:

PR #379, opened 2016-11-08
Pass uiSchema to the field template 

olzraiti:

I need to do some heavy customization on the rendering of the fields. Passing the `uiSchema` object to the field template allows the users to extend the functionality of this library.

There was discussion about if `uiSchema` should be passed to the template field in #304. There were many good reasons to not pass the `uiSchema`, but it was also implied that it could be passed in the future if given reason. The template field is a functionality for flexible UI. `uiSchema` is a description for the UI, and therefore I find it reasonable to be able to allows users to extend their own functionality to the UI rendering. Use cases would be for example more fine grained control of rendering different types of titles/descriptions, help block, tooltips which could be described in `uiSchema`.

This allows users to implement dirty hacks and abuse, like @n1k0 suggested in #304, but on the other hand this might make the hackering more clean, when hackering must be done to implement a feature.

### Checklist


n1k0:
Yeah, and I'd even go as far as passing the schema as well. The more you can. Invalid usage of these will be easy to spot I think.

Could you please alter the patch so we pass `schema` to the template component?

olzraiti:
Agree! I almost added schema to the fields being passed in the initial commit, but decided to go with baby steps.

n1k0:
Thank you so much for your reactivity, really appreciated :+1:

n1k0:

PR #381, opened 2016-11-08
Don't pass consumed error to child schema

olzraiti:

This is a bug fix.

I ended with an error being rendered twice. This happens when you have an `errorSchema` containing an error for the whole object, not a property of the object. `SchemaField` consumes `errorSchema.__errors` by creating an `ErrorList` from it and passing it to the `FieldTemplate`. Since `errorSchema.__errors` is consumed already this way, it shouldn't be passed to the child schema component.

### Checklist


src/components/fields/SchemaField.js
```diff
@@ -163,6 +163,7 @@ function SchemaField(props) {
       disabled={disabled}
       readonly={readonly}
       autofocus={autofocus}
+      errorSchema={{...errorSchema, __errors: undefined}}
```
n1k0: How about:

```jsx
const {...fieldErrorSchema, __errors) = errorSchema;
// ...
const field = (
     <FieldComponent {...props}
       schema={schema}
        disabled={disabled}
        readonly={readonly}
        autofocus={autofocus}
        errorSchema={{...fieldErrorSchema, __errors: undefined}}
```

Note that `__errors` is used later in the same function body.

n1k0:
Thank you! :+1:

PR #387, opened 2016-11-10
Fix #385: Avoid dynamic requires.

### LINKED ISSUE #385: 0.41.0 introduced some issues in Meteor
https://github.com/meteor/meteor/issues/8035

Dynamic require calls like this one (where the module identifier string is not a string literal) are not supported in client-bundled code, because it's hard/impossible to know at bundling time exactly what module is being imported.

Webpack works around this by including all modules matching a certain configuration option, but then you're potentially shipping unused modules to the client.

You can force certain specific modules to be included in the bundle by putting

require("react-jsonschema-form/lib/components/fields/SchemaField.js")
explicitly in your own code. Repeat for any other modules that might be imported by dynamic require calls at runtime.

The code in question appears to have been added two days ago, so I think this error was introduced by the release of react-jsonschema-form@0.40.0, not by the recent Meteor release.

n1k0:
Refs #385 

@dulguun0225 could you please confirm this would solve your issue?

@olzraiti I'd love a review from you on this.

### DIFF
--- a/src/components/fields/index.js
+++ b/src/components/fields/index.js
@@ -1,0 +1,22 @@
+import ArrayField from "./ArrayField";
+import BooleanField from "./BooleanField";
+import DescriptionField from "./DescriptionField";
+import NumberField from "./NumberField";
+import ObjectField from "./ObjectField";
+import SchemaField from "./SchemaField";
+import StringField from "./StringField";
+import TitleField from "./TitleField";
+import UnsupportedField from "./UnsupportedField";
+
+
+export default {
+  ArrayField,
+  BooleanField,
+  DescriptionField,
+  NumberField,
+  ObjectField,
+  SchemaField,
+  StringField,
+  TitleField,
+  UnsupportedField,
+};
--- a/src/components/widgets/index.js
+++ b/src/components/widgets/index.js
@@ -1,0 +1,40 @@
+import AltDateWidget from "./AltDateWidget";
+import AltDateTimeWidget from "./AltDateTimeWidget";
+import CheckboxWidget from "./CheckboxWidget";
+import CheckboxesWidget from "./CheckboxesWidget";
+import ColorWidget from "./ColorWidget";
+import DateWidget from "./DateWidget";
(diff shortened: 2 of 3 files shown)


olzraiti:
Looks all good to me!

n1k0:
In https://github.com/mozilla-services/react-jsonschema-form/pull/387/commits/d2b379261122e5dd09bbc2704fe5211dc33f6057 I avoid requiring on every call, which might save us a few nanoseconds.

Now waiting for @dulguun0225's feedback

dulguun0225:
I'm not sure. https://github.com/benjamn verified that it was dynamic requires caused the problem.

dulguun0225:
Yes, it would.

n1k0:

DanialK:
Hey guys,
I have just tried v0.41.2 and noticed that the object `SchemaField.defaultProps.registry`, is undefined. Basically as you can see in the logs, the first couple of times the getDefaultRegistry gets called it's returning undefined which is being set to the `SchemaField.defaultProps.registry` when doing 
```

SchemaField.defaultProps = {
  registry: getDefaultRegistry()
};
```

image

image


Just to give you some context to my issue, i'm using a customised version of [Kinto/formbuilder](https://github.com/Kinto/formbuilder) and [here](https://github.com/Kinto/formbuilder/blob/master/formbuilder/components/builder/Form.js#L13) is where the problem occurred.

I currently have no idea why this is happening but would appreciate if you guys had any idea!
Cheers

DanialK:
Hi guys,

The issue above is not fixed yet and is due to cyclical imports because fields and widgets in defaultRegistry all call the getDefaultRegistry

```
SchemaField.defaultProps = {
  uiSchema: {},
  errorSchema: {},
  idSchema: {},
  registry: getDefaultRegistry(),  <-------
  disabled: false,
  readonly: false,
  autofocus: false,
};

```

but then the defaultRegistry is not even loaded at that point so it returns undefined.
defaultRegistry codes should be moved to a seperate file that can be imported without causing this cyclic dependency issue.

Thanks

PR #392, opened 2016-11-11
Tab stop for Add button +

eeinowski:

This makes it possible to add an array item while tabbing through a form.

https://github.com/mozilla-services/react-jsonschema-form/issues/365


n1k0:
Yes of course!

n1k0:

olzraiti:
How about we make all array field buttons tabbable? Delete and order buttons have specifically `tabIndex` set to `-1`. I don't see why a user shouldn't be able to delete/order an item while tabbing if he/she can add an item while tabbing.

PR #397, opened 2016-11-16
fix: deal with edage case in errorPropertyToPath

Chun-Yang:

When there is '.' in side a key of an object, the errorPropertyToPath function failed to parse the string into a path array.

### Checklist

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -3,22 +3,22 @@ import {validate as jsonValidate} from "jsonschema";
 import {isObject, mergeObjects} from "./utils";
 
 
-const RE_ERROR_ARRAY_PATH = /\[\d+]/g;
+const reEscapeChar = /\\(\\)?/g;
+const reLeadingDot = /^\./;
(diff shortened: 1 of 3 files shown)


src/validate.js
```diff
@@ -3,22 +3,22 @@ import {validate as jsonValidate} from "jsonschema";
...
-    if (match) {
-      const nodeName = node.slice(0, node.indexOf("["));
-      const indices = match.map(str => parseInt(str.slice(1, -1), 10));
-      path = path.concat(nodeName, indices);
-    } else {
-      path.push(node);
-    }
-    return path;
-  }, []);
+  // copied from lodash https://github.com/lodash/lodash/blob/4.17.1/lodash.js#L6744
```
n1k0: Seems like we could use https://lodash.com/docs/4.17.2#toPath directly. I think lodash allows importing functions atomically, so this shouldn't bloat the build too much.

n1k0:
Thanks for contributing :+1:

n1k0:

PR #398, opened 2016-11-16
fix: clear error when submit is successful

Chun-Yang:

I have a form that after submit will still be on the screen. The error does not disappear after successful submit. Here is how to repeat it:

1. make sure liveValidate if off
2. make an error in the form
3. submit
4. fix the error in the form
5. submit

The form will submit, but the error message does not disappear.

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -106,10 +106,11 @@ export default class Form extends Component {
       }
     }
 
-    if (this.props.onSubmit) {
-      this.props.onSubmit(this.state);
-    }
-    this.setState({status: "initial", errors: [], errorSchema: {}});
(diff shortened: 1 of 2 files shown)


n1k0:
I can't reproduce the error you on the playground... Could you be more specific in what are the required conditions to reproduce?

crumblix:
Likewise I also couldn't reproduce this in the playground.  Perhaps see if you can reproduce it there?  If you can it will help alot in fixing it, and if not maybe you'll see something different between it and your code?

Chun-Yang:
Sorry to respond so late. I tried on the playground and cannot reproduce it either. It seems to related to my own uniq use case.
I will close this for now. If I can repeat it, I will set up some example you can play with and resubmit the PR.
Thanks!

PR #403, opened 2016-11-21
Fix CheckboxesWidget and RadioWidget style

maartenth:

Unfortunately I currently don't have time for the 'bigger' issues, but this is a simple style fix I think will make many forms look better.

Some context:
The label in a `CheckboxWidget` is bold, and rightly so, because the label of a `CheckboxWidget` is the field title and all field titles are bold. The labels in a `RadioWidget` are not bold, unless the option `inline: true` is set.

Issue 1:
The labels in the `CheckboxesWidget` are bold too. I think this is done for consistency with `CheckboxWidget`, but I think consistency with `RadioWidget` would be better, because the labels of a `CheckboxesWidget` are not titles but `enumNames`. Now it could lead to an enormous wall of bold text for large `enum`s, and the real field title does not stand out any more. I changed this in this pull request.

Issue 2:
The labels in `RadioWidget` and `CheckboxesWidget` were bold when `inline: true` was set. This was due to a styling issue with Bootstrap, which needs another DOM structure for inline widgets (class `radio-inline` or `checkbox-inline` must be set on `<label>` itself instead of on a wrapper `<div>`). I fixed this too.

Issue 3:
While I was working on it, I saw that in `CheckboxesWidget` the `disabled` class was not set for disabled inputs. I fixed this like it was done for `CheckboxWidget` and `RadioWidget`.

### Checklist

### DIFF
--- a/src/components/widgets/CheckboxesWidget.js
+++ b/src/components/widgets/CheckboxesWidget.js
@@ -20,23 +20,33 @@ function CheckboxesWidget(props) {
     <div className="checkboxes" id={id}>{
       enumOptions.map((option, index) => {
         const checked = value.indexOf(option.value) !== -1;
-        return (
-          <div key={index} className={`checkbox${inline ? "-inline" : ""}`}>
+        const disabledCls = disabled ? "disabled" : "";
+        const checkbox = (
+          <span>
+            <input type="checkbox"
+              id={`${id}_${index}`}
+              checked={checked}
+              disabled={disabled}
+              autoFocus={autofocus && index === 0}
+              onChange={(event) => {
+                const all = enumOptions.map(({value}) => value);
+                if (event.target.checked) {
+                  onChange(selectValue(option.value, value, all));
+                } else {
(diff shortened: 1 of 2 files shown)


n1k0:
Looks good to me. Thank you :)

n1k0:

wintercounter:
This is great, but you removed `<strong>` from the markup which I used to style my checkboxes: `input:checked + strong`
I was happy that at least checkboxes had a wrapper for text. Please wrap the text into something so it'll be possible to use the `:checked` selector for both radio and checkbox. Just an empty `span` at least.

maartenth:
@wintercounter I'm not the one who decides but I think a `<span>` would not hurt anyone, you could do a PR to fix it for both CheckboxWidget, CheckboxesWidget and RadioWidget.

n1k0:
I agree!

Le 6 déc. 2016 12:50 PM, "Maarten ter Horst" <notifications@github.com> a
écrit :

PR #414, opened 2016-11-23
Missing prototype in CheckboxesWidget.js

crumblix:

Missing prototype in CheckboxesWidget.js

Referencing issue:
https://github.com/mozilla-services/react-jsonschema-form/issues/413

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -220,7 +220,7 @@ class ArrayField extends Component {
   }
 
   renderMultiSelect() {
-    const {schema, idSchema, uiSchema, disabled, readonly, autofocus} = this.props;
+    const {schema, idSchema, uiSchema, disabled, required, readonly, autofocus} = this.props;
     const {items} = this.state;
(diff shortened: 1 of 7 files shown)


crumblix:
OK ... that didn't work the way I wanted .... I only want one of those patches in this PR :(

crumblix:
Also ... that failed auto-test was a syntax error fixed in a later commit.  Obviously this system requires each individual commit to be good, rather than the set?

Anyway really I just wanted 	c7a7b4a in this PR.  For the life of my though I can't find how to select/deselect the commits I want in my PR ... I'll keep looking and reading.

Sorry for the confusion, but I will get better at this!

n1k0:
Hey now worries, things always have to be learned at some point ;)

Try to learn about rebasing, this is exactly what you need here; the idea would be to force push your branch rebased with your single commit here. There are a lot of good tutos out there about this!

crumblix:
Thanks Nicholas ... will do! :)

n1k0:
Fixed by #416

PR #417, opened 2016-11-23
Add support for anyOf

strahius:

This change adds support for [anyOf](https://spacetelescope.github.io/understanding-json-schema/reference/combining.html#anyof). The logic is built in the `ArrayField` component and handles all the various supported types (number, string, array etc).

Areas that need improvement:
- The layout is maybe not the prettiest.
- UISchema is not supported for anyOf. The challenge here, as far as I understand it is that UISchema needs specific types defined in the JSONSchema, whereas with anyOf, the type is determined on the fly.
- Array with anyOf, cannot also have fixed items at the moment. 
- anyOf item of type object is not supported. null is also not supported, but I saw there is a pull request (#329) for that.

### Checklist

### DIFF
--- a/playground/samples/index.js
+++ b/playground/samples/index.js
@@ -12,6 +12,7 @@ import date from "./date";
 import validation from "./validation";
 import files from "./files";
 import single from "./single";
+import customArray from "./customArray";
 import anyOf from "./anyOf";
 
 export const samples = {
@@ -29,5 +30,6 @@ export const samples = {
   Validation: validation,
   Files: files,
   Single: single,
+  "Custom Array": customArray,
   "Any of": anyOf
 };
--- a/playground/samples/index.js
+++ b/playground/samples/index.js
@@ -12,6 +12,7 @@ import date from "./date";
 import validation from "./validation";
 import files from "./files";
 import single from "./single";
+import anyOf from "./anyOf";
 
 export const samples = {
   Simple: simple,
@@ -27,5 +28,6 @@ export const samples = {
   "Date & time": date,
   Validation: validation,
   Files: files,
-  Single: single
+  Single: single,
+  AnyOf: anyOf
 };
--- a/playground/samples/index.js
+++ b/playground/samples/index.js
@@ -29,5 +29,5 @@ export const samples = {
   Validation: validation,
   Files: files,
   Single: single,
-  AnyOf: anyOf
+  "Any of": anyOf
 };
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -43,6 +43,112 @@ function IconBtn(props) {
   );
 }
 
+// Used in the two templates
+function DefaultArrayItem(props) {
+  const btnStyle = {flex: 1, paddingLeft: 6, paddingRight: 6, fontWeight: "bold"};
+  return (
+    <div key={props.index} className={props.className}>
+
+      <div className={props.hasToolbar ? "col-xs-9" : "col-xs-12"}>
+        {props.selectWidget}
+        {props.children}
+      </div>
+
+      {props.hasToolbar ?
+        <div className="col-xs-3 array-item-toolbox">
+          <div className="btn-group" style={{display: "flex", justifyContent: "space-around"}}>
+
+            {props.hasMoveUp || props.hasMoveDown ?
+              <IconBtn icon="arrow-up" className="array-item-move-up"
+                tabIndex="-1"
+                style={btnStyle}
+                disabled={props.disabled || props.readonly || !props.hasMoveUp}
+                onClick={props.onReorderClick(props.index, props.index - 1)}/>
+            : null}
+
+            {props.hasMoveUp || props.hasMoveDown ?
+              <IconBtn icon="arrow-down" className="array-item-move-down"
+                tabIndex="-1"
+                style={btnStyle}
+                disabled={props.disabled || props.readonly || !props.hasMoveDown}
+                onClick={props.onReorderClick(props.index, props.index + 1)}/>
+            : null}
+
+            {props.hasRemove ?
+              <IconBtn type="danger" icon="remove" className="array-item-remove"
+                tabIndex="-1"
+                style={btnStyle}
+                disabled={props.disabled || props.readonly}
+                onClick={props.onDropIndexClick(props.index)}/>
+            : null}
+          </div>
+        </div>
+      : null}
+
+    </div>
+  );
+}
+
+function DefaultFixedArrayFieldTemplate(props) {
+  return (
+    <fieldset className={props.className}>
+
+      <ArrayFieldTitle
+          key={`array-field-title-${props.idSchema.$id}`}
+          TitleField={props.TitleField}
+          idSchema={props.idSchema}
+          title={props.title}
+          required={props.required}/>
+
+      {props.schema.description ? (
+        <div className="field-description" key={`field-description-${props.idSchema.$id}`}>
+          {props.schema.description}
+        </div>
+      ) : null}
+
+      <div className="row array-item-list"
+        key={`array-item-list-${props.idSchema.$id}`}>
+        {props.items && props.items.map(DefaultArrayItem)}
+      </div>
+
+      {props.canAdd ? <AddButton
+                        onClick={props.onAddClick}
+                        disabled={props.disabled || props.readonly}/> : null}
+    </fieldset>
+  );
+}
+
+function DefaultNormalArrayFieldTemplate(props) {
+  return (
+    <fieldset className={props.className}>
+
+      <ArrayFieldTitle
+        key={`array-field-title-${props.idSchema.$id}`}
+        TitleField={props.TitleField}
+        idSchema={props.idSchema}
+        title={props.title}
+        required={props.required}/>
+
+      {props.schema.description ? (
+        <ArrayFieldDescription
+          key={`array-field-description-${props.idSchema.$id}`}
+          DescriptionField={props.DescriptionField}
+          idSchema={props.idSchema}
+          description={props.schema.description}/>
+      ) : null}
+
+      <div className="row array-item-list"
+          key={`array-item-list-${props.idSchema.$id}`}>
+          {props.items && props.items.map(p => DefaultArrayItem(p))}
+      </div>
+
+      {props.canAdd ? <AddButton
+                        onClick={props.onAddClick}
+                        disabled={props.disabled || props.readonly}/> : null}
+    </fieldset>
+  );
+}
+
 class ArrayField extends Component {
   static defaultProps = {
     uiSchema: {},
@@ -59,7 +165,7 @@ class ArrayField extends Component {
     const formData = this.getStateFromProps(props);
     let anyOfItems = [];
     if (this.getAnyOfItemsSchema()) {
-      // We need to contruct the initial anyOfItems state, by searching for the props anyOf items
+      // We need to construct the initial anyOfItems state, by searching for the props anyOf items
       // in the available anyOf schema items
       anyOfItems = this.getAnyOfItemsFromProps(formData.items, props.schema.items.anyOf);
     }
@@ -159,7 +265,9 @@ class ArrayField extends Component {
 
   onDropIndexClick = (index) => {
     return (event) => {
-      event.preventDefault();
+      if (event) {
+        event.preventDefault();
+      }
       const {formData: {items}, anyOfItems} = this.state;
       const newItems = {
         items: items.filter((_, i) => i !== index)
@@ -173,8 +281,10 @@ class ArrayField extends Component {
 
   onReorderClick = (index, newIndex) => {
     return (event) => {
-      event.preventDefault();
-      event.target.blur();
+      if (event) {
+        event.preventDefault();
+        event.target.blur();
+      }
       const {formData: {items}, anyOfItems} = this.state;
 
       const reorder = (items, newIndex) =>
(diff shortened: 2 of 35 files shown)


cyxou:
Why it hasn't been merged yet?

jManji:
@cyxou I am wondering this myself... I am happy to re-work on it, if there are any comments, and update the branch (there's quite a few conflicts as ArrayField has been updated since my PR).

inostia:
+1, my project requires this and is on hold until this gets merged.

Natim:
> +1, my project requires this and is on hold until this gets merged.

Don't hesitate to use this branch in your project in the meantime.

kkarimi:
+1 for merging

Natim:
I guess the most important missing feature is the support of `uiSchema`. The anyOf feature is important to support but this PR doesn't support it yet at the standard we want for that lib.

We need at least to improve the UX of using anyOf as well as supporting the uiSchema.

Merging is easy but supporting an incomplete implementation of the spec in the lib is a path we are not prepared to take.

However this work is a great start and we are highly interesting in this, feel free to keep iterating and improving this Pull-Request and we will merge it as soon as we feel comfortable with maintaining it on the long term.

jManji:
Thanks for the detailed response @Natim! I wasn't aware what was wrong with it, but after your explanation, it makes sense. I remember updating the `uiSchema` was quite challenging, as it does not support dynamically added types... But now that I see there's some interest in this PR, I will keep working on it, first by updating it with master, then trying to add support for `uiSchema`

src/components/fields/ArrayField.js
```diff
@@ -228,35 +284,69 @@ class ArrayField extends Component {
...
   };
 
   onSelectChange = (value) => {
-    this.asyncSetState({items: value});
+    const newState = Object.assign({}, this.state, {formData: {items: value}});
+    this.asyncSetState(newState);
   };
 
+  anyOfOptions(anyOfItems) {
+    return anyOfItems.map(item => ({value: item.type, label: item.type}));
```
exchgr: Should it be `label: item.title`?

exchgr:
Will this be able to support inheritance? Suppose I have an array of widgets. A widget can be of several types: Button, Slider, and Select. Button and Slider have the same attributes, but Select allows the user to define what's in its list. 

The issue I'm running into is that they all need to be `"object"`s, and having multiple items in an `anyOf` array with the same `type` causes it to just select the first item with that `type`.

Here's an example of what I'm trying to do:

```JavaScript
items: {
  anyOf: [
    {
      type: "object",
      title: "Button",

      properties: {
        name: {
          type: "string",
          title: "Name"
        }
      }
    },

    {
      type: "object",
      title: "Slider",

      properties: {
        name: {
          type: "string",
          title: "Name"
        }
      }
    },

    {
      type: "object",
      title: "Select",

      properties: {
        name: {
          type: "string",
          title: "Name",
        },

        fields: {
          "$ref": "#/definitions/select_fields"
        }
      }
    }
  ]
}
```

src/components/fields/ArrayField.js
jManji: Good spot, thanks! I think it should, yes. Maybe `item.title || item.type`, in case it's not defined...

exchgr: Sounds good to me!

jManji:
@exchgr I see what you're saying. This was deliberate as I did not have a use case for multiple items with same type. I _think_ that part of code responsible is [this line](https://github.com/mozilla-services/react-jsonschema-form/pull/417/files#diff-049f0cb4d66526a998749db1e232f07cR338), that tries to find the selected item in the anyOf schema.

It could be _easily_ changed so that `title` is preferred instead `type`. But `title` is not an obligatory property and required by `object`. And it would break for cases where an `anyOf` schema has multiple types with the same title. I _think_ there is a simple custom change you can make if you want. If the `SelectWidget` uses the title as the identifier instead of the type, then it should just work...

I guess the problem lies in the fact that there is no unique identifier between `anyOf` items. Ideally I think it should not reference the items by type, but also by the order in the schema. Or maybe, now that I think about it, it could be just the order... Thanks for the comment, I will check whether order can be used, next time I have a look at it!

exchgr:
@jManji Thanks for the quick response. Going by the array index may solve the issue. I was also thinking that it'd be useful to have a way for the dropdown to influence an attribute that indicates the type on the backend.

Building on my example, suppose we have a few classes on the backend (in this case, Rails). Here's an example with polymorphic associations emulating inheritance across multiple tables (aka Multiple Table Inheritance):

```ruby
class DeviceType < ApplicationRecord
  has_many :controls
  # we're also able to work with each control's "controllable" association directly, through some methods not defined in this example
end

class Control < ApplicationRecord
  belongs_to :device_type
  belongs_to :controllable, polymorphic: true
end

class Button < ApplicationRecord
  has_one :control, as: :controllable
end

class Slider < ApplicationRecord
  has_one :control, as: :controllable
end

class Select < ApplicationRecord
  has_one :control, as: :controllable
end
```

Then in the JSON schema:
```JavaScript
items: {
  anyOf: [
    {
      type: "object",
      title: "Button",

      properties: {
        name: {
          type: "string",
          title: "Name"
        },

        controllable_type: {
          enum: ["Button"]
        }
      }
    },

    {
      type: "object",
      title: "Slider",

      properties: {
        name: {
          type: "string",
          title: "Name"
        },

        controllable_type: {
          enum: ["Slider"]
        }
      }
    },

    {
      type: "object",
      title: "Select",

      properties: {
        name: {
          type: "string",
          title: "Name",
        },

        controllable_type: {
          enum: ["Select"]
        }

        fields: {
          "$ref": "#/definitions/select_fields"
        }
      }
    }
  ]
}
```

Since I actually need this for something I'm working on, I'm going to take a stab at it today.

src/components/fields/ArrayField.js
```diff
@@ -228,35 +284,69 @@ class ArrayField extends Component {
...
+  }
+
+  setWidgetType(index, value) {
+    const {items} = this.state.formData;
+    const {registry} = this.props;
+    const {definitions} = registry;
+    const anyOfItemsSchema = this.getAnyOfItemsSchema();
+    const newItems = items.slice();
+    const foundItem = anyOfItemsSchema.find((element) => element.type === value);
```
glasserc: Does this work with things like `$ref`?

src/components/fields/ArrayField.js
```diff
@@ -481,6 +580,16 @@ class ArrayField extends Component {
     };
     has.toolbar = Object.keys(has).some(key => has[key]);
 
+    const selectWidget = anyOfItemsSchema ? (
+      <div className="form-group" style={{width: 120}}>
+        <SelectWidget
+          schema={{type: "integer", default: selectWidgetValue}}
+          id="select-widget-id"
```
glasserc: The same ID shouldn't be used more than once in a DOM.

src/components/fields/ArrayField.js
```diff
@@ -50,6 +50,7 @@ function DefaultArrayItem(props) {
     <div key={props.index} className={props.className}>
 
       <div className={props.hasToolbar ? "col-xs-9" : "col-xs-12"}>
+        {props.selectWidget}
```
glasserc: It seems kind of strange for `selectWidget` to be part of the props, and not part of the template.

b1r3k:
@glasserc Thanks for review. Argument for extending support for anyOf beyond array makes sense + it's very useful to me at this moment.

For example I would like to use this schema:

```
{
    "definitions": {
        "TPNode": {
            "title": "TPNode",
            "description": "TPNode",
            "type": "object",
            "properties": {
                "name": {
                    "type": "string"
                },
                "value": {
                    "anyOf": [
                        { "type": "string"},
                        {
                            "type": "array",
                            "items": {
                                "$ref": "#/definitions/TPNode"
                            }
                        }
                    ]
                }
            }
        }
    },
    "$ref": "#/definitions/TPNode"
};
```

[Here is how it should render](http://jeremydorn.com/json-editor/?schema=N4IgJgpgZglgdjALjA9nAziAXKAKgBQDkVJtRlEAbCbEA40gGnAnQGMAnGAB2TVvokazRAE9uNLCBQAjAFYQ2iEM24cUEjslZkQcAIYBbSeXGSQ6RFzgBzEAF9mAN32UAriZD64ogPJRsAG1TCVpLaztHEPN9Dg59URUQJAhDTBwQABIOaFoAYgB6SFgEPgwCwVJ7ewBdaurmbNypQuL4JFRyypp7IAA&value=N4IgdghgtgpiBcIQBoQDcIBsCud4G1RJYElUMc8kBfAXWqAA&theme=bootstrap2&iconlib=fontawesome4&object_layout=normal&show_errors=interaction)

@jManji are you intending to work this PR according to @glasserc suggestions any time soon?

src/components/fields/ArrayField.js
b1r3k: I've tried it & it doesn't support $ref

jManji:
Sorry for late reply guys. I've been quite busy lately, and didn't have time to properly look at this. @b1r3k yes, I would very much like to rework on this, when I have some time, based on @glasserc's suggestions. And there's a couple of things that came up from previous conversations here. I would say I should have something in the next couple of weeks. 

@glasserc, thanks a lot for the detailed CR, appreciated! Just trying to answer some of your questions without touching the code yet:
> How come the logic for this is part of the ArrayField class? Just to simplify, I guess?

I think that was because I wanted to make use of the `ArrayField` functionality that is there already, like adding multiple properties, all the UI stuff that comes with it, etc. `anyOf` looks and behaves a bit like an array, so I thought I could fit it in there (like for instance, how `additionalItems` feature does). Maybe a new component type makes more sense? That seemed much more work I guess.

> I also found the UI a little clunky -- rather than having a "select" to choose what kind of object it is, why not use something like Bootstrap's "radio" buttons functionality?

You think radio is better for this case? How would you render lots of options? Will it not occupy too much space and make the form look a bit inconsistent? Not sure, but that's and easy change, if that's what we want.

> Also, why did you move items from arrayField.state.items to arrayField.state.formData.items?

That's a good question... I must have had a reason, I will let you know on my next commit 😄. If I see no reason, I will just move it back to `state.items`

playground/samples/index.js
```diff
@@ -30,4 +31,5 @@ export const samples = {
   Files: files,
   Single: single,
   "Custom Array": customArray,
+  "Any of": anyOf
```
exchgr: In any case (😉), leaving "of" as lowercase is correct, even when you're using title case. http://grammar.yourdictionary.com/capitalization/rules-for-capitalization-in-titles.html

exchgr:
Perhaps if this supports `uiSchema`, we could let the developer decide whether to use radios or selects.

glasserc:
I like @exchgr's suggestion to support uischema. As a default, I'd propose this heuristic: if the anyOf options are only two or three, I think radio buttons are more convenient (just click the one you want); if they are four or more, a select is probably more respectful of screen real estate.

> Maybe a new component type makes more sense?

Could we maybe put it on SchemaField? That was my first instinct (coming from a classical OO background), since any schema could potentially be `anyOf`. On the other hand, that might make SchemaField too complicated. Maybe @n1k0 has an opinion?

n1k0:
I also like the idea of using uiSchema to configure the type selector a lot, though I fear the added complexity. Personally, I'd probably iterate here. Good support for anyOf is already hard to get right :)

Otherwise I concur with @glasserc, `SchemaField` feels like the natural place to implement such support.

Side note, @jManji could you please try to install [prettier](https://www.npmjs.com/package/prettier) v0.22.0 and run:

```
$ prettier --jsx-bracket-same-line --trailing-comma es5 '{playground,src,test}/**/*.js' --write
```

on your current branch, then try to merge/rebase it with latest master? I also fear we're gonna eventually diffing and conflicting a lot if we postpone this operation too much. Thanks!

jManji:
The challenge I see with `uiSchema` is that it is keyed by the property's key. You cannot just say, "I want `anyOf` items to be `textarea`" for instance. You have to define which `anyOf` item you refer to, and since we potentially want (currently this branch does not support it) multiple `anyOf` items of the same type, the `uiSchema` needs to probably reference the order/index of the `anyOf` item (?). I will definitely leave the `uiSchema` out of the scope of the next iteration. If we want, we can start by figuring out how a `uiSchema` for `anyOf` would look like.

@glasserc, @n1k0 Regarding moving code to `SchemaField`, it's been some time since I looked at the code, but currently this also feels challenging/weird to me. It will definitely make `SchemaField` more complicated as you said, and I think, it will need to have some kind of support for dynamically added widgets, like `ArrayField` has. But I am sure you know the codebase much better than me, so I will look into how this can be achieved.

> Side note, @jManji could you please try to install prettier v0.22.0 and run...

Yes, will do!

>... I also fear we're gonna eventually diffing and conflicting a lot if we postpone this operation too much. 

Yes, you are right, I know... I will try to work on this asap, going through every comment mentioned here and in the code review.

b1r3k:
@jManji any news regarding this PR? I'm willing to take it further if you do not have time. In fact I've forked your solution and fixed few mistakes in order to use it in our internal tooling but it has meant to be workaround and the further we go with development the more crippled we are due to divergence from origin/master

jManji:
@b1r3k I am very sorry but unfortunately I don't plan on working on it again, any time soon. I needed that feature for my work, but now we have shifted priorities. Please feel free to work on it!

I haven't checked in detail, but maybe #581 solves the same problem in a different way?

danrot:
I have checked #581, but it looks to me like this is only built to be used in dropdown and selects. From what I've seen it doesn't allow you to build more complex objects than simple strings. 

@b1r3k Have you continued working on it? Anything worth sharing? We would also desperately need a feature like this, although my use case would probably be a `oneOf`, although `anyOf` should also be good enough :-)

intersplice:
Is there any way to install this through npm as a node_module?

Issue #412, opened 2016-11-23
Readonly on radio and checkboxes


crumblix:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I was surprised to find that I could still update radio buttons and checkboxes when they were set to readonly.

While the controls themselves may not support the property directly it seemed to me that that "disabled" was closer than nothing at all.  At least the user could update the data.

While disabled and readonly are subtly different, my suggestion is that using disabled is possibly better than doing nothing.

### Steps to Reproduce

1. [First Step]
Set a field using a RadioWidget or CheckboxesWidget to readonly.

#### Expected behavior
The user can't update those fields

#### Actual behavior
The user can still update the field.

### Version
Latest

Suggestion is to use "disabled" when the readonly flag is passed in to a control that doesn't support readonly inherently.

Fixes in my fork:
Commit: 41053bb3a08672cf78d70cff1dd6624c5a10b463 [41053bb]
Commit: 2573f73a40b21182ff4baa8126874a6639eaa7c5 [2573f73]
Commit: dda05d567ecdf09656538e8c40a9413b11e48300 [dda05d5]

Definitely made a meal of that one too.  Sorry for three commits instead of one!

n1k0:
> While disabled and readonly are subtly different, my suggestion is that using disabled is possibly better than doing nothing.

I agree. Could you please send a PR with your changes so we can properly review and land them? Thanks!

crumblix:
Will do :)

n1k0:
Released in [v0.45.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.45.0).

PR #418, opened 2016-11-23
Readonly on radio and checkboxes #412

crumblix:

While the controls themselves may not support the property directly it seemed to me that that "disabled" was closer than nothing at all. At least the user could update the data.

While disabled and readonly are subtly different, my suggestion is that using disabled is possibly better than doing nothing.

Reference:
Readonly on radio and checkboxes #412

### Checklist

### DIFF
--- a/src/components/widgets/CheckboxesWidget.js
+++ b/src/components/widgets/CheckboxesWidget.js
@@ -20,7 +20,7 @@ function CheckboxesWidget(props) {
     <div className="checkboxes" id={id}>{
       enumOptions.map((option, index) => {
         const checked = value.indexOf(option.value) !== -1;
-        const disabledCls = disabled ? "disabled" : "";
+        const disabledCls = disabled || readonly ? "disabled" : "";
         const checkbox = (
           <span>
             <input type="checkbox"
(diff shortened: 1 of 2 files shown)


crumblix:
While this pull request is still valid, I forgot the recent change for disabledCls.

It should also be:
        const disabledCls = disabled || readonly ? "disabled" : "";

In both files.  But I'll wait for feedback first ... which I could find a way to kill it myself.  But happy to resubmit it with the above change as well.

glasserc:
Hi, sorry about the (unforgivably) long delay in review! I took a quick look at this and it basically looks OK, but it needs a rebase and it doesn't seem to cover `CheckboxWidget` (only `CheckboxesWidget`, which I think is only used for arrays). Could you update it please? I promise I'll review it faster this time  😞

n1k0:
We still want this feature. @crumblix, I feel really bad about the review delay, but would you be willing to keep working on this? Just say "no" and I'll take over the patch. Sorry again.

n1k0:
Bah I see that #470 is the current PR where work is happening. Closing this one.

crumblix:
Sure.  Probably still needs another rebase ... but I'd forgotten about the PR when you asked me last time and so created a new one.

It's more comprehensive.  I'd suggest we close this one and move discussion/review to that PR.

See:
https://github.com/mozilla-services/react-jsonschema-form/pull/470

Issue #411, opened 2016-11-23
Enable the required field on arrays and radio buttons 


crumblix:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description
Following on from my last suggestion:

https://github.com/mozilla-services/react-jsonschema-form/issues/410

I'd like to suggest enabling the "required" field on arrays and radio buttons.

If we make the changes I suggested in the above issue, we can enabled the "required" field on arrays.  The use case would be where we want the user to select "at least one" checkbox in a list.

Also, the RadioWidget input supports the "required" tag, which in addition to being a nicer prompt for the user is also consistent with it's usage elsewhere in the module.

### Steps to Reproduce

1. [First Step]
Attempt to make a radio button list mandatory, or an array mandatory.
2. [Second Step]
For the radio widget instead of a clean "please fill in the field" message that we get when we use the mandatory tag, we get the errorlist returned.
If issue 410 is also addressed then we can make an array mandatory.  Without 410 an empty array is still an array and so it won't work.  With it we can make an array mandatory.  We can't use the "required" HTML attribute, but at least the validation routines will return and prompt us with the error of the missing selection.

#### Expected behavior

For radio buttons:  They should act like any other input that supports the "required" attribute.
For arrays (with 410): We should be able to make it possible to force the user to select "at least one" element in an array.

#### Actual behavior

Radio buttons go straight to the validation routine.  While it works it's not as neat or consistent IMO.
You can't have mandatory arrays currently at all.

### Version
Latest

Again here is the change in my fork:
Commit: a02083712335bd8bac136aab31cf568498b37751 [a020837]

n1k0:
We've discussed this in the past already, I believe the jsonschema spec (which drives this lib implementation) wants `minItems` to be used to define a required array to have at least `n` items attached to an array field. I'm not sure this lib should override the spec with its own opinions for achieving the same results using different ways to express their constraints.

But maybe I'm missing subtleties here; could you please create a jsfiddle highlighting the specific use case we're trying to cover here? Thanks!

crumblix:
WIth regards arrays ... agreed.  When this was raised I wan't aware of minItems.  This has been resolved with better documentation.

[...]

PR #419, opened 2016-11-24
Enable the required field on the radio buttons

crumblix:

I'd like to suggest enabling the "required" field on radio buttons.

Since the RadioWidget input supports the "required" tag, I think it's a nicer prompt for the user is also consistent with it's usage elsewhere in the module.

Relates to Issue:
https://github.com/mozilla-services/react-jsonschema-form/issues/411
### Checklist

### DIFF
--- a/src/components/widgets/RadioWidget.js
+++ b/src/components/widgets/RadioWidget.js
@@ -22,6 +22,7 @@ function RadioWidget({
           <span>
             <input type="radio"
               name={name}
+              required={required}
               value={option.value}
               checked={checked}
               disabled={disabled}


crumblix:
For reference:

http://www.w3schools.com/tags/att_input_required.asp

_Note: The required attribute works with the following input types: text, search, url, tel, email, password, date pickers, number, checkbox, **radio**, and file._

crumblix:
Also ... this PR only is with regards to the RadioWidget and not the arrays mentioned in Issue #411.

I'll address that as a separate PR, since it may be more controversial.

crumblix:
This has been merged under another PR and can be closed :)

PR #431, opened 2016-12-22
Add onBlur event

### LINKED ISSUE #364: onBlur event
### Prerequisites
- [X] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
### Description

Just as we currently have an onChange event I'd love an onBlur, we have some forms that trigger validation or saving on field focus changes. Happy to write the PR if it's of interest.

jbalboni:

We want to be able to show validation errors only after a user has touched a field or tried to submit. We can do the submit part easily through the onError hook and formContext, but we need a way to track blur events. This adds that, similar to the onChange hook.

Should close #364 

### Checklist

  - [X] If I'm adding a new section, I've updated the Table of Content

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -487,7 +487,7 @@ class ArrayField extends Component {
           idSchema={itemIdSchema}
           required={this.isItemRequired(itemSchema)}
           onChange={this.onChangeForIndex(index)}
-            onBlur={onBlur}
+          onBlur={onBlur}
           registry={this.props.registry}
           disabled={this.props.disabled}
           readonly={this.props.readonly}
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -173,6 +173,7 @@ class ArrayField extends Component {
       disabled,
       readonly,
       autofocus,
+      onBlur
     } = this.props;
     const title = (schema.title === undefined) ? name : schema.title;
     const {items} = this.state;
@@ -208,7 +209,8 @@ class ArrayField extends Component {
               itemErrorSchema,
               itemData: items[index],
               itemUiSchema: uiSchema.items,
-              autofocus: autofocus && index === 0
+              autofocus: autofocus && index === 0,
+              onBlur
             });
           })
         }</div>
@@ -220,7 +222,7 @@ class ArrayField extends Component {
   }
 
   renderMultiSelect() {
-    const {schema, idSchema, uiSchema, disabled, readonly, autofocus} = this.props;
+    const {schema, idSchema, uiSchema, disabled, readonly, autofocus, onBlur} = this.props;
     const {items} = this.state;
     const {widgets, definitions} = this.props.registry;
     const itemsSchema = retrieveSchema(schema.items, definitions);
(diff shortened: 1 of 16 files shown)


n1k0:
I like the approach of this patch better than the proposed implementation in #444. I think it's worth having the field unique id passed as the handler first argument as it's the most obvious way to know which field has triggered the event.

Could you please rebase your patch against latest master so we can land this?

src/components/fields/ArrayField.js
```diff
@@ -439,6 +447,7 @@ if (process.env.NODE_ENV !== "production") {
     idSchema: PropTypes.object,
     errorSchema: PropTypes.object,
     onChange: PropTypes.func.isRequired,
+    onBlur: PropTypes.func.isRequired,
```
n1k0: We should probably leave this optional and test for its presence instead.

jbalboni: I think I can just remove the isRequired part here. There's no actual use of onBlur in this component.

jbalboni:
@n1k0 I'm assuming that there's always an onBlur function passed down to the widget components, and I have the conditional check for a user specified onBlur function in the Form component. I could add a conditional in front of the use of onBlur in the widget components instead (I think the other PR did this). Do you have a preference?

n1k0:
> Do you have a preference?

Actually I don't have a strong opinion on this. While the `onChange` fn seems mandatory, the `onBlur` one is more optional to me, so I'd probably mark it that way.

src/components/widgets/BaseInput.js
```diff
@@ -24,7 +25,8 @@ function BaseInput(props) {
       readOnly={readonly}
       autoFocus={autofocus}
       value={typeof value === "undefined" ? "" : value}
-      onChange={_onChange} />
+      onChange={_onChange}
+      onBlur={onBlur ? (event) => onBlur(inputProps.id, event.target.value) : undefined}/>
```
n1k0: I realize we could just do `onBlur && (event) => onBlur(inputProps.id, event.target.value)` here, but that may be harder to read. Opinions?

src/components/widgets/SelectWidget.js
```diff
@@ -18,6 +17,18 @@ function processValue({type, items}, value) {
...
+  let newValue;
+  if (multiple) {
+    newValue = [].filter.call(
+      event.target.options, o => o.selected).map(o => o.value);
+  } else {
+    newValue = event.target.value;
+  }
+
+  return newValue;
+}
```
n1k0: Let's take the opportunity to improve readability here:

```jsx
function getValue(event, multiple) {
  if (multiple) {
    return [].slice.call(event.target.options)
      .filter(o => o.selected)
      .map(o => o.value);
  } else {
    return event.target.value;
  }
}
```

src/components/widgets/BaseInput.js
jbalboni: I think that's better, actually.

n1k0: Let's go for it then!

n1k0:
Great work, thank you.

n1k0:

PR #432, opened 2016-12-22
Add transform hook for changing errors from json schema validation

jbalboni:

The error messages that come from the json schema spec/library (not sure where they come from) aren't very user friendly. They also clash with a pattern that the forms I work have for required validation, namely that those required errors should show up at the field level (rather than at the containing object).

One way of addressing those issues is to allow users to modify the error list sent back by the jsonschema library. Since that has all the error detail produced beyond just a message, it's easy to modify or write your own messages. This PR adds that hook.

I'm skipping updating tests and docs until I get some feedback on whether this is a good approach and something that this library should accommodate. 

Thanks!

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
   - [Form data validation](#form-data-validation)
      - [Live validation](#live-validation)
      - [Custom validation](#custom-validation)
+     - [Custom error messages](#custom-error-messages)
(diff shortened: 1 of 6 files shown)


n1k0:
This is actually smart, I like it. Sure I'd love if the jsonschema lib we used exposed a better API, but I think this nicely solves our current needs with this issue.

Could you please rebase your patch against latest master, and add a few tests and docs? Sorry for being so late to provide feedback.

n1k0:

PR #436, opened 2017-01-02
Don't render a multiselect for Array field if items have `ui:widget` property

olzraiti:

I'm rendering a string array with uniqueItems set true (I use it to get more strict validation) and the array is automatically rendered as a multiselect. If the schema items have a `ui:widget` property, it makes more sense to render a normal array and render the widget for the array items than to render a multiselect.

### Checklist


olzraiti:
I came up with a situation where a string array has an uiSchema for the items, but I DO want to use the multiselect for the rendering... In this case the uiSchema for the items changed & filtered some labels. I'm closing this pull request since I think it's not necessary wanted default behaviour after all.

KurtPreston:
@olzraiti What is the suggested workaround for this behavior?  If the schema has `uniqueItems`, is there any way to render a normal ArrayField?

olzraiti:
I can't remember how I solved this, and I can't come up with a workaround right now... Maybe an uiSchema option for forcing the default array rendering would be useful. :wink:

PR #439, opened 2017-01-12
Don't pass consumed SchemaField class names to child component

olzraiti:

Currently SchemaField passes it's `classNames` prop as-is to the child component. It shouldn't be passed, because SchemaField consumes the given class name. I tried to put some borders to a deeply nested field and I ended up with nested borders, because the nested fields use SchemaField and each of them rerendered the class name.

### Checklist


n1k0:
I don't properly understand the patch and its related use case, sorry. Would you mind adding a test case so we make clear what it's supposed to cover? Also, a comment above the modified line with a reference to this PR might be helpful to readers.

glasserc:
After poking around this code for a while, I think it's OK, and I don't see any reason not to merge it, since you can explicitly provide class names to any nested field you want already.

n1k0:

Issue #410, opened 2016-11-23
Suggestion: in onChange return "undefined" instead of "empty string" or "empty array"


crumblix:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

When there is no data for checkbox or text input ... return undefined instead of empty string or empty array to be consistent with other field types.

From what I can gather based on recent commits this year.  The correct behaviour for NULL or clearing of an input field is to return "undefined".  This will cause the field to be removed from the JSON result data.  This is good in two ways.
1.  It means that an empty string won't raise an error for being an invalid integer (for example).
2.  The mandatory check kicks in.

Let's say I have a Name string field which is required.  Now normally (say for strings) an empty string is not a problem.  As the HTML "required" tag will cause the browser to prompt the user to input a string.  

But now let's assume I have another error as well in my form ... could be anything, but for now assume it's a field whose widget (for whatever reason) doesn't support the "required" tag.   For me this was a custom checkbox widget to support a NULL value.   That's fine but when the error list is now displayed it tells me that my boolean value is required, but doesn't mention the empty Name string field.

Why?  Because an empty string is still a string.  From a validation point of view you HAVE a string.  Basically we're saved in most cases from this by virtue of the fact the "required" tag is kicking in.

I've noticed this on "string" fields and checkbox arrays ....

Yes ... mandatory on a checkbox array.  Although the current code doesn't support it, with this fix if you pass in "undefined" instead of empty array, and then also pass through the "required" parameter to Widget in Arrayfield it works.

You won't be able to use the "required" HTML tag for the checkbox list (that would mean you have to select every item), but the validation routine then picks up the error when you click submit.

I'll be submitting another bug (fix) for the required tag that would enable this shortly, but this would lay the groundwork for that as well.

#### Expected behavior

My suggestion is that to be consistent with other controls, when the string field is empty string "" it should return "undefined" like other controls.  Likewise an empty array [], should never be returned either in onClick, rather it should also return "undefined".

This makes it consistent, and allows the validation to work as well.

#### Actual behavior

String return an empty string "".  Which is still a string for validation purposes.
Arrays (like from checkboxes)  return an empty array []"  Which is still an array for validation purposes.

### Version

Latest.

[...]

Issue #402, opened 2016-11-18
Validation - field Email can not be empty if "touched"


fabarea:
Read the doc OK

### Description

Consider we have a field email with format: email. If we leave the field as pristine - and we don't require the field - we can save it without problem. But as soon as you start typing an email address and then remove it for some reason, the form is not considered valid anymore as field "email" wrongly expects a value.

Live: https://jsfiddle.net/fab1en/yn1ay7op/2/

#### Expected behavior

Field email should only be validated agains format email if it contains a value

atkawa7:
This is because email is not required. Look at 
```required: ["name"]```

fabarea:
I don't want "email" to be required. It must be **optional** in my use case. It only validate against "email" if it has a value.

maartenth:
Good catch, I think that this bug is due to the invisible difference between 'no value' and the empty string: if you don't touch the field, the resulting `formData` looks like `{name: '(name)'}`, but if you do touch the field, the result looks like `{name: '(name)', email: ''}`. I think the right approach for non-required fields would be to always `delete` empty non-required values from the `formData`. That would lead to a better UX in my opinion.

n1k0:
We'd need a reset button for each field, so it can be reinitialized to `undefined`. I could swear we had an issue for it, but couldn't find it back.

crumblix:
OK, so I came across this as part of a separate issue to do with consistency in mandatory fields. (Issue #410).

It is the old battle between empty string and NULL.

My 2cents worth (for what it's worth) is that although empty string is a valid string type, it is rarely if ever used that way in a web context.

An HTML input control does not support NULL.  Only empty string.  And an empty string triggers does not count "as a string" for a mandatory check (via the required attribute).

So while technically the JSON type supports both, it appears (that in a webcontext) most users expect 

We could add a clear or cross box, but no other websites that I know work that way.  If there is an empty string in a required text field, then every system I know of will treat that as NOT a string (for the purposes of being required) .. or in effect treat it as NULL.

As a result currently the HTML required tag kicks in to say that the an empty string is required .... however our validation check says it's a good string.  That's the inconsistency.  So when the error list is returned it may only contain some of the missing required fields.

The options as I see it are:

1.  Live with the inconsistency.  The returned error list on validation will never include a standard text field.
2. Add a clear button (or "cross" button) to be able to send a NULL (undefined) in addition to the empty string from the text input.
[...]

PR #442, opened 2017-01-18
Enable Arrays to be marked as required, and fixes for strings marked as required

crumblix:
Reasons for making this change

New feature:

Enable the arrays to be marked as "required". Meaning that the user must select "at least one" option from the list.

This references the array part of ticket:
#411

It also refers to the string part of
#410 and #402

### Checklist


README.md
```diff
@@ -558,6 +558,8 @@ const uiSchema = {
 };
 ```
 
+In addition it is valid to maked an array as "required".  This means the user must select at least one items to add to the array.
```
MoOx: to maked? Looks like a typo. Maybe "to mark"?

crumblix:
Nice pickup ... thank you!!! :)

src/components/fields/ArrayField.js
```diff
@@ -216,8 +216,12 @@ class ArrayField extends Component {
...
         event.preventDefault();
       }
+      var newitems = this.state.items.filter((_, i) => i !== index);
+      if (newitems.length == 0) {
+        newitems = undefined;
+      }
       this.asyncSetState({
-        items: this.state.items.filter((_, i) => i !== index)
+        items: newitems
       }, {validate: true}); // refs #195
```
```js
const items = this.state.items.filter((_, i) => i !== index);
this.asyncSetState({
  items: items.length > 0 ? items : undefined,
}, {validate: true}); // refs #195
```

src/components/widgets/BaseInput.js
```diff
@@ -8,21 +8,25 @@ function BaseInput(props) {
...
       className="form-control"
       readOnly={readonly}
       autoFocus={autofocus}
       value={typeof value === "undefined" ? "" : value}
-      onChange={(event) => onChange(event.target.value)}/>
+      onChange={(event) => {if (event.target.value == "") {
+        return _onChange(undefined);
+      } else {
+        return _onChange(event.target.value);
+      }}}/>
```
n1k0:
```jsx
const _onChange = ({target: {value}}) => {
  return props.onChange(value === "" ? undefined : value);
};
// ...
onChange={_onChange} />
```

src/components/widgets/CheckboxesWidget.js
```diff
@@ -10,7 +10,12 @@ function selectValue(value, selected, all) {
 }
 
 function deselectValue(value, selected) {
-  return selected.filter(v => v !== value);
+  var ret = selected.filter(v => v !== value);
+  if (ret.length == 0) {
+    return undefined;
+  } else {
+    return ret;    
+  }
```
n1k0:
```js
const ret = selected.filter(v => v !== value);
return ret.length === 0 ? undefined : ret;
```

src/components/widgets/CheckboxesWidget.js
```diff
@@ -19,7 +24,7 @@ function CheckboxesWidget(props) {
   return (
     <div className="checkboxes" id={id}>{
       enumOptions.map((option, index) => {
-        const checked = value.indexOf(option.value) !== -1;
+        const checked = (value === undefined) ? false : value.indexOf(option.value) !== -1;
```
n1k0: Nit: Unnecessary wrapping parenthesis.

src/components/widgets/TextareaWidget.js
```diff
@@ -22,7 +23,11 @@ function TextareaWidget({
       disabled={disabled}
       readOnly={readonly}
       autoFocus={autofocus}
-      onChange={(event) => onChange(event.target.value)}/>
+      onChange={(event) => {if (event.target.value == "") {
+        return _onChange(undefined);
+      } else {
+        return _onChange(event.target.value);
+      }}}/>
```
n1k0: See previous impl. suggestion above.

README.md
```diff
@@ -558,6 +558,8 @@ const uiSchema = {
 };
 ```
 
+In addition it is valid to mark an array as "required".  This means the user must select at least one item to add to the array.
```
crumblix: Fair call.  I wasn't aware of that directive.  If we can achieve a similar effect with existing code that is standards compliant I would certainly support that.

crumblix:
Thank you very much for the code review!  Much appreciated, I have no problem addressing a few nits :)

README.md
MoOx: Indeed that's a serious diff. Not sure I am comfortable with that but I guess we can't have the silver bullet.

crumblix: The SelectWidget is an interesting case.  In truth I don't use it much because I've written my own custom widget based on react-select which better suited my needs for a nullable and multiselect-able drop down widget.  (Happy to share if anyone wants it, but I thought since it was an extra dependency it was better off outside the core project.)

Although it flies in the face of this change, I would tend to opt the other direction for the SelectWidget ... if only because the programmer has explicitly set the valid return values so should be expecting it, so perhaps we don't need to try and interpret their intent.

olzraiti: This wouldn't affect my code base, but I'm not sure if we would benefit enough from adding this considering the consequences.

We aren't extending the json schema spec in a way that would make something possible that was impossible before (using `minItem` provides the functionality that is wanted here). We are merely making something more convenient than before, with the cost of breaking compatibility with json schema spec.

crumblix: Yep.  More than happy to leave the array case out.  If we can achieve the same result using minItem and it's more standards compliant I can't see why we wouldn't.

maartenth: I am in favor of this change. It is consequent with the empty string to undefined functionality. The error message you get when you don't select a value is then always the same, whether you never selected a value or you selected and then deselected a value. If something looks the same in the ui, I think it should be the same.

For JSON Schema, the empty array is of course a valid array, but JSON Schema is not a form builder and the difference between an empty array and undefined is simply not visible in a form and should as a consequence not be settable through the form, in my opinion.

n1k0: > JSON Schema is not a form builder

The issue is that people import existing schemas and expect them to work with forms with no need to tweak their schemas (uiSchemas have been introduced for that specific purpose). 

That could be addressed with an appropriate FAQ/Caveats section though. The same way we could document the `minItems` jsonschema directive.

> the difference between an empty array and undefined is simply not visible in a form

That's indeed a fair point. Another approach would be to make the difference actually visible, but that's rather tricky to get right, and would probably clutter UX a lot. Also this would have to work with bootstrap and other styles.

So my current thinking is:

1. Keep the *empty string as undefined value* feature from this patch.
1. Remove the *required array should contain at least one item* one and add proper docs describing how to achieve it using the `minItems` jsonschema directive.
1. Leave `SelectWidget` as it is, users should provide enum values matching their expectations.

What do people think?

src/components/fields/ArrayField.js
```diff
@@ -216,8 +216,9 @@ class ArrayField extends Component {
       if (event) {
         event.preventDefault();
       }
+      var items = this.state.items.filter((_, i) => i !== index);
```
dehli: I think should be a `const` or `let`. It looks like you never assign to it again, so a `const` would be the way to go.

README.md
dehli: I agree with you @n1k0 . I'm currently using minItems in my application b/c that's the `json-schema` way of doing it. If we added documentation describing it, then it'll be more obvious that it's the way to do it.

olzraiti: Sounds good to me! Empty string as undefined seems well-grounded.

maartenth: I'm OK with your proposed solution, @n1k0. I think it is an improvement to the current situation either way.

n1k0: Good, @crumblix would you mind updating the patch accordingly if you agree with this decision? :)

Thanks a bunch :+1:

crumblix:
Thanks again everyone for the feedback :)
I reverted the documentation and fixed the last nit to make the code cleaner and better.

I'll need to read up on minItems before making any changes to the documentation for that as I haven't used it before.

My plan is to try and get it working in the playground later before updating the documentation about it.

Either that or if someone who knows how it works already sends me a sentence I'll just include it.

crumblix:
OK, so I've reverted the array changes, played around with minItems in the playground, and tried to make it clear how to achieve a similar (and more feature rich) result by using it in the documentation.

Thanks for pointing out minItems!

README.md
```diff
@@ -547,6 +547,23 @@ const uiSchema = {
 };
 ```
 
+Unlike other types, array's do not support the `required` property.  You can however instead specify the minimum number of items the user must select with the `minItems` property.
```
n1k0: It actually supports the `required` property, it only checks that an array property value is provided, which may or may not be empty. Put another way, for

```js
schema = {
  type: "object",
  required: ["foo"],
  properties: {
    foo: {type: "array"}
  }
}
```

`{foo: []}` is valid against this schema while `{}` is not.

crumblix: No problem!  While it's true supports the required property, I suspect that it doesn't operate in the fashion most UI builders would expect, at the very least it's a potential source of confusion.  I'd try and update the documentation to be clear in that regard. :)

And yes to the tests as well! :)

n1k0: > I suspect that it doesn't operate in the fashion most UI builders would expect

I think we need to stay true to the jsonschema spec as much as possible, as it's our data model layer behind our form views. One of the goal of this lib is also to make forms and their underlying data model more robust and consistent. I wouldn't want to bring regular web forms flakiness to jsonschema, which would kind of defeat its purpose if you ask me :)

README.md
```diff
@@ -547,7 +547,7 @@ const uiSchema = {
 };
 ```
 
-Unlike other types, array's do not support the `required` property.  You can however instead specify the minimum number of items the user must select with the `minItems` property.
+Care should be taken when using the `required` property with arrays.  An empty array is sufficient to pass that validation check.  If you wish to ensure the user populates the array, you can specify the minimum number of items the user must select with the `minItems` property.
```
n1k0: This is perfect :+1:

crumblix: Excellent :)
I'll have to get to the tests over the weekend.  It's a part of the code base I haven't been into yet.

crumblix: Unfortunately there wasn't a text area widget section that I could see so I created one .... but it isn't right.  I will have to look at it over the weekend.  

What's checked in is passing, but it's not using the widget.

I've tried the change:

    const uiSchema = {"ui:widget": "textarea"};

    it("should handle an empty string change event", () => {
      const {comp, node} = createFormComponent({schema: {
        type: "string",
      }, uiSchema});

... which seems to me like it should work but that fails for reasons that currently escape me.  I'll need to look at it over the weekend.  Got to head off now though!

crumblix:
You're welcome :), but I think Textarea widget test is wrong ... isn't it?

n1k0:
> I think Textarea widget test is wrong ... isn't it?

It was, fixed by 8c4707a125bc191c9ac3bf6c39d706062a26aa2b :)

n1k0:

PR #450, opened 2017-01-24
Add support for rows attribute of textarea widget.

NessunKim:

Add support for rows attribute of textarea widget.
Now users can specify initial height of it.

### Checklist

### DIFF
--- a/src/components/widgets/TextareaWidget.js
+++ b/src/components/widgets/TextareaWidget.js
@@ -4,6 +4,7 @@ import React, {PropTypes} from "react";
 function TextareaWidget({
   schema,
   id,
+  options,
   placeholder,
   value,
   required,
@@ -15,7 +16,7 @@ function TextareaWidget({
 }) {
   const _onChange = ({target: {value}}) => {
     return onChange(value === "" ? undefined : value);
-  };  
+  };
   return (
     <textarea
       id={id}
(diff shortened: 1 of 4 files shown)


src/components/widgets/TextareaWidget.js
```diff
@@ -40,6 +42,10 @@ if (process.env.NODE_ENV !== "production") {
     schema: PropTypes.object.isRequired,
     id: PropTypes.string.isRequired,
     placeholder: PropTypes.string,
+    options: PropTypes.shape({
+      enumOptions: PropTypes.bool,
```
n1k0: We don't support `enumOptions` for textareas.

src/components/widgets/TextareaWidget.js
```diff
@@ -26,6 +27,7 @@ function TextareaWidget({
       disabled={disabled}
       readOnly={readonly}
       autoFocus={autofocus}
+      rows={options.rows}
```
n1k0: We need to ensure that options is always an object before accessing this property. We may want to define `{}` a default value for the `options` arg.

src/components/widgets/TextareaWidget.js
```diff
@@ -40,6 +42,10 @@ if (process.env.NODE_ENV !== "production") {
     schema: PropTypes.object.isRequired,
     id: PropTypes.string.isRequired,
     placeholder: PropTypes.string,
+    options: PropTypes.shape({
+      enumOptions: PropTypes.bool,
+      rows: PropTypes.number
+    }).isRequired,
```
n1k0: Please ensure that this is always actually required (I forgot about it).

n1k0:
Poke.

elisechant:
@NessunKim these changes are important so we can make a release.

n1k0:

PR #451, opened 2017-01-24
Allow empty option for enum fields

### LINKED ISSUE #449: Option to turn off enum fields defaulting to first item
### Prerequisites

### Description

We have string fields that use enums and are required fields. On the form, we'd like to not have them automatically default to the first item in the enum, so that the required validation is triggered if a user skips the field. Ideally we could use the `default` property in the schema for that, but in our case the default should really be `undefined`. So it would be good to have the ability to skip the enum specific default behavior and use the normal string field behavior.

This could be a Form component prop or an option in the uiSchema for a field, I think. I don't see a way to do it in the normal schema. At the form level would work for me.

I can submit a PR for this if you're open to it.

jbalboni:

This removes the logic that defaults enum fields to the first item and updates the date and select widgets to use an empty default option when a field is undefined.

This ended up being more difficult to get right at the widget level than I thought and I've essentially skipped making this change affect `multiple` arrays. This needs some careful checking for the cases that aren't on the simpler string field using a select widget path.

Closes #449.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
      - [Auto focus](#auto-focus)
      - [Placeholders](#placeholders)
      - [Form attributes](#form-attributes)
+     - [Enum fields](#enum-fields)
   - [Advanced customization](#advanced-customization)
      - [Field template](#field-template)
      - [Array field template](#array-field-template)
@@ -680,6 +681,10 @@ Form component supports the following html attributes:
   acceptcharset="ISO-8859-1"
   schema={} />
 ```
+
+### Enum fields
+
+String fields that use `enum` and a `select` widget will have an empty option in the options list. When a user selects that option, the field will be set to `undefined` (similar to how regular `string` fields work if the field is empty). This also means that if you have an empty string in your `enum` array, selecting that option will cause the field to be set to `undefined`.
 
 ## Advanced customization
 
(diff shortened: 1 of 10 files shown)


src/components/widgets/SelectWidget.js
```diff
@@ -7,7 +7,9 @@ import {asNumber} from "../../utils";
  * always retrieved as strings.
  */
 function processValue({type, items}, value) {
-  if (type === "array" && items && ["number", "integer"].includes(items.type)) {
+  if (value === "") {
+    return undefined;
```
n1k0: We need to add a warning somewhere in the docs, because if one adds `""` as an enum choice in their schema, it will be automatically converted to `undefined` and drop the property from any parent object - which is now consistent with our text inputs but may be surprising to users.

jbalboni: I added a note, but I'm not sure how well I explained it or if there's a better spot.

README.md
```diff
@@ -681,6 +682,10 @@ Form component supports the following html attributes:
   schema={} />
 ```
 
+### Enum fields
+
+String fields that use `enum` and a `select` widget will have an empty option in the options list. When a user selects that option, the field will be set to `undefined` (similar to how regular `string` fields work if the field is empty). This also means that if you have an empty string in your `enum` array, selecting that option will cause the field to be set to `undefined`.
```
n1k0: I think I'd move this in new section under the *Form data validation*, eg. *The case of empty strings*.

n1k0:
Perfect, thank you!

n1k0:

PR #470, opened 2017-02-14
Issue 412:  Readonly fixes to checkboxes, radion buttons, dropdowns, …

crumblix:
Issue #412:  Readonly fixes to checkboxes, radio buttons, dropdowns and color widget.

Many controls are still editable in the form when the schema is set to readonly.  Even though these controls may not explicitly support the readonly keyword, it is anticipated that disabled in a closer match to than ignoring the directive,

Initially this was just checkboxes and radio buttons, but in the course of testing in the playground I also encountered similar problems with the dropdowns and color select control.

Please review the `ColorWidget` in particular, not sure what styling you'd prefer.  I went for two "returns" and only unpacked the props if needed, but it obviously can be done differently if you'd prefer.

Tested in playground on by setting

```js
        "ui:readonly": true
```

Many controls affected on many tabs:

- Nested Tab:  "tasks"."items"."done"
- Arrays Tab:  "multipleChoicesList", "fixedItemsList"."a boolean value", "fixedNoToolbar"."a boolean"
- Numbers Tab: "numberEnum", "numberEnumRadio"
- Widgets Tab: "boolean"."default", "boolean"."radio", "boolean"."select", "strings"."color", "selectWidgetOptions"
- Errors Tab: "active", "multipleChoicesList"
- Large:  everything
- Date & Time: all alt-datetime dropdowns

**NOTE:**  The following items still do NOT respond to setting readonly in the playground:

In the custom tab the "geo" component.  As this is a custom component rather than a core component that is possibly acceptable.

In the "Custom Array" tab, the buttons still work ... but I'm not sure I'm the best person to tackle that fix.  I'd happily accept and include the change, or it can just go in another PR.

### Checklist

### DIFF
--- a/src/components/widgets/BaseInput.js
+++ b/src/components/widgets/BaseInput.js
@@ -7,6 +7,7 @@ function BaseInput(props) {
   const {
     value,
     readonly,
+    disabled,
     autofocus,
     onBlur,
     options,  // eslint-disable-line
@@ -23,6 +24,7 @@ function BaseInput(props) {
       {...inputProps}
       className="form-control"
       readOnly={readonly}
+      disabled={disabled}
       autoFocus={autofocus}
       value={typeof value === "undefined" ? "" : value}
       onChange={_onChange}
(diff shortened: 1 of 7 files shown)


n1k0:
> Would you be willing to update the patch?

I came to realize how tedious this was and did the dirty job in #525.

Issue #471, opened 2017-02-14
setting default value for the string input refills the field when emptied up


Liooo:
As is shown below from the playground page.
![bug](https://cloud.githubusercontent.com/assets/1630378/22921510/634f1912-f2dd-11e6-8517-c6d468801a1d.gif)

Is this a desired behaviour? If so, is there any options to set initial value while allowing the field to be empty?

n1k0:
This looks like a bug indeed!

n1k0:
This looks to be a side effect of #442; now we propagate `undefined` when a text field is cleared, the default value defined in the schema is set automatically.

n1k0:
I'm adding field clearing buttons in #472. That's always been my initial plan, I'm sticking to it.

Liooo:
@n1k0 Thanks for the quick response.

n1k0:
@Liooo I've deployed #472 to [gh-pages](https://mozilla-services.github.io/react-jsonschema-form/), would you mind providing feedback? thanks

crumblix:
The problem has simply shifted to "clear". (check out the example on the current live playground).

The real fix to is only apply defaults on create of the item. 

At that point you can have your cake and eat it too.  You can have a simple UI and have required strings that work as the user expects without "clearing" buttons, or complex ones with the clear button.

While I like the clear button, I do not expect my users will understand it in the context of strings and single checkboxes.   Yes they'll get the "clear" part on a string ... but understanding that that is difference from manually deleting the string is something they won't get.  And then we hit the same problem as earlier again.  So still very much against reverting #442.

Perhaps an option on whether to treat "empty string" as "undefined" once we have the clear boxes?  It can default either way, but I personally don't want the UI acting differently if the users hit the "clear" or manually deletes the string.

crumblix:
OK ... I've investigated this further.  It actually looks to me like it's a bug in StringField.js.

On line 39.  Simply replace:
    value={defaultFieldValue(formData, schema)}
with
    value={formData}

This way the default only applies on creation.  This should work independently I think on whether you clear it manually or use the clear button.

Currently in the live environment if you clear it displays "bazinga" on the screen but it's null in the formData.  I would expect this change would work with both that and #442.

Note there is also what appears to be a bug with the current implementation regardless of this change (and still exists afterwards).  The defaults in general appear on screen but not in the formData.  I tested this on the simple tab of the playground by making the following changes (necessarily in code):

      firstName: {
        type: "string",
[...]

PR #472, opened 2017-02-14
Refs #471: Introducing field clearing button.

n1k0:
WiP, introducing clear buttons for fields. The blank-string-as-undefined strategy introduced by #442 brings too much trouble, eg. #471. So I'm finally working on adding clear buttons, as initially intended.

### Todo

- [x] Text inputs (including numbers, textareas)
- [x] Boolean fields
- [x] Enums (select)
- [x] Handling `disabled` and `readonly`
- [x] Disabling the button for already cleared fields

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -5,7 +5,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>react-jsonschema-form playground</title>
-  <link rel="stylesheet" id="theme" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css">
+  <link rel="stylesheet" id="theme" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
   <script src="//cdn.polyfill.io/v2/polyfill.min.js"></script>
 </head>
 <body>
--- a/playground/index.prod.html
+++ b/playground/index.prod.html
@@ -5,7 +5,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>react-jsonschema-form playground</title>
-  <link rel="stylesheet" id="theme" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css">
+  <link rel="stylesheet" id="theme" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
   <link rel="stylesheet" href="./styles.css">
   <script src="//cdn.polyfill.io/v2/polyfill.min.js"></script>
 </head>
--- a/playground/samples/errors.js
+++ b/playground/samples/errors.js
@@ -7,7 +7,8 @@ module.exports = {
         type: "string",
         title: "First name",
         minLength: 8,
-        pattern: "\\d+"
+        pattern: "\\d+",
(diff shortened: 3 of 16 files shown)


n1k0:
Patch has been deployed to [gh-pages](https://github.com/mozilla-services/react-jsonschema-form) for easier testing.

Feedback? @Natim @leplatrem @olzraiti @crumblix and anybody interested

crumblix:
I think having the ability to clear a field is an excellent addition and the feature can be tremendously useful.

A couple of thoughts.  

I'd want an option to disable it, both entirely for the form and as an override on a specific control.  With so many buttons potentially on a page it could confuse users  It can make for a very busy form on some pages (particularly on forms like on the arrays tab in the playground), and I'd struggle to sell the "clear" on the checkbox to any non technical person!

My initial thoughts on the cross itself was I'd prefer it to be a little less obvious.  Perhaps the cross can go red on hover and be a little less visible when the mouse isn't over it?  However like many things once you get used to it, it grows on you, so perhaps it's fine as is.

We also now have two crosses for clearing, this one and the one for clearing an entry in an array.  Actually I think I prefer this one, but if we keep it I'd prefer it consistent and that may mean restyling the up/down/plus buttons (probably not a bad thing, they can be very big and in your face).

n1k0:
> I'd struggle to sell the "clear" on the checkbox to any non technical person!

After a good night of sleep, I have to agree... what the hell was I thinking :) Though we still need to expose a way to distinguish bewteen *blanking* and *clearing* (or resetting) a field value, as it strongly impacts the resultling submitted data. It's a hard problem to solve wrt UX, and this was my first attempt at tackling it, but there's obviously more work to be done.

Meanwhile having an option to enable these buttons is a good idea, though that will probably involve a bunch of supplementary code. I really like the idea of having the "simplest" (if we can still use this word with this lib) possible forms.

I'll spend some more time toying around with the UI today, to see if I can come with something less cluttered and invasive, something like [this](https://codepen.io/shidhincr/pen/ICLBD) maybe (but prettier).

crumblix:
Yes agreed :)

That example is much more palatable.  It isn't in your face and yet allows you an easy clear.

I think the option is good.  As you say sometimes we do need to distinguish between clearing and blanking ... but sometimes we also don't and confusing users in those situations doesn't help.

:)

n1k0:
I'm updating the u(i,x) on top of the previous patch so it only renders the clear button when the field has the focus:

demo

n1k0:
Went way too crazy with the focus stuff, back to working on something much lighter:



If we make this optional (and turned off by default), it may be just good enough. Thoughts?

crumblix:
Yes.  Optional and turned off by default sounds good.  I'd personally like to have fine grained access so (if I'm using it) I can choose the controls that require it, if only for checkboxes! :)

Personally I like losing the square around the cross.  That looks neater.  But I didn't mind the focus approach at all.  It makes the UI less busy when it's not an option and to my eye makes it less complicated.

n1k0:
This is way too complicated. I'm gonna close this in favor of #476 which seems to nicely address the concerns we have with nullable fields. We may want to revisit the opportunity to distinguish between blanked and cleared fields in the future, where this would be a good starting point.

mravenash:
> Patch has been deployed to [gh-pages](https://mozilla-services.github.io/react-jsonschema-form/) for easier testing.
> 
> Feedback? @Liooo @Natim @leplatrem @glasserc @MoOx @olzraiti @crumblix and anybody interested

Is this feature available?

mravenash:
is there a way clear fields for an input textfield/widgets with a button?

benedictantony:
Hey, I wish if we get a "Delete All" option for the array items elements!! Currently we have the option for deleting individual elements. Will it be possible to get a functionality for deleting the formData for all the items at a click?

Issue #473, opened 2017-02-14
[Playground] Error in "Single" after having loaded the "Validation" example 


n1k0:
### Steps to Reproduce

1. Head to the playground https://mozilla-services.github.io/react-jsonschema-form/
1. Head to the *Validation* section
1. Fiddle with the form
1. Now head to the *Single* section
1. Clear the value in the string field
1. Watch the browser console:

```
Uncaught TypeError: Cannot read property 'pass1' of undefined
    at validate (validation.js:4)
    at validateFormData (validate.js:156)
    at Form.validate (Form.js:181)
    at Form.proxiedMethod (createPrototypeProxy.js:44)
    at Form.getStateFromProps (Form.js:150)
    at Form.proxiedMethod (createPrototypeProxy.js:44)
    at Form.componentWillReceiveProps (Form.js:135)
    at Form.proxiedMethod (createPrototypeProxy.js:44)
    at eval (ReactCompositeComponent.js:608)
    at measureLifeCyclePerf (ReactCompositeComponent.js:74)
```

crumblix:
I can reproduced on the live hosted playground, but it didn't occur on my local copy.

Nevertheless, I am fairly certain that the problem is to do with the missing export of uiSchema from single.js.  Give it a go ... even if it doesn't end up fixing this specific problem it's a worthwhile patch and makes the example clearer.

knilink:
```js
// Head to the Validation section
// app component state
{
  schema: validationSchema,
  formData: validationFormData,
  validation: validationValidate,
}

// Now head to the Single section
setState({
  schema: singleSchema,
  formData: singleFormData,
});
// new app state
{
  schema: singleSchema,
  formData: singleFormData,
  validation: validationValidate,
}

// Clear the value in the string field
{
  schema: singleSchema,
  formData: undefined,
  validation: validationValidate,
}
validateFormData(undefined, singleSchema, validationValidate); 
```

epicfaace:
This seems to be gone now, with the new Monaco editor.

PR #475, opened 2017-02-14
Issue #473: Fix uiSchema in playground "Single" tab.

crumblix:

Missing uiSchema export for "Single" tab in playground.

[Playground] Error in "Single" after having loaded the "Validation" example #473

### Checklist


n1k0:
Thanks!

PR #478, opened 2017-02-17
Fix/add debounce to base input

matt-wright86:
The user would lose characters when entering text into an input
[Please describe them here]

### Checklist


n1k0:
Please send a PR properly describing what you're trying to achieve, and implementing a generic solution which would work without landing your specific needs and package name changes. 

Thanks for your understanding.

Issue #399, opened 2016-11-16
Integrating  with other widgets such as react-select or react-tags #396


atkawa7:
How do you make sure the rest of fields render with default widgets and then the custom field with your custom widget.

revolunet:
not sure i understand correctly the question but here's an example where i use some custom widget only for a given field (description)

```js
import Form from 'react-jsonschema-form';

// dummy widget example, just with value & onChange
const MyCustomWidget = ({ formData, onChange }) => <div>{ formData }</div>;

const formProps = {
  fields: {
    description: MyCustomWidget
  },
  schema: {
    title: 'Description',
    type: 'object',
    required: ['description'],
    properties: {
      description: {type: 'string', title: 'Text'},
      notes: {type: 'string', title: 'Notes'},
    }
  },
  uiSchema: {
    description: {
      'ui:field': 'description',
      'ui:options': {
        size: 'big'
      }
    }
  },
  formData: {
    description: "some description",
  }
};

const Demo2 = () => (
  <div>
    <Form { ...formProps } />
  </div>
)
```

atkawa7:
I am trying to create a list of genres using reactinput tags

here  is what I have but I am still getting error 

```
import React from 'react';
import Form from 'react-jsonschema-form';
import SchemaTagField from './SchemaTagField';

const schema = {
    type: 'object',
    required: [
        'name', 'genres'
    ],
    properties: {
        name: {
            type: 'string'
        },
        genres: {
            type: 'array',
            items: {
                type: 'string',
                enum: [],
                enumNames: []
            }
        }

    }
};

const fields = {
    items: SchemaTagField
};
const uiSchema = {
    genres: {
        'ui:field': 'items'
    }
};

export default class Test extends React.Component {
    onChange(event) {
        console.log(event);
    }
    render() {
        return (
            <div className='col-md-4 col-md-offset-4'>
                <Form schema={schema} uiSchema={uiSchema} fields={fields}/></div>
        );

import React from 'react';
import TagsInput from 'react-tagsinput';
import '../css/tags.css';

export default class SchemaTagField extends React.Component {
    constructor(props) {
        super(props)
        this.state = {
            tags: []
        }
    }

    handleChange(tags) {
        this.setState({
            tags
        }, function() {
            var obj = {
                [this.props.uiSchema['ui:field']]: this.state.tags
            };
            // console.log(obj);
            this.props.onChange(obj);
        }.bind(this));
    }

    render() {
[...]

PR #480, opened 2017-02-21
Remove setState in ObjectField and ArrayField

knilink:

The preserving state in "ObjectField" and "ArrayField" was unnecessary because the root Form component had done the job(Form.js line 152). It may be the root cause of #339. (Haven't test it in IE but it did fix my losing characters issue).

Performance test has been updated due to componentWillReceiveProps in "ObjectField" and "ArrayField" are no longer necessary and have been removed.

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -159,10 +159,6 @@ class ArrayField extends Component {
     autofocus: false,
   };
 
-  constructor(props) {
-    super(props);
-  }
-
   shouldComponentUpdate(nextProps, nextState) {
     return shouldRender(this, nextProps, nextState);
   }
@@ -196,7 +192,7 @@ class ArrayField extends Component {
         event.preventDefault();
       }
       this.props.onChange(
-        this.props.formData.filter((_, i)=> i !== index),
+        this.props.formData.filter((_, i) => i !== index),
         {validate: true} // refs #195
       );
     };
@@ -257,6 +253,7 @@ class ArrayField extends Component {
     const {
       schema,
       uiSchema,
+      formData,
       errorSchema,
       idSchema,
       name,
@@ -269,7 +266,6 @@ class ArrayField extends Component {
       onBlur
     } = this.props;
     const title = (schema.title === undefined) ? name : schema.title;
-    const items = this.props.formData;
     const {ArrayFieldTemplate, definitions, fields} = registry;
     const {TitleField, DescriptionField} = fields;
     const itemsSchema = retrieveSchema(schema.items, definitions);
@@ -277,18 +273,18 @@ class ArrayField extends Component {
 
     const arrayProps = {
       canAdd: addable,
-      items: items.map((item, index) => {
+      items: formData.map((item, index) => {
         const itemErrorSchema = errorSchema ? errorSchema[index] : undefined;
         const itemIdPrefix = idSchema.$id + "_" + index;
         const itemIdSchema = toIdSchema(itemsSchema, itemIdPrefix, definitions);
         return this.renderArrayFieldItem({
           index,
           canMoveUp: index > 0,
-          canMoveDown: index < items.length - 1,
+          canMoveDown: index < formData.length - 1,
           itemSchema: itemsSchema,
           itemIdSchema,
           itemErrorSchema,
-          itemData: items[index],
+          itemData: formData[index],
           itemUiSchema: uiSchema.items,
           autofocus: autofocus && index === 0,
           onBlur
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -12,8 +12,7 @@ import {
   retrieveSchema,
   toIdSchema,
   shouldRender,
-  getDefaultRegistry,
-  setState
+  getDefaultRegistry
 } from "../../utils";
 
 function ArrayFieldTitle({TitleField, idSchema, title, required}) {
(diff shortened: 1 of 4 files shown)


src/components/fields/ObjectField.js
```diff
@@ -71,21 +33,21 @@ class ObjectField extends Component {
...
-  }
-
   onPropertyChange = (name) => {
     return (value, options) => {
-      this.asyncSetState({[name]: value}, options);
+      const newFormData = Object.assign(
+        {},
+        this.props.formData,
+        {[name]: value}
+      );
```
n1k0: You can do:

```js
const newFormData = {...this.props.formData, [name]: value};
```

src/components/fields/ArrayField.js
```diff
@@ -161,19 +161,6 @@ class ArrayField extends Component {
 
   constructor(props) {
     super(props);
-    this.state = this.getStateFromProps(props);
-  }
```
n1k0: The constructor is now unnecessary. Also that means we don't have local state anymore, which means we should use a functional stateless component instead of a component class. Which is so nice.

src/components/fields/ObjectField.js
```diff
@@ -40,25 +21,6 @@ class ObjectField extends Component {
 
   constructor(props) {
     super(props);
-    this.state = this.getStateFromProps(props);
-  }
```
n1k0: Same remark as with `ArrayField`, this should become a stateless component.

src/components/fields/ArrayField.js
```diff
@@ -189,36 +176,29 @@ class ArrayField extends Component {
...
       itemSchema = schema.additionalItems;
     }
-    this.asyncSetState({
-      items: items.concat([
-        getDefaultFormState(itemSchema, undefined, definitions)
-      ])
-    });
+    this.props.onChange([
+      ...formData,
+      getDefaultFormState(itemSchema, undefined, definitions)
```
n1k0: It may be worth renaming `utils.getDefaultFormState` which now sounds a little confusing. How about something like `initDefaults`?

src/components/fields/ArrayField.js
```diff
@@ -189,36 +176,29 @@ class ArrayField extends Component {
...
   onDropIndexClick = (index) => {
     return (event) => {
       if (event) {
         event.preventDefault();
       }
-      this.asyncSetState({
-        items: this.state.items.filter((_, i) => i !== index)
-      }, {validate: true}); // refs #195
+      this.props.onChange(
+        this.props.formData.filter((_, i)=> i !== index),
```
n1k0: Nit: missing space before `=>`.

src/components/fields/ArrayField.js
```diff
@@ -228,33 +208,35 @@ class ArrayField extends Component {
         event.preventDefault();
         event.target.blur();
       }
-      const {items} = this.state;
-      this.asyncSetState({
-        items: items.map((item, i) => {
+      const items = this.props.formData;
```
n1k0: Nit: this can alternatively be written as `const {formData: items} = this.props;`, which I personally like a lot. But here we could probably just use `formData` as it makes sense that they're a list of items in an ArrayField.

src/components/fields/ArrayField.js
```diff
@@ -12,8 +12,7 @@ import {
   retrieveSchema,
   toIdSchema,
   shouldRender,
-  getDefaultRegistry,
-  setState
```
n1k0: We should check that this is still needed elsewhere. If not, we should remove that helper function, which is now obsolete.

src/components/fields/ObjectField.js
```diff
@@ -40,25 +21,6 @@ class ObjectField extends Component {
...
-
-  componentWillReceiveProps(nextProps) {
-    const state = this.getStateFromProps(nextProps);
-    const {formData} = nextProps;
-    if (formData && objectKeysHaveChanged(formData, this.state)) {
-      // We *need* to replace state entirely here has we have received formData
-      // holding different keys (so with some removed).
-      this.state = state;
-      this.forceUpdate();
```
n1k0: I'm so happy seeing this going away :+1:

Gurkenschreck:
Nice work, I hope this fixes #399 .
Can you tell me when this will be merged and published?

n1k0:
Once review comments are addressed.

src/components/fields/ArrayField.js
knilink: I'm not sure whether it is appropriate to do it at the moment because getDefaultFormState is used  by the root Form component as well.

n1k0:
:+1:

src/components/fields/ObjectField.js
MoOx: Me too :)

Gurkenschreck:
Yes I can do that.

MoOx:
Looks fantastic according to my test!

jManji:
If/when I get around adding support for `anyOf` as part of #417, I _think_ I will need to bring the state back, but probably only using it for `ArrayField`'s internal own stuff, and not for getting the `formData` as the code before this PR does.

Gurkenschreck:
Is anywhere stated what browsers are supported? The branch gh-pages cannot be seen in IE9. 

Console output of https://mozilla-services.github.io/react-jsonschema-form/ in IE9
```
SCRIPT438: Object doesn't support property or method 'bind' 
bundle.js, line 1 character 226
```

I would raise an issue if IE9 should be supported.

n1k0:
We're not planning on supporting IE9. It looks like it could be achieved by including the right polyfills though, but I'd like to let developers taking care of this.

But you're right we should document what our current target browser versions are.

n1k0:

israelshirk:
<3

PR #481, opened 2017-02-21
Fix nesting in uiSchema example

jyustman:

Fixes nesting issue with the object structure in the `uiSchema` example.

### Checklist


n1k0:
Woah, thanks. I can't believe this went unnoticed for so long...

Issue #434, opened 2016-12-28
Render empty array item fields when minItems is specified


n1k0:
When `minItems` is specified for an array field, we should initialize the form with as many empty fields as the number specified.

glasserc:
This is closed by #434.

n1k0:
Released in [v0.44.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.44.0).

dropdevcoding:
I recently upgraded from 0.43.0 to 0.45.0 and it seems that this change causes unexpected behaviour in my application:

I have defined an array with minItems: 1 but this array is NOT marked as required. This will cause validation errors when submitting again without changing anything, since the changed default behaviour produces invalid state.

Could you please add an additional check respecting the required property before applying the newly introduced default behaviour?

n1k0:
In my own understanding of the JSON schema spec, `required` and `minItems` are not interdependent. You can have this schema:

```js
{
  type: "object",
  properties: {
    foo: {
      type: "array",
      items: {type: "number"},
      minLength: 1
    }
  }
}
```

And both `{}` and `{foo: [42]}` would be valid payloads against that schema, while `{foo: []}` would not.

spacebaboon:
I'm having a problem with this change too. We're using an array with `"ui:widget": "checkboxes"` to render a checkbox group in a form, and `minItems=1` to validate that the user has selected at least one.

It can be seen easily in the playground, under Arrays. Add `"minItems": 1` to the `multipleChoicesList` section, and then delete the entry for `multipleChoicesList` in the formData section. When a form is first rendered, we have no formData, so this is the initial state, not an empty array.

Now click one of the checkboxes in the live demo and see what happens to the formData. A null entry is added after the selected values. Make it `"minItems": 2` and 2 null entries are added.

I thought that I could set a default for the field, by adding `default: []` in the schema, but this changes the nulls to empty arrays. 

It seems like minItems is serving two different purposes here. And perhaps default too. Perhaps `default` inside `items` could define the item default, and `default` outside could be the default for the array itself?

So my only option, unless I'm missing something, is to programmatically initialise the formData to be an empty array where I find an array type in the schema.

spacebaboon:
I think I'll post my comment above as a new issue, it looks like a bug to me, and this issue has already been closed.

PR #484, opened 2017-02-22
#434 - Render empty array item fields when minItems is specified

Reggino:

Trying to get my feet wet with this library and doing something useful on the way :smile: 

Implemented #434

BTW thanks for this awesome stuff!

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -139,7 +139,6 @@ function computeDefaults(schema, parentDefaults, definitions={}) {
       if (schema.minItems) {
         return new Array(schema.minItems).fill(computeDefaults(schema.items, defaults, definitions));
       }
-      return defaults || [];
   }
   return defaults;
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -123,15 +123,23 @@ function computeDefaults(schema, parentDefaults, definitions={}) {
   if (typeof(defaults) === "undefined") {
     defaults = schema.default;
   }
-  // We need to recur for object schema inner default values.
-  if (schema.type === "object") {
-    return Object.keys(schema.properties).reduce((acc, key) => {
-      // Compute the defaults for this node, with the parent defaults we might
-      // have from a previous run: defaults[key].
-      acc[key] = computeDefaults(
-        schema.properties[key], (defaults || {})[key], definitions);
-      return acc;
-    }, {});
+
+  switch (schema.type) {
+    // We need to recur for object schema inner default values.
+    case "object":
+      return Object.keys(schema.properties).reduce((acc, key) => {
+        // Compute the defaults for this node, with the parent defaults we might
+        // have from a previous run: defaults[key].
+        acc[key] = computeDefaults(
+          schema.properties[key], (defaults || {})[key], definitions);
+        return acc;
+      }, {});
+
+    case "array":
+      if (schema.minItems) {
+        return new Array(schema.minItems).fill(schema.items.default);
+      }
+      return defaults || [];
   }
   return defaults;
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -137,7 +137,7 @@ function computeDefaults(schema, parentDefaults, definitions={}) {
 
     case "array":
       if (schema.minItems) {
-        return new Array(schema.minItems).fill(schema.items.default);
+        return new Array(schema.minItems).fill(computeDefaults(schema.items, defaults, definitions));
       }
       return defaults || [];
   }
--- a/test/ArrayField_test.js
+++ b/test/ArrayField_test.js
@@ -295,6 +295,25 @@ describe("ArrayField", () => {
       expect(inputs[2].id).eql("root_foo_1_bar");
       expect(inputs[3].id).eql("root_foo_1_baz");
     });
+
+    it("should render enough inputs to match minItems in schema when no formData is set", () => {
+      const complexSchema = {
+        type: "object",
+        properties: {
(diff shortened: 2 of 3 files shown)


src/utils.js
```diff
@@ -123,15 +123,23 @@ function computeDefaults(schema, parentDefaults, definitions={}) {
...
+        // Compute the defaults for this node, with the parent defaults we might
+        // have from a previous run: defaults[key].
+        acc[key] = computeDefaults(
+          schema.properties[key], (defaults || {})[key], definitions);
+        return acc;
+      }, {});
+
+    case "array":
+      if (schema.minItems) {
+        return new Array(schema.minItems).fill(schema.items.default);
```
glasserc: Shouldn't this involve a recursive call to `computeDefaults` ? For example, what if the items are an object or have `$ref` ?

test/ArrayField_test.js
```diff
@@ -295,6 +295,25 @@ describe("ArrayField", () => {
...
+            items: {
+              type: "string"
+            }
+          }
+        }
+      };
+      const {node} = createFormComponent({schema: complexSchema, formData: { }});
+
+      const inputs = node.querySelectorAll("input[type=text]");
+      expect(inputs.length).eql(4);
```
glasserc: Should we test that the right default value is chosen? I see that there's no other tests for the "default" behavior for ArrayField, so this might be excessive.

Should we test that merging `formData` with `defaults` works correctly?

src/utils.js
n1k0: Indeed.

Reggino: Fixed

test/ArrayField_test.js
Reggino: IMHO `formData` and `defaults` can really be merged in case of an array: it is either set or not set. When set, the form should honor the given value. When not set it may compute a sane default. 

I've added a test for this behaviour

Reggino:
I've added an example to the playground.

test/ArrayField_test.js
```diff
@@ -296,23 +296,39 @@ describe("ArrayField", () => {
...
-      expect(inputs.length).eql(4);
+      let form = createFormComponent({schema: complexSchema, formData: { }});
+      let inputs = form.node.querySelectorAll("input[type=text]");
+      expect(inputs[0].value).eql("Default name");
+      expect(inputs[1].value).eql("Default name");
+
+      // whenever a value is set, honor it!
+      form = createFormComponent({schema: complexSchema, formData: { foo: [] }});
+      inputs = form.node.querySelectorAll("input[type=text]");
+      expect(inputs.length).eql(0);
```
n1k0: It would be better to distinguish between the two sets of assertions here, so we're better informed when one of these fails.

This can usually be achieved by either mounting a single component per `it` test.

test/ArrayField_test.js
```diff
@@ -295,6 +295,41 @@ describe("ArrayField", () => {
...
+      };
+      let form = createFormComponent({schema: complexSchema, formData: { }});
+      let inputs = form.node.querySelectorAll("input[type=text]");
+      expect(inputs[0].value).eql("Default name");
+      expect(inputs[1].value).eql("Default name");
+
+      // whenever a value is set, honor it!
+      form = createFormComponent({schema: complexSchema, formData: {foo: []}});
+      inputs = form.node.querySelectorAll("input[type=text]");
+      expect(inputs.length).eql(0);
```
n1k0: It would be better to distinguish between the two sets of assertions here, so we're better informed when one of these fails.

This can usually be achieved by either mounting a single component per `it` test.

n1k0:
Also, tested the patch again with the updated playground, works great!

src/utils.js
```diff
@@ -123,15 +123,23 @@ function computeDefaults(schema, parentDefaults, definitions={}) {
...
+        acc[key] = computeDefaults(
+          schema.properties[key], (defaults || {})[key], definitions);
+        return acc;
+      }, {});
+
+    case "array":
+      if (schema.minItems) {
+        return new Array(schema.minItems).fill(computeDefaults(schema.items, defaults, definitions));
+      }
+      return defaults || [];
```
n1k0: Why do we fallback to an empty array here? I understand it may be confusing but we need to distinguish between an empty array and no array value defined at all. I'd suggest we just `return defaults` here.

Reggino: IMHO `computeDefaults` would ideally be the only place where defaults are "made up". I actually tried to remove this line: 
https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/ArrayField.js#L153

But other tests would break. However, it would be more clear when _all_ defaults would originate in `computeDefaults`

test/ArrayField_test.js
Reggino: You're completely right. I've updated the PR

src/utils.js
Reggino: BTW when no value would be defined at all, that would lead to `[]` due to the `defaultProps` in ArrayField.js

n1k0: I agree, we'll have to work to make this more consistent in the future. Meanwhile here, I'm pretty sure we could remove the whole `return defaults || [];` line without breaking anything, as that would match previous behavior, don't you think?

Reggino: True. I've updated the PR

n1k0:
Thanks++

n1k0:

PR #489, opened 2017-02-25
Prevent multiple times rendering when formData change

### LINKED ISSUE #482: Error list disappears when updating a field in error
Sample form https://jsfiddle.net/x79eop03/ 

1. type a single char in the `name` field
2. submit the form
3. type another char in the same field

The top error list disapears as soon as the second character is entered, which shouldn't happen, especially with the `liveValidate` prop not being leveraged. This doesn't happen with a formData state wrapper: https://jsfiddle.net/vj172c7s/1/

knilink:

~~In [react livecycle](https://facebook.github.io/react/docs/react-component.html#updating), "render" function will be called after "componentWillReceiveProps". So, calling "setState" in it will cause multiple times rendering.~~
Also, calling "setState" in "onChange" function may potentially cause multiple times rendering because parent may try to update the form after receiving new formData.

This update fix #482 and the blinking ErrorList.

Since this update will break a lot of test cases, I would like to have some feedback before updating the test cases.

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -26,7 +26,7 @@ export default class Form extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState(this.getStateFromProps(nextProps));
+    this.state = {...this.state, ...this.getStateFromProps(nextProps)};
   }
 
   getStateFromProps(props) {
@@ -82,11 +82,18 @@ export default class Form extends Component {
       const {errors, errorSchema} = this.validate(formData);
       state = {...state, errors, errorSchema};
     }
-    setState(this, state, () => {
+
+    if ("formData" in this.props) {
       if (this.props.onChange) {
-        this.props.onChange(this.state);
+        this.props.onChange(state);
       }
-    });
+    } else {
+      setState(this, state, () => {
+        if (this.props.onChange) {
+          this.props.onChange(this.state);
+        }
+      });
+    }
   };
 
   onBlur = (...args) => {


src/components/Form.js
```diff
@@ -26,7 +26,7 @@ export default class Form extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState(this.getStateFromProps(nextProps));
+    this.state = {...this.state, ...this.getStateFromProps(nextProps)};
```
n1k0: At first I was frightened by this change as it's always been documented as a bad practice. But now I've read more about the lifecycle and the execution order of hooks, I it feels like we don't have many other alternatives. We may want to drop a comment here to explain why we're doing this, though.

src/components/Form.js
```diff
@@ -82,11 +82,18 @@ export default class Form extends Component {
...
-        this.props.onChange(this.state);
+        this.props.onChange(state);
       }
-    });
+    } else {
+      setState(this, state, () => {
+        if (this.props.onChange) {
+          this.props.onChange(this.state);
+        }
+      });
```
n1k0: We'll have to document this behavior very explicitly, and early in the docs, as it may be very surprising to users. Otherwise it makes sense, we can now distinguish between controlled and uncontrolled forms.

src/components/Form.js
elisechant: this could be `this.setState(this.getStateFromProps(nextProps);`

n1k0: Well, that would trigger a render, which is what this patch tries preventing.

knilink: I just did some investigation an found that "setSate" in "componentWillReceiveProps" will actually not trigger render.
And the [react document](https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops) said 
>  you may compare this.props and nextProps and perform state transitions using this.setState() in this method.
So I think the root cause was just the setState in onChange.
I will change this code back.

n1k0: Ah, nice!

epicfaace:
Closing due to inactivity.

PR #490, opened 2017-02-25
Centralized handling shouldComponentUpdate into SchemaField

knilink:

When one property is updated, the rest of the properties will be unnecessarily rendered.
This will cause a significant performance issue in a large object schema such as the demo of 'Large' in the playground.

### Checklist

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -1,4 +1,4 @@
-import React, {PropTypes, Component} from "react";
+import React, {PropTypes} from "react";
 
 import {
   isMultiSelect,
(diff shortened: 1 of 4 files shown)


src/components/fields/SchemaField.js
```diff
@@ -128,38 +129,50 @@ DefaultTemplate.defaultProps = {
...
-    return <div/>;
+class SchemaField extends Component {
+  shouldComponentUpdate(nextProps, nextState) {
+    let _props = {...this.props};
+    let _nextProps = {...nextProps};
+    // if schema and formData are the equal
+    // idSchema and errorSchema will be equal
+    delete _props.idSchema;
+    delete _nextProps.idSchema;
+    return !deepEquals(_props, _nextProps);
```
n1k0: I'm confused with this code:

1. the implementation doesn't reflect what the comment says; we're not dropping `errorSchema`;
1. the comment suggests we should only compare changes for `schema` and `formData`, which is not what we're doing;
1. I suspect we should take care of `uiSchema` changes as well.

I'm trying to summarize the points above in a sample implementation below:

```jsx
shouldComponentUpdate(nextProps) {
  return !deepEquals(this.props.schema, nextProps.schema) ||
         !deepEquals(this.props.uiSchema, nextProps.uiSchema) ||
         !deepEquals(this.props.formData, nextProps.formData);
}
```

We may obviously want to ensure we're not missing other necessary checks, but I'm finding this way much more explicit.

src/components/fields/SchemaField.js
```diff
@@ -128,38 +129,50 @@ DefaultTemplate.defaultProps = {
...
-  const FieldComponent = getFieldComponent(schema, uiSchema, fields);
-  const {DescriptionField} = fields;
-  const disabled = Boolean(props.disabled || uiSchema["ui:disabled"]);
-  const readonly = Boolean(props.readonly || uiSchema["ui:readonly"]);
-  const autofocus = Boolean(props.autofocus || uiSchema["ui:autofocus"]);
-
-  if (Object.keys(schema).length === 0) {
-    // See #312: Ensure compatibility with old versions of React.
-    return <div/>;
+class SchemaField extends Component {
```
n1k0: It's really sad there's no way to have the equivalent of `shouldComponentUpdate` for stateless components. I could totally see a decorator of some sort.

test/performance_test.js
```diff
@@ -49,49 +48,7 @@ describe("Rendering performance optimizations", () => {
...
-      };
-
-      const {comp} = createComponent(ArrayField, props);
-      sandbox.stub(comp, "render").returns(<div/>);
-
-      setProps(comp, {...props, formData: ["a", "b"]});
-
-      sinon.assert.notCalled(comp.render);
-    });
-  });
```
n1k0: Why has this test been removed?

knilink: The "shouldComponnetUpdate" of "ObjectField" and "ArrayField" were guaranteed by "SchemaField". So, I thought those two function were no longer necessary and the related tests were removed.

src/components/fields/SchemaField.js
knilink: I think React still need an instance to preserve props in order to have both "this.props" and "nextProps" to compare. [recompose](https://github.com/acdlite/recompose) is one library I know to do such a thing. It will turn functions into classes.

src/components/fields/SchemaField.js
knilink: The comment suggests that "schema" and "formData" didn't change => "idSchema" and "errorSchema" didn't change. So, comparing "schema" and "formData" is enough to tell wheter "idSchema" and "errorSchema" have change or not.
But, not comparing errorSchema will fail some test. I haven't figure out the reason yet. And, the performance with comparing errorSchema looks acceptable. 
I will update the comment.

src/components/fields/SchemaField.js
```diff
@@ -211,6 +212,21 @@ function SchemaField(props) {
...
+  shouldComponentUpdate(nextProps, nextState) {
+    // if schemas are equal idSchemas will be equal as well,
+    // so it is not necessary to compare
+    return !deepEquals(
+      {...this.props, idSchema: undefined},
+      {...nextProps,  idSchema: undefined}
+    );
+  }
+
+  render(){
```
n1k0: nit: space before opening bracket

test/performance_test.js
```diff
@@ -47,51 +46,61 @@ describe("Rendering performance optimizations", () => {
...
+        formData: ["const", "var0"]
+      });
 
-      setProps(comp, {...props, formData: ["a", "b"]});
+      const fields = scryRenderedComponentsWithType(comp, SchemaField)
+              .reduce( (fields, fieldComp) => {
+                sandbox.stub(fieldComp, "render").returns(<div/>);
+                fields[fieldComp.props.idSchema.$id] = fieldComp;
+                return fields;
+              });
```
n1k0: nit: let's apply a consistent coding style:

```jsx
const fields = scryRenderedComponentsWithType(comp, SchemaField)
  .reduce((fields, fieldComp) => {
    sandbox.stub(fieldComp, "render").returns(<div/>);
    fields[fieldComp.props.idSchema.$id] = fieldComp;
    return fields;
  });
```

n1k0:
Thank you :+1:

n1k0:

PR #495, opened 2017-02-27
lint playground too

Reggino:

My editor complains when opening playground code

### Checklist


n1k0:
Thanks!

Issue #435, opened 2016-12-29
Leverage the description schema prop for boolean checkboxes


n1k0:
Checkbox fields don't display the description at all; we could at least use a tooltip.

Reggino:
@n1k0 Would adding Jquery + Bootstrap JavaScript as (optional?) dependencies be an option to get something like [http://getbootstrap.com/javascript/#tooltips](this)? Or would you prefer a custom type of tooltip? 

Or would it be an option to show the description above the checkbox, like it is with regular text inputs?

n1k0:
Just a `title` attribute would work I think.

revolunet:
does https://github.com/mozilla-services/react-jsonschema-form/pull/498 fixes this ?

n1k0:
Yes!

PR #498, opened 2017-02-28
Show description for boolean fields

Reggino:

A fix for #435 . I chose to present the description in a similar way like the other fields, with an extra text above the input. The suggested `title` looked a bit "strange" since it was the only input with that kind of behavior and it would probably have issues in a mobile view. 

### Checklist

### DIFF
--- a/src/components/widgets/CheckboxWidget.js
+++ b/src/components/widgets/CheckboxWidget.js
@@ -1,5 +1,5 @@
 import React, {PropTypes} from "react";
-
+import DescriptionField from "../fields/DescriptionField.js";
 
 function CheckboxWidget({
   schema,
@@ -13,6 +13,10 @@ function CheckboxWidget({
 }) {
   return (
     <div className={`checkbox ${disabled ? "disabled" : ""}`}>
+      { schema.description
+        ? <DescriptionField description={ schema.description }/>
+        : null
+      }
       <label>
         <input type="checkbox"
           id={id}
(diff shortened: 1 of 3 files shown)


src/components/widgets/CheckboxWidget.js
```diff
@@ -13,13 +13,18 @@ function CheckboxWidget({
...
+        : null
+      }
       <label>
         <input type="checkbox"
           id={id}
           checked={typeof value === "undefined" ? false : value}
           required={required}
           disabled={disabled}
           autoFocus={autofocus}
+          title={schema.description}
```
n1k0: The title attribute should be set on the `<label>` element, so the tooltip is shown when hovering over the whole widget.

src/components/widgets/CheckboxWidget.js
```diff
@@ -13,13 +13,18 @@ function CheckboxWidget({
 }) {
   return (
     <div className={`checkbox ${disabled ? "disabled" : ""}`}>
+      { schema.description
+        ? <DescriptionField description={ schema.description }/>
+        : null
```
n1k0: Nit: this could be written as

```jsx
{ schema.description && <DescriptionField description={ schema.description }/> }
```

src/components/widgets/CheckboxWidget.js
Reggino: I removed the title-attribute completely, since it is redundant with the `DescriptionField` above the input.

Reggino:
Thanks for the feedback. Should be good to go now!

n1k0:
Thanks!

n1k0:

PR #199, opened 2016-05-10
Fixes #195: Preserve array field error index.

n1k0:
Refs #195.

r=? @leplatrem @glasserc /cc @nuclearspike

PR #503, opened 2017-03-03
Fix #221 Manually reorder array erros insteading of using validate when live validation is off

### LINKED ISSUE #221: Adding an item to an ArrayField triggers form validation even if liveValidate is false

knilink:
No validating but updating errorSchema base on array operation when live validation is off.

{validate: true/false} in the onChange function has been replaced by a updated errorSchema when live validation was setted to fail and array was reorder or related item was deleted. This can prevent validating the whole form when justing updating the array.

Limitation: Unexpected behaviours may occur if user forget to pass the update function to parent in their custom field.

This patch fix #221.

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -8,7 +8,7 @@ import {
   setState,
   getDefaultRegistry,
 } from "../utils";
-import validateFormData from "../validate";
+import validateFormData, {toErrorList} from "../validate";
 
 
 export default class Form extends Component {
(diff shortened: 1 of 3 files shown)


src/components/Form.js
```diff
@@ -75,12 +75,18 @@ export default class Form extends Component {
     return null;
   }
 
-  onChange = (formData, options={validate: false}) => {
-    const mustValidate = !this.props.noValidate && (this.props.liveValidate || options.validate);
+  onChange = (formData, newErrorSchema) => {
```
elisechant: Does parameter name `newErrorSchema` make sense here? does `toErrorList(arg)` not evaluate to a "newErrorSchema"?
Also, is this a bit scary to change `onChange`'s parameter API?

glasserc:
I don't understand this either -- the PR description says it's adding a function that lets you set error schema, but instead it looks like it's just making `onChange` always take `errorSchema`. That seems a bit strange to me because most changes don't have access to the error schema or need to update it.

src/components/Form.js
knilink: Base on my understanding so far, the "option.validate" was just to notify Form component to reorder or remove errorSchema in the array as well by using "validation" when the array was updated. If it wasn't the exiting "option" parameter that can be replaced with, I would probably not propose this solution for the moment.

knilink:
@glasserc I forgot to update the description, my bad. I found out latter that returning an updated errorSchema directly was the same so I change the solution.

coder-guy22296:
@n1k0 I am having an issue where removing a field from the field array triggers a form validation when live validate is off. I believe this is a fix for the issue, could you take a look and try to merge this one. The fix for  #195 (Pull request #199 ) introduced this bug.

lingxiaoyang:
+1

glasserc:
Sorry for taking so long to get back to this. I understand the purpose of this a lot better. I've taken the liberty of rebasing this and merging it to master. Thanks very much for your work!

PR #506, opened 2017-03-06
[WIP] feat(errorList): *naive* add links to fields in error in the Error section

elisechant:

https://github.com/mozilla-services/react-jsonschema-form/issues/86#issuecomment-284218618

*This Pull Request is WIP.*

This is a naive implementation of the Issue. I assume this is the sort of fix you meant by [it should be trivial](https://github.com/mozilla-services/react-jsonschema-form/issues/86#issue-143018959)?

Can I get some feedback on the following before I continue:
* I believe that scroll to links should work without hash anchors for the case where people are using Hash Routing
* Is this along the lines of the implementation that you had in mind?
* Should this behaviour be configurable? 

### Checklist

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -91,7 +91,7 @@ function DefaultTemplate(props) {
   }
 
   return (
-    <div className={classNames}>
+    <div id={id} className={classNames}>
       {displayLabel ? <Label label={label} required={required} id={id}/> : null}
       {displayLabel && description ? description : null}
       {children}
--- a/src/components/ErrorList.js
+++ b/src/components/ErrorList.js
@@ -1,7 +1,7 @@
 import React from "react";
 
 
-export default function ErrorList({errors}) {
+export default function ErrorList({errors, idSchema}) {
   return (
     <div className="panel panel-danger errors">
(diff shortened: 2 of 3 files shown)


src/components/fields/SchemaField.js
```diff
@@ -91,7 +91,7 @@ function DefaultTemplate(props) {
   }
 
   return (
-    <div className={classNames}>
+    <div id={id} className={classNames}>
```
n1k0: That would duplicate the id, while they should be unique to the document; we could suffix the target wrapping component id.

elisechant: yes. it would. this is a just a naive implementation so we can unpack what needs to be done.

elisechant: * is this the type of solution you had in mind? 
* should it be configurable from api?

n1k0:
> is this the type of solution you had in mind?

Yes! It's important to differentiate between the input id from the field one, as you'll want to scroll at a position where the whole field in error is rendered I think.

> should it be configurable from api?

Please no, let's be opinionated here. I think the docs are long and boring enough already.

epicfaace:
Closing due to inactivity.

PR #516, opened 2017-03-16
add css example

revolunet:

Another customisation example

### Checklist


n1k0:
Good idea ;)

PR #525, opened 2017-03-23
Fix #412: Readonly widgets behavior consistency

### LINKED ISSUE #412: Readonly on radio and checkboxes

n1k0:
Supersedes #470.


n1k0:
Released in [v0.45.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.45.0).

PR #537, opened 2017-04-05
Fix #536: Pass uiSchema to custom array templates.

### LINKED ISSUE #536: Multiple Array Types
I have two different kinds of arrays in my form which should be rendered differently. However, I cannot multiplex the ArrayField template because you do not pass the uiSchema as props to it (with which I would pass a 'ui:options' describing the array type).

Would it be possible to pass the uiSchema (or maybe there is another way to handle multiple custom array fields I'm not seeing).

### Version
v0.45.0

Thanks for the great library!

n1k0:
Fix #536.


n1k0:
Released in [v0.46.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.46.0).

PR #539, opened 2017-04-06
removing content from text field result in empty string instead of undefined

quentin-sommer:
As mentioned here before https://github.com/mozilla-services/react-jsonschema-form/pull/476#issuecomment-292175013
The new way of handling empty text fields (setting them to undefined) is undesirable as most of the time we want to keep an empty value.
If we want to delete a field we can do so after the form validation by for emptiness and delete accordingly.

Behavior change:
- sets text field to `""` instead of undefined when you remove everything inside it 

### Checklist

Please take in consideration that I'm fairly new to the repo and might have underestimated the change, I'd like your input if that's the case!

Quentin

### DIFF
--- a/README.md
+++ b/README.md
@@ -1294,7 +1294,7 @@ render((
 
 ### The case of empty strings
 
-When a text input is empty, the field in form data is set to `undefined`. String fields that use `enum` and a `select` widget work similarly and will have an empty option at the top of the options list that when selected will result in the field being `undefined`.
+When a text input is empty, the field in form data is set to empty string. String fields that use `enum` and a `select` widget will have an empty option at the top of the options list that when selected will result in the field being `undefined`.
 
 One consequence of this is that if you have an empty string in your `enum` array, selecting that option in the `select` input will cause the field to be set to `undefined`, not an empty string.
 
--- a/README.md
+++ b/README.md
@@ -1294,9 +1294,11 @@ render((
 
 ### The case of empty strings
 
-When a text input is empty, the field in form data is set to empty string. String fields that use `enum` and a `select` widget will have an empty option at the top of the options list that when selected will result in the field being `undefined`.
+When a text input is empty, the field in form data is set to `undefined`. String fields that use `enum` and a `select` widget will have an empty option at the top of the options list that when selected will result in the field being `undefined`.
 
 One consequence of this is that if you have an empty string in your `enum` array, selecting that option in the `select` input will cause the field to be set to `undefined`, not an empty string.
+
+If you want to have the field set to a default value when empty you can provide a `ui:emptyValue` field in the `uiSchema` object.
 
 ## Styling your forms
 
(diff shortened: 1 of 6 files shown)


n1k0:
I'm not sure we want to revert the old behavior. We spent weeks (months) aligning on this and it was hard to reach consensus. Please ensure reading comments in related issues, there are a lot of constraints which may not be obvious at first glance. 

Could you please describe a little more your specific use case so we can ensure we're not missing any piece of valuable information here?

Calling @crumblix @olzraiti @revolunet @ @Reggino @knilink @vinhlh @mplis-jetsetter @glasserc @MoOx and other people interested in this project to provide feedback.

quentin-sommer:
Hi,

When implementing data edit forms and dealing with strings I think we have 2 use cases
- string is used to hold a link information (such as an id to another resource or an enum value).
In this case, it's easier to mark the emptiness of the field as `null` or `undefined` as it's logical that it's either here or non-existant.

- string is used to hold an arbitrary information (such as a description o r a text). In this case, it is better to hold a sane default value for the field (it could be `""` or something else depending on the context).

One way to make is this is to leverage the `default` value and populate the field with it every time it becomes empty. If we want to represent it by `undefined` we just have to leave `default` blank. However, if we want to keep a default value at all times we provide it to `default` and we'll be assured that we will still be getting a good default in any case.

The behavior I'd like to see is the following: 
* when no `default` field is set and field is empty, output `undefined`
* when a `default` field is set and field is empty, output `default` value

mplis:
I agree that there are two different, valid use cases for an empty string field. I've come across both since I started using this library. I like giving the user some control to define the behavior. 

`default` does feel like a natural place for this, though I worry that that would be overloading the field. We'd be making the assumption that the value you want to populate the field with before the user interacts with it (the current use of `default`) is *always* the same value you want in the field when the user clears it, which I don't think is a safe assumption.

Reggino:
I agree with @quentin-sommer the current string-array handling feels weird and needs some attention :-)

I agree with @mplis it would simply be annoying whenever you would try to clear an input and it is instantly reset to some default value. 

Can't we just add new option to the uischema? Something like `ui:allowEmptyString`?

whollacsek:
I also agree with @mplis, how about something like `ui:emptyValue`, which when not set means `undefined`, so the developer can explicitly set it to `""` (empty string) or `0` (zero). This should keep the current behavior as is, and won't break anything from the issues @n1k0 is referring to.

quentin-sommer:
I've implemented the change on the pull request so you can try it out here 
https://quentin-sommer.github.io/react-jsonschema-form

the `firstName` field in the `simple` example as an `ui:emptyValue` of `""` 
and the `listOfStrings` field of the `array` example as an `ui:emptyValue` of `""`

n1k0:
This is really good, and it solves all the constraints I had in mind. Could you please add some documentation in the README so we can land this patch?

n1k0:
Released in [v0.46.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.46.0).

Issue #465, opened 2017-02-09
Support optional objects


mfulton26:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Add support for not specifying an optional object property.

### Steps to Reproduce

1. Define a schema with a property of type object which is not required.
2. Render the schema on a react-jsonschema-form and attempt to not specify the optional object property.

https://jsfiddle.net/f2y3fq7L/19/

#### Expected behavior

Have some sort of checkbox or something to enable/disable specifying the optional object property.

#### Actual behavior

There is no apparent way to not specify the optional object property.

### Version

0.42.0

n1k0:
Yeah, I've worked on a patch to make [fields collapsible in the past](https://github.com/mozilla-services/react-jsonschema-form/pull/271) but it was discarded in favor of theming; maybe we'd want to resurect it at some point, as this use case still seems valid to me.

elisechant:
@mfulton26 would the example from the docs work for you here? https://jsfiddle.net/69z2wepo/68259/

mfulton26:
I suppose although it does alter the schema and it seems like a lot of overhead if I have many optional objects.

dropdevcoding:
In fact the form actually also validates wrong in this case which is very annoying. When defining an optional object inside the schema which has required fields the required fields always get validated although the schema would allow the whole object not to be there

literarymachine:
I just had the same problem described here and in #331; a workaround seems to be to specify the property as type `array` with the `object` as its `items` and `maxItems` of 1: https://jsfiddle.net/z8b5s7t2/

epicfaace:
Closing this in favor of #675

PR #542, opened 2017-04-10
Corrected defaulting & non-required empty property handling

dropdevcoding:

In the current version imho the defaulting is broken:

* When having a non required array property minItems will be set though. Behavior should be: When required, minItems will be applied, when not required and no defaults are set, minItems should not be applied

The value handling imho also is broken for non required fields:

* When having defined an object which is not required but has required properties, the following should apply:

  * When one of the object's property is set the invariant rules shall apply
  * When no property of the object is set the whole object should be undefined since it is not required because otherwise an empty object is left in formData and the invariant rules will be applied which causes an incorrect validation error  

* When deleting the last item of a non required array, the following should apply:
  * When the last element of a non required array is removed, the whole array should be set to undefined, otherwise the array is left empty in formData and eventual invariants like minItems would be applied and cause validation errors although they should not in this case

#465 

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -12,6 +12,7 @@ import {
   retrieveSchema,
   toIdSchema,
   getDefaultRegistry,
+  cleanUpNonRequiredArray,
 } from "../../utils";
 
 function ArrayFieldTitle({ TitleField, idSchema, title, required }) {
@@ -220,7 +221,6 @@ class ArrayField extends Component {
   };
 
   onDropIndexClick = index => {
-    console.log("DROP INDEX");
     return event => {
       if (event) {
         event.preventDefault();
@@ -231,11 +231,7 @@ class ArrayField extends Component {
       // if field isn't required and doesn't contain any entries
       // remove the whole property from formData to guarantee correct validation
       if (!required) {
-        formData = formData.filter(item => {
-          return item !== null && item !== undefined;
-        });
-
-        formData = formData.length ? formData : undefined;
+        formData = cleanUpNonRequiredArray(formData);
       }
       // refs #195: revalidate to ensure properly reindexing errors
 
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -202,26 +202,45 @@ class ArrayField extends Component {
 
   onAddClick = event => {
     event.preventDefault();
-    const { schema, registry, formData } = this.props;
+    const { schema, registry, required } = this.props;
     const { definitions } = registry;
+    let formData = this.props.formData;
     let itemSchema = schema.items;
     if (isFixedItems(schema) && allowAdditionalItems(schema)) {
       itemSchema = schema.additionalItems;
     }
+
     this.props.onChange(
-      [...formData, getDefaultFormState(itemSchema, undefined, definitions)],
+      [
+        ...formData,
+        getDefaultFormState(itemSchema, undefined, definitions, required),
+      ],
       { validate: false }
     );
   };
 
   onDropIndexClick = index => {
+    console.log("DROP INDEX");
     return event => {
       if (event) {
         event.preventDefault();
       }
-      const { formData, onChange } = this.props;
+      const { onChange, required } = this.props;
+      let formData = this.props.formData;
+
+      // if field isn't required and doesn't contain any entries
+      // remove the whole property from formData to guarantee correct validation
+      if (!required) {
+        formData = formData.filter(item => {
+          return item !== null && item !== undefined;
+        });
+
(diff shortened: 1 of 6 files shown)


src/components/fields/ArrayField.js
```diff
@@ -202,26 +202,45 @@ class ArrayField extends Component {
...
+        formData = formData.filter(item => {
+          return item !== null && item !== undefined;
+        });
+
+        formData = formData.length ? formData : undefined;
+      }
       // refs #195: revalidate to ensure properly reindexing errors
-      onChange(formData.filter((_, i) => i !== index), { validate: true });
+
+      formData = formData ? formData.filter((_, i) => i !== index) : formData;
```
glasserc: Shouldn't we be doing this `filter` before the `if (!required)` check?

src/components/fields/ArrayField.js
```diff
@@ -250,12 +265,13 @@ class ArrayField extends Component {
   onChangeForIndex = index => {
     return value => {
       const { formData, onChange } = this.props;
-      const newFormData = formData.map((item, i) => {
+      let newFormData = formData.map((item, i) => {
```
glasserc: Nit: Does this still need to be `let`?

src/utils.js
```diff
@@ -133,32 +138,76 @@ function computeDefaults(schema, parentDefaults, definitions = {}) {
...
         },
         {}
       );
 
     case "array":
-      if (schema.minItems) {
+      if (required && schema.minItems) {
         return new Array(schema.minItems).fill(
-          computeDefaults(schema.items, defaults, definitions)
+          computeDefaults(schema.items, defaults, definitions, required)
```
glasserc: Should `schema.required` be relevant here?

src/utils.js
```diff
@@ -133,32 +138,76 @@ function computeDefaults(schema, parentDefaults, definitions = {}) {
...
+  if (!cleanValues.length) {
+    values = undefined;
+  }
+
+  return values;
+}
+
+export function cleanUpNonRequiredArray(values) {
+  values = values.filter(item => {
+    return item !== null && item !== undefined;
```
glasserc: Why do we need to filter out undefined values here?

src/utils.js
```diff
@@ -197,6 +246,8 @@ export function isObject(thing) {
 
 export function mergeObjects(obj1, obj2, concatArrays = false) {
   // Recursively merge deeply nested objects.
+  obj1 = obj1 || {};
+  obj2 = obj2 || {};
```
glasserc: What is this change about?

dropdevcoding:
Okay, I'll sum up the reasons for the changes:
Having given a schema like

```
{
  "type": "object"
  "properties": {
   "nonRequiredOuterProp": {
     "type": "object",
     "properties": {
       "requiredInnerProp1": {
         "type": "string"
       },
       "requiredInnerProp2": {
         "type":"array",
         "items": {
           "type": "string"
         }
       }
     },
     "required": [
      "requiredInnerProp1",
      "requiredInnerProp2"
     ]
   }
  }
}
```

Which means nonRequiredOuterProp may be undefined or if set must match the given criteria,
the form doesn't allow the user to restore the undefined state.

As soon as I fill in ```requiredInnerProp1``` and ```requiredInnerProp2```  the object turns from ```undefined``` to
```
{
   "nonRequiredOuterProp": {
     "requiredInnerProp1": "myInputForRequiredInnerProp1",
     "requiredInnerProp2": "myInputForRequiredInnerProp2"
}
```

which is correct, but when I remove the whole input the objects falls back to ```{}``` which is not valid, since ```requiredInnerProp1```and ```requiredInnerProp2```are not defined but required for ```nonRequiredProp```where ```nonRequiredProp``` is not required at all.

So the expected behaviour is:
When an object in *NOT* required, it *collapses* to undefined (initial state) if none of its properties is defined. Otherwise validation would fail (possibly ```jsonschema```does not handle this case right. I also introduced ```ajv```since I use it for backend validation).

As soon as the user fills out a property the empty object remains and the user hasn't got any chance to transfer it to initial state.

Same behaviour applies to non required arrays. 

I hope this will help to clarify. If you got further questions, please don't hesitate to ask :-)

eggyal:
If I've understood this issue correctly, then I think it's also fixed by #557?  In particular 9d0b741 and 132bc60?

glasserc:
Hi @dropdevcoding, thanks for explaining the reasoning behind the PR. What do you think about my objections to the behavior of the changes, and my comments on the code review?

epicfaace:
Closing this PR due to inactivity. If you'd still like to get this PR in, please make a new PR rebased to the latest code. Thanks!

Issue #446, opened 2017-01-19
On slow CPUs safeRenderCompletion needs to be set to make sure RJSF waits for setState before calling onChange


matiaslarsson:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I have custom widgets calling onChange only when the user is typing enter or unfocusing the text field to keep RJSF from updating the model for every key stroke.

Depending on CPU speed, submitting the form can result in the previous field value being submitted, probably due to the [utils.js setState function](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/utils.js#L463) only using the React setState callback functionality when `safeRenderCompletion` is true.

When the `safeRenderCompletion` prop is set on the `<Form>` component everything works as expected.

If I'm not doing something totally crazy here and this is expected behaviour maybe we should add something about `safeRenderCompletion` and this issue to the docs?

### Steps to Reproduce

1. Go to https://jsfiddle.net/matiasnu/gad1xm1L/
2. Turn on CPU throttling in Chrome (I'm using 50x, but the factor will most likely vary depending on your CPU)
3. Try updating the text field and clicking submit while keeping the text field focused. When not using `safeRenderCompletion` the submission will very often be the previous model state.
4. Turning off CPU throttling also makes the problem go away so it seems to be a timing issue.

#### Expected behavior

The submission should be the same as the model.

#### Actual behavior

On slow computers or CPU throttling activated the submission is often, but not always, the previous model value.

### Version

react-jsonschema-form@0.41.2

n1k0:
> maybe we should add something about safeRenderCompletion and this issue to the docs?

Sure, would you want to update the docs about it in a PR?

n1k0:
Could you please test again now that #480 has landed, which seems to have fixed #339 which looks very similar to your issue? Thanks.

matiaslarsson:
Sorry for the late reply! #480 did not fix the issue unfortunately. Tried both RJSF 0.43 and 0.45. Will create a PR with updated docs warning about this issue.

matiaslarsson:
@n1k0 docs PR submitted.

glasserc:
[...]

PR #544, opened 2017-04-12
Add a note about needing safeRenderCompletion and slow CPU:s

matiaslarsson:
On slow CPUs safeRenderCompletion needs to be set to make sure RJSF waits for setState before calling onChange.

See the discussion in #446. 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -895,7 +895,9 @@ The following props are passed to custom widget components:
 - `options`: A map of options passed as a prop to the component (see [Custom widget options](#custom-widget-options)).
 - `formContext`: The `formContext` object that you passed to Form.
 
-> Note: Prior to v0.35.0, the `options` prop contained the list of options (`label` and `value`) for `enum` fields. Since v0.35.0, it now exposes this list as the `enumOptions` property within the `options` object.
+> Notes: 
+> * Prior to v0.35.0, the `options` prop contained the list of options (`label` and `value`) for `enum` fields. Since v0.35.0, it now exposes this list as the `enumOptions` property within the `options` object.
+> * If your custom widget only calls `onChange` when the user is blurring the field (maybe to keep RJSF from updating the model for every key stroke) there is an edge case where users on devices with less processing power could potentially submit stale `formData` when focused in an input field while submitting the form due to [RJSF util setState](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/utils.js#L495) not using the React `setState` callback unless we set the `safeRenderCompletion` prop to `true` on the `<Form>`component.
 
 #### Custom component registration
 


README.md
```diff
@@ -895,7 +895,9 @@ The following props are passed to custom widget components:
 - `options`: A map of options passed as a prop to the component (see [Custom widget options](#custom-widget-options)).
 - `formContext`: The `formContext` object that you passed to Form.
 
-> Note: Prior to v0.35.0, the `options` prop contained the list of options (`label` and `value`) for `enum` fields. Since v0.35.0, it now exposes this list as the `enumOptions` property within the `options` object.
+> Notes: 
+> * Prior to v0.35.0, the `options` prop contained the list of options (`label` and `value`) for `enum` fields. Since v0.35.0, it now exposes this list as the `enumOptions` property within the `options` object.
+> * If your custom widget only calls `onChange` when the user is blurring the field (maybe to keep RJSF from updating the model for every key stroke) there is an edge case where users on devices with less processing power could potentially submit stale `formData` when focused in an input field while submitting the form due to [RJSF util setState](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/utils.js#L495) not using the React `setState` callback unless we set the `safeRenderCompletion` prop to `true` on the `<Form>`component.
```
glasserc: Wow, this is one complicated sentence. I'd like to break it up into smaller, more digestible chunks. I understand that there's a lot you want to convey -- there's an edge case, it happens in these circumstances, this is why it happens, and here's the workaround -- but I think each of those could even be a whole sentence. Maybe we should start with the symptom (submitting with stale data?) since that's the thing someone might notice?

I just noticed that we don't actually document `safeRenderCompletion` anywhere, which makes me think it's a hack used for tests. But because it's a prop of the form, rather than any specific widget, I wonder if we should document it somewhere more global instead?

n1k0:
`safeRenderCompletion` is only used for tests to ease writing them, where we need to ensure a component is **fully** rendered before performing assertions on the resulting DOM state. I never documented this option as I could never see any use case in real world webapps, as it slows execution a lot.

matiaslarsson:
The funny thing is, we went for only updating the `formData` when blurring the input since the RJSF + Material-UI combo was so slow when updating on every key stroke. That in turn causes the stale data issue I tried to describe above to appear (and solved by setting `safeRenderCompletion`).

But I believe that's more of a Material-UI problem which hopefully will be solved when they're done with their rewrite so I'm fine with this not getting merged or documented.

numbbbbb:
I waste 3 hours to figure out why my onChange function doesn't give me the changed data. I understand you may have some reason to add this config, but you really shouldn't set it to `false` by default and **SAY NOTHING ABOUT IT IN YOUR README**.

epicfaace:
Closing this because we no longer use safeRenderCompletion.

PR #547, opened 2017-04-15
Make the base input widget customizable.

### LINKED ISSUE #546: Overriding core components styles
Hey,

I'd like to perform minimal changes on the core widgets shipped with this package,
For example, I'd like to extend all components with my some CSS classes:

```javascript
import React, {PropTypes} from "react"
import _BaseInput from 'react-jsonschema-form/lib/components/widgets/BaseInput'

function BaseInput(props) {
    return <_BaseInput {...props} className='foo'/>
}
export default BaseInput
```
This won't work because `BaseInput` overrides the className field,

Do you find it okay to move the [inputProps](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/widgets/BaseInput.js#L23]) at the end of the list so default props could be overriden ?

Also, other components that don't inherit from `BaseInput` don't propagate properties, if you find this approach acceptable I can send some PR that make the core widgets extendable.

Thanks.

asaf:
Make it possible to extends the BaseInput widget and override some of its
props (e.g., "className")

resolves #546

### DIFF
--- a/src/components/widgets/BaseInput.js
+++ b/src/components/widgets/BaseInput.js
@@ -20,12 +20,12 @@ function BaseInput(props) {
   };
   return (
     <input
-      {...inputProps}
       className="form-control"
       readOnly={readonly}
       disabled={disabled}
       autoFocus={autofocus}
       value={value == null ? "" : value}
+      {...inputProps}
       onChange={_onChange}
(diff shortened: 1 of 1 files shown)


glasserc:
Sorry for the delay! Thanks!

n1k0:

PR #549, opened 2017-04-18
add prop-types package and add respective imports b/c it was deprecated

NathanQ:

Utilizing in a React 15.5 project and getting the console warning:

> Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

So, just adding the prop-types package and importing as Proptypes from the prop-types package where Proptypes were previously imported from React.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "mocha": "^2.5.3",
     "prettier": "^1.0.0",
     "prettier-check": "^1.0.0",
+    "prop-types": "^15.5.8",
     "react": "^15.0.0",
     "react-addons-test-utils": "^15.3.2",
     "react-codemirror": "^0.2.3",
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "prettier": "^1.0.0",
     "prettier-check": "^1.0.0",
     "prop-types": "^15.5.8",
-    "react": "^15.0.0",
+    "react": "^15.5.0",
     "react-addons-test-utils": "^15.3.2",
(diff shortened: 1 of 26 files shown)


elxris:
I think its `PropTypes` instead of `Proptypes`, that's why the CI failed.

NathanQ:
Doh! I'll resubmit. Thanks @elxris

elxris:
Prettier has changed a bit, and because CI install always the newest version you should run `npm run cs-format` to fix it

glasserc:
Thanks for your contribution! The code looks good, but I noticed that https://github.com/reactjs/prop-types says that prop-types is compatible with 15.3+. Maybe this would be a good opportunity to bump the `react` version in package.json?

It looks like this PR needs a rebase onto current master. Let me know if you don't have time to take care of it and I'll do it!

NathanQ:
@glasserc no problem and I don't mind. Thanks for the tip.

edi9999:
Can you release a new version of this package now that this MR was accepted ?

Thanks in advance !

n1k0:

package.json
```diff
@@ -74,7 +74,8 @@
     "mocha": "^2.5.3",
     "prettier": "^1.0.0",
     "prettier-check": "^1.0.0",
-    "react": "^15.0.0",
+    "prop-types": "^15.5.8",
```
nelix: This should be in peer or dependencies rather than devDependencies

PR #554, opened 2017-04-21
Handle references to deep schema definitions

### LINKED ISSUE #553: References to definitions deeper than immediate children of `#/definitions` are not supported
### Prerequisites

### Description

References to definitions deeper than immediate children of `#/definitions` are not supported.

### Steps to Reproduce

1. Define a schema that references definitions deeper than immediate children of `#/definitions`:
    ```json
    {
      "definitions": {
        "testdef": {
          "type": "object",
          "properties": {
            "bar": {
              "type": "string"
            }
          }
        }
      },
      "type": "object",
      "properties": {
        "foo": {
          "$ref": "#/definitions/testdef/properties/bar"
        }
      }
    }
    ```
2. Attempt to create a form using that schema ([shared playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJkZWZpbml0aW9ucyI6eyJ0ZXN0ZGVmIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7ImJhciI6eyJ0eXBlIjoic3RyaW5nIn19fX0sInR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7ImZvbyI6eyIkcmVmIjoiIy9kZWZpbml0aW9ucy90ZXN0ZGVmL3Byb3BlcnRpZXMvYmFyIn19fSwidWlTY2hlbWEiOnt9fQ==)).

#### Expected behavior

Form to be rendered.

#### Actual behavior

No form rendered.  Console shows:

> **Error: Could not find a definition for #/definitions/testdef/properties/bar.**

### Version

0.46.0

eggyal:

Fixes #553

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -314,10 +314,21 @@ export function optionsList(schema) {
 
 function findSchemaDefinition($ref, definitions = {}) {
   // Extract and use the referenced definition if we have it.
-  const match = /#\/definitions\/(.*)$/.exec($ref);
-  if (match && match[1] && definitions.hasOwnProperty(match[1])) {
-    return definitions[match[1]];
+  const match = /^#\/definitions\/(.*)$/.exec($ref);
+  if (match && match[1]) {
+    const parts = match[1].split("/");
+    let current = definitions;
+    for (const part of parts) {
+      if (current.hasOwnProperty(part)) {
+        current = current[part];
+      } else {
+        // No matching definition found, that's an error (bogus schema?)
+        throw new Error(`Could not find a definition for ${$ref}.`);
+      }
+    }
+    return current;
   }
+
   // No matching definition found, that's an error (bogus schema?)
   throw new Error(`Could not find a definition for ${$ref}.`);
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -318,7 +318,8 @@ function findSchemaDefinition($ref, definitions = {}) {
   if (match && match[1]) {
     const parts = match[1].split("/");
     let current = definitions;
-    for (const part of parts) {
+    for (let part of parts) {
+      part = part.replace(/~1/g, "/").replace(/~0/g, "~");
       if (current.hasOwnProperty(part)) {
         current = current[part];
       } else {
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -229,6 +229,27 @@ describe("Form", () => {
       expect(node.querySelectorAll("input[type=text]")).to.have.length.of(1);
     });
 
+    it("should handle references to deep schema definitions", () => {
+      const schema = {
+        definitions: {
+          testdef: {
+            type: "object",
+            properties: {
+              bar: { type: "string" },
+            },
+          },
+        },
+        type: "object",
+        properties: {
+          foo: { $ref: "#/definitions/testdef/properties/bar" },
+        },
+      };
+
+      const { node } = createFormComponent({ schema });
+
+      expect(node.querySelectorAll("input[type=text]")).to.have.length.of(1);
+    });
+
     it("should handle referenced definitions for array items", () => {
       const schema = {
         definitions: {
--- a/test/utils_test.js
+++ b/test/utils_test.js
@@ -365,6 +365,14 @@ describe("utils", () => {
       expect(retrieveSchema(schema, definitions)).eql(address);
     });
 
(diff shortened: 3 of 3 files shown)


glasserc:
Thanks! The code looks good, but the code format check failed, for reasons that aren't related to your change. Maybe this is because we just updated to `prettier 1.2.2`. Could you rebase onto current master?

test/Form_test.js
```diff
@@ -229,6 +229,27 @@ describe("Form", () => {
...
+          testdef: {
+            type: "object",
+            properties: {
+              bar: { type: "string" },
+            },
+          },
+        },
+        type: "object",
+        properties: {
+          foo: { $ref: "#/definitions/testdef/properties/bar" },
```
n1k0: Actually I couldn't find any reference in the spec about this capability. The only part of it using nested definitions is [this one](http://json-schema.org/latest/json-schema-core.html#rfc.section.8.2) but I'm not sure about accessing `properties` like your implemention does would actually comply.

Do you by chance have a link describing how implementations should access nested definitions?

glasserc: I couldn't find anything in the spec either, but I tried the given example on https://jsonschemalint.com/ and it works the way you would expect it to.

[Section 8.2.1](http://json-schema.org/latest/json-schema-core.html#rfc.section.8.2.1) says "Schemas can be identified by any URI that has been given to them, including a [JSON Pointer](https://tools.ietf.org/html/rfc6901)".

n1k0: Ah, indeed I could validate the example in #553 as well, I think we're good to go then (well once the prettier formatting issue is addressed).

src/utils.js
```diff
@@ -314,10 +314,21 @@ export function optionsList(schema) {
 
 function findSchemaDefinition($ref, definitions = {}) {
   // Extract and use the referenced definition if we have it.
-  const match = /#\/definitions\/(.*)$/.exec($ref);
-  if (match && match[1] && definitions.hasOwnProperty(match[1])) {
-    return definitions[match[1]];
+  const match = /^#\/definitions\/(.*)$/.exec($ref);
+  if (match && match[1]) {
+    const parts = match[1].split("/");
```
n1k0: Makes me think of properties containing a `/` in their name, which will fail here. I don't want us to over this use case, but it may be a good idea to document this limitation, or at least adding a comment about it here.

eggyal: @n1k0: Looking at the [JSON Pointer spec, §3. Syntax](https://tools.ietf.org/html/rfc6901#section-3), names containing `/` or `~` will be escaped as `~1` and `~0` respectively.  I will update to handle this, but I can't help wondering whether this project might instead use something like [json-schema-ref-parser](https://github.com/BigstickCarpet/json-schema-ref-parser) to handle these and other issues?

glasserc: I'd be in favor of that! But if you don't have time, then just a FIXME comment would be fine.

n1k0: I'd be in favor of it too, though their API being async it may be more work than needed in this specific case. A comment will be fine for now.

Also, please add a test case for this new covered edge case.

eggyal: @n1k0: Test case added.

n1k0:
Thank you :+1:

n1k0:
Released in [v0.48.0](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.48.0).

src/utils.js
eggyal: @epicfaace: Perhaps I had misunderstood what @glasserc was proposing a FIXME comment for—I assumed it was to handle `/`s, which I fixed in 95bee41 so comment no longer required?

epicfaace: > @epicfaace: Perhaps I had misunderstood what @glasserc was proposing a FIXME comment for—I assumed it was to handle `/`s, which I fixed in [95bee41](https://github.com/mozilla-services/react-jsonschema-form/commit/95bee418412810e67fb18c8d1a6c59c16d900ff6) so comment no longer required?

Oh, I assumed that the FIXME comment was supposed to be for using "json-schema-ref-parser to handle ... other issues."

Since you did fix the escaping of `/` and `~`, do you know if there's anything we're missing in this implementation which json-schema-ref-parser does well?

PR #557, opened 2017-04-22
Handle circular references

### LINKED ISSUE #556: Recursive references are not supported
### Prerequisites

### Description

Recursive references are not supported.

### Steps to Reproduce

1. Define a schema that utilises recursive referencing:
    ```json
    {
      "definitions": {
        "bar": { "$ref": "#/definitions/qux" },
        "qux": { "type": "string" }
      },
      "type": "object",
      "properties": {
        "foo": {
          "$ref": "#/definitions/bar"
        }
      }
    }
    ```

    or

    ```json
    {
      "definitions": {
        "testdef": { "$ref": "#/definitions/bar" },
        "bar": {
          "type": "object",
          "properties": {
            "qux": { "type": "string" }
          }
        }
      },
      "type": "object",
      "properties": {
        "foo": { "$ref": "#/definitions/testdef/properties/qux" }
      }
    }
    ```
2. Attempt to create a form using those schema.

#### Expected behavior

Form to be rendered.

#### Actual behavior

In the first case above ([shared playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJkZWZpbml0aW9ucyI6eyJiYXIiOnsiJHJlZiI6IiMvZGVmaW5pdGlvbnMvcXV4In0sInF1eCI6eyJ0eXBlIjoic3RyaW5nIn19LCJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJmb28iOnsiJHJlZiI6IiMvZGVmaW5pdGlvbnMvYmFyIn19fSwidWlTY2hlbWEiOnt9fQ==)), a form is rendered—but where the field is expected, the following message appears:

> **Unsupported field schema { "$ref": "#/definitions/qux" }**

In the second case above ([shared playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJkZWZpbml0aW9ucyI6eyJ0ZXN0ZGVmIjp7IiRyZWYiOiIjL2RlZmluaXRpb25zL2JhciJ9LCJiYXIiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsicXV4Ijp7InR5cGUiOiJzdHJpbmcifX19fSwidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsiZm9vIjp7IiRyZWYiOiIjL2RlZmluaXRpb25zL3Rlc3RkZWYvcHJvcGVydGllcy9xdXgifX19LCJ1aVNjaGVtYSI6e319)), no form is rendered.  Console shows:

> **Error: Could not find a definition for #/definitions/testdef/properties/qux.**

### Version

0.46.0

eggyal:

Fixes #556

### Checklist

### DIFF
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -45,7 +45,7 @@ class ObjectField extends Component {
       onBlur,
     } = this.props;
     const { definitions, fields, formContext } = this.props.registry;
-    const { SchemaField, TitleField, DescriptionField } = fields;
+    const { ObjectPropertyField, TitleField, DescriptionField } = fields;
     const schema = retrieveSchema(this.props.schema, definitions);
     const title = schema.title === undefined ? name : schema.title;
     let orderedProperties;
@@ -79,8 +79,12 @@ class ObjectField extends Component {
             formContext={formContext}
           />}
         {orderedProperties.map((name, index) => {
+          const isDefault =
+            !(name in formData) ||
+            typeof Object.getOwnPropertyDescriptor(formData, name).get ===
+              "function";
           return (
-            <SchemaField
+            <ObjectPropertyField
               key={index}
               name={name}
               required={this.isRequired(name)}
@@ -94,6 +98,7 @@ class ObjectField extends Component {
               registry={this.props.registry}
               disabled={disabled}
               readonly={readonly}
+              isDefault={isDefault}
             />
           );
         })}
--- a/src/components/fields/ObjectPropertyField.js
+++ b/src/components/fields/ObjectPropertyField.js
@@ -1,0 +1,74 @@
+import React, { Component, PropTypes } from "react";
+
+import { getDefaultRegistry } from "../../utils";
+
+class ObjectPropertyField extends Component {
+  static defaultProps = {
+    uiSchema: {},
+    errorSchema: {},
+    idSchema: {},
+    registry: getDefaultRegistry(),
+    disabled: false,
+    readonly: false,
+    autofocus: false,
+    isDefault: true,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = { present: props.required || !props.isDefault };
+  }
+
+  show = event => {
+    event.preventDefault();
+    this.setState({ present: true });
+  };
+
+  hide = event => {
+    event.preventDefault();
+    this.setState({ present: false });
+  };
+
+  render() {
+    const title = this.props.schema.title === undefined
+      ? this.props.name
+      : this.props.schema.title;
+
+    if (this.state.present) {
+      const SchemaField = this.props.registry.fields.SchemaField;
+      const { isDefault, ...props } = this.props;
+      return (
+        <div>
+          <SchemaField {...props} />
+          {this.props.required ||
+            <button onClick={this.hide}>REMOVE {title}</button>}
+        </div>
+      );
+    } else {
+      return <button onClick={this.show}>ADD {title}</button>;
+    }
+  }
+}
+
+if (process.env.NODE_ENV !== "production") {
+  ObjectPropertyField.propTypes = {
+    schema: PropTypes.object.isRequired,
+    uiSchema: PropTypes.object,
+    idSchema: PropTypes.object,
+    formData: PropTypes.any,
+    errorSchema: PropTypes.object,
+    registry: PropTypes.shape({
+      widgets: PropTypes.objectOf(
+        PropTypes.oneOfType([PropTypes.func, PropTypes.object])
+      ).isRequired,
+      fields: PropTypes.objectOf(PropTypes.func).isRequired,
+      definitions: PropTypes.object.isRequired,
+      ArrayFieldTemplate: PropTypes.func,
+      FieldTemplate: PropTypes.func,
+      formContext: PropTypes.object.isRequired,
+    }),
+    isDefault: PropTypes.bool,
+  };
+}
+
+export default ObjectPropertyField;
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -237,12 +237,14 @@ describe("Form", () => {
         definitions: {
           testdef: {
             type: "object",
+            required: ["bar"],
             properties: {
               bar: { type: "string" },
             },
           },
         },
         type: "object",
+        required: ["foo"],
         properties: {
           foo: { $ref: "#/definitions/testdef/properties/bar" },
         },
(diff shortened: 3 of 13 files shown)


glasserc:
Hi, this looks good, but I'd like to handle #554 before merging this -- I think it will make review easier.

Also, I saw "Schemas SHOULD NOT make use of infinite recursive nesting like this, the behavior is undefined." in http://json-schema.org/latest/json-schema-core.html, does this mean we can punt on handling the infinite recursion case? If we can't work on Node, that seems like a pretty big limitation...

eggyal:
@glasserc: I agree with handling #554 first—as you can see, I actually rebased to include that fix in here because they are actually dependent (my git-fu is not great however and I fear I've done something a bit weird with my branches).

Anyway, I fixed the Node issue in 80623ba — per [this comment](https://bugs.chromium.org/p/v8/issues/detail?id=6290#c6) on my bug report, the problem was that I hadn't defined `getOwnPropertyDescriptor()`.  However, whilst the existing tests are now passing I still need to add a circular reference test case because I think it's still getting stuck somewhere else.

As regards the spec, I did wonder about that... but am not 100% sure whether it's talking about the same thing?  I can't see how circularities can be avoided in many real world scenarios e.g. where the schema for "person" has a "children" property that is an array of "person" objects.  I ran into exactly this problem with the (draft) [UBL JSON schema](http://docs.oasis-open.org/ubl/UBL-2.1-JSON/v1.0/cnprd01/UBL-2.1-JSON-v1.0-cnprd01.html) (albeit with Locations that have AlternaltiveLocation children).  So, even if the JSON Schema spec currently discourages circularities I suspect that will eventually have to be changed?  [json-ref-schema-parser](https://github.com/BigstickCarpet/json-schema-ref-parser), which I mentioned previously, [happily handles](https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/docs/README.md#circular-refs) a [variety of different types of circularity](https://gist.github.com/BigstickCarpet/d18278935fc73e3a0ee1).

src/components/fields/ObjectField.js
```diff
@@ -45,7 +45,7 @@ class ObjectField extends Component {
       onBlur,
     } = this.props;
     const { definitions, fields, formContext } = this.props.registry;
-    const { SchemaField, TitleField, DescriptionField } = fields;
+    const { ObjectPropertyField, TitleField, DescriptionField } = fields;
```
eggyal: With circular schema, we must avoid rendering everything: object properties should only be rendered when necessary.  I define a new component, `ObjectPropertyField` that only renders the underlying `SchemaField` when necessary—i.e. if the schema requires the property to be present, or if the `formData` contains an explicit value for it.

src/components/fields/ObjectField.js
```diff
@@ -79,8 +79,12 @@ class ObjectField extends Component {
             formContext={formContext}
           />}
         {orderedProperties.map((name, index) => {
+          const isDefault =
+            !(name in formData) ||
+            typeof Object.getOwnPropertyDescriptor(formData, name).get ===
+              "function";
```
eggyal: Here I'm trying determine whether the value in `formData` was given explicitly, or if it's just a default value from the lazily-evaluated Proxy object (this is used in the `ObjectPropertyField` component to determine whether the underlying `SchemaField` should be rendered).  It may be better to pass explicit `formData` separately from default data?

src/components/fields/ObjectPropertyField.js
```diff
@@ -0,0 +1,74 @@
...
+      const { isDefault, ...props } = this.props;
+      return (
+        <div>
+          <SchemaField {...props} />
+          {this.props.required ||
+            <button onClick={this.hide}>REMOVE {title}</button>}
+        </div>
+      );
+    } else {
+      return <button onClick={this.show}>ADD {title}</button>;
```
eggyal: No doubt these `<button/>` elements should be replaced with a better UI?

glasserc: I'm pretty nervous about the UX in general. I know that @n1k0 took a swing at collapsibility before in https://github.com/mozilla-services/react-jsonschema-form/issues/268 and wasn't able to find a UX that he found felt right for most use cases, and this feels like a similar situation to me. Can we add this only if we are in a circular-schema situation? (N.B. I haven't actually tried running the code so this may not be as intrusive as I imagine it being.)

test/Form_test.js
```diff
@@ -273,12 +277,14 @@ describe("Form", () => {
...
           bar: {
             type: "object",
+            required: ["qux"],
             properties: {
               qux: { type: "string" },
             },
           },
         },
         type: "object",
+        required: ["foo"],
```
glasserc: These are great examples of what I don't think should be necessary.

micahscopes:
Hey, any progress on this?

epicfaace:
@eggyal now that the actual problem in #556 has been fixed, I guess the question is whether circular references should be supported.

How _do_ you envision a form should look like when it's created from a schema such as this?

```
{
        definitions: {
          bar: {
            type: "object",
            properties: {
              qux: { $ref: "#/definitions/baz" },
            },
          },
          baz: {
            type: "object",
            properties: {
              quz: { $ref: "#/definitions/bar" },
            },
          },
        },
        type: "object",
        properties: {
          foo: { $ref: "#/definitions/bar" },
        },
      }
```

eggyal:
@epicfaace:
> How _do_ you envision a form should look like when it's created from a schema such as this?

An object instance is valid even if properties are omitted (unless they are explicitly [`required`](https://json-schema.org/latest/json-schema-validation.html#rfc.section.6.5.3) by the schema).  A circular schema is only sane if it can terminate, which will occur when (at some depth) a property in the instance is so omitted.

So, the UI **must** provide some means of omitting optional properties.

Moreover, optional properties should not be rendered until they are required (either explicitly in the schema or implicitly by their existence in the document instance).  I provided in my original PR a UI through which such optional properties could be added to document instances, but @glasserc (quite rightly) felt it polluted the UI for simple cases.

Taking that into consideration, perhaps RJSF needs to:

1. Identify all circular paths within the schema;
2. Only render properties that are on circular paths if required (by schema or by instance);
3. Provide a "delete" button by which any rendered property that is not required by the schema can be removed (whether on circular path or not);
4. Provide an "add" button by which any not rendered property permitted by the schema can be added (whether on circular path or not).

This would still render entire forms in the simple cases (whilst also allowing optional properties to be properly deleted/omitted from the document, rather than merely left empty) but would furthermore support circularities.

epicfaace:
Closing this PR due to inactivity. It would be best to continue discussion of this issue in a new issue.

PR #558, opened 2017-04-22
Add UI option for omitting field label

tuunanen:

Sometimes you may want to render fields without label associated with them.

While it's possible to control the rendering of the field by creating a custom field template, or by hiding the label using CSS, they both feel non-ideal for basic use cases where you simply want to get rid of the label.

This PR introduces a new `label` option within `ui:options`. Setting the `label` to `false` will omit the label. The option won't have any effect on fields where the label is omitted already, even if it's set to `true`.

```jsx
const schema = {type: "string"};
const uiSchema = {
  "ui:options": {
    label: false
  }
};
```

I'm not sure if `ui:options` -> `label` is the best place put this under. Instead of `ui:showLabel` or `ui:hideLabel` I went with this solution to potentially support other than boolean values. For example, some `label` value could support rendering the label for screenreaders only.

After making the change, I noticed this subject has been discussed before in https://github.com/mozilla-services/react-jsonschema-form/pull/51

### Checklist

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -4,6 +4,7 @@ import {
   isMultiSelect,
   retrieveSchema,
   getDefaultRegistry,
+  getUiOptions,
   isFilesArray,
   deepEquals,
 } from "../../utils";
@@ -149,7 +150,8 @@ function SchemaFieldRender(props) {
     return <div />;
   }
 
-  let displayLabel = true;
+  const uiOptions = getUiOptions(uiSchema);
+  let { label: displayLabel } = uiOptions;
   if (schema.type === "array") {
     displayLabel = isMultiSelect(schema) || isFilesArray(schema, uiSchema);
   }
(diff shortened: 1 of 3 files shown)


src/components/fields/SchemaField.js
```diff
@@ -149,7 +150,8 @@ function SchemaFieldRender(props) {
     return <div />;
   }
 
-  let displayLabel = true;
+  const uiOptions = getUiOptions(uiSchema);
+  let { label: displayLabel } = uiOptions;
```
n1k0: I'm surprised this works without a default value set to true; hint?

tuunanen: Yeah, it did work because the default value is set in `DefaultTemplate.defaultProps`. 

Anyways, it's clearer for the reader if the default value is also set here. I made the necessary changes to the code.

n1k0:
Thank you :+1:

n1k0:

PR #579, opened 2017-05-16
getDefaultFormState populates default data to whole formData

olzraiti:

`getDefaultFormState` didn't populate formData with existing array items inside object properties with default data, and therefore wasn't working correctly with deeply nested data. I'm sorry `computeDefaults` had to be rewritten - I had to add `formData` to the parameters.

Since the logic is pretty complex and I had to rewrite it, feel free to suggest tests to add.

### Checklist


olzraiti:
Hehe yeah, for some reason all my pull requests show all the commits I've made combined in my previous pull requests.

My hands hands pretty full at the moment, so this will have to wait. I'll dig deeper into your feedback later - hopefully sometime next month. Thanks for the input :-)

olzraiti:
I found some issues when writing the tests. Local schema defaults weren't used if there was a `$ref` schema. Also defaults weren't filled properly when there was array formData with less items than `minItems`. `additionalItems` defaults weren't used.

The fixes broke one test - the logic is indeed inconsistent at the moment. Now `ArrayField List of inputs should honor given formData, even when it does not meet the minItems-requirement` is failing, and it is conflicting with `should fill defaults to an array with less items than minItems` which is succeeding. I didn't update the former yet, since this needs some discussion.

I'd remove the `ArrayField List of inputs should honor given formData, even when it does not meet the minItems-requirement`, since I think its also conflicting with #434.

> I'm glad you tackled this. I think you're exposing some very complicated and even strange behavior regarding the use of default. For example, we have an existing test "should keep parent defaults if they don't have a node level default" that I sort of disagree with. I think we need a good, clear, crisp definition of when we fill in defaults. To my mind, if a schema has a default, we use that default only when the field is completely missing. However, the test I mentioned fills in a schema's children even though the field is present!
> 
> Similarly, I am a little confused about when parentDefaults enter into play and when they override the child's defaults. If I have {default: {"a": 1}, properties: {"a": {default: 4}}}, my intuition is that the computed default should be {"a": 1} -- you fill in the default for the object, and then check its children. But that's at odds with your implementation, and I could be swayed to an alternate viewpoint/strategy, especially if it has a clear summary.

The logic I'm using is that we favor always the most nested default. I think your logic that the parent default should be filled first is logical, but somehow extending the parent default with child defaults seems intuitive to me. The situation with nested defaults is nonsensical, since no matter which way you compute the defaults, one of those defaults wouldn't be used. It's conflicting any way you look at it.

From a practical point of view, the only reason I can think of why someone would want to use nested defaults is when someone wants to extend an existing schema object, which has defaults and wants to override some child default by providing a default to the child. By using the most nested defaults, like in the logic I'm using here, that would be possible. It works the same way how we favor local schemas instead of a refered schema, without using a defined schema (I wrote a test case `should use local default in favor of ref default for object` which shows this behaviour).

Here's an example, since that was probably hard to read. We have a schema A `{default: {a: 1, b: 2}}`. Now, if I wanted to use that schema as a backbone (without creating a defition from it for some reason), but override the default for `a`, I would end up with schema B `{default: {a: 1, b: 2}, a: { default: 4}}` which would end up with `{a: 4, b: 2}` as default.

This is just one point of view - I'm not using nested defaults so I'm not sure what the practical point really is. :)

olzraiti:
I decided to revert the default population order to favor parents instead of children, because the newly merged pull requests #663 (see tests `"should render enough input to match minItems, populating the first with default values, and the rest empty"` and `"should render enough input to match minItems, populating the first with default values, and the rest with the item default"`) was incompatible with the idea. So you can ignore my previous comment, where I was gathering my thoughts about why the children-first logic could be valid :)

> For example, we have an existing test "should keep parent defaults if they don't have a node level default" that I sort of disagree with. I think we need a good, clear, crisp definition of when we fill in defaults. To my mind, if a schema has a default, we use that default only when the field is completely missing. However, the test I mentioned fills in a schema's children even though the field is present!

So this is the logic now. Hence, I made these changes to the existing tests:

- Renamed the test `"should keep parent defaults if they don't have a node level default"` to `"should keep parent defaults even if they have a node level default"` and updated the test to check that parent defaults are used. 
- Removed the test `"ArrayField List of inputs should honor given formData, even when it does not meet the minItems-requirement"`, since it is conflicting with `"should fill defaults to an array with less items than minItems"`

olzraiti:
Bumping, since this is ~3rd time I've got to handle a conflict with this PR.

This is a complicated issue and I've failed to express clearly why this solution is better than the current default populating. Partly that's because, as we have discussed, we currently don't have a clear reasoning with populating the defaults. My previous comment summarises the current conflicting approaches and the solution that this PR offers.

n7best:
does this solve #768 also?

olzraiti:
@n7best I made some changes and added a test case for defaults in dependencies, and I believe this PR would solve #768 now.

test/utils_test.js
```diff
@@ -168,7 +168,7 @@ describe("utils", () => {
         expect(
           getDefaultFormState(schema, { level1: { level2: { leaf4: 4 } } })
         ).eql({
-          level1: { level2: { leaf1: 1, leaf2: 2, leaf3: 3, leaf4: 4 } },
+          level1: { level2: { leaf1: 1, leaf2: 1, leaf3: 1, leaf4: 4 } },
```
glasserc: See, this doesn't make any sense to me. I would have expected `{ level1: { level2: { leaf3: 3, leaf4: 4 } } }`. I would have expected `level1.default` not to apply because there is already a `level1` field, and `level1.level2.default` to not apply because there is already a `level1.level2` field, and `level1.level2.leaf3.default` to apply because there is no `level1.level2.leaf3`. Could you summarize under what circumstances a default is filled in?

test/utils_test.js
```diff
@@ -242,6 +242,207 @@ describe("utils", () => {
...
+          },
+        },
+        $ref: "#/definitions/testdef",
+        default: { foo: 42 },
+      };
+
+      expect(getDefaultFormState(schema, undefined, schema.definitions)).eql({
+        foo: 42,
+      });
+    });
```
glasserc: 👍

test/utils_test.js
```diff
@@ -242,6 +242,207 @@ describe("utils", () => {
...
+      const schema = {
+        type: "array",
+        default: [1, 2],
+        items: {
+          type: "number",
+          default: 3,
+        },
+      };
+
+      expect(getDefaultFormState(schema, [0, undefined])).eql([0, 2]);
```
glasserc: Same concern here about filling in elements of `default` when the field itself already exists.

test/utils_test.js
```diff
@@ -242,6 +242,207 @@ describe("utils", () => {
...
+        items: {
+          type: "number",
+          default: 0,
+        },
+      };
+
+      expect(getDefaultFormState(schema, [1])).eql([1, 0]);
+    });
+
+    it("should use array  array default in favor or array item default even if it is a ref", () => {
```
glasserc: `array  array default`?

test/utils_test.js
```diff
@@ -242,6 +242,207 @@ describe("utils", () => {
...
+            type: "number",
+            default: 1,
+          },
+        },
+      };
+
+      expect(getDefaultFormState(schema, [undefined], schema.definitions)).eql([
+        0,
+      ]);
+    });
```
glasserc: Same concern about filling in the toplevel default when the element exists. I think this test would make sense to me if instead of `[undefined]`, you had `undefined`. However, as written, I would have expected it to be `[1]`.

test/utils_test.js
```diff
@@ -242,6 +242,207 @@ describe("utils", () => {
...
+            ],
+          },
+        },
+      };
+
+      expect(getDefaultFormState(schema, undefined)).eql({
+        "Do you have a pet?": "Yes",
+        "Does it bite?": true,
+      });
+    });
```
glasserc: Very cool!

olzraiti:
Hey,

Currently I don't have the time for this one. I read quickly your feedback (thanks!), and the axiom you are suggesting seems intuitive. Feel free to take the wheel :-)

martijnthe:
@olzraiti it's a bummer this hasn't moved along.
I'd like to see #768 get fixed and it looks like it does?
What's left here to do?

test/utils_test.js
epicfaace: I agree @glasserc

MatinF:
Any update on this?

epicfaace:
@glasserc has left some comments, so if anyone is interested in continuing from this PR and resolving the issues (or fixing this problem another way), feel free to do so.

olzraiti:
I'm closing this since the original issue is fixed by #1499 and I don't remember anymore if this PR had something else relevant in it, and there are so many conflicts that I don't find it likely that anyone would dig into this anymore.

Issue #532, opened 2017-03-31
enumNames is not in JSON Schema spec


dlax:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

react-jsonschema-form has implemented support for [labels in enum fields][] through the `enumNames` keyword following a [proposal](https://github.com/json-schema/json-schema/wiki/enumNames-%28v5-proposal%29) which then turned into issue json-schema-org/json-schema-spec#57 in which the proposal got refused. So the `enumNames` keyword does not appear and will probably not appear in JSON Schema specification format and react-jsonschema-form thus deviates from it.

[labels in enum fields]: https://github.com/mozilla-services/react-jsonschema-form/blob/v0.45.0/README.md#custom-labels-for-enum-fields

From json-schema-org/json-schema-spec#57, the recommended way to handle labels in enum is to use a `oneOf` with a set of schemas. So following the example from [documentation][labels in enum fields], it would be (in draft-v5):
```json
{
  "type": "number",
  "oneOf": [
    {"enum": [1], "title": "one"},
    {"enum": [2], "title": "two"},
    {"enum": [3], "title": "three"}
  ]
}
```
In draft-v6, a `const` keyword got added (see json-schema-org/json-schema-spec#139), so the example above could be simplified as:
```json
{
  "type": "number",
  "oneOf": [
    {"const": 1, "title": "one"},
    {"const": 2, "title": "two"},
    {"const": 3, "title": "three"}
  ]
}
```

### Version

0.45

Natim:
Well let see how it turns out, because the usability of the current solution is much better.

dlax:
> Do you think react-jsonschema-form should implement all of this specs? It sounds terrible to me.

Not sure what you mean by "all this specs", but I'd say it should rather implement something from the spec than something not from the spec (as `enumNames`). The fact that implementation on this side preceded adoption in JSON Schema format is rather unfortunate. At least the documentation could be clarified.

This description is indeed more verbose but it has the advantage of coupling tight the value and the title rather than relying on the order of two arrays. Another advantage is that is make use of an existing construct `oneOf`, which highlights the versatility of JSON Schema format.

Unfortunately (#52), `oneOf` is not supported by react-jsonschema-form.

[...]

PR #581, opened 2017-05-17
Support oneOf and anyOf as alternative of enum/enumNames

fbessou:

As mentioned by @dlax in #532 enumNames keyword is not in the JSON Schema specification (neither draft-04 nor draft-06).
Another way to support choices schemas is to write them with the following structure :
```js
{
  oneOf: [{
    enum: ["foo"],
    title: "Foo"
  }, {
    enum: ["bar"],
    title: "Bar"
  }]
}
```
or with draft-06's new `const` keyword:
```js
{
  oneOf: [{
    const: "foo",
    title: "Foo"
  }, {
    const: "bar",
    title: "Bar"
  }]
}
```
This PR adds basic support for oneOf/anyOf as an alternative to the non standard `enumNames` keywords.

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -277,13 +277,40 @@ export function orderProperties(properties, order) {
   return complete;
 }
 
+/**
+ * This function checks if the given schema matches a single
+ * constant value.
+ */
+export function isConstant(schema) {
+  return (
+    (Array.isArray(schema.enum) && schema.enum.length === 1) ||
+    schema.hasOwnProperty("const")
+  );
+}
+
+export function toConstant(schema) {
+  if (Array.isArray(schema.enum) && schema.enum.length === 1) {
+    return schema.enum[0];
+  } else if (schema.hasOwnProperty("const")) {
+    return schema.const;
+  } else {
+    throw new Error("schema cannot be inferred as a constant");
+  }
+}
+
 export function isMultiSelect(schema, definitions = {}) {
   if (!schema.uniqueItems || !schema.items) {
     return false;
   }
   const itemsSchema = retrieveSchema(schema.items, definitions);
-  return Array.isArray(itemsSchema.enum);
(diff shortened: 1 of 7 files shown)


test/utils_test.js
```diff
@@ -270,34 +270,59 @@ describe("utils", () => {
...
+        it("should be true if oneOf/anyOf schemas are all constants", () => {
+          const schema = {
+            items: { oneOf: [constantSchema, constantSchema] },
+            uniqueItems: true,
+          };
+          expect(isMultiSelect(schema)).to.be.true;
+        });
+      });
+
+      it("should retrieve reference schema definitions", () => {
```
glasserc: `referenced`

src/utils.js
```diff
@@ -294,20 +294,25 @@ export function toConstant(schema) {
...
   }
   return false;
 }
 
+export function isMultiSelect(schema, definitions = {}) {
+  if (!schema.uniqueItems || !schema.items) {
+    return false;
+  }
+  const itemsSchema = retrieveSchema(schema.items, definitions);
+  return isSelect(itemsSchema);
```
glasserc: How about `isSelect(schema.items, definitions)` instead?

fbessou:
Currently there isn't any JSDoc comment in the react-jsonschema-form sources. Should I add a simple comment or a JSDoc comment ?

glasserc:
We don't have any code that does anything specific with JSDoc comments and actually most functions are not documented using JSDoc, but adding a JSDoc comment is fine. It doesn't have to be complicated -- the example on validateFormData is fine.

fbessou:
@glasserc, I made all the requested changes except the renaming of `isSelect` for which I haven't found a better name.

glasserc:
Thanks!

glasserc:
Released in v0.50.0.

PR #582, opened 2017-05-18
Options disable feature added

VijayKrish93:

Selection widget option elements currently does not have the ability to disable an option/options through schema.I had to deal with a use case in my project where I needed to disable some parent option elements so that they become non selectable.I went through the code and I found that a extra attribute must be added.

So this PR introduces a feature to disable a StringField.UiWidget options through schema 

`{
 "type":"string",
 "enum":["foo","bar"],
 "enumNames":["Foo","Bar"],
 "enumDisabled":["bar"]
}`

### Checklist


glasserc:
Hi, thanks for your contribution! I'm not crazy about the use of `enumDisabled` in the schema. `enumNames` are already nonstandard (see #532), and disabling of an option seems like something that belongs in the `uiSchema`.

Also, can you add tests when you add a feature like this? Thanks!

VijayKrish93:
Hi @glasserc

glasserc:
This PR was superseded by #708.

PR #586, opened 2017-05-23
Added Code widget.

stepek:
We need code editor widget for xml yaml and javascript formats.

### Checklist


stepek:
@n1k0 Could you check this?

n1k0:
Sorry, I've been busy on other fronts lately. So that feature is interesting, though I feel like this shouldn't part of the core. Why don't you expose this feature in a separate repo/package, so people could register your custom code widget when they actually need it?

zboldyga:
@n1k0 I'm currently using this project and am about to add a couple of extra components as well - a typeahead field (e.g. https://github.com/twitter/typeahead.js), a complex table component (https://github.com/AllenFang/react-bootstrap-table), and at least one more that combines search in a text field with a table view list of items that are selected. I imagine these are not going to be ideal features to merge into your project. What approach would you recommend so that we won't clash with the direction of react-jsonschema-form, but so that we can still have a direct dependency to your project (e.g. we don't want to fork)?

n1k0:
> What approach would you recommend so that we won't clash with the direction of react-jsonschema-form

You can always pin a specific 0.n.x release version of this lib to secure yourself. Things will be more sane when we tag a 1.0 release.

n1k0:
I'm gonna close this as I'd really like this to live in its own independent package.

PR #587, opened 2017-05-23
Added possibility to change title and description by ui:schema.

stepek:
Added possibility to change title and description of field using ui:schema.

### Checklist


n1k0:
Title and description are part of the JSONSchema spec, why would one want to duplicate the source for these information?

stepek:
This is no duplication of data, because main purpose is add possibility to change title and description, (for e.g. locale settings); if ui:title is not setted, user will see jsonschema title.

n1k0:
Ok I could see use cases for people using generated schemas or things like that. Could you please add tests and update docs for this patch? Thanks.

stepek:
@n1k0 Done

PR #589, opened 2017-05-24
Fixes default value for object missing properties

vierno:

When the schema of a field of type object is missing the `properties` key, it raises an exception: `Cannot convert undefined or null to object`. This PR fixes it by defaulting the value to an empty object, this way, we can use custom fields to render objects without defined properties.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -128,7 +128,7 @@ function computeDefaults(schema, parentDefaults, definitions = {}) {
   switch (schema.type) {
     // We need to recur for object schema inner default values.
     case "object":
-      return Object.keys(schema.properties).reduce((acc, key) => {
+      return Object.keys(schema.properties || {}).reduce((acc, key) => {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -128,16 +128,20 @@ function computeDefaults(schema, parentDefaults, definitions = {}) {
...
-        // have from a previous run: defaults[key].
-        acc[key] = computeDefaults(
-          schema.properties[key],
-          (defaults || {})[key],
-          definitions
-        );
-        return acc;
-      }, {});
+      if (schema.properties) {
+        return Object.keys(schema.properties).reduce((acc, key) => {
```
n1k0: Instead of introducing an `if`, you could just do `return Object.keys(schema.properties || {})`.

vierno:
Hello @n1k0 , I made the requested changes, thanks!

n1k0:
Thanks you!

PR #590, opened 2017-05-25
Fixes array without items for custom fields

vierno:

The jsonschema of an array may omit the `items` property, similarly to #589, with this fix it's possible to create a field for custom arrays without items defined by the schema.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -278,12 +278,16 @@ export function orderProperties(properties, order) {
 }
 
 export function isMultiSelect(schema) {
-  return Array.isArray(schema.items.enum) && schema.uniqueItems;
+  return schema.items
+    ? Array.isArray(schema.items.enum) && schema.uniqueItems
+    : false;
 }
 
 export function isFilesArray(schema, uiSchema) {
   return (
-    (schema.items.type === "string" && schema.items.format === "data-url") ||
+    (schema.items &&
+      schema.items.type === "string" &&
+      schema.items.format === "data-url") ||
     uiSchema["ui:widget"] === "files"
   );
 }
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -278,12 +278,16 @@ export function orderProperties(properties, order) {
 }
 
 export function isMultiSelect(schema) {
-  return Array.isArray(schema.items.enum) && schema.uniqueItems;
+  return schema.items
+    ? Array.isArray(schema.items.enum) && schema.uniqueItems
+    : false;
```
n1k0: Should work just fine with `return schema.items && Array.isArray(schema.items.enum) && schema.uniqueItems`

test/utils_test.js
```diff
@@ -277,6 +283,19 @@ describe("utils", () => {
     });
   });
 
+  describe("isFilesArray()", () => {
+    it("should be true if items have data-url format", () => {
+      const schema = { items: { type: "string", format: "data-url" } };
+      const uiSchema = {};
+      expect(isFilesArray(schema, uiSchema)).to.be.true;
+    });
+    it("should be false if items is undefined", () => {
```
n1k0: Nit: please separate these two tests with a blank line

src/utils.js
vierno: This code would return `undefined`, no `false`, is it ok?

n1k0: Damn you're right, I'm doing too much Elm these days :D

vierno: hehe =)
I made other requested change.

n1k0:
:+1:

PR #596, opened 2017-05-31
Resolves  from definitions, detects circular references

vamsiampolu:

Adding support for `$id` within `definitions` when data is normalized using `normalizr` which requires an `id` attribute of some sort to aid it. The data is only one level deep within the `definitions` which is provided as the key to the  `schema.Entity` defined there as mentioned in https://github.com/mozilla-services/react-jsonschema-form/issues/594

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1395,6 +1395,98 @@ This library partially supports [inline schema definition dereferencing]( http:/
 }
 ```
 
+The library partially supports `$id` for any object placed within the `definitions` property. A property within the form can refer to a definition using a `$id` field like this:
+
+```json
+{
+  "title": "A registration form",
+    "description": "A simple form example.",
+    "type": "object",
+    "required": [
+      "firstName",
+    "lastName"
+    ],
+    "definitions": {
+      "firstName": {
+        "$id": "#firstName",
+        "type": "string",
+        "title": "First name"
+      },
+      "lastName": {
+        "$id": "#lastName",
+        "type": "string",
+        "title": "Last name"
+      },
+      "age": {
+        "$id": "#age",
+        "type": "integer",
+        "title": "Age"
+      },
+      "bio": {
+        "$id": "#bio",
+        "type": "string",
+        "title": "Bio"
+      },
+      "password": {
+        "$id": "#password",
+        "type": "string",
+        "title": "Password",
+        "minLength": 3
+      }
+    },
+    "properties": {
+      "firstName": {
+        "$ref": "#firstName"
+      },
+      "lastName": {
+        "$ref": "#lastName"
+      },
+      "age": {
+        "$ref": "#age"
+      },
+      "bio": {
+        "$ref": "#bio"
+      },
+      "password": {
+        "$ref": "#password"
+      }
+    }
+}
+```
+
+Sometimes normalizing a deep nested schema can result in a subschema which is accessed by a reference and has references of its own. Such references will be followed unless the subschema refers to itself or has a circular reference between two items, then the form will throw an Error due to the malformed schema.
+
+Self referencing schema:
+
+```json
+{
+  "definitions": {
+    "object": {
+      "$id": "#object",
+      "$ref": "#object"
+    }
+  }
+}
+```
+
+Circular reference within the schema:
+
+```json
+{
+  "definitions": {
+    "object1": {
+      "$id": "#object1",
+        "$ref": "#object2"
+    },
+
+    "object2": {
+        "$id": "#object2",
+        "$ref": "#object1"
+    }
+  }
+}
+```
+
 *(Sample schema courtesy of the [Space Telescope Science Institute](http://spacetelescope.github.io/understanding-json-schema/structuring.html))*
 
 Note that it only supports local definition referencing, we do not plan on fetching foreign schemas over HTTP anytime soon. Basically, you can only reference a definition from the very schema object defining it.
--- a/README.md
+++ b/README.md
@@ -1395,6 +1395,8 @@ This library partially supports [inline schema definition dereferencing]( http:/
 }
 ```
 
+*(Sample schema courtesy of the [Space Telescope Science Institute](http://spacetelescope.github.io/understanding-json-schema/structuring.html))*
+
 The library partially supports `$id` for any object placed within the `definitions` property. A property within the form can refer to a definition using a `$id` field like this:
 
 ```json
@@ -1404,7 +1406,7 @@ The library partially supports `$id` for any object placed within the `definitio
     "type": "object",
     "required": [
       "firstName",
-    "lastName"
+      "lastName"
     ],
     "definitions": {
       "firstName": {
@@ -1454,7 +1456,7 @@ The library partially supports `$id` for any object placed within the `definitio
 }
 ```
 
-Sometimes normalizing a deep nested schema can result in a subschema which is accessed by a reference and has references of its own. Such references will be followed unless the subschema refers to itself or has a circular reference between two items, then the form will throw an Error due to the malformed schema.
+Subschemas can reference other subschemas, which can reference other subschemas and so on. This chain of references will be followed, but if it appears to be infinite or circular, an error will be thrown.
 
 Self referencing schema:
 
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -314,26 +314,84 @@ export function optionsList(schema) {
...
+}
+
+function isJSONPointer($ref) {
+  const regex = /^#\/definitions\/(.*)$/;
+  return regex.test($ref);
+}
+
+function has$id($id, defs) {
+  let $idsWithFalsyValues = Object.keys(defs).map(key => defs[key].$id);
+  let $ids = compact($idsWithFalsyValues);
```
glasserc: You should use `filter` instead, maybe even before `map`.

src/utils.js
```diff
@@ -314,26 +314,84 @@ export function optionsList(schema) {
...
+        if (key$id === $ref) {
+          let current = definitions[key];
+          if (current.hasOwnProperty("$ref") && is$id(current.$ref)) {
+            processedIds = processedIds.concat($ref);
+            //One question refers to itself or to another question inside it, used for object and array schemas
+            if (processedIds.includes(current.$ref)) {
+              throw new Error(
+                `Circular or self reference detected when reading schema for ${current.$ref}`
+              );
+            }
```
glasserc: I don't think it's necessary for us to worry about circular references as part of this PR, but if we do, we should worry about them for both JSON Pointer `$ref` as well as `$ref` that matches an ID.

src/utils.js
```diff
@@ -314,26 +314,84 @@ export function optionsList(schema) {
...
+    }
+  } else if (is$id($ref)) {
+    if (has$id($ref, definitions)) {
+      let keys = Object.keys(definitions);
+      for (let key of keys) {
+        let { $id: key$id } = definitions[key];
+        if (key$id === $ref) {
+          let current = definitions[key];
+          if (current.hasOwnProperty("$ref") && is$id(current.$ref)) {
+            processedIds = processedIds.concat($ref);
```
glasserc: You can just use `processedIds.push($ref)`.

src/utils.js
```diff
@@ -314,26 +314,84 @@ export function optionsList(schema) {
...
+            if (processedIds.includes(current.$ref)) {
+              throw new Error(
+                `Circular or self reference detected when reading schema for ${current.$ref}`
+              );
+            }
+            return findSchemaDefinition(
+              current.$ref,
+              definitions,
+              processedIds
+            );
```
glasserc: There's already some recursive logic that does this in e.g. `computeDefaults`. I think you're right to have that logic in this method instead, but then it should be moved from `computeDefaults`, and `findSchemaDefinition` should be recursive for both JSON Pointer as well as `$id` cases.

src/utils.js
```diff
@@ -314,26 +314,84 @@ export function optionsList(schema) {
...
+          }
+          return current;
+        }
+      }
+    } else {
```
glasserc: Not every case above returns/throws, so you should probably not have this inside an `else` block.

README.md
```diff
@@ -1395,6 +1395,98 @@ This library partially supports [inline schema definition dereferencing]( http:/
...
+The library partially supports `$id` for any object placed within the `definitions` property. A property within the form can refer to a definition using a `$id` field like this:
+
+```json
+{
+  "title": "A registration form",
+    "description": "A simple form example.",
+    "type": "object",
+    "required": [
+      "firstName",
+    "lastName"
```
glasserc: nit: indentation

README.md
```diff
@@ -1395,6 +1395,98 @@ This library partially supports [inline schema definition dereferencing]( http:/
...
+
+    "object2": {
+        "$id": "#object2",
+        "$ref": "#object1"
+    }
+  }
+}
+```
+
 *(Sample schema courtesy of the [Space Telescope Science Institute](http://spacetelescope.github.io/understanding-json-schema/structuring.html))*
```
glasserc: I think this credit should come before your additional example.

README.md
```diff
@@ -1395,6 +1395,98 @@ This library partially supports [inline schema definition dereferencing]( http:/
...
+        "$ref": "#bio"
+      },
+      "password": {
+        "$ref": "#password"
+      }
+    }
+}
+```
+
+Sometimes normalizing a deep nested schema can result in a subschema which is accessed by a reference and has references of its own. Such references will be followed unless the subschema refers to itself or has a circular reference between two items, then the form will throw an Error due to the malformed schema.
```
glasserc: I don't really understand what `normalizing` is supposed to mean in this context. How about something like `Subschemas can reference other subschemas, which can reference other subschemas and so on. This chain of references will be followed, but if it appears infinite or circular, an error will be thrown.`

src/utils.js
vamsiampolu: In the new commit, I have added support to JSON Pointer for following references to another JSON Pointer or to a $id, the tests now reflect cases where a user creates a schema which tries to incorporate both $id and JSON Pointer as well.

src/utils.js
vamsiampolu: I think you would like to take this part of the `computeDefaults` logic and move it into `findSchemaDefinition`:

```js
  switch (schema.type) {
    // We need to recur for object schema inner default values.
    case "object":
      return Object.keys(schema.properties || {}).reduce((acc, key) => {
        // Compute the defaults for this node, with the parent defaults we might
        // have from a previous run: defaults[key].
        acc[key] = computeDefaults(
          schema.properties[key],
          (defaults || {})[key],
          definitions
        );
        return acc;
      }, {});

    case "array":
      if (schema.minItems) {
        return new Array(schema.minItems).fill(
          computeDefaults(schema.items, defaults, definitions)
        );
      }
  }
```

However, instead of resolving the default values, you want to resolve the schema definition for each `property` of an `ObjectField`'s schema. I know that `SchemaField` is rendered for each field within an ObjectField and it calls the `retrieveSchema` function  before delegating to the appropriate field and widget. I think the same happens for each item within the `ArrayField` when using the `toNormalArray` function.

I think this is good because a field's schema is only resolved when it is being rendered and not prematurely.

src/utils.js
vamsiampolu: I have added more cases to flesh out where errors are thrown, I think it should throw an Error only when an illegal item is found.

glasserc: Sorry, I misunderstood how this code worked. I thought it was possible to get to the end of the `if (isJSONPointer($ref))` block without returning or throwing. I see now that isn't possible because it's the same regular expression. I'd like to see this cleaned up; instead of `isJSONPointer`, how about `parseJSONPointer` that returns the match object or undefined? That way we don't have to have a second check against the same regular expression, nor an "always true" check for `if (match && match[1])`.

src/utils.js
glasserc: Never mind, it seems like `computeDefaults` already calls `findSchemaDefinition` when necessary. I'm not sure what I was thinking about here..

glasserc:
Thanks for updating your patch! I still don't really understand the use case you're trying to address, and I don't see this PR as on the path to complete support for `$id`, so I'm still not excited about merging this PR. Could you please elaborate on what you're trying to do?

vamsiampolu:
When I made the pull request I wanted to define a completely normalized form structure using the spec from `json-schema`.  I believed that it would improve handling of nested object schemas for slightly more complicated use cases. And I wanted this library to handle these concerns.

I still believe that normalizing the form schema is essential for some use cases which I describe below.  Firstly, it de-duplicates your data which you support using `JSON Pointers`.  `JSON Pointers` need to be resolved using a recursive path matching. I believed that I could denormalize my data using `normalizr`, it turns out that matching `id`  would also need a similar logic when using within RJSF. 

Also, normalizing data is good for lazy loading parts of your schema  especially if you are building multi page journeys or wizards. However, in such cases, it would be much better to do it outside the form using a different module such as normalizr or redux-orm instead of applying it within the form as:

epicfaace:
Hi @vamsiampolu , apologies this PR wasn't reviewed further at the time, but it's been so long since you created this PR that I think we should just go ahead and close it. If you are still interested in implementing this, please feel free to make another PR from the latest / newest code.

PR #599, opened 2017-06-01
Consider maxItems value before rendering the add button

adimascio:
jsonschema refuses to validate an array of items when the number of
items is greater than the maxItems constraint. Therefore the UI should
consisder this constraint and not provide an "add" button when the
maximum number of elements is reached.

The "add" button is rendered in the following cases:

- ui:options.addable is not false and items.length < schema.maxItems
- schema.maxItems is unspecified and ui:options.addable is not false

As before, if ui:options.addable is not specified, it is considered as
true by default.

The "add" button is *not* rendered in the following cases:

- ui:options.addable is false,
- items.length >= schema.maxItems, even if ui.options:addable is true

[Please describe them here]

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -200,6 +200,21 @@ class ArrayField extends Component {
     return itemSchema.type !== "null";
   }
 
+  canAddItem(formItems) {
+    const { schema, uiSchema } = this.props;
+    let { addable } = getUiOptions(uiSchema);
+    if (addable !== false) {
+      // if ui:options.addable was not explicitly set to false, we can add
+      // another item if we have not exceeded maxItems yet
+      if (schema.maxItems !== undefined) {
+        addable = formItems.length < schema.maxItems;
+      } else {
+        addable = true;
+      }
+    }
+    return addable;
+  }
+
   onAddClick = event => {
     event.preventDefault();
     const { schema, formData, registry = getDefaultRegistry() } = this.props;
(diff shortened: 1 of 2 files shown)


src/components/fields/ArrayField.js
```diff
@@ -434,14 +440,21 @@ class ArrayField extends Component {
...
       items = items.concat(new Array(itemSchemas.length - items.length));
     }
+    let { addable } = getUiOptions(uiSchema);
+    if (addable !== false) {
+      if (schema.maxItems !== undefined) {
+        addable = items.length < schema.maxItems;
+      } else {
+        addable = true;
+      }
+    }
```
glasserc: Should we maybe extract this logic into a utility function somewhere?

test/ArrayField_test.js
```diff
@@ -793,6 +843,65 @@ describe("ArrayField", () => {
       expect(node.querySelector(".array-item-add button")).to.be.null;
     });
 
+    it("[fixed-noadditional] should not provide an add button regardless maxItems", () => {
+      const maxItemsSchema = Object.assign({ maxItems: 3 }, schema);
```
glasserc: We tend to prefer object spread in this codebase, so you could write `{ maxItems: 3, ...schema }`.

adimascio:
Hi @glasserc 

Thanks for your comments. I've updated the pull request to take your requests into account. Is it good enough ?

Issue #79, opened 2016-03-22
Custom validator


alexeygolev:
It would be nice to pass a custom validator instance. One could define custom validations on that instance. I can submit a PR if needed:)

n1k0:
You mean a custom [jsonschema](https://www.npmjs.com/package/jsonschema) one? Sounds good, feel free to send a PR.

n1k0:
Closing in favor of #145.

PR #633, opened 2017-07-13
Allows injecting custom jsonschema Validator instance

simonschmidt:
The current custom validation does not have access to the schema
in the same way that a jsonschema validator does, by allowing
user to specify the validator many validation tasks become much
simpler.

This is related to #79 which was closed when the custom field validation was implemented,
however that is not as powerful as a jsonschema validator can be.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1303,6 +1303,33 @@ render((
 >   received as second argument.
 > - The `validate()` function is called **after** the JSON schema validation.
 
+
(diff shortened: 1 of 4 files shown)


epicfaace:
Now that we've switched over to ajv, is this PR still relevant?

epicfaace:
I'm closing this PR, but do comment if you think it's still relevant.

Issue #638, opened 2017-07-18
Removing properties from schema (as demonstrated in "conditional fields" example) does not remove field from rendered form


LyraelRayne:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Removing a property from (a clone of) the schema does not remove the field from the form

### Steps to Reproduce

1. Open the "conditional fields" example linked to from the project page (https://jsfiddle.net/69z2wepo/68259/)
2. Tick the box to cause the "Name" field to be displayed
3. Untick the box

#### Expected behavior

The name field should disappear

#### Actual behavior

The name field doesn't disappear

### Version

0.49.0

Natim:
I guess it is because you are adding it if it is `checked` but no removing it if it is `unchecked`.

Natim:
See https://jsfiddle.net/69z2wepo/83017/

LyraelRayne:
I didn't write the code, it's an example from https://github.com/mozilla-services/react-jsonschema-form#tips-and-tricks (so I presumed that it was meant to work as-is).

The code example does actually remove them, but it does it by mutating the schema.properties instead of replacing it with a modified clone. 

https://jsfiddle.net/69z2wepo/83018/ is a working version of the original example. I would suggest that you guys update the example link to the fixed version.

Natim:
@LyraelRayne would you like to send a pull-request with that modification?

LyraelRayne:
I'm not really sure how to go about that (never done one on github). Do I have to make a fork or something?

LyraelRayne:
Quick question: Do I have to keep the fork around once the pull request is merged?

Natim:
Nope :)

LyraelRayne:
Thanks. That was all a lot less scary than it seemed ;)

PR #640, opened 2017-07-18
Fix example of conditional fields so that it doesn't mutate

LyraelRayne:
The original example mutated schema.properties. This update links to a new version of the jsbin which adds the following line:

        schema.properties = Object.assign({},schema.properties)

See #638

The example does not currently work correctly.

### Checklist


Natim:
Thanks.

glasserc:
Released in v0.50.0.

PR #642, opened 2017-07-18
add allowMutation option to formContext

ssured:

I'd like to integrate MobX and RJSF. MobX is able to track object changes by invisibly adding listeners, getters and setters to regular javascript objects. See https://github.com/mobxjs/mobx and https://github.com/mobxjs/mobx-state-tree.

The issue I hit is that each change to the `formData` object in RJSF will trigger an `onChange` with a __copy__ of the `formData`. For many programs this is really safe, but for me my 'MobX' object is lost and changed into a regular object. 

This change proposes an extra setting. When passing `formContext: { allowMutation: true }` to the form props, your `formData` will be mutated on each change, instead of copied.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1113,6 +1113,8 @@ The registry is passed down the component tree, so you can access it from your c
 
 You can provide a `formContext` object to the Form, which is passed down to all fields and widgets (including [TitleField](#custom-titles) and [DescriptionField](#custom-descriptions)). Useful for implementing context aware fields and widgets.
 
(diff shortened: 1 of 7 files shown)


ssured:
What I dislike from my PR is that `formContext` is cluttered with a special property. Maybe adding a global config option on the `Form` component is a better solution?

adarshks:
I have found this PR very useful , as i also got an issue #644 . I request RJSF team to accept this commit.

ssured:
Well this does not work nice with react and rerendering the components. Will close for now

PR #643, opened 2017-07-21
custom setState is not used

ssured:
utils->setState depends on safeRenderCompletion, which is not an allowed prop on FileWidget

Declutter the code

### Checklist


epicfaace:
Closing this because we no longer use safeRenderCompletion.

PR #646, opened 2017-07-22
Add Advanced conditional fields example

tuunanen:

We have been building some complex forms using React JSON Schema form component at work. The requirement was to have a simple method of showing/hiding fields depending on the current state of the form. For example, “if the value of `name` field is 'bob' or 'alice' then show `age` field”.

Instead of doing it manually for each form in the onChange handler, the logic has been put into a custom “condition” property in the UI Schema that is processed dynamically upon changes in the form state.

Example

```js
const schema = {
  type: 'object',
  properties: {
    name: {
      type: 'string'
    },
    age: {
      type: 'number'
    }
  }
};
const uiSchema = {
  name: {},
  age: {
    // Show `age` field if `name` is bob or alice
    condition: 'name=bob,alice'
  }
};
```

Supported conditions in the example are:
- foo=bar
- foo!=bar
- foo=bar,baz
- foo!=bar,baz
- foo=bar&&bar=foo
- foo=bar||bar=foo

...and some permutations of these.

I commented and documented the example quite thoroughly. Hopefully it will be helpful for others too!

Link to the fiddle:
https://jsfiddle.net/cowbellerina/zbfh96b1/

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1412,6 +1412,7 @@ This component follows [JSON Schema](http://json-schema.org/documentation.html) 
  - Multi-step wizard: https://jsfiddle.net/sn4bnw9h/1/
  - Using classNames with uiSchema: https://jsfiddle.net/gfwp25we/1/
  - Conditional fields: https://jsfiddle.net/69z2wepo/83018/
+ - Advanced conditional fields: https://jsfiddle.net/cowbellerina/zbfh96b1/
  - Use radio list for enums: https://jsfiddle.net/f2y3fq7L/2/
  - Reading file input data: https://jsfiddle.net/f9vcb6pL/1/
  - Custom errors messages with transformErrors : https://jsfiddle.net/revolunet/5r3swnr4/


toni-h:
Wow really nice :) i have been fighting with this kind of conditioning for few weeks now. Thank you for your effort!

tuunanen:
No, unfortunately there’s no support for Array fields yet. Similarly nested properties, e.g. parent.child = foo, won’t work.

toni-h:
ok too bad. do you know any hacky way to do this? i only need it for one enum inside array :)

n1k0:
This is a pretty sweet example, I'm landing this. Cheers.

tuunanen:
@toni-h: I have updated the fiddle. Now it should support arrays too. The relevant change was done to lines 168-179. The condition value is now matched against an array of values instead of a single one.

```js
if (field && newFormData.hasOwnProperty(field)) {
    let currentValues = _.isArray(newFormData[field])
        ? newFormData[field]
        : [ newFormData[field] ];
    _.each(values, (value) => {
        if (invert) {
            visible = visible && _.indexOf(currentValues, value) === -1;
        }
        else {
            visible = visible || _.indexOf(currentValues, value) !== -1;
        }
    });
}
```

wstam88:
Does it work with booleans as well? Doesn't seem to work.

tuunanen:
Hi @wstam88: It should work with `foo=true` or `foo=0` etc. The values are cast as strings but they work nevertheless. Am I missing something?

glasserc:
Released in v0.50.0.

ml242:
Fiddle is broken, would it be possible to update the live playground?

tuunanen:
Hi @ml242, the fiddle is fixed now. It had broken because `react-with-addons.js` and `react-dom.js` files have been deprecated since React v15.5 and are no longer available through the CDN. The quick fix was to simply load the two scripts with explicit version numbering using React v15.

The conditional fields example covers your use case. While the fiddle might look intimidating, getting the functionality to your application is simpler than it looks. The only external dependency is to lodash library. The three utility functions at the bottom may be directly copy and pasted to your application. Then, depending on your application, the `processForm` function should be called from your Form onChange handler (see `MyComp` from the fiddle).

ml242:
I'll give it another look, thanks!

mr-madamin:
@tuunanen, this is a very useful example, thanks! The only thing that is missing for me is depth. Could you give some instructions or anything on how to make nested properties work as well?

marinav:
Hi,

Thank you very much for your example, that's exactly what I was looking for !
I just have a problem with boolean. @tuunanen you said it was working well but I can't make it work... Maybe I missed something...? Can you please help me ? Here a part of my code : 

```
const originalSchema = {
  "title": 'my schema',
  "type": 'object',
  "properties": {
    "input1": {
      "title": "Input 1",
      "type": "string",
      "minLength": 5
    },
    "input2": {
      "type": "string",
      "title": "Input 2",
      "type": "boolean",
      "default": false
    },
   }
}

const originalUISchema = {
  "ui:order": ["input1", "input2"],
    "input2": {
    "condition": 'input2=true',
    "ui:widget": "radio",
  },
}
```

Thank you !

tuunanen:
Hi @marinav, indeed it seems booleans no longer work. I suspect the changes I did to support arrays in Jul 27 have broken this. The comparison is now strict therefore e.g. `true === 'true'` fails to match the condition.

A boring solution for this is to transform the array of values before the comparison is done as follows (lines 165-175 in the example)

```js
const { field, values: stringValues, invert } = getConditionRule(rule);
let visible = invert;

const values = stringValues.map(value => {
    if (value === 'true') {
        value = true;
    } else if (value === 'false') {
        value = false;
    }
    return value;
});
```

I updated the example accordingly. Now it should work. Thanks for the bug report.

---

There are also two minor issues in the code example you provided. First, the type is set twice for "input2" in the schema. Second, the radio widget should be applied to "input2" in the UI schema. Now it's for "input1" which is a string field.

I made a fork of the example with your use case:
https://jsfiddle.net/cowbellerina/mf4xcbuz/

Hopefully this helps.

marinav:
Thank you very much @tuunanen ! It works very well know :)

My bad for the two mistakes, it's because of copy/cut. Thank you for the example, it's perfect 👍

PR #623, opened 2017-06-29
Easier Bootstrap Grid usage and basic BS4 compatibility

SampsonCrowley:
### Reasons for making this change

I really would like to be able to use the almost stock react-jsonschema-form, but there just needs to be a way to play nice with bootstrap. 

since we already lean more towards bootstrap than material, I added a custom "object" option in the schema to produce bootstrap rows inside the normal object fieldset

### Checklist

* [X] **I'm updating documentation**
  - [X] I've checked the rendering of the Markdown text I've added
  - [X] If I'm adding a new section, I've updated the Table of Content
* [X] **I'm adding or updating code**
  - [X] I've added and/or updated tests
  - [X] I've updated docs if needed
  - [X] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style
* [X] **I'm adding a new feature**
  - [X] I've updated the playground with an example use of the feature

Issue #268, opened 2016-07-07
Make form group collapsable


zengfenfei:
Since my schema has nested objects and is kind of large, it will be great help to collapse some of objects.

n1k0:
This is definitely something we want indeed.

n1k0:
As often, the hardest part is the UX here. I came with this for now, but am wondering how the UI could expose something convenient and not intrusive. Thoughts?

![](http://i.imgur.com/T57gvlu.png)

amarnus:
I feel that it will be better to keep this functionality outside this module and instead support widgets for Array and Object fields in a better way so that people can build this functionality by themselves as a widget. Collapsing is just one option. You can also show as Tabs, for instance.

n1k0:
Very true, and this lib does a lot already so we shouldn't bloat it up with more features most people won't ever need.

I'll look into #225 to see where we're at to have custom object and array widgets.

n1k0:
A field template system as described in https://github.com/mozilla-services/react-jsonschema-form/issues/275#issuecomment-231948348 would indeed expose a convenient API to implement this. Let's do this.

nuclearspike:
Being able to set a wrapping component of a fieldset would solve this. Then you could use tabs or collapsible sections, columns, or anything you wanted. I'd want to be able to define the wrapping component then an object that contains all the props to be passed to it (for tab titles, etc). I don't know that the field template system would be enough since isn't that only at the field level? Would we need a fieldset template?

n1k0:
> Would we need a fieldset template?

Yeah this is a good idea, please comment on #281 about it.

n1k0:
Now #304 has landed we can safely close this.

Issue #573, opened 2017-05-06
Custom array widget.


asaf:
Hey,

I would like to render a custom component such "multi select component) for a specific array type,
The component would have to know how to generate the  JSON portion for that array,

Does it make sense to skip the complex ArrayField logic and just render a widget in case `ui:widget` is set on the array level? other alternatives?

Thanks.

glasserc:
The current ArrayField implementation doesn't support a custom widget. As far as I can tell from the documentation, widgets are mainly intended for single fields, and arrays are much more complicated than that. It's possible to override ArrayField with your own custom field type using `ui:field`, and I think that's probably the best choice considering how complicated ArrayField is and how many things it supports.

asaf:
@glasserc Thanks, I managed by adding a custom ui:field.

Closing.

danrot:
Is there no plan to support this? I think using a field instead is a bit tedious, since it doesn't show a label by default... And maybe it also skips some other functionality?

glasserc:
Sorry, I don't really understand what support is missing. We don't have any plans to support overriding an "array widget" because handling arrays requires a certain amount of complexity. Could you elaborate on what you have in mind?

danrot:
If you don't plan to support it, that's fine. I just think it's a little bit cumbersome to already implement a custom field for that, because certain features like using a label are then harder than necessary. But I have already decided to solve that in another way anyway.

asaf:
Yea, multi values are tricky as their characteristic vary based on their visual layouts, when it comes to arrays we usually found ourself overriding the entire Field in different ways.

glasserc:
I should also mention that you can override how an array field is rendered by using the ArrayFieldTemplate option described [here](https://github.com/mozilla-services/react-jsonschema-form#array-field-template). Depending on what "Custom array widget" means, that might or might not be suitable for your needs.

PR #653, opened 2017-08-01
Add ObjectFieldTemplate

olzraiti:

`ObjectFieldTemplate` simplifies customizing how the fields are rendered, eg. creating a bootstrap grid layout or a flexbox layout is easier. Previously we had to override ObjectField to achieve custom layouts - like `FieldTemplate` and `ArrayFieldTemplate`, we let the users worry only about rendering. Admittedly,  there isn't that much logic behind `ObjectField`, but this pull requests adds also consistency and makes it possible to customize titles & descriptions without using `TitleField` and `DescriptionField` (`TitleField` and `DescriptionField` could be even deprecated at some point in order to make the code base more consistent - discussed in #354 also).

EDIT: These issues/pull requests would be solved by this: #623 #268  

### Checklist


olzraiti:
Would this be a more powerful feature, if we didn't pass `properties` to the template, but `renderProperty(propertyName)`? It could add some easy optimization for example if there is a large number of properties, but only few of them would be shown at a time? Would this just add unnecessary complexity?

I was thinking also about allowing the template to handle constructing the `SchemaField` for each property, but this would kind of be against the idea of a template, though I'd personally have some use for this with a component where I don't want to render labels for the properties.

README.md
```diff
@@ -870,6 +871,51 @@ The following props are part of each element in `items`:
...
+        ObjectFieldTemplate={ObjectFieldTemplate} />,
+), document.getElementById("app"));
+```
+
+Please see [customObject.js](https://github.com/mozilla-services/react-jsonschema-form/blob/master/playground/samples/customObject.js) for a better example.
+
+The following props are passed to each `ObjectFieldTemplate`:
+
+- `DescriptionField`: The generated `DescriptionField` (if you wanted to utilize it)
+- `TitleField`: The generated `TitleField` (if you wanted to utilize it).
```
glasserc: I'm kind of ambivalent about passing these field components to the template. Client code already has the opportunity to override the entire template, using any `DescriptionField` and `TitleField` definition they want, and they can re-import DescriptionField and TitleField themselves if they want them. So what's the value?

I don't think `generated` is the right word here -- it's the one taken from the registry.

How about changing `if you wanted to` to `in case you wanted to`?

README.md
```diff
@@ -870,6 +871,51 @@ The following props are part of each element in `items`:
...
+- `description`: A string value containing the description for the object.
+- `properties`: An array of object representing the properties in the array. Each of the properties represent a child with properties described below.
+- `required`: A boolean value stating if the object is required.
+- `schema`: The schema object for this object.
+- `uiSchema`: The uiSchema object for this object field.
+- `idSchema`: An object containing the id for this object & ids for it's properties.
+- `formData`: The form data for the object.
+- `formContext`: The `formContext` object that you passed to Form.
+
+The following props are part of each element in `items`:
```
glasserc: Is `items` here a typo for `properties`?

README.md
```diff
@@ -870,6 +871,51 @@ The following props are part of each element in `items`:
...
+
+- `children`: The html for the property's content.
+- `name`: A string representing the property name.
+- `disabled`: A boolean value stating if the object property is disabled.
+- `index`: A number stating the index the property occurs in the `properties`.
+- `readonly`: A boolean value stating if the property is readonly.
```
glasserc: Missing blank line.

README.md
```diff
@@ -870,6 +871,51 @@ The following props are part of each element in `items`:
...
+- `idSchema`: An object containing the id for this object & ids for it's properties.
+- `formData`: The form data for the object.
+- `formContext`: The `formContext` object that you passed to Form.
+
+The following props are part of each element in `items`:
+
+- `children`: The html for the property's content.
+- `name`: A string representing the property name.
+- `disabled`: A boolean value stating if the object property is disabled.
+- `index`: A number stating the index the property occurs in the `properties`.
```
glasserc: I don't think we need to expose this, because it should be the same as the index in the `children` array, right?

test/ObjectFieldTemplate_test.js
```diff
@@ -0,0 +1,77 @@
...
+        properties,
+        title,
+        description,
+      } = this.props;
+      return (
+        <div className="root">
+          <TitleField title={title} />
+          <DescriptionField description={description} />
+          <div>
+            {properties.map(({ children, idx }) => (
```
glasserc: Is `idx` even going to be valid here?

README.md
```diff
@@ -870,6 +871,51 @@ The following props are part of each element in `items`:
...
+- `required`: A boolean value stating if the object is required.
+- `schema`: The schema object for this object.
+- `uiSchema`: The uiSchema object for this object field.
+- `idSchema`: An object containing the id for this object & ids for it's properties.
+- `formData`: The form data for the object.
+- `formContext`: The `formContext` object that you passed to Form.
+
+The following props are part of each element in `items`:
+
+- `children`: The html for the property's content.
```
glasserc: I don't like `children` as the name for this property. How about `content`, or `rendered`?

README.md
olzraiti: I agree that passing `DescriptionField` and `TitleField` is useless. However, `ArrayField` also passes them to `ArrayFieldTemplate`, so at least we are consistenly useless :)

I copy pasted the descriptions from `ArrayFieldTemplate` manual entry. I changed the descriptions in both manual entries now to "The `DescriptionField/TitleField` from the registry (in case you wanted to utilize it)"

olzraiti:
> My main remaining concern is about whether this goes far enough, and whether the approach it takes can be applied to arrays as well as objects. But I don't see any reason why arrays would be much different, and I'm comfortable merging this and leaving arrays for later.

Hmm, I'm not sure I understand what you mean. The templating approach is taken directly from how `ArrayField` handles `ArrayFieldTemplate` and the implementation is very similar.

glasserc:
> Hmm, I'm not sure I understand what you mean. The templating approach is taken directly from how ArrayField handles ArrayFieldTemplate and the implementation is very similar.

Oops. Somehow I had in my mind the idea that we were less flexible about arrays because of the question in #573. However, looking back at it, I now see that this is about ArrayWidgets, not ArrayFields or their templates, and I had confused myself. OK, great! Let me look at it again.

README.md
```diff
@@ -870,6 +871,52 @@ The following props are part of each element in `items`:
...
+- `idSchema`: An object containing the id for this object & ids for it's properties.
+- `formData`: The form data for the object.
+- `formContext`: The `formContext` object that you passed to Form.
+
+The following props are part of each element in `properties`:
+
+- `content`: The html for the property's content.
+- `name`: A string representing the property name.
+- `disabled`: A boolean value stating if the object property is disabled.
+- `index`: A number stating the index the property occurs in the `properties`.
```
glasserc: I still don't really understand why we need to expose this. Knowing that it comes from `ArrayField`, I can understand why it's valuable there -- but here it seems like the property name is more valuable.

olzraiti: I missed this one... You're right, I removed it.

glasserc:
Thanks!

PR #658, opened 2017-08-04
Pass more props to ErrorList

olzraiti:

`formContext` should be passed to every component that renders something for proper on-the-fly localization, styling etc.

EDIT: I passed also `errorSchema`, `schema` and `uiSchema` since they allow you to create a much more powerful & integrated component (you can pick labels from the `schema`, styles & root id from `uiSchema` to create links to the HTML id anchors where the errors are etc).

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -899,6 +899,7 @@ render((
 The following props are passed to `ErrorList`
 
 - `errors`: An array of the errors.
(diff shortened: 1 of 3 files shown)


n1k0:
Thank you!

glasserc:
Released in v0.50.0.

Issue #616, opened 2017-06-19
Support dynamic field type in arrays based on array element field value


mfulton26:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I would like to be able to have the type of a field be different depending the value of another field. I have seen discussion of similar problems where the solution involved dynamically swapping out the JSON Schema for the form but this doesn't appear to work if I have an array of objects where each one can have a different value and therefore type. Please advise. Thank you.

### Details

[...]

Issue #522, opened 2017-03-23
Support dynamic presence of properties


mfulton26:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I would like a way to define a schema where some properties should be present if and only if another property has some value(s) and not present otherwise.

### Steps to Reproduce

1. Define a JSON Schema defining properties that sometimes should be present and sometimes should not be (e.g. using dependencies with oneOf, not, and required).
2. Render the schema on a react-jsonschema-form and attempt and try it out.

https://jsfiddle.net/07fm4n1b/

#### Expected behavior

react-jsonschema-form reacts to the "oneOf" array in the "dependencies" and intelligently displays only the relevant properties and when values change and some properties become relevant or no longer relevant they are shown or hidden appropriately.

#### Actual behavior

All of the properties are displayed requiring the user to understand which properties they need to fill and which ones they must leave blank.

### Version

0.44.0

n1k0:
That seems very complicated to implement properly and efficiently, no mention of maintainability. We usually recommend to process the schema when `onChange` is triggered to contextually augment the initial one with new definitions. This has worked great for us so far, so I'm having a hard time justifying spending a possibly huge amount of time implementing what you suggest.

I'm closing for now, though if you're willing to work on this, feel free to reopen.

misto:
I've recently implemented something like this using a custom field template. I've added additional entries to the ```uiSchema```:
~~~
{
  "ConditionallyEnabledfield": {
    "ui:options": {
      "enabledIf": "SomeOtherField",
      "hasValue": "yes"
    }
}
~~~

and then check them in the custom field rendering:
~~~
export class CustomFieldTemplate extends React.PureComponent {

  render() {

    const {id, label, displayLabel, rawHelp, description, rawErrors, children, schema, uiSchema} = this.props

    let childrenProps = children.props
    let disabled = false

    if(schema && schema.type === "object") {
      const { formData, uiSchema } = children.props      
      childrenProps = {
        ...childrenProps,
        uiSchema: Object.keys(uiSchema).reduce((newUiSchema, key) => {          
          const uiOptions = (uiSchema[key] || {})['ui:options']
          if(uiOptions
            && uiOptions.enabledIf
            && uiOptions.hasValue
[...]

Issue #430, opened 2016-12-15
Is there a way to have conditionnal elements?


mikaoelitiana:
### Prerequisites

- [X] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I want to add conditions in my forms so that some fields will be displayed depending on value of other fields.

#### Expected behavior

When I create a form with conditions, all fields are not displayed yet. When I enter or select an input value, the condition is evaluated to see if some other field should be enable or disabled

exist3nz:
As a suggestion, the [JSON Schema Dependencies](https://spacetelescope.github.io/understanding-json-schema/reference/object.html?highlight=dependencies) provides examples on how that would be achieved at the schema level and it's treating the case you are also referring to in the section Schema dependencies. I think that would be a good starting point. I'm currently also trying to figure out if the JSON Schema Form supports it.

n1k0:
This should be achieved by computing a new form schema from within the `onSubmit` handler, according to received values; that's certainly the best place to write your own conditional logic, using plain old js. This has been already discussed in issues, and I've provided examples how to do that.

I don't have the time for searching these atm, but if you find it back it'd be great to add an entry in the FAQ section of the README, as this question gets often asked.

mikaoelitiana:
Thanks @exist3nz and @n1k0

inostia:
For anyone else wondering about this, here's a working example: https://jsfiddle.net/69z2wepo/68258/

n1k0:
@mikaoelitiana thanks, it's a great example. Would you mind contributing it to the README? :)

biskit:
this example listed above may not be helpful as it steps outside the definition and requires conditional coding. if i have tens of conditional elements this does not scale. a better answe would be dependencies in the schema as suggested above.

glasserc:
rjsf doesn't support schema or property dependencies and at present we don't have any specific plans to implement it. (Nobody has even opened an issue about it.) Another approach is to implement `anyOf` as described in #52. But I think Niko's point in the above comments is that in React, we have other, more powerful tools available to us, such as updating the rjsf props to have a different schema according to the component's current state.

biskit:
understood, it's doable with VDOM. it's about being able to define in schema that is of concern to me. this way, all the characteristics are captured in it.

[...]

PR #659, opened 2017-08-05
Initial support for property & schema dependencies

mfulton26:

JSON Schema supports property and schema dependencies which allow changing the schema for an object (even a single object within an array of objects) based on the presence or value of certain properties.

This change brings initial support for property dependencies and schema dependencies which allows developers to define conditionally required fields, conditionally present fields, and dynamic schemas (e.g. changing a property type based on the value of another field).

See #616 and #430 (see also #522).

### Checklist

This is a work in progress but I am looking for some initial feedback as I have not contributed to this project before and I am looking for early feedback on the implementation, etc.


ksjogo:
Using that in my project now. Works great. Happy to provide any needed help for merging this.

faassen:
I've seen it argued in other issues such as https://github.com/mozilla-services/react-jsonschema-form/issues/430 that instead of using schema dependencies you can dynamically change the schema when data changes. This is indeed very powerful. But I've added a note there with some use cases for dependencies.

playground/samples/propertyDependencies.js
```diff
@@ -0,0 +1,86 @@
...
+    title: "Property dependencies",
+    description: "These samples are best viewed without live validation.",
+    type: "object",
+    properties: {
+      unidirectional: {
+        title: "Unidirectional",
+        src:
+          "https://spacetelescope.github.io/understanding-json-schema/reference/object.html#dependencies",
+        description:
+          "In the following example, whenever a `credit_card` property is provided, a `billing_address` property must also be present.",
```
n1k0: Markdown is not supported, so backticks won't turn into `<code>` blocks, though you could use JSX instead of a string here.

src/components/fields/ArrayField.js
```diff
@@ -312,14 +312,24 @@ class ArrayField extends Component {
     const arrayProps = {
       canAdd: this.canAddItem(formData),
       items: formData.map((item, index) => {
+        const itemSchema = retrieveSchema(
+          schema.items,
+          definitions,
+          formData[index]
```
n1k0: Nit: here and in some other places, this could be just `item`.

src/utils.js
```diff
@@ -379,17 +380,82 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+    const $refSchema = findSchemaDefinition(schema.$ref, definitions);
+    // Drop the $ref property of the source schema.
+    const { $ref, ...localSchema } = schema;
+    // Update referenced schema definition with local schema properties.
+    return retrieveSchema(
+      { ...$refSchema, ...localSchema },
+      definitions,
+      formData
+    );
+  } else if (schema.hasOwnProperty("dependencies")) {
```
n1k0: How about extracting dependency handling logic into its own dedicated function? This whole `retrieveSchema` one is getting huge.

src/utils.js
```diff
@@ -379,17 +380,82 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+    // Drop the dependencies from the source schema.
+    let { dependencies, ...localSchema } = schema;
+    // Process dependencies updating the local schema properties as appropriate.
+    for (const key of Object.keys(dependencies)) {
```
n1k0: Nit: why not just `for (const key in dependencies)`?

src/utils.js
```diff
@@ -379,17 +380,82 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+        }
+      } else if (isObject(value)) {
+        localSchema = retrieveSchema(
+          mergeObjects(
+            localSchema,
+            retrieveSchema(value, definitions, formData),
+            true
+          )
+        );
+        if (Array.isArray(value.oneOf)) {
```
n1k0: Hmm so this introduces partial support for `oneOf` while fields are still not handled at all, I'd have preferred introducing this after we get proper support for oneOf.

src/utils.js
```diff
@@ -379,17 +380,82 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+            if (conditionProperty) {
+              const conditionSchema = {
+                type: "object",
+                properties: {
+                  [key]: conditionProperty,
+                },
+              };
+              const { errors } = jsonValidate(formData, conditionSchema);
+              if (errors.length === 0) {
+                const dependentSchema = { ...subschema };
```
n1k0: Beware as cloning this way isn't recursive.

src/utils.js
```diff
@@ -379,17 +380,82 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+    );
+  } else if (schema.hasOwnProperty("dependencies")) {
+    // Drop the dependencies from the source schema.
+    let { dependencies, ...localSchema } = schema;
+    // Process dependencies updating the local schema properties as appropriate.
+    for (const key of Object.keys(dependencies)) {
+      // Skip this dependency if its trigger property is not present.
+      if (!formData[key]) {
+        // fixme: a falsey check may not be appropriate here, I'm not sure
+        continue;
```
n1k0: Yeah, maybe just check for undefined I think.

test/utils_test.js
```diff
@@ -496,6 +496,139 @@ describe("utils", () => {
...
+            type: "object",
+            properties: {
+              a: { type: "string" },
+              b: { type: "integer" },
+            },
+            required: ["a"],
+          });
+        });
+      });
+      describe("true condition", () => {
```
n1k0: Nit: empty line between describe statements.

playground/samples/propertyDependencies.js
mfulton26: Oh interesting, I didn't think of that. I'll take a look. Thank you.

src/components/fields/ArrayField.js
mfulton26: Good point, I felt like I was "hacking" a bit as I am still familiarizing myself with the code base. Thank you for the feedback. I will update.

src/utils.js
mfulton26: I was just thinking of this this morning when I started thinking about this pull request again due to others commenting on the pull request. I will see if I can have some updates up today or tomorrow. Thank you.

src/utils.js
mfulton26: Because I'm an EcmaScript noob? 😄

src/utils.js
mfulton26: Hmm, I'll try that but I may already have done so. If a field is left empty does it come in as `undefined` in the `formData` or as an empty string `""`? I guess I will find out. :-)

src/utils.js
mfulton26: Yeah, I'm still not certain how `oneOf` would work in general but I kind of view the usage here as special. The idea is that depending on values in `formData` the overall schema will dynamically update to one of several options but this still doesn't support a user selecting one of several schemas to interact with as one would need if using `oneOf` elsewhere in a schema. I have started updating README.md locally and I am planning on explaining just this as implementing support for this kind of `oneOf` is very independent of general support for `oneOf` (and vice versa I believe although I could be wrong).

src/utils.js
mfulton26: Correct. I am just trying to not mutate the incoming schema at all so I make copies as needed in the tree instead of recursively coping the entire tree (although I suppose the later could be done for readability/maintainability reasons although I'm not sure how to do that off the top of my head).

test/utils_test.js
mfulton26: It can be done! 😄

mfulton26: FYI: I appreciate the "Nit" comments. Every little bit counts.

src/utils.js
n1k0: AFAICR it depends if the field value has been edited already or not; by default a schema string field value is set to undefined (property is not preset), but as soon as users manipulate the widget, eg. typing then erasing the entered value, it's set to an empty string. So yeah, it might be a little tricky.

src/utils.js
n1k0: Right, we may just want to document this specific behavior here.

src/utils.js
n1k0: Deep copying objects is hard, Json serial/deserialization is know to be [quite efficient](https://stackoverflow.com/a/5344074/330911) but maybe not enough here as this operation may be executed on each keystroke in a field with this lib.

mfulton26:
@glasserc I took a quick glance at your input and I totally agree I've just been pretty busy this week with other things. I should be able to make some updates soon though. Thanks!

src/utils.js
```diff
@@ -393,71 +393,104 @@ export function retrieveSchema(schema, definitions = {}, formData = {}) {
...
+    return retrieveSchema(resolvedSchema, definitions, formData);
   } else {
     // No $ref or dependencies attribute found, returning the original schema.
     return schema;
   }
 }
 
+function resolveDependencies(schema, definitions, formData) {
+  // Drop the dependencies from the source schema.
+  let { dependencies, ...resolvedSchema } = schema;
```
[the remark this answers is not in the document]
mfulton26: Sounds good to me.

src/utils.js
```diff
@@ -393,71 +393,104 @@ export function retrieveSchema(schema, definitions = {}, formData = {}) {
...
+  // Process dependencies updating the local schema properties as appropriate.
+  for (const key of Object.keys(dependencies)) {
```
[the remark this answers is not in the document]
mfulton26: Yep, I missed that when I moved things around before.

src/utils.js
```diff
@@ -393,71 +393,104 @@ export function retrieveSchema(schema, definitions = {}, formData = {}) {
...
+  }
+  return resolvedSchema;
+}
+
+function resolvePropertyDependency(schema, properties) {
+  if (!properties) {
+    return schema;
+  }
+  schema = { ...schema };
+  schema.required = Array.isArray(schema.required) ? [...schema.required] : [];
```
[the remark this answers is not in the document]
mfulton26: I think using a `Set` makes for a good solution. I'll try that for now.

src/utils.js
```diff
@@ -393,71 +393,104 @@ export function retrieveSchema(schema, definitions = {}, formData = {}) {
...
+  for (const property of properties) {
+    if (schema.required.includes(property)) {
+      continue;
+    }
+    schema.required.push(property);
+  }
```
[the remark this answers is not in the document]
mfulton26: With the `Set` implementation this actually goes away. 👍

src/utils.js
```diff
@@ -379,17 +380,115 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+        formData,
+        key,
+        value
+      );
+    }
+  }
+  return resolvedSchema;
+}
+
+function resolvePropertyDependency(schema, properties) {
```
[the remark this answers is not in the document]
mfulton26: I don't like it either but I couldn't think of something better at the time. :smile:

I'm actually wonder if `withDependentProperties` would make sense. That way it is clear that this function returns a new schema object with the additional dependent properties (in the required property which is how property dependencies work).

Thoughts?

src/utils.js
```diff
@@ -379,17 +380,115 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+  let { oneOf, ...resolvedValue } = value;
+  schema = resolveConditionalSchema(
+    schema,
+    definitions,
+    formData,
+    resolvedValue
+  );
+  if (Array.isArray(oneOf)) {
+    schema = resolveDynamicSchema(schema, definitions, formData, key, oneOf);
+  }
```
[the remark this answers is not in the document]
mfulton26: Hmm, I agree. I haven't looked yet throughout this project. Do we ever display to the user that they have created a bad JSON schema? Or do we log warnings for this kind of thing in the console?

src/utils.js
```diff
@@ -379,17 +380,115 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+  return schema;
+}
+
+function resolveConditionalSchema(schema, definitions, formData, value) {
+  return retrieveSchema(
+    mergeObjects(schema, retrieveSchema(value, definitions, formData), true)
+  );
```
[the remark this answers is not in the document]
mfulton26: It might not be... I'l do some testing and see. I think I was too liberal with calling `retrieveSchema`

src/utils.js
```diff
@@ -379,17 +380,115 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+    formData,
+    resolvedValue
+  );
+  if (Array.isArray(oneOf)) {
+    schema = resolveDynamicSchema(schema, definitions, formData, key, oneOf);
+  }
+  return schema;
+}
+
+function resolveConditionalSchema(schema, definitions, formData, value) {
```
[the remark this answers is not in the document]
mfulton26: Agreed.

src/utils.js
```diff
@@ -379,17 +380,115 @@ function findSchemaDefinition($ref, definitions = {}) {
...
+        },
+      };
+      const { errors } = jsonValidate(formData, conditionSchema);
+      if (errors.length === 0) {
+        const dependentSchema = { ...subschema };
+        dependentSchema.properties = dependentProperties;
+        schema = mergeObjects(
+          schema,
+          retrieveSchema(dependentSchema, definitions, formData)
+        );
```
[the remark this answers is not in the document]
mfulton26: This probably actually needs to filter all of the subschemas and confirm that only one is valid. I will update.

mfulton26:
Things I still need to do:

- [ ] update README.md to explain special `oneOf` usage in `dependencies` only
- [ ] address "Warning: Failed prop type: The prop `id` is marked as required in `BaseInput`, but its value is `undefined`."
- [ ] re-review previous comments and code and see what else can be cleaned up

Thank you for the feedback, please feel free to reiterate and/or provide additional feedback as some of the decisions I've made in an attempt to "improve" things might not have been the best decisions. 😄

mfulton26:
Hmmm, I'm not sure why the build is working but I'm not going to look at this again until at least tomorrow so it can wait until at least then.

mfulton26:
Okay, this appears ready to be reviewed again.

src/utils.js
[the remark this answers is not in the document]
glasserc: I've changed this to throw, which we appear to do for some schemas that are too dysfunctional to work. (For example, `getDefaultFormState` throws on non-object schemas.)

src/utils.js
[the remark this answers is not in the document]
glasserc: I've added some tests that I think justify the need for `retrieveSchema`, as well as demonstrate a minor bug in this implementation, which I fixed.

src/utils.js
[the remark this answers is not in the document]
glasserc: (Oops, I posted this comment on the wrong line before...)

I added a test justifying the need for a call to `retrieveSchema`, plus fixed a minor bug that such a test uncovered.

glasserc:
I took the liberty of adding a little bit of documentation to the README explaining in the vaguest possible terms how this works and fixed up a few stylistic nits that bothered me. The tests seem to pass, so I'm merging it. Thanks so much for your work on this!

src/utils.js
```diff
@@ -422,7 +422,7 @@ function resolveDependencies(schema, definitions, formData) {
   // Process dependencies updating the local schema properties as appropriate.
   for (const dependencyKey in dependencies) {
     // Skip this dependency if its trigger property is not present.
-    if (formData[dependencyKey] === undefined) {
+    if (!formData.hasOwnProperty(dependencyKey)) {
```
mfulton26: @glasserc Unfortunately it looks like this change broke properties and schema dependencies. With the change conditionally required properties are always marked as required (with an asterisk (*)) and conditionally declared properties are always shown. See the playground for the examples.

Note that the dynamic schema still works correctly with the specific `oneOf` case.

mfulton26: I'm surprised this passed the tests... I must not have written very good ones. 😄

mfulton26:
Awesome. I'm happy you've made the additions. This is really great and thank you all for supporting the change!

mfulton26:
I think the hasOwnProperty change may have caused a hiccup though. Please see my comment on your commit. Thank you.

src/utils.js
glasserc: Yeah, that's unfortunate.. the way we `computeDefaults` fills missing fields with `undefined`s. I don't like that design but it's outside the scope of this PR. Thanks for letting me know that I broke it :)

mfulton26:
Just out of curiosity, how soon might we be able to get the latest version deployed to the playground? Thanks!

cdtinguria:
Guys, very nice feature to have. Any idea when this will be released?

mfulton26:
@glasserc This appears to be published to npm. I'm curious how we get the playground updated. Do you know? Thanks.

glasserc:
Sorry, I forgot to push it to the playground. It should be up now.

mfulton26:
Thanks!

PR #660, opened 2017-08-09
Remove status from Form state (BUG FIX)

olzraiti:

This fixes a bug that causes the error list to flicker when formData is edited. 

**To reproduce:** go to the "Errors" tab on the playground, select an input on the Form and start smashing the keyboard (you don't really even have to smash the keyboard, you can see the effect on any single form data change). The error list will flicker.

First I thought the bug was caused by some recent commits regarding the Error List, but actually the flickering is caused by the `status` state property, which will flip between `initial` and `editing` during typing.

The `status` property has been there from the very first commit, but I can't find any real use for it other than causing this bug. I don't understand why we wouldn't show the error list when the Form is edited. Even if there was some kind of buffering that would hide the error list only when the formData is updated in intervals less time than some milliseconds, it would affect the page scrolling position before and after the user stops typing. The only side effect that removing `status` causes that I can think of is that when  `onSubmit` calls `this.props.onSubmit(this.state)`, the `state` won't contain `status: "submitted"` - which seems kind of pointless to me :)

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -51,7 +51,6 @@ export default class Form extends Component {
       definitions
     );
     return {
-      status: "initial",
       schema,
       uiSchema,
       idSchema,
@@ -77,10 +76,10 @@ export default class Form extends Component {
   }
 
   renderErrors() {
-    const { status, errors, errorSchema, schema, uiSchema } = this.state;
+    const { errors, errorSchema, schema, uiSchema } = this.state;
     const { ErrorList, showErrorList, formContext } = this.props;
 
-    if (status !== "editing" && errors.length && showErrorList != false) {
+    if (errors.length && showErrorList != false) {
       return (
         <ErrorList
           errors={errors}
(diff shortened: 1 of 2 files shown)


n1k0:
The initial purpose for having a status was to distinguish between a pristine form and one that have been edited already. I agree this was never really used by our team, so I'm not against removing it entirely, though we should ensure mentioning it in the next release changelog because it's a breaking change of the public API.

olzraiti:
Hmm, okay. We could keep that functionality by adding `status: "submitted"` to every `onSubmit` propagation.

So we would change this part:
```
 if (this.props.onSubmit) {
  this.props.onSubmit(this.state);
}
```

To this: 
```
 if (this.props.onSubmit) {
  this.props.onSubmit({...this.state, status: "submitted"});
}
```

This way, we wouldn't have any breaking changes to the public API.

Either that, or just remove the functionality. Which would you prefer? :)

n1k0:
Yeah, keeping BC while fixing bugs always sounds like a great plan to me

cdtinguria:
Guys, when will this fix be released?

n1k0:
Probably next week

glasserc:
Released in v0.50.0.

PR #661, opened 2017-08-09
Fix minItems bug with multiSelect and formData

### LINKED ISSUE #534: Minimum Items validation in multiple choice inputs is not working
### Prerequisites

- [X] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);

### Description

I'm using an array with "ui:widget": "checkboxes" to render a checkbox group in a form, and minItems=1 to validate that the user has selected at least one item. 
When a form is first rendered, we have no formData, so this is the initial state is undefined formData for the field, not an empty array.
The minItems is causing an extra 'undefined' entry to be added to the formData, which means that the validation does not work, as this entry stays here no matter what the user selects.

### Steps to Reproduce

This can be seen easily in the playground, under Arrays. Add "minItems": 1 to the multipleChoicesList section, and then delete the whole entry for multipleChoicesList from the formData section. 

Now click one of the checkboxes in the live demo and see what happens to the formData. A null entry is added after the selected values. Make it "minItems": 2, delete the formData entry and try again, and 2 null entries are added. This means that the minimum items validation does not work.

I thought that I could set a default empty array for the field, by adding default: [] in the schema, but this changes the nulls to empty arrays.

#### Expected behavior

When the user has no items selected, a minimum items validation error message should be shown.

#### Actual behavior

The minimum items error is not shown.

### Version

0.44.0

spacebaboon:

Check to see whether the field is a multiselect (select or checkboxes widget) before adding extra items up to minLength in the defaults.
This was causing extra, invalid items to be added into formData and breaking the minItems validation when no formData is initially present.

Fixes #534 

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -140,9 +140,13 @@ function computeDefaults(schema, parentDefaults, definitions = {}) {
 
     case "array":
       if (schema.minItems) {
-        return new Array(schema.minItems).fill(
-          computeDefaults(schema.items, defaults, definitions)
-        );
+        if (!isMultiSelect(schema, definitions)) {
+          return new Array(schema.minItems).fill(
+            computeDefaults(schema.items, defaults, definitions)
+          );
+        } else {
+          return [];
+        }
       }
   }
   return defaults;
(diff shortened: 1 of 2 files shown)


glasserc:
Released in v0.50.0.

Issue #611, opened 2017-06-14
Unsupported field schema


gastonche:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);

### Description

I am trying to generate forms for a schema onthe playground but i keep getting an error, and I am unable to solve the error. The schema is tested valid on jsonschemalint.com the error i get is 
```javascript
Unsupported field schema { "definitions": { "address": { "type": "object", "properties": { "street_address": { "type": "string" }, "city": { "type": "string" }, "state": { "type": "string" } }, "required": [ "street_address", "city", "state" ] } }, "allOf": [ { "$ref": "#/definitions/address" }, { "properties": { "type": { "enum": [ "residential", "business" ] } } } ] }.
```

the json I am trying to use is 
```javascript
{
  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": {
          "type": "string"
        },
        "city": {
          "type": "string"
        },
        "state": {
          "type": "string"
        }
      },
      "required": [
        "street_address",
        "city",
        "state"
      ]
    }
  },
  "allOf": [
    {
      "$ref": "#/definitions/address"
    },
    {
      "properties": {
        "type": {
          "enum": [
            "residential",
            "business"
          ]
        }
      }
    }
  ]
}
```
I don't understand what is wrong. I intend to use this library for my project
### Steps to Reproduce

1. visit [https://mozilla-services.github.io/react-jsonschema-form/](https://mozilla-services.github.io/react-jsonschema-form/)
2.enter the obove schema in place of the json schema
3.replace the ui-schema and formData contents with { }

[...]

Issue #66, opened 2016-03-15
Enums don't require a type definition


magopian:
(splitted out from #25)

Example in http://json-schema.org/example2.html shows an `enum` field doesn't seem to require a `type` property defined:

``` js
"fstype": {
  "enum": ["ext3", "ext4", "btrfs"]
},
```

leplatrem:
In that case you assume `"string"`, but should we introspect and guess the type from enum values?

n1k0:
> should we introspect and guess the type from enum values?

And how about 

```js
"fstype": {
  "enum": ["str", 2, {foo: [1, 2]}]
},
```

Oh my.

n1k0:
> http://json-schema.org/latest/json-schema-validation.html#anchor76
> The value of this keyword MUST be an array. This array MUST have at least one element. 
> Elements in the array MUST be unique.
>
> Elements in the array MAY be of any type, including null.

epicfaace:
This was fixed by #668

PR #668, opened 2017-08-15
66 enum no type

penx:

JSON Schema states enums can be used without defining type => default type to string when not specified.

See #66 #611 

### Checklist


src/utils.js
```diff
@@ -62,7 +62,10 @@ export function getDefaultRegistry() {
 }
 
 export function getWidget(schema, widget, registeredWidgets = {}) {
-  const { type } = schema;
+  let { type } = schema;
+  if (!type && schema.enum) {
+    type = "string";
+  }
```
n1k0: I'd love to see this duplicated logic extracted to some `getSchemaType` helper function, so we could reuse it in SchemaFiled.js.

edi9999: +1 on this one, can you do that @penx so that we can merge this ?

penx: done

PR #673, opened 2017-08-18
Add errorSchema to onError prop callback.

theengineear:
Any qualms about exposing the `errorSchema` structure to a wrapping component in the `onError` method?

I'm using CSS-in-JS and looking to access the structured information in errorSchema. The flat structure of the errors array (currently passed back in `onError`) is good for printing out errors, but the names are missing then nested structure of the `errorSchema`.

### Checklist

^^ Before I do these things, I'm hoping to get an initial 👍 / 👎

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -128,7 +128,7 @@ export default class Form extends Component {
       if (Object.keys(errors).length > 0) {
         setState(this, { errors, errorSchema }, () => {
           if (this.props.onError) {
-            this.props.onError(errors);
+            this.props.onError(errors, errorSchema);
           } else {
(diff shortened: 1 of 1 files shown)


n1k0:
Sure, we should document this though.

n1k0:
> Sure, we should document this though.

@theengineear Could you please take care of this? Thanks a lot.

epicfaace:
@theengineear it's been a while -- do you still need this feature?

theengineear:
@epicfaace , sorry for never getting back to document this. No I no longer need this. I think the use case is still valid. But we should close so that we don't add baggage to the codebase unnecessarily.

PR #680, opened 2017-08-25
Fix typo in README

cuyu:

There is a typo in README, the `maximum` is actually works instead of `maximium`

### Checklist


glasserc:
Released in v0.50.0.

Issue #675, opened 2017-08-22
Optional object with required fields in react-jsonschema-form


davidMcneil:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [X] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Given a json schema like the one below, the react-jsonschema-form validator essentially requires both shipping_address and billing_address even though the billing_address is not listed as required. This is because the address type requires all three of its properties. How can I make the billing_address optional?

    {
      "definitions": {
        "address": {
          "type": "object",
          "properties": {
            "street_address": {
              "type": "string"
            },
            "city": {
              "type": "string"
            },
            "state": {
              "type": "string"
            }
          },
          "required": [
            "street_address",
            "city",
            "state"
          ]
        }
      },
      "type": "object",
      "properties": {
        "billing_address": {
          "title": "Billing address",
          "$ref": "#/definitions/address"
        },
        "shipping_address": {
          "title": "Shipping address",
          "$ref": "#/definitions/address"
        }
      },
      "required": [
        "shipping_address"
      ]
    }
### Steps to Reproduce

[...]

Issue #331, opened 2016-09-29
Allow marking an object subschema as non-required


n1k0:
Sample schema example:

``` js
{
  type: "object",
  properties: {
    foo: {
      type: "object",
      required: ["bar"],
      properties: {
        bar: {type: "string"}
      }
    }
  }
}
```

Here we'd expect the validation logic to allow submitting the form if no value is entered for `bar`; but currently this raises a `required` validation error - as `foo` is *not* required itself. 

Fix: we could consider an object formData as not yet provided by the user as long as the matching `formData` object for this field has no keys. As soon as one property is entered for this object, we should mark the field value as _entered_ by the user, so it should be submitted.

epicfaace:
Closing this in favor of #675

PR #682, opened 2017-08-26
Fix optional object with required fields

davidMcneil:

The library would require an optional object if it had required fields. (#675)

Note, this change does not take into account the HTML5 required property. As such I added a "HTML5 Validation" checkbox to the example app to allow for turning off HTML5 validation.

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -210,6 +210,8 @@ export function isObject(thing) {
 }
 
 export function mergeObjects(obj1, obj2, concatArrays = false) {
+  obj1 = obj1 instanceof Object ? obj1 : {};
+  obj2 = obj2 instanceof Object ? obj2 : {};
   // Recursively merge deeply nested objects.
   var acc = Object.assign({}, obj1); // Prevent mutation of source object.
   return Object.keys(obj2).reduce((acc, key) => {


playground/app.js
```diff
@@ -386,7 +393,7 @@ class App extends Component {
     });
   };
 
-  setLiveValidate = ({ formData }) => this.setState({ liveValidate: formData });
+  setValidate = ({ formData }) => this.setState({ ...formData });
```
n1k0: Just call `setState(formData)` here. You may even want to try using `onChange={this.setState}` later.

src/utils.js
```diff
@@ -224,6 +226,44 @@ export function mergeObjects(obj1, obj2, concatArrays = false) {
...
+export function trimObject(obj) {
+  // Recursively trim an object converting objects whose values are all undefined to `undefined`.
+  let result = Object.assign({}, obj); // Prevent mutation of source object.
+  for (const key of Object.keys(result)) {
+    if (result.hasOwnProperty(key) && isObject(result[key])) {
+      result[key] = trimObject(result[key]);
+    }
+  }
+  if (objectNeedsTrimmed(result)) {
+    result = undefined;
```
n1k0: Nit: just `return;`

test/utils_test.js
```diff
@@ -431,6 +432,60 @@ describe("utils", () => {
       expect(mergeObjects(obj1, obj2)).eql(expected);
     });
 
+    describe("trimObject()", () => {
+      it("should't mutate the provided objects", () => {
```
n1k0: Nit: *shouldn't*

test/utils_test.js
```diff
@@ -431,6 +432,60 @@ describe("utils", () => {
...
+        const obj = { a: undefined };
+        trimObject(obj);
+        expect(obj).eql({ a: undefined });
+      });
+
+      it("should trim empty object", () => {
+        const obj = {};
+        const res = trimObject(obj);
+        expect(res).eql(undefined);
```
n1k0: Please ensure using `expect(res).to.be.a("undefined")` for better accuracy.

src/utils.js
```diff
@@ -224,6 +226,44 @@ export function mergeObjects(obj1, obj2, concatArrays = false) {
   }, acc);
 }
 
+function objectNeedsTrimmed(obj) {
```
n1k0: How about `objectNeedsTrimming`?

src/utils.js
```diff
@@ -224,6 +226,44 @@ export function mergeObjects(obj1, obj2, concatArrays = false) {
...
+  // Recursively determine if an object needs to be simplified to `undefined`.
+  return Object.keys(obj).every(key => {
```
n1k0: Shouldn't we use `some` instead of `every` here?

test/utils_test.js
```diff
@@ -431,6 +432,60 @@ describe("utils", () => {
...
+        const res = trimObject(obj);
+        expect(res).eql({ a: true, o1: undefined, o2: undefined });
+      });
+
+      it("shouldn't trim object with empty array field", () => {
+        const obj = { a: [] };
+        const res = trimObject(obj);
+        expect(res).eql({ a: [] });
+      });
+    });
```
n1k0: I'd like to see more functional tests added at the component level for this behavior.

src/utils.js
```diff
@@ -224,6 +226,44 @@ export function mergeObjects(obj1, obj2, concatArrays = false) {
...
+}
+
+export function trimObject(obj) {
+  // Recursively trim an object converting objects whose values are all undefined to `undefined`.
+  let result = Object.assign({}, obj); // Prevent mutation of source object.
+  for (const key of Object.keys(result)) {
+    if (result.hasOwnProperty(key) && isObject(result[key])) {
+      result[key] = trimObject(result[key]);
+    }
+  }
```
n1k0: We may want to benchmark this, but I wonder how much slower would be to just use `JSON.parse(JSON.stringify(obj))`. I've read that these fns were heavily optimized in modern js engines. That would give us:

- Deep cloning for free,
- Getting rid of the `objectNeedsTrimmed` function call.

glasserc:
I'll take a look in a second, but before I forget, I want to note that this is related to #542 (and through it, #465 and #331).

src/utils.js
davidMcneil: I probably needed to do a better job explaining this in the PR, but we need to convert objects with all of their fields set to `undefined` to simply `undefined` and do this recursively. For example, convert

`{ "a": true, "b": { "c": undefined, "d": { "e": undefined} } }` to `{"a": true, "b": undefined}`

The reason is, assuming we have a schema where `b` is optional but its fields are required, if `b` is set to an object the validator requires all of its required fields to be set. So we must "trim" down its unspecified fields and `b` itself to simply `undefined`. The `every` is requiring all of an objects fields are `undefined` to do a "trim". I do not believe setting it to `some` would work because the user would be unable to set the fields of an object (the entire object would keep being set to `undefined`) unless they were somehow able to simultaneously set all fields at once. Thanks for the feedback!

src/utils.js
```diff
@@ -168,15 +168,17 @@ export function getDefaultFormState(_schema, formData, definitions = {}) {
...
+    result = defaults;
+  } else if (isObject(formData)) {
     // Override schema defaults with form data.
-    return mergeObjects(defaults, formData);
+    result = mergeObjects(defaults, formData);
+  } else {
+    result = formData || defaults;
   }
-  return formData || defaults;
+  return trimFormData(result);
```
glasserc: This means that we trim defaults, which feels like it could be a bit surprising?

src/utils.js
davidMcneil: I am not sure simply calling `JSON.parse(JSON.stringify(obj))` would work I  tried it and it converts `{"a": true, "b": {"c": undefined, "d": {"e": undefined}}}` to `{ "a": true, "b": {"d": {}}}` and I believe we would need it to convert it to `{ "a": true, "b":undefined}`. That being said we could now simply remove empty objects and get the same behavior, but we still have to walk through the fields of the object recursively.

test/utils_test.js
```diff
@@ -431,6 +432,60 @@ describe("utils", () => {
...
+      it("shouldn't trim object with fields set", () => {
+        const obj = { a: true, b: undefined, c: undefined };
+        const res = trimObject(obj);
+        expect(res).eql({ a: true, b: undefined, c: undefined });
+      });
+
+      it("should trim field object", () => {
+        const obj = { a: true, o: { b: undefined } };
+        const res = trimObject(obj);
+        expect(res).eql({ a: true, o: undefined });
```
glasserc: This might be a bit pedantic, but why return an object with `undefined` s in it, and not e.g. `{a: true}`? (Of course, this would require changes to your implementation.

davidMcneil:
@glasserc You are absolutely right this does not handle the case where we have a required  object with all optional fields. With the schema below, assuming neither `field1` or `field2` are set, I believe all the relevant cases we need to handle are listed in the table.

```
{
  "type": "object",
  "properties": {
    "object": {
      "type": "object",
      "properties": {
        "field1": {
          "type": "string"
        },
        "field2": {
          "type": "string"
        }
      }
    }
  }
}
```

| `object` | `field1` and `field2` | submit |
|---|---|---|
| required | none required | `{object: {}}` |
| optional | none required | `{object: {}}` or `{}` ? | 
| required | some required | invalid |
| optional | some required | `{}` |

I would appreciate any feedback on the "correct/best" way to handle the second state, optional `object` and optional fields, as well as thoughts on how best to implement this fix. I do not believe it is possible to simply "trim" the object by solely looking at the current `formData` the way the current state of this PR does. Somehow we need to know whether or not a field is required. Is it acceptable to simply pass in the `schema` to `trimFormData` and walk the `schema` as I walk the `formData` and check if a field is required or not before trimming it? Or is there a better way to do this?

glasserc:
My only real concern with trimming entire objects is that the error messages make sense. If you have a required object with required fields, I think the error messages belong on the required fields ("phone number is a required field") and not the containing object ("user is a required field"). So for me, I'm most concerned about the 3rd case. As for the second case, I think a user might expect either one depending on the application. I think what we should do here is maintain backwards compatibility (so `{option: {}}`). If someone complains then maybe we can add an option. But I don't want to add too much complexity at once, especially if we aren't sure we need it.

I don't have a strong opinion on how to implement it. I guess you have to examine both the `schema` (to determine whether the field is required) and the `formData` (to see what data is already there). Doing that in `trimFormData` seems fine to me architecturally, but that's without looking at any resulting code.

gdbassett:
I would suggest `{}` for the 3rd case as the existence of `object` in the formData may imply information.

For case 3, it is the lack of `object` that appears to be the error, rather than the lack of `field1` or `field2`.  As such I'd think the error message would reflect "user is a required field" rather than "phone number".

mirajp:
Any updates on this PR?

epicfaace:
@davidMcneil , it looks like you deleted the repository / made it private from which this PR was started from, so I can't review your changes.

epicfaace:
I'm closing this PR due to the above issue, but @davidMcneil please feel free to make another PR with your actual changes.

PR #684, opened 2017-08-28
Fix #681: Warn on invalid array schema provided.

### LINKED ISSUE #681: Omitting `items` from arrays causes it to throw.
### Prerequisites

### Description

An array without the `items` field, breaks the rendered form.

I noticed some schemas I have were breaking stuff heavily in the app I'm working on. And I noticed sending invalid schemas usually results in a proper message like "the schema is invalid" or something, instead of rendering the form, but for some schemas it would just break catastrophically impacting in my whole page some times.

Upon further investigation I noticed some of my schemas had properties of type "array" with no "items" field. And I was surprised to see that in my database because I validate everything before saving. Turns out these are actually valid schemas:

>"Omitting this keyword has the same behavior as an empty schema."
http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.9

I debugged react-jsonschema-form's source code a little bit and saw that it was really throwing here:
https://github.com/mozilla-services/react-jsonschema-form/blob/9d79b71f13d87831e5c9e2d8fc7ae5e5ca7865c4/src/utils.js#L398
that is a call from here:
https://github.com/mozilla-services/react-jsonschema-form/blob/d403edc20a1d839326b8f6d0a380f84697efba2b/src/components/fields/ArrayField.js#L367
And this code tells me you guys are assuming all arrays should have the `items` property and not really checking if `schema.items` is defined, causing it to throw an uncatched exception.

### Steps to Reproduce

1. Just run this fiddle https://jsfiddle.net/h6fsjLhj/

#### Expected behavior

I know it's probably not possible to render a form for this scenario, but the tool should ignore fields like this or probably provide feedback that this type of schema is not supported.

#### Actual behavior

The tool just throws, and since catching rendering exceptions in React is not always simple, it can cause unexpected behavior in the whole app.

### Version

0.49.0

Thanks! And let me know if I can provide more info, or if there is a workaround I can use.

n1k0:
Fix #681.

I've also updated the `UnsupportedField` component to provide more details:

Invalid array schema:

Invalid schema type:

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "jsdom": "^8.3.0",
     "lint-staged": "^3.3.1",
     "mocha": "^2.5.3",
-    "prettier": "^1.5.2",
+    "prettier": "^1.5.3",
     "react": "^15.5.0",
     "react-addons-test-utils": "^15.3.2",
     "react-codemirror": "^0.2.3",
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -276,10 +276,21 @@ class ArrayField extends Component {
   };
 
   render() {
-    const { schema, uiSchema, registry = getDefaultRegistry() } = this.props;
+    const {
+      schema,
+      uiSchema,
+      idSchema,
+      registry = getDefaultRegistry(),
+    } = this.props;
     const { definitions } = registry;
     if (!schema.hasOwnProperty("items")) {
-      return <UnsupportedField />;
+      return (
+        <UnsupportedField
+          schema={schema}
+          idSchema={idSchema}
+          reason="Missing items definition"
+        />
+      );
     }
     if (isFixedItems(schema)) {
       return this.renderFixedArray();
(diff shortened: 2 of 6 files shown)


glasserc:
Released in v0.50.0.

llamamoray:
@glasserc This change looks like it's part of the tagged commit 0.50.0, but the npm package doesn't seem to have the changed. 

This is the code I see in the published `lib/components/fields/UnsupportedField.js`:
```
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = UnsupportedField;

var _react = require("react");

var _react2 = _interopRequireDefault(_react);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

function UnsupportedField(_ref) {
  var schema = _ref.schema;

  // XXX render json as string so dev can inspect faulty subschema
  return _react2.default.createElement(
    "div",
    { className: "unsupported-field" },
    "Unsupported field schema ",
    JSON.stringify(schema, null, 2),
    "."
  );
}
```

Has something odd happened with the publish, or is it something local?

glasserc:
Oops, you're right, I published a bad version. I guess I should document how to release a version, since I keep forgetting. I've just pushed v0.50.1, can you tell me if that works correctly?

llamamoray:
0.50.1 looks good. Cheers!

PR #691, opened 2017-09-05
Allow overriding of unsupported field

llamamoray:

I would like to be able to create a custom component to display unsupported field errors. 

### Checklist

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -7,7 +7,7 @@ import {
   getDefaultRegistry,
   getUiOptions,
   isFilesArray,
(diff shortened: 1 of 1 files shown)


src/components/fields/SchemaField.js
```diff
@@ -33,8 +32,9 @@ function getFieldComponent(schema, uiSchema, idSchema, fields) {
   return componentName in fields
     ? fields[componentName]
     : () => {
+        const UF = fields["UnsupportedField"];
```
glasserc: I think `const UnsupportedField` is fine.

PR #698, opened 2017-09-13
Add picture preview functionality

shihuili:

[Please describe them here]

### Checklist


glasserc:
Thanks for the PR! This is a very interesting change! Do you think this should be optional, or do you think all users would want this functionality?

Note that this will also trigger for mimetypes like `application/x-apple-diskimage`, which I don't think you want.

epicfaace:
The picture preview should also be optional so it doesn't break existing functionality too. @shihuili do you still have time to work on this?

PR #699, opened 2017-09-15
Make BaseInput overridable

olzraiti:

All the other components in registry fields & widgets are overridable. BaseInput shouldn't be an exception. Customizing BaseInput is useful e.g. for propagating changes only on text input blur.

BTW there is something wrong with the npm commit hook - after pushing the first commit I noticed that some files were prettified after pushing, but the prettified changes were in git staging and weren't pushed. 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
      - [Custom widget components](#custom-widget-components)
         - [Custom component registration](#custom-component-registration)
         - [Custom widget options](#custom-widget-options)
+        - [Customizing widgets' text input](#customizing-widgets-text-input)
      - [Custom field components](#custom-field-components)
         - [Field props](#field-props)
         - [The registry object](#the-registry-object)
@@ -1037,6 +1038,10 @@ render((
 > Note: This also applies to [registered custom components](#custom-component-registration).
 
 > Note: Since v0.41.0, the `ui:widget` object API, where a widget and options were specified with `"ui:widget": {component, options}` shape, is deprecated. It will be removed in a future release.
+
+#### Customizing widgets' text input
+
+All the widgets that render a text input use the `BaseInput` component internally. If you need to customize all text inputs without customizing all widgets individially, you can provide a `BaseInput` component in the `widgets` property of `Form` (see [Custom component registration](#custom-component-registration).
 
 ### Custom field components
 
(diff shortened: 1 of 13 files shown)


glasserc:
Looks OK to me. Can we add a brief note to the documentation explaining that this is possible and why you might want to do this?

Regarding the commit hook, I've never heard of it accidentally not committing stuff. I'm more familiar with husky committing too much stuff, as in https://github.com/Kinto/kinto-admin/issues/419.

Chun-Yang:
Can we change 'BaseInput' to 'BaseInputWidget' to make it consistent?
Since BaseInput is not an open API, the name change should not affect anyone anyway.

glasserc:
That sounds OK to me.

olzraiti:
> Can we change 'BaseInput' to 'BaseInputWidget' to make it consistent?
> Since BaseInput is not an open API, the name change should not affect anyone anyway.

On the other hand, excluding the "Widget" part in the component/file name communicates that it's not a widget. It isn't a widget because you can't use `BaseInput` with `ui:widget` in `uiSchema`. Maybe we should move it away from `widgets` directory or use an underscore in the file name, but I think it's OK like it is now.

glasserc:
Thanks!

anlesk:
Hi guys,

I've just updated to the latest version of react-jsonschema-form and faced the issue with Jest snapshot tests. We have a custom widget that is actually a wrapper for original TextWidget.
With that PR being merged to master, snapshot test of custom widget is starting to fail with "unable to read property widgets of undefined". 

Shouldn't default registry be defined as a default prop for TextWidget? I found a workaround providing registry as a property for custom widget but that looks like a hack to me.
```
import { getDefaultRegistry } from 'react-jsonschema-form/lib/utils';
<CustomTextWidget  registry={getDefaultRegistry()} />
```

Please share your thoughts.

olzraiti:
The `registry` property must be passed to child fields/widgets always. If `BaseInput` doesn't receive `registry`, then it's likely that some of your custom fields/widgets doesn't pass the `registry` property properly.

We can't use the default registry as a default, because then the custom fields/widgets wouldn't be used.

anlesk:
@olzraiti, thanks for the response!

Shouldn't we define registry as a required propType than?

olzraiti:
@anlesk Yes, good idea. Feel free to open a pull request for that :)

Issue #702, opened 2017-09-20
Validation Error Display/CSS


sgentry:
### Prerequisites

- [ x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [ x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I've add required fields to my schema and I'm seeing odd behavior. When only a single field fails validation, the entire form is highlighted in red. 

### Steps to Reproduce

This behavior can be seen by looking at the "Simple" example in the playground app: [https://mozilla-services.github.io/react-jsonschema-form/](url)

1. Delete the text from the last name field. All the form inputs are highlighted.
2. Reload page and delete text from the first name field. No validation errors until you hit Submit.

#### Expected behavior

I would expect only the fields that failed validation to be highlighted in red. Also, the validation behavior seems to be handled differently between the first and last name fields, in the simple example. Maybe I'm incorrect in my assumption and am missing something obvious. Is this the expected behavior?

glasserc:
The validation in that example sure is confusing! What's happening is that there is an `emptyValue` for the `firstName` in the `uiSchema`, which means that the field is always "present" even if it's empty. Marking the field "required" here has no effect -- the schema could add a `minLength` constraint if the empty string is not valid.

There's actually a bunch of bugs here. One is that if you completely clear the password field, it doesn't show as an error. It only shows as an error if it has a string in it that is not empty and fewer than three characters.

The highlighting thing comes about because the actual JSON Schema error we get is that the instance is invalid because it's missing a field. This is not very user-friendly. I would be interested in reviewing a patch to fix it.

glasserc:
Thinking about it further, I guess the password field is not considered an error because clearing the whole field removes it from the object -- it's only invalid if it's present and less than three characters. So maybe that isn't a bug.

sgentry:
Thanks for your explanation!

I found this [issue](https://github.com/mozilla-services/react-jsonschema-form/issues/384) which is basically the behavior I'm seeing in my tests. There was mention of switching to the [ajv](https://github.com/epoberezkin/ajv) validator to assist resolving the issue. So I decided to give it a shot and see what happens. My approach was to transform the ajv error output to more closely match what this lib expects from jsonschema and minimize the changes I'd need to make. 

[...]

Issue #384, opened 2016-11-09
Add class like has-error to help style individual non-validating fields 


jmatsushita:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

The `has-error` class is applied to the parent of a nested object when validation of a required nested object property fails.

It means that until the user clicks "Submit" (which will display the " Please fill out this field" tooltip, there is no way (that I know off) to style individual fields that don't (live) validate.

### Steps to Reproduce

1.  Use schema:

```jsx
{
  "title": "A list of tasks",
  "type": "object",
  "required": [
    "title"
  ],
  "properties": {
    "title": {
      "type": "string",
      "title": "Task list title"
    },
    "tasks": {
      "type": "object",
      "title": "Tasks",
      "required": [
        "title"
      ],
      "properties": {
        "title": {
          "type": "string",
          "title": "Title",
          "description": "A sample title"
        },
        "details": {
          "type": "string",
          "title": "Task details",
          "description": "Enter the task details"
        },
        "done": {
          "type": "boolean",
          "title": "Done?",
          "default": false
        }
      }
    }
  }
}
```

2. The containing parent field have a `has-error` class

![image](https://cloud.githubusercontent.com/assets/356097/20149455/a28f1f8c-a6b1-11e6-8691-f0890004dc81.png)

#### Expected behavior

Rather desired behavior would be to have a `has-error` class on individual field elements that don't validate. Or maybe a `has-field-error` to help distinguish the failing nested objects from the fields themselves.

#### Actual behavior

The individual field doesn't have a class that could help with providing live feedback about problem fields:

![image](https://cloud.githubusercontent.com/assets/356097/20149486/cf28a9c8-a6b1-11e6-9ecf-582d0c4a50a1.png)

### Version

This was tested with the live version on https://mozilla-services.github.io/react-jsonschema-form/

jmatsushita:
I see this is being discussed here #351 and seems doable with `FieldTemplate`s already. I also now realise that the `required` constraint is attached to the parent property, so the default makes sense although it's not the best UX.

maartenth:
I think we can parse the error message and pass the error through to the missing field(s), instead of the entire object. That would be better UX in my opinion.

n1k0:
[...]

PR #711, opened 2017-10-04
Switch to ajv validator

sgentry:

Switching to the [ajv](https://github.com/epoberezkin/ajv) validator should help address validation problems outlined in these issues:

#384 
#702 

### Checklist

  - [x ] I've added and/or updated tests
  - [ x] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style


webpack.config.prod.js
```diff
@@ -39,6 +39,15 @@ module.exports = {
...
         ],
+      },
+      {
+        test: /\.json$/,
+        loader:"json-loader",
+        include: [
+          path.join(__dirname, "css"),
+          path.join(__dirname, "playground"),
+          path.join(__dirname, "node_modules"),
+        ],
```
glasserc: What's up with these json-loader changes? Do we import JSON explicitly anywhere in the project? I don't see any new uses of this functionality as part of this PR..

src/validate.js
```diff
@@ -99,6 +112,31 @@ function unwrapErrorHandler(errorHandler) {
...
+ */
+function transformAjvErrors(errors = []) {
+  if (errors === null) {
+    return [];
+  }
+
+  return errors.map(e => {
+    const { dataPath, keyword, message, params } = e;
+    // not having 'instance' in the property name breaks things further down the line.
+    let property = `instance${dataPath}`;
```
glasserc: Could you elaborate on this?

webpack.config.prod.js
sgentry: Ajv loads two json files so I added the loader for that.

ajv/lib/refs/json-schema-draft-06.json
ajv/lib/refs/$data.json

It uses draft 6 by default but we can specify an earlier revision if needed. Not sure if using v6 would be a breaking change for this lib.

src/validate.js
sgentry: Not having 'instance' in the property value caused an additional 8 tests to fail so I was thinking that there's a dependency in some of the components for this, and didn't dig further. Maybe I just need to update the tests. I'll take a closer look at the tests.

webpack.config.prod.js
glasserc: Ah, great, thanks. I think draft 6 is what we target at the moment.

src/validate.js
sgentry: I pushed another commit to remove the reliance on `instance` and updated the affected tests.

sgentry:
I agree that switching from `arguments` -> `params` could be a breaking change for anyone who may use it. I guess another option is to try and re-create `arguments` based on the ajv error list but I don't think I'd like to attempt that.

glasserc:
Thank you very much for this great work!

PR #714, opened 2017-10-06
UpDownWidget not correctly using BaseInput

domharrington:

Whilst attempting to override the BaseInput, I noticed it wasn't being applied for the UpDownWidget.
This change adds a test case for all widgets that use BaseInput to make sure they can be overridden successfully.

I've also removed all instances where BaseInput was required directly, and instead made it always come from the registry for consistency.

### Checklist


glasserc:
Thanks!

PR #715, opened 2017-10-06
Fix #622 - correctly parse JSON date into local date

### LINKED ISSUE #622: Bug in native datetime widget in playground
### Prerequisites

### Description

There are two bugs:
1. The native datetime widget alters the displayed time after it's completely filled out
2. After the native datetime widget is filled out, attempting to alter the time results in unusual behavior. For example, if attempting to increment the hour by 1, it instead increments by 5.

It appears that once the datetime widget is filled out or altered, the datetime is immediately cast to UTC (which is expected per #496), but the date and time in the widget are not shown in local time, causing unexpected behavior.

### Steps to Reproduce

1. Open [playground](https://mozilla-services.github.io/react-jsonschema-form/) and go to "Date & time" tab
2. Fill out the native datetime widget and notice that the time jumps several hours once you fill out all parts of the widget
3. Attempt to alter the time and notice that the displayed time isn't what you expect

#### Expected behavior

Date and time in the native datetime widget should be displayed in the user's local time.

#### Actual behavior

Date and time in the native datetime widget are being displayed in UTC.

### Version

Whatever version is in the playground

Chrome: v59.0.3071.109

My current timezone is EST.

mmmveggies:
Currently `datetime-local` is basically unusable and #622 sums up the reason quite well.
This makes sure the `fromJSONDate` returns a locally-adjusted time  (see [spec](https://html.spec.whatwg.org/multipage/input.html#local-date-and-time-state-(type%3Ddatetime-local))) string to set `value` to on the `HTMLInputElement`, and not the straight up JSON (GMT) value.

It seems that, in chrome 61 and edge 40 at least, a JSON-like ISO-8601 timestamp without the Z is interpreted as local time and the 'Z' appendage designates UTC.

```js
new Date('0000-01-01T00:00:00Z').toJSON()
"0000-01-01T00:00:00.000Z"
new Date('0000-01-01T00:00:00').toJSON()
"0000-01-01T07:00:00.000Z"
```

### Checklist

### DIFF
--- a/src/components/widgets/DateTimeWidget.js
+++ b/src/components/widgets/DateTimeWidget.js
@@ -1,11 +1,8 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { pad } from "../../utils";
 
-function pad(p, n) {
-  return String(n).padStart(p, 0);
-}
-
-function fromJSONDate(jsonDate) {
+export function fromJSONDate(jsonDate) {
   if (!jsonDate) {
     return "";
   }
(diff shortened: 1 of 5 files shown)


mmmveggies:
My current work-around is to just define 
```js
uiSchema.startDate['ui:options'].inputType = 'datetime-local'
```

https://github.com/mozilla-services/react-jsonschema-form#html5-input-types

glasserc:
Hi, thanks for digging into this! Your explanation is quite clear -- JSON dates are in UTC, but the `value` field of the input needs to be in local time.

I spent a little time trying to understand what role `toJSONDate` and `fromJSONDate` play and how they are supposed to work. It seems like now we actually have a pretty formal contract -- `toJSONDate` is actually more of a "localtime-to-UTC" function, and `fromJSONDate` is a "UTC-to-local" function. Can we maybe rename them to express this contract a little more clearly, and/or add an ESDoc comment?

I think `fromJSONDate` is complicated enough now that we ought to write a test explicitly for it, although I'm not really sure how to do so in a way that is isolated from the tz of the developer. What do you think?

mmmveggies:
I'm not quite sure how to spoof the timezone in the test runner, we could test using `getTimezoneOffset` but I feel that the implementation of the test would require more testing than the function itself. 

Right now, this isn't the case in any timezone non-GMT, so maybe we could add this for *most-of-the-way* consistency 

```js
const date = new Date().toJSON()

toJSONDate(fromJSONDate(date)) === date
```

src/components/widgets/DateTimeWidget.js
```diff
@@ -1,8 +1,27 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { pad } from "../../utils";
 
-function fromJSONDate(jsonDate) {
-  return jsonDate ? jsonDate.slice(0, 19) : "";
+export function fromJSONDate(jsonDate) {
```
glasserc: I know this is a bit outside the scope of the PR, but could you rename these functions something like `utcToLocal` and `localToUTC`?

mmmveggies: Happy to :)

PR #724, opened 2017-10-18
Add `checked` class name for RadioWidget

iplus26:

Add `checked` class name for RadioWidget, just like `disabled` class name it has. 

Sometimes we customize the style for `input[checked]` by hiding the default `input[type=radio]` itself and instead, we use a pseudo element like below: 

```css
input[type=radio] {
    opacity: 0;
}
.radio:before{
    content: "";
    box-sizing:border-box;
    position: relative;
    top: 2px;
    bottom: 0;
    width: 14px;
    height: 14px;
    z-index:2;
    border:1px solid #666;
    border-radius: 100%;
    display: inline-block;
    margin-right:9px;
}
.radio.checked:after{
    content: "";
    position: absolute;
    z-index:3;
    top:16px;
    left:3px;
    width:8px;
    height:8px;
    background: #666;
    border-radius: 100%;
}
```

Therefore, a `checked` class will be useful.

### Checklist

### DIFF
--- a/src/components/widgets/RadioWidget.js
+++ b/src/components/widgets/RadioWidget.js
@@ -39,7 +39,9 @@ function RadioWidget(props) {
         );
 
         return inline ? (
-          <label key={i} className={`radio-inline ${disabledCls} ${checkedCls}`}>
+          <label
+            key={i}
+            className={`radio-inline ${disabledCls} ${checkedCls}`}>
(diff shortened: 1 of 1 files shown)


glasserc:
Why is this better than adding styles to `.radio:checked:after`?

PR #726, opened 2017-10-18
Fixed unneeded values variable

keithweaver:

It causes linting errors with my own version and provides zero benefit. As the value object
is being created from the map function and used below.

### Checklist

  -  [x] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style

### DIFF
--- a/test/uiSchema_test.js
+++ b/test/uiSchema_test.js
@@ -354,7 +354,7 @@ describe("uiSchema", () => {
       };
 
       const CustomWidget = props => {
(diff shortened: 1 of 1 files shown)


keithweaver:
Ran a `npm run test` and nothing broke.

glasserc:
Hi, thanks for noticing this! Actually, I also caught this in #722, which I took the liberty of just merging. Sorry to make you do all that work for nothing!

PR #734, opened 2017-10-23
Updates conditional field example

OliDM:
Fixes errors with the dependencies on the conditional field example.

i tried to check out the example and i saw that the fiddle wasn't working due to setup.

### Checklist

* [x ] **I'm updating documentation**


glasserc:
Thanks!

PR #741, opened 2017-10-26
fix: Pass `disabled` prop to `FieldTemplate`.

theengineear:
If you want to change how your FieldTemplate is rendered based on
ui:disabled, you currently have to check `uiSchema["ui:disabled"]`,
which is unlike the pattern for `readonly` and other known `ui:*`
settings.

`ui:readonly -> props.readonly` in `FieldTemplate`, but `ui:disabled` does not. This changes that.

### Checklist


glasserc:
Thanks!

Issue #747, opened 2017-11-01
unique "key" prop warning when running the tests


faassen:
When I run the tests with ``yarn test`` I get this message on top of the test output:

```
Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `ObjectFieldTemplate`. See https://fb.me/react-warning-keys for more information.
    in div (created by ObjectFieldTemplate)
    in ObjectFieldTemplate (created by ObjectField)
    in ObjectField (created by SchemaField)
    in div (created by DefaultTemplate)
    in DefaultTemplate (created by SchemaField)
    in SchemaField (created by Form)
    in form (created by Form)
    in Form
```
I traced this to ``ObjectFieldTemplate_test.js``, line 33. It maps over properties like this:

```
            {properties.map(({ content, index }) => (
              <div key={index} className="property">
                {content}
              </div>
            ))}
```

But ``index`` doesn't appear to be a field of properties items. If I change the code to this the error goes away and the tests still pass:

```
            {properties.map(({ content }, index) => (
              <div key={index} className="property">
                {content}
              </div>
            ))}
```

If this looks all right I can make a pull request.

PR #748, opened 2017-11-01
Fix issue #747

faassen:
Shut up React key warning during tests. See issue #747


glasserc:
This was 100% my fault -- I made olzraiti take out "index" before merging #653. Thanks for fixing it.

PR #757, opened 2017-11-10
Proof of concept async validation

olzraiti:

Async validation is probably one of the most desired features that RJSF is missing. There's been at least one very good attempt at this: #401. I suspect that the PR is abandoned, so I wrote this quick implementation to push this feature forward.

I needed this feature NOW, so I wrote it and didn't fix the tests yet, because I hope that #401 would provide this feature instead (not very likely, I think)., as it is way more polished and though out.

EDIT: You can see it in action at the "Validation" tab on the playground and entering non-matching passwords.

### Checklist


olzraiti:
Closed by #1444

PR #765, opened 2017-11-15
Add span and class to label required symbol

ttbarnes:
Currently a label's required symbol is pure text string. There is no way to style this without creating a custom component/field/widget.

I've added a span and class name to the required symbol.

No unit tests have been added for this new small span element. I wanted to do this, but I don't have the time right now to learn how to do this with the test config/library(s) in use.

### Checklist


edi9999:
I think that makes sense

PR #766, opened 2017-11-15
Remove p tag surrounding submit button

ttbarnes:

- I don't see a need to have a paragraph surrounding a submit button
- having this wrapping p tag can cause unwanted styles (eg typically a margin)
- IMO for a submit button, a div is more semantic than a paragraph tag, since it's not a paragraph 😄 

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -208,11 +208,11 @@ export default class Form extends Component {
         {children ? (
           children
         ) : (
-          <p>
+          <div>
             <button type="submit" className="btn btn-info">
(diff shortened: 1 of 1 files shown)


MatejMazur:
Related #899 (basically all UI components are moved into the new separate UI package)
submit button -> `SubmitTemplate`

epicfaace:
LGTM -- what do you think, @glasserc ?

glasserc:
Yes, thanks! Sorry for taking so long to get to it!

PR #767, opened 2017-11-15
Rename some props to be lowercase

ttbarnes:

I noticed that a few props start with uppercase and this bothered me. EG you can end up with:

```
<Form
  schema={}
  onChange={}
  onSubmit={}
  onError={}
  ErrorList={}
  FieldTemplate={}
  ObjectFieldTemplate={}
  ArrayFieldTemplate={}
/>
```

I've renamed these 4 props since I think it's cleaner/more consistent.
```
ErrorList           === errorList
FieldTemplate       === fieldTemplate
ObjectFieldTemplate === objectFieldTemplate
ArrayFieldTemplate  === arrayFieldTemplate
```

### Checklist


edi9999:
Since this is a breaking change (the public API changes in a non-retro compatible way), we should postpone this to a new release (2.0)

glasserc:
I don't really agree with this change. These props are uppercase because they are the names of components. In JSX, `<foo>` is generally understood to be an HTML tag, and `<Foo>` is generally understood to be a component.

edi9999:
I understand your point. I'm not sure if the convention applies only to tagnames, or also to props.

For example, the `props.children` is lowercase and not `props.Children`

MatejMazur:
I think that if we find a way to merge #1013 then we don't need to bother with this because templates are moved under the `Form.props.templates`.

epicfaace:
Closing this due to inactivity. I agree with @glasserc 's justification for not doing this change.

PR #769, opened 2017-11-16
Add key to stop Array forms scroll up on playground

nilportugues:

On long forms with Array/Nested forms, onChange, the playground will scroll up. This is weird behavior and bad UX. By setting a key this is solved.

### Checklist


glasserc:
Hi, could you provide steps to reproduce? The playground does not scroll up when I change the name field of elements in the Arrays section. Why would setting a randomly-generated UUID as the key on each render() cause this to be different? There is no key property that I can find in the Codemirror documentation or source code, and the React documentation seems to suggest it's only important for lists.

edi9999:
Any updates on this @nilportugues ? 

I will be closing this for now, until further explanation.

Issue #635, opened 2017-07-15
Add `minRemovable` options


mfulton26:
RJSF supports [`addable`](https://github.com/mozilla-services/react-jsonschema-form#addable-option) and [`removable`](https://github.com/mozilla-services/react-jsonschema-form#removable-option) options but does not currently support hiding the remove buttons when `minItems` is met. I propose the following option (or similarly named one) be added:

#### `minRemovable` option

A remove button is shown by default even when the number of items is less than or equal to `minItems`. You can turn this off with the `minRemovable` option in `uiSchema`:

```jsx
const uiSchema = {
  "ui:options":  {
    minRemovable: false
  }
};

mfulton26:
I've since learned that the add button is not shown once `maxItems` is worth so I've updated this issue to only add a `minRemovable` option.

glasserc:
Hmm. I didn't realize we don't show an add button once we reach `maxItems`. (I even [wrote about it](https://github.com/mozilla-services/react-jsonschema-form#philosophy) in the README.) Personally, I would have said we should show buttons in all cases. I agree that some users want to prevent users from "coloring outside of the lines", as it were, but I feel like such a granular option as `minRemovable` is too granular -- in general, I would expect a developer to either want none of the safety guards, or all of them. So I would prefer to see a larger-grain option, perhaps at the root of the uiSchema, that adds all of the safety possible. What do you think about that? I understand it's a more invasive change, but I think it's a more holistic approach.

mfulton26:
Yeah, I like that. I'm currently involved more heavily in some other projects at the moment but when I circle back to the one where we're using rjsf I'll take a look at a more in depth change (unless someone beets me to it).

PR #776, opened 2017-11-24
Add minRemovable to ui:options

mfulton26:

Sometimes I'd like to prevent the users from removing array items when they've only entered `minItems` worth of items.

This implements #635.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
         - [orderable option](#orderable-option)
         - [addable option](#addable-option)
         - [removable option](#removable-option)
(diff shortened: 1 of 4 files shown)


edi9999:
+1 on this MR

epicfaace:
> After thinking about this, I think this should not be an option at all imo but rather a default behavior, who would want to have the icon if it is not possible to delete items ??

I agree. @mfulton26 will you be able to implement this as the default behavior and put that into this PR?

mfulton26:
I am no longer actively using JSON Schema, etc. Anyone is welcome to adapt the code I've posted here but I'm closing this PR as I won't personally be pursuing its completion.

PR #778, opened 2017-11-28
Generating idSchema based on dependency changes

redixhumayun:

[Library throwing an error regarding `id` prop not being passed. This should fix the issue]

### Checklist

* [] **I'm updating documentation**


edi9999:
I think this would fix #803. Can we merge this ?

src/components/Form.js
```diff
@@ -39,14 +40,18 @@ export default class Form extends Component {
     const mustValidate = edit && !props.noValidate && liveValidate;
     const { definitions } = schema;
     const formData = getDefaultFormState(schema, props.formData, definitions);
+
+    //create a retrievedSchema object here to get new idSchema
```
glasserc: I don't think this comment is necessary.

redixhumayun: Okay, made the necessary changes. Let me know if anything else needs to be done.

redixhumayun:
Let me know if this is good to be merged!

edi9999:
Looks good to me. I'm going to merge this and squash it into one commit, hope that's good for you too @glasserc

PR #814, opened 2018-01-16
Removed throwing error when there's an extraneous property in ui:order

igorgubernat:

In current implementation if you use dynamic schema dependencies, you cannot specify dependent fields in 'ui:order' list. It's because if you list such field in 'ui:order' and it is currently not displayed according to it's dependency condition, the error is thrown: 'uiSchema order list does not contain <field>'.

I do not see why we should throw an error instead of just filtering 'ui:order' array to contain fields that are present according to dependency condition. If you cannot order conditional fields along with unconditional, it significantly reduces usefulness of dynamic schema dependency feature.

My solution is to give user the ability to specify all properties in 'ui:order' and just sort them as needed.

### Checklist

### DIFF
--- a/test/ObjectField_test.js
+++ b/test/ObjectField_test.js
@@ -218,19 +218,6 @@ describe("ObjectField", () => {
       expect(labels).eql(["baz", "bar", "qux", "foo"]);
     });
 
-    it("should throw when order list contains an extraneous property", () => {
-      const { node } = createFormComponent({
-        schema,
-        uiSchema: {
-          "ui:order": ["baz", "qux", "bar", "wut?", "foo", "huh?"],
-        },
-      });
-
-      expect(node.querySelector(".config-error").textContent).to.match(
-        /contains extraneous properties 'wut\?', 'huh\?'/
-      );
-    });
-
     it("should throw when order list misses an existing property", () => {
       const { node } = createFormComponent({
         schema,
--- a/test/ObjectField_test.js
+++ b/test/ObjectField_test.js
@@ -218,6 +218,22 @@ describe("ObjectField", () => {
       expect(labels).eql(["baz", "bar", "qux", "foo"]);
     });
 
+    it("should use provided order also if order list contains extraneous properties", () => {
+      const { node } = createFormComponent({
+        schema,
+        uiSchema: {
+          "ui:order": ["baz", "qux", "bar", "wut?", "foo", "huh?"],
+        },
+      });
+
+      const labels = [].map.call(
+        node.querySelectorAll(".field > label"),
+        l => l.textContent
+      );
+
+      expect(labels).eql(["baz", "qux", "bar", "foo"]);
+    });
+
     it("should throw when order list misses an existing property", () => {
       const { node } = createFormComponent({
         schema,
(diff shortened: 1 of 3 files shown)


edi9999:
Can you update the test instead of deleting it ?

igorgubernat:
Added tests for orderProperties(). Thanks.

test/ObjectField_test.js
```diff
@@ -218,19 +218,6 @@ describe("ObjectField", () => {
       expect(labels).eql(["baz", "bar", "qux", "foo"]);
     });
 
-    it("should throw when order list contains an extraneous property", () => {
```
edi9999: Can you uncomment keep this test please and change the expectation ?

igorgubernat: Done

ahardworker:
When is this getting merged @edi9999 ?

edi9999:
I personnaly have found that the shown warning is useful, especially when you make typos , for example if I write in my json-schema :  

`users`

and in my ui-schema : 

`usesr`.

Maybe this could be an option that you set globally, such as : 

`allowExtraneousUIOrder={true}`

ahardworker:
Oh good point. You just gave me an idea, would it be better to show a warning in the console instead of having the warning inside the form? This way you could easily track any mispelled properties and extraneous properties by looking at your console without breaking the form. What do you think?

edi9999:
Yes, I think it makes sense to use `console.warning` for that.

glasserc:
Yes, the warning when you use a property name that doesn't exist can save a lot of aggravation. I understand it isn't really helpful in the case of dynamic schemas with property dependencies, which is why I suggested in #823 that we could maybe loosen the check in only those cases. I'd prefer logic that does the right thing to an option that the user has to know to set. @jaminthorns commented in that issue that `retrieveSchema` "resolves" dependencies, so we might have to refactor things in order to get this right.

SlashBin:
This would be very helpful for me as I'm starting to get into generating dynamic JSON Schemas based on various rules, I would like to not have to worry about also dynamically generating my UI Schema's "ui:order" lists if I don't have to to avoid spitting errors at my users because they don't need certain fields. It should be simple enough to tell the UI Schema "If the JSON Schema has this field, display it here on the form".
Isn't the point of the UI Schema more of a suggestion to how to render a JSON Schema? Most things seem to gracefully degrade functionality if the UI Schema doesn't make your JSON Schema.

SlashBin:
Perhaps we could just have a `*` or some other indicator for "optional" fields in `ui:order`. For example:

```
"ui:order": [
    "first_name",
    "last_name",
    "date_of_birth",
    "is_pediatric*",
    "additional_comments*"
]
```
I would suggest a more regex-y`?` but it looks like that's common in property names, according to your examples.

SlashBin:
Ran into an related issue today which is blocking our project, looking for advise.

We have two "enum" fields in a section that have dependencies, when you select "Other" there is a text entry (this is an extremely common pattern on the web, no?)

I have no way of telling `ui:schema` where to display these two "Other" sections which need to be directly below the selection.

```
"ui:order": [
      "description",
      "description_other",
      "frequency",
      "frequency_other",
      "duration",
      "planned_procedures",
      "additional_comments"
    ]
```
The above results in:
`Invalid plan object field configuration:uiSchema order list contains extraneous properties 'description_other', 'frequency_other'.`

Am I missing some more practical way of getting this behavior? Certainly this is pretty common, so I'm thinking maybe dependencies aren't the way to do it.

@glasserc

PR #825, opened 2018-01-26
a priori should be italicized for readibility

sinbonders:

Spotted a typo in the readme.
### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
 
 ## Philosophy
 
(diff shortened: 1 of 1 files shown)


glasserc:
Well, it's correct as written. It should probably be written in italics: _a priori_. See https://en.wiktionary.org/wiki/a_priori#Adjective. I agree that use of Latin phrases in a text of this nature is probably unnecessary, so maybe something like "ahead-of-time knowledge" would be better?

sinbonders:
@glasserc I did not know that one. Makes sense now, but _in situ_ it looks wrong at a glance. I agree with your suggestion though. Updated this PR to reflect that :)

glasserc:
Thanks -- _mea culpa_ :)

Issue #727, opened 2017-10-19
Passing errors down to Widgets


llamamoray:
### Prerequisites

- [x ] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [ ] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Is there any reason why the StringField, NumberField, BooleanField etc don't pass the error schema down to the widgets they render?

I have implemented most of the input components with a custom component library which can display errors as part of their functionality. It would be neater to show errors using these controls rather than displaying errors separately using a Field Template. Is there any fundamental reason that the errors aren't passed down to the widgets?

iplus26:
![image](https://user-images.githubusercontent.com/6082721/33238270-46092802-d2c5-11e7-831b-338a7d091222.png)
+1 This looks really wired. Is implementing all the input components (which is really not elegant) is the only way to avoid that?

llamamoray:
@glasserc is there any reason why you wouldn't want this implemented? 

I'm considering putting aside some time to implement, but want to make sure that it will be considered to be merged into the project.

glasserc:
I think showing errors near their source is definitely better. I don't know what the reasoning was for having them all together at the top. Be careful to also handle errors that do not have a specific location (such as violations of "required", although I'm not sure if that's still an issue).

Vandell63:
Hey there, Thx for this awesome project. But i try to implement my custom field with the materialUI React. But even i created a custom field i have to passing down errors array messages to my children ( Children props from my custom field template ) because material ui input field need an error context ( error && error message ). How can we do this ?

llamamoray:
@glasserc I've created a PR #826 which passes the raw errors down through the SchemaField, down to the widgets. This does not change any of the behaviour of the current implementation, but will allow people who are using custom widgets to display errors where they see fit.

llamamoray:
@iplus26 the first part of what you require has been merged in #826 so the fields are receiving raw errors. As I have entirely custom fields this covers my use case, however if you still require the default bootstrap controls to display errors differently you'll need to create a PR. 

@glasserc when do you plan to do the next release?

glasserc:
I've tried to cut version 1.0.1, please tell me if that provides what you need.

llamamoray:
@glasserc Just tried it out and working as expected. Thanks

[...]

PR #826, opened 2018-01-29
pass raw errors to  field widgets

llamamoray:

See issue #727 

The current implementation of ArrayField, BooleanField, and StringField don't pass the error schema down to the widgets they render. This makes rendering errors as part of a custom control tricky. 

This PR solves this by passing the raw errors down, but maintaining the default implementation for displaying errors with the libraries built in widgets.

### Checklist

* [ x] **I'm adding or updating code**
  - [ x] I've added and/or updated tests
  - [x ] I've updated docs if needed
  - [ x] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -8,7 +8,7 @@ import {
   getUiOptions,
   isFilesArray,
   deepEquals,
-  getSchemaType,
+  getSchemaType
 } from "../../utils";
 import UnsupportedField from "./UnsupportedField";
 
@@ -19,7 +19,7 @@ const COMPONENT_TYPES = {
   integer: "NumberField",
   number: "NumberField",
   object: "ObjectField",
-  string: "StringField",
+  string: "StringField"
 };
 
 function getFieldComponent(schema, uiSchema, idSchema, fields) {
@@ -103,7 +103,7 @@ function DefaultTemplate(props) {
     description,
     hidden,
     required,
-    displayLabel,
+    displayLabel
   } = props;
   if (hidden) {
     return children;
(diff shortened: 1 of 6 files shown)


src/components/widgets/BaseInput.js
```diff
@@ -22,14 +22,18 @@ function BaseInput(props) {
   const _onChange = ({ target: { value } }) => {
     return props.onChange(value === "" ? options.emptyValue : value);
   };
+
+  const cleanProps = { ...inputProps };
+  delete cleanProps.rawErrors;
```
glasserc: I think you can do `const {rawErrors, ...cleanProps} = inputProps;` See the "Rest in object destructuring section" of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring.

src/components/fields/SchemaField.js
```diff
@@ -113,7 +113,9 @@ function DefaultTemplate(props) {
     <div className={classNames}>
       {displayLabel && <Label label={label} required={required} id={id} />}
       {displayLabel && description ? description : null}
-      {children}
+      {React.Children.map(children, child => {
+        return React.cloneElement(child, { rawErrors: props.rawErrors });
+      })}
```
glasserc: This seems a bit heavy. Why is this necessary? Wouldn't it be better to pass the `rawErrors` prop when we constructed widgets?

llamamoray: I've just reviewed the code again and I see that this can be passed through the SchemaFieldRender function. I'll update this.

llamamoray:
Thinking about this more, I don't think the ArrayField or ObjectField even need to be passed the raw errors. The objective of the PR is to pass field level errors down to the fields so that custom field components can render errors in any way they wish. I think I can remove the props for ArrayField  and ObjectField.

llamamoray:
@glasserc any feedback on my latest changes?

PR #834, opened 2018-02-07
Add allOf schema dependency support

jaredhoyt:

The `oneOf` schema dependency support is nice when you've got one condition. But if there are multiple, non-mutually exclusive dependencies on a single field, then you also need `allOf`.

For example, if Field1 were an array with 3 options: A, B, and C. And Field2 was supposed to appear if option B were selected, and Field3 was supposed to appear if A and C were selected, _and_ you wanted both Field2 and Field3 to appear if A, B, and C were selected, then this would accomplish that.

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -466,21 +466,32 @@ function withDependentSchema(
   dependencyKey,
   dependencyValue
 ) {
-  let { oneOf, ...dependentSchema } = retrieveSchema(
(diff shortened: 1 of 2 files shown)


jaredhoyt:
Sorry, this was meant to be a PR into my own fork. 

Apologies for adding this into the incorrect repo.

PR #842, opened 2018-02-15
Fix warning on FileWidget

lauriii:
Following warning is thrown when using FileWidget:

`Warning: Setting defaultProps as an instance property on FileWidget is not supported and will be ignored. Instead, define defaultProps as a static property on FileWidget.`

It seems like default props have been set twice on the FileWidget so I just removed the instance that is causing the warnings. I checked and there are no other instances of this problem.

### Checklist

### DIFF
--- a/src/components/widgets/FileWidget.js
+++ b/src/components/widgets/FileWidget.js
@@ -60,10 +60,6 @@ function extractFileInfo(dataURLs) {
 }
 
 class FileWidget extends Component {
-  defaultProps = {
-    multiple: false,
-  };
-
   constructor(props) {
     super(props);
     const { value } = props;


src/components/widgets/FileWidget.js
```diff
@@ -60,10 +60,6 @@ function extractFileInfo(dataURLs) {
 }
 
 class FileWidget extends Component {
-  defaultProps = {
```
edi9999: On the bottom it shows : 

```

 FileWidget.defaultProps = {
   autofocus: false,
 };
 ```

so it seems the defaultprops are not the same, this is strange ...

renancouto: `multiple: false` should be added there, since `multiple` is not a required property

lauriii: The reason I initially left it out was to do the simplest possible change. When defaultProps is set as an instance property, they are ignored, meaning that if we add `multiple: false` to the defaultProps, it is small change in the functionality. If the maintainers think we should fix both problems on a single PR, I'm happy to make the change 🙂

edi9999: Ok, I think we should rather fix the other problem later, to not break backwards-compatibility.

harouny:
cs-check step is failing on CI. Also didn't work on my local machine and I fixed it in the last pull request here. Must be a breaking change somewhere.

edi9999:
Yes, I see the cs-check failing. I will open a PR very soon to fix it.

edi9999:
See https://github.com/mozilla-services/react-jsonschema-form/pull/892

PR #844, opened 2018-02-15
Fixed issue where form validation errors stay on the page after user fixed the causes

### LINKED ISSUE #793: Form validation errors stay on screen when re-rendering due to changed properties
### Prerequisites

### Description

Form validation errors stay on screen when re-rendering due to changed properties

The component which includes <Form> maintains form data internally and assigns it as formData prop to the <Form> component.

When user submits the form an gets validation errors, the errors stay on screen even after submitting valid data (validation succeeds, onSubmit fires).

I think internal <Form> state gets messed up when re-rendering due to changed properties.

Testing with `_.cloneDeep` (as proposed in #306) didn't help.

Although I think handling `onChange` should be required only for `liveValidate` mode, adding `onChange` with or without `_.cloneDeep` didn't help, either.

Maybe related:
* #306 

### Steps to Reproduce

Here's a fiddle with a mimimal case demonstrating this behavior:

https://jsfiddle.net/h2o/vz4nexpx/
(lodash available in `_` variable)

1. enter invalid data
2. submit form
3. correct data
4. submit form again

#### Expected behavior

Validation errors are hidden when submitting valid data.

#### Actual behavior

Validation errors remain on screen.

### Version

1.0.0

stepbackj:

This pull request should resolve #793 

Expected Behavior:
- During onSubmit, if the form validates successfully the form's state should reflect that by setting `state.errors = []` and `state.errorSchema = {}`

Actual behavior:
- Form.js's this.setState() isn't actually modifying the state of the component, so `state.errors` and `state.errorSchema` aren't modified properly after successful validation of a form.

Fix: Use util.setState and do the `this.props.onSubmit` callback after setting state. This is similar to how things are done in Form.js `onChange()` function.

### Checklist

### DIFF
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -190,8 +190,15 @@ describe("Form", () => {
   });
 
   describe("Custom submit buttons", () => {
-    it("should submit the form when clicked", () => {
-      const onSubmit = sandbox.spy();
+    it("should submit the form when clicked", done => {
+      let submitCount = 0;
+      const onSubmit = () => {
+        submitCount++;
+        if (submitCount === 2) {
+          done();
+        }
+      };
+
       const comp = renderIntoDocument(
         <Form onSubmit={onSubmit} schema={{}}>
           <button type="submit">Submit</button>
@@ -202,7 +209,6 @@ describe("Form", () => {
       const buttons = node.querySelectorAll("button[type=submit]");
       buttons[0].click();
       buttons[1].click();
-      sinon.assert.calledTwice(onSubmit);
     });
   });
 
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -930,9 +930,10 @@ describe("Form", () => {
       });
 
       it("should reset errors and errorSchema state to initial state after correction and resubmission", () => {
+        const onError = sandbox.spy();
         const { comp, node } = createFormComponent({
           schema,
-          onError: () => {},
+          onError,
         });
 
         Simulate.change(node.querySelector("input[type=text]"), {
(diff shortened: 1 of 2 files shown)


test/Form_test.js
```diff
@@ -922,6 +928,30 @@ describe("Form", () => {
           })
         );
       });
+
+      it("should reset errors and errorSchema state to initial state after correction and resubmission", () => {
+        const { comp, node } = createFormComponent({
+          schema,
+          onError: () => {},
```
edi9999: Why onError empty ?

Maybe this function should be called once, and thus add an expectation for that.

test/Form_test.js
```diff
@@ -922,6 +928,30 @@ describe("Form", () => {
...
+        });
+
+        Simulate.change(node.querySelector("input[type=text]"), {
+          target: { value: "short" },
+        });
+        Simulate.submit(node);
+
+        expect(comp.state.errorSchema).eql({
+          __errors: ["should NOT be shorter than 8 characters"],
+        });
```
edi9999: `+        expect(comp.state.errors.length).eql(1);`

to show that the size has changed.

test/Form_test.js
```diff
@@ -190,8 +190,15 @@ describe("Form", () => {
   });
 
   describe("Custom submit buttons", () => {
-    it("should submit the form when clicked", () => {
-      const onSubmit = sandbox.spy();
+    it("should submit the form when clicked", done => {
```
edi9999: Why did you have to change this ?

glasserc: Probably because calling `onSubmit` now happens asynchronously.

stepbackj:
@glasserc Just made a few changes

ojab:
Maybe make a release, since it's rather annoying issue?

prgpascal:
Can you please release a patch with this change? 
Thank you

Issue #71, opened 2016-03-17
Render without form tag?


zbyte64:
I want to use this project to extend existing forms but I cannot render the `Form` component inside another form. The two solutions I see is to convert the existing form code into schema json and merge it with the user supplied schema OR have the option to render without the form tag. Merging the fields should work but there are cases where the existing form layout would be difficult to represent with ui schema and jsonschema (examples: an "advanced" section which collapses down, code editor with codemirror, and embedded previews next to fields). It seems having control over whether to render the form tag would be more robust.

zbyte64:
If I sent a pull request that implemented `Fieldset` which acts like `Form` but without the form specifics, would it be accepted (assuming all contributing guidelines is followed)?

n1k0:
> If I sent a pull request that implemented Fieldset which acts like Form but without the form specifics, would it be accepted (assuming all contributing guidelines is followed)?

If usage remains backward compatible and the code is good, definitely yes :)

fermuch:
why don't you just use the component `SchemaField`?

Checkout how it is being used here: https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/Form.js

zbyte64:
Yes, I ended up using `SchemaField` and copying over all the `Form` methods except `onSubmit`. As to why not to do this: it is not DRY. Trying to factor out the common logic to another component either meant inheriting the common component or rendering a new sub-component which broke allot of unit tests that do internal state checks.

derekdon:
Any update on the best approach to this? Would like to drop a react-jsonschema-form powered fieldset into an existing form.

epicfaace:
This was fixed by #1345!

PR #849, opened 2018-02-20
Support Nested Forms (Forms in Forms)

CurtisHumphrey:
This allows nested forms because a form in a form is bad DOM

Use case:
* A form that has an link new record button that creates a new form (nested) that allow one to create this new record and then returns the record id to the first form for linking.  In my case the record is an uploaded picture.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
       "git add"
     ]
   },
-  "main": "lib/index.js",
+  "main": "src/index.js",
   "files": [
     "dist",
     "lib"
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -189,7 +189,7 @@ export default class Form extends Component {
     const _SchemaField = registry.fields.SchemaField;
 
     return (
-      <form
+      <div
         className={className ? className : "rjsf"}
         id={id}
         name={name}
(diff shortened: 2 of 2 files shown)


CurtisHumphrey:
Ok, I'll fix those errors here in the next day or so.

glasserc:
I don't really understand your use case, but even if a form within a form is bad, then certainly a form by itself should still be a `<form>`, no?

edi9999:
Can you describe your usecase @CurtisHumphrey ? 

This seems like it could be closed.

olzraiti:
I have actually exactly the same use case. Our images are uploaded to a separate database, and only the image ID is saved to the document. I display a RJSF for updating the metadata for the images, so we have two nested `<form>` elements (well, I actually don't since I solved this by displaying the image form in a modal, that is rendered outside the parent RJSF element).

Maybe we could render a `<form>` by default, but allow some parameter for changing the element type for special use cases like this?

edi9999:
I think you should be having a json-schema that has multiple levels instead of trying to have Forms and subforms. 

Seee for example the "Nested" demo in https://mozilla-services.github.io/react-jsonschema-form/

CurtisHumphrey:
I've re-put this on my list of ToDos for next week.  Sorry for the delays.

epicfaace:
@CurtisHumphrey , I'm closing this PR -- if we were to implement this, it should definitely by default render a `<form>` tag and only have the _option_ to render another HTML tag instead. That issue is tracked in #71.

epicfaace:
This was fixed by #1345!

PR #852, opened 2018-02-25
Update Radio Widget to use Bootstrap v4

davidbarratt:
Updates the Radio Widget to Support Bootstrap v4 based on their documentation:
https://getbootstrap.com/docs/4.0/components/forms/#checkboxes-and-radios

This change breaks backwards capability. Either the major version should be incremented, or a backwards-compability layer should be added (perhaps adding a bootstrap version prop?)

### Checklist

### DIFF
--- a/src/components/widgets/RadioWidget.js
+++ b/src/components/widgets/RadioWidget.js
@@ -14,17 +14,18 @@ function RadioWidget(props) {
   // Generating a unique field name to identify this set of radio buttons
   const name = Math.random().toString();
   const { enumOptions, inline } = options;
+  const inlineCls = inline ? "form-check-inline" : "";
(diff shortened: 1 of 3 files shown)


edi9999:
I think this can't be merged as is, since we do support bootstrap v3 for the moment.

davidbarratt:
Maybe we need some sort of config that specifies the bootstrap version you use?

glasserc:
Refs #899.

epicfaace:
I'm closing this PR, as it seems like if we ever do end up supporting Bootstrap v4, we would make rjsf framework-agnostic and have the option to configure which UI framework is used. That discussion is tracked in #899

PR #864, opened 2018-03-08
handle errors from ajv.validate to correctly display schema errors in form

### LINKED ISSUE #863: Schema errors not displaying
### Prerequisites

- [x ] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);

### Description

When you have an error in the schema for a form, the error is not being displayed.

### Steps to Reproduce

In the playground you can force an error such as setting required to a non-array value and the error isn't being displayed with a schema like:
```
{
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
      "title": "First name",
      "required": true
    }
  }
}
```

#### Expected behavior

I expect an error to display like this:

image

#### Actual behavior

Error is not displayed ( we only see an uncaught error in the console):

image

### Version

1.0.1, but I believe this was introduced when ajv was added as a dependency to the project.

llamamoray:

Schema errors not being displayed in form.

Fixes #863 

### Checklist

* [ x ] **I'm adding or updating code**
  - [ x ] I've added and/or updated tests
  - [ x ] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -153,7 +153,12 @@ export default function validateFormData(
   customValidate,
   transformErrors
 ) {
-  ajv.validate(schema, formData);
+  try {
+    ajv.validate(schema, formData);
+  } catch (e) {
+    // swallow errors thrown in ajv due to invalid schemas, these
+    // still get displayed
+  }
 
   let errors = transformAjvErrors(ajv.errors);
 
--- a/test/validate_test.js
+++ b/test/validate_test.js
@@ -134,6 +134,42 @@ describe("Validation", () => {
         expect(errors[0].message).to.equal(newErrorMessage);
       });
     });
+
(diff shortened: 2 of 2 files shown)


glasserc:
Thanks for putting this together. So, ajv throws when the schema is invalid. This behavior probably makes sense for ajv. I'm a little surprised that we don't do anything with the caught error, but fortunately (?) ajv also exposes the error in the schema the same way it would expose errors in the data.

llamamoray:
No problem @glasserc . Would it be possible to get this into a patch release soon? It's causing some issues in our project.

glasserc:
I have published 1.0.2 with this fix.

PR #871, opened 2018-03-23
Fix corner case of empty items array

jbalboni:

When upgrading from an old version I found that our code was creating array schemas in the shape shown in the test case, which was resulting in the `getDefaultFormState` function creating form data that had an array of `undefined`s, instead of using `additionalItems` as the item schema. This meant that required fields in the array item schema weren't triggering validation errors.

### Checklist


test/utils_test.js
```diff
@@ -241,6 +241,26 @@ describe("utils", () => {
           foo: 42,
         });
       });
+
+      it("should fill array with additional items schema when items is empty", () => {
```
jbalboni: This is not a great test description, but I'm not sure what a better one would be.

edi9999: I think it is quite ok, it describes what it is testing.

edi9999:
To me this looks good, I suggest to merge this soon.

epicfaace:
@glasserc Can you merge this?

glasserc:
Sorry for the delay! Thanks!

PR #874, opened 2018-03-24
Allow errorSchema to be passed into a form

### LINKED ISSUE #155: Allow the insertion of errors from outside JSONSchema validation

davidbarratt:
Fixes #155. Users should be able to control the `errorSchema` externally. For instance, if you receive errors from a server, you should be able to inject those errors into the form.

### Checklist

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -314,12 +314,19 @@ class App extends Component {
   constructor(props) {
     super(props);
     // initialize state with Simple data sample
-    const { schema, uiSchema, formData, validate } = samples.Simple;
+    const {
+      schema,
+      uiSchema,
+      formData,
+      errorSchema,
+      validate,
+    } = samples.Simple;
     this.state = {
       form: false,
       schema,
       uiSchema,
       formData,
+      errorSchema,
       validate,
       editor: "default",
       theme: "default",
@@ -365,6 +372,9 @@ class App extends Component {
 
   onFormDataEdited = formData => this.setState({ formData, shareURL: null });
 
+  onErrorSchemaEdited = errorSchema =>
+    this.setState({ errorSchema, shareURL: null });
+
   onThemeSelected = (theme, { stylesheet, editor }) => {
     this.setState({ theme, editor: editor ? editor : "default" });
     setImmediate(() => {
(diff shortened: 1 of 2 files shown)


edi9999:
Nice feature ! 

Can you add a unit test for this feature please ?

And also update the documentation ?

Maybe also in the playground ?

glasserc:
I'd also like to see tests. Does this merge the errors from the errorSchema prop with the errors generated by validation? It looks more like the validation errors *replace* the prop, which seems a bit strange to me.

davidbarratt:
If you have time to add some tests that would be fantastic! Please feel free to fork my fork and create a new PR on top of this one.

pieplu:
> I'd also like to see tests. Does this merge the errors from the errorSchema prop with the errors generated by validation? It looks more like the validation errors replace the prop, which seems a bit strange to me.

Maybe it should be work like the formData props. The parent component should save on his state the errorSchema, like the FormData, and make the merge if it want, and pass the new errorSchema to mozilla react-json form.

karianpour:
I just saw your implementation and I like the idea of giving the server side error back to the form.
But about the code, I have suggestion which might be simpler to implement, test and maintain.
In order to get the server side error we have to submit the form and get the feedback. Why don't we use the same function to get back the errors as the `props.onSubmit` return. Here is my suggestion, in the `Form.js`, in the `onSubmit` function, after the validation and custom validation, it execute the `props.onSubmit` and here is where we can give back the server side error (if any) back to the original `onSubmit` function : 

```
    setState(this, { errors: [], errorSchema: {} }, () => {
      if (this.props.onSubmit) {
        const {serverErrors, serverErrorSchema}  = this.props.onSubmit({ ...this.state, status: "submitted" });
        //new code
        //some how we have to be sure that the error objects returned from the props.submit function match the desired format
        if (serverErrors && serverErrorSchema) {
          setState(this, { errors: serverErrors, errorSchema: serverErrorSchema });
        }
        //new code end
      }
    });
```

what is the others opinion?

davidbarratt:
I don't like that because that's returning state via a callback rather than //passing// state via a prop (which is the typical way you pass state to another component in React). Also, I think users should be able to modify the errors whenever they want, not just on submit.

Jonjoe:
Hey guys! Whats holding this PR up? Anything I can do to help move it along? :)

olzraiti:
This would solve #401 #757 #197 neatly!

epicfaace:
@davidbarratt @olzraiti Could you explain how exactly being able to modify `errorSchema` would let you do async validation? Maybe a sample code snippet would be useful

epicfaace:
@davidbarratt just checking on this? Tests would help see how exactly this feature is used.

msalvi-cartera:
Is there any update on this change ? @epicfaace

davidbarratt:
I'm sorry I don't have time to update this, if someone else would like to take it over, that would be great. :)

Andi24:
Is this PR still relevant in light of https://github.com/rjsf-team/react-jsonschema-form/issues/1389?

epicfaace:
No, it's not. Thanks for pointing that out!

PR #876, opened 2018-03-26
added passing of raw errors to ArrayField template and components

llamamoray:

When I made change #826 I overlooked the fact that raw errors weren't being passed down from ArrayFields. I only came accross this when trying to integrate these errors into my custom components and ArrayFieldTemplate and seeing array level errors not being displayed (validation errors like min/max items etc).

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -366,6 +366,7 @@ class ArrayField extends Component {
       onBlur,
       onFocus,
       idPrefix,
+      rawErrors,
     } = this.props;
     const title = schema.title === undefined ? name : schema.title;
     const { ArrayFieldTemplate, definitions, fields, formContext } = registry;
@@ -411,6 +412,7 @@ class ArrayField extends Component {
       TitleField,
       formContext,
       formData,
+      rawErrors,
     };
 
     // Check if a custom render function was passed in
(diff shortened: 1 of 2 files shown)


llamamoray:
@glasserc would really appreciate if you could take a look at this PR... something I overlooked in my change a few weeks ago - sorry!

edi9999:
Thanks for this MR , seems good to me !

edi9999:
I don't have the rights to push a new version, but I would be interested in having them ? Is that possible @glasserc ?

glasserc:
Sorry for dropping the ball on this, let me cut a new release.

Do you think we will have a similar issue with ObjectFields, @llamamoray ?

glasserc:
I've released this as v1.0.3.

edi9999:
My username is the same as github,  edi9999

llamamoray:
@glasserc I can't think of any errors that would be from an ObjectField that wouldn't be a duplication of errors produced by Boolean, Schema, String, and Array Fields.

Is there anything that you can think of?

glasserc:
At the moment, I don't think there are any, but I think we could add some in the future, either on purpose or accidentally, and I think we should be prepared for it by handling it in the ObjectField case (the way we do with the ArrayField case).

PR #882, opened 2018-04-04
Updated README to mention the support for JSON Schema compliant enums

r4j4h:
Updated README to mention the support for JSON Schema compliant drop-down enums

PR #581 brought in support for `anyOf`-powered enums and added an Alternatives example in the playground but it still requires one to be aware of the difference in enums in that example or to find Issue #532 to understand how to use them so I added a section to the README.

Notably this was a task "I've updated docs if needed" on #581 that was left undone and this is trying to make that task done. :)

Although I added a new sub-section I did not add it to the Table of Contents as it still applies to the header present in the ToC but if you feel that would be good to do then I totally can. In that case I think we should probably add a sub-section title to the existing enumNames approach for clarity but I'm not sure what it should be called.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -559,6 +559,51 @@ This will be rendered using a select box that way:
 
 Note that string representations of numbers will be cast back and reflected as actual numbers into form state.
 
+#### Alternative JSON-Schema compliant approach
+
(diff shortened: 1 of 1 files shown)


edi9999:
Thanks for this doc, I actually didn't know myself that you can create selects without enumNames

PR #883, opened 2018-04-05
Add idPrefix option for oneOf

edi9999:

This is the follow up for https://github.com/mozilla-services/react-jsonschema-form/pull/806/files

It now correctly also handles idPrefix when using oneOf in the jsonSchema

### Checklist


src/components/fields/SchemaField.js
```diff
@@ -164,7 +166,12 @@ function SchemaFieldRender(props) {
     formContext,
     FieldTemplate = DefaultTemplate,
   } = registry;
+  let idSchema = props.idSchema;
   const schema = retrieveSchema(props.schema, definitions, formData);
+  idSchema = mergeObjects(
+    toIdSchema(schema, null, definitions, formData, idPrefix),
+    idSchema
+  );
```
glasserc: Why do we need to generate another `idSchema` at this level? What's wrong with the one that was generated for us at toplevel? That one was generated from the resolved schema, so it should have the `oneOf` accounted for, shouldn't it?

I'm a little concerned that we're restarting with `id=null` here -- won't this lose our place in the hierarchy?

src/components/widgets/BaseInput.js
```diff
@@ -4,6 +4,10 @@ import PropTypes from "prop-types";
 function BaseInput(props) {
   // Note: since React 15.2.0 we can't forward unknown element attributes, so we
   // exclude the "options" and "schema" ones here.
+  if (!props.id) {
+    console.log("No id for", props);
+    throw new Error(`no id for props ${JSON.stringify(props)}`);
+  }
```
glasserc: I think it would be better for this code not to come between the comment and the code the comment describes.

src/components/fields/SchemaField.js
knilink: Same question here, removing this piece of code doesn't seem to break any test.

PR #884, opened 2018-04-05
Fixed multiplicative errors on schema dependencies

damian-codilime:

When schema have dependencies with multiple branches, passing whole schema to validator will generate extraneous errors.
Here we are stripping schema from branches that we don't need to validate - since they are not present in the form.

Relates to the ticked i have created: https://github.com/mozilla-services/react-jsonschema-form/issues/829

### Checklist


edi9999:
Very good ! Thanks for the regression test too !

Issue #837, opened 2018-02-12
validation error when using format property


locales-wellington:
### Prerequisites

- [x ] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [ ] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Introduction of ajv validation prevents form submission when using unknown 'format' property

### Steps to Reproduce

https://jsfiddle.net/hdp1kgn6/1/

Add format: "textarea" to description and try and submit form

#### Expected behavior

I can add custom fields  which will format ui based on format property in data and not prevent submission

eg.

```
description: {type: "string". format: "textarea"}
```

```
const fields = {
 customField: CustomFieldComponent
}
const uiFields = { desc: { 'ui:field': 'customField'  };
```
#### Actual behavior

Error on form submission (ajv validation)
Uncaught Error: unknown format "textarea" is used in schema at path "#/properties/description"

### Version

1.0.x

You can usually get this information in your `package.json` or in the file URL if you're using the unpkg one.

locales-wellington:
7.2 "Implementations MAY add custom format attributes."
AJV allows the addition of custom formats [https://github.com/epoberezkin/ajv#formats](https://github.com/epoberezkin/ajv#formats) so it seems that it would be sensible to allow implementations using react-jsonschema-form to add these?

I am using ajv with react-jsonschema-form 0.51 and the 1.0.0 release breaks my implementation by not allowing that option.

glasserc:
You can already add widgets for any format you like as part of the registry.

locales-wellington:
But not custom formats in the schema correct? Because that will fail validation. For example: I have a schema that includes a format 'markup' for a string to display a markup editor. I should be able to add that custom format to the validator (ajv) so I don't get a validation error. The validator ajv that has been added provides that option so it seems odd that the react-schema-form api would not?

glasserc:
Yes, someone using rjsf should be able to interact with the ajv validator directly. See #783, #794.

locales-wellington:
Ah, righto - thanks :)

PR #889, opened 2018-04-11
Allow custom formats

harouny:
Related to issue: validation error when using format property #837 

- Allow ajv validation for custom (or unknown) schema formats i.e. (int32, int64. etc).
- Show fields with unknown formats using the default widget based on the type (same behavior as not passing format at all).
- Fix prettier setup as it didn't work as its meant to be in my box (possibly because its Windows).

### Checklist


package.json
```diff
@@ -7,8 +7,8 @@
     "build:lib": "rimraf lib && cross-env NODE_ENV=production babel -d lib/ src/",
     "build:dist": "rimraf dist && cross-env NODE_ENV=production webpack --config webpack.config.dist.js --optimize-minimize",
     "build:playground": "rimraf build && cross-env NODE_ENV=production webpack --config webpack.config.prod.js --optimize-minimize && cp playground/index.prod.html build/index.html",
-    "cs-check": "prettier -l $npm_package_prettierOptions '{playground,src,test}/**/*.js'",
-    "cs-format": "prettier --jsx-bracket-same-line --trailing-comma es5 --use-tabs false --semi --tab-width 2 '{playground,src,test}/**/*.js' --write",
+    "cs-check": "prettier -l \"{playground,src,test}/**/*.js\"",
```
harouny: I need someone to check that on lunix box just to be sure

src/utils.js
```diff
@@ -4,52 +4,67 @@ import validateFormData from "./validate";
...
+    },
   },
   integer: {
-    text: "TextWidget",
-    select: "SelectWidget",
-    updown: "UpDownWidget",
-    range: "RangeWidget",
-    radio: "RadioWidget",
-    hidden: "HiddenWidget",
+    default: "TextWidget",
```
harouny: I feel like `integer` default widget should be `UpDownWidget` but that's a breaking change, we can do that next major release.

package.json
```diff
@@ -19,7 +19,13 @@
     "tdd": "cross-env NODE_ENV=test mocha --compilers js:babel-register --watch --require ./test/setup-jsdom.js test/**/*_test.js",
     "test": "cross-env NODE_ENV=test mocha --compilers js:babel-register --require ./test/setup-jsdom.js test/**/*_test.js"
   },
-  "prettierOptions": "--jsx-bracket-same-line --trailing-comma es5 --semi --tab-width 2",
+  "prettier": {
+    "jsxBracketSameLine": true,
+    "trailingComma": "es5",
+    "useTabs": false,
```
harouny: Some of those options unnecessary passing the default values, I just passed them here because they used to be in `cs-format`

package.json
harouny: I guess cs-check must have run successfully on Travis so all good.

glasserc:
What does "custom" formats mean for you? How are you defining these "custom" formats and what behavior would you expect them to have?

I don't really like the approach taken in this PR. You can already add widgets for any format by putting them in your registry, or equally you could use uiSchema to use a custom field. Since other formats are not defined by JSON Schema, there's no way to actually validate input against them, which means there isn't really much difference between just using a custom field for those fields. At the same time, this PR weakens the protection against typoing a format like `uri` as `url`.

glasserc:
I don't think rjsf should necessarily support formats defined by OpenAPI, though I think it's good if a developer can add support for it themselves. I disagree that this is "formats not known at development time".

The JSON Schema spec says that validation *should* succeed for formats that are not among the ones that have been defined, and that we *should* offer an option to disable validation for the "format" keyword. But I think that behavior is bad. I think the maintainers of `ajv` agree with me, because in the README, about `unknownFormats: "ignore"`, is written "This option is not recommended, as it allows to mistype format name and it won't be validated without any error message. This behaviour is required by JSON Schema specification." Anyhow, because both of those requirements are given with "should", we're in compliance already and I don't see a great reason to add more complexity to support this feature.

Any developer who wants to add support for specific formats should probably follow #783 and add custom fields for the formats they want to support.

harouny:
> The JSON Schema spec says that validation should succeed for formats that are not among the ones that have been defined, and that we should offer an option to disable validation for the "format" keyword. But I think that behavior is bad.

I think if we are implementing a  spec, we should implement the spec correctly, otherwise integration with other components implementing the same spec will be nightmare. Plus the spec is there to make sure it is solving generic and wide range set of problems that a specific spec implementer might not even think about. Anyhow, thanks for your input.

PR #892, opened 2018-04-12
Update prettier and fix files

edi9999:

The ci is failing currently.

This is due to the new release of prettier a few hours ago : 1.12 .0 https://github.com/prettier/prettier/releases

I've updated the version of prettier explicitly, and runned cs:format command. The tests are all green with this MR


glasserc:
Thanks!

PR #894, opened 2018-04-13
Added Time input Widget as new String format

r4j4h:

react-jsonschema-form supports date and date-time String formats but not time itself while the underlying JSON parser used, ajv, supports the time format.

This relates to https://github.com/mozilla-services/react-jsonschema-form/issues/878 in that the problem sounds like what I ran into and solved for in Chrome and noted solution at https://github.com/r4j4h/react-jsonschema-form/commit/0a39e3c8568263ac099e38b88c54a393f3a317e8#diff-b936f552ddd7fd586fc97f5b3ac7975cR42. I recreated locally and will try the same solution and push up another PR if it works to solve it.

This relates to https://github.com/mozilla-services/react-jsonschema-form/issues/496 as it punts on time zones. I thought I read something that confirmed time inputs adhere to `partial-time` from [RFC3339](https://tools.ietf.org/html/rfc3339#section-5.6) but I can't find it now so maybe it is `full-time` - we can fix that now or in a subsequent PR as desired.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,8 @@ The built-in string field also supports the JSONSchema `format` property, and wi
 - `data-url`: By default, an `input[type=file]` element is used; in case the string is part of an array, multiple files will be handled automatically (see [File widgets](#file-widgets)).
 - `date`: By default, an `input[type=date]` element is used;
 - `date-time`: By default, an `input[type=datetime-local]` element is used.
+- `time`: By default, an `input[type=time]` element is used.
+
 
 ![](http://i.imgur.com/xqu6Lcp.png)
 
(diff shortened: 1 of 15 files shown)


r4j4h:
Odd, cs-format and cs-check both pass on my system. I am running in a container running node v9.10.1 and it's running in a VM but my actual host is a Mac. I've run into capitalization issues before where Linux is case sensitive and Mac isn't, perhaps that is the cause? I can't tell, but my other PR passed. :(

r4j4h:
Yeah I'm at a loss as to why CI hates my file 😢

r4j4h:
Great catch!! I had a npm lockfile in place and even after removing it I was at 1.11.1 but after a manual `npm install prettier` I am at 1.12.0 and got the same output as the CI 😌 

Also you are probably aware already but just incase I also added some background on the reasoning behind the seconds/milliseconds hacking in the other MR 895 https://github.com/mozilla-services/react-jsonschema-form/pull/895#issuecomment-381290620

Issue #878, opened 2018-03-29
Bug in native datetime widget in playground


kwakwaversal:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

The DateTimeWidget does not allow manually editing the year (and sometimes day/month) via the keyboard once a valid date has been set.

### Steps to Reproduce

1. Open [playground](https://mozilla-services.github.io/react-jsonschema-form/) and go to "Date & time" tab (alternatively see this [fiddle with just the datetime](https://jsfiddle.net/f2y3fq7L/271/)
2. Fill out the native datetime widget with a valid date
3. In the year type `0` (which will reset the year to `0000`), then try typing in `2018` as the year. It keeps getting reset so you can't get past entering the first two digits (if you're *really* quick).

#### Expected behavior

I'd expect the input element to allow you to type in the year (month, day etc.).

#### Actual behavior

You cannot enter more than 2 digits before it resets itself.

### Version

* Whatever version is in the playground.
* Chrome: Version 64.0.3282.186 (Official Build) (64-bit)
* Timezone: BST

edi9999:
Indeed, I can reproduce this issue.

Do you think you can find a fix and create a PR ?

kwakwaversal:
@edi9999 I tried to fix it but couldn't narrow it down after littering a few debug statements around in the `DateTimeWidget` and elsewhere.

If you could point me in the right direction I might be able to fix it. I can only assume that the "onChange" firing after every keystroke the widget it attempting to validate the value but fails, ultimately resetting it. I don't know the internals of RJF enough to start doing things like firing delayed `onChange` events.

r4j4h:
Hi, I was albe to reproduce this issue as well and I found the root cause in Chrome. I found a fix that works there, but am unsure if it is the right path.

heath-freenome:
After playing around with this in the current playground, this appears to be fixed. Closing

PR #895, opened 2018-04-13
Potential fix for DateTime Widget #878

r4j4h:
Eagerly truncating optional seconds/milliseconds when ambiguous to improve behavior in browsers when editing a DateTime widget populated with data

https://github.com/mozilla-services/react-jsonschema-form/issues/878 exists and is eerily similar to a problem I encountered while working on https://github.com/mozilla-services/react-jsonschema-form/pull/894. I recreated locally and the fix worked.

I say potential because this workaround may have side effects in other browsers that don't eagerly truncate the seconds/milliseconds when they are potentially optional.

### Checklist

### DIFF
--- a/src/components/widgets/DateTimeWidget.js
+++ b/src/components/widgets/DateTimeWidget.js
@@ -23,7 +23,16 @@ export function utcToLocal(jsonDate) {
   const ss = pad(date.getSeconds(), 2);
   const SSS = pad(date.getMilliseconds(), 3);
 
-  return `${yyyy}-${MM}-${dd}T${hh}:${mm}:${ss}.${SSS}`;
+  // Users can specify precision using step attribute otherwise (some) browsers automatically trim off when 0
+  // and this value difference breaks the cursor position from ambiguity in the change diff
+  // For example if an existing value is set and user focuses into the year field and types 1957, they
+  // go from 0001 and inbetween the 1 and the 9 the field renders current value in a format browser changes and
+  // so react sees the values as diff and resets cursor resulting in 0009 and then 0005 instead of the intended 1957.
+  return SSS === "000"
+    ? ss === "00"
+      ? `${yyyy}-${MM}-${dd}T${hh}:${mm}`
+      : `${yyyy}-${MM}-${dd}T${hh}:${mm}:${ss}`
+    : `${yyyy}-${MM}-${dd}T${hh}:${mm}:${ss}.${SSS}`;
 }
 
 export function localToUTC(dateString) {


glasserc:
Hmm, this is interesting. Your fix definitely fixes it, but I still don't really understand how. Could you please elaborate on what's going on without the fix?

I feel like there are a bunch of other interesting errors in the datetime widget. For instance, if the widget already shows a correct date, and then you enter an invalid year (e.g. "19834"), it completely loses its state. I'm not really sure that using React to manage this kind of widget like this is a great idea. I tried using `defaultValue` instead of `value`, the way you might do it in Elm, and that seemed to work a lot better, but requires more invasive changes.

r4j4h:
Sure, as I understand it the browser is taking liberal use of what constitutes a valid time string when no "step" attribute is defined and our function that converts the formData value to the HTML value attribute handed to the input is not following the same liberal truncation rules and so the switch between them is confusing React or the browser's native input control's cursor positioning.

https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-time-string states (emphasis mine):
> 4. **If second is nonzero**, or **optionally** if second is zero:
> A U+003A COLON character (:)
> Two ASCII digits, representing the integer part of second, in the range 0 ≤ s ≤ 59
> If second is not an integer, or optionally if second is an integer:
> A U+002E FULL STOP character (.)
> One, two, or three ASCII digits, representing the fractional part of second

and right above "Using form inputs" https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#Using_time_inputs MDN confirms:

> When a form including a time input is submitted, the value is encoded before being included in the form's data. The form's data entry for a time input will always be in the form "name=hh%3Amm", or "name=hh%3Amm%3ass" if seconds are included (see Using the step attribute).

So if my understanding is correct then the following is happening with the React components:

1. User has valid date and so components state is valid date, let's say 10:15 PM.
2. User wants to change minutes from 15 to 20 and so clicks into minutes and begins typing "2".
3. React catches the input change and when run through the code the returned value actually changes it to 10:02:00 PM.
4. I could be wrong but I think React JSON Schema Forms renders asynchronously and so the browser gets a chance to override the value using its logic which truncates the added **:00** optional seconds bit from 10:02:00 PM to the truncated 10:02 PM.
5. The user finishes typing "0" for the second character of "20" but React then sees this as radically different (10:15 PM -> 10:**02:00** PM) and not just a simple one character change (10:15 PM -> 10:**02** PM) and so it or the browser gets confused and understandably resets the cursor position. With the cursor position reset the newly entered character goes into an unexpected place and we get the "overwrite last character" behavior.

Based on this understanding my change attempts to follow this behavior a little more closely by excluding the milliseconds and seconds when they are zero. I honestly hate it, but I think it is the root cause because if we add the "step" attribute the appropriate numbers always show and if the step format matches our function's output (`1` for "seconds" resolution) the bug disappeared.

All that being said my fears are what other browsers will do.. are there some that don't eagerly truncate seconds/millis when "steps" is left undefined? What about browsers that fall back to plain inputs? The MDN page has lots to say. I think our widgets here have a unique ability to help enforce some of that pattern validation stuff in the onChange handler and value transformer functions which is a nicer place than if we were left on our own with only the native inputs! I hope to make an "alt-time" variant that uses dropdowns as well if time permits - I wanted to reuse the time portion of th existing alt one but I think we may want to re-create the 12-hour AM/PM behavior that browsers seem to prefer on the "time" input.

glasserc:
OK, thanks for the detail. It seems to me like the missing piece is step 4 above, where I think what's happening is that React noticed that the value is different than the input's existing value, and updates or replaces the element, which resets its internal state, causing subsequent keypresses to be interpreted as though they were the first keypress. Your fix works around this by trying to match the existing value as closely as possible. But as you note, it seems quite fragile, and we have no guarantee if this behavior will remain in future versions of Chrome, let alone other browsers.

Additionally, I think there are other issues that we won't be able to solve here, like the fact that if you add a fifth digit to the year field, the input gets reset. That's pretty surprising. I'm not a React expert but it seems to me that this kind of input element has too much internal state for us to be using it in the traditional "controlled" React component style. I think our options are:

1. Merge a PR like this. Status quo is mostly unchanged. Other quirks regarding use of datetime remain. This bug may rear its head again on other browsers or on later versions of Chrome.
2. Move to using `defaultValue` for this field and leaving `value` alone. The field's state is managed entirely by the DOM. Strange quirks go away entirely. Unfortunately, we will no longer be able to "update" the field's value by updating the formData, which makes this component much less "React-y".
3. Move to using `ref`, as documented in https://reactjs.org/docs/uncontrolled-components.html. I've never used this before and have no idea what it does or how it works, but I imagine this means we have to do a lot more work to get `value` out of the "uncontrolled" input, and then decide for ourselves when we want to update the value of the component, probably using some home-grown ad hoc logic that behaves more-or-less like React's, but possibly leading to a bunch of strange behaviors and bugs if we mess it up.

None of these approaches sound really great to me. What do you think? The only thing I think we should definitely do is add a note in the README explaining how fragile these native input widgets are.

r4j4h:
1. I like this route. We can do some browser testing and try to cover or document discovered cases. Unfortunately after poking around at the other options I think this or patching `ajv` (see rest of this comment) is the smoother way forward.

2. Using `defaultValue` almost works.
 
    We can use the `onChange` handler to keep our `formData` in sync with user changes.
    
    For keeping input `value` in sync with `formData` changes we can use `UNSAFE_componentWillReceiveProps` and to get the component to see the updated `defaultValue` we can provide and mutate a `key` property to get React to re-render the uncontrolled component back to its defaultValue.

    However, this resetting of the `key` to force redraw messes up rapid user input like the original fix was trying to correct. Not to mention this requires changes to `BaseInput` to support not always passing `value` and to support passing `defaultValue`. Or an alternative `BaseInput` that provides that behavior to be created.

3. Moving to `refs` works though. It goes in the same direction as Option 2 because uncontrolled components cannot have `value` directly set.
 
    We can define `onChange` handler and we can use `UNSAFE_componentWillReceiveProps` to set the `ref`'s input value. There we can directly update the `ref`'s DOM input's `value` to keep things in sync with `formData` without tricking redraws by changing `key`.
    
    But we also need to modify `BaseInput` to support `refs`. That means also not passing `value` and instead passing `defaultValue`, configuring a `ref`
reference, and using the `ref` reference in `UNSAFE_componentWillReceiveProps` to update the uncontrolled component's value accordingly (or we can keep passing `value` and have the `BaseInput` handle all the `ref`-related bits).

    Either way to support `refs` we also need to change `BaseInput` to not be a stateless functional component and to not pass `value` in favor of `defaultValue` (like option 2). It already passes `ref` but React does not allow to pass `value` when using `ref` or it throws warnings). Alternatively a specialized ref supporting `BaseInput` could be created for this purpose. In fact the only thing that *needs* the `ref` in this case is the `BaseInput` itself.

    This approach works but still has problems - namely that *runs into the exact same problem due to text formatting for ajv*. If we do not clean up the text
for ajv, the bug does not appear, but in some cases the form cannot be submitted.

So the real problem is beginning to look like the JSON Schema parser not supporting the browser's preferred time format of `HH:MM`.

A seemingly simple solution is to fix ajv so that it accepts HH:MM as a valid time string, or we handle that one edge case more
explicitly for AJV's sake for validation and less so for the browsers. Here are two ways we could go about that:

1. Main Option: Fix problem at source in AJV 

    They are going off of [RFC3339](https://tools.ietf.org/html/rfc3339#section-5.6) which is the same ballpark but does not have the optional clause around seconds. I think `partial-time`.
    
    > partial-time    = time-hour ":" time-minute ":" time-second
    
    The change would be easy enough:
    
    The `time` regular expression at 
    
    https://github.com/epoberezkin/ajv/blob/797dfc8c2b0f51aaa405342916cccb5962dd5f21/lib/compile/formats.js#L36
    
    would be changed from
    
    ```
      time: /^(?:[0-2]\d:[0-5]\d:[0-5]\d|23:59:60)(?:\.\d+)?(?:z|[+-]\d\d:\d\d)?$/i,
    ```
    
    to
    
    ```
      time: /^(?:[0-2]\d:[0-5]\d:[0-5]\d?|23:59:60|23:59)(?:\.\d+)?(?:z|[+-]\d\d:\d\d)?$/i
    ```
    
    But since only `time-numoffset` drops the `time-second` I am guessing they will be against this change. 
    
    > time-numoffset  = ("+" / "-") time-hour ":" time-minute
    
    So,

2. Alternative Option: Fix wrapping of AJV 

    Instead of fixing it at the ajv layer we could keep a shim in React JSON Schema Forms that adds the `:00` if necessary
before passing to AJV. This option has some similarities with my current implementation, aka option 1, but instead of trying to transparently handle swapping proper data with the browser it would prefer the browser data all the time except validation time when it would swap 'HH:MM' to 'HH:MM:SS' to prevent validation failure. Because of that it feels a little less brittle than hoping all browsers behave the same way, as it just relies on ajv behaving the same way instead.. that at least sounds convincing as I type this. 😊 

What a pain. Either way I think that noting something about this sort of behavior in the docs is a great idea.

PR #896, opened 2018-04-13
Prevent stale form errors appearing

laithalissa:
Call submit as setState callback to prevent stale errors appearing.

Stale form errors can sometimes appear when invalid fields are submitted, corrected, resubmitted, followed by rerendering the form. 

fixes mozilla-services/react-jsonschema-form#315

I've just seen that somebody else has made an identical change, but it was from last year so I'll continue with this pull request anyway. I've not added a test because I'm not sure it's really relevant. If somebody wants to suggest a test, I'll add it.

### Checklist

* [ x ] **I'm adding or updating code**
  - [ X ] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -147,10 +147,11 @@ export default class Form extends Component {
       }
     }
 
-    if (this.props.onSubmit) {
-      this.props.onSubmit({ ...this.state, status: "submitted" });
-    }
-    this.setState({ errors: [], errorSchema: {} });
+    this.setState({ status: "initial", errors: [], errorSchema: {} }, () => {
+      if (this.props.onSubmit) {
+        this.props.onSubmit({ ...this.state, status: "submitted" });
+      }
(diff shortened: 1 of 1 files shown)


glasserc:
What was the other PR you saw? Was it #844? That was only from a month ago, not a year.. Are there others?

glasserc:
I believe this is fixed by #844. If you find other PRs/issues about this issue, please feel free to let me know.

laithalissa:
Hi glasserc. Sorry for the late reply, I've not been working around this for a while and I had to find time to refresh my context on this issue. Looks like I was referring to a [comment](https://github.com/mozilla-services/react-jsonschema-form/issues/315#issuecomment-256930949), rather than a PR. I'll double check we're up to date with that change when I get a chance.

PR #909, opened 2018-04-25
Provide user errors to transformErrors

pushred:

Hello!

We only display one error at a time in our implementation in an effort to meet some UX goals:

- Compact forms
- Avoid jarring shifts of controls as messages come and go
- Avoid overeager messaging (e.g. pattern errors before a length constraint is met)

The order of errors matters. The default order coming from AJV seems OK, but may be so by chance. We can apply our own sort using transformErrors, but an issue remains: custom validation errors are always appended to the output.

It's reasonable to expect that custom validation errors will be of higher quality than those from AJV. So we want to order them first. However the only place we have access to the entire error list, the field template, lacks metadata to distinguish custom messages from others.

#### Solution

To provide an opportunity to sort the entire error list I have updated validateFormData to call transformErrors _after_ customValidate when it is provided. I flatten the custom validation schema in order to merge the custom errors into the flat array returned by AJV. Custom errors have a similar structure to AJV's errors and are named `custom` for selection.

### Checklist

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -14,7 +14,7 @@ ajv.addFormat(
   /^(#?([0-9A-Fa-f]{3}){1,2}\b|aqua|black|blue|fuchsia|gray|green|lime|maroon|navy|olive|orange|purple|red|silver|teal|white|yellow|(rgb\(\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*,\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*,\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*\))|(rgb\(\s*(\d?\d%|100%)+\s*,\s*(\d?\d%|100%)+\s*,\s*(\d?\d%|100%)+\s*\)))$/
 );
 
-import { isObject, mergeObjects } from "./utils";
+import { isObject } from "./utils";
 
 function toErrorSchema(errors) {
   // Transforms a ajv validation errors list:
@@ -52,13 +52,13 @@ function toErrorSchema(errors) {
       }
       parent = parent[segment];
     }
-    if (Array.isArray(parent.__errors)) {
+    if (Array.isArray(parent.__errors) && message) {
       // We store the list of errors for this node in a property named __errors
       // to avoid name collision with a possible sub schema field named
       // "errors" (see `validate.createErrorHandler`).
       parent.__errors = parent.__errors.concat(message);
-    } else {
-      parent.__errors = [message];
+    } else if (!Array.isArray(parent.__errors)) {
+      parent.__errors = message ? [message] : [];
     }
     return errorSchema;
   }, {});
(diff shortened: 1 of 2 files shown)


src/validate.js
```diff
@@ -52,13 +52,13 @@ function toErrorSchema(errors) {
       }
       parent = parent[segment];
     }
-    if (Array.isArray(parent.__errors)) {
+    if (Array.isArray(parent.__errors) && message) {
```
epicfaace: What is this code (changes made on lines 50-61) meant for?

src/validate.js
```diff
@@ -162,22 +205,36 @@ export default function validateFormData(
...
   if (typeof customValidate !== "function") {
+    const errorSchema = toErrorSchema(errors);
     return { errors, errorSchema };
   }
 
+  // Apply custom validations and allow transform of merged result
   const errorHandler = customValidate(formData, createErrorHandler(formData));
   const userErrorSchema = unwrapErrorHandler(errorHandler);
-  const newErrorSchema = mergeObjects(errorSchema, userErrorSchema, true);
+  const userErrors = flattenErrorSchema(userErrorSchema).map(error =>
```
epicfaace: Is there a reason why you're flattening & merging regardless of whether `transformErrors` is defined (as opposed to doing so only when `transformErrors` is defined)?

test/validate_test.js
```diff
@@ -112,63 +156,79 @@ describe("Validation", () => {
...
+        const result = validateFormData(
+          { foo: 42, [illFormedKey]: 41 },
+          schema,
+          validate,
+          transformErrors
+        );
+        errors = result.errors;
+
+        expect(errors).not.to.be.empty;
+        expect(errors[0].stack).to.equal("foo: custom");
```
epicfaace: Can you add an assertion to make sure that the `errors[0].name` is also equal to `custom`?

epicfaace:
Hey @pushred , could you respond to my comments when you get a chance? Would like to get this merged.

pushred:
@epicfaace sure I'll review next week, especially the backwards compatibility concern you raised.

src/validate.js
pushred: I don't think so, I'll conditionally do this

src/validate.js
pushred: This counters the message-less paths inserted at L129-133 that are used to create empty `__errors` arrays at those nodes that don't have any errors when the flattened data is expanded to the hierarchical error schema object. Without this a couple tests failed:

https://github.com/mozilla-services/react-jsonschema-form/blob/aebfab9baa94d35176b7197ecf080e2fc1258dcf/test/validate_test.js#L488-L542

https://github.com/mozilla-services/react-jsonschema-form/blob/aebfab9baa94d35176b7197ecf080e2fc1258dcf/test/validate_test.js#L544-L575

It [seems to be intentional](https://github.com/mozilla-services/react-jsonschema-form/blob/aebfab9baa94d35176b7197ecf080e2fc1258dcf/src/validate.js#L93
) that `__errors` are inserted on every node regardless of errors.

It's been awhile so I'm not 100% sure how these transformations work anymore, I'll have a deeper look next week and document the data structures provided to and expected out of transformErrors and customValidate, including the custom error handling I'm implementing here.

test/validate_test.js
pushred: That's actually not available here because the schemas are [converted to an errorList](https://github.com/mozilla-services/react-jsonschema-form/blob/18503088e57aaf11af0cc64d7a70d0e330c1127a/src/validate.js#L234-L239) whose items only have a `stack` prop for the return value of  `validateFormData`

`transformErrors` functions are given the flattened data that includes the `name` and here on L181 I sort by that. The `stack` assertion validates that the sorting worked. An AJV `foo: should be string` error comes first without sorting.

pushred:
> This might be a backwards incompatible change though, but the rationale make sense.

@epicfaace can you confirm that it's the reordering of the transformErrors call after the customValidate that you're concerned could be a breaking change? I can see that doing something unexpected in cases I'm not thinking about.

If so, to err on the side of caution do you have a major version coming up that you are thinking of including this in? If not, is another config prop warranted to control this behavior?

src/validate.js
pushred: So you added some docs in this area recently, haven't had a chance to review yet myself but will try to next week

epicfaace:
> > This might be a backwards incompatible change though, but the rationale make sense.
> 
> @epicfaace can you confirm that it's the reordering of the transformErrors call after the customValidate that you're concerned could be a breaking change? I can see that doing something unexpected in cases I'm not thinking about.
> 
> If so, to err on the side of caution do you have a major version coming up that you are thinking of including this in? If not, is another config prop warranted to control this behavior?

Exactly. According to the docs we say "If you need to change these messages or make any other modifications to the errors from the JSON Schema validation, you can define a transform function that receives the list of JSON Schema errors and returns a new list" -- so it seems that our API _does_ specify that transformErrors only transforms errors from the JSON Schema validation.

I think this would be better placed in the upcoming v2 version, as it might cause confusion if we do add another config prop. Could you update this PR to fix the merge conflicts in the meantime?

epicfaace:
Actually, looking at it, "JSON Schema validation" does sound a bit ambiguous -- it could be construed include custom schema validation as well. So maybe this change isn't backwards-incompatible (at least according to our published documentation) after all.

I can't really think of any meaningful use cases in which adding this change would break existing code; what were some of the use cases you had in mind @pushred ?

PR #916, opened 2018-05-05
Add initial support for if/then/else and not

### LINKED ISSUE #885: Add support for if/then/else from JSON Schema Draft 7
I recently learned about draft 7 and it's support for if/then/else which seems much easier to reason with than schema dependencies and oneOf (which we added support for not too long ago).

Should there be any concerns about pulling in draft 7 keywords? Does/Can rjsf even use ajv in draft 7 mode? Even if it doesn't/can't (yet) is there any concern about adding support for if/then/else?

Thank you.

mfulton26:
- if/then/else follows Draft 7 keyword definitions
- for special cases where you want to delete form data
  (e.g. based on some condition) `"not": {}` can be used
  to specify that nothing is allowed for a property.

Schema dependencies don't allow triggering conditional schema on the absence of some property and can sometimes be difficult to understand with their declarative nature (e.g. when using `oneOf`).

Add support for Draft 7's `if`/`then`/`else` makes it possible to define conditional schema attributes using any schema.

This also leads to cases where some input data should only be allowed and persisted in the form data when applicable and be removed when not applicable. As such, this change also includes very basic/limited support for the `not` keyword to specify that any associated value with a property key should be deleted via `"not": {}` (see examples and README for details).

Closes #885.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1749,9 +1749,6 @@ JSON Schema Draft 7 introduces support for `if`/`then`/`else as a more procedura
     }
   },
   "if": {
-    "required": [
-      "referrer"
-    ],
     "properties": {
       "referrer": {
         "const": "Other (please specify)"
--- a/README.md
+++ b/README.md
@@ -1737,35 +1737,37 @@ JSON Schema Draft 7 introduces support for `if`/`then`/`else as a more procedura
 {
   "type": "object",
   "properties": {
-    "referrer": {
-      "title": "How did you hear about us?",
+    "ownsCar": {
+      "title": "Do you own a car?",
       "type": "string",
       "enum": [
-        "Friends",
-        "Family",
-        "Internet",
-        "Other (please specify)"
+        "Yes",
+        "No"
       ]
     }
   },
   "if": {
+    "required": [
+      "ownsCar"
+    ],
     "properties": {
-      "referrer": {
-        "const": "Other (please specify)"
+      "ownsCar": {
+        "const": "Yes"
       }
     }
   },
   "then": {
     "properties": {
-      "referrerOther": {
-        "title": "Other",
-        "type": "string"
+      "carColor": {
+        "title": "What color is your car?",
+        "type": "string",
+        "default": "#ff0000"
       }
     }
   },
   "else": {
     "properties": {
-      "referrerOther": {
+      "carColor": {
         "not": {}
       }
     }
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1727,6 +1729,60 @@ Note that this is quite far from complete `oneOf` support!
...
+        "not": {}
+      }
+    }
+  }
+}
+```
+
+If the form data validates against the `if` schema then the `then` schema is merged into the form schema and rendered. An `else` schema can optionally be specified to be applied if the form data does not validate against the `if` schema.
+
+In the above example the `else` schema causes any previous input for `carColor` to be deleted when not applicable. Ommiting `"not": {}` would retain the value. See [not](#not) for more details.
```
glasserc: `Ommiting` should be `Omitting`.

src/utils.js
```diff
@@ -418,8 +418,49 @@ export function retrieveSchema(schema, definitions = {}, formData = {}) {
...
+        remainingSchema,
+        conditionalSchema || {}
+      );
+      return retrieveSchema(updatedSchema, definitions, formData);
+    } else {
+      // Use the remaining schema without merging anything with it
+      // Note that the remaining schema no longer contains any `then` or `else` schemas
+      // This is important so that and `if` schema that gets merged into the schema
+      // from a schema dependency or another if/then/else does not trigger an
+      // orphaned `then` or `else` schema.
```
glasserc: Should we error if `if` isn't an object?

elyobo:
> I also don't like that we have to explicitly remove fields rather than having them naturally fall away when the condition that added them is gone.

I have just run in to this with the existing dynamic schema dependencies (data related to a conditional field is submitted even if that field is not included in the current schema) and was about to open a bug for it; is that actually the expected behaviour @glasserc?

src/utils.js
```diff
@@ -418,8 +418,37 @@ export function retrieveSchema(schema, definitions = {}, formData = {}) {
   } else if (schema.hasOwnProperty("dependencies")) {
     const resolvedSchema = resolveDependencies(schema, definitions, formData);
     return retrieveSchema(resolvedSchema, definitions, formData);
+  } else if (schema.if || schema.then || schema.else) {
+    const {
+      if: ifSchema,
+      then: thenSchema,
+      else: elseSchema,
+      ...remainingSchema
+    } = schema;
```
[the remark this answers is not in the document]
nagaozen: Yes and Yes please xD

nagaozen:
I really need this feature... Hey @epicfaace, want me to open another PR to implement the changes proposed by @mfulton26 ? Removing the extra fields added by another conditional is paramount for my use case.

epicfaace:
Yes, @nagaozen , that would be great!

epicfaace:
Closing this in favor of #1666, which is more up-to-date

Issue #898, opened 2018-04-13
alt-date widget statically show years from 1900 to 2020


keul:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [ ] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

https://github.com/mozilla-services/react-jsonschema-form/blob/03b54d086036167d5e1aa970c9d4d103dd766617/src/components/widgets/AltDateWidget.js#L106

I need to provide a widget for selecting years from 1800 to 1880, but the `alt-date` widget is using a static list.

#### Expected behavior

Year select should be composable by providing a min and max props.

### Version

1.0.2

PR #917, opened 2018-05-07
Can now customize years shown when using alt-date widgets

keul:

See #898 

### Checklist

### DIFF
--- a/src/components/widgets/AltDateWidget.js
+++ b/src/components/widgets/AltDateWidget.js
@@ -53,6 +53,9 @@ class AltDateWidget extends Component {
     disabled: false,
     readonly: false,
     autofocus: false,
+    options: {
+      yearsRange: [1900, new Date().getFullYear() + 2],
+    },
   };
 
   constructor(props) {
@@ -100,10 +103,14 @@ class AltDateWidget extends Component {
   };
 
   get dateElementProps() {
-    const { time } = this.props;
+    const { time, options } = this.props;
     const { year, month, day, hour, minute, second } = this.state;
     const data = [
-      { type: "year", range: [1900, 2020], value: year },
+      {
+        type: "year",
+        range: options.yearsRange,
+        value: year,
+      },
       { type: "month", range: [1, 12], value: month },
       { type: "day", range: [1, 31], value: day },
     ];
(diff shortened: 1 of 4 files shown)


src/components/widgets/AltDateWidget.js
```diff
@@ -100,10 +103,15 @@ class AltDateWidget extends Component {
...
   get dateElementProps() {
-    const { time } = this.props;
+    const { time, options } = this.props;
     const { year, month, day, hour, minute, second } = this.state;
     const data = [
-      { type: "year", range: [1900, 2020], value: year },
+      {
+        type: "year",
+        range:
+          options.yearsRange || AltDateWidget.defaultProps.options.yearsRange,
```
keul: I don't like this change that much, however only using `options.yearsRange` is not working while documentation says:

> If you need to pass options to your custom widget, you can add a ui:options object containing those properties. If the widget has defaultProps, the options will be merged with the (optional) options object from defaultProps

This merge is not taking place here. If I don't provide any `yearsRange` to uiSchema options object only contains an `enumOptions` prop.

glasserc: Are you sure? There are a few tests that cover that the `ui:options` merge happens -- see uiSchema_test.js -- and when I take this code out, it seems to fall back safely to `[1900, 2020]` (i.e. this year plus 2).

keul: @glasserc I saw this test too, but trying to removing the `||` code break the demo app.

I investigated this a little more, seems that the issue is only with the _AltDateTimeWidget_ but not with _AltDateWidget_. That's weird as AltDateTimeWidget is only a very simple wrapper as far I can see.

Recap: if I try to rely on default values I have:

* _AltDateWidget_ with options `{yearsRange: Array(2), enumOptions: false}`
* _AltDateTimeWidget_ with options `{enumOptions: false}`

(I rebased on master to the sure of that, nothing changed).

glasserc: This is probably because of `AltDateTimeWidget` being a wrapper around `AltDateWidget` that doesn't have its own `defaultProps` set. I think the right thing to do here is to set `AltDateTimeWidget.defaultProps = AltDateWidget.defaultProps`.

keul: @glasserc uh! That was easy, sorry for not spotted this before.
Yes, it works properly now, PR updated!

keul:
@glasserc done!

glasserc:
Thanks!

rohitbishnoi:
This fix is not yet available on the CDN link given in the documentation.

Issue #823, opened 2018-01-24
Ordering of schema-depended fields


lingxiaoyang:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Hello there, thank you for the great work on this project. I am exploring the use of schema dependencies but I encounter a trouble in ordering the depended fields. I created a minimal example here: https://codesandbox.io/s/8kw2wq6z08

In this example, there are two selects. When every of them is selected "yes", a depended field will be triggered to display. So "Select A" triggers "Textbox A" and "Select B" triggers "Textbox B".

What I am trying to achieve is an ordering of "Select A", "Textbox A", "Select B", "Textbox B", but setting `ui:order` as is (you can modify the `uiSchema` to `uiSchema_not_working`) will throw the following error:

```
Invalid root object field configuration:uiSchema order list contains extraneous properties
```

If there is a way to correctly define the ordering, could you please kindly inform?

Copy-paste the schema here:
```json
{
  "type": "object",
  "properties": {
    "Show first name?": {
      "type": "string",
      "enum": ["yes", "no"]
    },
    "Show family name?": {
      "type": "string",
      "enum": ["yes", "no"]
    }
  },
  "dependencies": {
    "Show first name?": {
      "oneOf": [{
        "properties": {
          "Show first name?": {
            "enum": ["yes"]
          },
          "First name": {
            "type": "string"
          }
        }
      }, {
        "properties": {
          "Show first name?": {
            "enum": ["no"]
          }
        }
      }]
    },
    "Show family name?": {
      "oneOf": [{
        "properties": {
          "Show family name?": {
            "enum": ["yes"]
          },
          "Family name": {
            "type": "string"
          }
        }
      }, {
        "properties": {
          "Show family name?": {
            "enum": ["no"]
          }
        }
      }]
    }
  }
}
```

The uiSchema I was trying to set:
```json
{
  "ui:order": ["Show family name?", "Family name", "Show first name?", "First name"]
}
```

### Version

React 16.0.0
react-jsonschema-form latest

AdityaParameshwara:
Facing with the same issue

Pruxis:
You can use a wildcard like this:
```
    'ui:order': [
      'description',
      'method',
      'url',
      'headers',
[...]

PR #918, opened 2018-05-07
Make ui:order less strict

mfulton26:
Throwing an error when an extraneous property name is included
or when a property name is missing is problematic when ordering
properties in a dynamic schema (e.g. via schema dependencies,
etc.).

This change ignores extraneous property names and places places
missing properties at the end if not specified.

[Please describe them here]

### Checklist

### DIFF
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -100,6 +100,11 @@ class ObjectField extends Component {
       );
     }
 
+    // filter out any property names that are not currently applicable
+    orderedProperties = orderedProperties.filter(name =>
+      schema.properties.hasOwnProperty(name)
+    );
+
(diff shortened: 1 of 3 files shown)


glasserc:
Refs #823, #814. Per the discussion in the other threads, I think doing this unconditionally is dangerous. I'd much rather be smart about it.

mfulton26:
I just got done looking at those refs. Thank you. If I understand correctly, you would like for an error to be shown for any extraneous property name that is not in the schema properties nor any conditional schema properties defined in dependencies. That seems doable.

epicfaace:
> I just got done looking at those refs. Thank you. If I understand correctly, you would like for an error to be shown for any extraneous property name that is not in the schema properties nor any conditional schema properties defined in dependencies. That seems doable.

@mfulton26 I'm closing this for now, if you would like to implement the behavior mentioned above, please make another PR. Thanks

PR #927, opened 2018-05-14
Added `alt-time` Time input Widget as new String format

r4j4h:

react-jsonschema-form supports date and date-time String formats but not time itself while the underlying JSON parser used, ajv, supports the time format.

This adds a more browser-safe time input option through the already existing `alt` widget family.

#894 is an approach towards a "browser native" time input, while this is the more accessible drop down version.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,7 @@ Please note that while standardized, `datetime-local` and `date` input elements 
 
 - `alt-datetime`: Six `select` elements are used to select the year, the month, the day, the hour, the minute and the second;
 - `alt-date`: Three `select` elements are used to select the year, month and the day.
+- `alt-time`: Two `select` elements are used to select the hour and minutes.
 
 ![](http://i.imgur.com/VF5tY60.png)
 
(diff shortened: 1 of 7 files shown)


epicfaace:
Thanks for the PR, @r4j4h . I noticed that `AltTimeWidget.js` and `AltDateWidget.js` share a lot of common code, though. If you see the code for AltDateTimeWidget, it is using AltDateWidget under the hood and just passing a prop called `time`. https://github.com/mozilla-services/react-jsonschema-form/blob/51caab66b3895601d1a50ad5e2196948749d325b/src/components/widgets/AltDateTimeWidget.js#L7

Would it be possible to design AltTimeWidget so that it also similarly extends on the functionality of AltDateWidget, rather than writing the entire widget from scratch?

r4j4h:
Hi @epicfaace , sorry for the long delay. That sounds like a good request! I'm not currently using this code anymore but since it has a chance of getting merged in I wouldn't mind making that change. Not sure when I will have it done by, but hopefully soon.

epicfaace:
Ok, thanks!

Issue #805, opened 2018-01-10
v2 Roadmap


edi9999:
Currently this is just a list of PR that should be accepted for the v2 version (they have breaking changes).

1. https://github.com/mozilla-services/react-jsonschema-form/pull/767 (breaking-change)

However, any big topic that can't be implemented in v1 could be discussed in this thread too.

eberkund:
Perhaps making the templates framework agnostic can be something to focus on in v2?

i.e. Bootstrap v4, Material UI, etc.

epicfaace:
Closing this as some of the V2 issues have been added to this project: https://github.com/mozilla-services/react-jsonschema-form/projects/1

edi9999:
Ok, good idea !

PR #950, opened 2018-06-09
Refactor idSchema related code

knilink:

Simply refactoring, no functionality change.
The reasons is to generate a well formed idSchema at the root instead of re-generate it everytime in the children components.
Also with this change, users can customize idSchema easier.

_**!!!This change might potentially break custom array schema components rely on idSchema.!!!**_

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -619,35 +619,58 @@ export function shouldRender(comp, nextProps, nextState) {
   return !deepEquals(props, nextProps) || !deepEquals(state, nextState);
 }
 
-export function toIdSchema(
-  schema,
-  id,
-  definitions,
-  formData = {},
-  idPrefix = "root"
-) {
+export function toIdSchema(schema, id, definitions, formData, seperator = "_") {
   const idSchema = {
-    $id: id || idPrefix,
+    $id: id || "root",
   };
-  if ("$ref" in schema) {
+  if ("$ref" in schema || "dependencies" in schema) {
     const _schema = retrieveSchema(schema, definitions, formData);
-    return toIdSchema(_schema, id, definitions, formData, idPrefix);
+    return toIdSchema(_schema, id, definitions, formData, seperator);
   }
-  if ("items" in schema && !schema.items.$ref) {
-    return toIdSchema(schema.items, id, definitions, formData, idPrefix);
+  if ("items" in schema) {
+    for (const i in formData) {
+      const itemId = idSchema.$id + seperator + i;
+      if (Array.isArray(schema.items)) {
+        if (i < schema.items.length) {
+          idSchema[i] = toIdSchema(
+            schema.items[i],
+            itemId,
+            definitions,
+            formData[i]
+          );
+        } else if (schema.additionalItems) {
+          idSchema[i] = toIdSchema(
+            schema.additionalItems,
+            itemId,
+            definitions,
+            formData[i]
+          );
+        }
+        // exception: formData is longer the the given length
+      } else {
+        idSchema[i] = toIdSchema(
+          schema.items,
+          itemId,
+          definitions,
+          formData[i]
+        );
+      }
+    }
+    return idSchema;
   }
   if (schema.type !== "object") {
     return idSchema;
   }
+  formData = formData || {};
   for (const name in schema.properties || {}) {
     const field = schema.properties[name];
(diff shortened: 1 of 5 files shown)


test/utils_test.js
```diff
@@ -1068,13 +1074,11 @@ describe("utils", () => {
...
-          $id: "rjsf",
-          foo: { $id: "rjsf_foo" },
-          bar: { $id: "rjsf_bar" },
-        }
-      );
+      expect(toIdSchema(schema, "rjsf", schema.definitions, {})).eql({
+        $id: "rjsf",
+        foo: { $id: "rjsf_foo" },
+        bar: { $id: "rjsf_bar" },
+      });
```
glasserc: Previously `ui:rootFieldId` on the root element was not passed on to children; now I think it is. Is that intentional? This test almost seems like it was meant to ensure that behavior, but it's a bit hard to say for sure.

Assuming we don't care about preserving the `idPrefix` behavior, this section is called `"should handle idPrefix parameter"`, but there is no longer such a parameter, so maybe we should delete the whole thing.

src/utils.js
```diff
@@ -619,35 +619,58 @@ export function shouldRender(comp, nextProps, nextState) {
...
+          );
+        } else if (schema.additionalItems) {
+          idSchema[i] = toIdSchema(
+            schema.additionalItems,
+            itemId,
+            definitions,
+            formData[i]
+          );
+        }
+        // exception: formData is longer the the given length
```
glasserc: Do we have a preferred behavior for this case?

test/utils_test.js
knilink: This code here in the master branch
https://github.com/mozilla-services/react-jsonschema-form/blob/7c252de331cfa77c415adf49c04eeb25fd91d28b/src/utils.js#L621-L630
where `id=uiSchema["ui:rootFieldId"]` as shown bellow.
https://github.com/mozilla-services/react-jsonschema-form/blob/7c252de331cfa77c415adf49c04eeb25fd91d28b/src/components/Form.js#L65-L71
So, no matter what, the fact is they are doing the same thing.

### About passing down
Previously, `toIdSchema` didn't pass down to array I guess it's because it relied on array data, which was not given to `toIdSchema` in the very beginning. But now, since the `toIdSchema` has the form data, it should be able to generate a well formed idSchema at the root instead of being called again in `ObjectField` and `ArrayField`.

### About keeping idPrefix
I didn't remove `idPrefix` for compatibility reason because it's documented and someone might have been using it. If we decide to clean it up, it's fine to delete the whole thing.

src/utils.js
knilink: This happen when the schema is a fix length array but the form data is longer then it. If those exceed are not go be displayed on the form then it should be fine not assigning an id.

knilink:
Ideally passing a `pathJoin` function will make it simpler.
`idSchema[fieldName].$id` is equal to `pathJoin(parentId, field)`.

test/utils_test.js
glasserc: Wait, now I'm even more confused. I definitely misread the code at first, and you're right, it seems like `ui:rootFieldId` and `idPrefix` are doing the same thing, but we're not maintaining compatibility with `idPrefix` -- any `idPrefix` is going to be ignored completely.

I'm the one who merged the PR to add `idPrefix`. I should have paid more attention! But it's definitely documented and I bet @edi9999 uses it somewhere. We should either make an effort to maintain backwards compatibility or admit defeat and add it to #805. What do you think?

edi9999: Yes, I am indeed using idPrefix to uniquely identify form elements from our tests. It could be good if you keep backwards compatibility, but I would understand if you don't.

glasserc: @edi9999 Is there a reason you need `idPrefix` instead of putting `rootFieldId` in the `uiSchema`?

knilink:
closing this as it too old and no longer the best solution in my opinion.
my original thought was to use a different custom separator to avoid element id collision but this doesn't solve the issue entirely 
- the ideal solution IMO is to use [json pointer ](https://tools.ietf.org/html/rfc6901)
- or secondly, escape the separator character (.e.g `id = parentId + '_' + escape(name, '_'`)

PR #952, opened 2018-06-09
Export default template

### LINKED ISSUE #951: export DefaultTemplate or add to the registry object
### Prerequisites

### Description
It would be really helpful if `DefaultTemplate` was exported or added to the [registry object](https://github.com/mozilla-services/react-jsonschema-form#the-registry-object). As far as I can tell, it's impossible to use the `DefaultTemplate` in a custom field.

My preference would be to ensure that `registry.FieldTemplate` has a default value (which would be `DefaultTemplate`.

### Steps to Reproduce
1. Create a custom field.
2. Try to wrap it in the `DefaultTemplate`.

#### Expected behavior
Be able to wrap custom fields in the `DefaultTemplate` or the defined custom field template.

#### Actual behavior
Cannot access the `DefaultTemplate`.

### Version
`1.0.2`

davidbarratt:
Fixes #951

### Checklist

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -12,9 +12,9 @@ import {
   deepEquals,
   getSchemaType,
 } from "../../utils";
+import DefaultTemplate from "../templates/DefaultTemplate";
 import UnsupportedField from "./UnsupportedField";
 
-const REQUIRED_FIELD_SYMBOL = "*";
 const COMPONENT_TYPES = {
   array: "ArrayField",
   boolean: "BooleanField",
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -57,6 +57,7 @@ export function getDefaultRegistry() {
   return {
     fields: require("./components/fields").default,
     widgets: require("./components/widgets").default,
+    FieldTemplate: require("./components/templates/DefaultTemplate").default,
```
knilink: https://github.com/mozilla-services/react-jsonschema-form/blob/e0192b75d24024370d52a43293931c8ddd769752/src/components/Form.js#L162

I think the form doesn't actually receive the template.

src/components/fields/SchemaField.js
```diff
@@ -12,9 +12,9 @@ import {
   deepEquals,
   getSchemaType,
 } from "../../utils";
+import DefaultTemplate from "../templates/DefaultTemplate";
```
knilink: I think it's not necessary to import it here if we decided to pass it from the registry

src/utils.js
epicfaace: You're right -- @davidbarratt can you fix this?

src/components/fields/SchemaField.js
epicfaace: I agree

epicfaace:
@davidbarratt just bumping this, can you please make the changes requested so I can merge this?

epicfaace:
@davidbarratt bump?

davidbarratt:
I don't use this library anymore, so feel free to take over the PR or close it. :)

epicfaace:
Okay, I'll just close it, then.

PR #954, opened 2018-06-14
Allow schema $refs to not have to start with `#/definitions`

domharrington:

We're using this module to render forms for Swagger/OAS files. $refs
in swagger are stored in `components/schemas` at the top level of the object.

This change allows $ref strings to not always start with `#/definitions/`
which is a much easier solution than us having to recursively modify $refs
to prefix with this value.

All existing tests are passing, with a new test for this functionality.

The change was to make the definitions part of the regular expression an
optional non-capturing group.

Edit: any reason why you dont commit the package-lock.json file? 

Edit 2: also, is there any main reason (apart from keeping the repo clean) why you don't commit the built source files (in lib/ and dist/?) This makes it tricky to `npm install` from a fork, as `lib/index.js` doesn't exist in git.

### Checklist


domharrington:
Tests are passing locally on my machine. Anyone know why it would be failing the prettier checks for a file i have not modified?

domharrington:
Yes! We're doing it quite successfully over here: https://github.com/readmeio/api-explorer

Demo site: https://preview.readme.io/

domharrington:
If you're interested in a hosted solution for your docs, you should check out ReadMe: https://readme.io/. The API explorer that I linked to above is our open source project, but we have a whole host of other features built into the product including suggested edits for your docs, an API, user login with variables. 

I'm an engineer at ReadMe, please feel free to reach out to me dom[at]readme.io if you have any more questions.

glasserc:
Hey, sorry for not responding to this at the time.

You only change the regular expression not to start with `definitions`, but all definitions are still looked up starting from the schema `definitions`, so e.g. `#/components/schema/address` pulls stuff out of `#/definitions/components/schema/address`.

To be honest, we handle `definitions` specially but I'm not sure if we need to. If they're true JSON pointers, they should be resolvable against the schema root. Maybe we should be passing that around instead.

> Edit: any reason why you dont commit the package-lock.json file?

I'll admit ignorance here, but the reason is that rjsf is a library rather than an application, so constraining packages should happen in the calling code.

> Edit 2: also, is there any main reason (apart from keeping the repo clean) why you don't commit the built source files (in lib/ and dist/?) This makes it tricky to npm install from a fork, as lib/index.js doesn't exist in git.

As you say, it's just to keep the repo clean.

domharrington:
Hey! No worries. I made the `definitions` part of the regex optional because the refactor to change everywhere else in the code and examples to *not* start with definitions, plus it probably would constitute a major version bump to do so.

This change just allowed the possibility for it to not have to start with `definitions`. We're rendering OAS files, and it's much easier to just patch on the components object directly onto each schema than it is to recursively crawl the object for all $refs and modify them all to start with `definitions`.

Hope that makes it a little clearer for the reasoning for this change. I'm using a fork of this library currently but it would be great to get back onto the main repo. I've got one more PR too: #1001.

Thanks in advance.

glasserc:
I think no longer treating `definitions` as special but just resolving JSON pointers against the root of the schema is largely backwards-compatible. Schemas that used to work should continue to work, for instance. I think the only user-facing place where we use `definitions` is in `registry`, and I think we could produce a backwards-compatible way to manage that (combine the user's `schema` and `definitions` using `realSchema = {...schema, definitions: registry.definitions}`). I know that would be more work, but I think it would be much clearer than having a regexp with an optional string buried deep in the code, plus I think it would be cleaner for your use case. What do you think?

domharrington:
Hey @epicfaace yup i agree with everything that was said before. It's on my todo list to update this PR with the more generic solution, i've just not had the capacity to get round to it yet.

epicfaace:
This PR fixes #988 .

domharrington:
I took a stab at finishing this off today but kept hitting upon problems around this block here: https://github.com/mozilla-services/react-jsonschema-form/blob/bece2a5a847e216c2386d4f7f0ce9cf25b549562/src/utils.js#L402-L404

Since the $ref and the definitions are coming from the same object now, it appears to get stuck in an infinite loop during that while block. E.g.

```js
const schema = {
  definitions: {
    testdef: { type: "string" },
  },
  $ref: "#/definitions/testdef",
};
```

The code attempts to resolve the $ref before it gets to the point where it looks in the current object: https://github.com/mozilla-services/react-jsonschema-form/blob/bece2a5a847e216c2386d4f7f0ce9cf25b549562/src/utils.js#L405

Can anyone with more knowledge of this code point me in the right direction or suggest a way forward?

domharrington:
Hey @epicfaace, any suggestions on a way forward for this? This PR as it is is backwards compatible and though it's not completely ideal, when I tried to refactor completely I came up against some issues.

minhnq013:
Anyone know what's the status for this one?
I have several schemas that are un-usable due to this.

Thanks

domharrington:
I'm using a fork of this module, which only has this change. You can use it by running this command:

```
npm i "github:domharrington/react-jsonschema-form#dist-committed"
```

epicfaace:
@domharrington Can you share the code for the approach you tried? I'm trying to understand what was your approach there. I'm thinking that a successful approach might need to make `findSchemaDefinition` accept the entire schema object rather than the `definitions` object.

I'm not comfortable with the way it's implemented now, because having `#/components/schemas/address` actually point to `#/definitions/components/schemas/address` seems to be unintuitive.

domharrington:
It's been over 2 months since I attempted it, so the code I put together at the time is probably lost to the ether or in a git stash somewhere, so I cannot remember what I tried.

The fix I implemented makes having `definitions` completely optional in the $ref path, it doesn't have it present in one place whilst actually pointing somewhere else. It's a backwards compatible fix

domharrington:
Any insight you can give into getting this merged would be amazing! I couldn't figure it out at the time.

epicfaace:
Hmm, not sure, will look into it and let you know

erunion:
@epicfaace @glasserc Is there anything else that needs to be done to this PR before it gets merged in?

epicfaace:
Closing this in favor of #1506 .

PR #967, opened 2018-06-30
Add encodeUriComponent to file name

jorgerobles:

Fix for non standard file names (with spaces, apostrophes, etc) that prevent to be used.
For example uploading a whitespace named file prevents to be used as background url.

### Checklist

### DIFF
--- a/test/StringField_test.js
+++ b/test/StringField_test.js
@@ -1501,9 +1501,6 @@ describe("StringField", () => {
   describe("FileWidget", () => {
     const initialValue = "data:text/plain;name=file1.txt;base64,dGVzdDE=";
 
-    const nonUriEncodedValue = "fileáéí óú1.txt";
-    const uriEncodedValue = "file%C3%A1%C3%A9%C3%AD%20%C3%B3%C3%BA1.txt";
-
     it("should render a color field", () => {
       const { node } = createFormComponent({
         schema: {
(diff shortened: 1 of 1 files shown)


src/components/widgets/FileWidget.js
```diff
@@ -4,6 +4,7 @@ import PropTypes from "prop-types";
 import { dataURItoBlob, shouldRender, setState } from "../../utils";
 
 function addNameToDataURL(dataURL, name) {
+  name=encodeURIComponent(name)
```
graingert: you've got a lint error here.

Do this instead:
```
  function addNameToDataURL(dataURL, name) {
    return dataURL.replace(";base64", `;name=${encodeURIComponent(name)};base64`); 
}
```

jorgerobles: Sorry @graingert, did not see your comment until now.

epicfaace:
Great, thanks @jorgerobles ! Can you also add a test? It would be similar to the tests already existing: https://github.com/mozilla-services/react-jsonschema-form/blob/1941e760d8af93b118bbc193dd98b88f64b32645/test/StringField_test.js#L1503

jorgerobles:
@epicfaace I've added the test. I'm not very familiarised with mocha, hope is well.

epicfaace:
@jorgerobles Thanks, almost there -- instead of modifying the existing test, can you leave the existing test case "should reflect the change into the dom" in place and add a new test case right next to it? (you can copy and paste the existing test case and make the changes that you did)

test/StringField_test.js
```diff
@@ -1501,6 +1501,9 @@ describe("StringField", () => {
   describe("FileWidget", () => {
     const initialValue = "data:text/plain;name=file1.txt;base64,dGVzdDE=";
 
+    const nonUriEncodedValue = "fileáéí óú1.txt";
```
epicfaace: Please move these two lines inside the "should encode file name with encodeURIComponent" test case.

jorgerobles: 👍

epicfaace:
@jorgerobles can you help with #824? It's pretty similar to this one.

Issue #674, opened 2017-08-21
Decimal numbers ending in zero


spacebaboon:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description
A field with type `number` does not validate when the entered value is a decimal ending in a 0, as it is stored in `formData` as a string.
This means that, e.g. `2.1` validates correctly as a number but `2.0` fails with an error '`instance.number is not of a type(s) number'`.

### Steps to Reproduce

1. In the Playground, Numbers tab, enter `2.1`. No error.
2. Change the number to `2.0`. Error is shown.

#### Expected behavior
`2.0` (and any decimal numbers ending in a `0`) validate correctly as numbers.

This is more complex than I initially thought, and I'm not sure what to suggest as a fix. 
As soon as a number ends in a . or a `0`, it is returned as a string. A number will be represented as simply as possible, e.g. `2.0` as a number is just `2`. But in order to input numbers like `2.01` with live validation on, you need to stop this behaviour, otherwise `2.0` would immediately be rewritten as `2` as soon as the user hits the `0`.
Json Schema only has `integer` and `number` data types, `decimal` is not (yet) supported.
I wonder if the `uiSchema` could include something like a format string, a number of decimal digits, or a decimal flag. That way, the field input could still be displayed in a certain way (maybe using `toFixed(n)`) but upon validation, the form could send the value as a number. 

### Version
0.49.0

spacebaboon:
angular-schema-form handles this nicely. it looks like formdata is the number, but the input field is just a text input. change the Name field's type to number:
http://schemaform.io/examples/bootstrap-example.html

glasserc:
We explicitly treat strings with a decimal point and a 0 at the end as strings -- see src/utils.js. This behavior is because the JS Number type does not support distinguishing between `2` and `2.0`, and we need to preserve exactly what the user entered in the input field. I'd be happy to review solutions to this problem, but I don't have any myself.

PR #971, opened 2018-07-05
Add new number input for type number

### LINKED ISSUE #966: Default widget for number fields
### Prerequisites

### Description

Default widget for numbers seems to be text fields.

### Steps to Reproduce
```js
const Form = JSONSchemaForm.default;
const schema = {
  title: "react-jsonschema-form demo",
  type: "object",
  properties: {
  	rate: {type: "number"},
    percent: {type: "number"}
  }
};
const uiSchema = {
  percent: {
    "ui:widget": "range"
  }
};

ReactDOM.render(<Form schema={schema} uiSchema={uiSchema}/>, 
             document.getElementById("main"));
```

#### Expected behavior

I would expect the serialized value to be an integer or float

#### Actual behavior

Currently the value is quoted as string

leplatrem:

When schema has `type: number`, we should use a `type=number` otherwise with a text widget the serialized value is a string.

Fixes #966 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ render((
 
 ### Alternative widgets
 
-The uiSchema `ui:widget` property tells the form which UI widget should be used to render a field. 
+The uiSchema `ui:widget` property tells the form which UI widget should be used to render a field.
 
 Example:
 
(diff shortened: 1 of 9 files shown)


glasserc:
Refs #958, #674. When you write "the serialized value is a string", what does "serialized" mean? Is the value in `formData` a string, even when the entered value is a number?

leplatrem:
> When you write "the serialized value is a string", what does "serialized" mean? Is the value in formData a string, even when the entered value is a number?

Yes, the field value in the data passed to `onSubmit` is a string even if the entered value is a number. 

The original issue is this https://bugzilla.mozilla.org/show_bug.cgi?id=1468878

glasserc:
This is a duplicate of #674. Specifically, the number being entered is `1.0`, which cannot be represented as a JavaScript number. I don't see how changing the type of the input field will fix the issue.

leplatrem:
> This is a duplicate of #674

Ok closing then.

Issue #798, opened 2018-01-05
Symbol is undefined in IE11


rahul1059:
IE11 doesn't support Symbol and error is displayed in console whenever form has any validation error.
I don't want to add huge babel-polyfill for this. Is there any way to fix it quickly.
![image](https://user-images.githubusercontent.com/5712189/34603342-98e6583c-f229-11e7-98f1-fd5a81a562fd.png)

Natim:
I would be glad if you could investigate this error and start suggesting a fix for it.

edi9999:
It seems to be in the bundle.js file without a test for typeof Symbol === "function" beforehand : 

```
...
for(var l,u=n[Symbol.iterator]();!(i=(l=u.next()).done);i=!0){var c=l.value;if(c=c.replace(/~1/g,"/").replace(/~0/g,"~"),!o.hasOwnProperty(c))throw new Error("Could not find a definition for "+e+".");o=o[c]}
...
```

Line 29, Character 29661

KiranChaudhariJM:
This solution will work for sure, it worked for me when I encountered the error: 'Symbol' is undefined in IE . It worked earlier in Chrome and Firefox but IE was throwing this error.It took me few hours to find this solution.
I am using the latest React at this time react "react": "^16.5.0" on windows machine.

    1. Install babel-polyfill
    npm install --save-dev babel-polyfill
    In package.json, it should have the following entries

    "devDependencies": {
        "babel-core": "^6.26.3",
        "babel-loader": "^8.0.2",
        "babel-polyfill": "^6.26.0",
        "babel-preset-react": "^6.24.1"
      }

    2. In index.js, add
    
        import babelPolyfill from 'babel-polyfill';

Problem should get solved

PR #982, opened 2018-07-19
support old browsers without polyfills

graingert:

Currently react-jsonschema-form forces downstream projects to use babel-polyfill, and includes duplicated babel helpers.

This removes those duplicate helpers and adds ponyfills for modern globals like Set, Promise and Objects.assign.

This PR also replaces methods like `Array.prototype.includes` and `Array.prototype.fill` with ponyfills from `core-js`

If this is related to existing tickets, include links to them as well:

#798

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
   "dependencies": {
     "ajv": "^5.2.3",
     "babel-runtime": "^6.26.0",
+    "core-js": "^2.5.7",
     "lodash.topath": "^4.5.2",
     "prop-types": "^15.5.8",
     "setimmediate": "^1.0.5"
--- a/package.json
+++ b/package.json
@@ -45,8 +45,7 @@
     "babel-runtime": "^6.26.0",
     "core-js": "^2.5.7",
     "lodash.topath": "^4.5.2",
-    "prop-types": "^15.5.8",
-    "setimmediate": "^1.0.5"
+    "prop-types": "^15.5.8"
   },
   "devDependencies": {
     "atob": "^2.0.3",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "ajv": "^5.2.3",
-    "babel-preset-env": "^1.7.0",
+    "babel-runtime": "^6.26.0",
     "lodash.topath": "^4.5.2",
     "prop-types": "^15.5.8",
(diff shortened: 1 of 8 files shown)


src/components/widgets/SelectWidget.js
```diff
@@ -3,18 +3,16 @@ import PropTypes from "prop-types";
 
 import { asNumber } from "../../utils";
 
+const nums = new Set(["number", "integer"]);
```
magopian: Shouldn't this `Set` be taken from `core-js`? Or maybe the same change as above, using the polyfilled `includes` from `core-js`.

graingert:
transform-runtime automatically ponyfills globals like Set Promise setImmediate and Object.assign

magopian:
oh, wait, no, I borked it. Let me try again...

graingert:
@magopian let me take over

package.json
```diff
@@ -42,10 +42,10 @@
   },
   "dependencies": {
     "ajv": "^5.2.3",
-    "babel-preset-env": "^1.7.0",
```
graingert: woops, I put this in the wrong place in my last pr, should have been a devDep

graingert:
@magopian done

glasserc:
Semver says it's a minor bump only if new functionality is added *to the API*. The API hasn't changed, so it's a patch version.

glasserc:
@graingert , could you please look at #991 and #1082 and tell me what you think? I am worried that we may have broken the way some of our users are including rjsf, but I'm not an expert in JS packaging so I'm hoping you can help.

graingert:
Ok I'll have a look tomorrow

graingert:
@glasserc I took a look - looks like it's a simple fix to the webpack babel-loader config to exclude babel loader on babel and core-js in webpack.

This is why it works when people use the package from npm directly

glasserc:
To elaborate on the above comment, I spoke with @graingert via IRC and he explained that he thought the problem was that we're running babel on all our dependencies, including the ones that provide ponyfills such as corejs and babel-runtime, and that this is due to a webpack misconfiguration. I'll try to fix it and open a PR today.

graingert:
@glasserc currently I think you only *need* to exclude `node_modules\/core-js\/.*` from `babel-loader`

It may be prudent to switch to a whitelist, however

PR #984, opened 2018-07-19
use babel-preset-env over deprecated es2015

graingert:


magopian:
I don't understand why [some packages are downgraded](https://registry.npmjs.org/core-js)?
Also, just reading the first paragraph of [the docs](https://babeljs.io/docs/en/next/babel-preset-env.html), it says that

> We don't recommend using preset-env this way because it doesn't take advantage of its ability to target specific browsers.

I'm not sure if we should target specific browsers, maybe not, so that it's usable on most (all ?) of them?

graingert:
There's not much point in targeting ie11 because that just enables all features

graingert:
@magopian any chance of a release and an update to the playground?

glasserc:
Hi, I just cut a release and deployed it to the playground. Please let me know if you notice anything wrong with it.

graingert:
@glasserc tried out the playground in IE11 and it works!

graingert:
Awesome thanks!

PR #986, opened 2018-07-23
listen to FileReader errors

graingert:


src/components/widgets/FileWidget.js
```diff
@@ -11,6 +11,7 @@ function processFile(file) {
   const { name, size, type } = file;
   return new Promise((resolve, reject) => {
     const reader = new window.FileReader();
+    reader.onerror = reject;
```
graingert: sadly I couldn't find any tests for this

PR #1000, opened 2018-08-06
Bug/multiple of with a float

### LINKED ISSUE #997: Regression for multipleOf precision issue after change to ajv
### Prerequisites

### Description

I believe that after the change to use ajv the issue where "multipleOf" is set to a decimal and fails validation due to floating point precision has appeared again. The original ticket to fix it was #671 

I beleive this can be fixed by setting the option `multipleOfPrecision` in ajv : https://github.com/epoberezkin/ajv#options 

### Steps to Reproduce

In the playground set the schema to:
```
{
  "type": "object",
  "properties": {
    "price": {
      "title": "Price per task ($)",
      "type": "number",
      "multipleOf": 0.01,
      "minimum": 0
    }
  }
}
```

set the value in the form to 0.14 and the field will display an error "should be multiple of 0.01".

### Version
1.x.x

llamamoray:

Fixes #997 

### Checklist

### DIFF
--- a/src/components/fields/UnsupportedField.js
+++ b/src/components/fields/UnsupportedField.js
@@ -5,7 +5,8 @@ function UnsupportedField({ schema, idSchema, reason }) {
   return (
     <div className="unsupported-field">
       <p>
-        Unsupported field schema{idSchema &&
+        Unsupported field schema
+        {idSchema &&
           idSchema.$id && (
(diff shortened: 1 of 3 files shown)


llamamoray:
@n1k0 @glasserc would you be able to take a look at this?

PR #1001, opened 2018-08-07
Add wrapper element around required asterisk in TitleField

domharrington:
This is so that it can be separately styled easily via CSS.
It is already done this way in https://github.com/mozilla-services/react-jsonschema-form/blob/aadbbe9701c4db9df761e86f0280e1ecafc509f8/src/components/fields/SchemaField.js#L59

### Checklist


glasserc:
Thanks!

Issue #992, opened 2018-07-30
Remove value from formData when property is removed (not required)


zrzka:
### Description

I'm trying to achieve something like this:

* if `enable` property is `true`, it MUST contain `ssid` property
* if `enable` property is `false`, it MUST NOT contain `ssid` property

### Steps to Reproduce

1. Open this [example](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0aXRsZSI6IkNvbmZpZ3VyYXRpb24iLCJkZXNjcmlwdGlvbnMiOiJEZXBlbmRlbmNpZXMgZXhhbXBsZXMiLCJ0eXBlIjoib2JqZWN0IiwicmVxdWlyZWQiOlsiZW5hYmxlIl0sInByb3BlcnRpZXMiOnsiZW5hYmxlIjp7InRpdGxlIjoiRW5hYmxlIFdpRmk/IiwidHlwZSI6ImJvb2xlYW4iLCJkZWZhdWx0IjpmYWxzZX19LCJkZXBlbmRlbmNpZXMiOnsiZW5hYmxlIjp7Im9uZU9mIjpbeyJhZGRpdGlvbmFsUHJvcGVydGllcyI6ZmFsc2UsInByb3BlcnRpZXMiOnsiZW5hYmxlIjp7ImVudW0iOlt0cnVlXX0sInNzaWQiOnsidGl0bGUiOiJOZXR3b3JrIFNTSUQiLCJ0eXBlIjoic3RyaW5nIn19LCJyZXF1aXJlZCI6WyJzc2lkIl19LHsiYWRkaXRpb25hbFByb3BlcnRpZXMiOmZhbHNlLCJwcm9wZXJ0aWVzIjp7ImVuYWJsZSI6eyJlbnVtIjpbZmFsc2VdfX19XX19fSwidWlTY2hlbWEiOnt9fQ==)
2. Enable WiFi
3. Fill in SSID
4. Disable WiFi
5. Error `['ssid'] should NOT have additional properties` appears

If the link in 1. wont work, here's the [schema](https://gist.github.com/zrzka/7dd4028e312dd0b1afe0e93e2d81e050).

#### Expected behavior

No `['ssid'] should NOT have additional properties` error. I expect that there will be no `ssid` value in `formData` when I disable WiFi.

#### Actual behavior

[...]

Issue #902, opened 2018-04-18
Nested schema objects with required properties in dependencies validate incorrectly


Steffan-Ennis:
### Prerequisites

- [x ] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x ] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Nested schema objects with required properties in dependencies validate incorrectly

### Steps to Reproduce
I'm providing a [sample JSFiddle](https://jsfiddle.net/Ldqc584z/6/) .

Select Address in the sections drop down
Click Submit

#### Expected behavior

The Form validates correctly and renders errors on Street and Street Number

#### Actual behavior

The Form validates with no errors and invokes the onSubmit prop

#### Debug info
The error seems to be with `ajv.js` `Ajv` will not validate against nested schemas in dependencies unless the formData for the schema has been touched. 

Ie setting `formData.address = {}` will validate correctly

![screen shot 2018-04-19 at 10 41 01 am](https://user-images.githubusercontent.com/16796843/38965953-d1745b5e-43be-11e8-9491-0bd3b078393c.png)

It may be worth opening an issue in AJV

claybridges:
I can repro this with a much simpler schema. See [this JSFiddle](https://jsfiddle.net/tp8y330f/1/).

phatNfqAsia:
As my last check, the form behaves correctly. So you might want to close this issue though @Steffan-Ennis 
![image](https://user-images.githubusercontent.com/39822773/120170549-46138a00-c22b-11eb-85ee-2d99936c3bd3.png)

SleeplessByte:
It doesn't behave correctly when it's defined as:

- root schema has required: ["a"]
- property "b" has required: ["foo"]

When "b" is not present, the error will be "b.foo" is required.

jeudryWepsys:
THis is not finished, i have the same problem, the problem is when you have a dependency inside a nested object and then a required property inside that dependency.

Issue #848, opened 2018-02-19
Dependencies yields spurious validation error: "... should NOT have additional properties"


krallin:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

When using a [dynamic](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#dynamic) that includes a "controlled" property (i.e. one that shows up only if the "controlling" property is set to a specific value), then if the controlled property is an object, it fails validation. 

The form reports the controlled property is invalid because it has extra fields, but those fields are actually defined in the schema (presumably, if they weren't, then the form wouldn't even draw that field).

### Steps to Reproduce

[...]

PR #1018, opened 2018-08-28
Document a gotcha with `additionalProperties`

loganvolkers:
Relates to #848 #902 #992 (and maybe others).

It seems like a lot of people are running into the same problem, this will help them avoid it earlier and be more clear about the current limitations of the library.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1744,6 +1744,7 @@ This component follows [JSON Schema](http://json-schema.org/documentation.html) 
 * `anyOf`, `allOf`, and `oneOf`, or multiple `types` (i.e. `"type": ["string", "array"]`
     Nobody yet has come up with a PR that adds this feature with a simple and easy-to-understand UX.
     You can use `oneOf` with [schema dependencies](#schema-dependencies) to dynamically add schema properties based on input data but this feature does not bring general support for `oneOf` elsewhere in a schema.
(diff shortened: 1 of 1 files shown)


glasserc:
Thanks for submitting this, and sorry for taking so long to respond to it. This reads like a pretty mild warning but it almost sounds like it's describing a bug. Is it actually a bug? Is using `additionalProperties: false` incorrect somehow? If not, is there a workaround -- something else that should be used?

loganvolkers:
#848 has a pretty good example of the behaviour. Using `additionalProperties: false` will produce incorrect results and break things. There may be a fix for this that would delete additional properties from the output, but would be hard to work.

Either way, it is a limitation that seems to be commonly encountered, so it should be more clearly documented.

README.md
```diff
@@ -1744,6 +1744,7 @@ This component follows [JSON Schema](http://json-schema.org/documentation.html)
 * `anyOf`, `allOf`, and `oneOf`, or multiple `types` (i.e. `"type": ["string", "array"]`
     Nobody yet has come up with a PR that adds this feature with a simple and easy-to-understand UX.
     You can use `oneOf` with [schema dependencies](#schema-dependencies) to dynamically add schema properties based on input data but this feature does not bring general support for `oneOf` elsewhere in a schema.
+* `"additionalProperties":false` when used with [schema dependencies](#schema-dependencies) will not remove the extra properties and produce invalid schemas. It is recommended to avoid setting `"additionalProperties":false`. See [#848](https://github.com/mozilla-services/react-jsonschema-form/issues/848) [#902](https://github.com/mozilla-services/react-jsonschema-form/issues/902) [#992](https://github.com/mozilla-services/react-jsonschema-form/issues/992)
```
epicfaace: Can you change the wording here? I can't tell if this means "will not remove the extra properties and will produce invalid schemas" or "will not remove the extra properties and will not produce invalid schemas"

epicfaace: Also, can you make your change in the `docs` directory instead of the README? Thanks

epicfaace: Ok, that's great! Let me know what kind of features / bugs you come across; I'll try to be as responsive as possible to new ideas / PRs.

PR #1021, opened 2018-09-03
Add preliminary support for object "additionalProperties"

christianclent:

This will add the ability for the form to understand "additionalProperties" definitions in the schema (see [json-schema properties documentation](https://json-schema.org/understanding-json-schema/reference/object.html#properties)).

Related issue: https://github.com/mozilla-services/react-jsonschema-form/issues/228

### Checklist

### Details

Specifically, this will enable the following behaviors when additionalProperties is enabled:
* If the form data contains properties that are not explicitly defined, those fields will be automatically added to the form
* An "expand" button will be [optionally] added that allows the form user to manually add new properties.
* In any case in which an additional property has been added, a new text field will be rendered to allow the user to rename the key

Opportunities for improvement:
* Ability to remove additional properties
* Somewhat better test coverage (particularly around type detection)

### DIFF
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -1,3 +1,4 @@
+import AddButton from "../AddButton";
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 
@@ -5,9 +6,28 @@ import {
   orderProperties,
   retrieveSchema,
   getDefaultRegistry,
+  getUiOptions,
 } from "../../utils";
 
 function DefaultObjectFieldTemplate(props) {
+  const canExpand = function canExpand() {
+    const { formData, schema, uiSchema } = props;
+    if (!schema.additionalProperties) {
+      return false;
+    }
+    let { expandable } = getUiOptions(uiSchema);
+    if (expandable !== false) {
+      // if ui:options.expandable was not explicitly set to false, we can add
+      // another property if we have not exceeded maxProperties yet
+      if (schema.maxProperties !== undefined) {
+        expandable = Object.keys(formData).length < schema.maxProperties;
+      } else {
+        expandable = true;
+      }
+    }
+    return expandable;
+  };
+
   const { TitleField, DescriptionField } = props;
   return (
     <fieldset>
@@ -27,6 +47,13 @@ function DefaultObjectFieldTemplate(props) {
         />
       )}
       {props.properties.map(prop => prop.content)}
+      {canExpand() && (
+        <AddButton
+          className="object-property-expand"
+          onClick={props.onAddClick(props.schema)}
+          disabled={props.disabled || props.readonly}
+        />
+      )}
     </fieldset>
   );
 }
@@ -41,6 +68,13 @@ class ObjectField extends Component {
     disabled: false,
     readonly: false,
   };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      additionalProperties: {},
+    };
+  }
 
   isRequired(name) {
     const schema = this.props.schema;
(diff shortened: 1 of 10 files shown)


src/components/fields/ObjectField.js
```diff
@@ -1,13 +1,33 @@
...
+  getUiOptions,
 } from "../../utils";
 
 function DefaultObjectFieldTemplate(props) {
+  const canExpand = function canExpand() {
+    const { formData, schema, uiSchema } = props;
+    if (!schema.additionalProperties) {
+      return false;
+    }
+    let { expandable } = getUiOptions(uiSchema);
```
graingert: No need to make expandable a let: 

```
const { expandable } = getUiOptions(uiSchema);
if (expandable !== false) {
      // if ui:options.expandable was not explicitly set to false, we can add
      // another property if we have not exceeded maxProperties yet
      if (schema.maxProperties !== undefined) {
        return Object.keys(formData).length < schema.maxProperties;
      } else {
        return true;
      }
    }
return expandable;
```

src/components/fields/ObjectField.js
```diff
@@ -42,6 +69,13 @@ class ObjectField extends Component {
     readonly: false,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
```
graingert: it looks like we've got the class fields proposal enabled, so you can do this with:
```
state = {
   additionalProperties: {}, 
}
```

src/components/fields/ObjectField.js
```diff
@@ -63,6 +97,64 @@ class ObjectField extends Component {
...
+        return 0;
+      case "object":
+        return {};
+      default:
+        // We don't have a datatype for some reason (perhaps additionalProperties was true)
+        return "New Value";
+    }
+  }
+
+  handleAddClick = schema => {
```
graingert: `handleAddClick = schema => () = >`

src/components/fields/ObjectField.js
```diff
@@ -141,7 +234,12 @@ class ObjectField extends Component {
       formData,
       formContext,
     };
-    return <Template {...templateProps} />;
+    return (
+      <Template
+        {...templateProps}
+        onAddClick={this.handleAddClick.bind(this)}
```
graingert: handleAddClick is an arrow function, this bind is a noop

src/utils.js
```diff
@@ -402,25 +404,75 @@ function findSchemaDefinition($ref, definitions = {}) {
...
-    return retrieveSchema(resolvedSchema, definitions, formData);
-  } else {
-    // No $ref or dependencies attribute found, returning the original schema.
-    return schema;
+    schema = retrieveSchema(resolvedSchema, definitions, formData);
+  }
+  if (schema.hasOwnProperty("additionalProperties")) {
+    schema = stubExistingAdditionalProperties(schema, definitions, formData);
   }
+  return schema;
```
graingert: avoid reassigning params

christianclent:
Thanks for the great feedback @graingert!

mjasinski5:
🙏

mjasinski5:
@christianclent CI failed because of the timeout.. I think all you have to do is to rerun it ;) It would be a big loss if such a remarkable work couldnt be merged 👍

derberg:
@christianclent maybe close this and open new PR as this will trigger travis again https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit
otherwise I'm not sure if anyone will treat this PR seriously

christianclent:
Thank you for the hot tip @derberg!

I've been using this branch in one of my personal projects for a few weeks now, and there are definitely areas I would like to improve.

- The meat of the change in is SchemaField (with some ancillary code in ObjectField), but I wonder if I should have contained the change to ObjectField only
- Because the new keys are not known ahead of time, it typically isn't feasible to apply uiSchemas to the additional properties.
- It doesn't handle ordering very well
- Because ids of underlying DOM elements incorporate object keys, it will annoyingly bork your tab order when changing an additional property's key.
- Maybe more?

Some of those may be feasible to fix in the next few weekends if they are deal-breakers for anybody.

pkosiec:
Hi @christianclent,
Thank you very much for your changes! I think the improvements you've mentioned could be made in another pull request. We really need object support in `additionalProperties`!

@graingert Could you please take another look at this PR? Thanks!

src/components/fields/ObjectField.js
```diff
@@ -1,13 +1,32 @@
...
+    const { expandable } = getUiOptions(uiSchema);
+    if (expandable !== false) {
+      // if ui:options.expandable was not explicitly set to false, we can add
+      // another property if we have not exceeded maxProperties yet
+      if (schema.maxProperties !== undefined) {
+        return Object.keys(formData).length < schema.maxProperties;
+      }
+      return true;
+    }
+    return expandable;
```
glasserc: You could refactor this as 

```
if (expandable === false) {
  return false;
}
if (schema.maxProperties !== undefined) {
  return Object.keys(formData).length < schema.maxProperties;
}
return true;
```

This would save a level of indentation. But that's just a personal preference.

src/components/fields/ObjectField.js
```diff
@@ -63,6 +93,62 @@ class ObjectField extends Component {
...
+      const newFormData = { ...this.props.formData };
+      const property = newFormData[oldValue];
+      delete newFormData[oldValue];
+      newFormData[value] = property;
+      this.props.onChange(
+        newFormData,
+        errorSchema &&
+          this.props.errorSchema && {
+            ...this.props.errorSchema,
+            [name]: errorSchema,
```
glasserc: Should `name` here be `value`? Or more generally, where does `name` come from in this scope?

src/utils.js
```diff
@@ -423,6 +474,21 @@ export function retrieveSchema(schema, definitions = {}, formData = {}) {
   }
 }
 
+export function retrieveSchema(schema, definitions = {}, formData = {}) {
+  const resolvedSchema = resolveSchema(schema, definitions, formData);
+  const hasAdditionalProperties =
+    resolvedSchema.hasOwnProperty("additionalProperties") &&
+    resolvedSchema.hasAdditionalProperties !== false;
```
glasserc: What is the `hasAdditionalProperties` property? Is this maybe supposed to be `additionalProperties`?

pkosiec:
@glasserc Thank you for merging this! Do you plan a new release soon?

glasserc:
@pkosiec I don't know. It seems like 1.0.5 is broken so I might cut another one once we figure out what to do about #1042.

PR #1028, opened 2018-09-14
Added ui:enumDisabled to RadioWidget.js

stamp:

I noticed that `ui:enumDisabled` didn't work for the `RadioWidget` so I added it the same way it is implemented in the `SelectWidget`

### Checklist


glasserc:
Thanks!

PR #1034, opened 2018-09-18
call props.onChange on Form component mount to set any defaults in sc…

### LINKED ISSUE #1033: Form not setting default schema values when using onChange during initialisation
### Prerequisites

### Description

When there are defaults set in the schema such as 
```
  schema: {
    title: "A registration form",
    type: "object",
    properties: {
      firstName: {
        type: "string",
        title: "First name",
        default: "Chuck",
      }
   }
}
```

the default is correctly displayed in the form (and stored internally), but if you are using the `onChange` prop on the form and using that state for say an external submission button you never receive the defaults.

An example of this is can be seen in the playground by changing `/playground/samples/simple.js` by removing `formData.firstName` and chainging the definition of the schema property to:
```
firstName: {
  type: "string",
  title: "First name",
  default: "Chuck",
},
```

The displayed form data on the page doesn't include a firstName field as the playground is using the `onChange` prop to display that value.

I think on the form component mount it should call `onChange` once mounted to ensure any consumers of the form data have the correct values.

llamamoray:

This fixes #1033 where schema defaults aren't being set on any model that is being updated using the form `onChange` prop.

### Checklist

### DIFF
--- a/playground/samples/simple.js
+++ b/playground/samples/simple.js
@@ -8,6 +8,7 @@ module.exports = {
       firstName: {
         type: "string",
         title: "First name",
+        default: "Chuck",
       },
       lastName: {
         type: "string",
@@ -60,7 +61,6 @@ module.exports = {
     },
   },
   formData: {
-    firstName: "Chuck",
     lastName: "Norris",
     age: 75,
     bio: "Roundhouse kicking asses since 1940",
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -9,6 +9,7 @@ import {
   toIdSchema,
   setState,
   getDefaultRegistry,
+  deepEquals,
 } from "../utils";
 import validateFormData, { toErrorList } from "../validate";
 
@@ -25,10 +26,10 @@ export default class Form extends Component {
   constructor(props) {
     super(props);
     this.state = this.getStateFromProps(props);
-  }
-
-  componentDidMount() {
-    if (this.props.onChange) {
+    if (
+      this.props.onChange &&
+      !deepEquals(this.state.formData, this.props.formData)
+    ) {
       this.props.onChange(this.state);
     }
   }
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -35,7 +35,14 @@ export default class Form extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState(this.getStateFromProps(nextProps));
+    const nextState = this.getStateFromProps(nextProps);
+    this.setState(nextState);
+    if (
+      !deepEquals(nextState.formData, nextProps.formData) &&
+      this.props.onChange
+    ) {
+      this.props.onChange(nextState);
+    }
   }
 
   getStateFromProps(props) {
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -27,6 +27,12 @@ export default class Form extends Component {
     this.state = this.getStateFromProps(props);
   }
 
+  componentDidMount() {
+    if (this.props.onChange) {
+      this.props.onChange(this.state);
+    }
+  }
+
   componentWillReceiveProps(nextProps) {
     this.setState(this.getStateFromProps(nextProps));
   }
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -44,35 +44,60 @@ describe("Form", () => {
     });
   });
 
-  describe("componentDidMount", () => {
-    describe("when props.onchange is set", () => {
-      let comp;
-      let onChangeProp;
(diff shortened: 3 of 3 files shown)


llamamoray:
Hi @glasserc @n1k0 

Would you be able to take a look at this PR?

glasserc:
Thanks for the great write-up! (And sorry for taking so long to get to this.)

My intuition is that if we use a `default` anywhere in the `formData`, then the data has changed, but if the default isn't used (because e.g. there's a value for that property already) then the data has not changed. I'd prefer to see `onChange` only called when the data actually changes, but I'm willing to accept something like this if figuring that out is difficult.

Looking at the code, I see that Form calls `getStateFromProps` in `componentWillReceiveProps`. Does that mean we need to add a call to `onChange` somewhere in that part of the lifecycle too?

llamamoray:
Thanks for taking a look at this. 

I see what you're saying about the onChange callback only being called when the data has actually changed. I can do a deep compare of the the state vs the props formData in the `componentDidMount` and only call onChange in this case:
```js
  componentDidMount() {
    if (this.props.onChange
      && !deepEquals(this.state.formData, this.props.formData)) {
      this.props.onChange(this.state);
    }
  }
```

Would you be happy with that? 

To your second point about the other lifecycle events, I think in theory `onChange` _could_ be called in `componentWillReceiveProps` if and only if the the schema has changed. Something like this:

```js
// Form.js
  componentWillReceiveProps(nextProps) {
    const nextState = this.getStateFromProps(nextProps);
    this.setState(nextState);
    if (!deepEquals(this.props.schema, nextProps.schema)) {
      if (this.props.onChange) {
        this.props.onChange(nextState);
      }
    }
  }
```

The problem with this is that unless the formData prop is cleared on the form before the schema prop is changed, it will have no effect on the internal formData state of the rjsf form. This is because `getStateFromProps` calls `getDefaultFormState` which effectively takes the current formData and merges in the default values (which wouldn't occur as they've already been set). I guess the question is: is it worth doing a deep compare of the schema on every prop change when with the current implementation it doesn't set the default in any case?

glasserc:
Regarding `componentWillReceiveProps`,  I was thinking in terms of the caller of rjsf changing the `formData` or `schema` props out from under us. If the `schema` adds a field with a default, or adds a default to an existing field, wouldn't that involve setting the default? But then I realized I didn't know if `componentWillReceiveProps` is the appropriate lifecycle method or not, and when I went to the React docs, and I see that it's being renamed to [UNSAFE_componentWillReceiveProps](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops). Oh boy. Well, maybe it's better not to rely on such a method. I get the feeling that we need to really rethink all this `deepEquals` stuff that we're doing all over the place, but that's way outside the scope of this PR. So let's drop this point for the time being.

One suggestion in the docs is that we update state in the component constructor. Maybe we should do the calls to `onChange` in the constructor -- what do you think? Otherwise, yes, doing `deepEqual` like you propose is fine.

llamamoray:
Moving the `componentDidMount` logic into the contructor makes sense 👍  - I'll go ahead and do this with the deep equals check.

I think the correct method to use to replace `componentWillReceiveProps` is probably [componentDidUpdate](https://reactjs.org/docs/react-component.html#componentdidupdate). This would involve doing a deep compare of all those props and also moving the library from react 15 to React 16.x.

For now `componentWillReceiveProps` is the best way to hook into the lifecycle event for changes to the schema or form data props, so I propose checking for changes to those inside there and firing onChange if required. 

A bigger piece of work will be to upgrade the library to a newer version of react - a lot of the components may need to be changed. For now they are supporting the legacy lifecycle methods until version 17 albeit with the UNSAFE_ prefix which we could run [this](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) codemod to update the library.

graingert:
I don't think this is the correct solution - I think rjf should expose a function to clean formData and crash if it receives unclean formData

ml242:
Getting an onChange error and endless loop now with my form -- it seems like updating state in the componentWillReceiveProps is the problem. Investigating further now.

llamamoray:
@graingert what do you mean by "clean data"? The problem I was seeing was if you had a form where you were using the onChange prop to store form state outside of rjsf and the schema contained some defaults, then your "external" form state would not reflect the defaults that rjsf sets in it's internal form state on initialisation.

llamamoray:
@ml242 how did your investigations go? Any chance you could create a jsfiddle demonstrating the problem?

graingert:
@llamamoray currently rjsf calls onChange with cleaned data immediately during construction. This shouldn't happen

Issue #1022, opened 2018-09-03
Tips & Tricks don't seem to work with 1.0.4 release


struankl:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Try out any of the tips & tricks links to jsfiddle. None work and error in console: 
Uncaught ReferenceError: JSONSchemaForm is not defined

Downgrading to 1.0.3 the fiddles work as expected.

loganvolkers:
Is this related? https://github.com/mozilla-services/react-jsonschema-form/issues/991

glasserc:
I also think this is a duplicate of #991.

Issue #994, opened 2018-07-31
Getting Uncaught TypeError: a is not a function error on the CDN version of the library.


rohitbishnoi:
![screen shot 2018-07-31 at 5 37 23 pm](https://user-images.githubusercontent.com/6151462/43458639-96eb377a-94e8-11e8-8e0d-640110c302d1.png)

Uncaught TypeError: a is not a function
    at Object.<anonymous> (<anonymous>:1:9802)
    at t (<anonymous>:1:385)
    at Object.<anonymous> (<anonymous>:1:9677)
    at t (<anonymous>:1:385)
    at Object.<anonymous> (<anonymous>:1:9626)
    at t (<anonymous>:1:385)
    at Object.<anonymous> (<anonymous>:1:9332)
    at t (<anonymous>:1:385)
    at Object.<anonymous> (<anonymous>:1:9142)
    at t (<anonymous>:1:385)

Note: This used to work 3 days ago, looks like something is broken in the latest version.
 
I am using https://unpkg.com/react-jsonschema-form/dist/react-jsonschema-form.js CDN link.

loganvolkers:
Can you confirm that this is still the case? Can you provide a [SSCCE](http://sscce.org/), for example on codesandbox.io?

M-Reimer:
I don't know how long this will still be available, but I don't get the errors with:

`    <script scr='https://unpkg.com/react-jsonschema-form@1.0.3/dist/react-jsonschema-form.js'></script>
`

It would be really helpful if Mozilla provided a simple example on how to use the "CDN version" of this library. Just plain HTML with  embedded <script> directly in HTML and *without* JSX.

GuyPaddock:
I too am seeing this issue with 1.0.4. It does not happen with 1.0.3.

glasserc:
I think this is a duplicate of #991.

Issue #991, opened 2018-07-27
npm unpkg verion 1.0.4 is broken, 1.0.3 works


pieper:
### Prerequisites

- [x ] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [ x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I get an error loading a site with the current version on unpkg

### Steps to Reproduce

http://pieper.github.io/sites/schemaform/index.html

Gives this error:
```
Uncaught TypeError: a is not a function
    at Object.<anonymous> (es6.object.define-property.js:3)
    at t (bootstrap 7fcccad7da765f90bc6a:19)
    at Object.<anonymous> (define-property.js:1)
    at t (bootstrap 7fcccad7da765f90bc6a:19)
    at Object.<anonymous> (define-property.js:1)
    at t (bootstrap 7fcccad7da765f90bc6a:19)
    at Object.<anonymous> (react-jsonschema-form.js:879)
    at t (bootstrap 7fcccad7da765f90bc6a:19)
    at Object.<anonymous> (_hide.js:1)
    at t (bootstrap 7fcccad7da765f90bc6a:19)
```

If I change the script to use 1.0.3 it works:
```
  <script src="https://unpkg.com/react-jsonschema-form@1.0.3/dist/react-jsonschema-form.js"></script>
```

#### Expected behavior

1.0.4 shouldn't raise an error
#### Actual behavior

The error happens independent of my code.

### Version

1.0.4

graingert:
this is because we're transpiling core-js with babel-transform-runtime

graingert:
the fix is to either whitelist packages to transform with babel (eg react and react-dom) or blacklist `node_module\/((babel-[\w\d-]+)|core-js)/.*`

graingert:
@pieper I get `TypeError: titleEleement is null[Learn More]` on firefox

graingert:
@pieper can you open another bug - looks like onChange is getting called immediately after rendering

graingert:
@pieper looks like it was this change: https://github.com/mozilla-services/react-jsonschema-form/commit/9813483ab19f235690b85f9337610a4189979d73

PR #1039, opened 2018-10-01
Exclude core-js and babel-runtime from babel transformations

glasserc:

Discussion with @graingert suggests that issues #994, #991, and #1022 might be caused by a strange recursive webpack-babel-corejs configuration. Specifically, babel-runtime normally injects certain utility functions, or uses those of core-js, when it sees JS code that needs them. Our webpack configuration causes babel to be used on all our dependencies, including babel-runtime and core-js. Babel then injects the utility methods into the babel-runtime and core-js files that define the utility methods, leading to a circular dependency.

This fix excludes just babel-runtime and core-js from the `dist` Webpack configuration.

I'm not really sure how to test this without cutting a release. What I did to verify it at least a little bit:

- Remove `--optimize-minimize` from `build:dist` in `package.json`
- Run `npm run dist`
- Look through `dist/react-jsonschema-form.js`. `defineProperty` seems to be defined in `29`, which requires `30`, which requires `31`, which requires `32`, which requires `29`
- Make changes to webpack.config.dist.js
- Rerun `npm run dist`
- Look through `dist/react-jsonschema-form.js`. `29` requires `30`, which requires `31`, which does not require anything.

There may be more clever ways to tell Webpack not to do anything silly, but we're running Webpack v1.13.3 (current is v4.20.2).

See also the discussion at the end of #982.

### Checklist

* (n/a) **I'm updating documentation**
  - (n/a) I've checked the rendering of the Markdown text I've added
  - (n/a) If I'm adding a new section, I've updated the Table of Content
* (n/a) **I'm adding or updating code**
  - (n/a) I've added and/or updated tests
  - (n/a) I've updated docs if needed
  - (n/a) I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style
* (n/a) **I'm adding a new feature**
  - (n/a) I've updated the playground with an example use of the feature

### DIFF
--- a/webpack.config.dist.js
+++ b/webpack.config.dist.js
@@ -33,6 +33,10 @@ module.exports = {
       {
         test: /\.js$/,
         loaders: ["babel"],
+        exclude: [
+          path.join(__dirname, "node_modules", "core-js"),
+          path.join(__dirname, "node_modules", "babel-runtime"),
+        ],
       },
       {
         test: /\.json$/,


webpack.config.dist.js
```diff
@@ -33,6 +33,10 @@ module.exports = {
       {
         test: /\.js$/,
         loaders: ["babel"],
+        exclude: [
+          path.join(__dirname, "node_modules", "core-js"),
+          path.join(__dirname, "node_modules", "babel-runtime"),
```
graingert: I think it might work without this

graingert: (eg delete line 38, you only need line 37)

glasserc: When I removed the `babel-runtime` element of the list, I got the same results as before (described in the PR notes as `29` including `30` including `31` including 32` including `29`).

graingert: Oh ok, that's odd. LGTM otherwise

graingert:
@glasserc it might be a good idea to release a .pre version to npm then test it out in a website via `<script src="https://unpkg..."`

glasserc:
Good idea, thanks! I'll try that.

glasserc:
I tried one of the tips-and-tricks ones (specifically, https://jsfiddle.net/hdp1kgn6/1/), and it failed before I published -pre.1 and it seems to work now. I'll cut a real release..

graingert:
had a bit of a think about this - it might be better to use the regex form and exclude anywhere where '/node_module/core-js/' or '/node_modules/babel-runtime/' exists in the module path. That way it can cover nested deps with different babel-runtime or core-js versions

glasserc:
Thanks for the suggestion. I opened #1040 as you can see to try to get webpack into modern shape. I'm not sure whether the changes are even still necessary -- looking at the output, it doesn't seem like anything is including itself any more, even without the change. Are you familiar with any change that would make webpack smarter in this regard? Think it's worth cutting another prerelease release without the exclude clause and see what happens?

PR #1054, opened 2018-10-18
Add x-enumNames support

RicoSuter:
First: Thanks for this awesome project, works like a charm...

Add support for the x-enumNames property which has the same semantics as enumNames but uses the x- prefix to declare it as a custom/extension property. I'm maintaining the NJsonSchema C# library which generates these properties from .NET enums. See https://github.com/RSuter/NJsonSchema/wiki/Enums

I think it makes sense to support this scenario because this prefix is required/defined in the JSON Schema spec. It seems that the proposal for the property enumNames has been rejected in favor of const (see below). 

**Alternatives**

There are also const enums: https://github.com/json-schema-org/json-schema-spec/issues/67#issuecomment-300235090

Spec for const: https://github.com/json-schema-org/json-schema-spec/issues/58

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -369,7 +369,9 @@ export function allowAdditionalItems(schema) {
 export function optionsList(schema) {
   if (schema.enum) {
     return schema.enum.map((value, i) => {
-      const label = (schema.enumNames && schema.enumNames[i]) || String(value);
+      const label = (schema.enumNames && schema.enumNames[i]) || 
+                    (schema["x-enumNames"] && schema["x-enumNames"][i]) || 
+                    String(value);
       return { label, value };
     });
   } else {


RSuter:
x-enumNames is not in the spec but so is enumNames which is already supported - enumNames even violates json schema as custom extensions must start with x-. I maintain a json schema library which generate with x-...

epicfaace:
That's a good point. It looks like enumNames [was added in the past because it was in the v5 draft](https://github.com/mozilla-services/react-jsonschema-form/issues/25), and retrospectively it probably shouldn't have been added, given that it would have been better to wait until draft-06 was finalized.

It's worth noting, though, that adding custom attributes technically doesn't violate JSON Schema according to [draft-07](https://json-schema.org/latest/json-schema-core.html#rfc.section.6.4):

> Implementations MAY define additional keywords to JSON Schema. Save for explicit agreement, schema authors SHALL NOT expect these additional keywords to be supported by peer implementations. Implementations SHOULD ignore keywords they do not support.

@RSuter , where do you see in the spec that custom extensions should start with x-? The only place I can find that rule is with [Swagger](https://swagger.io/docs/specification/openapi-extensions/).

glasserc:
I don't even think `enumNames` ever even made it into draft 5 -- see #532. The alternative the committee proposed was to use `oneOf` with `const` and a `title`. Maybe the best thing would be to add special support for that? Alternatively, maybe we could add support for finding `enumNames` in the `uiSchema`?

glasserc:
Actually, I guess the support for `anyOf` and `title` has existed for a while #581. @RSuter, could you just export it that way instead?

RicoSuter:
I really dislike the way json schema handles enums: There is a native/clean way to describe enums but as soon as you need to add more metadata (names) you need to switch to a completely different construct which is less concise and much more verbose... I probably will not switch to oneOf enums as i’m pretty sure that most json schema/swagger/openapi tools (code gens) will not support that...

epicfaace:
Closing due to inactivity.

PR #1055, opened 2018-10-19
Added hideNowButton and hideClearButton options to alt-date widget

stamp:

I used the alt-date widget in a form for the user to add his/her birthdate and then the "now" button is a little out of place. So I added an option to hide it. And when i already was on it I added an option to hide the clear button as well.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,7 +352,7 @@ Please note that, even though they are standardized, `datetime-local` and `date`
 
 ![](http://i.imgur.com/VF5tY60.png)
 
-You can customize the list of years displayed in the `year` dropdown by providing a ``yearsRange`` property to ``ui:options`` in your uiSchema:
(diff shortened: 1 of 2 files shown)


glasserc:
This is starting to come up against the limit of what I think we should support. Maybe the right fix for your use case is just to define your own widget. What do you think?

stamp:
Ah, i could easily hide them with css or by a custom component so it does not matter that much to me. I just thought it could be nice if there were easy to hide and the picker works great without them (even better in some cases). 

One could argue that many of the other included widgets are highly customizeable and just the alt-date/alt-datetime widgets actually doesn't have that many options.

I think this is a line that you maintainers has to draw. I'm just glad to help any way :)

The react-jsonschema-form project is an awesome tool when creating webapps. I just love it!

swetapatnaik:
Is there an example for this? Or can you say what should be the code.

swetapatnaik:
Thanks a ton @stamp

Issue #854, opened 2018-02-28
Allow marking complete form as disabled / read only


k-nut:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);

### Description

I would like to be able to render a complete form `readonly` or `disabled`. This is to show it to users later as a validation step.  As far as I can tell this is not possible right now and I would have to add the respective property (`ui:disabled`/`ui:readonly`) for each input in the `UISchema`. 

Would you agree that it might be useful to just set this globally on the form? If so I'd start preparing a pull request for it.

glasserc:
I don't have a strong opinion about this. The standard HTML form controls support `readonly` and `disabled`, but only on form controls, not on the form element itself. I don't know if we should try to be doing anything more powerful than the underlying HTML abstraction does. I almost feel like we should be making it easier to perform other kinds of global transformations on schemas, perhaps by exporting utility functions that allow users to `map` over a schema to produce a UI schema or something?

ojab:
I'm also interested in this, it's useful for displaying whole form with only some fields enabled based on various conditions.
Basically it means that `props.disabled || uiSchema["ui:disabled"]` order should be changed [here](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/SchemaField.js#L170).

Maybe hide it behind some `uiSchema` flag, so it will not break current behaviour?

k-nut:
Now thinking about it again I think that I agree that it would make sense to not implement it like this but instead make it easier to generate a `uiSchema` from the `schema` or something like that. For the time being let me close this issue.

eberkund:
I am trying to make a view mode which displays the data but does not allow editing. So far I've gone with something like this,

    Object.keys(schema.properties).map(function (key, index) {
        uiSchema[key] = {'ui:readonly': true};
    });

I think going beyond standard HTML controls is necessary because the whole form is getting rendered by this component. Unless there is a way to disable or remove the 'Submit' button that I am not seeing.

IMO having a readonly attribute on the form could remove the submit button and set all the inner controls to readonly as well sort of like a display mode.

comwt:
For anyone looking into this in the future, the originally desired behavior (that of making the entire form readonly) can easily be accomplished, as follows:
`uiSchema = { "ui:readonly": true }`
No need to parse each property name of a JSON doc.

eberkund:
[...]

PR #1056, opened 2018-10-19
Added form disabled prop

stamp:

I needed a simple way to disable the whole form while storing the data in our backend. So I added a `disabled` prop on the `Form` component that propagates down to each widget in the form. It uses the already existing disable mechanics in the widgets so the change was pretty simple. 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ A [live playground](https://mozilla-services.github.io/react-jsonschema-form/) i
      - [Field labels](#field-labels)
      - [HTML5 Input Types](#html5-input-types)
      - [Form attributes](#form-attributes)
+     - [Form disable](#form-disable)
   - [Advanced customization](#advanced-customization)
      - [Field template](#field-template)
      - [Array Field Template](#array-field-template)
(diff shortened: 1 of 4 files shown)


bhunwal:
@stamp this is an awesome feature. can you please with what is the rationale behind having 
  disabled has property instead of a feature flag? IMO if we have a feature flag we can enable disable run time.

stamp:
I totally missed #854. I did it like this mostly because it follows the same pattern as when disabling any form element, like button for exampe:
```
<button onClick... disabled />
```

I use it like this:
```
class Demo exends Component {
  
  onSubmit = () => ({ formData }) => {
    this.setState({ disableForm: true });

    dispatch(saveChanges(formData)).then(() => {
      this.setState({ disableForm: false });
    });
  }

  render() {
    const { disableForm } = this.state;
    return (
      <Form
        schema={schema}
        disabled={disableForm}
        onSubmit={this.onSubmit()}
      />
    )
  }
}
```

landsman:
Great! Thanks!

Issue #500, opened 2017-03-01
Unable to trigger form submit/validation externally from the form


aackerman:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Requirements from my design team require the submit buttons to be logically separate in structure, both is usage of React and the DOM. I'm looking for a way to submit and/or validate the form externally or pass in errors that I have validated externally. It doesn't seem possible right now and I couldn't find a way to do it that didn't seem like a complete hack relying on the internals of RJSF.

### Steps to Reproduce

The following is essentially the crux of the structure. Our design team wants the buttons to be above the form in a header instead of below and the structure of the application prevents wrapping the header inside the form. Any help with the ability to submit/validate/display errors without having buttons as children of the form would be useful.

```js
<div>
  <Header>
    <button>Submit</button>
    <button>Cancel</button>
  </Header>
  <Form {...formProps}>
    <span/>
  </Form>
</div>
```

### Version

0.41.2

n1k0:
> I'm looking for a way to submit and/or validate the form externally or pass in errors that I have validated externally

While external submission isn't supported, the lib has support for [custom validation](https://github.com/mozilla-services/react-jsonschema-form/#custom-validation). 

> Our design team wants the buttons to be above the form in a header instead of below and the structure of the application prevent wrapping the header inside the form

This is rather weird semantically speaking. I'd probably try to position the regular form children buttons using CSS, but I'm not in your shoes.

glasserc:
Related to #155.

elisechant:
@aackerman 

Some idea: 

1. you might be able to make 2 linked forms with the same ui:rootFieldId and then use css or some other magic to hide and show the bits you need? https://github.com/mozilla-services/react-jsonschema-form#autogenerated-widget-ids

2. you could transclude the buttons and then add callback refs to them and then your additional buttons above the form, could trigger click on them. Or they could just trigger form submission: 

```
const YourForm = () => {
  return (
    <div>
      <button onClick={() => this.form.submit()}>Submit from above</button>
      <Form ref={(el => this.form = el)}
            schema={{foo:'bar'}} />
    </div>
  )
};
```

Not tested.

lucaas:
In our application we also have found the need to trigger form submission externally.

[...]

PR #1058, opened 2018-10-19
Add sumbit method to enable the programmatic submission

sunweiye:

In some user case, the form should be submitted by the javascript program programmatically, for example, the submit button in the UI design is separated from the form.

### Checklist


glasserc:
Refs #500. Thanks!

PR #1065, opened 2018-10-26
Issue-881 : Added support for onFocus and onBlur for BooleanField

kunal28889:

onFocus and onBlur callbacks are not currently supported for Checkboxes i.e BooleanFields. Added the code for passing on these callbacks as part of the props variable.

https://github.com/mozilla-services/react-jsonschema-form/issues/881

### Checklist

* [ ] **I'm updating documentation** (Not Needed, since onFocus is mentioned as already supported in the docs)

### DIFF
--- a/src/components/fields/BooleanField.js
+++ b/src/components/fields/BooleanField.js
@@ -60,6 +60,8 @@ if (process.env.NODE_ENV !== "production") {
     uiSchema: PropTypes.object,
     idSchema: PropTypes.object,
     onChange: PropTypes.func.isRequired,
+    onFocus: PropTypes.func,
+    onBlur: PropTypes.func,
     formData: PropTypes.bool,
     required: PropTypes.bool,
     disabled: PropTypes.bool,
--- a/src/components/fields/BooleanField.js
+++ b/src/components/fields/BooleanField.js
@@ -21,6 +21,8 @@ function BooleanField(props) {
     readonly,
     autofocus,
     onChange,
+    onFocus,
+    onBlur,
     rawErrors,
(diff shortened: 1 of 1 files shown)


src/components/fields/BooleanField.js
```diff
@@ -56,6 +60,8 @@ if (process.env.NODE_ENV !== "production") {
     uiSchema: PropTypes.object,
     idSchema: PropTypes.object,
     onChange: PropTypes.func.isRequired,
+    onFocus: PropTypes.func.isRequired,
+    onBlur: PropTypes.func.isRequired,
```
glasserc: Are these really required?

kunal28889: Agreed. Will back these out... I'd put them in there to keep the code similar to onChange event...

glasserc:
Shouldn't they remain in PropTypes but without `.required`?

kunal28889:
> Shouldn't they remain in PropTypes but without `.required`?

I'd thought since it's not necessary to keep it in there, I went ahead and removed it. But I agree, for the sake of conforming to React coding guidelines, we should keep it. Going to make the change now. Thanks for the inputs!

glasserc:
Fantastic, thanks!

kunal28889:
Thank you!

kunal28889:
@glasserc We're trying to pull v1.0.5, but this commit doesn't seem to be checked in for that version. Could you help out with when these changes would be visible ?

kunal28889:
Awesome! Appreciate the help!

PR #1066, opened 2018-10-26
Fix for new submit() method (PR #1058) also submitting the HTML form

sbusch:
Mew submit() method from PR #1058 also submitting the HTML form, which leads to navigating away from current page, at least in Firefox.

Reason: dispatched event was not cancelable, so preventDefault in onSubmit
couldn't cancel it.

Further information:

* <https://stackoverflow.com/a/40916998>
* <https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable>

### Checklist

  - ~[ ] I've added and/or updated tests~ (just fixing PR #1058 which didnt add new test)
  - ~[ ] I've updated docs if needed~ n/a

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -189,7 +189,7 @@ export default class Form extends Component {
 
   submit() {
     if (this.formElement) {
-      this.formElement.dispatchEvent(new Event("submit"));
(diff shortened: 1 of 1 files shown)


glasserc:
Thanks! I wonder if there's a better way to implement `submit()`..

PR #1068, opened 2018-10-26
Synchronous call to onSubmit() from props

sbusch:

Due to the use of `setImmediate()` hack in `setState` utility function from `utils.js`,
the `onSubmit()` handler from props is called asynchronously.

This leads to problems for submit handlers requiring browser "trusted events", like `window.open()` or
when programmatically submitting forms with target `_blank`.

Because `onSubmit()` should not need the performance-related `setImmiate()` hack, I
replaced call to `setState` utility function with proper `this.setState()` from
React, which fixed the problem.

### Checklist

  - [ ] I've added and/or updated tests: couldnt figure out how to properly test this.

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -165,7 +165,7 @@ export default class Form extends Component {
       }
     }
 
-    setState(this, { errors: [], errorSchema: {} }, () => {
+    this.setState({ errors: [], errorSchema: {} }, () => {
       if (this.props.onSubmit) {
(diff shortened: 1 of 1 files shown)


sbusch:
@althalus86

Do you have a reproducible full example?

Did you install via "npm install react-jsonschema-form"?

Is node_modules/react-jsonschema-form a regular directory or a symlink? 

Which react, react-dom and typescript version are you using? (npm ls output for each of them)

PR #1071, opened 2018-10-30
Add `prepare` script to package.json

leoc:
That way we can install from github master branch if a feature has not yet been released.

```
npm install --save mozilla-services/react-jsonschema-form
```

I want to use some features that are not yet available/released (submitting form programmatically).

### Checklist

It does not match that well with the checklist.


glasserc:
This is cool. Does this mean we no longer need to run `npm run dist` before publishing releases? If so, can we update the docs?

leoc:
I think you could, but that would mean that each user of the lib builds the dist themself from the same source. From my point of view, it still makes sense to build all distribution files once per release and serve them as part of the npm package, no?

glasserc:
Great, thanks for explaining that!

PR #1088, opened 2018-11-28
Pass disabled and readonly to object field template

olzraiti:

The template might want to render something if the object is `readonly` or `disabled`.

### Checklist


glasserc:
Thanks!

PR #1089, opened 2018-11-29
Fix infinite loop when setting formData to falsey value in onChange handler

### LINKED ISSUE #1086: Not able to control the`formData` in redux state
### Description

We get a formData object from the server that we allow the user to edit and we are using lodash `._set`-object to update parts of that formData depending on what part of the formData the user updates.

### Steps to Reproduce

We are trying to control the state of formData in redux state. We try to use `componentWillReceiveProps`, but can't get the updated state back as a props. We put our action in onChange function and it changes the state correctly when we type in text. But when we finish typing it return back to empty string value as it can't get updated state to the component as a props, and then the onChange function triggers the action once again because value of the input is empty.

We are trying to get the current value locally by setting the state inside the component but getting this error;

```Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops```

This is our function;

    onChange = ({ formData }) => {
    const { path, editKeyword } = this.props;
    const inputValue =formDataData[editKeyword];

    let prevformData = this.props.formData;

    // update the object using
    const updatedFormData = _.set(prevFormData, path, inputValue);
    
    // action
    this.props.setCurrentFormData(updatedformData);};

We also tried the following with no luck;

    onChange = ({ formData }) => {
    const { path, editKeyword } = this.props;
    const inputValue = formData[editKeyword];
    console.log("input value ", inputValue);

    let prevFormData = this.props.formData;

    const updatedFormData = _.set(prevformData, path, inputValue);

    // when we try to update local state
    this.setState({ FormData: updatedformData})

  };

Anyone encountered anything similar? We can't directly use the formData from the form-input  because we are merging values from server and the user input.

LucianBuzzo:
Fixes #1086

Change-type: patch

This fixes a bug where the `<Form>` component would go into an infinite loop if the `onChange` handler set `formData` to a falsey value. See #1086 for more details.

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -38,13 +38,14 @@ export default class Form extends Component {
 
   componentWillReceiveProps(nextProps) {
     const nextState = this.getStateFromProps(nextProps);
-    this.setState(nextState);
     if (
       !deepEquals(nextState.formData, nextProps.formData) &&
+      !deepEquals(nextState.formData, this.state.formData) &&
       this.props.onChange
     ) {
       this.props.onChange(nextState);
     }
+    this.setState(nextState);
   }
 
   getStateFromProps(props) {
--- a/test/Form_test.js
+++ b/test/Form_test.js
@@ -5,7 +5,11 @@ import { renderIntoDocument, Simulate } from "react-addons-test-utils";
 import { findDOMNode } from "react-dom";
 
 import Form from "../src";
-import { createFormComponent, createSandbox } from "./test_utils";
+import {
+  createComponent,
+  createFormComponent,
+  createSandbox,
+} from "./test_utils";
 
 describe("Form", () => {
   let sandbox;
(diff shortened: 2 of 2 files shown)


LucianBuzzo:
@glasserc Any chance of reviewing this PR?

LucianBuzzo:
@JamieDixon Not to my knowledge. The last published version was 2 months ago. Perhaps @edi9999 Can increment a patch version and publish the package?

LucianBuzzo:
@JamieDixon If you take a look at my repsonse here https://github.com/mozilla-services/react-jsonschema-form/issues/1086#issuecomment-442739138 you might be able to tweak your code to avoid this issue.

JamieDixon:
@LucianBuzzo Thanks. For now I've forked and published a scoped package under 1.0.7.

RE your comment, it seems it's not always `falsey` values that cause the issue. See here a modified version of your minimum case where I set the formData to be an object with a single key-value pair: https://codesandbox.io/s/xvwyylz17o This example also has the issue described.

I can happily confirm that your fix does solve my issue though so this is great news and I'm very grateful.

edi9999:
I don't have access to the npm package but I would happily publish if I had the rights to do so. My npm username is edi9999

LucianBuzzo:
@JamieDixon v1.0.7 has been released and includes this fix

JamieDixon:
@LucianBuzzo Thank you! 🚀 🙂

Issue #1099, opened 2018-12-06
multiple bugs with additionalProperties component


aerfio:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

- There is no option to remove key-value pair once it's added
- Key changes name when doubleclicked
- Key-value pair also sometimes change position, when you add more pairs, change their keys and doubleclick again

### Steps to Reproduce

1. Loot at provided jsfiddle
2. Doubleclick "newKey Key"
3. It changes name to "newKey-1 Key"
4. Doubleclick again
5. "newKey-1-1 Key"
6. add one more pair
7. click key again
8. they should change their order

https://jsfiddle.net/aerfio/mynuxdc4/

#### Expected behavior

I expect to have option to remove once added key-value pair and to preserve order of those pairs.

#### Actual behavior

No "remove/delete/X" button, ordering is not preserved 

### Version

1.0.6

Related: #228

aerfio:
@christianclent, can you have a look at this? Are working on this, can we help somehow?

domharrington:
Also when your cursor is inside the "key" input, pressing tab has the same effect of renaming to append `-1` to the name.

domharrington:
It would also be great to get an additional class name here so this can be styled:
https://github.com/mozilla-services/react-jsonschema-form/blob/0ac0da7eb3fb99bb76017572dd10ef7a11ffadcd/src/components/fields/SchemaField.js#L133-L134

loganvolkers:
Wow, so many bugs with additional properties. 

This should have been released on an alpha branch, as it's clearly early functionality that create backwards incompatible renderings of forms and functionality.

glasserc:
#1123 is listed as fixing all these issues. Please try it and let me know.

`additionalProperties` was marked as "preliminary" in #1021 and I didn't expect it to be perfect, but as forms don't get the functionality unless they include an `additionalProperties` field, it seemed like it was opt-in and not a problem to just release. Sorry if the preliminary nature of the feature wasn't communicated clearly. However, I don't think "releasing it on an alpha branch" would have helped since my impression is that people only want to use rjsf via the latest npm release.

PR #1101, opened 2018-12-07
Key-value pair in additional properties retains order when edited

aerfio:

When you change key it shows on the bottom of the list, instead of keeping its place, keeping the order of key-value pairs.

### Checklist

Solves two things of #1099

### DIFF
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -106,11 +106,14 @@ class ObjectField extends Component {
     return (value, errorSchema) => {
       value = this.getAvailableKey(value, this.props.formData);
       const newFormData = { ...this.props.formData };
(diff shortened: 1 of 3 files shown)


aerfio:
Hey guys, any comment on this?

edi9999:
Can you add a test case in the same form as https://github.com/mozilla-services/react-jsonschema-form/blob/master/test/ObjectField_test.js#L582 that fails before your patch, and succeeds afterwards ?

aerfio:
@edi9999 Done

edi9999:
Great , merging !

aerfio:
@edi9999 When do you guys plan to realease 1.0.7? Do you have it scheduled? :d

Issue #86, opened 2016-03-23
Add links to fields in error in the Error section at the top of the form


n1k0:
This should be trivial as we have unique ids for each widget.

tiemevanveen:
Not [RadioWidget](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/widgets/RadioWidget.js)..

elisechant:
this would be awesome! any plans to make this happen?

n1k0:
No, would you want to work on this?

elisechant:
Ok, I can.

sahithi2018:
is some one working on this,if not i can take it up?

PR #1102, opened 2018-12-08
Give ids to array and object fields

epicfaace:
This change:
- Gives ids to the `<fieldset>` tag in the default array field template and object field template
- Gives the id of "root" to the topmost `<fieldset>` in the form.

This way, we are more consistent with the idSchema, so that every id mentioned in the idSchema actually has a corresponding element rendered in the DOM. This allows us to create anchor links that go to specific elements, even if they are not `<input>` or `<select>` elements.

For example, on the "Arrays" example in the [playground](https://mozilla-services.github.io/react-jsonschema-form/), here is the idSchema:
image
However, there are no elements with ids equal to "root", "root_fixedItemsList", etc. This change fixes that.

This change was especially motivated by looking into #86 -- sometimes you may have a validation error that refers to an *array* element (such as minimum items not being reached). If there is no corresponding DOM ID corresponding to the idSchema value for that element, then we cannot create anchor links that go to the specific element with the error.

### Checklist


glasserc:
Thanks for the great PR description!

glasserc:
Also, sorry for taking so long to get to this!

Issue #1104, opened 2018-12-12
BUG: Unwanted error with usage of schema dependencies oneOf


pieplu:
### Prerequisites

- [X] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

![image](https://user-images.githubusercontent.com/3260285/49846160-e47bae80-fd97-11e8-94c1-09b9179aaf44.png)

In some case, a false error appears. (When a schema use dependencies, oneOf, and a "format" on type number)

After testing and read the code, it's looks like the ajv object used for the validation, still have errors on the "ajv.errors" properties.

The function `validateFormData` is used to validate the form AND to [retrieve Schema](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/utils.js#L584).

With a schema who have dependencies, `retieveSchema` is called many time for small piece of the schema, and get error each time formData don't fit with oneOf schema dependencies.

It's looks like, when the schema contain a number with a format (and maybe other case), the ajv object don't clean up the last errors.

Because of that, we have a error schema with a persisting false error, at the root.

### Steps to Reproduce

1. Go on https://codesandbox.io/s/o1jp047xmz
2. Change **typeRequisition** value to "*TONNAGE*"
3. Submit

An error should be visible on the top Panel (but not on the field)

#### Expected behavior

No error

#### Actual behavior

False and not conform error (in more, at the root of the errorSchema...)

### Version

1.0.6 (same on 1.0.3)

### Fix

A PR with fix will follow, a simple fix look's like to set ajv.errors to null each time before validate

PR #1105, opened 2018-12-12
Fix: #1104 Clear ajv errors before each validate

pieplu:

Resolve issue #1104

### Checklist

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -155,7 +155,7 @@ export default function validateFormData(
   transformErrors
 ) {
   try {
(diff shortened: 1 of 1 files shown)


glasserc:
Yikes. Thanks for the bug report and the fix. Sorry for taking so long to get to this. The fact that `ajv.errors` is how state gets reported is unfortunate. Fortunately, nobody else should have access to our `ajv` instance, so I think clearing it after we get the errors out will be sufficient. I'm going to merge this and clean it up a little bit myself.

PR #940, opened 2018-06-01
[WIP]feat: supporting anyof

picodoth:
### Reasons for making this change

Supporting anyof type including refs
demo

Continuing work of supporting https://github.com/mozilla-services/react-jsonschema-form/issues/52 based on https://github.com/mozilla-services/react-jsonschema-form/pull/417

TODO items

- [ ] adding test

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've checked the rendering of the Markdown text I've added
  - [ ] If I'm adding a new section, I've updated the Table of Content
* [x] **I'm adding or updating code**
  - [ ] I've added and/or updated tests
  - [ ] I've updated docs if needed
  - [ ] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style
* [x] **I'm adding a new feature**
  - [x] I've updated the playground with an example use of the feature

PR #1118, opened 2019-01-03
Add preliminary support for the anyOf keyword

LucianBuzzo:
This change introduces support for the `anyOf` keyword. Other PRs that add support for `anyOf`/`oneOf` exist, but are stale and/or incomplete ( #302, #417, #940 ).
The keyword is NOT supported in arrays (using the `items` keyword).
Options provided in the `anyOf` keyword are rendered using a new component called `AnyOfField`, this can be overridden with a custom field using the `fields` property.

I'd like to follow-up on this PR with work to add support for `anyOf` in arrays and `oneOf`. I decided against submitting that work in this PR to help keep reviews and discussion manageable.

anyof-form

### Checklist


LucianBuzzo:
Note that the unit tests are failing due to the New year rollover, which I've fixed/band-aided in a separate PR here https://github.com/mozilla-services/react-jsonschema-form/pull/1119

cyplo:
This looks good to me as a change and would be extremely useful if merged. Thanks ! :)

LucianBuzzo:
@edi9999 Could you review this PR please?

test/anyOf_test.js
```diff
@@ -0,0 +1,272 @@
...
+      properties: {
+        foo: { type: "string" },
+      },
+    };
+
+    const { node } = createFormComponent({
+      schema,
+    });
+
+    expect(node.querySelectorAll("select#root_anyof_select")).to.have.length.of(
```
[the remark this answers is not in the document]
LucianBuzzo: Good point, I'll remove the ID selectors

test/anyOf_test.js
```diff
@@ -0,0 +1,272 @@
...
+      1
+    );
+  });
+
+  it("should render a select element if the anyOf keyword is present", () => {
```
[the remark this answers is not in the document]
LucianBuzzo: This is a typo, the whole test block is duplicated by mistake

LucianBuzzo:
@edi9999 I've addressed your review comments

src/components/fields/AnyOfField.js
```diff
@@ -0,0 +1,147 @@
...
+              return (
+                <option key={index} value={index}>
+                  {option.title || `Option ${index + 1}`}
+                </option>
+              );
+            })}
+          </select>
+        </div>
+
+        {options.map((option, index) => {
```
glasserc: Using `map` here is a little unusual, no? Why not something like `let option = selectedOption && options[selectedOption]`?

src/components/fields/AnyOfField.js
```diff
@@ -0,0 +1,147 @@
...
+    formData: PropTypes.any,
+    errorSchema: PropTypes.object,
+    registry: PropTypes.shape({
+      widgets: PropTypes.objectOf(
+        PropTypes.oneOfType([PropTypes.func, PropTypes.object])
+      ).isRequired,
+      fields: PropTypes.objectOf(PropTypes.func).isRequired,
+      definitions: PropTypes.object.isRequired,
+      ArrayFieldTemplate: PropTypes.func,
+      ObjectFieldTemplate: PropTypes.func,
```
glasserc: Why do we need `ArrayFieldTemplate` or `ObjectFieldTemplate`?

src/components/fields/SchemaField.js
```diff
@@ -296,7 +304,35 @@ function SchemaFieldRender(props) {
...
+  return (
+    <FieldTemplate {...fieldProps}>
+      {field}
+
+      {/*
+        If the schema `anyOf` can be rendered as a select control, don't
+        render the `anyOf` selection and let `StringField` component handle
+        rendering
+      */}
+      {schema.anyOf && !isSelect(schema) && (
```
glasserc: I think this logic should move to `getFieldComponent`.

src/components/fields/AnyOfField.js
```diff
@@ -0,0 +1,147 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { guessType } from "../../utils";
+
+class AnyOfField extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      selectedOption: 0,
```
glasserc: It just occurred to me that this might not be safe if you are rendering a form from existing data. Should this try to guess the index based on the formData/anyOf schemas?

src/components/fields/AnyOfField.js
LucianBuzzo: Although, looking at this code further, is this not just describing the shape of the registry object?

src/components/fields/AnyOfField.js
LucianBuzzo: Yeah I think thats sensible behaviour, I'll add some tests for it as well

src/components/fields/SchemaField.js
LucianBuzzo: Could you elaborate on this? Should `getFieldComponent` return the `AnyOfField` component?

src/components/fields/AnyOfField.js
glasserc: To be honest, I'm not a React expert so I'm not really sure myself. It feels silly for us to have to redescribe the registry object's shape in every single component. If you want to leave it as-is for now and file a bug to clean it up later, that's fine.

src/components/fields/SchemaField.js
glasserc: Hmm. I see your concern -- the `AnyOfField` as it is right now isn't really the field, but rather something that is used alongside a field. I guess this is workable but it seems bad to me because it splits the logic about what components are used and when. Is it maybe possible to make `AnyOfField` a wrapper around another field (i.e. renders itself and then renders `props.children`)? Then we could have `getFieldComponent` return the `AnyOfField` component.

README.md
```diff
@@ -1803,7 +1803,12 @@ This component follows [JSON Schema](http://json-schema.org/documentation.html)
 * `additionalItems` keyword for arrays
     This keyword works when `items` is an array. `additionalItems: true` is not supported because there's no widget to represent an item of any type. In this case it will be treated as no additional items allowed. `additionalItems` being a valid schema is supported.
 * `anyOf`, `allOf`, and `oneOf`, or multiple `types` (i.e. `"type": ["string", "array"]`
-    Nobody yet has come up with a PR that adds this feature with a simple and easy-to-understand UX.
+    The `anyOf` keyword is supported but has the following caveats:
+    - The `anyOf` keyword is not supported when used inside the `items` keyword
+      for arrays.
+    - Properties declared inside the `anyOf` should not overlap with properties
+      "outside" of the `anyOf`.
```
glasserc: We can also update the FAQ entry about how no one has ever come up with a decent PR for `anyOf`.

src/components/fields/AnyOfField.js
LucianBuzzo: @glasserc I think we can use `PropTypes.shape()` to create a re-usable definition for the registry

src/components/fields/SchemaField.js
LucianBuzzo: I don't know if that would work either, as `AnyOfField` could render multiple different field components. In `AnyOfField`, rendering is delegated back to `SchemaField` which in turn calls `getFieldComponent`, so `getFieldComponent` still ultimately controls how a schema entry is rendered.
I could put the `AnyOfField` inside `getFieldComponent`, but we then have a situation where we have to add more wrapper markup (because of supporting react 15 we don't have access to `React.Fragment`) to contain the two elements and all of the properties for the `AnyOfField` component need to be passed to `getFieldComponent` so that the `AnyOfField` can be rendered correctly. This seems like a lot of additional complexity for little benefit.

glasserc: OK, you've convinced me. Thanks for thinking this through.

glasserc:
Fantastic, thanks. I'll cut 1.2.0 in a few minutes.

strahius:
Is this missing from the playground?

LucianBuzzo:
@strahius I think the playground hasn't been updated, you can see it in action if you run the playground from source.

strahius:
You are right, it just hasn't been deployed yet probably. Thanks a lot, good job!

PR #1119, opened 2019-01-03
Fix failing unit tests related to time widget

LucianBuzzo:

It appears that due to the recent change from 2018 to 2019, there is one additional year in the select list, which caused the unit tests to fail.

### Checklist

### DIFF
--- a/test/StringField_test.js
+++ b/test/StringField_test.js
@@ -773,7 +773,8 @@ describe("StringField", () => {
       );
 
       expect(lengths).eql([
-        121 + 1, // from 1900 to 2020 + undefined
+        // from 1900 to current year + 2 (inclusive) + 1 undefined option
+        new Date().getFullYear() - 1900 + 3 + 1,
         12 + 1,
         31 + 1,
         24 + 1,
@@ -1019,7 +1020,8 @@ describe("StringField", () => {
       );
 
       expect(lengths).eql([
-        121 + 1, // from 1900 to 2020 + undefined
+        // from 1900 to current year + 2 (inclusive) + 1 undefined option
+        new Date().getFullYear() - 1900 + 3 + 1,
         12 + 1,
         31 + 1,
       ]);


test/StringField_test.js
```diff
@@ -773,7 +773,7 @@ describe("StringField", () => {
       );
 
       expect(lengths).eql([
-        121 + 1, // from 1900 to 2020 + undefined
+        122 + 1, // from 1900 to 2020 + undefined
```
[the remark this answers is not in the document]
LucianBuzzo: I tested this by setting my computers date to various different years and it worked great :+1:

edi9999:
For some reason this failed on travis

LucianBuzzo:
@edi9999 Strange, I'm going to try and debug it

edi9999:
One idea of mine is that for some reason `new Date().getFullYear() ` returns 2018.

Can you add a test for this ?

LucianBuzzo:
@edi9999 The test passes when using `.only`, and the available options appear to be the same in both cases:
- https://travis-ci.org/mozilla-services/react-jsonschema-form/jobs/476279908#L2190
- https://travis-ci.org/mozilla-services/react-jsonschema-form/jobs/476277009

LucianBuzzo:
@edi9999 I was being stupid, it was actually another test that was failing -> https://github.com/mozilla-services/react-jsonschema-form/blob/master/test/StringField_test.js#L1007 I'll update it with the same fix

edi9999:
Thanks @LucianBuzzo  !

test/StringField_test.js
```diff
@@ -773,7 +773,8 @@ describe("StringField", () => {
       );
 
       expect(lengths).eql([
-        121 + 1, // from 1900 to 2020 + undefined
+        // from 1900 to current year + 2 (inclusive) + 1 undefined option
```
cyplo: A nitpick - but would it make sense to extract a function that provides this, to avoid duplication between tests ?

edi9999: That is imo not really necessary, in tests duplication is not so bad.

PR #1125, opened 2019-01-06
Handle sort arrays better

christianeide:

The current sortfunctions for arrayFields will only work if you shuffle an item 1 step up or down. If you create your own ArrayFieldTemplate with drag and drop, users can move items in larger steps, and this library will at this time not handle this. 

This pull simply sorts the moved array item from its old index to its new index.

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -277,17 +277,16 @@ class ArrayField extends Component {
         }
       }
       onChange(
-        formData.map((item, i) => {
-          // i is string, index and newIndex are numbers,
-          // so using "==" to compare
-          if (i == newIndex) {
-            return formData[index];
-          } else if (i == index) {
-            return formData[newIndex];
-          } else {
-            return item;
-          }
-        }),
+        function() {
+          // Copy item
+           let newFormData = formData.slice();
+ 
+           // Moves item from index to newIndex
+           newFormData.splice(index, 1);
+           newFormData.splice(newIndex, 0, formData[index]);
+ 
+           return newFormData
+         }(),
         newErrorSchema
       );
     };
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -277,16 +277,16 @@ class ArrayField extends Component {
         }
       }
       onChange(
-        function() {
+        (function() {
           // Copy item
-           let newFormData = formData.slice();
- 
-           // Moves item from index to newIndex
-           newFormData.splice(index, 1);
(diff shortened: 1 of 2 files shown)


src/components/fields/ArrayField.js
```diff
@@ -277,17 +277,16 @@ class ArrayField extends Component {
...
+        (function() {
+          // Copy item
+          let newFormData = formData.slice();
+
+          // Moves item from index to newIndex
+          newFormData.splice(index, 1);
+          newFormData.splice(newIndex, 0, formData[index]);
+
+          return newFormData;
+        })(),
```
glasserc: IIFE here is a bit ugly.

christianeide: Agree, was just trying to keep the same structure as before with an inline map-function. Moved out to its own function now

christianeide:
I think this change is mostly covered by two existing tests, `it("should move down a field from the list")` and `it("should move up a field from the list")`.

I have however tried to create a new test that moves an object more than one index. But then, as far as I understand, I have to have access to the onReorderClick-function thats being passed down to elements in a custom ArrayField. My skills in chai is a little bit limited, so I haven't found a way to get this data. Any suggestions on how I could set up a test like that?

glasserc:
The `onReorderClick` is passed to the template the `ArrayField` uses to render all the items, so I think we should try to define a custom `ArrayFieldTemplate` within the test. Since it's just for testing, we could have it render a button for each item and for each position. Something like this (untested!):

```
function moveAnywhereArrayItemTemplate(props) {
  const buttons = [];
  for (let i = 0; i < 3; i++) {
    buttons.append(
      <button 
        className={"array-item-move-to-"+i}
        onClick={props.onReorderClick(props.index, i)}>
        {i}
      </button>
    );
  }
  return (
    <div class={"item-" + props.index}>
      {props.children}
      {buttons}
    </div>
  );
}

function moveAnywhereArrayFieldTemplate(props) {
  return (
    <div class="array">
      {props.items.map(moveAnywhereArrayItemTemplate)}
    </div>
  );
}

const { node } = createFormComponent({
  schema,
  formData: ["foo", "bar", "baz"],
  ArrayFieldTemplate: moveAnywhereArrayFieldTemplate,
});

Simulate.click(node.querySelectorAll(".item-0 .array-item-move-to-2"));
// This part is cribbed from "should move down a field from the list"
const inputs = node.querySelectorAll(".field-string input[type=text]");
expect(inputs[0].value).eql("bar");
expect(inputs[1].value).eql("baz");
expect(inputs[2].value).eql("foo");
```

PR #1129, opened 2019-01-16
Fix bug where matching anyOf branch is not selected correctly

LucianBuzzo:
This change improves the logic that selects a matching anyOf branch
based on form data. Previously the behaviour was to just check if the
form data was valid against an anyOf branch, however due to the
permissive nature of JSON schema this has unexpected behaviour. For
example, given the following schema:

```json
{
  "type": "object",
  "anyOf": [
    {
      "properties": {
        "foo": {
          "type": "string"
        }
      }
    },
    {
      "properties": {
        "bar": {
          "type": "string"
        }
      }
    }
  ]
}
```

The form data `{ bar: 'baz' }` will actually match the first branch. To
mitigate this, when doing the matching, the branch schema is augmented
to require at least one of the keys in the branch. For example the
schema above would become:

```json
{
  "type": "object",
  "anyOf": [
    {
      "properties": {
        "foo": {
          "type": "string"
        }
      },
      "anyOf": [
        { "required": [ "foo" ] }
      ]
    },
    {
      "properties": {
        "bar": {
          "type": "string"
        }
      },
      "anyOf": [
        { "required": [ "bar" ] }
      ]
    }
  ]
}
```

### Checklist

### DIFF
--- a/src/components/fields/MultiSchemaField.js
+++ b/src/components/fields/MultiSchemaField.js
@@ -30,7 +30,49 @@ class AnyOfField extends Component {
 
   getMatchingOption(formData, options) {
     for (let i = 0; i < options.length; i++) {
-      if (isValid(options[i], formData)) {
+      const option = options[i];
+
+      // If the schema describes an object then we need to add slightly more
+      // strict matching to the schema, because unless the schema uses the
+      // "requires" keyword, an object will match the schema as long as it
+      // doesn't have matching keys with a conflicting type. To do this we use an
+      // "anyOf" with an array of requires. This augmentation expresses that the
+      // schema should match if any of the keys in the schema are present on the
+      // object and pass validation.
+      if (option.properties) {
+        // Create an "anyOf" schema that requires at least one of the keys in the
+        // "properties" object
+        const requiresAnyOf = {
+          anyOf: Object.keys(option.properties).map(key => ({
+            required: [key],
+          })),
+        };
+
+        let augmentedSchema;
+
+        // If the "anyOf" keyword already exists, wrap the augmentation in an "allOf"
+        if (option.anyOf) {
+          // Create a shallow clone of the option
+          const { ...shallowClone } = option;
+
+          if (!shallowClone.allOf) {
+            shallowClone.allOf = [];
+          } else {
+            // If "allOf" already exists, shallow clone the array
+            shallowClone.allOf = shallowClone.allOf.slice();
+          }
+
+          shallowClone.allOf.push(requiresAnyOf);
+
+          augmentedSchema = shallowClone;
+        } else {
+          augmentedSchema = Object.assign({}, option, requiresAnyOf);
+        }
+
+        if (isValid(augmentedSchema, formData)) {
+          return i;
+        }
+      } else if (isValid(options[i], formData)) {
         return i;
       }
     }
--- a/test/anyOf_test.js
+++ b/test/anyOf_test.js
@@ -300,6 +300,235 @@ describe("anyOf", () => {
     expect(node.querySelector("select").value).eql("1");
   });
 
+  it("should not change the selected option when entering values", () => {
+    const schema = {
+      type: "object",
+      anyOf: [
+        {
+          title: "First method of identification",
+          properties: {
+            firstName: {
+              type: "string",
+            },
+            lastName: {
+              type: "string",
+            },
+          },
+        },
+        {
+          title: "Second method of identification",
+          properties: {
+            idCode: {
+              type: "string",
+            },
+          },
+        },
+      ],
+    };
+
+    const { node } = createFormComponent({
+      schema,
+    });
+
+    const $select = node.querySelector("select");
+
+    expect($select.value).eql("0");
+
+    Simulate.change($select, {
+      target: { value: $select.options[1].value },
+    });
+
+    expect($select.value).eql("1");
+
+    Simulate.change(node.querySelector("input#root_idCode"), {
+      target: { value: "Lorem ipsum dolor sit amet" },
+    });
+
+    expect($select.value).eql("1");
+  });
+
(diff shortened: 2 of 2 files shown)


LucianBuzzo:
@glasserc If you have time a review would be great! This fixes an annoying bug with my original `anyOf` PR.

src/components/fields/AnyOfField.js
```diff
@@ -30,7 +30,45 @@ class AnyOfField extends Component {
...
+
+        // If the "anyOf" keyword already exists, wrap it in an "allOf"
+        if (option.anyOf) {
+          const { anyOf, ...rest } = option;
+
+          if (!rest.allOf) {
+            rest.allOf = [];
+          }
+
+          rest.allOf.push(...[requiresAnyOf, { anyOf }]);
```
glasserc: I think we should be careful not to mutate `rest.allOf` (because `rest.allOf` is a shallow copy of `option.allOf`).

I think `arr.push(...[a, b])` is the same as `arr.push(a, b)`. I don't think we need to add the existing `anyOf` to `allOf` -- each schema keyword is implicitly `allOf`-ed, right?

test/anyOf_test.js
```diff
@@ -299,4 +299,51 @@ describe("anyOf", () => {
...
+      target: { value: $select.options[1].value },
+    });
+
+    expect($select.value).eql("1");
+
+    Simulate.change(node.querySelector("input#root_idCode"), {
+      target: { value: "Lorem ipsum dolor sit amet" },
+    });
+
+    expect($select.value).eql("1");
```
glasserc: Do we need a test this complicated? Can't we just render it with `{idCode: "something"}` and see if the select shows the right value?

src/components/fields/AnyOfField.js
LucianBuzzo: Good catch on the shallow copy, I'll fix that and add a test that checks for mutation.
You're right, the spread operator in `.push` is unnecessary - I'll remove it.
Regarding the `allOf/anyOf` merging, I read back through the code, and you're correct, its not necessary in this case. I think this is a result of some refactoring I was doing to simplify the code paths, I'll refactor again and clean it up.

test/anyOf_test.js
LucianBuzzo: This test specifically checks to see that the selected value remains as expected *after* the user inputs a value, which is subtly different to checking that the correct value is selected when data is already present. The other validations in the test are sanity checks to make sue the form is in the state we expect before inputting a value.
It might seem a little verbose, but the verbosity will help ensure that there are no future regressions or that the test reports false positives.

LucianBuzzo:
@glasserc Thanks for the detailed review :+1: . I've addressed your comments and the PR should be good to go now.

epicfaace:
@LucianBuzzo I'm wondering, why did you wrap the required attribute in another `anyOf` -- couldn't you have created an augmented schema like this?:

```
{
  "type": "object",
  "anyOf": [
    {
      "properties": {
        "foo": {
          "type": "string"
        }
      },
      "required": [ "foo" ]
    },
    {
      "properties": {
        "bar": {
          "type": "string"
        }
      },
      "required": [ "bar" ]
    }
  ]
}
```

LucianBuzzo:
@epicfaace Wrapping the `required` in `anyOf` is a trick to allow you to require at least one of a list of keys, which is what you need to do if the subschema is a more complex object, for example:

```
{
  "type": "object",
  "anyOf": [
    {
      "properties": {
        "foo": {
          "type": "string"
        },
        "bar": {
          "type": "string"
        }
      },
      "anyOf": [
        { "required": [ "foo" ] },
        { "required": [ "bar" ] }
      ]
    },
    {
      "properties": {
        "buzz": {
          "type": "string"
        }
      },
      "anyOf": [
        { "required": [ "buzz" ] }
      ]
    }
  ]
}
```

If you just use a plain `required` statement, then the first object wouldn't match if you start typing into the `foo` field, but the `bar` field is still empty.

src/components/fields/MultiSchemaField.js
```diff
@@ -30,7 +30,49 @@ class AnyOfField extends Component {
...
+        // "properties" object
+        const requiresAnyOf = {
+          anyOf: Object.keys(option.properties).map(key => ({
+            required: [key],
+          })),
+        };
+
+        let augmentedSchema;
+
+        // If the "anyOf" keyword already exists, wrap the augmentation in an "allOf"
```
epicfaace: I'm guessing that once rjsf implements "allOf", we'll need additional logic here -- maybe we should either 1) proactively add it in or just 2) add a FIXME comment?

test/anyOf_test.js
```diff
@@ -300,6 +300,109 @@ describe("anyOf", () => {
     expect(node.querySelector("select").value).eql("1");
   });
 
+  it("should not change the selected option when entering values", () => {
```
epicfaace: I think you should add a test just like this, but when the `anyOf` keyword already exists within one of the `anyOf` branches?

src/components/fields/MultiSchemaField.js
LucianBuzzo: @epicfaace Unless I'm mistaken, this should already work with a schema that contains `allOf`. If the `allOf` key is present, we shallow clone it and append the new subschema.

epicfaace: Oh, yeah, you're right -- thanks!

test/anyOf_test.js
epicfaace: Similarly a test in which the `allOf` keyword exists? (if that's even possible to do, given that rjsf doesn't support allOf yet)

LucianBuzzo: Good idea, I'll add them in

PR #1130, opened 2019-01-16
Add more schemas to validate against

aerfio:

Right now we have no clear way to validate data using anything other then draft-6, which is default for ajv version 5. By updating ajv to newest available version (6.7.0) and making changes listed here: https://github.com/epoberezkin/ajv#using-version-6 we would be able to validate data using other drafts. There are no breaking changes between v5 and v6: https://github.com/epoberezkin/ajv/releases/tag/v6.0.0

### Checklist

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -49,7 +49,7 @@ render((
 
 ### Custom schema validation
 
-To have your schemas validated against any other meta schema than draft-07 (the current version of [JSON Schema](http://json-schema.org/)), make sure your schema has `$schema` field that enables validator to use correct meta schema. For example:
+To have your schemas validated against any other meta schema than draft-07 (the current version of [JSON Schema](http://json-schema.org/)), make sure your schema has a `$schema` attribute that enables the validator to use the correct meta schema. For example:
 
 ```json 
 {
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -68,7 +68,7 @@ render((
         metaSchema={metaSchema} />
 ), document.getElementById("app"));
 ```
-`metaSchema` also accepts array of meta schemas:
+`metaSchema` also accepts an array of meta schemas:
 
 ```jsx
 render((
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -58,7 +58,7 @@ To have your schemas validated against any other meta schema than draft-07 (the 
 }
 ```
 
-Then pass your meta schemas as `metaSchema` prop to your to your `Form` component:
+Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas, you need to install `ajv` and import them, then pass your meta schemas through the `metaSchema` prop to your `Form` component:
 
 ```jsx
 const metaSchema=YOUR_SCHEMA;
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -65,13 +65,13 @@ const additionalMetaSchemas = require("ajv/lib/refs/json-schema-draft-04.json");
 
 render((
   <Form schema={schema} 
-        additionalMetaSchemas={additionalMetaSchemas}/>
+        additionalMetaSchemas={[additionalMetaSchemas]}/>
 ), document.getElementById("app"));
 ```
 
 In this example `schema` passed as props to `Form` component can be validated against draft-07 (default) and by draft-04 (added), depending on the value of `$schema` attribute.
 
-`additionalMetaSchemas` also accepts an array of meta schemas:
+`additionalMetaSchemas` also accepts more than one meta schema:
 
 ```jsx
 render((
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -61,7 +61,7 @@ To have your schemas validated against any other meta schema than draft-07 (the 
 Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas, you need to install `ajv` and import them, then pass your meta schemas through the `metaSchema` prop to your `Form` component:
 
 ```jsx
-const metaSchema=YOUR_SCHEMA;
+const metaSchema= require("ajv/lib/refs/json-schema-draft-04.json");
 
 render((
   <Form schema={schema}
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -61,7 +61,7 @@ To have your schemas validated against any other meta schema than draft-07 (the 
 Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas, you need to install `ajv` and import them, then pass your meta schemas through the `metaSchema` prop to your `Form` component:
 
 ```jsx
-const metaSchema= require("ajv/lib/refs/json-schema-draft-04.json");
+const metaSchema = require("ajv/lib/refs/json-schema-draft-04.json");
 
 render((
   <Form schema={schema}
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -51,29 +51,32 @@ render((
 
 To have your schemas validated against any other meta schema than draft-07 (the current version of [JSON Schema](http://json-schema.org/)), make sure your schema has a `$schema` attribute that enables the validator to use the correct meta schema. For example:
 
-```json 
+```json
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   ...
 }
 ```
 
-Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas, you need to install `ajv` and import them, then pass your meta schemas through the `metaSchema` prop to your `Form` component:
+Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas pass them through the `additionalMetaSchemas` prop to your `Form` component:
 
 ```jsx
-const metaSchema = require("ajv/lib/refs/json-schema-draft-04.json");
+const additionalMetaSchemas = require("ajv/lib/refs/json-schema-draft-04.json");
 
 render((
-  <Form schema={schema}
-        metaSchema={metaSchema} />
+  <Form schema={schema} 
+        additionalMetaSchemas={additionalMetaSchemas}/>
 ), document.getElementById("app"));
 ```
-`metaSchema` also accepts an array of meta schemas:
+
+In this example `schema` passed as props to `Form` component can be validated against draft-07 (default) and by draft-04 (added), depending on the value of `$schema` attribute.
+
+`additionalMetaSchemas` also accepts an array of meta schemas:
 
 ```jsx
 render((
-  <Form schema={schema}
-        metaSchema={[metaSchema1,metaSchema2]} />
+  <Form schema={schema} 
+        additionalMetaSchemas={[metaSchema1, metaSchema2]} />
 ), document.getElementById("app"));
 ```
 
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -47,6 +47,36 @@ render((
 >   received as second argument.
 > - The `validate()` function is called **after** the JSON schema validation.
 
+### Custom schema validation
+
+To have your schemas validated against any other meta schema than draft-07 (the current version of [JSON Schema](http://json-schema.org/)), make sure your schema has `$schema` field that enables validator to use correct meta schema. For example:
+
+```json 
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  ...
+}
+```
+
+Then pass your meta schemas as `metaSchema` prop to your to your `Form` component:
+
+```jsx
+const metaSchema=YOUR_SCHEMA;
+
+render((
+  <Form schema={schema}
+        metaSchema={metaSchema} />
+), document.getElementById("app"));
+```
+`metaSchema` also accepts array of meta schemas:
+
+```jsx
+render((
+  <Form schema={schema}
+        metaSchema={[metaSchema1,metaSchema2]} />
+), document.getElementById("app"));
+```
+
 ### Custom error messages
 
 Validation error messages are provided by the JSON Schema validation by default. If you need to change these messages or make any other modifications to the errors from the JSON Schema validation, you can define a transform function that receives the list of JSON Schema errors and returns a new list.
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -102,17 +102,19 @@ function ErrorList(props) {
   if (errors.length === 0) {
     return <div />;
   }
+
   return (
     <div>
-      <p />
       <ul className="error-detail bs-callout bs-callout-info">
-        {errors.map((error, index) => {
-          return (
-            <li className="text-danger" key={index}>
-              {error}
-            </li>
-          );
-        })}
+        {errors
+          .filter(elem => !!elem)
+          .map((error, index) => {
+            return (
+              <li className="text-danger" key={index}>
+                {error}
+              </li>
+            );
+          })}
       </ul>
     </div>
   );
--- a/src/validate.js
+++ b/src/validate.js
@@ -182,8 +182,16 @@ export default function validateFormData(
   } catch (err) {
     validationErrors = err;
   }
-
   let errors = transformAjvErrors(ajv.errors);
+  if (validationErrors && validationErrors.message) {
+    errors = [
+      ...errors,
+      {
+        stack: validationErrors.message,
+      },
+    ];
+  }
+
   // Clear errors to prevent persistent errors, see #1104
   ajv.errors = null;
 
--- a/src/validate.js
+++ b/src/validate.js
@@ -70,9 +70,7 @@ function toErrorSchema(errors) {
       // "errors" (see `validate.createErrorHandler`).
       parent.__errors = parent.__errors.concat(message);
     } else {
-      if (message) {
-        parent.__errors = [message];
-      }
+      parent.__errors = [message];
     }
     return errorSchema;
   }, {});
--- a/src/validate.js
+++ b/src/validate.js
@@ -159,7 +159,7 @@ export default function validateFormData(
   transformErrors,
   metaSchema
 ) {
-  //add more schemas to validate against
+  // add more schemas to validate against
   if (!addedMetaSchemas && metaSchema) {
     ajv.addMetaSchema(metaSchema);
     addedMetaSchemas = true;
--- a/src/validate.js
+++ b/src/validate.js
@@ -219,7 +219,6 @@ export default function validateFormData(
   return {
     errors: newErrors,
     errorSchema: newErrorSchema,
-    validationErrors: validationErrors,
   };
 }
 
--- a/src/validate.js
+++ b/src/validate.js
@@ -160,7 +160,12 @@ export default function validateFormData(
   additionalMetaSchemas
 ) {
   // add more schemas to validate against
-  if (!addedMetaSchemas && additionalMetaSchemas) {
+  if (
+    !addedMetaSchemas &&
+    additionalMetaSchemas &&
+    Array.isArray(additionalMetaSchemas) &&
+    additionalMetaSchemas.length > 0
+  ) {
     ajv.addMetaSchema(additionalMetaSchemas);
     addedMetaSchemas = true;
   }
--- a/src/validate.js
+++ b/src/validate.js
@@ -1,25 +1,32 @@
 import toPath from "lodash.topath";
 import Ajv from "ajv";
-const ajv = new Ajv({
-  errorDataPath: "property",
-  allErrors: true,
-  multipleOfPrecision: 8,
-  schemaId: "auto",
-});
-// flag indicating whether we've already added custom schemas
-let addedMetaSchemas = false;
+let ajv = createAjvInstance();
+import { deepEquals } from "./utils";
 
-// add custom formats
-ajv.addFormat(
-  "data-url",
-  /^data:([a-z]+\/[a-z0-9-+.]+)?;(?:name=(.*);)?base64,(.*)$/
-);
-ajv.addFormat(
-  "color",
-  /^(#?([0-9A-Fa-f]{3}){1,2}\b|aqua|black|blue|fuchsia|gray|green|lime|maroon|navy|olive|orange|purple|red|silver|teal|white|yellow|(rgb\(\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*,\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*,\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*\))|(rgb\(\s*(\d?\d%|100%)+\s*,\s*(\d?\d%|100%)+\s*,\s*(\d?\d%|100%)+\s*\)))$/
-);
+// flag indicating whether we've already added custom schemas
+let formerMetaSchema = null;
 
 import { isObject, mergeObjects } from "./utils";
+
+function createAjvInstance() {
+  const ajv = new Ajv({
+    errorDataPath: "property",
+    allErrors: true,
+    multipleOfPrecision: 8,
+    schemaId: "auto",
+  });
+
+  // add custom formats
+  ajv.addFormat(
+    "data-url",
+    /^data:([a-z]+\/[a-z0-9-+.]+)?;(?:name=(.*);)?base64,(.*)$/
+  );
+  ajv.addFormat(
+    "color",
+    /^(#?([0-9A-Fa-f]{3}){1,2}\b|aqua|black|blue|fuchsia|gray|green|lime|maroon|navy|olive|orange|purple|red|silver|teal|white|yellow|(rgb\(\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*,\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*,\s*\b([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\b\s*\))|(rgb\(\s*(\d?\d%|100%)+\s*,\s*(\d?\d%|100%)+\s*,\s*(\d?\d%|100%)+\s*\)))$/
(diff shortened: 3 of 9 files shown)


aerfio:
Such a schema fails to be validated: 
```json
{
  "$ref": "#/definitions/Dataset",
  "$schema": "http://json-schema.org/draft-04/schema#",
  "definitions": {
    "Dataset": {
      "properties": {
        "datasetId": {
          "description": "A user-specified unique ID for this BigQuery dataset.",
          "maxLength": 1024,
          "minLength": 1,
          "pattern": "^[_a-zA-Z0-9]+$",
          "title": "Dataset ID",
          "type": "string"
        },
        "defaultTableExpirationMs": {
          "description": "yadayada",
          "title": "Default table expiration",
          "type": "string"
        },
        "description": {
          "description": "A user-friendly description of the BigQuery dataset.",
          "title": "Description",
          "type": "string"
        },
        "friendlyName": {
          "description": "A descriptive name for the BigQuery dataset.",
          "title": "Friendly name",
          "type": "string"
        },
        "labels": {
          "additionalProperties": { "type": "string" },
          "description": "To organize your project, add arbitrary labels as key/value pairs to the BigQuery dataset. Use labels to indicate different environments, services, teams, and so on.",
          "title": "Labels",
          "type": "object",
          "x-googleProperty": { "type": "LABELS" }
        },
        "location": {
          "default": "US",
          "description": "The geographic location where the BigQuery dataset should reside.",
          "enum": ["US", "EU"],
          "title": "Location",
          "type": "string"
        }
      },
      "required": ["datasetId"],
      "type": "object"
    }
  },
  "form": [
    {
      "key": "datasetId",
      "validationMessage": "Must contain only letters, numbers, or underscores. Max length 1024."
    },
    "location",
    "friendlyName",
    "description",
    "labels"
  ]
}
```
And there are no errors in console and they do not get displayed, as written 
 here:https://github.com/mozilla-services/react-jsonschema-form/blob/4c81b02742270e5225c826025f5e99efdc35ea32/src/validate.js#L161

The error is following 
image

aerfio:
@glasserc @edi9999 Any thought on this?

src/validate.js
```diff
@@ -4,7 +4,11 @@ const ajv = new Ajv({
   errorDataPath: "property",
   allErrors: true,
   multipleOfPrecision: 8,
+  schemaId: "auto",
 });
+//add more schemas to validate against, draft-7 remains default
+ajv.addMetaSchema(require("ajv/lib/refs/json-schema-draft-04.json"));
+ajv.addMetaSchema(require("ajv/lib/refs/json-schema-draft-06.json"));
```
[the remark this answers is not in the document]
aerfio: Then maybe hide them under some kind of flag? Something like
```jsx
<Form enableDraft4 ... />
```
and in `validate.js`: 
```js
if(enableDraft4){
   ajv.addMetaSchema(require("ajv/lib/refs/json-schema-draft-04.json"));
}
```
What do you think?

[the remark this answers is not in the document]
epicfaace: @aerfio Perhaps is there some way to call `addMetaSchema` based on the value for `$schema` in the schema definition itself?

aerfio:
@epicfaace Did you mean something like this? regexp matches both "http://json-schema.org/draft-04/schema#" and "http://json-schema.org/draft-04/schema" ("#" sign at the end), because both of those two adresses work.

@edi9999 what do you think about this?

src/validate.js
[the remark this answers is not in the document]
edi9999: when this is compiled by webpack, even if they are `if` statements, all the JSON files will be in the bundled file.

I think we should rather have a way of doing : 

```
<Form
     jsonSchema={require("ajv/lib/refs/json-schema-draft-04.json"}
</Form>
```

epicfaace:
@aerfio , I think @edi9999 's proposal is that we do **not** include draft-04 or draft-06 within react-jsonschema-form, as it would unnecessarily increase our bundle size. Rather, if someone wants to use draft-04, they must install `ajv` themselves and use the following code:

```
<Form
     jsonSchema={require("ajv/lib/refs/json-schema-draft-04.json"}
</Form>
```

Did that make sense? Would you be able to change your PR to work this way instead?

edi9999:
Looks good to me !!

src/validate.js
```diff
@@ -4,7 +4,11 @@ const ajv = new Ajv({
   errorDataPath: "property",
   allErrors: true,
   multipleOfPrecision: 8,
+  schemaId: "auto",
 });
+//flag indicating whether we've already added custom schemas
+let addedAdditionalSchemas = false;
```
edi9999: Why would we be disallowed from adding schemas multiple times ?

I could see a case where you would have schema-06 for one form, and schema-04 for an other form.

aerfio: @glasserc You can't add the same schema multiple times, ajv throws errors: 
image

If you want to you can add multiple schemas at once, like so `additionalSchema ={[ schema1, schema2 ]}`

And if you want to validate one form using draft-6 and one using draft-4 you'd just use correct `$schema` in your schema, like in example provided in my second comment, read this for clarification https://github.com/epoberezkin/ajv#validateschemaobject-schema---boolean .

> If schema has $schema property, then the schema with this id (that should be previously added) is used to validate passed schema.
also  
>By default this method is called automatically when the schema is added, so you rarely need to use it directly.

@epicfaace They're additional jsonSchemas :P variable name can be changed, no problem

aerfio:
@epicfaace According to http://json-schema.org/latest/json-schema-core.html#rfc.section.6.4

>This extended meta-schema SHOULD be referenced using the "$schema" keyword, to allow tools to follow the correct behaviour.

src/validate.js
edi9999: Ok makes sense !

docs/validation.md
```diff
@@ -47,6 +47,36 @@ render((
...
+
+```jsx
+const metaSchema=YOUR_SCHEMA;
+
+render((
+  <Form schema={schema}
+        metaSchema={metaSchema} />
+), document.getElementById("app"));
+```
+`metaSchema` also accepts array of meta schemas:
```
epicfaace:
```suggestion
`metaSchema` also accepts an array of meta schemas:
```

docs/validation.md
```diff
@@ -47,6 +47,36 @@ render((
...
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  ...
+}
+```
+
+Then pass your meta schemas as `metaSchema` prop to your to your `Form` component:
+
+```jsx
+const metaSchema=YOUR_SCHEMA;
```
epicfaace:
```suggestion
const metaSchema= require("ajv/lib/refs/json-schema-draft-04.json");
```

docs/validation.md
```diff
@@ -47,6 +47,36 @@ render((
 >   received as second argument.
 > - The `validate()` function is called **after** the JSON schema validation.
 
+### Custom schema validation
+
+To have your schemas validated against any other meta schema than draft-07 (the current version of [JSON Schema](http://json-schema.org/)), make sure your schema has `$schema` field that enables validator to use correct meta schema. For example:
```
epicfaace:
```suggestion
To have your schemas validated against any other meta schema than draft-07 (the current version of [JSON Schema](http://json-schema.org/)), make sure your schema has a `$schema` attribute that enables the validator to use the correct meta schema. For example:
```

docs/validation.md
```diff
@@ -47,6 +47,36 @@ render((
...
+
+```json 
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  ...
+}
+```
+
+Then pass your meta schemas as `metaSchema` prop to your to your `Form` component:
```
epicfaace:
```suggestion
Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas, you need to install `ajv` and import them, then pass your meta schemas through the `metaSchema` prop to your `Form` component:
```

src/validate.js
```diff
@@ -4,7 +4,11 @@ const ajv = new Ajv({
   errorDataPath: "property",
   allErrors: true,
   multipleOfPrecision: 8,
+  schemaId: "auto",
 });
+//flag indicating whether we've already added custom schemas
```
epicfaace:
```suggestion
// flag indicating whether we've already added custom schemas
```

src/validate.js
```diff
@@ -152,8 +156,15 @@ export default function validateFormData(
   formData,
   schema,
   customValidate,
-  transformErrors
+  transformErrors,
+  metaSchema
 ) {
+  //add more schemas to validate against
```
epicfaace:
```suggestion
  // add more schemas to validate against
```

src/validate.js
```diff
@@ -152,8 +156,15 @@ export default function validateFormData(
...
+  if (!addedMetaSchemas && metaSchema) {
+    ajv.addMetaSchema(metaSchema);
+    addedMetaSchemas = true;
+  }
+
   try {
     ajv.validate(schema, formData);
   } catch (e) {
```
epicfaace: Can we add a `console.error` here so if the schema is invalid, the user gets an error?

aerfio:
I'll probably do those tests tomorrow, don't merge for now! :D

test/ArrayField_test.js
```diff
@@ -132,7 +132,7 @@ describe("ArrayField", () => {
       const matches = node.querySelectorAll("#custom");
       expect(matches).to.have.length.of(1);
       expect(matches[0].textContent).to.eql(
-        "should NOT have less than 2 items"
+        "should NOT have fewer than 2 items"
```
epicfaace: Do you know why these error messages have changed? Is it maybe because of the ajv upgrade?

aerfio: @epicfaace yeah, they changed after ajv upgrade

docs/validation.md
```diff
@@ -47,6 +47,36 @@ render((
 >   received as second argument.
 > - The `validate()` function is called **after** the JSON schema validation.
 
+### Custom schema validation
+
+To have your schemas validated against any other meta schema than draft-07 (the current version of [JSON Schema](http://json-schema.org/)), make sure your schema has a `$schema` attribute that enables the validator to use the correct meta schema. For example:
```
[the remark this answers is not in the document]
aerfio: that is intentional, look at `(the current`

aerfio:
On one hand I agree with you @epicfaace, but on the other it would be taking out features, that ajv already supports. 

It would be useful in a case when we don't really know which draft will finally use, (like getting data from some kind of backend) so we can preemptively add e.g. draft-4 and draft-6. And prop name `metaSchema` (singular) is intentional - usual case would be that user adds _one_ meta schema (while still allowing to use more than one at a time)

test/validate_test.js
```diff
@@ -675,5 +721,56 @@ describe("Validation", () => {
...
+          liveValidate: true,
+          metaSchema: require("ajv/lib/refs/json-schema-draft-04.json"),
+          onSubmit,
+          onError,
+        });
+        comp = withMetaSchema.comp;
+        node = withMetaSchema.node;
+      });
+
+      it("should be used to validate schema", () => {
```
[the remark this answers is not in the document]
aerfio: Both done 👍 @epicfaace 
All test cases are now added I think

aerfio:
@epicfaace What do you think?

epicfaace:
Yeah, I see your logic here. If I set `metaSchema={require("ajv/lib/refs/json-schema-draft-04.json")}` then this means the form will support both draft-04 and draft-07, right? If so, can you clarify on the documentation that that will be the case?

src/validate.js
```diff
@@ -152,13 +156,20 @@ export default function validateFormData(
...
+    addedMetaSchemas = true;
+  }
+  let validationErrors = null;
   try {
     ajv.validate(schema, formData);
-  } catch (e) {
-    // swallow errors thrown in ajv due to invalid schemas, these
-    // still get displayed
+  } catch (err) {
+    console.error(err);
```
aerfio: only problem with this console.error is that in test in few places we have now 
image

One such error is in my test (intentional), two are in old tests

aerfio:
ok @epicfaace, I've changed the prop name and updated docs

docs/validation.md
```diff
@@ -51,30 +51,33 @@ render((
...
-```json 
+```json
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   ...
 }
 ```
 
-Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas, you need to install `ajv` and import them, then pass your meta schemas through the `metaSchema` prop to your `Form` component:
+Note that react-jsonschema-form only supports the latest version of JSON Schema, draft-07, by default. To support additional meta schemas pass them through the `additionalMetaSchemas` prop to your `Form` component:
```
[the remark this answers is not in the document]
epicfaace: That's great, thanks for the update. Although all of `ajv` is still downloaded when working on it in a development environment, if you don't explicitly import `ajv/lib/refs/json-schema-draft-04.json` then that schema won't be added to the generated bundle -- so @glasserc 's point still holds. :)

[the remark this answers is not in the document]
aerfio: It is added
image

It's also in CRA, even without installing `RJSF`, `ajv` is dependency of `webpack` and it seems that it always installs itself with draft-04 few versions. Those 4 are already installed with RJSF 
image

epicfaace:
@aerfio one thing I didn't think of before when reviewing -- what if the user does the following:

```
  <Form schema={schema} 
        additionalMetaSchemas={metaSchemas}/>
```

And then `metaSchemas` changes from `[schema1, schema2]` to just `[schema1]`?

As of now, the code you have will still have an `ajv` validator which validates both schema1 and schema2 (while what you expect is that the validator only validates schema1). You will need to add logic to remove a meta schema from `ajv`, I think; is there a method to do so?

aerfio:
There is no `removeMetaSchema` function in ajv unfortunately but if you provide schema with proper `$schema` field, (or no `$schema` to use default draft which is draft-07) then your schema will get only validated by that metaSchema, so there is no situation in which adding "too many" metaSchemas would be detrimental (functionality wise ofc)

aerfio:
Aaaaand tests fail due to https://github.com/terser-js/terser/pull/254 😞. But changes have been made @glasserc

glasserc:
Since I keep getting tagged on this PR (:laughing:), I restarted the tests, which now pass.

aerfio:
Whoops! Wrong @ :stuck_out_tongue_winking_eye: 
But thanks for the tests :+1: 
Ofc I meant: @epicfaace

epicfaace:
> There is no `removeMetaSchema` function in ajv unfortunately but if you provide schema with proper `$schema` field, (or no `$schema` to use default draft which is draft-07) then your schema will get only validated by that metaSchema, so there is no situation in which adding "too many" metaSchemas would be detrimental (functionality wise ofc)

That makes sense. The case that I'm thinking of is this:

What if you pass in the prop `metaSchemas=[draft04, draft06]` and want the form _only_ to validate, say, draft-04 and draft-06 -- then, later, you change the prop to `metaSchemas=[draft04]`? One would expect that the form would only validate draft-04 at this case. However, the way the code is structured, it would continue to validate draft-04 and draft-06, which is not the behavior that people may expect.

I feel that this should be fixed -- unless you think this might not be a realistic use case?

Since there is no `removeMetaSchema` function, we can work around that: when meta schemas are removed, just create a new ajv instance and add the remaining meta schemas to it.

aerfio:
@epicfaace I think it's either: add to docs that this prop shouldn't be changed once it's set, or do it as you say with new Ajv instance, but I think that changing `additionalMetaSchemas` would be such an obscure usecase that no one ever would use that :p and in example you have provided -> at least user would get his/her data validated, right now validation just doesn't work, so it's either they make mistake and they get suprised that it still works (and gets validated), or they make mistake and nothing works at all.

epicfaace:
@aerfio , the problem here now is that the tests are all affecting each other though (i.e., when ajv is modified through additionalMetaSchemas in one test, the same modifications carry over to the next test.) Just try it -- set additionalMetaSchemas to [] in the following line: https://github.com/mozilla-services/react-jsonschema-form/pull/1130/files#diff-6654a6c320ad438fbbb53fa54fe3b95dR771 , and the test will still pass!

So I think we will need to make a way to create a new `ajv` instance each time a `Form` is instantiated.

aerfio:
@epicfaace It doesn't pass for me
image
Did you do `rm -rf node_modules dist lib && npm i` before running tests locally? Or am I somehow mistaken?

And as I understand new `ajv` instance is made when `Form` is created, so each time `createFormComponent` is called, so basically in each test, 'cause it's called in `beforeEach` block

aerfio:
By the way, could you please do one final review? We'd love to see this finally released 😀

aerfio:
Hey @epicfaace

pushred:
Please note that I have opened a separate PR based on this, https://github.com/mozilla-services/react-jsonschema-form/pull/1186, that implements the ability to specify custom formats. It's not directly related but I thought I should mention it.

If there's any interest in proceeding with #1186 it may make sense to consolidate such AJV passthrough arguments in the validateFormData function as an object. the signature of that function is getting a bit unwieldy.. sorry if that delays this in any way! :grimacing:

epicfaace:
@aerfio Apologize for the delay.

Here is the test I changed:
https://github.com/mozilla-services/react-jsonschema-form/commit/6d9440672c78669b0f1565fa43f0c394abf7cc88

And here is evidence that the tests are still passing:
https://travis-ci.org/mozilla-services/react-jsonschema-form/builds/496286121?utm_source=github_status&utm_medium=notification

It's strange that you're getting different results on your end, though. Am I missing something?

epicfaace:
> Please note that I have opened a separate PR based on this, #1186, that implements the ability to specify custom formats. It's not directly related but I thought I should mention it.
> 
> If there's any interest in proceeding with #1186 it may make sense to consolidate such AJV passthrough arguments in the validateFormData function as an object. the signature of that function is getting a bit unwieldy.. sorry if that delays this in any way! 😬

@pushred Thanks for mentioning this. I think we can just keep the current code structure for now and make a separate issue/PR for refactoring `validateFormData`; since `validateFormData` is an internal method it shouldn't be a big deal to change its arguments, anyway.

src/validate.js
```diff
@@ -152,13 +156,19 @@ export default function validateFormData(
   formData,
   schema,
   customValidate,
-  transformErrors
+  transformErrors,
+  additionalMetaSchemas
 ) {
+  // add more schemas to validate against
+  if (!addedMetaSchemas && additionalMetaSchemas) {
+    ajv.addMetaSchema(additionalMetaSchemas);
```
epicfaace: Technically, this still allows `additionalMetaSchemas` to not be an array.

aerfio: Yeah, good catch, fix'd 👍

aerfio:
https://travis-ci.org/mozilla-services/react-jsonschema-form/jobs/496498711
https://github.com/mozilla-services/react-jsonschema-form/pull/1130/commits/c0a52c0f656fbc0df4ee52d4ddb4898969aa8cf9

Finally it fails like it should. Ajv instance was not changing between tests, so now I've enforced creation of new instance when `additionalMetaSchemas` changed

aerfio:
@epicfaace I found what was going on and fixed it, now it's predictable

@pushred good, keep going!

epicfaace:
Good, I've added a test that allows additionalMetaSchemas to change and fixed a small bug in which a new ajv instance is not created if additionalMetaSchemas is not an array.

One more thing. Although validateFormData now returns a validationErrors object, validationErrors is not accessible by the end user at all, so it will fail silently. Perhaps we need to add validationErrors to the `<Form>`'s state.

In that case, would we need validationErrors to be another prop that is passed down to all fields (just like errorSchema), as documented here: https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/#field-props I'm worried that having error, errorSchema, and validationErrors might make it more confusing to a developer using this library. I'd appreciate any suggestions for how to do this.

src/components/fields/SchemaField.js
```diff
@@ -102,17 +102,19 @@ function ErrorList(props) {
   if (errors.length === 0) {
     return <div />;
   }
+
   return (
     <div>
-      <p />
```
aerfio: Was it intentional or just a leftover?

aerfio:
@epicfaace OK, I have deleted `validationErrors` and merged it into `errors` and `errorSchema`, so that they're handled by existing components: 
image

Only thing that may matter is that error produced by validation errors doesn't have all the fields that normal error have, it only has `stack` https://github.com/mozilla-services/react-jsonschema-form/pull/1130/files#diff-6ca5d88062570937214ee7ea244b05c5R194, but then again, those incomplete errors only occur in case of failed validation, so normal errors don't even show up, and previously we had *no* errors whatsoever in such a case, so I don't introduce any breaking changes

Unless we always guarantee that errors have specific fields, but I couldn't find anything about it in docs

src/components/fields/SchemaField.js
epicfaace: Not sure, do you know what commit caused this deletion?

epicfaace:
@aerfio I think that what you did works well and makes sense -- nice work!

One thing to keep in mind is that ajv actually throws an error when the $schema given is invalid (rather than adding it to the `errors` object -- see [this fiddle](https://jsfiddle.net/y5x09nu6/)). However, it makes sense on our end to instead add this error to the `errors` object, given that we're making user-facing forms.

src/validate.js
```diff
@@ -152,24 +165,44 @@ export default function validateFormData(
...
+  }
+
+  let validationErrors = null;
   try {
     ajv.validate(schema, formData);
-  } catch (e) {
-    // swallow errors thrown in ajv due to invalid schemas, these
-    // still get displayed
+  } catch (err) {
+    validationErrors = err;
```
epicfaace: @aerfio `validationErrors` is an object, not an array, right? Would it make more sense to rename it to `validationError`?

epicfaace:
@aerfio regarding error documentation, yes, we don't have a documentation for that yet. I'm adding that in [this PR](https://github.com/mozilla-services/react-jsonschema-form/pull/1140/commits/0de5673abaaced7f50545a303e5af61796e3a576) though, so once it's merged we can update the documentation to mention that errors regarding a $schema problem will only have the `stack` attribute.

src/validate.js
```diff
@@ -53,13 +63,16 @@ function toErrorSchema(errors) {
...
     }
+
     if (Array.isArray(parent.__errors)) {
       // We store the list of errors for this node in a property named __errors
       // to avoid name collision with a possible sub schema field named
       // "errors" (see `validate.createErrorHandler`).
       parent.__errors = parent.__errors.concat(message);
     } else {
-      parent.__errors = [message];
+      if (message) {
```
epicfaace: @aerfio when would `message` be undefined?

src/components/fields/SchemaField.js
aerfio: I mean I deleted this `<p />`, but I'm kinda curious why was it here in the first place 😛

src/validate.js
aerfio: If there is no `message` field in particular error, like in that special case I've introduced
If I'd make validation error like 
```
{
   stack: something,
   message: something
}
```
then it would look like this:
image
Which is bad because we have the same info in errorList and in the footer

aerfio: @epicfaace Error thrown by ajv has only name and stack fields, name is `Error`, stack is like in pic above

Advantage of using message field is that while form has validation errors, then all fields are red, opposite to https://github.com/mozilla-services/react-jsonschema-form/pull/1130#issuecomment-467088376

aerfio:
@epicfaace The case you are refering to, when ajv throws errors connected to bad schema are already handled, it's even tested behaviour https://github.com/mozilla-services/react-jsonschema-form/pull/1130#issuecomment-467088376 

And regarding docs - could you please add it in that PR, if you're already writing it? Choose what version you like, with `message` field in error or without it, and I'll change this PR accordingly (or leave it as it is)

epicfaace:
> @epicfaace The case you are refering to, when ajv throws errors connected to bad schema are already handled, it's even tested behaviour https://github.com/mozilla-services/react-jsonschema-form/pull/1130/files#diff-6654a6c320ad438fbbb53fa54fe3b95dR307
> 
> And regarding docs - could you please add it in that PR, if you're already writing it? Choose what version you like, with `message` field in error or without it, and I'll change this PR accordingly (or leave it as it is)

Will do. One more thing -- since we're adding meta schema errors (i.e., "no schema with key or ref \"http://json-schema.org/draft-04/schema#\"") to the `errors` object, we might also need to add these errors to the `errorSchema`. Can you update that?

aerfio:
@epicfaace We are already doing it, because we're merging `{stack: something}` into errors before `toErrorSchema` call: https://github.com/mozilla-services/react-jsonschema-form/pull/1130/files#diff-6ca5d88062570937214ee7ea244b05c5R205.

But those specific errors do not have `message` field so they do not get used in errorSchema, so we're back to this discussion about `message` field.

src/components/fields/SchemaField.js
epicfaace: Probably an accident then

epicfaace:
@aerfio I think it would be better if errorSchema has every error that is in errors. Maybe one way to do it is like this:
Imagine an invalid schema like this:
```
{
  "type": "test"
}
```
The errorSchema constructed looks like this (and then the error does not show up twice):
```
{
  "type": {
    "__errors": [
      "should be equal to one of the allowed values",
      "should be array",
      "should match some schema in anyOf"
    ]
  }
}
```
So if we have a problem with the wrong $schema, can we construct an error list like this?
```
{
  "$schema": {
    "__errors": [
      "no schema with key or ref \"test\""
    ]
  }
}
```

epicfaace:
Also, here's another problem: it seems like an extra error is added for _any_ validation error, not just a schema validation error. 

For example, imagine I have an invalid schema as follows:
```
{
  "type": "test"
}
```

Then, this code adds an extra error ("schema is invalid: data.type should be equal to one of the allowed values, data.type should be array, data.type should match some schema in anyOf") which is redundant.
image

We need to make sure that the try-catch statement is specific enough so that it only adds $schema-specific errors to the errors list.

aerfio:
@epicfaace I have reduced validation errors to only those with error message starting with `no schema with key or ref ` which I think is as specific as it gets.

And I have added 
```
{
  "$schema": {
    "__errors": [
      "no schema with key or ref yada yada"
    ]
  }
}
```
to errorSchema.

epicfaace:
@aerfio everything looks good, thanks!

aerfio:
@epicfaace thanks! when can we expect next release? :grinning:

PR #1133, opened 2019-01-17
Expand anyOf support to also support oneOf

LucianBuzzo:

This is a follow-up to #1118 that expands the code to support `oneOf` as well.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1792,8 +1792,6 @@ The JSON Schema standard says that the dependency is triggered if the property i
 
 In this example the user is prompted with different follow-up questions dynamically based on their answer to the first question.
 
-Note that this is quite far from complete `oneOf` support!
-
 In these examples, the "Do you have any pets?" question is validated against the corresponding property in each schema in the `oneOf` array. If exactly one matches, the rest of that schema is merged with the existing schema.
 
 ## JSON Schema supporting status
@@ -1803,13 +1801,9 @@ This component follows [JSON Schema](http://json-schema.org/documentation.html) 
 * `additionalItems` keyword for arrays
     This keyword works when `items` is an array. `additionalItems: true` is not supported because there's no widget to represent an item of any type. In this case it will be treated as no additional items allowed. `additionalItems` being a valid schema is supported.
 * `anyOf`, `allOf`, and `oneOf`, or multiple `types` (i.e. `"type": ["string", "array"]`
-    The `anyOf` keyword is supported but has the following caveats:
-    - The `anyOf` keyword is not supported when used inside the `items` keyword
-      for arrays.
-    - Properties declared inside the `anyOf` should not overlap with properties
-      "outside" of the `anyOf`.
+    The `anyOf`  and `oneOf` keywords are supported,  however, properties declared inside the `anyOf/oneOf` should not overlap with properties "outside" of the `anyOf/oneOf`.
 
-    You can use `oneOf` with [schema dependencies](#schema-dependencies) to dynamically add schema properties based on input data but this feature does not bring general support for `oneOf` elsewhere in a schema.
+    You can also use `oneOf` with [schema dependencies](#schema-dependencies) to dynamically add schema properties based on input data.
 
 ## Tips and tricks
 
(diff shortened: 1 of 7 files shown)


LucianBuzzo:
@glasserc If you have time a review would be great!

jericmason:
I'm trying to understand how this will work? I saw the previous PR for **anyOf** and it looks similar to what is already supported via schema dependencies. There is currently a problem with schema dependencies in that data will stick around when you change selections, which in combination with **additionalProperties: false** will result in invalid schemas. I feel that is a pretty big bug as the schema form should never generate invalid schemas. That behaviour is more relevant with **anyOf** and seems to work correctly, however I find the UI a bit confusing as **anyOf** should be a multi-select that allows input supporting one or more of the sub-schemas at a given time.

With respect to oneOf, I believe this solution is correct, but I'm curious as to whether it too will generate an invalid schema if you update more than one of the possible options.

jericmason:
@epicfaace I just checked the updated examples and **oneOf** does indeed work as expected. It won't generate invalid json for the schema. It would be nice if the current implementation of schema dependencies used the same **oneOf** functionality (I've added a ticket with examples for [that](https://github.com/mozilla-services/react-jsonschema-form/issues/1148)).

I'm still a little confused about **anyOf**; my understanding is that **anyOf** should support _one or more_ of the provided sub-schemas, but in this implementation it will support the _base schema and exactly one sub-schema_. Maybe I'm missing something here. I really do appreciated these features and feel like they will be powerful additions to this project.

epicfaace:
@jericmason thanks for the ticket, I've added my thoughts to that.

You are exactly right about oneOf and anyOf. Note how an example for the [playground for oneOf](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsb3JlbSI6ImFzZHNhZCJ9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsIm9uZU9mIjpbeyJwcm9wZXJ0aWVzIjp7ImxvcmVtIjp7InR5cGUiOiJzdHJpbmcifX0sInJlcXVpcmVkIjpbImxvcmVtIl19LHsicHJvcGVydGllcyI6eyJsb3JlbSI6eyJ0eXBlIjoic3RyaW5nIn19fV19LCJ1aVNjaGVtYSI6e319) will error when both schemas are the same (because if both schemas are the same, then the form data can't just match *one of* those two schemas), while the [playground for anyOf ](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsb3JlbSI6ImFzZHNhZCJ9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsImFueU9mIjpbeyJwcm9wZXJ0aWVzIjp7ImxvcmVtIjp7InR5cGUiOiJzdHJpbmcifX0sInJlcXVpcmVkIjpbImxvcmVtIl19LHsicHJvcGVydGllcyI6eyJsb3JlbSI6eyJ0eXBlIjoic3RyaW5nIn19fV19LCJ1aVNjaGVtYSI6e319) will not error. Additionally, anyOf can be used with arrays.

alex-pex:
@glasserc Can you release a new version with oneOf support ?

LucianBuzzo:
@alex-pex `oneOf` support has already been released - see https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsb3JlbSI6Imlwc3VtIn0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0Iiwib25lT2YiOlt7InByb3BlcnRpZXMiOnsibG9yZW0iOnsidHlwZSI6InN0cmluZyJ9fSwicmVxdWlyZWQiOlsibG9yZW0iXX0seyJwcm9wZXJ0aWVzIjp7Imlwc3VtIjp7InR5cGUiOiJzdHJpbmcifX0sInJlcXVpcmVkIjpbImlwc3VtIl19XX0sInVpU2NoZW1hIjp7fX0=

Issue #945, opened 2018-06-06
Support oneOf syntax in BooleanField with radio widget.


JamesGuthrie:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Recently support for the oneOf syntax was added for enums, it can be used to replace the existing enum/enumNames syntax which does not conform to the json schema spec.

### Steps to Reproduce

json schema
```
{
  "title": "An example of bad behaviour",
  "type": "object",
  "properties": {
    "test": {
      "type": "boolean",
      "oneOf": [
        {
          "const": true,
          "title": "True"
        },
        {
          "const": false,
          "title": "False"
        }
      ]
    }
  }
}
```
UI schema:
```
{
  "test": {
    "ui:widget": "radio"
  }
}
```

I've set up a [sample JSFiddle](https://jsfiddle.net/f2y3fq7L/306/) demonstrating the issue.

#### Expected behavior

The `title` values in the oneOf objects are used for titles of the radios

#### Actual behavior

The titles of the radios are yes/no

### Version

v1.0.3

Issue #782, opened 2017-12-04
oneOf{const,title} reorders boolean options unlike enum/enumNames


mfulton26:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Bug

### Steps to Reproduce

1. Define a schema with a boolean type property
2. Define an enum using the alternative oneOf{const,title} for that boolean type with `false` before `true` in the array
3. Render the form and note how the options have been reordered. 😢 

Here is a [shared playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJib29sIjp7InR5cGUiOiJib29sZWFuIiwib25lT2YiOlt7ImNvbnN0IjpmYWxzZSwidGl0bGUiOiJObyJ9LHsiY29uc3QiOnRydWUsInRpdGxlIjoiWWVzIn1dfX19LCJ1aVNjaGVtYSI6eyJib29sIjp7InVpOndpZGdldCI6InJhZGlvIn19fQ==) demonstrating the issue.

#### Expected behavior

Order of boolean options can be controlled

#### Actual behavior

Order of boolean options cannot be controled

### Version

You can usually get this information in your `package.json` or in the file URL if you're using the unpkg one.

1.0.0

glasserc:
Both this and #781 seem like legitimate bugs and I would be happy to review a PR to fix them.

pushred:
@glasserc I've opened a PR that fixes these issues as well as #945, please review https://github.com/mozilla-services/react-jsonschema-form/pull/1137

Issue #781, opened 2017-12-04
oneOf{const,title} with boolean type properties causes titles to be displayed in lower case


mfulton26:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Bug

### Steps to Reproduce

1. Define a schema with a boolean type property
2. Define an enum using the alternative oneOf{const,title} for that boolean type
3. Use some uppercase letters in the enum/oneOf titles
4. Render the form and note how the uppercase letters have been converted to lowercase. 😢 

Here is a [shared playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJib29sIjp7InR5cGUiOiJib29sZWFuIiwib25lT2YiOlt7ImNvbnN0IjpmYWxzZSwidGl0bGUiOiJObyJ9LHsiY29uc3QiOnRydWUsInRpdGxlIjoiWWVzIn1dfX19LCJ1aVNjaGVtYSI6eyJib29sIjp7InVpOndpZGdldCI6InJhZGlvIn19fQ==) demonstrating the issue.

#### Expected behavior

Mixed case radio/select options

#### Actual behavior

Lowercase only radio/select options

### Version

You can usually get this information in your `package.json` or in the file URL if you're using the unpkg one.

1.0.0

mfulton26:
Also, it appears that oneOf{const,title} reorders the boolean options unlike enum/enumNames. Should I open a separate issue for such?

PR #1137, opened 2019-01-19
Fix custom boolean option labels using oneOf schemas

pushred:

As previously reported in the following issues specifying boolean options with a oneOf schema does not work as expected.

- #945 - Support oneOf syntax in BooleanField with radio widget.
- #781 - oneOf{const,title} with boolean type properties causes titles to be displayed in lower case
- #782 - https://github.com/mozilla-services/react-jsonschema-form/issues/782

These issues all essentially result from a lack of support for `oneOf` in BooleanWidget altogether. This implements that support with a similar approach to the [`optionsList` utility](https://github.com/mozilla-services/react-jsonschema-form/blob/b068705b32c9409058b826c7a7d88aa090ec8d29/src/utils.js#L372).

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,6 +659,49 @@ This will be rendered as follows:
 </select>
 ```
 
+This also works for radio buttons:
+
+```js
+const schema = {
+  "type": "boolean",
+  "oneOf": [
+    {
+      "const": true,
+      "title": "Yes"
+    },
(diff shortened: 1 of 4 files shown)


epicfaace:
Merged, thanks @pushred !

PR #1144, opened 2019-01-21
Fix uiSchema for additionalProperties

### LINKED ISSUE #1132: Can't apply uiSchema to additionalProperties
### Prerequisites

### Description

With arrays, you can apply a `uiSchema` to the items in the array using `items` and `additionalItems`. However, the same functionality does not seem to exist for objects, for example using `properties` and `additionalProperties`.

An example is, using the "Additional Properties" example on https://mozilla-services.github.io/react-jsonschema-form/, I can't find a way to set help text (or any other `uiSchema` properties) on all additional properties. You can set it only if you know the name of the property, so adding `"assKickCount": {"ui:help": "Help text"}` works in the example but the text disappears if you rename the key. There should be a way to pass the `uiSchema` down to all `additionalProperties`.

epicfaace:

Fixes #1132:
- allows additionalProperties to be specified in uiSchema
- Document additionalProperties
- Add tests for additionalItems/additionalProperties uiSchema configuration
- Breaking change (but not sure if this is really a big deal @glasserc; the previous behavior might have been a bug) -- uiSchema for an additional property will now come from the additionalProperties attribute, not the attribute of the name of the additional property itself.
  - For example, if you previously had an additional property called "property1", you could style it using the "property1" attribute in the uiSchema. With this PR, the only way to style additional properties is through the "additionalProperties" attribute of the uiSchema, regardless of the value of "property1" in the uiSchema.
### Checklist


chirayu11:
Thanks for implementing this feature, it was useful for us :-)

PR #1147, opened 2019-01-21
Fixes #824

### LINKED ISSUE #824: Name should not be required for properties in data-url format

jorgerobles:

See Issue #824, made name not required.

### Checklist

### DIFF
--- a/test/validate_test.js
+++ b/test/validate_test.js
@@ -79,29 +79,29 @@ describe("Validation", () => {
       });
     });
 
-    describe("data-url validation", () => {
-      let errors;
-
+    describe("Data-Url validation", () => {
       const schema = {
         type: "object",
-        required: ["dataUrlWithName", "dataUrlWithoutName"],
         properties: {
           dataUrlWithName: { type: "string", format: "data-url" },
           dataUrlWithoutName: { type: "string", format: "data-url" },
         },
       };
 
-      beforeEach(() => {
+      it("Data-Url with name is accepted", () => {
         const formData = {
           dataUrlWithName: "data:text/plain;name=file1.txt;base64,x=",
+        };
+        const result = validateFormData(formData, schema);
+        expect(result.errors).to.have.length.of(0);
+      });
+
+      it("Data-Url without name is accepted", () => {
+        const formData = {
           dataUrlWithoutName: "data:text/plain;base64,x=",
         };
         const result = validateFormData(formData, schema);
-        errors = result.errors;
-      });
-
-      it("Both should be fine", () => {
-        expect(errors).to.have.length.of(0);
+        expect(result.errors).to.have.length.of(0);
       });
     });
 
--- a/test/validate_test.js
+++ b/test/validate_test.js
@@ -79,6 +79,32 @@ describe("Validation", () => {
       });
     });
 
+    describe("data-url validation", () => {
+      let errors;
+
+      const schema = {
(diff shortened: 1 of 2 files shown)


jorgerobles:
Just a regex tweak.

epicfaace:
Thanks @jorgerobles, good work 😀-- can you add a test as you did in https://github.com/mozilla-services/react-jsonschema-form/pull/967/files?

edi9999:
It would be better indeed to have a regression test.

test/validate_test.js
```diff
@@ -79,6 +79,32 @@ describe("Validation", () => {
...
+      beforeEach(() => {
+        const formData = {
+          dataUrlWithName: "data:text/plain;name=file1.txt;base64,x=",
+          dataUrlWithoutName: "data:text/plain;base64,x=",
+        };
+        const result = validateFormData(formData, schema);
+        errors = result.errors;
+      });
+
+      it("Both should be fine", () => {
```
epicfaace: Thanks for the test -- don't mean to be nitpicky, but it would really be great and cleaner if you could have separate `it` functions for the dataUrlWithName and a test case for dataUrlWithoutName.

glasserc:
Thanks!

PR #1163, opened 2019-02-02
Feature/fix enum empty bug

huhuaaa:
Fixed the bug. The selector have an empty option, when use enum and the default value is 0 or false or ''.

[Please describe them here]

Fixed the bug. The selector have an empty option, when use enum and the default value is 0 or false or ''.

### Checklist


epicfaace:
@huhuaaa Can you explain what issue you are solving? When the default value is 0 or false or "", what do you expect the dropdown to look like?

When the default value is "" or false or "" for me, it looks like this:
image

huhuaaa:
Yes, i use "typeof default === 'undefined'" to check the default value.

epicfaace:
Why would we _not_ want to have an empty option when the default is 0 or false or ''?

huhuaaa:
Yes, when the default is 0 or false or '' , we need an option. But i think an empty option is not the best one.  And there's another bug here，i set enum = [0,1] and enumNames = ['zero', 'one'] , the empty option also showing. This result is not i expected.

[code sample](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJmaXJzdE5hbWUiOmZhbHNlLCJsYXN0TmFtZSI6Ik5vcnJpcyIsImFnZSI6NzUsImJpbyI6IlJvdW5kaG91c2Uga2lja2luZyBhc3NlcyBzaW5jZSAxOTQwIiwicGFzc3dvcmQiOiJub25lZWQiLCJzdGF0dXMiOjB9LCJzY2hlbWEiOnsidGl0bGUiOiJBIHJlZ2lzdHJhdGlvbiBmb3JtIiwiZGVzY3JpcHRpb24iOiJBIHNpbXBsZSBmb3JtIGV4YW1wbGUuIiwidHlwZSI6Im9iamVjdCIsInJlcXVpcmVkIjpbImZpcnN0TmFtZSIsImxhc3ROYW1lIl0sInByb3BlcnRpZXMiOnsic3RhdHVzIjp7InR5cGUiOiJudW1iZXIiLCJ0aXRsZSI6InN0YXR1cyIsImVudW0iOlswLDFdLCJlbnVtTmFtZXMiOlsiemVybyIsIm9uZSJdLCJkZWZhdWx0IjowfX19LCJ1aVNjaGVtYSI6eyJmaXJzdE5hbWUiOnsidWk6YXV0b2ZvY3VzIjp0cnVlLCJ1aTplbXB0eVZhbHVlIjoiIn0sImFnZSI6eyJ1aTp3aWRnZXQiOiJ1cGRvd24iLCJ1aTp0aXRsZSI6IkFnZSBvZiBwZXJzb24iLCJ1aTpkZXNjcmlwdGlvbiI6IihlYXJ0aGlhbiB5ZWFyKSJ9LCJiaW8iOnsidWk6d2lkZ2V0IjoidGV4dGFyZWEifSwicGFzc3dvcmQiOnsidWk6d2lkZ2V0IjoicGFzc3dvcmQiLCJ1aTpoZWxwIjoiSGludDogTWFrZSBpdCBzdHJvbmchIn0sImRhdGUiOnsidWk6d2lkZ2V0IjoiYWx0LWRhdGV0aW1lIn0sInRlbGVwaG9uZSI6eyJ1aTpvcHRpb25zIjp7ImlucHV0VHlwZSI6InRlbCJ9fX19)

image

epicfaace:
Oh, I see, so this is solving #1177

epicfaace:
@huhuaaa Can you add some tests for this?

huhuaaa:
OK

huhuaaa:
I had commited the tests.

test/enum_test.js
```diff
@@ -0,0 +1,132 @@
...
+
+    const { node } = createFormComponent({
+      schema,
+    });
+
+    const options = node.querySelectorAll("option");
+    expect(options[0].innerHTML).eql("0");
+  });
+
+  it("should render a select element and it's first option is 'false', if set the enum and the default value is false.", () => {
```
epicfaace: Can you also add a test in which the default value is '', and then it shouldn't render two empty options?

huhuaaa: Thank you for your suggestions, i have adjusted it.

PR #1169, opened 2019-02-06
Fix multiple bugs related to switching between anyOf/oneOf options

### LINKED ISSUE #1168: Broken behaviour for oneOf inside array
### Prerequisites

### Description

Using `oneOf` inside an array, where one branch is an integer and the other branch is an object is broken. When selecting the second option, the correct fields are not rendered at all.

```
{
  "type": "object",
  "properties": {
    "myArray": {
      "type": "array",
      "items": {
        "oneOf": [
          {
            "type": "integer"
          },
          {
            "type": "object",
            "properties": {
              "elt1": {
                "type": "integer"
              },
              "elt2": {
                "type": "string"
              }
            },
            "required": [
              "elt1",
              "elt2"
            ]
          }
        ]
      }
    }
  }
}
```

[playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJteUFycmF5Ijp7InR5cGUiOiJhcnJheSIsIml0ZW1zIjp7Im9uZU9mIjpbeyJ0eXBlIjoiaW50ZWdlciJ9LHsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsiZWx0MSI6eyJ0eXBlIjoiaW50ZWdlciJ9LCJlbHQyIjp7InR5cGUiOiJzdHJpbmcifX0sInJlcXVpcmVkIjpbImVsdDEiLCJlbHQyIl19XX19fX0sInVpU2NoZW1hIjp7fX0=)

Originally reported by @Rasstrelnikov here https://github.com/mozilla-services/react-jsonschema-form/issues/1141#issuecomment-460835801

LucianBuzzo:
Fixes #1168

- Fixed a bug that would prevent input fields from rendering when
switching between a non-object type option and an object type option
- Fixed a bug where options would incorrectly change when entering
values if a subschema with multiple required fields is used
- Fixed a bug where switching from an object type option to a non-object
type option would result in an input field containing the value [Object object]

Change-type: patch

### Checklist

### DIFF
--- a/src/components/fields/MultiSchemaField.js
+++ b/src/components/fields/MultiSchemaField.js
@@ -69,6 +69,10 @@ class AnyOfField extends Component {
           augmentedSchema = Object.assign({}, option, requiresAnyOf);
         }
 
+        // Remove the "required" field as it's likely that not all fields have
+        // been filled in yet, which will mean that the schema is not valid
+        delete augmentedSchema.required;
+
         if (isValid(augmentedSchema, formData)) {
           return i;
         }
@@ -85,7 +89,14 @@ class AnyOfField extends Component {
     const selectedOption = parseInt(event.target.value, 10);
     const { formData, onChange, options } = this.props;
 
-    if (guessType(formData) === "object") {
+    const newOption = options[selectedOption];
+
+    // If the new option is of type object and the current data is an object,
+    // discard properties added using the old option.
+    if (
+      guessType(formData) === "object" &&
+      (newOption.type === "object" || newOption.properties)
+    ) {
       const newFormData = Object.assign({}, formData);
 
       const optionsToDiscard = options.slice();
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -237,7 +237,7 @@ class ObjectField extends Component {
               errorSchema={errorSchema[name]}
               idSchema={idSchema[name]}
               idPrefix={idPrefix}
-              formData={formData[name]}
+              formData={(formData || {})[name]}
               onKeyChange={this.onKeyChange(name)}
               onChange={this.onPropertyChange(
                 name,
--- a/src/utils.js
+++ b/src/utils.js
@@ -732,7 +732,9 @@ export function toIdSchema(
       field,
       fieldId,
       definitions,
-      formData[name],
+      // Its possible that formData is not an object, this can happen if an
+      // array item has just been added, but not populated with data yet
+      (formData || {})[name],
       idPrefix
     );
   }
--- a/src/utils.js
+++ b/src/utils.js
@@ -732,7 +732,7 @@ export function toIdSchema(
       field,
       fieldId,
       definitions,
-      // Its possible that formData is not an object, this can happen if an
+      // It's possible that formData is not an object, this can happen if an
       // array item has just been added, but not populated with data yet
       (formData || {})[name],
       idPrefix
--- a/src/utils.js
+++ b/src/utils.js
@@ -732,7 +732,7 @@ export function toIdSchema(
       field,
       fieldId,
       definitions,
-      // It's possible that formData is not an object, this can happen if an
+      // It's possible that formData is not an object -- this can happen if an
       // array item has just been added, but not populated with data yet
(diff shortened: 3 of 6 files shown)


src/components/fields/MultiSchemaField.js
```diff
@@ -85,7 +89,14 @@ class AnyOfField extends Component {
...
 
-    if (guessType(formData) === "object") {
+    const newOption = options[selectedOption];
+
+    // If the new option is of type object and the current data is an object,
+    // discard properties added using the old option.
+    if (
+      guessType(formData) === "object" &&
+      (newOption.type === "object" || newOption.properties)
+    ) {
```
epicfaace: I'm having trouble understanding how this piece of code fixes the issue "should empty the form data when switching from an option of type 'object'." It looks like in that case, `newOption` would not be an object, so then the properties added by the old option would not be deleted?

src/components/fields/ObjectField.js
```diff
@@ -237,7 +237,7 @@ class ObjectField extends Component {
               errorSchema={errorSchema[name]}
               idSchema={idSchema[name]}
               idPrefix={idPrefix}
-              formData={formData[name]}
+              formData={(formData || {})[name]}
```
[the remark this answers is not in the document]
LucianBuzzo: `formData` can be null if you have just added an item to an array

src/utils.js
```diff
@@ -727,7 +727,7 @@ export function toIdSchema(
       field,
       fieldId,
       definitions,
-      formData[name],
+      formData,
```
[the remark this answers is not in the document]
LucianBuzzo: This fixes and issue where an idSchema is being created, but `formData` is null. It turns out that the code path that this effects (nested object schemas) isn't tested, so I'm going to improve test coverage here and rework this change.

test/anyOf_test.js
```diff
@@ -530,24 +530,23 @@ describe("anyOf", () => {
   });
 
   describe("Arrays", () => {
-    it("should correctly render form inputs for anyOf inside array items", () => {
```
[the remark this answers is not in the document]
LucianBuzzo: I think this was a mistake

src/components/fields/MultiSchemaField.js
LucianBuzzo: `newOption` is the schema that the form has changed to. When we move into this code path, `formData` will always get set to an object, which is undesirable if the new object is, for example, a string. Previously, if the value of `newOption` was `{ "type": "string" }`, and the `formData` gets set to an object, then the UI would display an input with the text `[Object object]`.

LucianBuzzo:
@epicfaace Can you please re-review this PR when you have a chance?

loctn:
Do not mean to rush you @epicfaace , but excited for this fix to land ^_^.

src/components/fields/MultiSchemaField.js
epicfaace: Makes sense -- thanks!

src/utils.js
```diff
@@ -732,7 +732,9 @@ export function toIdSchema(
       field,
       fieldId,
       definitions,
-      formData[name],
+      // Its possible that formData is not an object, this can happen if an
```
epicfaace:
```suggestion
      // It's possible that formData is not an object, this can happen if an
```

src/utils.js
```diff
@@ -732,7 +732,9 @@ export function toIdSchema(
       field,
       fieldId,
       definitions,
-      formData[name],
+      // It's possible that formData is not an object, this can happen if an
```
epicfaace:
```suggestion
      // It's possible that formData is not an object -- this can happen if an
```

src/utils.js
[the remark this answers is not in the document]
epicfaace: Makes sense -- thanks for the fix! Not sure how your initial change to `formData` fixes it, but it makes sense now with your change to `(formData || {})[name]`.

PR #1170, opened 2019-02-06
Generate code coverage reports using nyc

LucianBuzzo:
Change-type: patch

Having code coverage reports is useful for detecting uncovered code paths.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,7 @@ site
 dist
 lib
 yarn.lock
+
+# Code coverage
+coverage
+.nyc_output
--- a/docs/index.md
+++ b/docs/index.md
@@ -258,6 +258,11 @@ $ npm test
 $ npm run tdd
 ```
 
+#### Code coverage
+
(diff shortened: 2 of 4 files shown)


epicfaace:
Looks good, but I think we should find a way to publish the reports. Could we maybe integrate travis ci and nyc? Or, maybe it's better to use a tool such as [code climate](https://codeclimate.com/oss/) instead so it can easily be published?

epicfaace:
> How much does this slow down the tests ?
> 
> Can you measure time before / after ?
> 
> If it slows down significantly, I would suggest to create a special command in package.json for this (npm test-coverage for example)

Yeah, I think making a special command is the right way to go, as one wouldn't need to create code coverage reports each time one runs the tests, right?

LucianBuzzo:
@edi9999 @epicfaace I ran some tests locally and saw, on average, a time increase of ~5s. Whilst this is not a huge increase it is proportionally significant when you consider that the average test run without coverage is only 10s. I'll add a separate command for running the coverage report. 
Code climate looks interesting, how about we open a new issue about integrating it into this project?

epicfaace:
Looks good; let's do that.

PR #1172, opened 2019-02-06
Return original submit event as second param

travisdahl:
See https://github.com/mozilla-services/react-jsonschema-form/issues/1167

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -151,6 +151,7 @@ export default class Form extends Component {
 
   onSubmit = event => {
     event.preventDefault();
+    event.persist();
 
     if (!this.props.noValidate) {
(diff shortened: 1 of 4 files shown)


src/components/Form.js
```diff
@@ -151,6 +151,7 @@ export default class Form extends Component {
 
   onSubmit = event => {
     event.preventDefault();
+    event.persist();
```
epicfaace: Would we need to ever access the event asynchronously? Do you have a sample use case for that?

travisdahl: If you are going to pass the original event past this function you need to persist or you get react warnings.

epicfaace:
Thanks!

PR #1183, opened 2019-02-15
Improve handling of decimal points and trailing zeroes in numbers

### LINKED ISSUE #674: Decimal numbers ending in zero

### LINKED ISSUE #958: Annoying live validation with numeric prefixes

### LINKED ISSUE #966: Default widget for number fields

LucianBuzzo:
Fixes #674, fixes #958, fixes #966 

Change-type: patch

### Checklist

### DIFF
--- a/src/components/fields/NumberField.js
+++ b/src/components/fields/NumberField.js
@@ -3,14 +3,89 @@ import React from "react";
 import * as types from "../../types";
 import { asNumber } from "../../utils";
 
-function NumberField(props) {
-  const { StringField } = props.registry.fields;
-  return (
-    <StringField
-      {...props}
-      onChange={value => props.onChange(asNumber(value))}
-    />
-  );
+// Matches a string that ends in a . character, optionally followed by a sequence of
+// digits followed by any number of 0 characters up until the end of the line.
+// Ensuring that there is at least one prefixed character is important so that
+// you don't incorrectly match against "0".
+const trailingCharMatcherWithPrefix = /\.([0-9]*0)*$/;
+
+// This is used for trimming the trailing 0 and . characters without affecting
+// the rest of the string. Its possible to use one RegEx with groups for this
+// functionality, but it is fairly complex compared to simply defining two
+// different matchers.
+const trailingCharMatcher = /[0.]0*$/;
+
+/**
+ * The NumberField class has some special handling for dealing with trailing
+ * decimal points and/or zeroes. This logic is designed to allow trailing values
+ * to be visible in the input element, but not be represented in the
+ * corresponding form data.
+ *
+ * The algorithm is as follows:
+ *
+ * 1. When the input value changes the value is cached in the component state
+ *
+ * 2. The value is then normalized, removing trailing decimal points and zeros,
+ *    then passed to the "onChange" callback
+ *
+ * 3. When the component is rendered, the formData value is checked against the
+ *    value cached in the state. If it matches the cached value, the cached
+ *    value is passed to the input instead of the formData value
+ */
+class NumberField extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      lastValue: props.value,
+    };
+  }
+
+  handleChange = value => {
+    // Cache the original value in component state
+    this.setState({ lastValue: value });
+
+    // Normalize decimals that don't start with a zero character in advance so
+    // that the rest of the normalization logic is simpler
+    if (`${value}`.charAt(0) === ".") {
+      value = `0${value}`;
+    }
+
+    // Check that the value is a string (this can happen if the widget used is a
+    // <select>, due to an enum declaration etc) then, if the value ends in a
+    // trailing decimal point or multiple zeroes, strip the trailing values
+    let processed =
+      typeof value === "string" && value.match(trailingCharMatcherWithPrefix)
+        ? asNumber(value.replace(trailingCharMatcher, ""))
+        : asNumber(value);
+
+    this.props.onChange(processed);
+  };
+
+  render() {
+    const { StringField } = this.props.registry.fields;
+    const { formData, ...props } = this.props;
+    const { lastValue } = this.state;
+
+    let value = formData;
+
+    if (typeof lastValue === "string" && value) {
+      // Construct a regular expression that checks for a string that consists
+      // of the formData value suffixed with zero or one '.' characters and zero
+      // or more '0' characters
+      const re = new RegExp(`${value}`.replace(".", "\\.") + "\\.?0*$");
+
+      // If the cached "lastValue" is a match, use that instead of the formData
+      // value to prevent the input value from changing in the UI
+      if (lastValue.match(re)) {
+        value = lastValue;
+      }
+    }
+
+    return (
+      <StringField {...props} formData={value} onChange={this.handleChange} />
+    );
+  }
 }
 
 if (process.env.NODE_ENV !== "production") {
--- a/src/components/widgets/BaseInput.js
+++ b/src/components/widgets/BaseInput.js
@@ -23,7 +23,32 @@ function BaseInput(props) {
     ...inputProps
   } = props;
 
-  inputProps.type = options.inputType || inputProps.type || "text";
+  // If options.inputType is set use that as the input type
+  if (options.inputType) {
+    inputProps.type = options.inputType;
+  } else if (!inputProps.type) {
+    // If the schema is of type number or integer, set the input type to number
+    if (schema.type === "number") {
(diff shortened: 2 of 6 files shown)


src/components/fields/NumberField.js
```diff
@@ -3,14 +3,71 @@ import React from "react";
...
-  const { StringField } = props.registry.fields;
-  return (
-    <StringField
-      {...props}
-      onChange={value => props.onChange(asNumber(value))}
-    />
-  );
+/**
+ * The NumberField class has some special handling for dealing with trailing
+ * decimal points and/or zeroes. This logic is desgined to allow trailing values
```
epicfaace:
```suggestion
 * decimal points and/or zeroes. This logic is designed to allow trailing values
```

src/components/fields/NumberField.js
```diff
@@ -3,14 +3,71 @@ import React from "react";
...
+      lastValue: props.value,
+    };
+  }
+
+  handleChange = value => {
+    // Cache the original value in component state
+    this.setState({ lastValue: value });
+
+    // Check that the value is a string (if the widget used is a select (due to
+    // an enum declaration etc) and if the value ends in a trail decimal point
```
epicfaace:
```suggestion
    // an enum declaration etc), and if the value ends in a trailing decimal point
```

src/components/fields/NumberField.js
```diff
@@ -3,14 +3,71 @@ import React from "react";
...
+  render() {
+    const { StringField } = this.props.registry.fields;
+    const { formData, ...props } = this.props;
+    const { lastValue } = this.state;
+
+    let value = formData;
+
+    if (typeof lastValue === "string" && value) {
+      // Construct a regular expression that checks for a string that consists
+      // of the formdata value suffixed with zero or one '.' characters and zero
```
epicfaace:
```suggestion
      // of the formData value suffixed with zero or one '.' characters and zero
```

test/NumberField_test.js
```diff
@@ -129,18 +129,86 @@ describe("NumberField", () => {
...
+          output: 2.003,
+        },
+        {
+          input: "2.00300",
+          output: 2.003,
+        },
+      ];
+
+      tests.forEach(test => {
+        it(`Should work with an input value of ${test.input}`, () => {
```
epicfaace:
```suggestion
        it(`should work with an input value of ${test.input}`, () => {
```

test/NumberField_test.js
```diff
@@ -129,18 +129,86 @@ describe("NumberField", () => {
       expect(node.querySelector(".field input").value).eql("2");
     });
 
-    it("should not cast the input as a number if it ends with a dot", () => {
+    describe("when inputting a number thats ends with a dot and/or zero it should normalize it, without changing the input value", () => {
```
epicfaace:
```suggestion
    describe("when inputting a number that ends with a dot and/or zero it should normalize it, without changing the input value", () => {
```

src/components/fields/NumberField.js
```diff
@@ -3,14 +3,71 @@ import React from "react";
...
+    this.state = {
+      lastValue: props.value,
+    };
+  }
+
+  handleChange = value => {
+    // Cache the original value in component state
+    this.setState({ lastValue: value });
+
+    // Check that the value is a string (if the widget used is a select (due to
```
epicfaace: @LucianBuzzo can you fix the parentheses here?

Also, what is the scenario, exactly, in which there's a select with type=number with an option that has a trailing decimal point or multiple zeroes? Won't the schema never have these? Maybe if you can add a test for this it would be clearer.

src/components/widgets/BaseInput.js
```diff
@@ -44,7 +69,6 @@ function BaseInput(props) {
 }
 
 BaseInput.defaultProps = {
-  type: "text",
```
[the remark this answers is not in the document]
LucianBuzzo: @epicfaace With the type prop set to "text" by default there is no way of knowing wether the prop is there intentionally or not. Removing the default type allows the type prop to inferred correctly in the render step.

src/components/fields/NumberField.js
LucianBuzzo: @epicfaace This logic needs to be here, because as you type in a value such as `0.001` each keystroke will trigger a change event, so the input will have to handle inputs like `0.00` as the user is typing. Additionally, you may have someone copy and paste values like `201.00` into the form, which is a valid number that you would expect to be processed correctly.

LucianBuzzo:
@epicfaace I've addressed you're review comments. If it's alright with you I'd like to merge this PR and then review the changes to `asNumber` independently.

src/components/fields/NumberField.js
```diff
@@ -3,14 +3,71 @@ import React from "react";
...
+    };
+  }
+
+  handleChange = value => {
+    // Cache the original value in component state
+    this.setState({ lastValue: value });
+
+    // Check that the value is a string (this can happen if the widget used is a
+    // <select>, due to an enum declaration etc) then, if the value ends in a
+    // trail decimal point or multiple zeroes, strip the trailing values
```
epicfaace:
```suggestion
    // trailing decimal point or multiple zeroes, strip the trailing values
```

LucianBuzzo:
@epicfaace I couldn't reproduce the issue with pasting the value "201.00", could you post the schema you used to test it? 
The issue with "0.001" is repeatable and a bit of a problem, I'll add some tests and fix it.

epicfaace:
This also fixes #966

epicfaace:
@LucianBuzzo I can't seem to reproduce the "201.00" issue, so I probably was using the wrong version earlier.

Other issues:
- I can't type in ".00"
- If I type in ".01", the number input only ends up with a "1"
- I can't paste in ".00" or "0.00"

LucianBuzzo:
@epicfaace What do you think of automatically prefix `.` with a `0` so that if `.00` is added it automatically becomes `0.00`?

epicfaace:
Looks good! Will merge.

Issue #754, opened 2017-11-07
Schema dependencies conditional. formdata is not deleted if dependent field data is deleted


asifmom:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

Schema dependencies conditional. formdata is not deleted if dependent field data is deleted

### Steps to Reproduce
Take example of sample json provided in documentation
```
{
  "type": "object",

  "properties": {
    "name": { "type": "string" },
    "credit_card": { "type": "number" }
  },

  "required": ["name"],

  "dependencies": {
    "credit_card": {
      "properties": {
        "billing_address": { "type": "string" }
      },
      "required": ["billing_address"]
    }
  }
}
````
#### Expected behavior

Lets say you enter following data:
1) Name :"abc"
2) credit_card: 123

At this point,  billing_address field is visible (which is expected).  Now enter some value in billing address field: example abc

The formData is . 
```
{
  "name": "abc",
  "credit_card": 123,
  "billing_address": "abc"
}
```
Now delete value is credit_card field. billing_address field disappears (expected)

[What you expected to happen]
The formData should be 
```
{
  "name": "Joe",
}
```
#### Actual behavior

However it is 
```
{
  "name": "Joe",
  "billing_address": "abc"
}
```
### Version

1.0.0

lauriii:
There are use cases where it is also useful to have the data stored. Let's say if a user changes a form value by a mistake, it could lead user losing data from the form. I think this is something that should be addressed in the submission phase instead of in the internal data storage. Maybe the behaviour could be configurable?

iplus26:
I agree with @lauriii 
I am considering whether the validation of not-dependent fields can be ignored when submission, instead of modifying user data.

jimmy-guo:
Seems like the billing_address field no longer shows up after entering a credit card number in the schema dependencies example?

https://mozilla-services.github.io/react-jsonschema-form/ -> schema dependencies tab

tbutman:
Did anyone find a workaround for this issue when using schema dependencies? Is there a way to determine which form fields are visible during the time of submission in the onSubmit handler, so that we can filter out formData for fields that are not visible. Currently I don't see any way of customizing that behavior to not have all formData (whether visible or not) submitted.

[...]

PR #1190, opened 2019-02-23
Omit Unused Data Prop

travisdahl:
As mentioned in https://github.com/mozilla-services/react-jsonschema-form/issues/754 , 

We have a working solution for only returning the data actually used in the form. In addition to the problems stated in the reported issue, we also dont want passthrough unused data, even if we provided it originally. We just want to get the data from whatever is in the visible form. As stated, there is no super straight fwd solution, but we think we found a pretty reasonable and not too intrusive solution.

The idea is....

• Add a `name` attribute to all of the fields that is the json path to the form data, e.g. `foo.bar.1.name` (something that IMO should be there regardless of this solution).

* Checkboxes no longer have randomly generated names

• `onSubmit` we can get all of the form elements and make a list of name attributes (the paths).

• use lodash `_.pick(formData, [ARRAY_OF_FIELD_NAMES])` to get only the used data and return that as the formData.

We do this if they set the `omitUnusedData` prop to `true` otherwise it defaults to `false` and works as is.

My coworker @codegains helped me out a bit on this too.

==================

A couple things I wish were better... 

- ~~I wish there was a more reliable way to generate the paths rather than doing string manipulation on the `id`.  But it was hard to access all the necessary data at the point of adding the `name` attribute.~~

- ~~The `idToPath` could fail if the consumer provides a `idPrefix` prop with an underscore .~~ 

**@codegains update addresses both of the above. no longer an issue**

The bonus of having these paths as the name attributes on the fields is that in the future you could potentially simplify your submit and changes with something like https://github.com/maxatwork/form2js (this is super old but basically it gathers formData based on the name attributes more like traditional html forms).


src/components/Form.js
```diff
@@ -149,12 +150,30 @@ export default class Form extends Component {
...
+    return _pick(this.state.formData, inputNames);
+  };
+
   onSubmit = event => {
     event.preventDefault();
     event.persist();
+    let newFormData = this.state.formData;
+
+    if (this.props.omitUnusedData) {
+      newFormData = this.getUsedFormData(event, this.state.formData);
```
epicfaace: @travisdahl Can you explain why you decided to call `getUsedFormData` only on `onSubmit` and not during `onChange`?

src/components/Form.js
```diff
@@ -149,12 +150,30 @@ export default class Form extends Component {
...
+
+    return _pick(this.state.formData, inputNames);
+  };
+
   onSubmit = event => {
     event.preventDefault();
     event.persist();
+    let newFormData = this.state.formData;
+
+    if (this.props.omitUnusedData) {
```
epicfaace:
```suggestion
    if (this.props.omitUnusedData === true) {
```

src/components/widgets/CheckboxesWidget.js
```diff
@@ -29,6 +30,7 @@ function CheckboxesWidget(props) {
             <input
               type="checkbox"
               id={`${id}_${index}`}
+              name={idToPath(`${id}.${index}`)}
```
epicfaace: Did you mean to put an underscore here?

```suggestion
              name={idToPath(`${id}_${index}`)}
```

src/components/widgets/RadioWidget.js
```diff
@@ -32,7 +31,7 @@ function RadioWidget(props) {
             <input
               type="radio"
               checked={checked}
-              name={name}
+              name={idToPath(id)}
```
epicfaace: @travisdahl this is a backwards-incompatible change; why do you think `idToPath(id)` is more appropriate in this case rather than the previously randomly-generated field name?

docs/index.md
```diff
@@ -118,6 +118,8 @@ render((
 ), document.getElementById("app"));
 ```
 
+> Note: If there are fields in the `formData` that are not represented in the schema, they will be retained by default. If you would like to omit those unused values on form submission, set the `omitUnusedData` prop to `true`.
+
```
epicfaace: @travisdahl can you document the `name` prop for all the widgets/fields somewhere in the docs?

src/components/Form.js
CodeGains: Hello! I worked on this with @travisdahl so I think I can answer. We decided to only add it to onSubmit because if we omitted the data in onChange then you would lose the functionality of hidden/dependent fields being populated with data when being shown.

src/components/widgets/CheckboxesWidget.js
CodeGains: The '.' is the correct format for the path in order to use lodash's pick() method.

src/components/widgets/RadioWidget.js
CodeGains: Is this really not backwards compatible? I don't think anyone can rely on using the random generated name in their code in some way. We we're thinking that the name only needed to be unique. It should still be unique in this case because the paths are unique for the form. It also seemed to make more sense to have the name be related to the form/field rather than a random name, more in line with a traditional html form.

CodeGains:
About the name `omitUnusedData` we were definitely mulling over a lot of names and this was the best we could come up with. Some other ideas were `omitUnownedData` or `omitNonFormData`. We didn't want to name it something like `omitNonSchemaData` because some data could be tied to some field in the schema, but not rendered because of a dependent not being met. Definitely open to suggestions though. Thanks!

src/components/widgets/CheckboxesWidget.js
epicfaace: Right -- but `idToPath` takes in a path with underscores in it, then outputs a path with "."

CodeGains: Ohh I see what you are saying. Ya I can change that for consistency, however it would get converted right back to a '.'

src/components/Form.js
```diff
@@ -149,12 +150,30 @@ export default class Form extends Component {
     }
   };
 
+  getUsedFormData = event => {
+    const fields = event.target.elements;
```
epicfaace: @travisdahl @CodeGains It seems like `event.target.elements` will only return [the following elements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/elements): `<button> <fieldset> <input> <object> <output> <select> <textarea>`.

How would we include custom widgets that don't render any of these tags?

I think that, instead of this approach, we may have to go with using ajv's [`removeAdditional`](https://github.com/epoberezkin/ajv#filtering-data) method.

src/components/widgets/CheckboxesWidget.js
epicfaace: Yes, please do!

src/components/widgets/RadioWidget.js
epicfaace: That makes sense; this is definitely an improvement over the randomly-generated name.

My only concern is that now the name doesn't reflect the data path anymore. Let's say I select "foo" and "fuzz" in the playground:
image

Then the checkboxes with names "multipleChoicesList.0" and "multipleChoicesList.2" are checked.

Contrast this with if I didn't use a checkbox widget:
image

In this case, the `<select>` elements with names "multipleChoicesList.0" and "multipleChoicesList.1" are the ones having the values.

I would think that the latter is more natural, as, really, the value of `multipleChoicesList` is an array with two elements, and having names "multipleChoicesList.0" and "multipleChoicesList.1" would reflect that.

But maybe I'm just worried about a non-issue -- do you think the concern I raised might not really practically matter?

epicfaace:
> About the name `omitUnusedData` we were definitely mulling over a lot of names and this was the best we could come up with. Some other ideas were `omitUnownedData` or `omitNonFormData`. We didn't want to name it something like `omitNonSchemaData` because some data could be tied to some field in the schema, but not rendered because of a dependent not being met. Definitely open to suggestions though. Thanks!

Some more ideas: `omitDisallowedData` or `omitExtraData`

src/components/Form.js
CodeGains: I'll have to look more into the custom widgets. What kind of a widget would not have any of these elements? 

Perhaps we also need to add an additional prop for `name` for the custom widgets. The current custom widgets props for reference: https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/#custom-widgets-and-fields

src/components/widgets/RadioWidget.js
CodeGains: Good observation. You are correct that it doesn't reflect the data path anymore, but instead it really represents its location on the form. In practicality, it does not matter for the result of `omitUnusedData` because for the arrays fields, only the selected/present values will be in the formData and all of the array paths are present in the `event.target.elements`, so all of the array values will be picked.

src/components/Form.js
epicfaace: Good point about the `name` prop.

Regarding a custom widget without any of those elements off the top of my head, but one could imagine, say, using a [meter](https://www.w3schools.com/tags/tag_meter.asp) element that could be controlled by buttons.

Or, maybe one creates a custom widget for money with a div that contains two textboxes -- one which controls the number of dollars and the other which controls the number of cents -- and the "div" element itself will report out the value.

Basically, the way a custom widget is designed, we currently don't impose the restriction that elements `<button> <fieldset> <input> <object> <output> <select> <textarea>` have to report out the value; rather, we let it be flexible by using the `onChange` prop. Ideally, we would keep it that way.

src/components/Form.js
travisdahl: Yeah @epicfaace , we went back and forth about this, on whether to always omit it or only on submit.  In the issue thread people were arguing they dont want to lose data while showing, hiding dependent fields which is understandable, so we thought that when you "submit" a form that is a more final action, then you have applied your changes, saved, etc. But now that im saying it aloud, I guess it really depends on how you use the component. We do some stuff on change but dont save to our redux store until submit. Other may only use `onChange` that would send data to the store on each change. We can change it to be `onChange` instead but then you will lose data while hiding showing fields on dependencies. Not to sure what the right expectation would be, would love to hear your opinion.

src/components/Form.js
travisdahl: its defaulted to `false`  in the default props and prop type is set to boolean, so we thought this was reasonable, but can update that no problem.

src/components/Form.js
travisdahl: I see your theoretical point, BUT i would say that is bad practice to not set some kind of input value, even if its a hidden input.  No matter what your visual UI is, if its a form element, it should set a value somewhere, for example if i made any of the custom widgets you suggested..

• a meter with buttons. On up and down button clicks, maybe a graphical ui shows a meter at 50%, but I would set a hidden input value to `50` and that input would be tracked. 

• in the money example, same thing, your dollar and cents can be 2 fields without an onchange but they should be setting a caoncatted value in a hidden input that is being tracked


Thats said, I have an idea on how to support all elements like you are asking, but they would still need to provide a `name` attribute to that element. or maybe we use something like `data-path`. I just like the name attribute because its a more standardized, traditional pattern on how to use forms.

epicfaace: @travisdahl here's a more realistic example: using a `<div contenteditable>`

https://stackoverflow.com/questions/22677931/react-js-onchange-event-for-contenteditable

travisdahl: ah yes, I see your point. That is definitely a valid use case.

src/components/Form.js
epicfaace: I'd like for it to happen on `onChange` if possible. I don't quite get what you mean when you say "you will lose data while hiding showing fields on dependencies" -- could you give an example of this situation?

src/components/Form.js
epicfaace: I think ajv's [`removeAdditional`](https://github.com/epoberezkin/ajv#filtering-data) method would work well to overcome this problem though?

src/components/Form.js
travisdahl: lets use the example in the docs...

If I enter a credit card number, then the billing address field would be shown. If I then entered an address, then cleared my credit card number. The address field would be hidden (and data discarded). The if I entered a new credit card number, I would have to re-enter my address. Which on one hand, is what i would expect if I set omit extra data to true BUT its kind of a bummer for the user who has to re-enter the address again.

src/components/Form.js
CodeGains: If we used `removeAdditional` wouldn't we need the schema with the dependent fields removed if they weren't being used?

epicfaace: @CodeGains can you elaborate on what you mean here?

src/components/Form.js
epicfaace: How about this?
- if `liveValidate` is true, then we perform the extra data omission during `onChange`
- if `liveValidate` is false, then we perform the extra data omission during `onSubmit`

CodeGains: That sounds good, I was just about to push my onChange changes, so I'll modify it to work like this.

src/components/Form.js
CodeGains: Unless I am misunderstanding how `removeAdditional` works, you would need the schema in order to see what fields in the formData are 'additional'. But in the case like #754, `billing_address` still exists when `credit_card` is removed. However, `billing_address` will always be in the schema, so `removeAdditional` will not remove `billing_address` from the formData.

jb-san:
any word on this MR ? would be super useful..

CodeGains:
@epicfaace Is there anything else in the PR that you think we need to address?

src/components/Form.js
```diff
@@ -123,19 +124,39 @@ export default class Form extends Component {
     return null;
   }
 
+  getUsedFormData = formData => {
+    const fields = this.formElement.querySelectorAll("[name]");
+    const inputNames = Object.keys(fields).reduce((filtered, i) => {
```
epicfaace: I think it would be simpler to replace this with 
```
fields.map(e => e.name).filter(e => e.name)
```

epicfaace: You could actually just put

```
fields.map(e => e.name)
```

Because the lodash pick function doesn't mind if you pass in null/undefined arguments in the array.

CodeGains: I think i can simplify, but unfortunately `querySelectorAll` returns a `NodeList` instead of an array so I can't use map.

epicfaace:
I'm still kind of uncomfortable about requiring every custom widget to have a `name` attribute. For example, if I were to create a `<div contenteditable>` widget, then it would have to have a `name` attribute, which would be non-standard.

If that's the only way to go about this, then I understand -- but might not there be another way to do this that is agnostic to the DOM elements generated themselves?

travisdahl:
@epicfaace we could potentially use a `data-` attribute instead. Something like `data-name`.  I like `name` better for form inputs but you are correct in saying it would be non-standard on a div element.  

Another idea is to support both, so when we check the form for inputs we could look for `name` attributes OR the custom `data-name` attribute. I think that would solve your concerns. But I dont think there is any way around having to set an attribute of some kind if they want to use the `omit` prop

epicfaace:
Yeah, good suggestions. I think using something like `removeAdditional` might be a feasible solution that is independent of the DOM, as long as it works with dependencies -- if https://github.com/epoberezkin/ajv/issues/970 is fixed. Is my thinking right here -- or am I missing another problem with this approach?

CodeGains:
@epicfaace Well epoberezkin/ajv#970 is most likely the biggest problem with that approach, so I'm interested to see what they say, but I'm somewhat not optimistic that they will support that behavior. If you see this section of the docs https://github.com/epoberezkin/ajv#filtering-data, they mention that the standard is that everything not included in the `properties` is considered considered an additional property. Actual quote:

> With the option removeAdditional: true the validation will pass for the object { "foo": "abc"} but will fail for the object {"bar": 1}. It happens because while the first subschema in oneOf is validated, the property bar is removed because it is an additional property according to the standard (because it is not included in properties keyword in the same schema).

However, I do agree with you that it should work like you are suggesting because of the fact that the `validate` returns `false` when there is a dependency and `"additionalProperties": false` is in the schema. That seems like an oversight in either the standard or ajv to me.

Another issue I foresee would be the the fact that we would have to add the `additionalProperties: false` config to all the sub-schemas programmatically for nested objects before filtering the data. Maybe it's not a big deal, but it could be messy. I've tried using `removeAdditional: "all"` in this fiddle: https://jsfiddle.net/tjaeger/aLd7zn6o/3/, but it doesn't seem to work at all. In epoberezkin/ajv#634, he says that adding `additionalProperties: false` to all objects and using `removeAdditional: true` is the correct approach.

CodeGains:
Actually, basically the same problem as epoberezkin/ajv#970 happens with `anyOf`, `oneOf`, `allOf`, etc.  as well. See: https://jsfiddle.net/tjaeger/c98axmd0/24/, which they deemed to be working correctly. So, I wouldn't know how to get around this if we were to use `removeAdditional`.

epicfaace:
@CodeGains I see, that makes sense, thanks for the detailed explanation.

What if we do this -- instead of querying for elements that are shown with a `name` attribute, we use the `idSchema` itself and see which keys are present there? That way our solution is not dependent on the DOM.

CodeGains:
Hmm that might work, I'll try it out and see.

CodeGains:
@epicfaace the problem I'm trying to overcome with the `idSchema` approach is that the array indexes are not in the `idSchema` object.

epicfaace:
@CodeGains I don't think that should pose a problem. I wouldn't imagine needing to omit specific indexes using omitExtraData, although please correct me if I'm wrong.

Your idSchema could, for example, just look like this:
```
{
    "$id": "root",
    "tasks": {
        "$id": "root_tasks"
   }
}
```

And if your formData looks like this:
```
{
     "tasks": ["a", "b", "c"]
}
```

Then doing `pick(formData, ["tasks"])` will still get you the data you need.

CodeGains:
@epicfaace That's not really the problem I was thinking about. If the array was an array of objects and some object had extra values then if you picked the whole array, you would get those extra values.

epicfaace:
@CodeGains you're right, here's an example: https://jsfiddle.net/epicfaace/dj37hvge/

It doesn't even seem that idSchema is generated correctly; I'd expect it to contain `root_b` once "b" shows up.

CodeGains:
@epicfaace The `idSchema` is generated the correctly in the playground. The problem I'm seeing now is that the `Form` state update for idSchema happens after `onChange`. However, it looks like I can generate the correct schema using `getStateFromProps` and passing in the `formData` from `onChange`.

But the part that makes it complicated is when there are nested arrays. I will have to account for the indexes not being listed in the schema. I'll keep working on it when I get the chance.

CodeGains:
@epicfaace Wasn't able to work on this for a while, but now I think I'm pretty close to having a solution that doesn't depend on the DOM. Just to let you know not to close this yet. Thanks!

epicfaace:
@codegains Perfect, thank you so much. Look forward to seeing it!

ketysek:
Yaay, I'm looking forward to see this merged in master! ;)

ketysek:
Yaay, I'm looking forward to see it live! Does it solve the issue replicable also in Schema dependencies on https://mozilla-services.github.io/react-jsonschema-form/ that:

1. You **change** the **Conditional/Do you have any pets?*** field to **Yes: One** and **fill** the value in **How old is your pet?***
2. You **change** the **Conditional/Do you have any pets?*** field to **Yes: More than one** and **select** some option for **Do you want to get rid of any?***
3. Now you can see that the formData contains values for both fields: **How old is your pet?*** and also **Do you want to get rid of any?***

CodeGains:
Yes, it should fix that. Sorry for the delay, now I'm having to create my own schema object similar to idSchema and refactor some stuff, because my solution was not working for dependencies within array objects. The schema object I am creating for the final solution should have contain arrays with each object in the array listing which fields are in use.

CodeGains:
Need to rip out old changes, update docs, update tests. May just make another PR instead because it might be easier. 

However, the current solution should be working in this branch if anyone wants to test it out. Set the `defaultProp` for `omitExtraData` to `true` in `Form.js`.

ValYouW:
@CodeGains Hi, I've tried the branch, it seems to have issues with arrays where the items are `$ref`
meaning something like:
```js
{
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "intro": {
      "type": "object",
      "properties": {
        "slides": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Slide"
          }
        }
      }
    }
  },
  "definitions": {
    "Slide": {
      ...
      ...
    }
  }
}
```
Using the above schema on some data it will remove "intro.slides" array from data...

travisdahl:
@ValYouW  i think this project would really benefit from trying to resolve refs and allOfs first thing, before it does anything with it. It would keep the code in this project much simpler.. see example in this comment....

https://github.com/mozilla-services/react-jsonschema-form/issues/1237#issuecomment-477394478

ValYouW:
Ah, thx for the reference...

travisdahl:
@epicfaace what do you think about adding the libraries mentioned in https://github.com/mozilla-services/react-jsonschema-form/issues/1237#issuecomment-477394478 ?

If you always resolve the refs and allOfs right off the bat, we can get rid of some of the complexity in the current `resolveSchema`.  Let me know what you think. I would be willing to try and figure it out and make a PR if you think it would be worthwhile.

CodeGains:
@ValYouW Nice catch, I'll try to fix that now. Worst case, is that I can resolve the references myself (like @travisdahl suggested) in my solution before building up my name schema.

ValYouW:
I ended up using `@cloudflare/json-schema-ref-loader` which deref during "build time"...

CodeGains:
I've created PR #1283 with the updated solution. I also fixed the problem @ValYouW mentioned.

travisdahl:
close in favor of https://github.com/mozilla-services/react-jsonschema-form/pull/1283/files

PR #1191, opened 2019-02-23
Remove unnecessary asNumber logic

epicfaace:
This is for PR #1183. I believe we don't need this logic anymore, with the changes implemented in the aforementioned PR.

### DIFF
--- a/packages/core/test/utils_test.js
+++ b/packages/core/test/utils_test.js
@@ -1152,12 +1152,11 @@ describe("utils", () => {
     });
 
     it("should return the raw value if the input ends with a dot", () => {
-      expect(asNumber("3.")).eql("3.");
+      expect(asNumber("3.")).eql(3);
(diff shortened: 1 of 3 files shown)


glasserc:
This looks like it got a bad merge somehow? I think you just intended to include the last commit.

LucianBuzzo:
@epicfaace Can you rebase this PR on top of master please?

epicfaace:
@LucianBuzzo done!

packages/core/test/utils_test.js
```diff
@@ -1152,12 +1152,11 @@ describe("utils", () => {
     });
 
     it("should return the raw value if the input ends with a dot", () => {
```
epicfaace:
```suggestion
    it("should convert the value to an integer if the input ends with a dot", () => {
```

PR #1194, opened 2019-02-25
Limit cases where "required" attribute is used on checkboxes

### LINKED ISSUE #338: Boolean checkbox can't send false
### Prerequisites
- [ x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme).
### Description

Boolean checkbox's require you to tick the box, you cannot send a false.  
### Steps to Reproduce

Create a  form where a boolean field is required
#### Expected behavior

Either a unticked / ticked state will be accepted
#### Actual behavior

Only a ticked box is accepted.  

I'm using a jsonschema-form to send a  json to an api that uses https://github.com/geraintluff/tv4 as a validator using the same schema.  tv4's interpretation of a required boolean is it can be true or false as long as it is present.  Which implementation is right?
### Version

^0.38.0

LucianBuzzo:
Fixes #338

If the "required" attribute is applied to a checkbox, then HTML5
validation will only pass if the checkbox is ticked. This is at odds
with the JSON schema definition of "required" that simply requires that
the key exists.

This PR will make it so that the "required" HTML5 attribute will only be
applied if the schema uses `{ const: true }` or `{ enum: [ true ] }`.

[Please describe them here]

### Checklist

### DIFF
--- a/src/components/widgets/CheckboxWidget.js
+++ b/src/components/widgets/CheckboxWidget.js
@@ -2,12 +2,40 @@ import React from "react";
 import PropTypes from "prop-types";
 import DescriptionField from "../fields/DescriptionField.js";
 
+// Check to see if a schema specifies that a value must be true
+function schemaRequiresTrueValue(schema) {
+  // Check if const is a truthy value
+  if (schema.const) {
+    return true;
+  }
+
+  // Check if an enum has a single value of true
+  if (schema.enum && schema.enum.length === 1 && schema.enum[0] === true) {
+    return true;
+  }
+
+  // If anyOf has a single value, evaluate the subschema
+  if (schema.anyOf && schema.anyOf.length === 1) {
+    return schemaRequiresTrueValue(schema.anyOf[0]);
+  }
+
+  // If oneOf has a single value, evaluate the subschema
+  if (schema.oneOf && schema.oneOf.length === 1) {
+    return schemaRequiresTrueValue(schema.oneOf[0]);
+  }
+
+  // Evaluate each subschema in allOf, to see if one of them requires a true
+  // value
+  if (schema.allOf) {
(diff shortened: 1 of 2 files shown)


epicfaace:
@LucianBuzzo What about this schema?

```
{
      "type": "boolean",
      "oneOf": [
        {
          "const": true
        }
      ]
}
```

src/components/widgets/CheckboxWidget.js
```diff
@@ -16,6 +15,14 @@ function CheckboxWidget(props) {
     onFocus,
     onChange,
   } = props;
+
+  // Because an unchecked checkbox will cause html5 validation to fail, only add
+  // the "required" attribute if the field value must be "true", due to the
+  // "const" or "enum" keywords
+  const required =
+    schema.const ||
```
epicfaace: Do you think we need to put this, or is it unnecessary?:
```suggestion
    schema.const === true ||
```

LucianBuzzo: It feels a bit unnecessary, we only really want to check that `const` is truthy

epicfaace:
> @LucianBuzzo What about this schema?
> 
> ```
> {
>       "type": "boolean",
>       "oneOf": [
>         {
>           "const": true
>         }
>       ]
> }
> ```

@LucianBuzzo Just bumping this.

LucianBuzzo:
@epicfaace Thanks for the reminder, I'll hopefully get time to add this case this afternoon

LucianBuzzo:
@epicfaace I've added additional logic for handling cases where `anyOf`, `oneOf` and `allOf` would require the field to be `true`

epicfaace:
Looks great!

PR #1212, opened 2019-03-07
Fix id of anyof select

epicfaace:
This fixes a change that was done in #1118. It seems like the id should be suffixed with a double underscore, not a single underscore.

For example, see examples here (https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/ArrayField.js#L27)


epicfaace:
@LucianBuzzo could you please review this?

edi9999:
I mean a test like here : https://github.com/mozilla-services/react-jsonschema-form/blob/master/test/BooleanField_test.js#L40

Where you render that component and verify that the id is the expected one.

epicfaace:
I've added tests -- but I realized that both the oneof and anyof widgets have the same id (i.e., "__anyof_select" instead of "__oneof_select" and "__anyof_select"). Was this intentional or do you think we should fix this as well?

edi9999:
Good catch, I think it is not intentional and we should fix it.

epicfaace:
This is technically a backwards incompatible change anyway, so let's just add this in v2.

epicfaace:
@LucianBuzzo / @edi9999  could you review?

PR #1213, opened 2019-03-08
Basic support for 'nullable' types

warrenseymour:

Modifies the implementation of `getSchemaType` so that when a property has a type of `[<any>, 'null']`, then `<any>` will be passed. This allows the library to render and validate the field correctly. For the sake of simplicity, this change does not attempt any further assumptions, coercions or transformations.

References #465 

### Checklist


warrenseymour:
I wanted to land this PR in draft status before the end of the week in the hope that it'll get some eyes over the weekend. If this feature is desirable to the project maintainers, I'll add some tests and a Playground example when I return on Monday.

warrenseymour:
@epicfaace Forgive the direct tag - this is ready for review, but I can't get the PR out of draft status. Hopefully the controls will appear for project maintainers.

epicfaace:
Looks good! Thanks for the PR. One question: how does the code actually know to put `null` as the value when you leave a field empty that has `type: ["string", "null"]` ? Is this coming from ajv?

Can you add a test in which it renders a form with a nullable field and checks to make sure that when the field is emptied, its value is null?

warrenseymour:
@epicfaace This PR doesn't actually go that far; having looked the history of closed, unmerged PRs that attempted to implement UX changes or other additional handling of such nullable types, I intentionally decided against that. This allows end-users to handle such scenarios themselves, either through custom widget implementations or uiSchema adjustments.

To quote from the FAQ entry change:

> However, the actual rendering and handling of the field is unchanged; you are free to handle this using an approach (`'ui:defaultValue': null`, for example) best-suited to your use case.

Should I expand on this in the FAQ and/or elsewhere to make this point more clear?

docs/index.md
```diff
@@ -281,6 +281,8 @@ $ git push --tags origin master
...
 
+In addition, "nullable" types are supported in a narrow sense: If a property declares a type of `["<some-type>", "null"]`, then `"some-type"` will be passed through as the type used to determine which widget to use for rendering the field. However, the actual rendering and handling of the field is unchanged; you are free to handle this using an approach (`'ui:defaultValue': null`, for example) best-suited to your use case.
```
epicfaace:
```suggestion
In addition, "nullable" types are supported in a narrow sense: If a property declares a type of `["<some-type>", "null"]`, then `"some-type"` will be passed through as the type used to determine which widget to use for rendering the field. However, the actual rendering and handling of the field is unchanged; you are free to handle this using an approach (`'ui:emptyValue': null`, for example) best-suited to your use case.
```

epicfaace:
@warrenseymour oh I see, thanks for the explanation. Just one small typo then and I think we'll be good to go! Thanks again!

I do think it might be nice to have some default handling of nullable types, but that can be added in the future as this change already goes pretty far.

FriOne:
@epicfaace hi, could you bump this change into npm?

epicfaace:
I don't have access, @glasserc does

FriOne:
Thought this solves validation problem when you want 3 types of values - any, undefined and null, but it is for widget only :(

epicfaace:
@FriOne I don't think it's possible to set a value equal to `undefined` in json schema.

FriOne:
@epicfaace sorry, I meant that this change doesn't allow null in values. I can set string or undefined and validation will be passed but can't set null.

epicfaace:
@FriOne it should allow null in values. I've just updated the playground to the latest version; can you make an example on the playground and share it with me?

FriOne:
@epicfaace the problem is on my side, my bad :)

PR #1224, opened 2019-03-17
Add schema name to console warning on anyOf error

paintedbicycle:

This PR tries to resolve https://github.com/mozilla-services/react-jsonschema-form/issues/1188

It adds the name of the schema to the console warning when an anyOf validation fails.

I don't know the codebase well, so I'm not sure if this validation is reused in other places, nor if this change would also be beneficial in other places. Feel free to update before merge.

**Update**: I also see I committed the package-lock file. I'm assuming you can remove/revert this before merging?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-jsonschema-form",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/utils.js
+++ b/src/utils.js
@@ -603,8 +603,11 @@ function withExactlyOneSubschema(
     }
   });
   if (validSubschemas.length !== 1) {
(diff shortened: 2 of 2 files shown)


paintedbicycle:
Can you point out why this doesn't work?

In this case, the developer chose a unique identifier, "a", and the error message is telling them that there is a problem with one of their subschemas, identifiable by the unique id they chose. Perhaps I'm missing a subtlety here though.

You know the codebase a lot better than me. When I logged the variables in order to try to get a "name" I could output, I used what was available to me.

epicfaace:
@paintedbicycle I meant that in the example I linked, I would expect the error to say `"ignoring oneOf in dependencies for 'Do your have any pets?' because there isn't exactly one subschema that is valid"` -- the error has nothing to do with the property `a`.

epicfaace:
@paintedbicycle just checking on this.

Issue #1222, opened 2019-03-16
Create a `withTheme` higher-order component


epicfaace:
Ideally, we should have a `withTheme` HOC in which fields, widgets, and/or templates can be passed -- this can be used to implement custom themes.

This was implemented in #1013, but that included several other breaking changes -- so I'd like to see a PR which just implements the `withTheme` component so it can be added to the library.

Example usage:
```
import withTheme from 'react-jsonschema-form/lib/components/withTheme';

import widgets from './my-theme/widgets';
import templates from './my-theme/templates';

const Form = withTheme('MyTheme', { widgets, templates });
```

I'm listing the popular / currently maintained libraries that fork off of react-jsonschema-form and implement a UI framework other than bootstrap 3, so that the maintainers of those repos are aware of the progress on this issue -- or if anyone wants to jump in and do this, that would be great! @peterkelly @cwhatley @vip-git @selfkeyfoundation @eddyzhang1986 @twoabove @nilportugues

BS4
https://github.com/peterkelly/react-jsonschema-form-bs4
Material
https://github.com/cwhatley/material-ui-react-jsonschema-form
https://github.com/vip-git/react-jsonschema-form-material-ui
https://github.com/SelfKeyFoundation/react-jsonschema-form-material-theme
https://github.com/TwoAbove/jsonschema-form-for-material-ui
Antd
https://github.com/eddyzhang1986/antd-jsonschema-form
Semantic UI
https://github.com/nilportugues/react-jsonschema-form-semanticui

See #899

TwoAbove:
Woah, never though I'd be mentioned here!

Not sure that I'll be helpful here, but when this is implemented I'll refactor my fork to use this withTheme. 

What is 'MyTheme' for? Just identification?

Kenny11CZ:
I already have implemented something like `withTheme(themeObject)`, where themeObject is containing *widgets*, *fields* and *templates*. It was not that hard but, since this is my first HOC I might have done something wrong 😄

epicfaace:
@Kenny11CZ that sounds perfect! It would be great if you could make a PR with that 😄

epicfaace:
@TwoAbove 
> What is 'MyTheme' for? Just identification?

That's a good point, I don't think the first argument is needed.

[...]

PR #1226, opened 2019-03-18
Added withTheme HOC

danbalarin:
Implemented withTheme HOC in order to be able to create custom theme without having to create forks. #1222

Example usage:
```jsx
import React, { Component } from 'react'
import { withTheme } from 'react-jsonschema-form'
import Bootstrap4Theme from 'react-jsonschema-form-theme-bs4'

class Demo extends Component {
  render() {
    return withTheme(Bootstrap4Theme)
  }
}
```

Theme object can provide *widgets*, *templates*, *fields* and *form* in order to customize generated form.

[Example Bootstrap 4 theme](https://github.com/Kenny11CZ/react-jsonschema-form-theme-bs4)

### Checklist


andronat:
@epicfaace any news regarding this?

src/withTheme.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import Form from "./";
+
+function withTheme(data) {
+  return class extends Component {
+    render() {
+      let { templates, widgets, fields, ...restData } = this.props;
```
epicfaace: Can you rename `restData` to `otherProps`?

Kenny11CZ: Sure, I'm very bad at naming variables.

Kenny11CZ:
So I added documentation, but I'm pretty bad at writing those things, so it would be nice if someone could look at it, fix typos, spelling errors and so on...

epicfaace:
Thank you! Can you also add some tests? (using `withTheme` with a simple theme and making sure the corresponding widgets, templates, fields, etc. are used)

Kenny11CZ:
I can try it, but I've written one test in javascript so far. And I don't have a lot of time, so don't know when I will do it...

epicfaace:
No problem, I can add the tests if you are not comfortable doing so.

Kenny11CZ:
That would be awesome, I have no idea how to test this functionality, so at least I would learn something new

andronat:
Any news regarding this?

TwoAbove:
Nice! Any news on this? Can't wait to implement the usage in my repo!

epicfaace:
I'll be writing a test for this soon! Then will be ready to merge

epicfaace:
@danbalarin I don't understand why the ability to pass in a custom form component to the theme is needed. Would there really be different kinds of data handling that each theme would require?

danbalarin:
Umm, actually I used it in [Bootstrap 4](https://getbootstrap.com/docs/4.0/components/forms/#custom-styles) since there might be needed to add `was-validated` class after validation. This is because when you have some validation made by HTML5 input itself (*number*, *email*, *regex*,...) when you submit the form, it's browser dependent how it shows the error (Firefox shows error input slightly differently from Chromium) and therefore inconsistent. But by providing custom `<form>`, there comes other issues like remembering the data filled in and with each submit refill all inputs (the stuff default form does...).

epicfaace:
@danbalarin Can you remove the custom form option for now and perhaps move that to a later PR? I'm not convinced that a custom form is the right way to go, because then it seems like it involves _too much_ customization; one can pass in an arbitrary component (not even one based on react-jsonschema-form's `Form`) for `form`.

For the case you mentioned, maybe the option of "adding a class after validation" should be implemented as a feature of react-jsonschema-form itself?

danbalarin:
I've removed custom form.

epicfaace:
Thanks! Can you update the documentation too? Then we should be all good to go.

danbalarin:
> Thanks! Can you update the documentation too? Then we should be all good to go.

Done

mirajp:
Question: Why require the templates be packaged into a temporary object when the underlying Form component itself does not do so? Think it'd be simpler/more direct to just allow template overrides be passed through the same way the other props are; otherwise, users of the ThemedForm would also need to package any custom field template overrides they want to perform on an instance-by-instance basis inside a makeshift "templates" object (thus diverging from how the native Form does it, and thereby requiring each theme provider to document this usage that differs from/contradicts the original docs).

I also don't think this HOC should limit itself to just widgets, fields(, and templates) and should pretty much allow for the passing through of all custom props (the data object). For instance, a "theme" object may want to turn off the native errors list (`showErrorList` prop because users (at least mine) may not want a separate error list on the top of the form when each field template already has a custom error message shown under the field. Thus, it would be nice to simply create a themed Form by doing something like this:
```
const ThemedForm = withTheme({
    FieldTemplate: MyThemeFieldTemplate,
    ...,
    showErrorList: false,
    fields: MyThemeFields,
    widgets: MyThemeWidgets
});
```
I get that at this point this "withTheme" HOC really just overrides every property (not just what some people would deem "theme" related), but I think it would prove much more useful. Otherwise, any Form customizations would have to extend the native Form or HOC class component in order to attach nice-to-have defaults for props like "showErrorList".

Proposed tweaks to this PR:
```
function withTheme(themeProps) {
  return class extends Component {
    render() {
      let { fields, widgets, ...directProps } = this.props;
      fields = { ...themeProps.fields, ...fields };
      widgets = { ...themeProps.widgets, ...widgets };
      return (
        <Form
          {...themeProps}
          {...directProps}
          fields={fields}
          widgets={widgets}
        />
      );
    }
  };
}
```
You may also want to consider a deep merge of the themeProps and the directProps to account for theme providers that want to provide defaults for other object-type properties (e.g. `formContext` or `customFormats`).

epicfaace:
@mirajp good idea. What do you mean when you said that "templates [are] packaged into a temporary object"?

mirajp:
@epicfaace based on how the HOC function is currently designed in this PR:
```
function withTheme(data) {
  return class extends Component {
    render() {
      let { templates, widgets, fields, ...otherProps } = this.props;
      templates = { ...data.templates, ...templates };
      widgets = { ...data.widgets, ...widgets };
      fields = { ...data.fields, ...fields };
      return (
        <Form
          {...otherProps}
          {...templates}
          widgets={widgets}
          fields={fields}
        />
      );
    }
  };
}
```
`templates` has to be a special object (keyed "templates") in the `data` object passed into the HOC function, and likewise instances of this themed Form would need to pass in a similar object to override any specific template:
```
const ThemedForm = withTheme({
    fields: {...}
    templates: {ArrayFieldTemplate: ThemedArrayFieldTemplate},
    widgets: {...},
});

<ThemedForm
    ...
    templates={{ ArrayFieldTemplate: MySpecialArrayFieldTemplate }}
/>
```
However, the actual rjsf Form has no such special `templates` object as a prop; instead, any field template override is done directly as a props/jsx attribute: https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/#array-field-template

```
<Form
    ...
    ArrayFieldTemplate={MySpecialArrayFieldTemplate}
    ObjectFieldTemplate={MySpecialObjectFieldTemplate}
/>
```
Note that in the native rjsf Form, only `fields`, `widgets`, `customFormats`, and `formContext` are object-type properties/jsx attributes. Thus, with my proposed tweak in my original comment, the proposed `withTheme({ ... })` HOC would allow more props to fall through (including all the *FieldTemplate props). 

If the implementation was even less heavy-handed/opinionated (currently only merges together the `fields` and `widgets` objects) and did a deep merge on all object properties from both the `data` props of withTheme (I think this should be renamed to themeProps) and the form instance props (`this.props`), you could even allow themes to specify their own formContext and customFormats values that developers could amend on an instance-by-instance basis if any configs needed to be stored there for any runtime logic performed in the custom/themed widgets, etc.

epicfaace:
@mirajp that's a good point. The `templates`, etc. structure of `withTheme` was based off of this PR (https://github.com/mozilla-services/react-jsonschema-form/pull/1013), but since that PR is probably not going to be merged at this point, and I don't see the structure of the form props changing in the near future, it might make more sense to make it like what you discussed. Would you be able to modify this PR to have that behavior?

mirajp:
@epicfaace I don't think I can modify this PR since I'm not a collaborator of this repo nor an authorized contributor of @danbalarin's master branch that this PR is merging from. All I can independently do is make a PR against his master branch or against rjsf's master branch -- may be simpler to just wait for @danbalarin to check back and copy and paste my snippet.

danbalarin:
Originally I made it this way so you can define theme in a separate file and use default export. An object of this size is negligible because all it does is holding three pointers, so is passing it as a param, its only passing pointer not the whole object.

@mirajp I'm sorry, but I have a lot of work that needs to be done, I added you as a collaborator, so you can add proposed changes.

mirajp:
@danbalarin I believe you should still be able to still the default export on the Theme options/props object, just this object will have all the field templates not nested under a templates object (I think the # of pointers would actually decrease (by 1 -- no pointer to a new templates object) and # of objects would also decrease (by 1 -- no templates object) :P -- though this is pretty insignificant in the grand scheme of things). I can review your branch and modify the PR and confirm your tests still work when I get a chance later today.

mirajp:
@danbalarin @epicfaace updated the PR.

docs/theme-customization.md
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const ThemeObject = {widgets: myWidgets};
+export default ThemeObject;
+```
+
+The above can be similarly done for **fields**.
+
+#### Templates
+Each template should be passed directly into the theme object just as you would into the rjsf Form component. In [here](/advanced-customization/#array-field-template) and [here](/advanced-customization/#error-list-template) are two examples of custom templates, below is example how to use these two custom templates in the theme object:
```
epicfaace:
```suggestion
Each template should be passed directly into the theme object just as you would into the rjsf Form component. Here is an example of how to use a custom [ArrayFieldTemplate](/advanced-customization/#array-field-template) and [ErrorListTemplate](/advanced-customization/#error-list-template) in the theme object:
```

docs/theme-customization.md
```diff
@@ -0,0 +1,94 @@
...
+const ThemeObject = {
+    ArrayFieldTemplate: MyArrayFieldTemplate, 
+    ErrorList: MyErrorListTemplate,
+    widgets: myWidgets
+};
+
+export default ThemeObject;
+```
+
+### Overriding other Form properties
```
epicfaace:
```suggestion
### Overriding other Form props
```

mkdocs.yml
```diff
@@ -8,6 +8,7 @@ nav:
   - Definitions: definitions.md
   - Dependencies: dependencies.md
   - Form Customization: form-customization.md
+  - Customizing with other frameworks: theme-customization.md
```
epicfaace:
```suggestion
  - Theme Customization: theme-customization.md
```

test/withTheme_test.js
```diff
@@ -0,0 +1,34 @@
...
+
+  beforeEach(() => {
+    sandbox = createSandbox();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe("With widgets", () => {
```
epicfaace: Can you add some more tests here that test the widgets, templates, and also the overriding?

mirajp: How do I even get the tests to run with changes? They keep running without any of my changes to the same file lol... (and only runs this one file, none of the other test files) I can literally delete all the tests and it still runs it - is there some weird build step involved and it only runs whatever got built?
I think I got it to work once by running cs-format and cs-check, but can't get `npm run test` to pick up any new changes to the file.

test/withTheme_test.js
```diff
@@ -0,0 +1,34 @@
+import { expect } from "chai";
+import React from "react";
+
+import { withTheme } from "../src";
+import { createComponent, createSandbox } from "./test_utils";
+
+describe.only("withTheme", () => {
```
epicfaace:
```suggestion
describe("withTheme", () => {
```

@mirajp change `describe.only` to `describe` and it should solve the problem you mentioned about tests 😉

mirajp: Never mind, I had to manually call `touch` on the file... Idk how chai/whatever this test infrastructure works but I guess it caches copies of the file and doesn't know it's updated?

test/withTheme_test.js
mirajp: @epicfaace added a few tests. I tried to get a custom ObjectFieldTemplate to render, but it never worked/was never called by Form =/ dunno why.

PR #1228, opened 2019-03-19
trim empty objects to allow correct validation of optional objects with mandatory fields

numical:

**NOT FOR MERGE YET!**
Test failures remain - see discussion below.

This PR offers a potential solution for the issue of allowing optional objects with required fields to pass validation when none of their values are defined.

The approach is similar to that discussed in [Issue 675](https://github.com/mozilla-services/react-jsonschema-form/issues/675) and proposed, then withdrawn, in [PR 682](https://github.com/mozilla-services/react-jsonschema-form/pull/682).

Prior to validation, a clone of formData is generated where all objects with no defined values are removed (note: empty arrays are **not** removed).

This works nicely - **but changes the semantics of the errors**. Primarily 'is a required property' errors become 'should be object'.  Currently I have **not**  changed the tests so as to display that.

I have raised this PR to essentially ask whether this approach could be used and, if so, to seek advice on what to about the changed errors (I note the conversation [here](https://github.com/mozilla-services/react-jsonschema-form/pull/682#issuecomment-326076044)).

If this is related to existing tickets, include links to them as well:
https://github.com/mozilla-services/react-jsonschema-form/issues/675
https://github.com/mozilla-services/react-jsonschema-form/pull/682

### Checklist

* [X ] **I'm adding or updating code**
  - [ X] I've added and/or updated tests
  - [X ] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,6 +124,7 @@ yarn.lock
 
 # IDE
 .vscode
+.idea
 
 # Code coverage
 coverage
--- a/src/utils.js
+++ b/src/utils.js
@@ -844,3 +844,34 @@ export function rangeSpec(schema) {
   }
   return spec;
 }
+
+function trimObject(object) {
+  return Object.entries(object).reduce((acc, [key, value]) => {
+    const trimmed = trimEmptyValues(value);
+    if (trimmed) {
+      if (!acc) {
+        acc = {};
+      }
+      acc[key] = trimmed;
+    }
+    return acc;
+  }, undefined);
+}
+
+function trimArray(array) {
+  return array.reduce((acc, value) => {
+    const trimmed = trimEmptyValues(value);
+    if (trimmed) {
+      acc.push(trimmed);
+    }
+    return acc;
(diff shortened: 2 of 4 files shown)


epicfaace:
Hi @numical , thanks for the PR.

I see that the tests are failing due to the errors saying "should be object"; however, when I try to replicate this error on my local playground, I still get the expected error, "is a required property". Do you know why this is the case?

image

Also, your changes work, but if HTML5 validation is enabled, then I still get the "required" dialog. Do you think there might be a way to programmatically change this in this PR as well (to conditionally add the `required` attribute)?

image

epicfaace:
@numical just checking on this.

Morriz:
any progress here? It would be reallllly beneficial to not get errors on empty nested props with required values

heath-freenome:
The issue attempted to be fixed by this PR was actually fixed in #3436

PR #1234, opened 2019-03-26
Fix #1233 : Exclude File object from isObject detection

### LINKED ISSUE #1233: File object becomes a basic js object
### Description
When we use the From component in a wrapper component and pass formData via props., the Form component transforms Props to State in `componentWillReceiveProps ` . Problem is when we use a custom widget with an input file, the `File` object is transformed to a basic object in [mergeObjects function](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/utils.js#L242) because the  `File` object is detected as an object.

I guess the solution is to exclude File from [isObject](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/utils.js#L232) function. 

Do you think it's the way to go? I don't see any exception than `File` object which could be bad detected.

### Steps to Reproduce

1. Open https://codesandbox.io/s/n4mlxwq3ml
2. Open the browser console
3. Choose a file
4. See in the console onChange event is triggered twice and the File object changes.

#### Expected behavior
The File object stays a File object

#### Actual behavior

The File object becomes a basic js object

### Version

1.3.0

oterral:
Fix https://github.com/mozilla-services/react-jsonschema-form/issues/1233

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -229,6 +229,9 @@ export function getUiOptions(uiSchema) {
 }
 
 export function isObject(thing) {
+  if (thing instanceof File) {
+    return false;
+  }
   return typeof thing === "object" && thing !== null && !Array.isArray(thing);
 }
 
--- a/test/utils_test.js
+++ b/test/utils_test.js
@@ -517,6 +517,17 @@ describe("utils", () => {
       expect(mergeObjects(obj1, obj2)).eql(expected);
     });
 
+    it("shouldn't recursively merge File object", () => {
+      const file = new File(["test"], "test.txt");
+      const obj1 = {
+        a: {},
+      };
+      const obj2 = {
+        a: file,
+      };
+      expect(mergeObjects(obj1, obj2).a).instanceOf(File);
+    });
+
     describe("concatArrays option", () => {
       it("should not concat arrays by default", () => {
         const obj1 = { a: [1] };
--- a/test/utils_test.js
+++ b/test/utils_test.js
@@ -517,7 +517,7 @@ describe("utils", () => {
       expect(mergeObjects(obj1, obj2)).eql(expected);
     });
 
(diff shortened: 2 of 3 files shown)


oterral:
@epicfaace

src/utils.js
```diff
@@ -229,7 +229,9 @@ export function getUiOptions(uiSchema) {
 }
 
 export function isObject(thing) {
-  return typeof thing === "object" && thing !== null && !Array.isArray(thing);
+  return thing instanceof File
```
epicfaace: Can you actually add an if statement here? For readability, and if we need to do additional checks in the future

oterral: I do that

oterral: Done

test/utils_test.js
```diff
@@ -517,6 +517,17 @@ describe("utils", () => {
       expect(mergeObjects(obj1, obj2)).eql(expected);
     });
 
+    it("shouldn't recursively merge File object", () => {
```
```suggestion
    it("should recursively merge File objects", () => {
```

oterral:
Thx @epicfaace

davidb-e4s:
Since 1.4.0 I'm now getting this error when trying to create a form

ReferenceError: File is not defined at isObject (/srv/node_modules/react-jsonschema-form/lib/utils.js:325:24) at getDefaultFormState (/srv/node_modules/react-jsonschema-form/lib/utils.js:288:8) at Form.getStateFromProps (/srv/node_modules/react-jsonschema-form/lib/components/Form.js:157:53) at new Form (/srv/node_modules/react-jsonschema-form/lib/components/Form.js:129:25)

1.3.0 works fine

epicfaace:
@davidb-e4s thanks for noting that, looks like a browser compatibility issue. Can you make an issue with this and put down your browser version?

glasserc:
The issue in question is #1252.

PR #1238, opened 2019-03-27
Add NullField

faissalMT:

The JSON Schema type 'null' is currently validated and inferred as a type but shows an error when trying to render. This corrects the issue such that null fields will render nothing, allowing for things like additional help text or use as a temporary placeholder field.  
  
#1232

### Checklist

### DIFF
--- a/src/components/fields/NullField.js
+++ b/src/components/fields/NullField.js
@@ -3,7 +3,9 @@ import PropTypes from "prop-types";
 
 class NullField extends Component {
   componentDidMount() {
-    this.props.onChange(null);
+    if (this.props.formData === undefined) {
+      this.props.onChange(null);
+    }
   }
 
   render() {
--- a/src/components/fields/NullField.js
+++ b/src/components/fields/NullField.js
@@ -1,5 +1,5 @@
 import { Component } from "react";
-import PropTypes from "prop-types";
+import * as types from "../../types";
 
 class NullField extends Component {
   componentDidMount() {
@@ -14,9 +14,7 @@ class NullField extends Component {
 }
 
 if (process.env.NODE_ENV !== "production") {
-  NullField.propTypes = {
-    onChange: PropTypes.func,
-  };
+  NullField.propTypes = types.fieldProps;
 }
 
 export default NullField;
--- a/src/components/fields/NullField.js
+++ b/src/components/fields/NullField.js
@@ -1,5 +1,20 @@
-function NullField(props) {
-  return null;
+import { Component } from "react";
+import PropTypes from "prop-types";
+
+class NullField extends Component {
+  componentDidMount() {
+    this.props.onChange(null);
+  }
+
+  render() {
+    return null;
+  }
+}
+
+if (process.env.NODE_ENV !== "production") {
+  NullField.propTypes = {
+    onChange: PropTypes.func,
+  };
 }
 
 export default NullField;
--- a/src/components/fields/NullField.js
+++ b/src/components/fields/NullField.js
@@ -1,0 +1,5 @@
+function NullField(props) {
+  return null;
+}
+
+export default NullField;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-jsonschema-form",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -4599,7 +4599,6 @@
       "resolved": "https://registry.npmjs.org/glob-parent/-/glob-parent-2.0.0.tgz",
       "integrity": "sha1-gTg9ctsFT8zPUzbaqQLxgvbtuyg=",
       "dev": true,
-      "optional": true,
       "requires": {
         "is-glob": "^2.0.0"
       }
@@ -5268,8 +5267,7 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/is-extglob/-/is-extglob-1.0.0.tgz",
       "integrity": "sha1-rEaBd8SUNAWgkvyPKXYMb/xiBsA=",
-      "dev": true,
-      "optional": true
+      "dev": true
     },
     "is-finite": {
       "version": "1.0.2",
@@ -5291,7 +5289,6 @@
       "resolved": "https://registry.npmjs.org/is-glob/-/is-glob-2.0.1.tgz",
       "integrity": "sha1-0Jb5JqPe1WAPP9/ZEZjLCIjC2GM=",
       "dev": true,
-      "optional": true,
       "requires": {
         "is-extglob": "^1.0.0"
       }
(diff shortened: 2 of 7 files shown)


epicfaace:
@faissalMT , thanks for the PR. In #1232, you mention that null fields should render the title only and not the input box -- but this PR makes null fields render nothing entirely. Was this intentional?

Additionally, I'm wondering how this change plays with the changes done in #1213. If this PR makes it such that "null" fields render nothing, then will our current handling of nullable types be unintuitive?

(#1213 makes a field with type `["string", "null"]` still render the input box for a string. If "null" fields actually render nothing, could one intuitively expect a field with type `["string", "null"]` to actually render a OneOf widget with either an input box or nothing in it?)

Maybe this isn't really a concern though so I'd like to hear your thoughts @faissalMT @warrenseymour

warrenseymour:
@epicfaace I've not taken the time to clone this locally and experiment, but as far as I can tell, from a code perspective, this *shouldn't clash* with what I did in #1213, since that does a reasonably extensive check that the field type is `[<T>, "null"]` or `["null, <T>]` and then simply passes the type on as `<T>`. So I think in the case presented here, `type: "null"` should be handled as @faissalMT intends.

From a UX perspective, however, I agree that explicitly adding a `Null` widget may lead to confusion and would inclined to push this sort of decision onto the consumer of RJSF - if they want to explicit enable a `Null` widget, it should be on them to opt-into by bringing their own Widget, or at least explicitly enabling that option if it is built-in.

Finally, trying to infer/envisage a legitimate use-case for this I looked at a commit that @faissalMT cited in #1232. It *looks* as if he's just trying to render a piece of inert content in the form. Would a `uiSchema` entry using a custom widget be a reasonable way of solving that issue?

faissalMT:
@epicfaace  
>but this PR makes null fields render nothing entirely. Was this intentional?  
  
RJSF itself is taking care of handling the rendering of those things (as per the playground example I added), so there was no need for me to implement it explicitly.  
  
>I agree that explicitly adding a Null widget may lead to confusion and I would be inclined to push this sort of decision onto the consumer of RJSF  
  
I would expect any native JSON schema type to be renderable by RJSF, if the user doesn't like the handling they can always override it. I think this makes the most sense and is significantly better than an error.  
  
>Would a uiSchema entry using a custom widget be a reasonable way of solving that issue?  
  
In that project we have a certain level of preprocessing that generates a uiSchema, however we must provide it with a non-null type at the moment or RJSF will refuse to render our custom widget. This also works well for creating placeholder fields.

epicfaace:
@faissalMT thanks, I didn't realize that it was rendering the title/description as well.

It makes sense how rjsf should have some kind of default widget for null types. However, my primary concern is this: when you make a field such as `helpText` with `type: null`, then you would expect `helpText: null` to be in the formData. However, in this case, there is no key for `helpText` in the formData.

image

What if we make `helpText` required? Then we get stuck with an "is a required property" error, which is only resolved when you actually manually set `helpText: null` in the formData.

image

Perhaps the solution for this is to make sure that a `NullWidget` calls the `onChange` handler with `null`, so that `helpText: null` is actually set in the formData. I'd be ready to merge this PR if we decide it makes sense to do that.

I don't know if, as @warrenseymour noted, this entirely fits the use case to "render a piece of inert content in the form". Do we want to clutter our formData by adding in extra null keys when we just want to add in additional content in the form? Perhaps the real solution fo this _is_ in the uiSchema though.

faissalMT:
@epicfaace  
  
>Perhaps the solution for this is to make sure that a NullWidget calls the onChange handler with null  
  
I like this solution and have committed it. An alternative could also simply be to use a default value, but that seems more tacky to me.  
  
>Do we want to clutter our formData by adding in extra null keys when we just want to add in additional content in the form? Perhaps the real solution fo this is in the uiSchema.  
  
I feel like modifying uiSchema to work with non-existent fields as if they are existing fields could quickly get ugly. The uiSchema may serve to further expand on the details of this content (such as use of a custom widget etc), but items in the uiSchema should always correspond to a real field or I believe some fundamental assumptions this project has made are broken.

src/components/fields/NullField.js
```diff
@@ -1,5 +1,20 @@
-function NullField(props) {
-  return null;
+import { Component } from "react";
+import PropTypes from "prop-types";
+
+class NullField extends Component {
+  componentWillMount() {
```
epicfaace: Is there a reason why you did this in componentWillMount instead of componentDidMount?

faissalMT: I just forgot it was deprecated

faissalMT:
Bump?

src/components/fields/NullField.js
```diff
@@ -0,0 +1,20 @@
+import { Component } from "react";
+import PropTypes from "prop-types";
+
+class NullField extends Component {
+  componentDidMount() {
+    this.props.onChange(null);
```
epicfaace: Perhaps it would be better to call onChange only if `props.value === undefined`?

faissalMT: Agreed, that behaviour might come in handy.

epicfaace:
@faissalMT there's just one thing, thanks for the reminder. I was hesitant to merge this because unlike other widgets, this is the only widget that automatically mutates the formData without any user input (by calling `onChange` on load). I'm not sure if that could potentially cause problems, or if it is okay -- what do you think?

faissalMT:
@epicfaace The system doesn't really have anything that would distinguish between actors and this is a really simple manipulation so I don't think there'll be any issues.

epicfaace:
Okay, thanks. What do you mean by "the system doesn't really have anything that would distinguish between actors"?

faissalMT:
@epicfaace I mean that nothing outside the widgets themselves in RJSF has implemented any assumptions about what kind of actor is entering the data (although they may have been programmed with that assumption), there would be nothing preventing a widget being set by fetched data for example. Our project currently implements a number of automatically filled out fields and it really isn't an issue for individual widgets acting on their own data. RJSF only cares about which widget is entering the data for which form field, and we're not breaking the assumption that a widget will only change the data of its field.

epicfaace:
Sounds good, thanks for the explanation.

epicfaace:
Should `NullField` have proptypes of the other fields, such as in:

https://github.com/mozilla-services/react-jsonschema-form/blob/ab0be5ec42939fc699a60a3779d743ea595ad496/src/components/fields/StringField.js#L60

faissalMT:
@epicfaace I didn't realise we had a central proptypes object for fields, I've changed it now.

epicfaace:
Thanks! Looks good.

tomasgvivo:
When will this changes be released?

rmdevos:
@epicfaace   Will this PR eventually be merged or will the issue with null as default value be resolved in an entirely different way?

epicfaace:
@rmdevos this PR has already been merged -- however, it only addresses fields with `type: null`, not necessarily fields with `default: null`.

PR #1248, opened 2019-04-05
use __id instead of $id to avoid collision

a-b-r-o-w-n:

it is possible to have a property named `$id` which causes issues when creating id's.

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -23,7 +23,7 @@ function ArrayFieldTitle({ TitleField, idSchema, title, required }) {
   if (!title) {
     return null;
   }
-  const id = `${idSchema.$id}__title`;
+  const id = `${idSchema.__id}__title`;
   return <TitleField id={id} title={title} required={required} />;
 }
 
@@ -31,7 +31,7 @@ function ArrayFieldDescription({ DescriptionField, idSchema, description }) {
   if (!description) {
     return null;
   }
-  const id = `${idSchema.$id}__description`;
+  const id = `${idSchema.__id}__description`;
   return <DescriptionField id={id} description={description} />;
 }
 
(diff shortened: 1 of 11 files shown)


epicfaace:
What if we have a property named `__id` though?

a-b-r-o-w-n:
@epicfaace That is also a possibility. Maybe namespacing it more would be appropriate? Something like `REACT_JSONSCHEMA_FORM__ID`?

epicfaace:
@a-b-r-o-w-n hmm, maybe. That just looks kind of ugly though...

If you have an idSchema with the value
```
{
        $id: "root",
        $id: { $id: "root_foo" },
}
```
Why is this a problem in the first place?

a-b-r-o-w-n:
So should I update with a more unique property name?

epicfaace:
@a-b-r-o-w-n What do you think about the other possible collisions; for example, you cannot style a property named `classNames` in the uiSchema?

a-b-r-o-w-n:
It probably should. I haven't run into that one specifically though. A schema our team is using defines the `$id` property though which generates ids like `[Object object]_foo_bar`. I'm happy to apply a prefix to other places if you point them out to me.

a-b-r-o-w-n:
@epicfaace what direction should I take this in?

epicfaace:
@a-b-r-o-w-n I believe a prefix like what you mentioned is the right way to go. I don't think this can be implemented in 1.x, though, as there may already be widgets that develop on the `$id` format of the idSchema. It can go in 2.0.

ronny-rentner:
I would propose to find a more generic solution that lets you move fields out of the way in case you need to use the field name internally. You could do some kind of map that says `{ id: "MyCustomID", className: "__className" }` and then map away what is in your way. Not sure where to put the mapping array, could be outside of the schema.

And then to fetch a value, instead of doing `  const id = \`${idSchema.__id}__title\`;`

you do

`const id = \`${resolveMap(idSchema, "id")}__title\`;`

which resolves to 

`  const id = \`${idSchema.MyCustomID}__title\`;`

PR #1256, opened 2019-04-16
✨ Add support for a custom style in uiSchema

justtonytone:

More flexibility to use inline styles. Handles https://github.com/mozilla-services/react-jsonschema-form/issues/1200 

### Checklist

### DIFF
--- a/docs/advanced-customization.md
+++ b/docs/advanced-customization.md
@@ -17,9 +17,9 @@ A field template is basically a React stateless component being passed field-rel
 
 ```jsx
 function CustomFieldTemplate(props) {
-  const {id, classNames, label, help, required, description, errors, children} = props;
+  const {id, classNames, style, label, help, required, description, errors, children} = props;
   return (
-    <div className={classNames}>
+    <div className={classNames} style={style}>
       <label htmlFor={id}>{label}{required ? "*" : null}</label>
       {description}
       {children}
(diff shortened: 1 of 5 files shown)


epicfaace:
Looks good! Can we make style an attribute under ui:options? That way, this change won't impact schemas that already have properties named "style".

justtonytone:
@epicfaace, that is a good point. I wanted to keep it under `uiSchema` to keep it consistent with `classNames`.

Just wanted to make sure I understood your concern. When you say schemas that already have properties named `style`, do you mean under the `properties` key? Or do you mean under the `uiSchema` key?

epicfaace:
@justtonytone Both. If a schema has a property named `style` under `properties`, then its corresponding uiSchema entry will have a property named `style` under `uiSchema`.

I think it's unfortunate how `classNames` was designed, but at least it's been around for a while and so schema creators know not to call a field with key "classNames". At this point, though, adding a `styles` attribute that is not under `ui:options` could break existing schemas that already have the property `style` in them. See https://github.com/mozilla-services/react-jsonschema-form/pull/1248 for more problems with namespacing.

epicfaace:
@justtonytone I'd be glad to merge this PR if you are able to change it so that the style attribute is accepted as follows:

```
"ui:options": {
    "style": "..."
}
```

jaxomlotus:
hi @epicfaace do you have an eta on when this feature might be included? thanks!

heath-freenome:
Reimplemented in #3378

PR #1262, opened 2019-04-20
Blob-URL Widget

james-gc:
#### Data-URL takes too long, and hangs main-thread (specially on mobile browser)
Problem with the data-url was, it was taking too long on base64 conversion, in our project we faced it, and we resolved it by creating new field, and implemented blob-url.
we can upload it by later on submit event by :
`let blob = await fetch(url).then(r => r.blob());` 

### Checklist

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "jsxBracketSameLine": true,
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "useTabs": false
+}
--- a/devServer.js
+++ b/devServer.js
@@ -12,10 +12,12 @@ const webpackConfig = require("./webpack.config." + env);
 const compiler = webpack(webpackConfig);
 const app = express();
 
-app.use(require("webpack-dev-middleware")(compiler, {
-  publicPath: webpackConfig.output.publicPath,
-  noInfo: true
-}));
+app.use(
+  require("webpack-dev-middleware")(compiler, {
+    publicPath: webpackConfig.output.publicPath,
+    noInfo: true,
+  })
+);
 
 app.use(require("webpack-hot-middleware")(compiler));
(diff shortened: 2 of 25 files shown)


james-gc:
this PR creates a blob-url rather than data-url on FileWidget.
Problem with the data-url was, it was taking too long on base64 conversion, in our project we faced it, and we resolved it by creating new field, and implemented blob-url.
we can upload it by later on submit event by :
`let blob = await fetch(url).then(r => r.blob());`

james-gc:
Removed prettier options configuration from package.json.
Now it resides from .prettierrc.
Previously it conflicts with the vscode auto-formatter, by having `.prettierrc` we can use same config for vscode and precommit hook

epicfaace:
@james-gc thanks for the PR. However, we can't implement this change as it is right now without breaking backwards compatibility (others may depend on this library storing files as a `data-url` by default).

Can you modify this PR perhaps to support an _additional_ format called `blob-url` which stores it as a blob-url?

james-gc:
@epicfaace thanks for reviewing, it will do that.
and so on I also changed styling for `choose` file button, and now it can also show the file-preview

james-gc:
@epicfaace updated all test cases for new widget `BlobFileWidget`

james-gc:
@epicfaace documentation for `BlobFileWidget` also updated

kevin24greylabs:
👍  looks good

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
epicfaace: Can you remove this prettier config changes? Would like to keep this PR self-contained to only have the blob url changes.

james-gc: :+1: Sure, I will update it.

james-gc:
@epicfaace Closing this PR

Issue #598, opened 2017-05-31
Question: hide/show functionality for large forms?


gdbassett:
### Prerequisites

- [X] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [ ] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

I have a rather large schema (roughly 150 potential values to input).  Is it possible to condense the form in either an accordion or a hide/show manner to make navigation easier?  Is this something that could be added to a form generated with the existing library or would the library itself require updating?

glasserc:
Hi, thanks for your question. #304 was meant to allow use cases like this but I don't really know what is possible using existing libraries or what you would need to do to use them together with RJSF.

glasserc:
See also https://github.com/mozilla-services/react-jsonschema-form/issues/268, which I finally just found again.

acouch:
It would be awesome if this could be included in UISchema. Maybe ``ui:order`` could contain a grouping field so you could do something like:

```yml
ui:order:
- field1
- field2
  group:
    - field 3
    - field 4
- field5
```

If this is something the maintainers would accept I could take a crack at it.

EvertLagerberg:
@glasserc How would I use Field Template to achieve this?

gdbassett:
I'm in the same boat as @EvertLagerberg.  I don't know how to use Fix #284 to collapse empty sections and ignore required fields in them.  I had expected https://github.com/mozilla-services/react-jsonschema-form/pull/542 would address it, but it still seems to be pending feedback from the author.  If it's possible to address this with #284, it may be helpful if someone can document a quick blog post or such showing how to do it.

glasserc:
There's no specific setting or configuration that #304 permits. However, with your own field template, you can write any kind of HTML you like, including one that collapses according to user input. If someone took the time to do this, an example in the docs would be great!

fantapop:
@EvertLagerberg were you able to acheive this yet?

loganvolkers:
@fantapop Here's an example using a custom `ObjectFieldTemplate`: https://codesandbox.io/s/0y7787xp0l

fantapop:
Thank you for taking the time to put that together!

neilyoung:
@fantapop Looks good, but produces tons of warnings, mostly because of this:

[...]

PR #1272, opened 2019-04-30
Add documentation example for dynamically hiding widgets

jglev:
First, thank you *very much* for your work on this project! I'm grateful for the work that's gone into the project to date.

This PR updates the "Hidden Widget" documentation to include an example for dynamically hiding / filtering aspects of large forms. This new documentation would answer #598. The change I've proposed does not illustrate using a search box and the component's React state, but could be expanded to.

Is this a change that you think would be useful for others in the community to have available as an example?

### Checklist

### DIFF
--- a/docs/form-customization.md
+++ b/docs/form-customization.md
@@ -184,6 +184,167 @@ Notes:
  - Hiding widgets is only supported for `boolean`, `string`, `number` and `integer` schema types;
  - A hidden widget takes its value from the `formData` prop.
 
+##### Filtering forms dynamically with hidden widgets
+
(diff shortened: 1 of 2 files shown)


epicfaace:
Thanks for the PR -- perhaps it would be more appropriate to add this as a codesandbox example instead of being its own section in the docs?

publicus:
Sure, I can re-jig it to follow that approach!

publicus:
@epicfaace, is 2aba7a6 in line with what you had in mind?

(`prettier` also automatically removed a few errant spaces in that commit; I can revert those if you prefer.)

publicus:
Explaining my force-push just now: I rebased onto current `master`, so that my commits are on top of all of the latest changes.

PR #1283, opened 2019-05-13
Added an 'Omit extra data' prop to remove formData fields not represented in the schema

CodeGains:

Please see #1190 for details. This PR is accomplishes the same goal as #1190, but this solution does not rely on changing the DOM.

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -150,7 +150,7 @@ export default class Form extends Component {
             newPaths = newPaths.map(path => `${path}.${key}`);
             getAllPaths(_obj[key], acc, newPaths);
           }
-        } else if (key === "name") {
+        } else if (key === "$name") {
           paths.forEach(path => {
             const formValue = _get(formData, path);
             if (typeof formValue !== "object") {
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -1,5 +1,7 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
+import _pick from "lodash/pick";
+import _get from "lodash/get";
 
 import { default as DefaultErrorList } from "./ErrorList";
 import {
@@ -10,6 +12,7 @@ import {
   setState,
   getDefaultRegistry,
   deepEquals,
+  toNameSchema,
 } from "../utils";
 import validateFormData, { toErrorList } from "../validate";
 
@@ -22,6 +25,7 @@ export default class Form extends Component {
     safeRenderCompletion: false,
     noHtml5Validate: false,
     ErrorList: DefaultErrorList,
+    omitExtraData: false,
   };
 
   constructor(props) {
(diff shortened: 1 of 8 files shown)


src/components/Form.js
```diff
@@ -1,5 +1,7 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
+import _pick from "lodash/pick";
```
epicfaace: Should we be calling `lodash.pick` and `lodash.get` (and adding `lodash.get` to `package.json`) instead?

src/utils.js
```diff
@@ -753,6 +753,49 @@ export function toIdSchema(
   return idSchema;
 }
 
+export function toNameSchema(schema, name = "", definitions, formData = {}) {
```
epicfaace: This is great -- and we can later pass on the `nameSchema` to all widgets if we need them to have a name. Would it be more appropriate to call it `pathSchema` instead though?

Additionally, given https://github.com/mozilla-services/react-jsonschema-form/pull/1248, could we potentially namespace `$name`?

Also, how would we distinguish an array element from an actual object property called "0" or "1" or "2"?

CodeGains: Ya, I think that's a good idea. 

Also, you are right about those array indexes. I can rename them to be like "$0", "$1", "$2" ...

src/components/Form.js
CodeGains: Looks like they are both being used, but I forgot to add lodash.get to the package json when I was moving the code over from the other branch. Will update it

src/utils.js
CodeGains: @epicfaace got some time to work on this now, the namespace thing you guys were talking about -- that is something that doesn't exist currently, right? We could potentially namespace  `$name` or something similar in the future.

As for the array indexes, I think even having "$0", "$1", etc. would be too common. What about something crazy for now like having escaped quotes? 

For example:

```
{
  list: {
    '0': { ... },
    '1': { ... },
    '2': { ... }
}
```

Would it be possible to namespace something like the array indexes? `somePrefix+number`?

epicfaace:
@CodeGains The complexity of this PR worries me (although if that complexity is needed to implement the functionality, then that's fine). I'm trying to understand why exactly you needed to define a nameSchema -- could you explain the reason(s) you decided to go with that? I see that, unlike idSchema, nameSchema correctly handles array items that are showing up on the form -- is there anything else that it's supposed to handle?

I think we should also add tests for `toNameSchema`. That will also help enumerate the use cases for it.

CodeGains:
@epicfaace 

Ya it got a little complex, but I tried to make it so that the changes wouldn't impact other functionality. That is the main reason why `toNameSchema` is its own thing, rather than trying to modify some existing thing like `toIdSchema`. So the complexity is all contained within this feature since nothing uses the `nameSchema`. As a plus, it could be nice to have for other features in the future.

You are correct about the `nameSchema`. It will have each array element listed separately so that dependencies in array elements are handled independently. For example:

One array element could have a dependency met and data in the corresponding dependent field. Another element might have the same dependency not met and also data in the dependent field. Data in the dependent field should only be omitted in the second element.

There's lots of tests for `toNameSchema` in the utils test. Do you think there should be more?

epicfaace:
Oh, sorry, forgot to expand the utils.js file (I remember seeing the tests before and wondered where it went this time haha)

src/utils.js
CodeGains: Just updated this comment because I was thinking about it I changed my mind and don't think it would be a good idea to differentiate array indexes from object properties named like "0", "1", "2", etc. because they are handled the same way when using lodash `pick`. In the future, if someone needed to use the name/path schema for something else and wanted to know if a number path was an array index or property, then they could check it against the schema or form data.

PR #1292, opened 2019-05-21
Default type to `object` if undefined

### LINKED ISSUE #1265: Default type to `object` if undefined
### Prerequisites

### Description
Can we default the `type` to `object` if it's undefined? We accept user submitted JSON Schemas and this is a common thing that people miss. I'm not 100% sure if it's valid JSON Schema or not, but https://www.jsonschemavalidator.net/ appears to say the following is valid:

```json
{
  "properties": {
    "a": {
      "type": "string"
    }
  }
}
```

But when passing that into this module, we get the following error:

image

Playground link: https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJwcm9wZXJ0aWVzIjp7ImEiOnsidHlwZSI6InN0cmluZyJ9fX0sInVpU2NoZW1hIjp7fX0=

### Steps to Reproduce
1. Pass in a JSON Schema without a `type`
2. Observe that the [UnsupportedField](https://github.com/mozilla-services/react-jsonschema-form/blob/2d63594e075bba3b134ef09709c773a9ea037923/src/components/fields/UnsupportedField.js) is instantiated

#### Expected behavior
Default `type` to `object` if it is undefined
If `type` is set to something that is unknown, still create the `UnsupportedField` e.g. `type: 'someothertype'`

Affected blocks of code:
Could check for undefined in here and return `object`: https://github.com/mozilla-services/react-jsonschema-form/blob/2d63594e075bba3b134ef09709c773a9ea037923/src/utils.js#L68-L84

This is where it will bubble up to, but I don't think this code would have to change at all: https://github.com/mozilla-services/react-jsonschema-form/blob/2d63594e075bba3b134ef09709c773a9ea037923/src/components/fields/SchemaField.js#L49-L59

Happy to send over a PR for this if it'll get accepted! 

### Version
v1.5.0

domharrington:
Fixes #1265

Added tests and confirmed working on the local test app:
http://localhost:8080/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJwcm9wZXJ0aWVzIjp7ImEiOnsidHlwZSI6InN0cmluZyJ9fX0sInVpU2NoZW1hIjp7fX0=

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -76,6 +76,10 @@ export function getSchemaType(schema) {
     return "string";
   }
 
+  if ((!type && schema.properties) || schema.additionalProperties) {
+    return "object";
+  }
+
   if (type instanceof Array && type.length === 2 && type.includes("null")) {
     return type.find(type => type !== "null");
   }
--- a/src/utils.js
+++ b/src/utils.js
@@ -76,7 +76,7 @@ export function getSchemaType(schema) {
     return "string";
   }
 
-  if ((!type && schema.properties) || schema.additionalProperties) {
+  if (!type && (schema.properties || schema.additionalProperties)) {
     return "object";
   }
 
--- a/test/utils_test.js
+++ b/test/utils_test.js
@@ -1544,11 +1544,22 @@ describe("utils", () => {
         schema: { type: ["integer", "null"] },
         expected: "integer",
       },
+      {
+        schema: { properties: {} },
+        expected: "object",
+      },
+      {
+        schema: { additionalProperties: {} },
+        expected: "object",
+      },
     ];
 
     it("should correctly guess the type of a schema", () => {
       for (const test of cases) {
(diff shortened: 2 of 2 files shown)


src/utils.js
```diff
@@ -76,6 +76,10 @@ export function getSchemaType(schema) {
     return "string";
   }
 
+  if ((!type && schema.properties) || schema.additionalProperties) {
```
epicfaace: Is this necessary? Could we just make it

```suggestion
  if (schema.properties || schema.additionalProperties) {
```
Or does it need to be
```suggestion
  if (!type && (schema.properties || schema.additionalProperties)) {
```

domharrington: Agree the latter is better. Making that change!

epicfaace:
Thanks!

PR #1301, opened 2019-05-29
Pass along label/placeholder/registry for custom multiselect widgets.

fsteger:

Inconsistency between properties passed to custom select widgets. When widget is for a question of type string with an enum, the custom widget receives props for label, placeholder, and registry. When widget is used on a multi-select field, these properties are missing.

Sample Schema:
```
{
	type: 'object',
	properties: {
		selectQuestion: {
			title: 'Select Question',
			type: 'string',
			enum: ['foo', 'bar', 'baz', 'qux'],
			enumNames: ['foo', 'bar', 'baz', 'qux']
		},
		multiSelectQuestion: {
			title: 'Multi-Select Question',
			type: 'array',
			uniqueItems: true,
			items: {
				type: 'string',
				enum: ['foo', 'bar', 'baz', 'qux'],
				enumNames: ['foo', 'bar', 'baz', 'qux']
			}
		}
}
```

Sample UI Schema:
```
{
	selectQuestion: {
		'ui:widget': 'select'
	},
	multiSelectQuestion: {
		'ui:widget': 'select'
	}
}
```

### Checklist

  - [X I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -422,6 +422,8 @@ class ArrayField extends Component {
       disabled,
       readonly,
       required,
+      label,
+      placeholder,
       autofocus,
       onBlur,
       onFocus,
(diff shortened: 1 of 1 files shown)


epicfaace:
Thanks! The default MultiSelectWidget doesn't do anything with label/placeholder though -- do you think there is something we should do to change that?

PR #1302, opened 2019-05-29
Allowing AnyOfField or OneOfField to use schema

SlashBin:

I have a use case where I'd like to be able to define an anyOf in the schema (for validation purposes), but still render items from the schema properties. This should be allowable if I create my own AnyOfField, but currently SchemaField doesn't pass through the schema to the custom Field.

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -373,6 +373,7 @@ function SchemaFieldRender(props) {
           baseType={schema.type}
           registry={registry}
           safeRenderCompletion={props.safeRenderCompletion}
+          schema={schema}
           uiSchema={uiSchema}
         />
       )}
@@ -391,6 +392,7 @@ function SchemaFieldRender(props) {
           baseType={schema.type}
           registry={registry}
           safeRenderCompletion={props.safeRenderCompletion}
+          schema={schema}
           uiSchema={uiSchema}
         />
       )}


epicfaace:
hi @slashbin, can you show me an example of AnyofField/OneofField using a schema, and add some tests, so that I can see what exactly is the use case for this?

epicfaace:
Ok, took a look again and this makes sense. Thanks for the PR!

SlashBin:
Thanks for merging! Sorry I didn't see your reply earlier. 

Just for posterity's sake, the use case here would be that I'd like to render an `ObjectField` from the `properties`, as normal, and rely on the `anyOf` only for validation on a subset of the `properties`.

So, as a simple (untested) example, I could have something like the following, which I believe is valid JSON Schema and AJV seems fine with. However, it requires that I render all properties and validate on the `anyOf` conditions. I understand that isn't necessarily supported out of the box yet, but we shouldn't preclude someone from creating their own `AnyOfField` that does this (by omitting `schema`, as `SchemaField` was doing)

```
{
	"type": "object",
	"properties": {
		"is_exempt": {
			"type": "boolean"
		},
		"overtime_hours": {
			"type": "array",
			"items": {
				"type": "number"
			}
		}

	},
	"anyOf": [{
		"properties": {
			"is_exempt": {
				"const": true
			}
		}
	}, {
		"properties": {
			"overtime_hours": {
				"minItems": 1
			}
		}
	}]
}
```

Issue #998, opened 2018-08-03
Schema Dependencies Merged Incorrectly


arahansen:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

When merging schema with dependencies, enums are appended instead of replaced.

### Steps to Reproduce

[...]

PR #1306, opened 2019-06-03
Add test to concat required dependencies

epicfaace:

Add a test to demonstrate the use of concatArrays in #998.


edi9999:
Looks good to me, I'm merging since it fixes the `.only` and is a quite simple MR

PR #1312, opened 2019-06-07
Feature/1290 sort errors based on uischema

an-aessid:

As discussed here (https://github.com/mozilla-services/react-jsonschema-form/issues/1290) it would be nice to have the errors sorted based on the uiSchemas ui:order properties.
The reason was that for cases where the formSchema is sent by a backend over HTTP(s), the ordering of the formSchema in the payload can differ and will result in a "randomly" ordered error message list.

I added the sorting functionality (as it was suggested) to the `validate.js` file and created a `validate_errorSorting_test.js` test-suite because all the test cases are consuming quite a lot of space. 
Basically I'm flattening the `_errorSchema` to retrieve an easily accessible object containing all errors for a specific property. Then I'm recursively going through the `uiSchemas` `ui:order` properties (where they are given...), sorting in the matching errors.  

! The solution I guess can be optimized in speed, but I would like to not pre-maturely optimise things.

### Checklist

I'll go through the checklist and do actions as required.

  - [x] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed (I've added something to the "Object fields ordering" section in form-customization.md)


Gurkenschreck:
Question1: 
HELP NEEDED:
There is one test case in the test suite for the arrays
* error sorting -> with arrays -> containing nested objects -> "should have the order foo - wibble - plugh - garply - grault - wobble - bar"

There are 18 errors but in my calculation there should only be 14. There are errors given by ajv like.

Question 2:
In the checklist it states the following. Is this a new feature? I'll update the "errors" playground example to include the ui:orders.
I'm adding a new feature
    I've updated the playground with an example use of the feature

src/validate.js
```diff
@@ -159,6 +160,449 @@ function transformAjvErrors(errors = []) {
...
+ *     ".bar[n].qux[4].wibble": [* errors like: (a) *],
+ *     "": [* errors like: does-not-match-oneof-schema *]
+ * }
+ *
+ * @param {object} _target - the object to add the key to.
+ * @param {object[]} _errors - the errors of ajv
+ * @param {object} _errorSchema - the error schema constructed by the `toErrorSchema` function
+ * @param {string} _currentPath - the path of the current nested object
+ */
+function flattenErrorSchema(_target, _errors, _errorSchema, _currentPath = "") {
```
Gurkenschreck: It's kind of complicated, it's one of the first things I added :disappointed:. 

But I didn't want to introduce new library dependencies like https://www.npmjs.com/package/flat to add to the bundle size, but If you say I can add it, I'd check it out if it's usable.

src/validate.js
```diff
@@ -159,6 +160,308 @@ function transformAjvErrors(errors = []) {
...
+ * @return {Array}
+ */
+function orderErrorsByUiSchema(_uiSchema, _errors, _errorSchema) {
+  if (!Array.isArray(_errors) || _errors.length === 0) {
+    return [];
+  }
+  console.log(`in`, _errorSchema);
+  const flattenedErrorSchema = {};
+  flattenErrorSchema(flattenedErrorSchema, _errors, _errorSchema);
+  console.log(`out`, flattenedErrorSchema);
```
Gurkenschreck: My added `Question 3` can be checked by adding the `.only` to the suite and looking at these loggings.

src/validate.js
epicfaace: Yeah, if it's possible to use a library to do this, I'd say simplifying the complexity is worth the increase in bundle size.

Gurkenschreck: I tried using this lib but I cannot easily use it because it does not care about arrays and this function also flattens the error schema path depending on the given errors.

I would like to be able to distinguish between `foo[0].bar` and `foo.0.bar`.

epicfaace: Why do you need to be able to distinguish between foo[0].bar and foo.0.bar?

Gurkenschreck: Good question. I believe I did it like this because it's easier to figure out that `bar` is a property to look for in the uiSchema, and not `0`. 

ATM when it's foo[0].bar I know that foo is an array and "bar" would appear as it's own property in the matching uiSchema path, so I can simply access that property. When it's foo.0.bar I know foo is an object, with a property that is actually named `0` so I need to look that up in the uiSchema.

If it always would be flattened like foo.0.bar I would not have this info. Then I need to look up `0` and if it is not given, `bar` in the uiSchema.

Does this make any sense?

Gurkenschreck: I realized I can achieve the same thing by iterating over all errors and combining all errors for one faulty schema property (`error.property`).

Should be much easier now!

epicfaace:
> There are 18 errors given but in my calculation there should only be 14. There are errors given by ajv like.
> 
> ```
> { name: 'type',
>     property: '.foo[0].plugh[0]',
>     message: 'should be array',
>     params: { type: 'array' },
>     stack: '.foo[0].plugh[0] should be array',
>     schemaPath: '#/properties/foo/items/properties/plugh/items/type' }
> ```
> 
> which I don't understand since they are arrays in the `formData`. Does anybody know what's wrong?

@Gurkenschreck You need to change this line from `type: "array"` to `type: "object"`. Otherwise ajv is thinking that `foo` is an array of arrays, not an array of objects. https://github.com/mozilla-services/react-jsonschema-form/pull/1312/files#diff-1852564c7626956ec1edd2e5419a5a04R595

Gurkenschreck:
> @Gurkenschreck You need to change this line from `type: "array"` to `type: "object"`. Otherwise ajv is thinking that `foo` is an array of arrays, not an array of objects. https://github.com/mozilla-services/react-jsonschema-form/pull/1312/files#diff-1852564c7626956ec1edd2e5419a5a04R595

Ah, of course! I guess I totally oversaw this thing.

fsteger:
If there is not a ui:order defined, could this fall back to follow the order of the properties in the schema?

epicfaace:
How would such an order be defined? Are you thinking of ownPropertyKeys?

fsteger:
I was thinking something along the lines of the `orderProperties` function inside of `utils.js`

Gurkenschreck:
> If there is not a ui:order defined, could this fall back to follow the order of the properties in the schema?

From my understanding, this is what happens. Depending on how the schema is written, `ajv` seems to generate a list of errors based on the schemas properties order.
If there are no `ui:order`s defined at all or for only a part of the uiSchema, `orderErrorsByUiSchema` will simply take the ajv order for the errors that have no `ui:order` and put them in place.

fsteger:
> > If there is not a ui:order defined, could this fall back to follow the order of the properties in the schema?
> 
> From my understanding, this is what happens. Depending on how the schema is written, `ajv` seems to generate a list of errors based on the schemas properties order.
> If there are no `ui:order`s defined at all or for only a part of the uiSchema, `orderErrorsByUiSchema` will simply take the ajv order for the errors that have no `ui:order` and put them in place.

Sounds good; thanks for clarifying.

fsteger:
@Gurkenschreck I took a look at the recent build failures, and it looks like running on node v6/v7 is choking on some of the spacing inside `validate_errorSorting_test.js`. I was able to get `npm run cs-check` to successfully complete locally on both v6 and v7 with the following changes:
```
          throw new Error(
            `The amount of sored errors (${
              sortedErrors.length
            }) does not match the expected ones (${expectedOrder.length})`
          );
```
to
```
          throw new Error(
            `The amount of sored errors (${sortedErrors.length}) does not match the expected ones (${expectedOrder.length})`
          );
```

And changing
```
        it(
          "should have the order in which the properties appear in the schema (foo.wibble - foo.wobble - foo - bar.wibble - bar.wobble - bar -" +
            " oneOfError) with two oneOfs",
          () => {
```
to
```
        it("should have the order in which the properties appear in the schema (foo.wibble - foo.wobble - foo - bar.wibble - bar.wobble - bar - oneOfError) with two oneOfs", () => {
```
I also fixed the indentation on that function.

Gurkenschreck:
It's working now after merging rjsforms origin/master, thanks!

src/validate.js
```diff
@@ -159,6 +160,368 @@ function transformAjvErrors(errors = []) {
...
+ * [
+ *    {property: '.foo', ...errorForThisProp},
+ *    {property: '.bar', ...errorForThisProp},
+ *    {property: '.bar[2]', ...errorForThisProp},
+ *    {property: '.bar[2].qux', ...errorForThisProp},
+ *    {property: '.bar', ...errorForThisProp}
+ *    {property: '', ...errorForThisProp}
+ * ]
+ * // sample output
+ * // different cases possible of an error.properties; here are some EXAMPLES
```
epicfaace:
```suggestion
 * // different cases possible for error.properties; here are some examples
```

test/validate_errorSorting_test.js
```diff
@@ -0,0 +1,1600 @@
...
+
+import validateFormData from "../src/validate";
+
+describe("Validation Error Sorting", () => {
+  describe("validate.validateFormData()", () => {
+    describe("error sorting", () => {
+      const expectOrder = (sortedErrors, expectedOrder) => {
+        if (sortedErrors.length !== expectedOrder.length) {
+          throw new Error(
+            `The amount of sored errors (${
```
epicfaace:
```suggestion
            `The number of sorted errors (${
```

src/validate.js
```diff
@@ -159,6 +160,368 @@ function transformAjvErrors(errors = []) {
...
+    );
+  });
+  return flattenedErrorSchema;
+}
+
+function removeDotPrefixFromString(string) {
+  return string.startsWith(".") ? string.substring(1) : string;
+}
+
+function addErrorToArrayIfAbsent(array) {
```
epicfaace: Hey @Gurkenschreck , thanks for the change. I haven't gotten a chance to step through all your logic completely yet, but given that you've added around 12 different helper functions, I want to make sure that when adding this much complexity, we offset this complexity by adding unit tests. You do have tests for the uiSchema ordering -- which is great -- but could you also look into adding unit tests for some of the more important helper functions? That will help a ton, both in reviewing the code to see what each function is supposed to do, and in also changing / maintaining this logic in the future.

Gurkenschreck:
I'll check the changes soon.

src/validate.js
Gurkenschreck: Changed this.

test/validate_errorSorting_test.js
Gurkenschreck: Changed this.

src/validate.js
Gurkenschreck: Hi,
I see your point and yes, it would help. But I'd want to avoid exporting those functions, making the helper functions public, just for the sake of testing (for the same reasons as one would not make every method public in e.g. Java just to test it).
Would it be enough to think about more test cases that I can add to test the (now renamed) `sortErrorsByUiSchema` function?
If not, I don't wanna press this paradigm on this project, so I can export those functions and test them. What do you think?

Thanks for the feedback!

epicfaace: I see your point, but I think it's ok to export them because we do want to make sure each helper function is working as intended. Also, it's not super clear to me what functions like `filterErrorsByUnmentionedUiorder` and `recursivelyAddErrorsToArrayForNode` exactly do (unless I step through the code exactly), so adding maybe 1-2 tests for each of these functions would help to document exactly what each of these functions is supposed to do.

The tests may not need to be exhaustive, given that you already have many tests for `sortErrorsByUiSchema`.

PR #1313, opened 2019-06-07
fix: populate defaults for nested dependencies when formData is undefined

### LINKED ISSUE #1311: Release 1.6.0 editor crashes when dependencies are used
### Prerequisites

### Description

With the latest release 1.6.0 we now see an issue for all fields where we use dependencies. The issue causes the editor to not render at all. In our own build we see the console.log stating that the dependency-based fields are not defined.

Everything works in 1.5.0 - but we're eager to shift to 1.6.0 to enable default values in anyOf dependencies, as listed in issue #1293 

### Steps to Reproduce
For us, all fields that have dependencies now cause the editor to break - incl. the playground sample shared above (you can verify this by copying the Schema into a new playground and observe that it does not render):
[example (worked in 1.5.0, but not 1.6.0)
](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJydGMiOnsic3luYyI6MH19LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsicnRjIjp7InRpdGxlIjoiQ29uZmlndXJlIENsb2NrIiwidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsic3luYyI6eyJ0aXRsZSI6IlRpbWUgU3luYyBUeXBlIiwidHlwZSI6ImludGVnZXIiLCJkZWZhdWx0IjowLCJhbnlPZiI6W3sidGl0bGUiOiJSZXRhaW4gdGltZSIsInR5cGUiOiJpbnRlZ2VyIiwiZW51bSI6WzBdfSx7InRpdGxlIjoiQXV0byB0aW1lIiwidHlwZSI6ImludGVnZXIiLCJlbnVtIjpbMV19XX19LCJkZXBlbmRlbmNpZXMiOnsic3luYyI6eyJvbmVPZiI6W3sicHJvcGVydGllcyI6eyJzeW5jIjp7ImVudW0iOlswXX19fSx7InByb3BlcnRpZXMiOnsic3luYyI6eyJlbnVtIjpbMV19LCJ0b2xlcmFuY2UiOnsidGl0bGUiOiJUaW1lIHN5bmMgdG9sZXJhbmNlIiwidHlwZSI6ImludGVnZXIiLCJtaW5pbXVtIjoxLCJtYXhpbXVtIjozNjAwLCJkZWZhdWx0IjozMH19fV19fX19fSwidWlTY2hlbWEiOnt9fQ==)

In contrast, the same example will still render if I completely remove the "dependencies" tab - but of course that's probably not the intention. I assume some issue has arisen during the implementation of a fix for default values.
[same example after removing dependencies (works in 1.5.0 and 1.6.0)](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJydGMiOnsic3luYyI6MH19LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsicnRjIjp7InRpdGxlIjoiQ29uZmlndXJlIENsb2NrIiwidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsic3luYyI6eyJ0aXRsZSI6IlRpbWUgU3luYyBUeXBlIiwidHlwZSI6ImludGVnZXIiLCJkZWZhdWx0IjowLCJhbnlPZiI6W3sidGl0bGUiOiJSZXRhaW4gdGltZSIsInR5cGUiOiJpbnRlZ2VyIiwiZW51bSI6WzBdfSx7InRpdGxlIjoiQXV0byB0aW1lIiwidHlwZSI6ImludGVnZXIiLCJlbnVtIjpbMV19XX19fX19LCJ1aVNjaGVtYSI6e319)

Martin

epicfaace:

Fixes #1311. It looks like the problem is that formData is not defaulted to `{}` when populating defaults for dependencies, so then when formData is undefined / null, it crashes.

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -144,7 +144,13 @@ export function hasWidget(schema, widget, registeredWidgets = {}) {
   }
 }
 
-function computeDefaults(schema, parentDefaults, definitions, formData = {}) {
+function computeDefaults(
+  schema,
+  parentDefaults,
+  definitions,
+  rawFormData = {}
+) {
+  const formData = isObject(rawFormData) ? rawFormData : {};
   // Compute the defaults recursively: give highest priority to deepest nodes.
   let defaults = parentDefaults;
   if (isObject(defaults) && isObject(schema.default)) {
--- a/src/utils.js
+++ b/src/utils.js
@@ -144,7 +144,7 @@ export function hasWidget(schema, widget, registeredWidgets = {}) {
   }
 }
 
-function computeDefaults(schema, parentDefaults, definitions, formData) {
+function computeDefaults(schema, parentDefaults, definitions, formData = {}) {
   // Compute the defaults recursively: give highest priority to deepest nodes.
   let defaults = parentDefaults;
   if (isObject(defaults) && isObject(schema.default)) {
--- a/test/utils_test.js
+++ b/test/utils_test.js
@@ -615,7 +615,7 @@ describe("utils", () => {
         });
       });
 
-      it("should populate defaults for nested dependencies when formData is undefined", () => {
+      it("should populate defaults for nested dependencies when formData passed to computeDefaults is undefined", () => {
         const schema = {
           type: "object",
           properties: {
@@ -661,7 +661,7 @@ describe("utils", () => {
         });
       });
 
-      it("should populate defaults for nested dependencies when formData is null", () => {
+      it("should not crash for defaults for nested dependencies when formData passed to computeDefaults is null", () => {
         const schema = {
           type: "object",
           properties: {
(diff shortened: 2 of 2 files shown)


jrose-carecloud:
If formData is `null` this wont work as it wont default to the empty object

epicfaace:
@jrose-carecloud I added a test with `null` and it seems to work; was there something specifically that didn't work when you tried it?

jrose-carecloud:
I'l admit that I didnt run the test, more of a just glancing sort of thing. But argument defaults only occur when the argument is undefined, not null.
Consider:

```js
function foo(bar = {}) {
   console.log(bar);
}
```

If called like this:

```js
foo();
``` 
We will log out `{}` as expected. But if called like this:

```js
foo(null)
```
We will log out, `null`.

If `null` is not a possible/acceptable value to be received here then please disregard. I'm just keeping a close eye on this issue as its effecting our app at the moment.

epicfaace:
That's a good point, thanks. I don't think my `null` test accounted for that case, because it was still probably passing `undefined` down to the actual `computeDefaults` function. I'll fix this PR accordingly.

PR #1322, opened 2019-06-17
Upgrade to Babel 7

ankurkaushal:

Using my work account (@ankurkaushal360), I opened an issue here, #1319 mentioning `babel` upgrade & I was told to it would be a good idea.

So this pull upgrades the Babel 6 to Babel 7 using `npx babel-upgrade --write` and does the following things:

- Removes deprecated `babel-plugin-react-transform` with `@babel/plugin-transform-react-jsx`
- Updates a few files with `default` exports as they were throwing warnings while running `dev` like this:

image

As far as I can tell, everything should work as it is. Even all the tests are passing too. I don't think I need to add / update any tests for this change. 

Also, I haven't updated the `package.json` version yet. Though I think I need to update it. Would you prefer `major` or `minor` version bump?

### Checklist


.babelrc
```diff
@@ -1,20 +1,22 @@
...
   "plugins": [
-    "transform-object-rest-spread",
-    "transform-class-properties",
-    "transform-runtime"
+    "@babel/plugin-proposal-object-rest-spread",
+    "@babel/plugin-proposal-class-properties",
+    [
+      "@babel/plugin-transform-runtime",
+      {
+        "corejs": 2
```
epicfaace: Is there a reason why you chose 2 to be the value for corejs?

ankurkaushal: I think it was `npx babel-upgrade --write` that did it. And we're using `core-js` v2 right, so it should be that version right?

ankurkaushal:
@epicfaace Marked this PR as 'ready for review' :)

epicfaace:
Thanks!

PR #1330, opened 2019-06-21
pass formContext to SchemaField

ealtunyay:

1. [Documentation states, "Props passed to a custom SchemaField are the same as the ones passed to a custom field.", but formContext is not passed](https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/#custom-schemafield)
2. Use case: Make context aware modifications to errors in the ui. These are the options I considered:
    1. Modify my json-schema. Not ideal. These changes have side-effects on other business related things.
    1. Handle state myself and pass formData, but formData is only for form values.
    1. Use a custom field template. However, errors are already rendered by the time the template render is reached, and you cannot rerender them using the default error render function located in SchemaField.
    2. Use custom fields, but custom fields do not handle errors.
    1. Use a custom SchemaField. Close, but no formContext. If it was passed one could theoretically make error related changes and pass everything along to the default SchemaField component.
3. Number of open issues seemed high (224). I thought this might be better :)

1. [Not really related, but searching through issues before forking I found this issue that expressed a similar use case but for custom fields](https://github.com/mozilla-services/react-jsonschema-form/issues/925). Seems like it was answered.

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -220,6 +220,7 @@ export default class Form extends Component {
       acceptcharset,
       noHtml5Validate,
       disabled,
+      formContext,
     } = this.props;
 
     const { schema, uiSchema, formData, errorSchema, idSchema } = this.state;
@@ -249,6 +250,7 @@ export default class Form extends Component {
           errorSchema={errorSchema}
           idSchema={idSchema}
           idPrefix={idPrefix}
+          formContext={formContext}
           formData={formData}
           onChange={this.onChange}
           onBlur={this.onBlur}


ealtunyay:
After some further development I realize I missed some things:
1. My change was insufficient to propogate `formContext` fully through the tree. Further plumbing would be necessary.
2. `formContext` is passed to custom SchemaFields through the `registry` object making my change redundant.

Perhaps the docs could be a little clearer? The "Custom SchemaField" section and "The `registry` Object" sections are close to each other, but the quote I referenced above ("Props passed to a custom SchemaField are the same as the ones passed to a custom field.") is still misleading. Up until this point I considered the registry object a place to register custom components. I didn't expect to find formContext in there, and so I didn't read it carefully enough.

Only after working through my usecase did I stumble across formContext inside the registry. Only in hindsight do I see that the section on the registry object is clear: "The registry is passed down the component tree, so you can access it from your custom field and SchemaField components.", and `formContext` is clearly on the list of properties above it.

I'll leave this open for now in case anyone cares to comment or suggest making a change in the documentation.

epicfaace:
It looks like we have a few things here:
- `formContext` is passed down to all fields except for `SchemaField`
- `registry` is passed down to all fields except for `TitleField` and `DescriptionField`

For consistency's sake, I think it would be best to pass `formContext` to `SchemaField`. In a later version where we can make breaking changes it might be best to just pass down `registry` to all fields, not `formContext`.

PR #1332, opened 2019-06-23
[WIP] Live omit

epicfaace:

Add a `liveOmit` option. I have a use case in which I want omitExtraData to happen during onChange, but I do not want liveValidate.

### Checklist

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -118,7 +118,7 @@ render((
 ), document.getElementById("app"));
 ```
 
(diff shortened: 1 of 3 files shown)


epicfaace:
@CodeGains it turned out I had to include this for my use case -- if we end up having a `liveOmit` property, do you think it's still necessary to automatically live omit when `liveValidate` is true? Or would that end up complicating things and might it be better just to only live omit when `liveOmit` is true (regardless of the value of `liveValidate`)?

CodeGains:
@epicfaace I think it makes sense to only live omit when `liveOmit` is set so we don't force the live omit functionality on a user that just wants to `liveValidate`.

epicfaace:
Thanks -- good point, will do that.

PR #1335, opened 2019-06-25
Array items now have unique, stable keys (#1046)

### LINKED ISSUE #1046: Array item keys should not be based on array index
### Prerequisites

### Description

In arrays, the key is currently based on the array index, see https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/ArrayField.js#L49. 

This is not recommended and can give rise to a number of issues, see https://reactjs.org/docs/lists-and-keys.html

In particular this:
image

### Steps to Reproduce

Run https://jsfiddle.net/wt4hkpy2/, and try to remove 'item 1' or 'item 2'. The last item in the array gets removed.

This is because before removal there are 3 items with in index 0, 1 and 2. After removal there are two items with index 0 and 1. Thus the react reconciler infers that the item with index 2 must have been removed.

(I realise that by making the input controlled in the above example the problem disappears, but I purposely used the 'props to state' antipattern as it is the smallest way I could think of to demonstrate the key issue)

#### Expected behavior

The correct array item is removed

#### Actual behavior

The last array item is removed

### Version

1.0.2, but problem exists in current master as well

### LINKED ISSUE #1333: Removing or reordering an ArrayField breaks app
### Prerequisites

### Description
I've created a jsfiddle here: https://jsfiddle.net/louisbyrne/q6yghz04/34/

Issue 1: When removing or reordering the "list 1" item in my jsfiddle (which uses the ArrayField component) I recieve the following console error:
`ArrayField.js:370 Uncaught TypeError: n.map is not a function
    at t.value (ArrayField.js:370)
    at t.value (ArrayField.js:343)
    at p._renderValidatedComponentWithoutOwnerOrContext (react.min.js:13)
    at p._renderValidatedComponent (react.min.js:13)
    at p._updateRenderedComponent (react.min.js:13)
    at p._performComponentUpdate (react.min.js:13)
    at p.updateComponent (react.min.js:13)
    at p.receiveComponent (react.min.js:13)
    at Object.receiveComponent (react.min.js:15)
    at Object.updateChildren (react.min.js:13)`

This is because, the following item, which is a string field, now uses the ArrayField component. As the default value is a string and not an array, it breaks the component when it tries to map over the default value. 

I've created a second jsfiddle here: https://jsfiddle.net/louisbyrne/ed7jvx01/2/
Delete the "string 1" entry. Now all other entries are using the enum field's widget (two should still be using the string fields widget

It seems that the fields components are not correctly updating when the formData array items are being updated.

I feel this may be related to #1046 

#### Expected behavior
Fields components correctly update when formData changes

#### Actual behavior
When updating an array, the components are not being updated correctly, resulting in the app breaking due to incorrect data being passed into components.

### Version
1.6.1

fsteger:

Array items did not have stable, unique keys and instead had to rely on the index of the item.
Fixes #1046
Fixes #1333

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -46,10 +46,7 @@ function DefaultArrayItem(props) {
     fontWeight: "bold",
   };
   return (
-    <div
-      key={props.key}
-      id={`array-item-${props.key}`}
-      className={props.className}>
+    <div key={props.key} className={props.className}>
       <div className={props.hasToolbar ? "col-xs-9" : "col-xs-12"}>
         {props.children}
       </div>
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -1,6 +1,7 @@
 import AddButton from "../AddButton";
 import IconButton from "../IconButton";
 import React, { Component } from "react";
+import { polyfill } from "react-lifecycles-compat";
 import includes from "core-js/library/fn/array/includes";
 import * as types from "../../types";
 
@@ -217,9 +218,9 @@ class ArrayField extends Component {
     };
   }
 
-  componentWillReceiveProps(nextProps) {
+  static getDerivedStateFromProps(nextProps, prevState) {
     const nextFormData = nextProps.formData;
-    const previousKeyedFormData = this.state.keyedFormData;
+    const previousKeyedFormData = prevState.keyedFormData;
     const newKeyedFormData =
       nextFormData.length === previousKeyedFormData.length
         ? previousKeyedFormData.map((previousKeyedFormDatum, index) => {
@@ -229,29 +230,10 @@ class ArrayField extends Component {
             };
           })
         : generateKeyedFormData(nextFormData);
-    this.setState({
-      keyedFormData: newKeyedFormData,
-    });
-  }
-
-  /*
-  // React 16.3 replaces componentWillReceiveProps with getDerivedStateFromProps
-  //
-  static getDerivedStateFromProps(nextProps, prevState) {
-    const nextFormData = nextProps.formData;
-    const previousKeyedFormData = prevState.keyedFormData;
-    const newKeyedFormData = (nextFormData.length === previousKeyedFormData.length) ?
-      previousKeyedFormData.map((previousKeyedFormDatum, index) => {
-        return {
-          key: previousKeyedFormDatum.key,
-          item: nextFormData[index]
-        };
-      }) : generateKeyedFormData(nextFormData);
     return {
-      keyedFormData: newKeyedFormData
+      keyedFormData: newKeyedFormData,
     };
   }
-  */
 
   get itemTitle() {
     const { schema } = this.props;
@@ -770,5 +752,7 @@ class ArrayField extends Component {
 if (process.env.NODE_ENV !== "production") {
   ArrayField.propTypes = types.fieldProps;
 }
+
+polyfill(ArrayField);
 
 export default ArrayField;
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -45,7 +45,10 @@ function DefaultArrayItem(props) {
     fontWeight: "bold",
   };
   return (
-    <div key={props.index} className={props.className}>
+    <div
+      key={props.key}
+      id={`array-item-${props.key}`}
+      className={props.className}>
       <div className={props.hasToolbar ? "col-xs-9" : "col-xs-12"}>
         {props.children}
       </div>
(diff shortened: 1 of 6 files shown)


fsteger:
When using componentWillReceiveProps, the incoming props may be updated via onChange before the setState call completes. Latest change moves onChange to setState callback to ensure it's updated first.

This didn't appear to be an issue with getDerivedStateFromProps.

src/components/fields/ArrayField.js
```diff
@@ -185,6 +205,51 @@ class ArrayField extends Component {
...
+        : generateKeyedFormData(nextFormData);
+    this.setState({
+      keyedFormData: newKeyedFormData,
+    });
+  }
+
+  /*
+  // React 16.3 replaces componentWillReceiveProps with getDerivedStateFromProps
+  //
+  static getDerivedStateFromProps(nextProps, prevState) {
```
epicfaace: I think it might be cleaner (and easier to migrate to React 16) just to use getDerivedStateFromProps and https://github.com/reactjs/react-lifecycles-compat for now?

src/components/fields/ArrayField.js
```diff
@@ -185,6 +205,51 @@ class ArrayField extends Component {
...
+    const previousKeyedFormData = this.state.keyedFormData;
+    const newKeyedFormData =
+      nextFormData.length === previousKeyedFormData.length
+        ? previousKeyedFormData.map((previousKeyedFormDatum, index) => {
+            return {
+              key: previousKeyedFormDatum.key,
+              item: nextFormData[index],
+            };
+          })
+        : generateKeyedFormData(nextFormData);
```
epicfaace: Won't this reset _all_ the keys when an array element is added? Do you think that's ok behavior?

fsteger: If insertion is done via onAddClick, the newly added entry will already be in the state with an assigned key. If insertion is done externally, then it is likely safer to reassign keys as there's no way to reconcile any other possible changes in the form data with that in the state.

epicfaace: Makes sense, thanks!

epicfaace:
@fsteger do you think there is any way to pass on this benefit to those who make their own `ArrayFieldTemplate`s? As of now, they will only get this benefit if they use the default template.

fsteger:
> @fsteger do you think there is any way to pass on this benefit to those who make their own `ArrayFieldTemplate`s? As of now, they will only get this benefit if they use the default template.

Yeah; that was my intention. Each item in the array of `props.items` in `ArrayFieldTemplate` should now have access to the key property. I had missed this initially for fixed arrays, but it should consistently be present now.

I updated the `customArray.js` in the playground to use the key, and updated the "should move from first to last in the list" test (`ArrayField_test.js`), which uses a custom ArrayFieldTemplate, to include the keys as well.

src/components/fields/ArrayField.js
fsteger: Done in latest commit.

src/components/fields/ArrayField.js
```diff
@@ -44,7 +46,10 @@ function DefaultArrayItem(props) {
     fontWeight: "bold",
   };
   return (
-    <div key={props.index} className={props.className}>
+    <div
+      key={props.key}
+      id={`array-item-${props.key}`}
```
epicfaace: I don't think there's any point in exposing the key to the user as the `id`? Currently ids are only used as per the idSchema (for example, `root_listOfStrings_0`) so having another function for id may also not be the best.

fsteger: What would you recommend here?

epicfaace: Can we just remove the id?

Or is there a particular reason why you wanted to add it?

fsteger: I'm using the id right now to ensure that the key does not change inside the ArrayField tests. If you remove the id, the tests for inserting new rows will fail. The re-order and delete tests only pass when the attribute is removed because the id both before and after the action are null, but these should really test to ensure an id exists as well.

If we remove the id, I'm not sure how these tests should be updated to verify the correct behavior as the key doesn't appear to be exposed anywhere else.

Regarding the `idSchema`, were you thinking of just passing the key into the `toIdSchema` method for each item, and then accessing it via `props.idSchema.key`?

epicfaace: Got it. I don't think we should add it to the idSchema or expose it as the id attribute, though, given that the array item ids have no meaning apart from being unique and stable (as opposed to the other ids in the idSchema).

Here are two possible solutions:
- Can we expose the key as `data-rjsf-id` or something instead?
- Or, we don't expose the key at all -- in our tests, create a custom ArrayFieldTemplate that exposes the key in the DOM so that the tests can access it.

fsteger: I removed the id attribute, and updated the tests to use a custom template.

epicfaace:
Great, thanks!

PR #1338, opened 2019-06-27
Fixed schema type not being deduced correctly in compute defaults #1334

t-moe:
# Reasons for making this change
As described in bug report #1334, the "default" values of schemas are not correctly applied to anyOf/oneOf properties.

This PR fixes that. 
anyOf/oneOf schema don't have 'type' properties sometimes, so we needed to use `getSchemaType(schema)` instead of `schema.type` in `computeDefaults`

### Checklist

  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed.

### DIFF
--- a/test/anyOf_test.js
+++ b/test/anyOf_test.js
@@ -85,6 +85,36 @@ describe("anyOf", () => {
     expect(comp.state.formData).eql({ foo: "defaultbar" });
   });
 
+  it("should assign a default value and set defaults on option change with 'type': 'object' missing", () => {
+    const { comp, node } = createFormComponent({
+      schema: {
+        type: "object",
+        anyOf: [
+          {
+            properties: {
+              foo: { type: "string", default: "defaultfoo" },
+            },
+          },
+          {
+            properties: {
+              foo: { type: "string", default: "defaultbar" },
+            },
+          },
+        ],
+      },
+    });
+
+    expect(comp.state.formData).eql({ foo: "defaultfoo" });
+
+    const $select = node.querySelector("select");
+
+    Simulate.change($select, {
+      target: { value: $select.options[1].value },
+    });
+
+    expect(comp.state.formData).eql({ foo: "defaultbar" });
+  });
+
   it("should render a custom widget", () => {
     const schema = {
       type: "object",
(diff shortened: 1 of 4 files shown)


epicfaace:
@t-moe thanks! Can you add a test to make sure this bug doesn't regress?

t-moe:
@epicfaace done

test/anyOf_test.js
```diff
@@ -85,6 +85,36 @@ describe("anyOf", () => {
     expect(comp.state.formData).eql({ foo: "defaultbar" });
   });
 
+  it("should assign a default value and set defaults on option change with a different schema", () => {
```
epicfaace:
```suggestion
  it("should assign a default value and set defaults on option change with 'type': 'object' missing", () => {
```

test/oneOf_test.js
```diff
@@ -85,6 +85,36 @@ describe("oneOf", () => {
     expect(comp.state.formData).eql({ foo: "defaultbar" });
   });
 
+  it("should assign a default value and set defaults on option change with a different schema", () => {
```
epicfaace:
```suggestion
  it("should assign a default value and set defaults on option change with 'type': 'object' missing", () => {
```

test/utils_test.js
```diff
@@ -359,6 +359,23 @@ describe("utils", () => {
         });
       });
 
+      it("should populate object defaults for oneOf", () => {
```
```suggestion
      it("should populate defaults for oneOf when 'type': 'object' is missing", () => {
```

goralight:
Is there an Eta for this to be merged into master?

epicfaace:
Nothing pending, I just forgot to check on this. Thanks for the reminder @goralight !

goralight:
@epicfaace No worries! Thank you for merging and thank you @t-moe for fixing the bug. Could we expect a release soon ?

MatinF:
@epicfaace As mentioned, I'm not sure this is yet fully resolved in the latest commit. Would be great to hear your thoughts on this. It looks like the default values are rendered in the editor, but in some cases these default values are not actually part of the submitted formData.

PR #1342, opened 2019-06-30
Replace react-addons-test-utils with react-dom/test-utils

epicfaace:
This is because react-addons-test-utils is deprecated and included within react-dom. We get this warning each time we run tests:

```
Warning: ReactTestUtils has been moved to react-dom/test-utils. Update references to remove this warning.
```


epicfaace:
@edi9999 can you review?

PR #1343, opened 2019-06-30
Upgrade packages such as eslint

epicfaace:
Upgrade packages such as eslint to fix some security vulnerabilities detected by npm.

In the process of upgrading eslint, had to change some code so that it confirmed with https://eslint.org/docs/rules/no-prototype-builtins. This has the side effect that, for example, a schema can now have a property called `hasOwnProperty` and not break rjsf.


edi9999:
Hello @epicfaace , 


I would suggest to update to the latest eslint, and inside the travis runner, specify to use the latest version for the lint (as we want to support node 6 for end-users, not for developpers of rjsf).

That is what I do for another project of mine for about a year or more to continue testing on "older" nodejs versions but still be able to use eslint or tooling that works only on newer versions : https://github.com/open-xml-templating/docxtemplater/blob/master/.travis.yml

There is the magic command : `- nvm use $TRAVIS_NODE_VERSION` which allows me to use the base travis node version after doing the linting and running babel.

PR #1345, opened 2019-07-05
Allows for a custom tagName to be used in place of "form"

juliankigwana:

Sometimes we need a way to nest the RJSF components so that we can utilise the rendering capabilities within our components. For example, when rendering a custom `Array<Object>` component, the child properties can be rendered in a child form.  However doing this will produce invalid html, a nasty error in the console, and possibly cause unforeseen problems across browsers.

### Checklist

### DIFF
--- a/docs/form-customization.md
+++ b/docs/form-customization.md
@@ -722,3 +722,15 @@ It's possible to disable the whole form by setting the `disabled` prop. The `dis
 ```
 
 If you just want to disable some of the fields, see the [`ui:disabled`](#disabled-fields) parameter in the `uiSchema` directive. 
+
+
+### Changing the tag name
+
+
+It's possible to change the default `form` tag name to a different HTML tag, which can be helpful if you are nesting forms. However, native browser form behaviour, such as submitting when the `Enter` key is pressed, may no longer work.
+
+```jsx
+<Form
+  tagName="div"
+/>
+```
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -285,6 +285,7 @@ export default class Form extends Component {
       id,
       idPrefix,
       className,
+      tagName,
       name,
       method,
       target,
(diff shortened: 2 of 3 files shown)


epicfaace:
Looks great, thanks! Can you add a test for it?

I'm also wondering if not having a `form` tag will break anything -- nothing comes to mind, but let me know if you think of something.

juliankigwana:
I added a test. I cannot think of any obvious other than the default native browser behaviour might not work. But I would imagine that if you were manually changing the `tagName`, then you would be aware of this.

epicfaace:
@juliankigwana that makes sense. Can you add that bit as a warning in the documentation? (that the native browser behavior might not work)

epicfaace:
Additionally, the only allowed values for tagName will be valid JSX HTML tags, correct? Can you add a test in which one tries an invalid tagName?

juliankigwana:
> Additionally, the only allowed values for tagName will be valid JSX HTML tags, correct? Can you add a test in which one tries an invalid tagName?

Screen Shot 2019-07-08 at 11 09 11

I'm not sure we can test for it. Most browsers will render an unknown `banana` tag as a block element. It would be up to the developer to know what is valid HTML. 

I shall update the documentation with the warning and something saying it expects a valid JSX or HTML tag.

docs/form-customization.md
```diff
@@ -722,3 +722,15 @@ It's possible to disable the whole form by setting the `disabled` prop. The `dis
 ```
 
 If you just want to disable some of the fields, see the [`ui:disabled`](#disabled-fields) parameter in the `uiSchema` directive. 
+
+
+### Changing the tag name
+
+
+It's possible to change the default `form` tag name to a different HTML tag, which can be helpful if you are nesting forms. However, native browser form behaviour such as submitting when the `Enter` key is pressed, may no longer work.
```
epicfaace:
```suggestion
It's possible to change the default `form` tag name to a different HTML tag, which can be helpful if you are nesting forms. However, native browser form behaviour, such as submitting when the `Enter` key is pressed, may no longer work.
```

juliankigwana:
Thanks @epicfaace. I did apply the suggestion, but it's still showing here as changes requested. Do I need to do anything else?

epicfaace:
Nope, looks good -- thanks!

PR #1360, opened 2019-07-16
Allow "0." to be typed in to a text widget

### LINKED ISSUE #1355: '0.*' not valid number value when trying to add value
### Prerequisites

### Description
When entering a float number such as `0.1' the number field isn't allowing the insertion of the '.'.

It works as expected with '3.14' for example but whenever the string starts with '0' the user input of '.' is not allowed to be entered.

The schema is confiured as below
```json
 "fieldA": {
       "type": "number",
        "default": 1.0,
        "minimum": 0.0,
        "maximum": 1.0
   },
```
**NOTE:** You are able to enter the number as expected if you put a character infront of the 0 e.g '@0.1' and then deleting the '@'

### Steps to Reproduce

1. Delete default value
2. Type in '0.'
3. Keep trying to enter the '.'

#### Expected behavior

The '.' should be allowed to be entered 

#### Actual behavior

'.' character is not allowed to be entered after  a '0'

### Version
^1.2.1

I expect it is to do with the `asNumber` function (have commented on the line I think that is stopping it)
```js
export function asNumber(value) {
  if (value === "") {
    return undefined;
  }
  if (/\.$/.test(value)) {
    // "3." can't really be considered a number even if it parses in js. The
    // user is most likely entering a float.
    return value;
  }
  if (/\.0$/.test(value)) { <<---- HERE does this also handle 0.07
    // we need to return this as a string here, to allow for input like 3.07
    return value;
  }
  const n = Number(value);
  const valid = typeof n === "number" && !Number.isNaN(n);

  if (/\.\d*0$/.test(value)) {
    // It's a number, that's cool - but we need it as a string so it doesn't screw
    // with the user when entering dollar amounts or other values (such as those with
    // specific precision or number of significant digits)
    return value;
  }

  return valid ? n : value;
}
```

epicfaace:

Fixes #1355.

Changed some of the number field tests to run both on the default uiSchema and the uiSchema with the input type set to "text".

### Checklist

### DIFF
--- a/src/components/fields/NumberField.js
+++ b/src/components/fields/NumberField.js
@@ -69,7 +69,7 @@ class NumberField extends React.Component {
 
     let value = formData;
 
-    if (typeof lastValue === "string" && value) {
+    if (typeof lastValue === "string" && typeof value === "number") {
       // Construct a regular expression that checks for a string that consists
       // of the formData value suffixed with zero or one '.' characters and zero
       // or more '0' characters
--- a/test/NumberField_test.js
+++ b/test/NumberField_test.js
@@ -207,7 +207,7 @@ describe("NumberField", () => {
       });
 
       expect(comp.state.formData).eql(0);
-      expect($input.value).eql("0");
+      expect($input.value).eql(".00");
     });
 
     it("should update input values correctly when formData prop changes", () => {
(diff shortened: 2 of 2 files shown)


epicfaace:
@edi9999 / @LucianBuzzo can you review when you get a chance?

PR #1365, opened 2019-07-24
Fix onAddClick required event parameter

bresson:
Now it is optional. 

Passing the event object for onAddClick should not be mandatory. Without a fix, the event listener will throw an error. The event object is already available via lexical scope.

https://github.com/mozilla-services/react-jsonschema-form/issues/1314

### Checklist

* [Y ] **I'm updating documentation**
  - [ Y] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [Y ] **I'm adding or updating code**
  - [Y] I've added and/or updated tests
  - [Y] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [N] **I'm adding a new feature**
  - [NA ] I've updated the playground with an example use of the feature

### DIFF
--- a/docs/advanced-customization.md
+++ b/docs/advanced-customization.md
@@ -120,7 +120,7 @@ The following props are passed to each `ArrayFieldTemplate`:
 - `disabled`: A boolean value stating if the array is disabled.
 - `idSchema`: Object
 - `items`: An array of objects representing the items in the array. Each of the items represent a child with properties described below.
-- `onAddClick: (event) => void`: A function that adds a new item to the array.
+- `onAddClick: ([event]) => void`: A function that adds a new item to the array. The "event" object is optional as it is already accessible via lexical scope.
 - `readonly`: A boolean value stating if the array is read-only.
 - `required`: A boolean value stating if the array is required.
 - `schema`: The schema object for this array.
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -262,7 +262,7 @@ class ArrayField extends Component {
     return addable;
   }
 
-  onAddClick = event => {
+  onAddClick = (event = event) => {
     event.preventDefault();
     const { schema, registry = getDefaultRegistry(), onChange } = this.props;
     const { definitions } = registry;


epicfaace:
Hey @bresson, is this what you mentioned as "Option B" in the issue? Or is this a slightly different approach?

If someone wanted to call onAddClick programmatically with no event, then I don't think this fix would work because there would be no `event` in the lexical scope.

bresson:
@epicfaace Oh I apologize for the mistake! The pull request is actually option A, not the lexical scope option. The default event parameter in the commit is the SyntheticEvent object which I've understood React passes to each event handler automatically ( perhaps I misunderstood! ) :
"Your event handlers will be passed instances of SyntheticEvent, a cross-browser wrapper around the browser’s native event." 
I can change the commit and PR messaging if you want to go forward with this option.

I also think there are at least two other options if the commit needs to be reworked: 

Option C: 
```
onAddClick = event => {
  if ( event ) {
      event.preventDefault();
  }
...
}
```

Option D -- more investigation needed but possibly no event is needed at all 
```
onAddClick = () => {
  const { schema, formData, registry = getDefaultRegistry() } = this.props;
  ...
}
```
Thanks -- please let me know how you want to proceed.

epicfaace:
@bresson thanks for the clarification. I don't think Option A would work for the case in which I mentioned -- in which someone calls `onAddClick` _without the event handler_ in the argument -- though. For example, imagine the ArrayFieldTemplate looked like this:

```
(props) => <div><button onClick={e => props.onAddClick()} /></div>
```

In this case, no synthetic event would be passed in to `onAddClick` (or so I believe), and thus `event` will be undefined -- so I believe only Option C will properly handle without an error. Perhaps the best thing to do this write a test that uses the ArrayFieldTemplate above and checks for this behavior!

bresson:
@epicfaace Thank you for pointing that out. ```e => props.onAddClick()```  is an interesting scenario and Option C would provide the proper affordance. I'll update the commit and include relevant tests. Thank you for your patience 🙏

epicfaace:
@bresson any updates on this?

bresson:
@epicfaace thank you for following. Sorry for the delay and I’m taking another look at this later in the week ... been out of town and blitzed at work.

epicfaace:
@bresson any updates?

epicfaace:
closing this in favor of #1730

PR #1376, opened 2019-07-30
Inject defaults in arrays

ddelbondio:

When using dependencies in array items, defaults are now injected.
This adjusts injecting arrays into the form data from defaults to be
in line with the previous behaviour. So the defaults are injected when
there is either no value defined in the form data or when an array is
already set in the form data, injects only into existing entries.
This makes sure that we don't change any existing behaviour by creating
array entries where we did not create them before.

For example for following schema:
```
{
  "type": "array",
  "minItems": 2,
  "items": {
    "title": "asd",
    "type": "object",
    "properties": {
      "item": {
        "type": "string",
        "default": "foo"
      }
    }
  }
}
```
and following form data:
```
[{}]
```

The result will be (note how it did not full up the array to min items)
``` 
[{ "item": "foo" }]
``` 

### Checklist

### DIFF
--- a/playground/samples/defaults.js
+++ b/playground/samples/defaults.js
@@ -1,0 +1,70 @@
+module.exports = {
+  schema: {
+    title: "Schema default properties",
+    type: "object",
+    properties: {
+      valuesInFormData: {
+        title: "Values in form data",
+        $ref: "#/definitions/defaultsExample",
+      },
+      noValuesInFormData: {
+        title: "No values in form data",
+        $ref: "#/definitions/defaultsExample",
+      },
+    },
+    definitions: {
+      defaultsExample: {
+        type: "object",
+        properties: {
+          scalar: {
+            title: "Scalar",
+            type: "string",
+            default: "scalar default",
+          },
+          array: {
+            title: "Array",
+            type: "array",
+            items: {
+              type: "object",
+              properties: {
+                nested: {
+                  title: "Nested array",
+                  type: "string",
+                  default: "nested array default",
+                },
+              },
+            },
+          },
+          object: {
+            title: "Object",
+            type: "object",
+            properties: {
+              nested: {
+                title: "Nested object",
+                type: "string",
+                default: "nested object default",
+              },
+            },
+          },
+        },
+      },
+    },
+  },
+  uiSchema: {},
+  formData: {
+    valuesInFormData: {
+      scalar: "value",
+      array: [
+        {
+          nested: "nested array value",
+        },
+      ],
+      object: {
+        nested: "nested object value",
+      },
+    },
+    noValuesInFormData: {
+      array: [{}, {}],
+    },
+  },
+};
--- a/src/utils.js
+++ b/src/utils.js
@@ -181,10 +181,10 @@ function computeDefaults(
       includeUndefinedValues
     );
   } else if (isFixedItems(schema)) {
-    defaults = schema.items.map(itemSchema =>
+    defaults = schema.items.map((itemSchema, idx) =>
       computeDefaults(
         itemSchema,
-        undefined,
+        Array.isArray(parentDefaults) ? parentDefaults[idx] : undefined,
         definitions,
         formData,
         includeUndefinedValues
@@ -223,6 +223,17 @@ function computeDefaults(
       }, {});
 
     case "array":
+      // inject defaults into existing array defaults
+      if (Array.isArray(defaults)) {
+        defaults = defaults.map((item, idx) => {
+          return computeDefaults(
+            schema.items[idx] || schema.additionalItems || {},
+            item,
+            definitions
+          );
+        });
+      }
+
       // deeply inject defaults into already existing form data
       if (Array.isArray(rawFormData)) {
         defaults = rawFormData.map((array, idx) => {
--- a/src/utils.js
+++ b/src/utils.js
@@ -299,7 +299,6 @@ export function getDefaultFormState(
  * - scalars are overwritten/set by form data
  */
 export function mergeDefaultsWithFormData(defaults, formData) {
-  // Recursively merge deeply nested objects.
   if (Array.isArray(formData)) {
     const leftArray = Array.isArray(defaults) ? defaults : [];
     return formData.map((value, idx) => {
--- a/src/utils.js
+++ b/src/utils.js
@@ -223,6 +223,17 @@ function computeDefaults(
       }, {});
 
     case "array":
+      // deeply inject defaults into already existing form data
+      if (Array.isArray(rawFormData)) {
+        defaults = rawFormData.map((array, idx) => {
+          return computeDefaults(
+            schema.items,
+            (defaults || {})[idx],
+            definitions,
+            array
+          );
+        });
+      }
       if (schema.minItems) {
         if (!isMultiSelect(schema, definitions)) {
           const defaultsLength = defaults ? defaults.length : 0;
@@ -269,14 +280,46 @@ export function getDefaultFormState(
     // No form data? Use schema defaults.
     return defaults;
   }
-  if (isObject(formData)) {
-    // Override schema defaults with form data.
-    return mergeObjects(defaults, formData);
+  if (isObject(formData) || Array.isArray(formData)) {
+    return mergeDefaultsWithFormData(defaults, formData);
   }
   if (formData === 0) {
     return formData;
   }
   return formData || defaults;
+}
+
+/**
+ * When merging defaults and form data, we want to merge in very specific way:
+ * - objects are deeply merged
+ * - arrays are merged in such way that:
+ *   - when the array is set in form data, only array entries set in form data
+ *     are deeply merged, additional entries from the defaults are ignored
+ *   - when the array is not set in form data, the default is copied over
+ * - scalars are overwritten/set by form data
+ */
+export function mergeDefaultsWithFormData(defaults, formData) {
+  // Recursively merge deeply nested objects.
+  if (Array.isArray(formData)) {
+    const leftArray = Array.isArray(defaults) ? defaults : [];
+    return formData.map((value, idx) => {
+      if (leftArray[idx]) {
+        return mergeDefaultsWithFormData(leftArray[idx], value);
+      }
+      return value;
+    });
+  } else if (isObject(formData)) {
+    const acc = Object.assign({}, defaults); // Prevent mutation of source object.
+    return Object.keys(formData).reduce((acc, key) => {
+      acc[key] = mergeDefaultsWithFormData(
+        defaults ? defaults[key] : {},
+        formData[key]
+      );
+      return acc;
+    }, acc);
+  } else {
+    return formData;
+  }
 }
 
 export function getUiOptions(uiSchema) {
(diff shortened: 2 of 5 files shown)


epicfaace:
Should we also replace `mergeObjects` here (https://github.com/mozilla-services/react-jsonschema-form/pull/1376/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R159) with mergeDefaultsWithFormData?

src/utils.js
```diff
@@ -240,13 +251,45 @@ export function getDefaultFormState(_schema, formData, definitions = {}) {
...
+ * When merging defaults and form data, we want to merge in very specific way:
+ * - objects are deeply merged
+ * - arrays are merged in such way that:
+ *   - when the array is set in form data, only array entries set in form data
+ *     are deeply merged, additional entries from the defaults are ignored
+ *   - when the array is not set in form data, the default is copied over
+ * - scalars are overwritten/set by form data
+ */
+export function mergeDefaultsWithFormData(defaults, formData) {
+  // Recursively merge deeply nested objects.
```
epicfaace: Probably don't need this line anymore

ddelbondio: Good point. 
Do you want me to squash any changes I do or should I do multiple commits?

ddelbondio:
> Should we also replace `mergeObjects` here (https://github.com/mozilla-services/react-jsonschema-form/pull/1376/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R159) with `mergeDefaultsWithFormData`?

The goal would be to merge defaults into array items defined at parent level, right? 
I don't think that will be solved by just using mergeDefaultsWithFormData() in this place. It probably needs some juggling in the rest of the function as well. Since the spec doesn't define how to interpret defaults at all, it makes sense to bikeshed the wanted behaviour here, since this is already not directly obvious (and it doesn't get much better with this PR) how the defaults work in rjsf.

src/utils.js
epicfaace: Multiple commits are best -- I'll squash when merging.

epicfaace:
@ddelbondio sorry, I didn't understand what you meant. Could you try explaining it again? 🙂

ddelbondio:
@epicfaace Try to let me explain with code :)
I have added documentation and a new example explaining the merging. I have also added the merging of array defaults in the schema itself. 

Can you please check the new documentation chapter if that makes sense to you? The merging within the schema itself works slightly different than merging with the form data (it fully overwrites arrays when defined in the defaults, since this makes more sense IMHO). That is what I meant in my comment.

I have added [a test](https://github.com/mozilla-services/react-jsonschema-form/pull/1376/files#diff-0de170772e45d2602c103dd916607787R342) which would fail of we changed the `mergeObjects` to `mergeDefaultsWithFormData` in that place. 

I also rebased, since there were changes which conflicted with master.

test/utils_test.js
```diff
@@ -306,6 +307,210 @@ describe("utils", () => {
...
+            },
+          },
+        };
+
+        expect(getDefaultFormState(schema, {})).eql({
+          level1: { level2: [{ item: "parent-default-1" }, {}] },
+        });
+      });
+
+      it("should merge schema array item defaults from sam item for overlapping default definitions", () => {
```
```suggestion
      it("should merge schema array item defaults from the same item for overlapping default definitions", () => {
```

test/utils_test.js
```diff
@@ -306,6 +307,210 @@ describe("utils", () => {
...
+                },
+              ],
+            },
+          },
+        };
+
+        expect(getDefaultFormState(schema, {})).eql({
+          level1: [
+            { item: "property-default-1" },
+            { item: "additional-default" },
```
epicfaace: Is this an intuitive / logical result? (If a schema author purposefully makes the second item in the array default `{}`, then the second item is injected by the default of `additionalItems`)

test/utils_test.js
```diff
@@ -306,6 +307,210 @@ describe("utils", () => {
...
+            },
+          },
+        };
+
+        expect(getDefaultFormState(schema, {})).eql({
+          level1: [
+            { item: "property-default-1" },
+            { item: "child-default-2" },
+            { item: "additional-default" },
+          ],
```
epicfaace: Can you add a test in which you make sure that one of the root-defaults are injected? (Similar to the grandparent test, but a bit simpler since it's only one level)

test/utils_test.js
```diff
@@ -617,6 +859,120 @@ describe("utils", () => {
...
+                        },
+                      },
+                    ],
+                  },
+                },
+              },
+            },
+          },
+        };
+        expect(getDefaultFormState(schema, [{ foo: { name: "Name" } }])).eql([
```
epicfaace: Can you add another test in which the formData is empty, name has a default of "Name", and thus the result should have both name and grade?

test/utils_test.js
ddelbondio: It's in line with merging of objects in defaults and merging of array with the form data.
It's logical from the point that we try to merge as much deeply as we possible can. I tried to keep the special rules about merging as small as possible here.

test/utils_test.js
ddelbondio: I am not sure I understand what we should test there additionally which is not covered yet by the grandparent test?

test/utils_test.js
ddelbondio: Nice catch. That actually fails (when formData is`[{}]`, which is the only time the code actually injects the defaults here). But this already fails when dependencies are not used within arrays.
I have to check if/how we can fix this. 
But I guess it does make sense that the dependencies are evaluated (again) after injecting the defaults.

src/utils.js
```diff
@@ -269,16 +291,47 @@ export function getDefaultFormState(
...
+ * - objects are deeply merged
+ * - arrays are merged in such way that:
+ *   - when the array is set in form data, only array entries set in form data
+ *     are deeply merged, additional entries from the defaults are ignored
+ *   - when the array is not set in form data, the default is copied over
+ * - scalars are overwritten/set by form data
+ */
+export function mergeDefaultsWithFormData(defaults, formData) {
+  if (Array.isArray(formData)) {
+    const leftArray = Array.isArray(defaults) ? defaults : [];
```
epicfaace: Can we just do

```suggestion
    if (!Array.isArray(defaults)) {
        defaults = [];
    }
```

src/utils.js
```diff
@@ -223,6 +223,28 @@ function computeDefaults(
       }, {});
 
     case "array":
+      // inject defaults into existing array defaults
```
epicfaace:
```suggestion
      // Inject defaults into existing array defaults
```

src/utils.js
```diff
@@ -223,6 +223,28 @@ function computeDefaults(
...
+        defaults = defaults.map((item, idx) => {
+          return computeDefaults(
+            schema.items[idx] || schema.additionalItems || {},
+            item,
+            definitions
+          );
+        });
+      }
+
+      // deeply inject defaults into already existing form data
```
epicfaace:
```suggestion
      // Deeply inject defaults into already existing form data
```

src/utils.js
```diff
@@ -223,6 +223,28 @@ function computeDefaults(
...
+      if (Array.isArray(rawFormData)) {
+        defaults = rawFormData.map((array, idx) => {
```
epicfaace: Can we rename `array` to `item`? (`array` isn't actually an array, so the name sounds misleading)

test/utils_test.js
```diff
@@ -306,6 +307,210 @@ describe("utils", () => {
...
+              properties: {
+                level2: {
+                  type: "array",
+                  items: [
+                    {
+                      type: "string",
+                      default: "child-default-1",
+                    },
+                    {
+                      type: "string",
```
epicfaace: What if I _don't_ want the default to be set for this item to "root-default-2" (but still want to keep the grandparent default the same as it is)? Might there be an intuitive way to "unset" a default? (this is probably beyond the scope of this PR, but could be added in the future)

test/utils_test.js
epicfaace: Oh, makes sense -- I missed that `items` has a length of one here, so it makes sense for the default from `additionalItems` to apply to any items after the first one.

Can you add another test with the same schema, except for with `items` of length two, and thus, it should _not_ overwrite the default value of item 2?

test/utils_test.js
epicfaace: I had meant to have a "parent" test instead of a "grandparent" test. You are right, though, the grandparent test already seems to cover that logic.

playground/samples/defaults.js
```diff
@@ -0,0 +1,70 @@
...
+    definitions: {
+      defaultsExample: {
+        type: "object",
+        properties: {
+          scalar: {
+            title: "Scalar",
+            type: "string",
+            default: "scalar default",
+          },
+          array: {
```
epicfaace: Can you rename `array` to something else? (it would be clearer in the playground if the array were not named `array`, just because `type: array` is already there)

Also for `object` too.

test/utils_test.js
epicfaace: Yeah, if you can fix it now, please do that, otherwise, we can fix that in a follow-up PR.

test/utils_test.js
```diff
@@ -617,6 +859,120 @@ describe("utils", () => {
...
+          {
+            foo: {
+              name: "Name",
+              grade: "A",
+            },
+          },
+        ]);
+      });
+
+      it("should populate defaults for nested dependencies in arrays when matching enum values in oneOf", () => {
```
epicfaace: Why did you add this test? (was there a time when "should populate defaults for nested dependencies in arrays" passed but this test failed?)

test/utils_test.js
```diff
@@ -956,6 +1312,93 @@ describe("utils", () => {
...
+      };
+      const obj2 = {
+        a: file,
+      };
+      expect(mergeDefaultsWithFormData(obj1, obj2).a).instanceOf(File);
+    });
+  });
+
   describe("mergeObjects()", () => {
     it("should't mutate the provided objects", () => {
```
```suggestion
    it("shouldn't mutate the provided objects", () => {
```

test/utils_test.js
```diff
@@ -956,6 +1312,93 @@ describe("utils", () => {
...
   });
 
+  describe("mergeDefaultsWithFormData()", () => {
+    it("should't mutate the provided objects", () => {
+      const obj1 = { a: 1 };
+      mergeDefaultsWithFormData(obj1, { b: 2 });
+      expect(obj1).eql({ a: 1 });
+    });
+
+    it("should't mutate the provided arrays", () => {
```
```suggestion
    it("shouldn't mutate the provided arrays", () => {
```

test/utils_test.js
```diff
@@ -956,6 +1312,93 @@ describe("utils", () => {
     });
   });
 
+  describe("mergeDefaultsWithFormData()", () => {
+    it("should't mutate the provided objects", () => {
```
```suggestion
    it("shouldn't mutate the provided objects", () => {
```

epicfaace:
@ddelbondio just checking on this?

epicfaace:
@ddelbondio if you don't have time at the moment to address these I'll just take over this PR

epicfaace:
Closing this in favor of #1499.

playground/samples/defaults.js
epicfaace: nvm

ddelbondio:
@epicfaace: thanks for taking this over!

PR #1383, opened 2019-08-02
Add extraErrors prop (allows async validation)

thijssteel:

This is another attempt at adding async validation. The method used is similar to PR #874.
I have already outlined my idea in issue #1381.

Basically, what this PR does is add an extraErrors prop to the Form component. Errors in this prop are added to the other validation errors. This makes async validation possible because the surrounding component can perform async validation either on submit or on validation.

I have attempted to write tests but failed miserably, any help would be appreciated.

### Checklist

### DIFF
--- a/playground/samples/single.js
+++ b/playground/samples/single.js
@@ -5,4 +5,5 @@ module.exports = {
   },
   formData: "initial value",
   uiSchema: {},
+  extraErrors: [],
 };
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -64,7 +64,7 @@ export default class Form extends Component {
     const uiSchema = "uiSchema" in props ? props.uiSchema : this.props.uiSchema;
     const edit = typeof inputFormData !== "undefined";
     const extraErrorsChange =
-      !prevProps || props.extraErrors !== prevProps.extraErrors;
+      !prevProps || (props.extraErrors !== prevProps.extraErrors);
     const liveValidate = props.liveValidate || this.props.liveValidate;
     const mustValidate =
       edit && !props.noValidate && (liveValidate || extraErrorsChange);
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -43,7 +43,11 @@ export default class Form extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const nextState = this.getStateFromProps(nextProps, nextProps.formData);
+    const nextState = this.getStateFromProps(
+      nextProps,
+      nextProps.formData,
+      this.props
+    );
     if (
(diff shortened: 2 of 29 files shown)


playground/samples/single.js
```diff
@@ -5,4 +5,5 @@ module.exports = {
   },
   formData: "initial value",
   uiSchema: {},
+  extraErrors: [],
```
epicfaace: Why do you need an empty extraErrors prop for each item in the playground?

src/components/Form.js
```diff
@@ -53,20 +58,29 @@ export default class Form extends Component {
...
   }
 
-  getStateFromProps(props, inputFormData) {
+  getStateFromProps(props, inputFormData, prevProps) {
     const state = this.state || {};
     const schema = "schema" in props ? props.schema : this.props.schema;
     const uiSchema = "uiSchema" in props ? props.uiSchema : this.props.uiSchema;
     const edit = typeof inputFormData !== "undefined";
+    const extraErrorsChange =
+      !prevProps || props.extraErrors !== prevProps.extraErrors;
```
epicfaace:
```suggestion
      !prevProps || (props.extraErrors !== prevProps.extraErrors);
```

playground/samples/single.js
thijssteel: That way it overwrites it every time you switch

thijssteel:
Damn, i did not think of that. I'll try to think of a fix

src/components/Form.js
```diff
@@ -239,9 +255,14 @@ export default class Form extends Component {
       newFormData = this.getUsedFormData(this.state.formData, fieldNames);
     }
 
+    let errors = [];
+    let errorSchema = {};
     if (!this.props.noValidate) {
-      const { errors, errorSchema } = this.validate(newFormData);
-      if (Object.keys(errors).length > 0) {
+      const validationResult = this.validate(newFormData);
+      errors = validationResult.errors;
```
thijssteel: i tried doing this with a deconstructor, but got a boolean is not a function error, very strange.
let canSubmit
({errrors, errorSchema, canSubmit} = this.validate(newFormData)

thijssteel:
I'm going to have to rewrite this, keep getting stuck with the submission blocking

thijssteel:
i've completely rewritten this PR in a separate branch. Would you rather have me open a new PR or shall i change the source branch on this PR?

epicfaace:
I think another PR would be good, so we can keep this PR for a record of the earlier attempt. Thanks!

epicfaace:
Closed in favor of #1444

PR #1403, opened 2019-08-13
Encode checkbox settings in playground

### LINKED ISSUE #1400: Playground URL should serialize checkboxes as well
### Description

Currently, if I check a checkbox such as "Live validation" or "Omit extra data" and then copy a link from a playground, the link does not encode that I have checked the checkbox. It only encodes the schema, uiSchema, and formData.

### Solution

Encode the checkbox values as well, so that links can be more easily shared.

GrooChu:

[Please describe them here]

Fixes #1400 
Serialize the checkbox values in the playground

### Checklist

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -384,12 +384,14 @@ class App extends Component {
     this.setState({ formData, shareURL: null });
 
   onShare = () => {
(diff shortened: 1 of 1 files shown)


epicfaace:
Thanks! 😄

Issue #1396, opened 2019-08-11
Array of objects data entry not working with liveOmit


epicfaace:
### Description

Array of objects data entry not working with liveOmit.

### Steps to Reproduce

1. Go to [this playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJkZWZpbml0aW9ucyI6e30sInR5cGUiOiJhcnJheSIsIml0ZW1zIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7Im5hbWUiOnsidHlwZSI6InN0cmluZyJ9fX19LCJ1aVNjaGVtYSI6e319)
2. Type in a value for "name".

#### Expected behavior

The form data should be

```
[{name: "asdasd"}]
```

#### Actual behavior

The form data is `{}` -- somehow liveOmit is not adding in the data properly.

![image](https://user-images.githubusercontent.com/1689183/62828438-44cb2600-bb9b-11e9-8b48-a7d90b7d759a.png)

### Version

1.7.0+ (latest version from master)

This seems to be a regression added by PR #1371 .

[PR #1371 demo](https://deploy-preview-1371--trusting-agnesi-d3347d.netlify.com/#eyJmb3JtRGF0YSI6eyJzaG93Ijp0cnVlfSwic2NoZW1hIjp7ImRlZmluaXRpb25zIjp7fSwidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsic2hvdyI6eyJ0eXBlIjoiYm9vbGVhbiJ9fSwiZGVwZW5kZW5jaWVzIjp7InNob3ciOnsib25lT2YiOlt7InByb3BlcnRpZXMiOnsic2hvdyI6eyJjb25zdCI6dHJ1ZX0sInBhcnRpY2lwYW50cyI6eyJ0eXBlIjoiYXJyYXkiLCJpdGVtcyI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJuYW1lIjp7InR5cGUiOiJzdHJpbmcifX19fX19XX19fSwidWlTY2hlbWEiOnt9fQ==)
[
[...]

PR #1406, opened 2019-08-14
Fix omitData for top level array

Tonexus:

Enabling omitData currently converts a top level array to an object. Now includes a check if the form's top level type is an array, and converts back to an array if necessary.

#1396

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -139,7 +139,12 @@ export default class Form extends Component {
       return formData;
     }
 
-    return _pick(formData, fields);
+    let data = _pick(formData, fields);
+    if (typeof formData === "array") {
+      return Object.keys(data).map((k) => data[k]);
+    } else {
+      return data;
+    }
   };
 
   getFieldNames = (pathSchema, formData) => {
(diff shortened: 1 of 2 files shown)


epicfaace:
Thanks @Tonexus ! Can you add a test?

Tonexus:
Ok, added one test to make sure the top level array stays as an array.

src/components/Form.js
```diff
@@ -139,7 +139,12 @@ export default class Form extends Component {
       return formData;
     }
 
-    return _pick(formData, fields);
+    let data = _pick(formData, fields);
+    if (Array.isArray(formData)) {
+      return Object.keys(data).map(key => data[key]);
```
epicfaace: This won't work when formData is a non-empty array (for example, if formData is `["a","b","c"]`).

Tonexus: Not sure what you mean. Suppose we have `formData = ["a","b","c"]` and `fields = ["0", "2"]`. Then, `data = _pick(formData, fields)` results in `data = {0: "a", 2: "c"}`. `Object.keys(data)` evaluates to `["0", "2"]`, and `["0", "2"].map(key => data[key])` evaluates to `[data["0"], data["2"]] = ["a", "c"]`, which is what we want.

epicfaace: Oops, I thought for a second that fields would be an empty array. That makes sense, thanks!

PR #1408, opened 2019-08-14
[V2] Upgrade to react 16

### LINKED ISSUE #1275: Requiring React 16
I'm working on a PR on `react-jsonschema-form` for passing an Ajv instance from outside (allowing custom keywords etc.), and I need to use the Context API which has been introduced in [React 16.3](https://reactjs.org/blog/2018/03/29/react-v-16-3.html).

Even better would be availability of `contextType` for React class components, introduced in [React 16.6](https://reactjs.org/blog/2018/10/23/react-v-16-6.html).

Would you accept a PR which bumps up React version to 16.3, 16.6 or even current 16.8.6? and peerDependency to >= 16.3 or >= 16.6, respectively?

### LINKED ISSUE #1427: react 16: array keys keep regenerating
### Prerequisites

### Description

With react 16 the component lifecycle of `ArrayField` is different that with react 15.
The `getDerivedStateFromProps` method is first called with an the old props but updated state (e.g. adding a single item) and then recomputes all the keys because the lengths do not match. After that the props update and the keys get regenerated again because the lengths do not match.

Here you can see a log of `nextProps` and `prevState` args of `getDerivedStateFromProps`:

react 15:
Screenshot from 2019-08-22 12-25-02

react 16:
Screenshot from 2019-08-22 12-23-06

### Steps to Reproduce

1. add a console.log statement
1. open the array example on the playground
1. click the add button twice

#### Expected behavior

The keys should update the same way as they are doing with react 15.

### Version

v2 branch

epicfaace:

Upgrade to react 16, fixes #1275, fixes #1427 .

### Checklist


epicfaace:
@fsteger do you know why these tests may be failing on upgrade to React 16:

  1) ArrayField
       List of inputs
         should retain existing row keys/ids when adding new row:

  2) ArrayField
       List of inputs
         should allow inserting anywhere in list:

  3) ArrayField
       List of inputs
         should retain row keys/ids of remaining rows when a row is removed:

zepatrik:
I had a look into this and it probably depends on some deprecated lifecycle methods used. I can rewrite everything to functional components using hooks as I already have some experience doing that on a large scale. Will be done by today and open a PR then.

zepatrik:
It is actually more work than expected. I'm still working on that, biggest problem is that keys were used wrong. Seems like react 15 is passing the key prop to the child although it is only for internal use. With react 16 that prop is not accessible.

E.g. [this](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/ArrayField.js#L768) is not accessible [here](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/ArrayField.js#L49)

But getting there.

epicfaace:
@zepatrik thanks for the update. What exactly is the change between react 15 and react 16 regarding keys? Is it that you can't pass on a key prop to a child? (that sounds strange, because if the key prop is accessible, then you can pass it on, right...?)

zepatrik:
I could not really find a breaking change in reacts release notes but as you can see [here](https://jsfiddle.net/uyxq86m3/) (also look into the console) it is not possible to access the key prop (and apparently it was before)

zepatrik:
Also `static getDerivedStateFromProps()` lifecycle [is used](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/ArrayField.js#L218) which was [added in react 16.3.0](https://github.com/facebook/react/blob/master/CHANGELOG.md#react-dom-17). The tests worked until now because the derived state was not used.

epicfaace:
What do you mean by "the derived state was not used"? We did use `static getDerivedStateFromProps` but in combination with a polyfill from "react-lifecycles-compat". Will probably need to remove the polyfill though for react 16.

zepatrik:
well because it appeared to not have been called, the state differed on the react version

zepatrik:
Here are some debug information. Logging the `nextProps` and `prevState` args of `getDerivedStateFromProps` in `ArrayField`

This is also what the first failing test detects.

react 15:
Screenshot from 2019-08-22 12-25-02

react 16:
Screenshot from 2019-08-22 12-23-06

Steps to reproduce:

1. add a console.log statement
1. open the array example on the playground
1. click the add button twice

As you can see for some reason the component updates twice, once with an updated state and once with updated props.

zepatrik:
tracked that as issue #1427

edi9999:
Very cool !

PR #1414, opened 2019-08-15
Fix - additionalProperties' type-guessing

### LINKED ISSUE #1412: additionalProperties' type-guessing does not work when value is false
### Steps to Reproduce

1. Create a schema with `additionalProperties: true` and with an additional property equal to `true` ([playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJhc2Rhc2QiOnRydWV9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsImFkZGl0aW9uYWxQcm9wZXJ0aWVzIjp0cnVlfSwidWlTY2hlbWEiOnt9fQ==)).
2. Uncheck the checkbox.

#### Expected behavior

Unchecking a checkbox should result in an unchecked checkbox.

#### Actual behavior

The checkbox turns into an empty textbox!

### Version

1.7.0 (latest playground)

### Discussion

The issue seems to be this block of code: https://github.com/mozilla-services/react-jsonschema-form/blob/6728977b0085ef37c310b4d40d076a2622664ba7/src/utils.js#L545-L550

The idea of using `{ type: guessType(formData[key]) }` is that, when `additionalProperties` is set to `true`, it should guess the input type based on existing form data. Somehow, unchecking a checkbox ends up sending a value to `guessType` that returns "string".

It's probably caused by this piece of code, which sets `value` to an empty string if it is equal to false: https://github.com/mozilla-services/react-jsonschema-form/blob/310b8a5309bd55fd1c3e1533c93b3ce135b03dda/src/components/fields/ObjectField.js#L85-L94

rbrishabh:
Fixes #1412 

Hey @epicfaace 

This patch seems to fix the problem locally. Is this okay or this patch will create problems somewhere else?

Also, should I change this [comment](https://github.com/mozilla-services/react-jsonschema-form/blob/310b8a5309bd55fd1c3e1533c93b3ce135b03dda/src/components/fields/ObjectField.js#L92) here to something like

```set empty values to false```

Thanks!

### DIFF
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -90,7 +90,7 @@ class ObjectField extends Component {
         // fields which are "mandated" by the schema, these fields can
         // be set to undefined by clicking a "delete field" button, so
         // set empty values to the empty string.
-        if (typeof value == "boolean") {
+        if (typeof value === "boolean") {
           value = false;
         } else {
           value = "";
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -90,7 +90,11 @@ class ObjectField extends Component {
         // fields which are "mandated" by the schema, these fields can
         // be set to undefined by clicking a "delete field" button, so
         // set empty values to the empty string.
-        value = "";
+        if (typeof value == "boolean") {
+          value = false;
+        } else {
+          value = "";
+        }
       }
       const newFormData = { ...this.props.formData, [name]: value };
       this.props.onChange(
--- a/src/components/fields/ObjectField.js
+++ b/src/components/fields/ObjectField.js
@@ -92,7 +92,7 @@ class ObjectField extends Component {
         // set empty values to the empty string.
         if (typeof value === "boolean") {
           value = false;
-        } else {
+        } else if (value === "" || value === undefined) {
           value = "";
         }
       }

### CODE REVIEW

**Comment by epicfaace** [comment]
@rbrishabh thanks for the PR. I don't think this will work, because we don't want to set the value of additionalProperties to a boolean false in all cases (for example, when someone is clearing a textbox). We only want to set it to false when the type is already a boolean and the user unchecks the checkbox, making it false. You can also see that your change breaks some tests here: https://travis-ci.org/mozilla-services/react-jsonschema-form/jobs/572155050

I think the issue might be that `if (!value && addedByAdditionalProperties) {` is being evaluated when `value` is equal to `false`, and thus it is setting value to a string and turning it into a textbox.

**Comment by rbrishabh** [comment]
Ah! Thanks! I will look into this and get back to you today! :)

**Comment by rbrishabh** [comment]
I am sorry for accidentally closing this PR. 

The change I made now seems a little hacky, but it works, and all the test cases pass locally as well!

Let me know if this needs any changes! Thanks =)!

**Comment by epicfaace** [comment]
Thanks for updating the PR. However, this doesn't catch other issues in which one types in falsy input; for example, see [this playground example](https://deploy-preview-1414--rjsf.netlify.com//#eyJmb3JtRGF0YSI6eyJmaXJzdE5hbWUiOiJDaHVjayIsImxhc3ROYW1lIjoiTm9ycmlzIiwibmV3S2V5IjoiIn0sInNjaGVtYSI6eyJ0aXRsZSI6IkEgY3VzdG9taXphYmxlIHJlZ2lzdHJhdGlvbiBmb3JtIiwiZGVzY3JpcHRpb24iOiJBIHNpbXBsZSBmb3JtIHdpdGggYWRkaXRpb25hbCBwcm9wZXJ0aWVzIGV4YW1wbGUuIiwidHlwZSI6Im9iamVjdCIsInJlcXVpcmVkIjpbImZpcnN0TmFtZSIsImxhc3ROYW1lIl0sImFkZGl0aW9uYWxQcm9wZXJ0aWVzIjp7InR5cGUiOiJudW1iZXIifSwicHJvcGVydGllcyI6eyJmaXJzdE5hbWUiOnsidHlwZSI6InN0cmluZyIsInRpdGxlIjoiRmlyc3QgbmFtZSJ9LCJsYXN0TmFtZSI6eyJ0eXBlIjoic3RyaW5nIiwidGl0bGUiOiJMYXN0IG5hbWUifX19LCJ1aVNjaGVtYSI6eyJmaXJzdE5hbWUiOnsidWk6YXV0b2ZvY3VzIjp0cnVlLCJ1aTplbXB0eVZhbHVlIjoiIn19LCJsaXZlU2V0dGluZ3MiOnsidmFsaWRhdGUiOnRydWUsImRpc2FibGUiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlfX0=) -- when someone types in "0" in the additionalProperties' value, it doesn't let the user type that in as the form immediately sets the value to an empty string.

I think a simpler fix would involve changing this line

```
if (!value && addedByAdditionalProperties) {
```

to something like

```
if (typeof value === 'undefined' && addedByAdditionalProperties) {
```

That way the condition is never executed if value is falsy (`null`, `0`, `false`, etc.) so that the user can still type in falsy inputs.

Let me know if that makes sense -- I haven't tried out this change yet so there's a chance something might be missing in this approach.

rbrishabh:
> Let me know if that makes sense -- I haven't tried out this change yet so there's a chance something might be missing in this approach.

Yep, everything works great, but there's one small problem - 
For values that are defined, (e.g. ""), ```if (typeof value === 'undefined' && addedByAdditionalProperties) {``` completely deletes the object pair and then this test fails -

> deleting content of value input should not delete pair

I am making a small patch which I believe solves the issue, look into it, and please let me know if it has any problem!
 Thanks =)

rbrishabh:
Hey @epicfaace ! It's been a while since I heard back from you! I was wondering if this needs any changes!

Thanks! =)

epicfaace:
@rbrishabh sorry for the delay. Could you please add a test?

rbrishabh:
Sure, I will do that! ;)

epicfaace:
@rbrishabh bump

PR #1415, opened 2019-08-17
[Feat] Introduce a More Flexible Way to Trigger Form Submit Programmatically

ZheyangSong:
+ This change adds a new utility function, `submitForm`, to trigger a form's submission.

This change is an enhancement to the existing programmatic form submission functionality. This change introduces a more flexible way to submit a form programmatically. This will make using the form and a submit trigger separately easier and more flexible (They don't necessarily have to be coded under a common parent component so that the form's ref can be passed to the trigger component).

Example of usage is documented here:
https://github.com/mozilla-services/react-jsonschema-form/pull/1415/files#diff-1a523bd9fa0dbf998008b37579210e12R168-R179

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -66,6 +66,41 @@ export function getDefaultRegistry() {
   };
 }
 
+const currentFormInstances = {};
+
+export function registerFormNode(formId, formNode) {
+  if (typeof formId !== "undefined") {
+    currentFormInstances[formId] = formNode;
+  }
+}
+
+export function unregisterFormNode(formId) {
+  delete currentFormInstances[formId];
+}
+
+function getRegisteredFormNode(formId) {
+  const formNode = currentFormInstances[formId];
+
+  if (!formNode && process.env.NODE_ENV === "development") {
+    console.error(
+      `No registered form node found with given id (${formId}). Please check if the target form has proper id assigned.`
+    );
+  }
+
+  return formNode;
+}
+
+export function submitForm(formId) {
+  const formNode = getRegisteredFormNode(formId);
+
+  if (formNode) {
+    formNode.dispatchEvent(new window.Event("submit", { cancelable: true }));
+    return true;
+  } else {
+    return false;
+  }
+}
+
 export function getSchemaType(schema) {
   let { type } = schema;
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -163,6 +163,22 @@ render((
 yourForm.submit();
 ```
 
+Alternatively, if your form component is given an `id`, instead of getting a reference to the `Form` component, you can use the utility function `submitForm` to submit the form programmatically. Advantage of using this approach is that you now don't have to pass the form component reference around, if your implementation needs to trigger the form submit in another component (say, a modal) or as part of a data flow (say, a redux middleware after some AJAX finishes).
+
+```js
+import Form, { submitForm } from 'react-jsonschema-form';
+
+const onSubmit = ({formData}) => console.log("Data submitted: ",  formData);
+const formId = 'my-form';
+
+render((
+  <Form id={formId} schema={schema}
+        onSubmit={onSubmit} ref={(form) => {yourForm = form;}}/>
+), document.getElementById("app"));
+
+submitForm(formId);
+```
+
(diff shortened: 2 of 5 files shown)


src/utils.js
```diff
@@ -66,6 +66,41 @@ export function getDefaultRegistry() {
...
+  }
+
+  return formNode;
+}
+
+export function submitForm(formId) {
+  const formNode = getRegisteredFormNode(formId);
+
+  if (formNode) {
+    formNode.dispatchEvent(new window.Event("submit", { cancelable: true }));
```
ZheyangSong: I have to explicitly use `window.Event` here because of a test environment error (it seems related to test infra setup). If a reviewer can teach me how to fix below test environment issue. I'll update this PR by removing the `window`.

Screen Shot 2019-08-17 at 12 28 24 AM

epicfaace:
@ZheyangSong thanks for the PR. Why wouldn't it be enough just to use the ref? Is there a situation in which using the ref would be difficult/impossible but submitForm would work?

ZheyangSong:
@epic I haven't found a situation that using ref will be impossible. And ref works in all my cases. So, this fact makes me add this new `submitForm` as an alternative approach. The motivation for this new way is mainly writing clean code.

Before Fiber, if the form is wrapped multiple times for some reasons, the ref has to be created on the form component and all the wrappers so that the ref to form can be accessible. The resulting code is a little clumsy. Though Fiber's ForwardRef makes passing ref to form easier, it'd still be better if I can avoid creating a ref.

Here is a simplified example to illustrate the idea:

Using ref:
```jsx
// This CompA can be doing other stuff before rendering the form. I skipped them for simplicity. This is applied to the following components. 
class CompA extends Component {
  formRef = null;

  render() {
    return <Form ref={this.getFormRef} />;
  }

  getFormRef = (form) => this.formRef = form;

  submit = () => this.formRef.submit();
}

class CompB extends Component {
  compARef = null;

  state = {};

  render() {
    return (
      <div>
        <CompA ref={this.getCompARef} onChange={this.formChange}/>
        <LiveVisualizedCompAInput config={this.state.config} />
      </div>
    );
  }
  getCompARef = (compA) => this.compARef = compA;

  formChange = (config) => this.setState({ config });

  saveChange = () => this.compARef.submit();
}

class CompC extends Component {
  compBRef = null;

  render() {
    return (
       <div>
          <NavBar>
             <button onClick={this.submit} />
          </NavBar>
          <div>
             <CompB ref={this.getCompBRef} />
          </div>
       </div>
     );
  }

  getCompBRef = (compB) => this.compBRef = compB;

  submit = () => this.compBRef.saveChange();
}
```

Using utility function, `submitForm`:
```jsx
class CompA extends Component {
  render() {
    return <Form id={this.props.id} />;
  }
}

class CompB extends Component {
  state = {};

  render() {
    return (
      <div>
        <CompA id={this.props.id} onChange={this.formChange}/>
        <LiveVisualizedCompAInput config={this.state.config} />
      </div>
    );
  }

  formChange = (config) => this.setState({ config });
}

class CompC extends Component {
  targetFormId = "my-form-with-live-visualization";

  render() {
    return (
       <div>
          <NavBar>
             <button onClick={this.submit} />
          </NavBar>
          <div>
             <CompB id/>
          </div>
       </div>
     );
  }

  submit = () => submitForm(targetFormId);
}
```

You can see. In this case, using a utility function can be less verbose and the resulting code focuses more on the business logic.

In general, this approach promotes building a component and its controller/pairing component in a way similar to native html tags like:
* <label /> and <input />
* aria-labelledby and id

It removes the requirement where these components are put in the HTML.

Another example:
```jsx
class SmarterForm extends Component {
  render () {
    return <Form id={this.props.id} />
  }
}

class SmarterFormSubmitBtn extends Component {
  render () {
    return <button onClick={this.submitTarget} />;
  }

  submitTarget = () => submit(this.props.formId}
}
// WIth the above reusable components set, I can structure my layout in any way I want without worry about passing ref around
```

The main effort here is to improve the syntax and code verbosity.

At last, the feature request is optional. It can be closed without merge.

PR #1418, opened 2019-08-19
Live omit fixes

epicfaace:
### Reasons for making this change

Fixes a bunch of live omit issues, yay. Fixes #1388 and fixes #1396.

- If there is an array with empty objects in it, `getFieldNames` now returns values such as `participants.0` so that the empty objects are not removed (previously it would only return values like `participants.0.name`, which would cause the empty objects to be removed using lodash's `_pick`)
- Add tests for the omitExtraData bugs
- Run all the tests in Form_test under three conditions:
```{omitExtraData: false}```
```{omitExtraData: true}```
```{omitExtraData: true, liveOmit: true}```

Todo:
- add more exhaustive tests

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #1419, opened 2019-08-19
Fix omitExtraData bugs for nested empties and nonspecified objects

### LINKED ISSUE #1388: Can't enter in data with omitExtraData and liveOmit
### Description

Can't enter in data when `type: object` is not specified.

### Steps to Reproduce

1. Go to [this playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJmb28iOiJhc2QifSwic2NoZW1hIjp7InByb3BlcnRpZXMiOnsiZm9vIjp7InR5cGUiOiJzdHJpbmcifX0sImRlcGVuZGVuY2llcyI6eyJmb28iOnsib25lT2YiOlt7InByb3BlcnRpZXMiOnsiZm9vIjp7ImNvbnN0IjoiYSJ9LCJiYXIiOnsidHlwZSI6ImFycmF5IiwiaXRlbXMiOnsidHlwZSI6InN0cmluZyJ9LCJtaW5JdGVtcyI6M319fV19fX0sInVpU2NoZW1hIjp7ImZpcnN0TmFtZSI6eyJ1aTphdXRvZm9jdXMiOnRydWUsInVpOmVtcHR5VmFsdWUiOiIifSwiYWdlIjp7InVpOndpZGdldCI6InVwZG93biIsInVpOnRpdGxlIjoiQWdlIG9mIHBlcnNvbiIsInVpOmRlc2NyaXB0aW9uIjoiKGVhcnRoaWFuIHllYXIpIn0sImJpbyI6eyJ1aTp3aWRnZXQiOiJ0ZXh0YXJlYSJ9LCJwYXNzd29yZCI6eyJ1aTp3aWRnZXQiOiJwYXNzd29yZCIsInVpOmhlbHAiOiJIaW50OiBNYWtlIGl0IHN0cm9uZyEifSwiZGF0ZSI6eyJ1aTp3aWRnZXQiOiJhbHQtZGF0ZXRpbWUifSwidGVsZXBob25lIjp7InVpOm9wdGlvbnMiOnsiaW5wdXRUeXBlIjoidGVsIn19fX0=).
2. Turn on "omit extra data" and "live omit".
2. Type something in the textbox -- nothing comes up.

### Version

1.7.0

### LINKED ISSUE #1396: Array of objects data entry not working with liveOmit

Tonexus:

Fixes #1388 and fixes #1396. Alternative to #1418.

Modifies behavior of `toPathSchema` in `utils.js` (afaik the function is only used in omitting the extra data, let me know if this breaks anything else) to create `$name` key for arrays as well to be consistent with behavior for objects. Modifies behavior of `getFieldNames` in `Form.js` to also add a path to the array of paths if that path points to something empty. Refactors the surrounding code to be more concise and, hopefully, readable. Adds tests for `omitExtraData` and `liveOmit`.

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -79,7 +79,6 @@ export default class Form extends Component {
       formData,
       props.idPrefix
     );
-    const pathSchema = toPathSchema(retrievedSchema, "", definitions, formData);
     return {
       schema,
       uiSchema,
@@ -89,7 +88,6 @@ export default class Form extends Component {
       errors,
       errorSchema,
       additionalMetaSchemas,
-      pathSchema,
     };
   }
 
@@ -180,12 +178,19 @@ export default class Form extends Component {
     let newFormData = formData;
 
     if (this.props.omitExtraData === true && this.props.liveOmit === true) {
-      const newState = this.getStateFromProps(this.props, formData);
-
-      const fieldNames = this.getFieldNames(
-        newState.pathSchema,
-        newState.formData
+      const retrievedSchema = retrieveSchema(
+        this.state.schema,
+        this.state.schema.definitions,
+        formData
       );
+      const pathSchema = toPathSchema(
+        retrievedSchema,
+        "",
+        this.state.schema.definitions,
+        formData
+      );
+
+      const fieldNames = this.getFieldNames(pathSchema, formData);
 
       newFormData = this.getUsedFormData(formData, fieldNames);
       state = {
@@ -231,11 +236,22 @@ export default class Form extends Component {
     event.persist();
     let newFormData = this.state.formData;
 
-    const { pathSchema } = this.state;
-
     if (this.props.omitExtraData === true) {
-      const fieldNames = this.getFieldNames(pathSchema, this.state.formData);
-      newFormData = this.getUsedFormData(this.state.formData, fieldNames);
+      const retrievedSchema = retrieveSchema(
+        this.state.schema,
+        this.state.schema.definitions,
+        newFormData
+      );
+      const pathSchema = toPathSchema(
+        retrievedSchema,
+        "",
+        this.state.schema.definitions,
+        newFormData
+      );
+
+      const fieldNames = this.getFieldNames(pathSchema, newFormData);
+
+      newFormData = this.getUsedFormData(newFormData, fieldNames);
     }
 
     if (!this.props.noValidate) {
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -144,17 +144,14 @@ export default class Form extends Component {
   };
 
   getFieldNames = (pathSchema, formData) => {
-    const getAllPaths = (_obj, acc = [], paths = []) => {
+    const getAllPaths = (_obj, acc = [], paths = [""]) => {
       Object.keys(_obj).forEach(key => {
         if (typeof _obj[key] === "object") {
-          if (!paths.length) {
-            getAllPaths(_obj[key], acc, [key]);
-          } else {
-            let newPaths = paths.map(path => `${path}.${key}`);
-            getAllPaths(_obj[key], acc, newPaths);
-          }
-        } else if (key === "$name") {
+          let newPaths = paths.map(path => `${path}${key}.`);
+          getAllPaths(_obj[key], acc, newPaths);
+        } else if (key === "$name" && _obj[key] !== "") {
           paths.forEach(path => {
+            path = path.replace(/\.$/, "");
             const formValue = _get(formData, path);
             if (typeof formValue !== "object" || _isEmpty(formValue)) {
               acc.push(path);
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -2,6 +2,7 @@ import React, { Component } from "react";
 import PropTypes from "prop-types";
 import _pick from "lodash/pick";
 import _get from "lodash/get";
+import _isEmpty from "lodash/isEmpty";
 
 import { default as DefaultErrorList } from "./ErrorList";
 import {
@@ -149,17 +150,13 @@ export default class Form extends Component {
           if (!paths.length) {
             getAllPaths(_obj[key], acc, [key]);
           } else {
-            let newPaths = [];
-            paths.forEach(path => {
-              newPaths.push(path);
-            });
-            newPaths = newPaths.map(path => `${path}.${key}`);
+            let newPaths = paths.map(path => `${path}.${key}`);
             getAllPaths(_obj[key], acc, newPaths);
           }
         } else if (key === "$name") {
           paths.forEach(path => {
             const formValue = _get(formData, path);
-            if (typeof formValue !== "object") {
+            if (typeof formValue !== "object" || _isEmpty(formValue)) {
               acc.push(path);
             }
           });
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -150,11 +150,11 @@ export default class Form extends Component {
     const getAllPaths = (_obj, acc = [], paths = [""]) => {
       Object.keys(_obj).forEach(key => {
         if (typeof _obj[key] === "object") {
-          let newPaths = paths.map(path => `${path}${key}.`);
+          let newPaths = paths.map(path => `${path}.${key}`);
           getAllPaths(_obj[key], acc, newPaths);
         } else if (key === "$name" && _obj[key] !== "") {
           paths.forEach(path => {
-            path = path.replace(/\.$/, "");
+            path = path.replace(/^\./, "");
             const formValue = _get(formData, path);
             if (typeof formValue !== "object" || _isEmpty(formValue)) {
               acc.push(path);
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -156,6 +156,8 @@ export default class Form extends Component {
           paths.forEach(path => {
             path = path.replace(/^\./, "");
             const formValue = _get(formData, path);
+            // adds path to fieldNames if it points to a value
+            // or an empty object/array
             if (typeof formValue !== "object" || _isEmpty(formValue)) {
               acc.push(path);
             }
--- a/src/utils.js
+++ b/src/utils.js
@@ -880,7 +880,6 @@ export function toPathSchema(schema, name = "", definitions, formData = {}) {
     return toPathSchema(_schema, name, definitions, formData);
   }
   if (
-    schema.type === "array" &&
     schema.hasOwnProperty("items") &&
     Array.isArray(formData) &&
     formData.length > 0
@@ -893,7 +892,7 @@ export function toPathSchema(schema, name = "", definitions, formData = {}) {
         element
       );
     });
-  } else if (schema.type === "object" && schema.hasOwnProperty("properties")) {
+  } else if (schema.hasOwnProperty("properties")) {
     for (const property in schema.properties || {}) {
       pathSchema[property] = toPathSchema(
         schema.properties[property],
--- a/src/utils.js
+++ b/src/utils.js
@@ -873,7 +873,7 @@ export function toIdSchema(
 
 export function toPathSchema(schema, name = "", definitions, formData = {}) {
   const pathSchema = {
-    $name: name.replace(/\.$/, ""),
+    $name: name.replace(/^\./, ""),
   };
   if ("$ref" in schema || "dependencies" in schema) {
     const _schema = retrieveSchema(schema, definitions, formData);
@@ -887,7 +887,7 @@ export function toPathSchema(schema, name = "", definitions, formData = {}) {
     formData.forEach((element, i) => {
       pathSchema[i] = toPathSchema(
         schema.items,
-        `${name}${i}.`,
+        `${name}.${i}`,
         definitions,
         element
       );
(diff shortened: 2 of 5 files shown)


src/utils.js
```diff
@@ -873,43 +873,34 @@ export function toIdSchema(
 
 export function toPathSchema(schema, name = "", definitions, formData = {}) {
   const pathSchema = {
-    $name: name,
+    $name: name.replace(/\.$/, ""),
```
epicfaace: Why is this needed?

src/utils.js
```diff
@@ -873,43 +873,34 @@ export function toIdSchema(
...
+    formData.forEach((element, i) => {
+      pathSchema[i] = toPathSchema(
+        schema.items,
+        `${name}${i}.`,
+        definitions,
+        element
+      );
+    });
+  } else if (schema.hasOwnProperty("properties")) {
+    for (const property in schema.properties || {}) {
```
epicfaace: Can we just make this:

```suggestion
    for (const property in schema.properties) {
```

src/utils.js
```diff
@@ -873,43 +873,34 @@ export function toIdSchema(
 
 export function toPathSchema(schema, name = "", definitions, formData = {}) {
```
epicfaace: Are you changing the functionality of toPathSchema in some way or just refactoring it?

epicfaace:
> Modifies behavior of `getFieldNames` in `Form.js` to also add a path to the array of paths if that path points to something empty.

My PR #1418 also does this, although the way it does so isn't very good -- it relies on looking for the first "." in the string (which could be an issue with specially-named keys later). Your method seems better!

https://github.com/mozilla-services/react-jsonschema-form/pull/1418/files#diff-07a63ede6fe55fca89398d42da737369R170

src/utils.js
Tonexus: As a result of refactoring and because the initial call starts with name as an empty string, I ended up with a period prepended to all of the `$name`s, so I swapped it such that the period is appended, then removed when the `$name` value is set. e.g. here

https://github.com/Tonexus/react-jsonschema-form/blob/ff66b13dd0e74a36394f3e45e546137b024d709e/src/utils.js#L890

changes `` `${name}.${i}` `` to `` `${name}${i}.` ``.

Now that I think about it, it was honestly was kind of an arbitrary change. If you think the old style is more clear, I can change it back.

src/utils.js
Tonexus: I am changing the functionality of `toPathSchema`. Previously, if you had

```
schema = {
    type: "object",
    properties: {
        values: {
            type: "array",
            items: {
                type: "string"
            }
        }
    }
}

formData = {
    values: ["asdf", "ghkj"]
}
```

`toPathSchema(schema, formData = formData)` would not give the array a `$name` key, and would be

```
pathSchema = {
    $name: "",
    values: {
        0: {
            $name: "values.0"
        },
        1: {
            $name: "values.1"
        }
    }
}
```

My change results in

```
pathSchema = {
    $name: "",
    values: {
        $name: "values",
        0: {
            $name: "values.0"
        },
        1: {
            $name: "values.1"
        }
    }
}
```

which I believe is more consistent since there's a `$name` for ALL nesting. Again, I made this functionality change because it looks like `toPathSchema` is only used for omitting extra data and does not break anything else. This makes it easier to traverse the path schema and get the path if the array is empty.

src/utils.js
Tonexus: I think that should be fine. I'll check it out.

src/utils.js
epicfaace: I think you can change it to the old style, unless there are any additional advantages to this new style.

Tonexus:
@epicfaace I added your tests, and if I understand correctly, `describeRepeated` essentially parameterizes the tests and runs the 3 relevant combinations of `omitExtraData` and `liveOmit` for the enclosed tests. In using the function in `Form_test.js`, you repeat all of the current tests for the 3 combinations, right? So then there are no tests to make sure that `omitExtraData` actually omits the correct data yet, right?

epicfaace:
> @epicfaace I added your tests, and if I understand correctly, `describeRepeated` essentially parameterizes the tests and runs the 3 relevant combinations of `omitExtraData` and `liveOmit` for the enclosed tests. In using the function in `Form_test.js`, you repeat all of the current tests for the 3 combinations, right?

Yes.

> So then there are no tests to make sure that `omitExtraData` actually omits the correct data yet, right?

There are tests at the end of Form_test.js that are not within the `describeRepeated` block; I believe they include a test that tests that behavior of `omitExtraData`.

epicfaace:
By the way, the test that is failing is due to issues with form defaults when omitExtraData and/or liveOmit are true. I fixed them in the other PR with these changes: https://github.com/mozilla-services/react-jsonschema-form/pull/1418/files#diff-07a63ede6fe55fca89398d42da737369R249-R253

Tonexus:
Yeah I see what you are doing, but I think that that logic can be moved to `onChange` instead. Right now, you have to recompute `pathSchema` on submit to fix the bug, but it should really be computed whenever the form data changes, as the paths don't actually change on submit.

epicfaace:
@Tonexus ok, but if `omitExtraData` is true and `liveOmit` is false, then don't we only need to compute the `pathSchema` on submit? Computing it during `onChange` is unnecessary and might slow things down.

Tonexus:
@epicfaace That's a good point, if `liveOmit` is false, `pathSchema` probably doesn't need to be computed. However, it looks like `pathSchema` is already being computed on component creation and on props change without checking for `liveOmit`, and I wanted to be consistent with those cases. Should I change those cases to check for `omitExtraData` and `liveOmit` as well?

Also, right now it looks like `pathSchema` is not propagated to the state in `onChange` even if both `omitExtraData` and `liveOmit` are true, so I'll go ahead and change it so that `pathSchema` is propagated there as well.

epicfaace:
> @epicfaace That's a good point, if `liveOmit` is false, `pathSchema` probably doesn't need to be computed. However, it looks like `pathSchema` is already being computed on component creation and on props change without checking for `liveOmit`, and I wanted to be consistent with those cases. Should I change those cases to check for `omitExtraData` and `liveOmit` as well?

Yes, please do.

> Also, right now it looks like `pathSchema` is not propagated to the state in `onChange` even if both `omitExtraData` and `liveOmit` are true, so I'll go ahead and change it so that `pathSchema` is propagated there as well.

Sounds good, thanks!

Tonexus:
@epicfaace Actually, should `pathSchema` even be part of the state? It is only computed on submit or, with `liveOmit` true, whenever `formData` changes. And then it is only applied directly after those two cases, so it doesn't really need to persist.

On a side note, a lot of the logic in `onChange` and `onSubmit` seem redundant, maybe some of it can be refactored in another PR...

Tonexus:
@epicfaace I'm now having issues with the very last test case in the file.

https://github.com/Tonexus/react-jsonschema-form/blob/d23c6480175cf5eb1c041bdb1b7df67eabd0a515/test/Form_test.js#L2663-L2687

Can you explain the reasoning for that one? It looks like it is correctly applying the new `schema` to the new `formData`. Is it supposed to use the new `formData` with the old `schema` that then removes the `baz` property?

epicfaace:
> @epicfaace Actually, should `pathSchema` even be part of the state?

Yeah, you're right.

For the very last test case, you're right, I think the test case is wrong. I based it off an original test that tested based on an _external_ (props-based) formData change, which used `comp.componentWillReceiveProps`. I meant to change the test to be based on an _internal_ (user interaction-based) formData change, using `Simulate.change`. However, this test (as well as the one before it) mixes up both types of props changes in the same test case. Essentially, I meant to remove the `comp.componentWillReceiveProps` call, but neglected to.

It may be good to split this up; for each test in the final "Schema and formData updates" `describe` block, make one test that uses `comp.componentWillReceiveProps` to test an external formData change and another test that tests `Simulate.change` to test the internal formData change.

Tonexus:
@epicfaace Ok, made those changes. There are now 4 new tests for each combo of `omitExtraData` and `liveOmit`, and only one actually omits data on form change. Should be ready for review/merge now.

src/utils.js
epicfaace: Got it, thanks!

epicfaace:
Oh, I remember why pathSchema was initially part of the state. The idea was that it could be passed on to widgets/fields eventually, so that they would be able to know the "path" of the formData that they refer to. However, given that pathSchema is only computed live when `liveOmit` is true, there would be no point keeping it in the state for now. We can re-add pathSchema into the state later on (later PR) when also giving access to it for widgets/fields.

src/components/Form.js
```diff
@@ -148,23 +147,16 @@ export default class Form extends Component {
...
-          }
-        } else if (key === "$name") {
+          let newPaths = paths.map(path => `${path}.${key}`);
+          getAllPaths(_obj[key], acc, newPaths);
+        } else if (key === "$name" && _obj[key] !== "") {
           paths.forEach(path => {
+            path = path.replace(/^\./, "");
             const formValue = _get(formData, path);
-            if (typeof formValue !== "object") {
+            if (typeof formValue !== "object" || _isEmpty(formValue)) {
```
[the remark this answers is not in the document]
Tonexus: The empty array behavior is intentional. It fixes the behavior where if there is an empty array inside of an array, `_pick` will remove both arrays if the path to the inner (empty) array is not provided. Also, it looks like the default behavior if `liveOmit` is off is that deleting all array items does not remove the key/value pair containing the array, so `liveOmit` being true should behave the same, as the property is still defined in the schema. Maybe there needs to be a fix for the default behavior of removing the last element in an array in general.

In regards to the conditional, I added a comment, and I think that makes it clear enough. The code already uses the fact `typeof [] === "object"` in some other places as well, so I think that we should just assume that that fact is common knowledge.

Tonexus:
@epicfaace Added most of the requested changes. See review comment.

src/components/Form.js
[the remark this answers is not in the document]
epicfaace: Makes sense, thanks for the explanation!

MatinF:
Happy to do some tests of this to see if it resolves #1396 and #1425 - is the playground link updated?

epicfaace:
You can use the playground link for the deploy preview of this PR: https://deploy-preview-1419--rjsf.netlify.com/

PR #1429, opened 2019-08-22
add forwardRef to withTheme

### LINKED ISSUE #1405: withTheme does not forward ref
### Description

The `withTheme` HOC is not forwarding it's ref. Therefore it is e.g. not possible to [submit a form programmatically](https://react-jsonschema-form.readthedocs.io/en/latest/#submit-form-programmatically). React.forwardRef is available from 16.3.0, an implementation would look like this:

``` es6
import React, { forwardRef } from "react";
import Form from "./";

function withTheme(themeProps) {
  return forwardRef(({ fields, widgets, ...directProps }, ref) => (
    <Form
      {...themeProps}
      {...directProps}
      fields={{ ...themeProps.fields, ...fields }}
      widgets={{ ...themeProps.widgets, ...widgets }}
      ref={ref}
    />
  ));
}

withTheme.propTypes = {
  widgets: PropTypes.object,
  fields: PropTypes.object,
};

export default withTheme;

```

For now I will use my own HOC instead of withTheme. Please fix this as soon as you bump the React peer version. I will happily provide an PR then, testcase also kinda done:

``` es6
  it('should forward the ref', () => {
    let ref = undefined;
    const schema = {};
    const uiSchema = {};

    createComponent(withTheme({ ref: (form) => ref = form }), {
      schema,
      uiSchema,
    });

    expect(ref.submit).to.exist();
  });
```

zepatrik:

fixes #1405 
I didn't update the docs because I would assume this is actually default behavior.

### Checklist


timkindberg:
Why is there a bunch of material-ui files in this PR?? Shouldn't this be limited to just adding forwardRef??

zepatrik:
I didn't add them they probably where on master when I started this PR

zepatrik:
The whole commit history for this pr looks weird btw. Should I just transfer my changes to a fresh branch?

epicfaace:
Yeah, thanks for doing that @zepatrik

PR #1432, opened 2019-08-23
Fix #1431 - ie11 not able to submit

### LINKED ISSUE #1431: dispatchEvent submit does not work in ie11
### Prerequisites

### Description
In Internet Explorer when attempting to submit the form, the following error occurs:

image

### Steps to Reproduce
Submit a form.

#### Expected behavior
The form to submit.

#### Actual behavior
Error in ie11.

### Version
Latest

I'll probably have to come up with a way of polyfilling the Event on the form but I'm wondering why dispatchEvent was chosen in the first place over calling `.submit` directly on the form element?

martsie:
Whilst Internet Explorer supports the `Event` class it's read only and doesn't support instantiation therefore submitting a json schema form programatically results in a syntax error. I haven't found a way to polyfill the Event class.

The native browser `CustomEvent` class is polyfillable and does the same thing as the Event class.

Fixes #1431 -> see that issue for more information and the screenshot of the error in Internet Explorer.

### Checklist
  - [x] I've added and/or updated tests (no tests for this functionality)

### Overcoming this error without this fix
If you run into this issue before the PR gets merged and released you can extend the JSONForm class and override the submit method to use `CustomEvent` instead of `Event`.

```
import JSONForm from 'react-jsonschema-form';

export default class BrowserCompatibleJsonSchemaForm extends JSONForm<{}> {
  private formElement: HTMLFormElement|null = null;

  submit() {
    if (this.formElement) {
      this.formElement.dispatchEvent(new CustomEvent("submit", { cancelable: true }));
    }
  }
}
```

### Cross Browser
[x] Firefox (latest)
[x] Chrome (latest)
[x] Safari (latest)
[x] Internet Explorer (11)
[x] Edge (latest)


KKS1:
Thanks for looking into the fix. When is this targeted for merge and release creation ?

epicfaace:
Once we get the playground working on IE11 (granted, it does seem like the playground was never working on IE11). Otherwise, it's difficult to test this change.

epicfaace:
Ok, after adding the polyfill, the playground works in IE11, so I can verify it. Thanks for the patience!

PR #1438, opened 2019-08-28
fix: issue with false as formData on radio components

chanceaclark:

Passing `formData` as `false` for `type="boolean"` caused `formData` to come back as `undefined` on a non-nested JSONSchema object. To see it break, change `formData` to `false` in this example: [playground](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6dHJ1ZSwic2NoZW1hIjp7InRpdGxlIjoiQSBzaW5nbGUtZmllbGQgZm9ybSIsInR5cGUiOiJib29sZWFuIn0sInVpU2NoZW1hIjp7InVpOndpZGdldCI6InJhZGlvIn0sImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6dHJ1ZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==).

I didn't create an issue for this, but I still want to discuss if there was a point of not allowing `formData` to be `false`.

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-jsonschema-form",
-  "version": "1.7.0",
+  "version": "1.8.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/utils.js
+++ b/src/utils.js
@@ -273,7 +273,7 @@ export function getDefaultFormState(
     // Override schema defaults with form data.
     return mergeObjects(defaults, formData);
   }
-  if (formData === 0) {
+  if (formData === 0 || formData === false) {
     return formData;
   }
   return formData || defaults;
--- a/test/utils_test.js
+++ b/test/utils_test.js
@@ -49,6 +49,17 @@ describe("utils", () => {
         ).to.eql(0);
       });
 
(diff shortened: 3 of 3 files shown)


chanceaclark:
@epicfaace so I ran into an issue with setting `formData` to `null`. In one test we check for `onChange` to be called when the `formData` is set to `null` so that we can reset the form. So in the case that we check for `null` or essentially most falsy values, `componentWillReceiveProps` on the `Form` component will not throw the `onChange` event since `deepEquals` checks for `null`. 

And if you follow the link in the comments where the deepEquals function was extracted they mention this:
> `===` only tests objects and functions by reference. `null` is an object. Any pairs of identical entities failing this test are therefore objects (including `null`), which need to be recursed into and compared attribute by attribute.

The issue is `deepEquals` treats `null` as an object and not a value, so I'm not sure how to get around this, any ideas? We may also want to push it to another PR since it is a little more involved?

epicfaace:
Yeah, let's push it to another issue/PR given that it's not a simple fix.

PR #1441, opened 2019-08-29
Add propertyNames enum and pattern validation (#1250)

### LINKED ISSUE #1250: additionalProperties form - enum for key
### Prerequisites

### Description

Would it be possible to add a feature that would enable the user to set a given set of all possible keys of `additionalProperties`?

At the moment, I validate the keys with `propertyNames` which works, but it would be much cleaner for the user to have it set as an enum.

#### Expected behavior

`additionalProperties` are rendered with a dropdown key selection if enum given.

#### Actual behavior

`additionalProperties` have to options for key customizations and `propertyNames` are ignored in the rendering phase (but used for validation).

### Version

Latest.

v1vendi:

JSON Schema Draft 6 supports validation propertyNames validation
Fixes #1250 

### Checklist

### DIFF
--- a/playground/samples/index.js
+++ b/playground/samples/index.js
@@ -20,6 +20,7 @@ import alternatives from "./alternatives";
 import propertyDependencies from "./propertyDependencies";
 import schemaDependencies from "./schemaDependencies";
 import additionalProperties from "./additionalProperties";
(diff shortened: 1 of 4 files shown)


v1vendi:
Removing it to recreate without merge PR

PR #1443, opened 2019-09-01
Small functional changes

httpsOmkar:

Some lint issues and missing hasOwnProperty while iterating over `for in`.

### Checklist

* [*] **I'm adding or updating code**

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,66 +1,66 @@
-import React from "react";
-import * as ReactIs from "react-is";
-import fill from "core-js/library/fn/array/fill";
-import validateFormData, { isValid } from "./validate";
+import React from 'react';
+import * as ReactIs from 'react-is';
+import fill from 'core-js/library/fn/array/fill';
+import validateFormData, { isValid } from './validate';
 
-export const ADDITIONAL_PROPERTY_FLAG = "__additional_property";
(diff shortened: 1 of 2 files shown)


epicfaace:
Thanks! Can you update the title to be more specific as to what it does (the commit summary when merging will have the title name by default)

src/utils.js
```diff
@@ -901,14 +901,16 @@ export function toPathSchema(schema, name = "", definitions, formData = {}) {
...
     for (const property in schema.properties) {
-      pathSchema[property] = toPathSchema(
-        schema.properties[property],
-        `${name}.${property}`,
-        definitions,
-        // It's possible that formData is not an object -- this can happen if an
-        // array item has just been added, but not populated with data yet
-        (formData || {})[property]
-      );
+      if (schema.properties.hasOwnProperty(property)) {
```
epicfaace: Is there an example of a schema in which the original code would fail? Could you add a test with it?

httpsOmkar: https://stackoverflow.com/questions/12735778/for-in-and-hasownproperty

epicfaace:
@httpsOmkar it's been a while since this PR was updated, so I'm going to close it for now. Feel free to make a new PR as long as it 1) fixes a specific issue that is described in the PR description and 2) adds test cases that cover what you're fixing.

Issue #583, opened 2017-05-18
Converting source to TypeScript?


KurtPreston:
Would you consider a Pull Request converting react-jsonschema-form to TypeScript?

zhang-ning:
+1 TypeScript

glasserc:
Not at the present time, although I guess that could change later. (If it does, I'll reopen.)

loganvolkers:
@glasserc interesting discussions about the Jest typescript migration https://news.ycombinator.com/item?id=18918038

epicfaace:
I wouldn't mind, honestly, though it would be a large rewrite to do so.

thijssteel:
I'd be willing to help with this.
However, i suck at everything webpack and compiler related. If someone has time to get me started with that I believe a full conversion should be possible in a couple of days work.

chanceaclark:
This is the :dream:. Let me know if you need some help as well.

sbusch:
@epicfaace is this a goal for v2?

We are heavily using TypeScript and are currently maintaining a patched set of TS definitions to have *some* (quite inexact) type safety for our custom fields and widgets. A native TS implementation with good type coverage will be a huge step not only for us, but IMO for the project (stability, types as documentation, confidence when doing some refactorings, see my latest comment at #1286)

So I also offer my help for this migration, and support in case of problems.

epicfaace:
Yeah, let's make it a goal for v2. @sbusch could you send over whichever of your TS definitions might be helpful? I think starting with those as well as the DefinitelyTyped definitions would give us a good starting point with which to add the rest of the types needed.

thijssteel:
https://github.com/wix-incubator/as-typed

do you guys think this might be useful?

jacqueswho:
we could possibly use [tsdx](https://tsdx.io/)

heath-freenome:
An effort has begun to convert the `@rjsf/core` package to typescript in the upcoming v5 version

KurtPreston:
@heath-freenome Nice! I am no longer using this project, but it's great to hear that the TypeScript conversion is underway.

ksbrar:
@heath-freenome If you're still looking for extra hands, would be interested in participating!

heath-freenome:
@ksbrar We've managed to complete the conversion of the `core` package to Typescript but would love help with the `antd` theme

heath-freenome:
We are 100% typescript in the 5.x

PR #1446, opened 2019-09-04
Typescript webpack setup

epicfaace:

For #583 and @thijssteel. Sets up webpack configuration needed to use typescript in the rjsf source. I've converted Form.js -> Form.tsx (though have not really added the types) as an example.

When you convert the source to typescript, could you rename files that contain React components to ".tsx" and rename other files to ".ts"? Best of luck and thanks for helping 😄


package.json
```diff
@@ -61,7 +61,8 @@
     "@babel/plugin-transform-runtime": "^7.4.4",
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
-    "@babel/register": "^7.4.4",
+    "@babel/preset-typescript": "^7.3.3",
+    "./babel-register.js": "^7.4.4",
```
epicfaace: Not sure why this popped up. This should be a local file...

epicfaace: Fixed.

src/components/Form.tsx
```diff
@@ -18,7 +18,12 @@ import {
 } from "../utils";
 import validateFormData, { toErrorList } from "../validate";
 
-export default class Form extends Component {
+interface IFormProps {
```
chanceaclark: Has there been any discussion to export typings with the build or will you maintain typings within [@types/react-jsonschema-form](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/a819fdd3bc950160780f1e6e69c55a2bac8940f0/types/react-jsonschema-form)?

To add onto that, has there been any discussion regarding naming conventions for types?

epicfaace: Not yet. It would be great to start one!

What would be the advantages/disadvantages of exporting typings with the build vs maintaining typings within @types/react-jsonschema-form? Do other libraries based in typescript typically do the former?

chanceaclark: > What would be the advantages/disadvantages of exporting typings with the build vs maintaining typings within @types/react-jsonschema-form?

A big drawback with using DefinitelyTyped is that it's community maintained and can sometimes get out of sync with the actual library. That repository is mainly for JS packages that don't have typings for TS projects. And sometimes DefinitelyTyped can take a while to merge PR's so you'd have to wait for a release before you could use new typings. 

With that in mind, it's usually better to export typings with the package so your types are always 1:1 and you don't have to maintain two separate packages.

> Do other libraries based in typescript typically do the former?

Usually yeah. Check out [nestjs](https://github.com/nestjs/nest) or even vscode. Go explore some more TS packages here in [GH](https://github.com/topics/typescript)

epicfaace:
superseded by #1532

PR #1449, opened 2019-09-05
fix: playground links directing to old playground

chanceaclark:

Updating documentation to point to the right playground.

### Checklist


epicfaace:
Thanks a lot for this!

PR #1454, opened 2019-09-08
Removed setState from util

mirajp:
Removed setSstate from util

From #1297:

> Remove the util setState call from Form.onChange because it sets the state too early or something? Hard to describe, I came upon this fix (for #1281) after trying a bunch of things like using a leaner `componentDidUpdate` and removing `shouldComponentUpdate` (so it uses the default React Component function which always returns true) instead of the existing `componentWillReceiveProps` and `shouldComponentUpdate`. Note that removing `componentWillReceiveProps` breaks a lot of existing tests because for some reason they rely on `componentWillReceiveProps` (not sure why tests call that function directly, but I'm just a little more used to Jest than the raw 'react-addons-test-utils' library), so that's not an option anyway.
\
> I'm not sure how to add tests for the bug I'm trying to fix, so feel free to edit this PR.
\
> If this is related to existing tickets, include links to them as well.
#1281

### Checklist


epicfaace:
Actually, `master` now has the v2 code, so I've changed the base.

mirajp:
@epicfaace good catch. I removed `safeRenderCompletion` usages. Made mistake with commit message because I was rereading #1197 -- I believe setImmediate is still necessary to wait for DOM changes, though perhaps setTimeout could work in its place, e.g.:

```
/* ArrayField_test.js */

// existing setImmediate usage
new Promise(setImmediate).then(() =>
    expect(comp.state.formData).eql([
        'data:text/plain;name=file1.txt;base64,x=',
        'data:text/plain;name=file2.txt;base64,x=',
    ])
);

// can be replaced with:
setTimeout(() => {
    expect(comp.state.formData).eql([
        'data:text/plain;name=file1.txt;base64,x=',
        'data:text/plain;name=file2.txt;base64,x=',
    ]);
}, 100);
```

mirajp:
@epicfaace not sure why Travis build agents are failing -- all tests pass locally for me (Windows + WSL).

epicfaace:
Can you merge the latest `master` into this branch? That will take care of most of the failures. There's still a single build failure, but that's on node 6 and not sure why it's happening (https://github.com/rjsf-team/react-jsonschema-form/runs/217099766)

mirajp:
@epicfaace I actually have quite a few test failures (locally) once I've merged in master (node v10.16.0)... If only node v6 is failing, I don't think it's too much of a big deal as it's at EOL (https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true)? I recall getting random issues with node v6 at work and I pretty much just gave up trying to support it and changed all pipelines to only use v10 ;)

chanceaclark:
If Node 6 is out of support, rjsf should at least support Node 8+. Is there any benefit to supporting 6?

With @epicfaace most recent commit onto master the only issue I can see is on Node 6 with an [`AssertionError`](https://travis-ci.com/rjsf-team/react-jsonschema-form/jobs/232576062#L1953) 

>Anyway, here are some of the failures I'm seeing locally now (in master and now my branch which has master): 

For clarification, are you using node 6 with these errors?

epicfaace:
> Also, shouldn't the pipeline be building/testing the attempted merge (i.e. the projected state with master + feature)? I don't really get how the PR build produced a failure when neither master nor the feature branch without master failed...

Master _is_ failing, although not with the errors you mentioned. Master failed in the first place because I cherry-picked #1408 from `v2` to `master` -- I probably should have made a PR to make sure the tests passed first.

epicfaace:
> If Node 6 is out of support, rjsf should at least support Node 8+. Is there any benefit to supporting 6?

Yeah, I think you're right. It was preferable to keep support for Node 6 in 1.x to keep backwards compatibility, but at this point we can change it to Node 8+.

chanceaclark:
I'll go ahead and submit a PR then.

test/test_utils.js
```diff
@@ -14,7 +14,7 @@ export function createComponent(Component, props) {
 }
 
 export function createFormComponent(props) {
-  return createComponent(Form, { ...props, safeRenderCompletion: true });
+  return createComponent(Form, { ...props });
```
epicfaace:
```suggestion
  return createComponent(Form, props);
```

epicfaace:
Thanks @mirajp !

PR #1462, opened 2019-09-11
chore: remove node version < 8 from ci

chanceaclark:

From discussion in #1454, removing node versions less than 8 due to no longer support.


epicfaace:
Thanks!

PR #1476, opened 2019-09-29
Rewrite mergeSchemas to fix schema dependencies merging

### LINKED ISSUE #998: Schema Dependencies Merged Incorrectly

epicfaace:

Fixes #998. This makes sure that when dependencies are merged with the main schema, only arrays under the key `required` are merged. Other arrays are overwritten.

With this PR, `mergeSchemas` uses lodash's `union` operator to merge required arrays, so that the arrays do not have duplicate values when merged.

### Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -775,8 +775,9 @@ export function mergeSchemas(obj1, obj2) {
     if (obj1 && obj1.hasOwnProperty(key) && isObject(right)) {
       acc[key] = mergeSchemas(left, right);
     } else if (
-      (getSchemaType(obj1 || {}) === "object" ||
-        getSchemaType(obj2 || {}) === "object") &&
+      obj1 &&
+      obj2 &&
+      (getSchemaType(obj1) === "object" || getSchemaType(obj2) === "object") &&
       key === "required" &&
       Array.isArray(left) &&
       Array.isArray(right)
--- a/src/utils.js
+++ b/src/utils.js
@@ -67,6 +67,7 @@ export function getDefaultRegistry() {
   };
 }
 
+/* Gets the type of a given schema. */
 export function getSchemaType(schema) {
   let { type } = schema;
 
@@ -774,6 +775,8 @@ export function mergeSchemas(obj1, obj2) {
     if (obj1 && obj1.hasOwnProperty(key) && isObject(right)) {
       acc[key] = mergeSchemas(left, right);
     } else if (
+      (getSchemaType(obj1 || {}) === "object" ||
+        getSchemaType(obj2 || {}) === "object") &&
       key === "required" &&
       Array.isArray(left) &&
       Array.isArray(right)
--- a/src/utils.js
+++ b/src/utils.js
@@ -2,6 +2,7 @@ import React from "react";
 import * as ReactIs from "react-is";
 import fill from "core-js/library/fn/array/fill";
 import validateFormData, { isValid } from "./validate";
+import { union } from "lodash";
 
 export const ADDITIONAL_PROPERTY_FLAG = "__additional_property";
 
@@ -763,7 +764,8 @@ function withExactlyOneSubschema(
 // Recursively merge deeply nested schemas.
 // The difference between mergeSchemas and mergeObjects
 // is that mergeSchemas only concats arrays for
-// values under the "required" keyword.
+// values under the "required" keyword, and when it does,
+// it doesn't include duplicate values.
 export function mergeSchemas(obj1, obj2) {
   var acc = Object.assign({}, obj1); // Prevent mutation of source object.
   return Object.keys(obj2).reduce((acc, key) => {
@@ -776,7 +778,9 @@ export function mergeSchemas(obj1, obj2) {
       Array.isArray(left) &&
       Array.isArray(right)
     ) {
-      acc[key] = left.concat(right);
+      // Don't include duplicate values when merging
+      // "required" fields.
+      acc[key] = union(left, right);
     } else {
       acc[key] = right;
     }
--- a/src/utils.js
+++ b/src/utils.js
@@ -760,8 +760,28 @@ function withExactlyOneSubschema(
   );
 }
 
-function mergeSchemas(schema1, schema2) {
-  return mergeObjects(schema1, schema2, true);
+// Recursively merge deeply nested schemas.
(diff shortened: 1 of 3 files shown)


test/utils_test.js
```diff
@@ -1058,6 +1059,94 @@ describe("utils", () => {
     });
   });
 
+  describe("mergeSchemas()", () => {
+    it("should't mutate the provided objects", () => {
```
```suggestion
    it("shouldn't mutate the provided objects", () => {
```

src/utils.js
```diff
@@ -760,8 +760,28 @@ function withExactlyOneSubschema(
...
+// values under the "required" keyword.
+export function mergeSchemas(obj1, obj2) {
+  var acc = Object.assign({}, obj1); // Prevent mutation of source object.
+  return Object.keys(obj2).reduce((acc, key) => {
+    const left = obj1 ? obj1[key] : {},
+      right = obj2[key];
+    if (obj1 && obj1.hasOwnProperty(key) && isObject(right)) {
+      acc[key] = mergeSchemas(left, right);
+    } else if (
+      key === "required" &&
```
edi9999: Can you add some kind of check for the parent property ?

I mean techically speaking, only `properties.required` field should be overwritten, not all fields with `required` (I know it is most unlikely, so if you think this test is enough that would be ok for me)

travisdahl:
pulled it down and tested in our project, fixes our problem 👍

tsunamiq:
it's looking good! Appreciate all the work on this; you guys are a life saver! 

@epicfaace do you have an idea of when you'll have time to merge the pr? Just curious to know so I can work on scheduling on my side. Again, appreciate all the help with this issue!

src/utils.js
epicfaace: I've added a check to make sure that `obj1` or `obj2` are objects. Not sure if that's the right logic, though, feedback welcome!

epicfaace:
@edi9999 this PR should be ready for review now. I made a few changes:
- add an integration test
- make sure it uses lodash's `union` operator to merge `required` arrays, so that the arrays do not have duplicate values when merged
- merge the `required` attribute only if one of the schemas is an object type

src/utils.js
edi9999: I've tried with this on your PR, and it still works : 

```
obj1 && obj2 && getSchemaType(obj1) === "object" &&
        getSchemaType(obj2) === "object"
```

I think this is more correct to check for both the schematypes of both objects.

edi9999:
The integration test looks good to me !

src/utils.js
edi9999: you've written `getSchematype(obj1) === "object" || getSchemaType(obj2) === "object"`.

Is the `or` on purpose ? 

To me it seems that it should be an `and` (`&&`), that way only two objects schemas of type object which have a required property will be merged that way ?

epicfaace: I was thinking about the case in which you want to merge, say, `{properties: {...}, required: ["a"]}` and `{title: "a"}`. In this case, the second schema is not unambiguously an object, although the first schema is. It would still be nice to merge the two schemas together, though. This would only work if we use `||` instead of `&&`.

edi9999: Ok

PR #1482, opened 2019-10-15
Use setTimeout instead of setImmediate

edi9999:

At one place in the library code, we use setImmediate. 
However this function is not standard : 

> This method is not expected to become standard, and is only implemented by recent builds of Internet Explorer and Node.js 0.10+. It meets resistance both from Gecko (Firefox) and Webkit (Google/Apple).
[Source](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate)

Since it is not standard, the setImmediate is transformed by babel to : `var _setImmediate2 = _interopRequireDefault(require("@babel/runtime-corejs2/core-js/set-immediate"));` which adds a lot of unneeded code to the build.

In my application tests, we use react-json-schema-form and are mocking the time with lolex, however, this "polyfil" by babel seems to not react to lolex.

setTimeout(function, 0) has the same behavior, it brakes no tests.

It would be great to release a patch version with those changes :-)

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-jsonschema-form",
-  "version": "1.7.0",
+  "version": "1.8.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/playground/app.js
+++ b/playground/app.js
@@ -122,7 +122,7 @@ class GeoPosition extends Component {
   onChange(name) {
     return event => {
       this.setState({ [name]: parseFloat(event.target.value) });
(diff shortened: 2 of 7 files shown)


edi9999:
@glasserc please let me know if you have any comments.

edi9999:
Hello @epicfaace , can you publish a patch version (1.8.1) which contains this fix ? It would be helpful for me :-)

edi9999:
up also @glasserc

edi9999:
up also @glasserc

PR #1485, opened 2019-10-16
fix: update arrays correctly when changing index

### LINKED ISSUE #1156: Rename autocomplete to autoComplete
### Prerequisites

### Description

We should rename the `autocomplete` attribute on `<Form>` to `autoComplete`. This is because that's how React does it on the actual `<form>` element, so it will be a more familiar attribute name for React developers.

Perhaps we could support both options and deprecate `autocomplete`.

https://github.com/mozilla-services/react-jsonschema-form/blob/51caab66b3895601d1a50ad5e2196948749d325b/src/components/Form.js#L208

### LINKED ISSUE #1479: onChange callback gets obsolete after an array item is removed
### Prerequisites

### Description

In react < 16.4 removing an array item due to using generated keys correctly doesn't cause unaffected items to be rerendered, so the onChange callback keeps an obsolete index.

(Due to this bug https://github.com/rjsf-team/react-jsonschema-form/issues/1427 react > 16.4 regenerates keys preventing this issue) 

### Steps to Reproduce

1. Add an array item to have at least 3 items (Reproducible in react-jsonschema-form playground)
2. Remove an item to have at least 2 items left next to the removed one
3. Type anything in the item that replaced the removed one. 

#### Expected behaviour

Typed text appears in the field it was typed

#### Actual behaviour

Typed text appears in the next field it was actually typed

### Version
1.8.0^

SuriGill:
SchemaField needs to be updated when items are removed or repositioned in the array with the correct indices. If the index is not passed down from ArrayField, shouldComponentRender in SchemaField returns false as it detects nothing has changed for the array children.

[Please describe them here]
fixes #1479 

### Checklist

### DIFF
--- a/src/components/fields/ArrayField.js
+++ b/src/components/fields/ArrayField.js
@@ -750,6 +750,7 @@ class ArrayField extends Component {
     return {
       children: (
         <SchemaField
+          index={index}
           schema={itemSchema}
           uiSchema={itemUiSchema}
           formData={itemData}
--- a/test/ArrayField_test.js
+++ b/test/ArrayField_test.js
@@ -616,6 +616,23 @@ describe("ArrayField", () => {
       expect(inputs[0].value).eql("bar");
     });
 
+    it("should delete item from list and correct indices", () => {
+      const { node } = createFormComponent({
+        schema,
+        formData: ["foo", "bar", "baz"],
+      });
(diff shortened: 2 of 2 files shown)


epicfaace:
Thanks @SuriGill !

PR #1491, opened 2019-10-21
fix: Update anyOf schema to correctly update items in an array

### LINKED ISSUE #1486: anyOf field incorrectly updating after change
### Prerequisites

- [] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net/n1k0/f2y3fq7L/6/) or a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### Description

Since updating to react 16, the anyOf field doesn't correctly update after receiving props. It will override the selected option incorrectly. This can be resolved by migrating to the componentDidUpdate method instead of the componentWillReceiveProps that's currently marked unsafe. componentWillReceiveProps will update state synchronously and componentDidUpdate will update state asynchronously.

### Steps to Reproduce

1. Go to anyOf field in storybook of the latest version of master
2. Add a new entry and keep it selected as option 1 and enter some text in the input field
3. Add another entry and select option 2 and enter some text in the input field
4. Move the option 2 up in the array.
anyOfError

#### Expected behavior

The array should have the two positions switched.

#### Actual behavior

The array only has one position switched correctly, while the other has the wrong selected option even though the state has been changed.

### Version

v.1.8.0

SuriGill:

componentWillReceiveProps is no longer recommended by React after updating to React 16 as it has some design flaws. Instead, I've updated to a componentDidUpdate that will change after formData changes specifically. I've added a test to demonstrate the bug that was causing an update twice and selectedOption to become out of sync when moving indices in an array using anyOf.

fixes first issue in #1486 

### Checklist

### DIFF
--- a/src/components/fields/MultiSchemaField.js
+++ b/src/components/fields/MultiSchemaField.js
@@ -8,6 +8,7 @@ import {
   retrieveSchema,
   getDefaultFormState,
   getMatchingOption,
+  deepEquals,
 } from "../../utils";
 
 class AnyOfField extends Component {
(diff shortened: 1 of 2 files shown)


src/components/fields/MultiSchemaField.js
```diff
@@ -21,17 +22,24 @@ class AnyOfField extends Component {
...
+    ) {
+      const matchingOption = this.getMatchingOption(
+        this.props.formData,
+        this.props.options
+      );
 
-    if (matchingOption === this.state.selectedOption) {
-      return;
-    }
+      if (!prevState || matchingOption === this.state.selectedOption) {
```
epicfaace: Why did you add the `!prevState` check @SuriGill ? It seems like the tests already pass without it.

PR #1505, opened 2019-11-08
fix: schema of type string shows defaults when cleared

### LINKED ISSUE #1369: Root schema of type string shows defaults when cleared
### Prerequisites

### Description

When the root schema is of type `string`, with `ui:emptyValue` set to `""`, clearing the text box does not properly set the formData to an empty string; thus, default values show back up when clearing it.

### Steps to Reproduce

1. Go to this [playground link](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6ImEiLCJzY2hlbWEiOnsidHlwZSI6InN0cmluZyIsImRlZmF1bHQiOiJhIn0sInVpU2NoZW1hIjp7InVpOmVtcHR5VmFsdWUiOiIifX0=).
2. Type in something.
3. Clear the textbox.

#### Expected behavior

The textbox is cleared.

#### Actual behavior

The textbox is not cleared, and the default value is set to "a".

To fix this, we need to make sure that when a text box is cleared, the formData becomes `""` instead of `undefined`. This is currently how string inputs work when they are properties of an object; this behavior should be consistent even when a schema is of type `string`.

### Version

1.7.0

Kexin-Li:

Fixed #1369 

### Checklist


src/utils.js
```diff
@@ -279,7 +279,7 @@ export function getDefaultFormState(
     // Override schema defaults with form data.
     return mergeObjects(defaults, formData);
   }
-  if (formData === 0 || formData === false) {
+  if (formData === 0 || formData === false || formData === "") {
```
epicfaace: @Kexin-Li thanks for your change! Can you add a test for this?

It can be modeled off of this -- https://github.com/rjsf-team/react-jsonschema-form/blob/18b2ec42c9cfc4aad2cf6996342c8edde979ee68/test/StringField_test.js#L76 -- basically, you want to test that when a textbox is cleared, its value should also be "" instead of the default value.

Kexin-Li: Sure!

test/StringField_test.js
```diff
@@ -146,6 +146,21 @@ describe("StringField", () => {
       expect(comp.state.formData).eql("default");
     });
 
+    it("should handle an empty string change event with textbox is cleared", () => {
```
epicfaace:
```suggestion
    it("should handle an empty string change event even with defaults set", () => {
```

epicfaace:
```suggestion
    it("should handle an empty string change event with defaults set", () => {
```

PR #1508, opened 2019-11-11
Remove ui:order console warning about extraneous properties

aliserin:

When we need to use ui:order with oneOf, "uiSchema order list contains extraneous 'some property'" warning is shown. It is a problem in production. I added a property (`isDebugMode`) into `ObjectField` so that it is shown only when it is debug mode is enabled.  

### Checklist

### DIFF
--- a/packages/core/src/components/fields/ObjectField.js
+++ b/packages/core/src/components/fields/ObjectField.js
@@ -67,7 +67,6 @@ class ObjectField extends Component {
     required: false,
     disabled: false,
     readonly: false,
-    isDebugMode: false,
   };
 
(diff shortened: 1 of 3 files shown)


epicfaace:
@aliserin I'd actually be fine with removing this warning entirely, as this commonly happens when having to use a uiSchema for a schema with dependencies / oneOf -- this can go in v2.0 because it's a backwards incompatible change. Could you update the PR to do that instead?

aliserin:
@epicfaace I removed the warning completely. :+1:

epicfaace:
Thanks!

PR #1511, opened 2019-11-12
fix: form to accept react component

chanceaclark:

In development, we weren't able to pass a react component into the `tagName` prop. This is valid in React but was unable to test in development due to `PropTypes`.

It also looked like some packages weren't properly locked, so I've updated those.

_Side Note:_ We may want to look into migrating to `yarn`. Just a 💭 

### Checklist

### DIFF
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -428,7 +428,7 @@ if (process.env.NODE_ENV !== "production") {
     onSubmit: PropTypes.func,
     id: PropTypes.string,
     className: PropTypes.string,
-    tagName: PropTypes.string,
+    tagName: PropTypes.elementType,
     name: PropTypes.string,
     method: PropTypes.string,
     target: PropTypes.string,
--- a/packages/core/docs/form-customization.md
+++ b/packages/core/docs/form-customization.md
@@ -756,3 +756,14 @@ It's possible to change the default `form` tag name to a different HTML tag, whi
   tagName="div"
 />
 ```
+
+You can also provide a class/function component.
+
+
+```jsx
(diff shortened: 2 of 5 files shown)


packages/core/src/components/Form.js
```diff
@@ -428,7 +428,7 @@ if (process.env.NODE_ENV !== "production") {
     onSubmit: PropTypes.func,
     id: PropTypes.string,
     className: PropTypes.string,
-    tagName: PropTypes.string,
+    tagName: PropTypes.oneOfType([PropTypes.string, PropTypes.elementType]),
```
chanceaclark: I think I should be able to make this just `PropTypes.elementType` 🤔

epicfaace: Hmm, does PropTypes.elementType include strings?

chanceaclark: Yeah, it's a `ReactComponentLike`:
```ts
export type ReactComponentLike =
    | string
    | ((props: any, context?: any) => any)
    | (new (props: any, context?: any) => any);
```

chanceaclark:
@epicfaace I agree, but if we wanted to change it, it would be a `BREAKING CHANGE`. I guess since v2 is in alpha it would be fine to add a `BC`?

epicfaace:
Yeah, true. I'll go ahead and merge this for now, though. Do you have any suggestions for what a better prop name might be?

chanceaclark:
We could go for the `as` prop. It's used in some libraries, like `styled-components`. Material-ui and Atlaskit uses the `component` prop, which IMO, prefer.

PR #1513, opened 2019-11-15
Pass state formData to array field template

olzraiti:

Currently the `items` and `formData` in array field template props are out of sync, since `ArrayField` passes the `items` from the state and `formData` from the props

### Checklist


epicfaace:
Thanks for the PR. It's not clear to me when this out-of-sync behavior could cause a problem, though, given that `onChange(keyedToPlainFormData(newKeyedFormData))` is called by the arrayfield anyway. Could you come up with a situation in which the current behavior causes a bug and then add a test case for it?

olzraiti:
I added a test case.

The bug happens when you click on a custom ArrayFieldTemplate's add button, and then it renders the items from state, which contain one more item than the `props.formData` since the `onChange` hasn't bubblet yet to the Form root and then propagated back to the ArrayFieldTemplate component.

I noticed that the bug also effects `canAdd` and `canMoveDown` props.

olzraiti:
I made the test case more clear.

Another, more cleaner approach would be providing the `formData` prop to the template `items` prop instead of the template root props. However, BW compability is broken right now. The latest release is marked 2.0, so maybe breaking BW compability would be ok. In that case I would recommend removing the `formData` prop completely from the template root props and add it to the `items` props.

packages/core/test/ArrayFieldTemplate_test.js
```diff
@@ -325,5 +326,27 @@ describe("ArrayFieldTemplate", () => {
         ArrayFieldTemplate,
       });
     });
+
+    it("should pass formData", () => {
```
epicfaace:
```suggestion
    it("should pass formData so it is in sync with items", () => {
```

epicfaace:
Thanks, that makes sense.

Do you know why it is the case that "the `onChange` hasn't bubbled yet to the Form root and then propagated back to the ArrayFieldTemplate component"?

Also, yes, we can make backwards incompatible changes. Why would you recommend "removing the `formData` prop completely from the template root props and add it to the `items` props"? (given that `formData` will be the same throughout the array, shouldn't it just be a root-level prop?)

olzraiti:
> Do you know why it is the case that "the onChange hasn't bubbled yet to the Form root and then propagated back to the ArrayFieldTemplate component"?

Because ArrayField's `onChange` sets `state` before calling the parent `onChange`. 

> Also, yes, we can make backwards incompatible changes. Why would you recommend "removing the formData prop completely from the template root props and add it to the items props"? (given that formData will be the same throughout the array, shouldn't it just be a root-level prop?)

After thinking through this again, I wouldn't remove the `formData` from the template root props since the root might need it...

jacqueswho:
this may fix the issue I had with my array Field template #1572

Issue #632, opened 2017-07-13
Support for examples . http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.4


mallikarjuna91:
There is no support for expamples which is supported by json schema.

http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.4

PR #1517, opened 2019-11-22
Support for "examples" keyword

89erik:

The [examples](https://json-schema.org/understanding-json-schema/reference/generic.html) keyword is not currently supported, as noted in [issue 632](https://github.com/rjsf-team/react-jsonschema-form/issues/632). This pull requests adds support for this using the [HTML datalist tag](https://www.w3schools.com/tags/tag_datalist.asp).

### Checklist

### DIFF
--- a/packages/core/src/components/widgets/BaseInput.js
+++ b/packages/core/src/components/widgets/BaseInput.js
@@ -61,7 +61,7 @@ function BaseInput(props) {
     return props.onChange(value === "" ? options.emptyValue : value);
   };
 
-  return (
+  return [
     <input
       className="form-control"
       readOnly={readonly}
@@ -69,11 +69,19 @@ function BaseInput(props) {
       autoFocus={autofocus}
       value={value == null ? "" : value}
       {...inputProps}
+      list={schema.examples ? `examples_${inputProps.id}` : null}
       onChange={_onChange}
       onBlur={onBlur && (event => onBlur(inputProps.id, event.target.value))}
       onFocus={onFocus && (event => onFocus(inputProps.id, event.target.value))}
-    />
-  );
+    />,
+    schema.examples ? (
+      <datalist id={`examples_${inputProps.id}`}>
+        {schema.examples.concat([]).map(example => (
+          <option key={example} value={example} />
+        ))}
+      </datalist>
+    ) : null,
+  ];
 }
 
 BaseInput.defaultProps = {
--- a/packages/core/src/components/widgets/BaseInput.js
+++ b/packages/core/src/components/widgets/BaseInput.js
@@ -76,11 +76,13 @@ function BaseInput(props) {
     />,
     schema.examples ? (
       <datalist id={`examples_${inputProps.id}`}>
-        {schema.examples
-          .concat(schema.default ? [schema.default] : [])
-          .map(example => (
-            <option key={example} value={example} />
-          ))}
+        {[
+          ...new Set(
+            schema.examples.concat(schema.default ? [schema.default] : [])
+          ),
+        ].map(example => (
+          <option key={example} value={example} />
+        ))}
       </datalist>
     ) : null,
   ];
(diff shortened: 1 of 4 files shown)


packages/core/src/components/widgets/BaseInput.js
```diff
@@ -61,19 +61,29 @@ function BaseInput(props) {
...
-  return (
+  return [
     <input
       className="form-control"
       readOnly={readonly}
       disabled={disabled}
       autoFocus={autofocus}
       value={value == null ? "" : value}
       {...inputProps}
+      list={schema.examples ? `examples_${inputProps.id}` : null}
```
epicfaace: Can you make `idPrefix` be prepended to this `list` attribute and the `id` attribute in `<datalist>`?

packages/core/src/components/widgets/BaseInput.js
```diff
@@ -61,19 +61,29 @@ function BaseInput(props) {
...
       onChange={_onChange}
       onBlur={onBlur && (event => onBlur(inputProps.id, event.target.value))}
       onFocus={onFocus && (event => onFocus(inputProps.id, event.target.value))}
-    />
-  );
+    />,
+    schema.examples ? (
+      <datalist id={`examples_${inputProps.id}`}>
+        {schema.examples
+          .concat(schema.default ? [schema.default] : [])
```
epicfaace: Can you add some tests that test:
- the rendering of a form with a schema with `examples`
- the rendering of a form with a schema with `examples` and a `default` defined
- the rendering of a form with a schema with no `examples` and a `default` defined

89erik: Done :)

packages/core/src/components/widgets/BaseInput.js
89erik: After looking briefly through the rest of the code, idPrefix seems to me, as far as I can tell, to only be used higher up, e.g. SchemaField, ObjectField, to generate an ID to pass on to lower level components. So for the BaseInput type, which seems to be the very lowest, this is property is not available, and would need to be passed on through a couple of layers, alongside the generated ID, for this to work. Are you sure you want this?

packages/core/src/components/widgets/BaseInput.js
```diff
@@ -76,11 +76,13 @@ function BaseInput(props) {
...
     schema.examples ? (
       <datalist id={`examples_${inputProps.id}`}>
-        {schema.examples
-          .concat(schema.default ? [schema.default] : [])
-          .map(example => (
-            <option key={example} value={example} />
-          ))}
+        {[
+          ...new Set(
+            schema.examples.concat(schema.default ? [schema.default] : [])
```
epicfaace: Why did you decide to include `schema.default` within the examples that show up?

89erik: I did this because of this sentence in the [jsonschema specification](https://json-schema.org/understanding-json-schema/reference/generic.html): 
> There is no need to duplicate the default value in the examples array, since default will be treated as another example.

I agree if you think this is a bit of an odd behavior, seeing as default does not receive this special treatment when there are no examples. I do think this is most in accordance with the jsonschema spec, but I don't feel strongly about either way.

packages/core/src/components/widgets/BaseInput.js
epicfaace: Oh, I see. Looking at this code again, adding `idPrefix` seems to be unnecessary because `inputProps.id` will already include it.

packages/core/src/components/widgets/BaseInput.js
epicfaace: Thanks for referring to that. I believe you linked to a summary. Looking at the spec, it seems to be really our choice as to whether we should implement it (https://datatracker.ietf.org/doc/draft-handrews-json-schema-validation/?include_text=1):

> Implementations MAY use the value(s) of "default", if present, as an
   additional example.  If "examples" is absent, "default" MAY still be
   used in this manner.

I think that what you've implemented, though, in which default is an additional example, but only shows up in "examples" if the "examples" array is defined, is a sensible way to do it.

89erik:
Thanks for approving my pull request, @epicfaace! Are there any more actions required by me for getting this merged and released? (This is my first pull request, so I don't know the process.)

epicfaace:
Nope, nothing else -- merging now. Thanks!

PR #1522, opened 2019-11-27
Refactor AltDateWidget to functional component

jimmycallin:
Thought I could help you with updating a component that still relied on componentWillReceiveProps :)

- componentWillReceiveProps is deprecated and will not work with future versions of React. I personally find the functional component easier to understand.
- some test cases relied on timestamps with non-zero milliseconds. previous implementation would store the milliseconds in the formData state, but when triggering the onChange and onSubmit function, the milliseconds would disappear. By adding optional support for milliseconds, it will now correctly call onChange and onSubmit function with the provided values.

### Checklist


jimmycallin:
@epicfaace it's a bit unfortunate how the tests are built, since their reliance on asserting on the internal state makes it impossible to use functional components. i suggest that a better solution would be to pass an onChange function in the tests and assert that it has been called with the correct formData. i could help with this if you think it's a good idea.

epicfaace:
@jimmycallin yes, definitely, testing internal state is bad. Your suggestion makes sense -- it would be great if you could do that!

If you do do that, could you refactor the tests in a separate PR so we can make sure they still pass on the old code?

jimmycallin:
@epicfaace this one's now (finally!) ready for review! phew.

jimmycallin:
@epicfaace Honestly I don't quite remember, it was a while ago I made this PR.

But if you look at this test case: https://github.com/rjsf-team/react-jsonschema-form/pull/1522/files#diff-6e1d509b716d825e64173a54baf0a7fdR877

You can see that currently it is possible to add a datetime with milliseconds as default, that returns milliseconds on submit. I THINK i thought it was a weird discrepancy between allowing support of milliseconds, but not being able to include them when using the "Now" button. So I chose to include support for them as well.

jimmycallin:
@epicfaace Is this still interesting? I understand if you want to keep it consistent to class components since that is what you are using elsewhere, so I don't mind closing this PR if you think so.

epicfaace:
@jimmycallin sorry for the delay. Yeah, I'm a bit hesitant to merge this PR because all the other widgets are class-based and I don't see a huge advantage in moving to a functional component. I think it would be great to, at least for now, try to remove componentWillReceiveProps from the class-based component, though.

PR #1525, opened 2019-12-01
Refactor some tests to assert on change events instead of internal state

jimmycallin:
Refactored a few test fields to assert on change events instead of internal state.

This is just the beginning :) If you like this I'll refactor the rest as well.

[Please describe them here]

### Checklist

### DIFF
--- a/packages/core/test/NumberField_test.js
+++ b/packages/core/test/NumberField_test.js
@@ -130,7 +130,9 @@ describe("NumberField", () => {
         });
 
         submitForm(node);
-        expect(onSubmit.lastCall.args[0].formData).eql(undefined);
+        sinon.assert.calledWithMatch(onSubmit.lastCall, {
+          formData: undefined,
+        });
       });
 
       it("should assign a default value", () => {
@@ -157,7 +159,9 @@ describe("NumberField", () => {
           target: { value: "2" },
         });
 
-        expect(onChange.lastCall.args[0].formData).eql(2);
+        sinon.assert.calledWithMatch(onChange.lastCall, {
+          formData: 2,
+        });
       });
 
       it("should handle a blur event", () => {
@@ -267,7 +271,9 @@ describe("NumberField", () => {
               target: { value: test.input },
             });
 
-            expect(onChange.lastCall.args[0].formData).eql(test.output);
+            sinon.assert.calledWithMatch(onChange.lastCall, {
+              formData: test.output,
+            });
             expect($input.value).eql(test.input);
           });
         });
@@ -287,7 +293,9 @@ describe("NumberField", () => {
           target: { value: ".00" },
         });
 
-        expect(onChange.lastCall.args[0].formData).eql(0);
+        sinon.assert.calledWithMatch(onChange.lastCall, {
+          formData: 0,
+        });
         expect($input.value).eql(".00");
       });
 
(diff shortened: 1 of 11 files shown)


packages/core/test/test_utils.js
```diff
@@ -43,3 +53,9 @@ export function describeRepeated(title, fn) {
     );
   }
 }
+
+export async function submitForm(node) {
```
epicfaace: Why does this function (and the function wrapped in `act`) need to be async?

packages/core/test/NumberField_test.js
```diff
@@ -117,13 +122,15 @@ describe("NumberField", () => {
         expect(node.querySelector(".field-description").textContent).eql("bar");
       });
 
-      it("should default state value to undefined", () => {
-        const { comp } = createFormComponent({
+      it("should default state value to undefined", async () => {
```
epicfaace: Can you change the test name here?

epicfaace (commented):
Thanks, looks good!
- When would one want to test `onChange` versus `onSubmit`? I don't see a clear distinction between when you decided to test each one in the tests you added.
- Could we replace something like `onChange.lastCall.args[0].formData` with the less verbose `onChange.calledOnceWithExactly({})`? This is a great way to refactor the tests, but it might be good to figure out what our style of assertions will look like so we can standardize across all tests.

packages/core/test/test_utils.js
jimmycallin: apparently it didn't :) still haven't quite figured out when and when not you need `act` and when to make it async.

jimmycallin:
`calledOnceWithExactly` became a bit too verbose imo since you needed to pass a bunch of for the test unrelated parameters. took a middle road with `calledWithMatch` to only assert that formData is set correctly.

jimmycallin:
> When would one want to test onChange versus onSubmit? I don't see a clear distinction between when you decided to test each one in the tests you added.

it seems like `onChange` is not called when you pass `formData`, but when you set default values  `onChange` is indeed called. the tests should (now) reflect that `submitForm` is only called when you pass formData on initiation and want to verify that it is set correctly.

epicfaace:
> `calledOnceWithExactly` became a bit too verbose imo since you needed to pass a bunch of for the test unrelated parameters. took a middle road with `calledWithMatch` to only assert that formData is set correctly.

Good idea. Maybe you could replace `sinon.assert.calledWithMatch` with `spy.calledWithMatch` to make it even less verbose?

jimmycallin:
> Good idea. Maybe you could replace sinon.assert.calledWithMatch with spy.calledWithMatch to make it even less verbose?

the output of `spy.calledWithMatch` is only `true` or `false`, which means you have to wrap it in expect(...).eql(...). in case of mismatch, the error message is very unhelpful (e.g. "true should be false"), while I found using sinon's built in assertion much more helpful in pinpointing the error. but if you insist on using `expect(spy.calledWithMatch(...)).eql(...)` i'll change it :)

epicfaace:
Ok, that makes sense -- since sinon's assertion error messages are better than using `expect` and `eql`, I'm all for it.

epicfaace:
> it seems like `onChange` is not called when you pass `formData`, but when you set default values `onChange` is indeed called.

Note, however, that we'll probably change this behavior in v2 so that `onChange` is not called when setting default values (#1201).

epicfaace (approved):
Looks good -- thanks! It'd be great if you are able to help refactor the rest as well.

epicfaace (approved):
Changes look good, thanks! Thanks for making several of the tests more robust as well.

PR #1528, opened 2019-12-05
Display changes in extraErrors when noValidate is true or liveValidate is false

olzraiti:

My validation logic is quite complex including logic for handling simultaneously error, warnings, live-errors and live-warnings, and async for all of these. Also when I'm submitting, I need to delay the validations until all async form changes have completed. Because of these conditions, it's much more simpler to handle the validation process entirely on my client code instead of relying on RJSF handling the validation for JSON schema. That's why I'm using the flags `noValidate` and I'm passing all the errors through `extraErrors`.

I noticed that the `getStateFromProps` has has a state issue with `errorSchema` when `noValidate` is used. I added a test case to cover this.    

### Checklist


olzraiti:
I updated the title to describe the bug better, and made the test case a bit more clear.

The test case documents the bug quite well IMO but here's a further explanation: 
1. Take a look at https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/Form.js#L68-L77
2.  Set `noValidate` to true and pass some errors in `extraErrors`
3. Pass empty `extraErrors` in order to clear the errors
4.  The errors aren't cleared, because in the highlighted code block we are merging the previous errors into the new `extraErrors`.  if `noValidate` is true, we should never merge state errors into the props errors, since the errors are handled completely externally.

olzraiti:
Related #1547

packages/core/src/components/Form.js
```diff
@@ -68,8 +68,8 @@ export default class Form extends Component {
     let { errors, errorSchema } = mustValidate
       ? this.validate(formData, schema, additionalMetaSchemas, customFormats)
       : {
-          errors: state.errors || [],
-          errorSchema: state.errorSchema || {},
+          errors: (!props.noValidate && state.errors) || [],
```
epicfaace: This fix only works for when `noValidate` is true, but it doesn't fix the issue when `liveValidate` is false (#1547). I've went ahead and added a test for that one. Do you have an idea of what might fix that one?

epicfaace: Could you clean up the code that's here a bit? I think over time & many commits it's become difficult to read. Something like

```
let errors = !props.noValidate ? state.errors : [];
```

olzraiti: The bug for `liveValidate` was very similar to the bug being fixed: `mustValidate` being false, state errors were used and they contained the old `extraErrors`. I updated the code to check for both `liveValidate` and `noValidate` before using state errors.

olzraiti: I moved the logic for getting state errors to a separate function.

olzraiti:
Hey, could we get this merged?

pedramtalebi:
Any update on this?

olzraiti:
This PR isn't waiting for any changes (I made changes already according to the review) and is ready to merge if there are no further issues.

epicfaace:
@olzraiti sorry for the delay.

Looks good, except I have one question. Let's say `liveValidate` is set to `false`, the user submits and gets a normal validation error (not from `extraErrors`). Then, `extraErrors` is changed to add a new validation error -- and the original validation error will disappear and only the error from `extraErrors` will be added.

Rather than this behavior, it might make more sense to keep the original validation errors when the `extraErrors` prop is changed.

olzraiti:
Thanks for the input! Good point, I added a new test and modified the logic to keep old schema validation errors when `extraErrors` are updated and `liveValidate` is false. I had to add some new variables to the state in order to store the schema validations.

epicfaace:
I think tests are failing because we call `this.props.onChange(nextState);` where nextState comes from `this.getStateFromProps()`. Thus, there are two extra keys in nextState,:

```
  schemaValidationErrorSchema: undefined,
  schemaValidationErrors: undefined,
```

olzraiti:
Right! fixed that.

packages/core/src/components/Form.js
```diff
@@ -131,6 +131,23 @@ export default class Form extends Component {
       schemaValidationErrors,
```
epicfaace, line 131: Could we simplify this by just removing these two lines (then we would have no need for `getStateForEmit`)?

```suggestion
```

packages/core/src/components/Form.js
```diff
@@ -131,6 +131,23 @@ export default class Form extends Component {
...
       schemaValidationErrorSchema,
```
epicfaace, line 132:
```suggestion
```

packages/core/src/components/Form.js
olzraiti: Yes, thanks!

packages/core/src/components/Form.js
```diff
@@ -58,21 +58,58 @@ export default class Form extends Component {
...
-      ? this.validate(formData, schema, additionalMetaSchemas, customFormats)
-      : {
-          errors: state.errors || [],
-          errorSchema: state.errorSchema || {},
+
+    const getCurrentErrors = () => {
+      if (props.noValidate) {
+        return { errors: [], errorSchema: {} };
+      } else if (!props.liveValidate) {
+        return {
```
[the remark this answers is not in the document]
saimanoharnethi: Should we set the state.schemaValidationErrors and schemaValidationErrorSchema in line 373 here  - 
https://github.com/rjsf-team/react-jsonschema-form/pull/1528/files#diff-a3830d0e95feb07e61c8e35cf02715c655766e69871111b67373658e5d98398bR373

PR #1535, opened 2019-12-08
More test refactoring to move away from internal component state assertion

jimmycallin:

- Refactoring a large number of tests to no longer assert on internal component state.
- Fix key validation issue in BaseInput: By adding the `key` prop you no longer get unique key prop warnings when running e.g. the `validation_test.js`.

### Checklist


packages/core/src/components/Form.js
```diff
@@ -266,7 +266,10 @@ export default class Form extends Component {
     }
 
     if (!this.props.noValidate) {
-      let { errors, errorSchema } = this.validate(newFormData);
+      let { errors, errorSchema } = this.validate(
+        newFormData,
+        this.state.schema
```
jimmycallin: when i refactored the test "should validate formData when the schema is updated" in Form_test.js I noticed that it never called the onError event as I expected, but rather only updated the internal state. I tracked it down to be due to this.validate were still using the previous schema as passed by props. solved this by passing state.schema explicitly.

packages/core/test/ArrayField_test.js
```diff
@@ -1181,15 +1230,15 @@ describe("ArrayField", () => {
         .not.to.be.null;
     });
 
-    it("should handle a change event", () => {
```
epicfaace: thanks for fixing this

packages/core/test/Form_test.js
```diff
@@ -1102,29 +1120,39 @@ describeRepeated("Form common", createFormComponent => {
       };
 
       it("should update form state from new formData prop value", () => {
```
epicfaace: Can you rename this test to not mention state?

packages/core/test/Form_test.js
```diff
@@ -1289,13 +1331,14 @@ describeRepeated("Form common", createFormComponent => {
...
-          const { comp, node } = createFormComponent({ schema });
+          const { node, onChange } = createFormComponent({ schema });
 
           Simulate.change(node.querySelector("input[type=text]"), {
             target: { value: "short" },
           });
-
-          expect(comp.state.errorSchema).eql({});
+          sinon.assert.calledWithMatch(onChange.lastCall, {
+            onError: undefined,
```
epicfaace: Should this be
```suggestion
            errorSchema: undefined,
```

packages/core/test/Form_test.js
```diff
@@ -1419,22 +1468,6 @@ describeRepeated("Form common", createFormComponent => {
         minLength: 8,
       };
 
-      it("should update the errorSchema on form submission", () => {
```
epicfaace: Why did you remove this test? Is there already another test that makes this one redundant?

packages/core/test/Form_test.js
```diff
@@ -1495,11 +1530,18 @@ describeRepeated("Form common", createFormComponent => {
       };
 
       it("should reflect the contextualized error in state", () => {
```
epicfaace: Can you rename this test?

packages/core/test/Form_test.js
```diff
@@ -1524,13 +1566,26 @@ describeRepeated("Form common", createFormComponent => {
...
```
epicfaace: Can you rename this test?

packages/core/test/Form_test.js
```diff
@@ -1573,15 +1628,19 @@ describeRepeated("Form common", createFormComponent => {
...
```
epicfaace: Can you rename this test?

packages/core/test/Form_test.js
```diff
@@ -2195,22 +2301,30 @@ describeRepeated("Form common", createFormComponent => {
...
-      const { comp } = createFormComponent(formProps);
+      const { comp, node, onError } = createFormComponent(formProps);
 
-      expect(comp.state.errorSchema).eql({});
+      submitForm(node);
+      sinon.assert.notCalled(onError);
 
       setProps(comp, {
         ...formProps,
+        onError,
```
epicfaace: Do we need this line? (also line 2356)

packages/core/test/uiSchema_test.js
```diff
@@ -1703,15 +1743,19 @@ describe("uiSchema", () => {
       });
 
       it("should map widget value to a typed state one", () => {
```
epicfaace: Can you rename this?

packages/core/test/uiSchema_test.js
```diff
@@ -1654,11 +1690,13 @@ describe("uiSchema", () => {
           },
         });
 
-        expect(comp.state.formData).eql({ foo: true });
+        sinon.assert.calledWithMatch(onChange.lastCall, {
+          formData: { foo: true },
+        });
       });
 
       it("should update state when false is selected", () => {
```
epicfaace: Can you rename this?

packages/core/test/uiSchema_test.js
```diff
@@ -1286,7 +1302,7 @@ describe("uiSchema", () => {
       });
 
       it("should update state when value is updated", () => {
```
epicfaace: Can you rename this? (and all the ones mentioning state below)

packages/core/test/validate_test.js
```diff
@@ -684,19 +643,15 @@ describe("Validation", () => {
...
-          liveValidate: true,
-        });
-        comp.UNSAFE_componentWillReceiveProps({ formData });
-
-        expect(comp.state.errorSchema).eql({
-          0: { __errors: [] },
-          1: { __errors: [] },
-          2: { __errors: [] },
-          __errors: ["Forbidden value: bbb"],
+          formData,
```
epicfaace: I see that you removed liveValidate for several of these tests. Can you add one test in which you use both liveValidate and a custom validate function, to make sure that functionality is still covered by our tests?

packages/core/test/Form_test.js
jimmycallin: yes, just below this one

packages/core/test/Form_test.js
jimmycallin: If not, the setProps function will not include an onError spy function that we can assert on. There might be a better way to fix this, but that felt out of scope for this PR.

jimmycallin:
@epicfaace think i've resolved all your comments now :)

jimmycallin:
@epicfaace i think there's just one thing left to make the tests "strict mode ready", and that's the performance_test.js. i've tried too look into how to assert number of rerenders that covers both functional and class components, and i haven't found a best practice way to do this. i suggest that we simply remove this test file completely.

epicfaace:
@jimmycallin good point, it seems like performance_test.js is testing too specific (something that should already be done by React) so we don't need those tests. Do you think there might be value in unit-testing `shouldComponentUpdate` directly, though?

Also, what did you mean by making the tests "[strict mode](https://reactjs.org/docs/strict-mode.html) ready"?

jimmycallin:
I removed the performance tests!

> @jimmycallin good point, it seems like performance_test.js is testing too specific (something that should already be done by React) so we don't need those tests. Do you think there might be value in unit-testing shouldComponentUpdate directly, though?

I think the unit tests here: https://github.com/rjsf-team/react-jsonschema-form/blob/4524376debbb6f7ce99aa0a1dc97b9c1e6e47e61/packages/core/test/utils_test.js#L2336

already seem to cover the necessary unit tests for shouldComponentUpdate. do you agree?

> Also, what did you mean by making the tests "strict mode ready"?

I mostly meant that the tests are not dependent on deprecated class methods like componentWillReceiveProps that aren't valid when wrapping the app in `<React.StrictMode>`. :) In theory we should be able to wrap the `test_utils.createComponent` output in `<React.StrictMode>`, but haven't tried it myself.

epicfaace:
> already seem to cover the necessary unit tests for shouldComponentUpdate. do you agree?

Makes sense!

packages/core/src/components/Form.js
epicfaace: It seems that when I remove this change, the updated "should validate formData when the schema is updated" test is still passing, though. Do you know why this might be the case?

jimmycallin: really weird, i cannot reproduce the error anymore. must have been something else that messed up the test while i was doing the refactoring. i reverted the change, thanks for checking!

packages/core/src/components/widgets/BaseInput.js
```diff
@@ -63,6 +63,7 @@ function BaseInput(props) {
 
   return [
     <input
+      key={inputProps.id}
```
epicfaace: What issue did this fix? Could you add that to the PR description?

jimmycallin: Before adding the key you got key validation errors in e.g. validation_test.js, since it's wrapped in an array:

```
      Custom Form validation
Warning: Each child in a list should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
    in BaseInput (created by TextWidget)
    in TextWidget
    in Unknown (created by StringField)
    in StringField (created by SchemaField)
    in div (created by WrapIfAdditional)
    in WrapIfAdditional (created by DefaultTemplate)
    in DefaultTemplate (created by SchemaField)
    in SchemaField (created by Form)
    in form (created by Form)
    in Form
```

I updated the PR.

PR #1539, opened 2019-12-10
Combine all themes into a single playground

epicfaace:
Fixes #1536.

### Description
- Make a separate package called `@rjsf/playground`. This exports a `Playground` component that takes an object of themes in its props.
- This package will be published to npm. It will have a single "playground" within it which will have all themes, such as the following:
```
import React from "react";
import { Theme as MuiTheme } from "rjsf-material-ui";
import Playground from "../src/index";
import { render } from "react-dom";

const themes = {
  default: {
    stylesheet:
      "//maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css",
    theme: {},
  },
  "material-ui": {
    stylesheet: "",
    theme: MuiTheme,
  },
};

render(<Playground themes={themes} />, document.getElementById("app"));
```
- Each package can have a "development" playground with the following code:
```
import React from "react";
import { render } from "react-dom";
import Playground from "@rjsf-team/playground";

const themes = [
  {
    default: {
      stylesheet:
        "//maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css",
      theme: {},
    },
  },
];

render(<Playground themes={themes} />, document.getElementById("app"));
```

PR #1561, opened 2019-12-29
Enable antd (ant-design) support

delyanr:

Implement antd (ant-design) theme, fields and widgets as separate sub-package. Usage is as simple as importing the `Form` component from the package.

Pending to-dos:
* [x] Add `README.md` file.
* [x] Enable playground. Deps: #1539 
* [x] Remove `withTheme` HOC. Deps: #1498 
* [ ] Publish package to npm repository

Regarding the last to-do, could you let me know what needs to be done to publish the package (I've never used lerna before)?

### Checklist


delyanr:
Hi. I've added a `README.md` file. I believe this can already be merged without the playground and especially without the `withTheme` dependencies, so people can start testing it. Once these dependent pull requests are merged, I can update my code as well. @epicfaace, let me know what you think?

packages/antd/package.json
```diff
@@ -0,0 +1,79 @@
...
+    "eslint-plugin-flowtype": "^2.39.1",
+    "eslint-plugin-import": "^2.7.0",
+    "eslint-plugin-jsx-a11y": "^5.1.1",
+    "eslint-plugin-react": "^7.4.0",
+    "jest": "24.9.0",
+    "less": "^3.10.3",
+    "less-loader": "^5.0.0",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "react-jsonschema-form": "^2.0.0-alpha.1",
```
epicfaace: I think you can move some of these (`react`, `react-dom`, `react-jsonschema-form`) to `devDependencies`.

packages/antd/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "@react-jsonschema-form/antd",
```
epicfaace:
```suggestion
  "name": "@rjsf/antd",
```

This is the name under which the package will be published. Can you rename to the above? Once this PR is merged, I can publish it under the `rjsf` npm organization.

packages/antd/README.md
```diff
@@ -0,0 +1,83 @@
...
+- [Usage](#usage)
+- [Roadmap](#roadmap)
+- [Contributing](#contributing)
+- [License](#license)
+
+## Getting Started
+
+### Prerequisites
+
+- `antd == 3` (not tested yet with version 4)
```
epicfaace: Do you envision having to make a separate theme to support antd 4, or can one theme support both?

packages/antd/README.md
```diff
@@ -0,0 +1,83 @@
...
+- `react-jsonschema-form >= 1.6.0`
+
+```sh
+npm install antd @react-jsonschema-form/core
```
epicfaace:
```suggestion
npm install antd react-jsonschema-form
```

packages/antd/package.json
epicfaace: Also please replace other occurrences in the readme.

delyanr: Just to double-check, will the core package remain `react-jsonschema-form`, or will that also migrate to the `rjsf` npm organization?

packages/antd/README.md
delyanr: Unfortunately, antd 4 introduced breaking changes related to the Form component. I haven't looked into the details yet, but I think we need to fix a period of time where we support two themes, and then slowly deprecate antd 3.

packages/antd/README.md
delyanr: Ditto

packages/antd/package.json
epicfaace: I'm leaning more towards keeping it called `react-jsonschema-form`. It would be easier to upgrade that way; and that way, if someone installs the `@latest` version of it, they won't get the v2.0.0-alpha.1 alpha version.

jeremypeters:
Hi. I'm using Ant Design and would like to use react-jsonschema-form. I need to settle on an approach very shortly, and I was just wondering if there is a rough ETA for this merge?

delyanr:
@jeremypeters Just saw that the dependencies from OP are now merged. So I will have a look and finalise this pull request in the next week or so. I'm very busy at the moment, unfortunately, but will get to it.

jeremypeters:
Thanks @delyanr . No worries at all - totally understand. I'll just use the Ant Design CSS for the time-being :thumbsup:

packages/antd/package.json
epicfaace: Hey, just to let you know, I've updated this -- changed the name back to `@rjsf/core` for the core package -- it seemed best to keep all rjsf packages under the `@rjsf` organization to maintain a clear separation between org-owned packages and other community packages. We can still keep this called `@rjsf/antd` though.

delyanr:
@epicfaace Almost at the finish line now with this pull request. Seems like the checks are failing because the `@rjsf/antd` must be published first; otherwise the playground cannot find the dependency. Is that something you can do?

Finally, I spent a decent amount of time getting the antd styles to work in the playground but failed miserably. Any help to get that one, so we can finally merge would be greatly appreciated. Any ideas why that might not be working correctly? Thanks!

erikhofer:
Hey! Any update on this? :)

wenerme:
Base on previous work, I created a workable antd 4.0 with 2.0.0 rjsf, storybook here https://apis.wener.me/storybook/?path=/story/rjsf-demo--playground ,not pub to npm yet, consider to make a pr, but for now, anyone want to use antd, just copy and paste https://github.com/wenerme/apis/tree/master/packages/ui/stories/rjsf/antd. then, everything should just works!

delyanr:
@erikhofer I'm going through an internal update to v4 this week, so will have a look at this during the weekend to try and finalise it (and maybe split into v3 and v4 versions). In the meantime, what @wenerme pointed to should work. Thanks

wenerme:
Current antd rjsf is followed the bootstrap style, but should not follow every style, antd's component can be more expressive,like the title and description, reorder should also support dnd, can support more input widgets.

epicfaace:
@delyanr thanks for your help! Now that we have the infrastructure for tests / etc. for material-ui ready, we should be able to merge this in soon.

Could you add some simple snapshot tests such as the ones in material-ui (https://github.com/rjsf-team/react-jsonschema-form/tree/master/packages/material-ui/test)?

delyanr:
@epicfaace Hi! I finally got some time to finalise this PR. Note that I decided to migrate entirely to antd v4, since it's been a while now. I also added some snapshot tests as per your previous comment.

Importantly, the current playground does not allow for styles to be injected into the iframe tag, other than for the material-ui theme. I didn't spend much time on a solution there, so you'll notice the stylesheet is coming from a CDN link. We should probably improve on this, since the small number of styles defined in rjsf-antd are not working in the playground (the index.less file).

Let me know if there is anything else.

epicfaace:
If I select a datetime and then click outside of the box instead of clicking "OK", the input box stays empty. Is this expected behavior?

image

epicfaace:
Should we make the titles for objects larger as in bootstrap / other themes?:

bootstrap:

image

antd:

image

epicfaace:
When I go to the "datetime" example, all the dropdowns under "alt-datetime" and "alt-date" show a "-1":

image

epicfaace:
Could we add support for additionalProperties?

boostrap:
image

antd:
image

epicfaace:
Could we also fix the default submit button (same issue as with https://github.com/rjsf-team/react-jsonschema-form/issues/1725 for material-ui)?

packages/antd/package.json
```diff
@@ -0,0 +1,97 @@
+{
+  "name": "@rjsf/antd",
+  "version": "2.0.0",
+  "description": "Ant Design theme, fields and widgets for react-jsonschema-form",
+  "main": "dist/index.js",
+  "module": "dist/rjsf-antd.js",
+  "scripts": {
+    "build": "npm run build:dist && npm run build:lib",
+    "build:dist": "rimraf dist && cross-env NODE_ENV=production webpack --config webpack.config.dist.js",
+    "build:lib": "rimraf lib && cross-env NODE_ENV=production babel -d lib/ src/",
```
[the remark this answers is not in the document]
delyanr: You are right. Will remove and commit.

delyanr:
> If I select a datetime and then click outside of the box instead of clicking "OK", the input box stays empty. Is this expected behavior?
> 
> image

Yes. See: https://ant.design/components/date-picker/

delyanr:
@epicfaace See my replies above.

delyanr:
- Removed the useless build:lib script
- Added support for additionalProperties

We are waiting for ant design PRs to fix the floating containers (hopefully).

delyanr:
@epicfaace In the interest of getting this merged ASAP, I implemented the alternative date pickers, even though I'm still not sure why we need them. I've also fixed the popups for the native pickers not aligning with the fields. I believe we should be merging this now, thanks.

packages/antd/src/widgets/SelectWidget/index.js
```diff
@@ -0,0 +1,112 @@
...
+      onFocus={!readonly ? handleFocus : undefined}
+      placeholder={placeholder}
+      style={SELECT_STYLE}
+      value={typeof value !== 'undefined' ? stringify(value) : emptyValue}
+    >
+      {enumOptions.map(({ value: optionValue, label: optionLabel }) => (
+        <Select.Option
+          disabled={enumDisabled && enumDisabled.indexOf(value) !== -1}
+          key={String(optionValue)}
+          value={String(optionValue)}
```
cdussud, line 99: shouldn't this just be `value={optionValue}`? OptionValue can be a number

packages/antd/src/widgets/SelectWidget/index.js
```diff
@@ -0,0 +1,112 @@
...
+    Array.isArray(currentValue) ? value.map(String) : String(value);
+
+  return (
+    <Select
+      autoFocus={autofocus}
+      disabled={disabled || (readonlyAsDisabled && readonly)}
+      getPopupContainer={getPopupContainer}
+      id={id}
+      mode={typeof multiple !== 'undefined' ? 'multiple' : undefined}
+      name={id}
```
cdussud, line 87: I don't think name is one of the Antd Select props -- any reason this is here?

packages/antd/src/widgets/SelectWidget/index.js
delyanr: I did this a while ago now, but I remember I had some issues with boolean values. Playing around with the latest Antd version, I see that they work well, but are not part of the typescript props, so prefer to still pass through strings to be honest.

packages/antd/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "ant-design",
+    "react",
+    "react-jsonschema-form",
+    "jsonschema",
+    "json-schema",
+    "json",
+    "rjsf-antd"
+  ],
+  "author": "Delyan Ruskov <d.ruskov@gmail.com>",
+  "license": "MIT",
```
epicfaace, line 91: Are you fine changing this to an Apache 2 license so the entire repo is under a single license?

delyanr: Done.

delyanr:
> This looks really good -- thanks for your work on this. Can you address the license issue -- and then we can merge?
> 
> Some things which we should fix (but need not block this merge):
> 
> * When I update the value of altdatewidget / alttimewidget, it doesn't call onChange properly -- the formData isn't updating on the playground.
> * When I change the name of the key in additionalProperties and focus out, the key name changes, but then the textbox for the key becomes blank (when it should actually still have the same value as before).
> * It would be nice to (later) convert this source to typescript and use tsdx instead of webpack for building, so we have some kind of unified build process for all themes (the hope is that we can do this with @rjsf/core eventually, too).
> * If someone has already imported the antd css in their project, is there a way to still use @rjsf/antd without causing duplicate imports (as it stands, I believe we will end up having this css duplicated)?

I've fixed the first two points. Not sure I have the time to address the other two at this time, sorry. Once this is merged, others are free to contribute, etc. Thanks

packages/antd/src/widgets/SelectWidget/index.js
```diff
@@ -64,8 +64,6 @@ const SelectWidget = ({
 
   const { enumOptions, enumDisabled } = options;
 
-  const emptyValue = multiple ? [] : '';
```
epicfaace: Why remove this? I think this is required for multiple select support?

epicfaace:
For some reason the deploy preview (https://deploy-preview-1561--rjsf.netlify.app/) is not working. Looking into it...

packages/antd/src/widgets/SelectWidget/index.js
delyanr: I thought so too, but seems like Ant Design component handles it just fine. Having the `emptyValue` however breaks the placeholder, so that's a bug there, and hence why I removed it. You can try it in the playground.

epicfaace:
@delyanr is there an alternative way that lets us not include the `index.less` file? This is because `index.less` already includes the default antd theme -- how would a user use a custom theme in this case? Is there an easy way to do this when the less file is already included?

epicfaace:
Looks like we may need to do this for the above problem: https://ant.design/docs/react/customize-theme#How-to-avoid-modifying-global-styles

Issue #1494, opened 2019-10-22
Allow object selection in dropdowns


tdamsma:
I want to present a dropdown with options that when selected populate the formData with an object. 

Consider this schema:

```json
{
  "type": "object",
  "title": "Location",
  "required": [
    "name",
    "latitude",
    "longitude"
  ],
  "oneOf": [
    {
      "const": {
        "name": "Home",
        "latitude": 12.4,
        "longitude": 2.8
      },
      "title": "Home"
    },
    {
      "const": {
        "name": "Work",
        "latitude": 62.4,
        "longitude": 4
      },
      "title": "Work"
    }
  ]
}
```

I would like a dropdown that gives the options `Work` and `Home` but that doesn't store a string but the entire object with name, lat, lon. The schema above does not render properly at all. 

The closest I could get to the intended behavior is the following:
```json
{
  "type": "object",
  "title": "Location",
  "required": [
    "name",
    "latitude",
    "longitude"
  ],
  "oneOf": [
    {
      "properties": {
        "name": {
          "const": "Home",
          "default": "Home",
          "readOnly": true
        },
        "latitude": {
          "const": 12.4,
          "default": 12.4,
          "readOnly": true
        },
        "longitude": {
          "const": 2.8,
          "default": 2.8,
          "readOnly": true
        }
      },
      "title": "Home"
    },
    {
      "properties": {
        "name": {
          "const": "Work",
          "default": "Work",
          "readOnly": true
        },
        "latitude": {
          "const": 62.4,
          "default": 62.4,
          "readOnly": true
        },
        "longitude": {
          "const": 4,
          "default": 4,
          "readOnly": true
        }
      },
      "title": "Work"
    }
  ]
}
```
This renders all the fields disabled. What I would want is to nut render the name/lat/lon fields at all (only the title dropdown), but do pass them to the formData.

KKS1:
Hi @tdamsma , how about if you use `"ui:widget": "hidden"` in uiSchema instead of `readOnly: true` in schema.

tdamsma:
@KKS1 that would work as a work around to hide the fields, thanks! Still makes using this option slightly cumbersome, as the jsonschema's I use are dynamically generated so this would require generating an accompanying uiSchema. Would be nice if the uiSchema could be merged with the jsonschema, but that is a another discussion

epicfaace:
Yeah, merging this into #701 then.

PR #1562, opened 2019-12-30
Use index for value in SelectWidget

tdamsma:
This PR changes the way options are selected in the SelectWidget by using the options array  index instead of a serializing the selected value to the DOM (cast to string). This allows removal of a lot of hacky code to guess type of deserialized values, and also allows for selecting complete objects via a dropdown, addressing #1494.

I've added a playground example. I haven't looked at fixing the tests yet as  would like feedback on the approach first.

### DIFF
--- a/packages/core/src/components/widgets/SelectWidget.js
+++ b/packages/core/src/components/widgets/SelectWidget.js
@@ -1,7 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import _findIndex from "lodash/findIndex";
-import _indexOf from "lodash/indexOf";
+import _isEqual from "lodash/isEqual";
 
 function getValue(event, enumOptions, multiple) {
   if (event.target.value === "") {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/widgets/SelectWidget.js
```diff
@@ -1,49 +1,47 @@
...
+  if (typeof value === "undefined") {
+    return "";
+  }
+
+  for (const i in enumOptions) {
+    if (_isEqual(enumOptions[i].value, value)) {
+      return i;
+    }
+  }
+  return "";
```
epicfaace: Why return an empty string by default?

Could we just return `undefined` instead -- so that it would also be possible to select `""`? (an additional improvement from the old code)

packages/core/src/components/widgets/SelectWidget.js
```diff
@@ -63,41 +61,40 @@ function SelectWidget(props) {
...
-        onChange(processValue(schema, newValue));
+        onChange(getValue(event, enumOptions, multiple));
       }}>
       {!multiple && schema.default === undefined && (
         <option value="">{placeholder}</option>
       )}
       {enumOptions.map(({ value, label }, i) => {
         const disabled = enumDisabled && enumDisabled.indexOf(value) != -1;
         return (
-          <option key={i} value={value} disabled={disabled}>
```
epicfaace: This is likely a breaking change because it changes what's rendered in the DOM (the `value`)... but it's fine to go into version 2, and worth it to allow for better selection of any kinds of enums.

packages/core/src/components/widgets/SelectWidget.js
tdamsma: sounds like a good idea, will look into it

tdamsma:
> This would be a great improvement!

Good to hear, what can I do help this get landed? Not really sure how to proceed. Lots of tests are failing, but as they are written now they would become kind of pointless as most tests I saw focus on checking the values in the DOM render which are just a bunch of integers with this change.

epicfaace:
It seems that many of the tests are failing because `o` is undefined in `.map(o => o.value)`: https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/widgets/SelectWidget.js#L43

heath-freenome:
Implemented this fully for all widgets that use `enumOptions` (not just `SelectWidget`) in #3411

PR #1564, opened 2020-01-03
Fix oneOf/anyOf nested defaults

delyanr:

Somewhat related to (but does not fully address): #1558 

There is a cheeky bug whereby nested defaults in oneOf/anyOf remain in the `formData` when the option is changed. See [playground example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJxMSI6eyJsb3JlbSI6Im9wdGlvbiAxIn19LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsIm9uZU9mIjpbeyJwcm9wZXJ0aWVzIjp7InExIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7ImxvcmVtIjp7InR5cGUiOiJzdHJpbmcifX0sImRlZmF1bHQiOnsibG9yZW0iOiJvcHRpb24gMSJ9fX0sInJlcXVpcmVkIjpbInExIl19LHsicHJvcGVydGllcyI6eyJxMiI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJsb3JlbSI6eyJ0eXBlIjoic3RyaW5nIn19LCJkZWZhdWx0Ijp7ImxvcmVtIjoib3B0aW9uIDIifX19LCJyZXF1aXJlZCI6WyJxMiJdfV19LCJ1aVNjaGVtYSI6e30sImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6dHJ1ZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==). Select "Option 2" and watch how there is stale data in `formData`.

However, note that there are still some issues when options to omit data are selected. This needs to be investigated further as it goes deeper into the logic of several utility functions.

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -199,10 +199,10 @@ function computeDefaults(
     );
   } else if ("oneOf" in schema) {
     schema =
-      schema.oneOf[getMatchingOption(undefined, schema.oneOf, definitions)];
+      schema.oneOf[getMatchingOption(formData, schema.oneOf, definitions)];
   } else if ("anyOf" in schema) {
     schema =
-      schema.anyOf[getMatchingOption(undefined, schema.anyOf, definitions)];
+      schema.anyOf[getMatchingOption(formData, schema.anyOf, definitions)];
   }
 
   // Not defaults defined for this node, fallback to generic typed ones.
--- a/packages/core/test/utils_test.js
+++ b/packages/core/test/utils_test.js
@@ -1150,6 +1150,55 @@ describe("utils", () => {
         };
         expect(getDefaultFormState(schema, formData)).to.eql(result);
       });
+
+      it("shouldn't add in default objects to formData", () => {
+        const schema = {
+          type: "object",
+          oneOf: [
+            {
+              properties: {
+                first: {
+                  type: "object",
+                  properties: {
+                    shared: {
+                      type: "string",
+                    },
+                  },
+                  default: {
+                    shared: "first",
+                  },
+                },
+              },
+            },
+            {
+              properties: {
+                second: {
+                  type: "object",
+                  properties: {
+                    shared: {
+                      type: "string",
+                    },
+                  },
+                  default: {
+                    shared: "first",
+                  },
+                },
+              },
+            },
+          ],
+        };
+        const formData = {
+          second: {
+            shared: "second",
+          },
+        };
+        const result = {
+          second: {
+            shared: "second",
+          },
+        };
+        expect(getDefaultFormState(schema, formData)).to.eql(result);
+      });
     });
   });
 


delyanr:
One more thing, to fully address #1558, we need to fix the issue where shared fields defaults are not resetting across options. See [playground example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsb3JlbSI6Im9wdGlvbiAxIn0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwiYW55T2YiOlt7InByb3BlcnRpZXMiOnsibG9yZW0iOnsidHlwZSI6InN0cmluZyJ9fSwiZGVmYXVsdCI6eyJsb3JlbSI6Im9wdGlvbiAxIn19LHsicHJvcGVydGllcyI6eyJsb3JlbSI6eyJ0eXBlIjoic3RyaW5nIn19fV19LCJ1aVNjaGVtYSI6e30sImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6dHJ1ZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==). Select "Option 2" and watch how the default remains in `formData` even though there is no default value in this option.

I suggest we add `includeUndefinedValues = true` to line 92 of `MultiSchemaField.js` which would ensure `computeDefaults` does not inject the wrong defaults. However, this fix would not work for schemas with dependencies since they don't use that field. Let me know what you think?

epicfaace:
@delyanr can you update this branch with the latest from master?

delyanr:
@epicfaace Seems like this fix is no longer working on the latest master. I suggest we actually spend some time fleshing out the `Form` component, since some of the problems come from the fact that we have derived state (see `getStateFromProps`). This means that when the oneOf/anyOf components tries to clear the formData from any defaults, in some cases they will simply be repopulated as the derived state calls `getDefaultFormState`. I hope this makes sense, as it is a bit difficult to explain. Also, we can use the time to convert `Form` from a class component to a functional component.

MingruiZhang:
@delyanr @epicfaace Any updates on this issue? We would love this issue fixed as well!

isphrazy:
any update here?

kjoneandrei:
any update on this being merged in?

Technologeek:
Any updates on this issue at all?

epicfaace:
Sorry, it seems that this fix isn't working on the latest master, as per @delyanr's comment above . I don't have the bandwidth to do the required changes, but if someone is able to fix it in a separate PR, I'd gladly merge it.

heath-freenome:
Closing this PR as #2676 seems to have done the same job

PR #1565, opened 2020-01-07
Allow defaults to be skipped on load

fsteger:

User may choose to clear a default value. If the user saves and returns to finish the form at a later date, it should be possible to load the form data without the defaults populated. In addition, if the user then makes a change to the form that causes additional fields to appear, such as adding a new array row or selecting a different field trigger, the newly shown fields should have their defaults populated.

Related issue - #1271 

### Checklist


fsteger:
I discovered a case where my initial implementation was not working as expected when using anyOf. When the initial selection has defaults that are skipped, selecting a different value and then re-selecting the initial value would not bring in the defaults as expected. This seems to have been an underlying issue with handling of value changes inside the MultiSchemaField. I added more tests for the anyOf as well. Reverting the changes in "MultiSchemaField" should give a couple failing tests for comparison with previous behavior. Incidentally, this may resolve #1558.

fsteger:
It looks like the Netlify checks are just hanging; is there any way to manually kick them off again?

fsteger:
I'm unable to reproduce the failures from the CI. Running on node v11.0.0, the cs-check locally flags a single file:
test/ArrayField_test.js

This is different than the two files flagged by CI using node 11.x:
test/Form_test.js
test/utils_test.js

Please advise as I'd like to resolve this issue quickly so this PR can proceed.

fsteger:
Hurray; everything's green again, and conflicts have been resolved.

Please let me know if there's anything else I can do to help move this along.

fsteger:
I've synced this branch again. Like before, please let me know if there's anything else I can do to help move this along.

packages/core/test/test_utils.js
```diff
@@ -37,13 +37,15 @@ export function setProps(comp, newProps) {
...
-/* Run a group of tests with different combinations of omitExtraData and liveOmit as form props.
+/* Run a group of tests with different combinations of omitExtraData and liveOmit,
+ *  and omitDefaultLoad, as form props.
  */
 export function describeRepeated(title, fn) {
   const formExtraPropsList = [
     { omitExtraData: false },
     { omitExtraData: true },
     { omitExtraData: true, liveOmit: true },
+    { omitDefaultLoad: false },
```
epicfaace, line 48: Did you mean to do `omitDefaultLoad: true` here? (since it's set to false by default)

fsteger: No; some of the pre-existing tests using this rely on the default behavior. I included it more to be explicit about testing the default behavior, but I can yank it out as it is a bit superfluous.

packages/core/test/Form_test.js
```diff
@@ -3144,3 +3145,431 @@ describe("Form omitExtraData and liveOmit", () => {
...
+
+  beforeEach(() => {
+    sandbox = createSandbox();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("should not populate empty default for root property on load", () => {
```
epicfaace, line 3160:
```suggestion
  it("should not populate default for root property on load", () => {
```

packages/core/test/Form_test.js
```diff
@@ -3144,3 +3145,431 @@ describe("Form omitExtraData and liveOmit", () => {
...
+    });
+
+    sinon.assert.calledWithMatch(onChangeProp, {
+      formData: {
+        foo: undefined,
+      },
+    });
+  });
+
+  it("should not populate empty default for array property on load", () => {
```
epicfaace, line 3188:
```suggestion
  it("should not populate default for array item property on load", () => {
```

packages/core/test/Form_test.js
```diff
@@ -3144,3 +3145,431 @@ describe("Form omitExtraData and liveOmit", () => {
...
+    });
+
+    sinon.assert.calledWithMatch(onChangeProp.lastCall, {
+      formData: {
+        foo: "bar",
+      },
+    });
+  });
+
+  it("should remain emtpty when changing anyOf selection to option with no default", () => {
```
epicfaace, line 3481:
```suggestion
  it("should remain empty when changing anyOf selection to option with no default", () => {
```

PR #1588, opened 2020-02-11
Fix anyof and oneof inside additional properties.

sempostma:

oneOf and anyOf properties are not working properly inside additional properties. 

### Checklist

### DIFF
--- a/packages/core/src/components/fields/ObjectField.js
+++ b/packages/core/src/components/fields/ObjectField.js
@@ -83,7 +83,7 @@ class ObjectField extends Component {
 
   onPropertyChange = (name, addedByAdditionalProperties = false) => {
     return (value, errorSchema) => {
-      if (!value && addedByAdditionalProperties) {
+      if (typeof value === "undefined" && addedByAdditionalProperties) {
         // Don't set value = undefined for fields added by
         // additionalProperties. Doing so removes them from the
         // formData, which causes them to completely disappear
--- a/packages/core/src/components/fields/ObjectField.js
+++ b/packages/core/src/components/fields/ObjectField.js
@@ -83,7 +83,7 @@ class ObjectField extends Component {
 
   onPropertyChange = (name, addedByAdditionalProperties = false) => {
     return (value, errorSchema) => {
-      if (typeof value === "undefined" && addedByAdditionalProperties) {
+      if (value === undefined && addedByAdditionalProperties) {
         // Don't set value = undefined for fields added by
         // additionalProperties. Doing so removes them from the
         // formData, which causes them to completely disappear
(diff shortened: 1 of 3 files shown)


sempostma:
@epicfaace  I have added the tests in https://github.com/rjsf-team/react-jsonschema-form/pull/1588/commits/9be4b1a4d15ba9afa980045a46c5cc6e10b217a1.

sempostma:
@epicfaace Is this still awaiting review?

packages/core/src/components/fields/ObjectField.js
```diff
@@ -83,7 +83,7 @@ class ObjectField extends Component {
 
   onPropertyChange = (name, addedByAdditionalProperties = false) => {
     return (value, errorSchema) => {
-      if (!value && addedByAdditionalProperties) {
+      if (typeof value === "undefined" && addedByAdditionalProperties) {
```
epicfaace, line 86: This change seems unrelated to the oneOf change -- could you add another test to cover it? (I believe this fixes an issue in which you can't type in "0" in a oneOf textbox, I think)

epicfaace, line 86: Let's do this, because the variable `value` is always going to be defined (whereas `value`'s value may be equal to undefined)

```suggestion
      if (value === undefined && addedByAdditionalProperties) {
```

packages/core/src/utils.js
```diff
@@ -201,11 +201,21 @@ function computeDefaults(
       )
     );
   } else if ("oneOf" in schema) {
-    schema =
-      schema.oneOf[getMatchingOption(undefined, schema.oneOf, rootSchema)];
+    const matchingKey = getMatchingOption(undefined, schema.oneOf, rootSchema);
+    if (schema.oneOf.hasOwnProperty(matchingKey) === false) {
+      // oneOf is an empty array so there is no schema available.
```
epicfaace, line 206: Can you add tests that cover this case as well (and update the PR description)?

packages/core/src/utils.js
```diff
@@ -201,11 +201,21 @@ function computeDefaults(
       )
     );
   } else if ("oneOf" in schema) {
-    schema =
-      schema.oneOf[getMatchingOption(undefined, schema.oneOf, rootSchema)];
+    const matchingKey = getMatchingOption(undefined, schema.oneOf, rootSchema);
+    if (schema.oneOf.hasOwnProperty(matchingKey) === false) {
```
epicfaace, line 205: Would it be clearer to instead check if `schema.oneOf.length === 0`, so we don't have to unnecessarily call `getMatchingOption`?

heath-freenome:
@memalloc Given the changes in v5 beta, the code here needs refactoring. Is this something you are willing to do?

heath-freenome:
Closing this PR as it is reimplemented in #3435

PR #1594, opened 2020-02-18
Overridable DescriptionField in BooleanField

89erik:

When supplying the fields-object in Form's props, the normal behavior is to override the fields globally. For example, 
`<Form fields={{DescriptionField: ()=><div>overriden</div>}}/>`
will cause all descriptions to show "overriden". Except for BooleanField, which always uses the built-in implementation of DescriptionField. This commit causes the BooleanField to use the overridden Description, if supplied.

### Checklist


89erik:
> Thanks! Can you add a test which uses the overriden DescriptionField?

Done :)

Issue #1409, opened 2019-08-14
v2 branch and roadmap


epicfaace:
We should get out a v2 soon so we can resolve some of the issues that require backwards-compatible changes. Let's use this issue for discussion of a roadmap, plans, and issues.

Tracking some of the things we might want in v2 [in this GitHub Project board](https://github.com/mozilla-services/react-jsonschema-form/projects/1). Feel free to suggest what you think should / shouldn't be on this board.

I've also created a [v2 branch](https://github.com/mozilla-services/react-jsonschema-form/tree/v2). If you are able to help out with some of those changes, please make PRs to the v2 branch. As we continue with work on v1, we can periodically merge master in v2. 

Once v2 is released, we will rename the current master branch to v1 and rename v2 to master. In the future, v1 may only be updated with bugfixes and/or security enhancements.

cefn:
I am currently using react-jsonschema-form from npm.

What's the compatibility status of the following?
* npm version
* repo master
* repo v2 

I am wondering if any of them have Bootstrap4 support as such, or if you're just expecting that Bootstrap3 will mostly work with Bootstrap4 (excepting glyphicons ...  https://github.com/rjsf-team/react-jsonschema-form/issues/1434#issuecomment-529235555 ) and that v2 is heading in the general direction of increased Bootstrap4 compatibility.

I've assumed there are no versioning issues with React although I'm using mainly hooks-based functional components which render to include a react-json-schema Form component.

epicfaace:
react-jsonschema-form is going in the direction of being UI-framework-agnostic, with themes being able to be added through the `withTheme` component. So, in the future, there will be a bootstrap 3 theme and a bootstrap 4 theme.

Support for bootstrap 4 isn't blocked by a v2 version as such; it could be implemented now if someone made a bootstrap 4 theme (like the material-ui theme in #1420). v2 is more a way to add in breaking changes (such as requiring React 16 instead of React 15).

white54503:
Any sense of the approximate timeline for a v2 release?

epicfaace:
v2 is now out. Thanks to all who helped get this out!

kim1882:
Thanks for this release. However I'm having some issues while updating the version with npm.
"No matching version found for react-jsonschema-form@2.0.0." Can you publish to https://www.npmjs.com/package/react-jsonschema-form ? Thanks in advance.

> v2 is now out. Thanks to all who helped get this out!

epicfaace:
[...]

PR #1604, opened 2020-02-24
chore: move docs to separate folder

chanceaclark:
## What

This PR is more of a proposal, in which I will outline below:

To get the repo into a state where we can effectively use TypeScript, I propose we use separation of concerns for the individual aspects of the repo. With this in mind we would be able to split the load on converting each package individually instead of an all at once approach. This would allow us to slowly convert to Typescript. Proposal (not in any particular order, unless noted):
- Move `docs/` into it's own package folder.
- Add `configs` package, for shared dev deps configs (i.e. `eslint`, `babel`, `tsconfig`, etc...).
- Move `playground/` into it's own package folder.
  - Convert to `NextJS` (**Optional/further discussion needed**). This would probably fall later in the process.
- Convert Webpack to Rollup. I think this might simplify the build process. Converting to Next would also benefit this step.

#1532, #1409, #583

ping @epicfaace


epicfaace:
Thanks for bringing this up. A few thoughts:
- What's the need for a separate package for docs?
- I did some work on creating a `@rjsf/playground` package in #1539, but got stuck because it wasn't able to build the playground package. Let me know if you could take a look and have any idea what may fix it!

chanceaclark:
> * What's the need for a separate package for docs?
Just to keep the `core/` package only for main functionality. Essentially just separation of concerns.

> * I did some work on creating a `@rjsf/playground` package in #1539, but got stuck because it wasn't able to build the playground package. Let me know if you could take a look and have any idea what may fix it!
Let me take a stab at it and see what I can do!

epicfaace:
> Just to keep the `core/` package only for main functionality. Essentially just separation of concerns.
> 
> >

Right, makes sense. I'm wondering why it should be in `packages`, though; perhaps it should be in the root directory instead?

chanceaclark:
> Right, makes sense. I'm wondering why it should be in `packages`, though; perhaps it should be in the root directory instead?

Good call! We should assume that any folder inside `packages` contains a `package.json`. I'll make that adjustment.

docs/mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 site_name: react-jsonschema-form documentation
-docs_dir: docs
+docs_dir: src
```
epicfaace:
```suggestion
docs_dir: docs/src
```

Just curious, why did you prefer using `docs/src` instead of `docs`?

chanceaclark: From [mkdocs documentation](https://www.mkdocs.org/user-guide/configuration/#docs_dir) it can be relative or absolute, so in this case it should be correct.

bricemciver:
I think moving the docs broke "readthedocs"

chanceaclark:
> I think moving the docs broke "readthedocs"

@epicfaace is there a process in which you build docs? If so, would you mind building them?

epicfaace:
Oops, my bad, should have tested to make sure it worked before merging it (maybe deploying the docs from that branch).

The docs get built automatically from master from readthedocs. I've just reverted this PR and that should fix it.

PR #1608, opened 2020-02-26
Maintain json schema error properties when using custom validation

RoboPhred:
This is a fix for 1.x.  After it lands, I will cherry-pick to 2.x and adjust as nescessary.

As described in #1596 

Enabling custom validation strips json schema errors of useful properties such as "property", "params", and "schemaPath".  All properties save "stack" are removed, preventing the error consumer from knowing where the error occurred or what rule it validated.

Additionally, custom errors do not provide a "property" prop, so the error consumer likewise does not know what property is encountering the custom error, just that the error happened 'somewhere'.

This PR fixes both these issues:

JSON Schema errors now keep their properties in the presence of custom validation.  This is done by not regenerating the entire error list, but instead keeping the old list and appending the custom errors to this list.

For custom errors, this PR modifies `toErrorList` to track the path of the error and generate a `property` prop containing the property path to the property that encountered the error.

I am not adding documentation, as the actual format of the `onError` errors array seems to be undocumented.  I can attempt to fill in some of the gaps here if you want to adopt the current ajv error objects as official api.

### Checklist

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -252,10 +252,8 @@ export default function validateFormData(
   const errorHandler = customValidate(formData, createErrorHandler(formData));
   const userErrorSchema = unwrapErrorHandler(errorHandler);
   const newErrorSchema = mergeObjects(errorSchema, userErrorSchema, true);
-  // XXX: The errors list produced is not fully compliant with the format
-  // exposed by the jsonschema lib, which contains full field paths and other
-  // properties.
-  const newErrors = toErrorList(newErrorSchema);
+  // Append the custom errors to the current error list.
+  const newErrors = [].concat(errors, toErrorList(userErrorSchema));
 
   return {
     errors: newErrors,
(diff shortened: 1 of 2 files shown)


epicfaace:
@RoboPhred thanks for the PR. Could you actually make this PR to master instead? Would like to keep updates to v1.x to be minimal changes such as security vulnerability fixes if possible.

Additionally, can you add a test that ensures that the error properties are maintained?

bmmpt:
@epicfaace Any reason this as gone stale? Are there plans to merge this in any time soon?

bmmpt:
@epicfaace thanks for the thumbs up and forgive my persistence. I don't mean to be pushy, but I didn't really get an answer to any of my questions.

epicfaace:
> Additionally, can you add a test that ensures that the error properties are maintained?

Yeah, it's the lack of unit tests, and also the fact that this PR isn't to the master branch.

bmmpt:
@epicfaace Thanks for the quick response. This is something that I will need pretty soon and was even considering forking and roll with forked code until this is completed. Instead of that, as soon as I have a few minutes I will add the missing tests if no one beats me to it.

PR #1610, opened 2020-02-27
Implementation of If Then Else

### LINKED ISSUE #850: Implement if/then/else syntax for conditional fields
### Prerequisites

### Description
The `dependencies` syntax is very powerful, however limitations are very quickly hit when dealing with fields across different objects.
A usecase we have internally is:
when: entity.legal_status != 'limited_company'
then: ask for applicant.address

As additional incentive, there is a [warning](https://github.com/json-schema-org/json-schema-spec/commit/6947eb1d29307ccdc062e0905336d7936e568d9f) in the spec that `dependencies` could disappear.

#### Expected behavior
based on the example in the [readme](https://github.com/mozilla-services/react-jsonschema-form/blob/5391d23528e9c2f9342e9e708452a4d6e3bd2c16/README.md#dynamic):

This can be written as chained if/then/else statements:
```json
{
  "if": {"properties": {"Do you have any pets?": {"enum": ["No"]}}},
  "else": {
    "if": {"properties": {"Do you have any pets?": {"enum": ["Yes: One"]}}},
    "then": {
      "properties": {"How old is your pet?": {"type": "number"}},
      "required": ["How old is your pet?"]
    },
    "else": {
      "...": "etc"
    }
  }
}
```

or using a oneOf:
```json
{
  "oneOf": [
    {
      "if": {"properties": {"Do you have any pets?": {"enum": ["No"]}}},
      "else": false
    },
    {
      "if": {"properties": {"Do you have any pets?": {"enum": ["Yes: One"]}}},
      "then": {
        "properties": {"How old is your pet?": {"type": "number"}},
        "required": ["How old is your pet?"]
      },
      "else": false
    },
    "..."
  ]
}
```
*the `false` on `else` is to force the validity to evaluate falsely, so `oneOf` moves on*

While implementing this, I came across a potential issue, in that [`utils.retrieveSchema`](https://github.com/mozilla-services/react-jsonschema-form/blob/5391d23528e9c2f9342e9e708452a4d6e3bd2c16/src/utils.js#L414) doesn't do a deep merge when resolving `$refs`.
Given that the `if`/`then`/`else` syntax opens up the possibilities of doing cross-object conditioning, is this something that might need fixing?
The main issue caused was that lists would not get appended, meaning if a `$ref` defined `required`, it would get discarded in favour of applied within a conditional.

If this is functionality people want here, I can prep a pull request.

### LINKED ISSUE #1636: Access enum value in schema
**Version: 1.8.1**

```
alertLevels: {
        type: "array",
        minItems: 1,
        maxItems: 2,
        hasRemove: true,
        items: {
          type: "object",
          required: ["severity", "upperThreshold", "lowerThreshold"],
          properties: {
            severity: {
              type: "string",
              title: "Alert Severity",
              enum: ["Critical", "Warning"]
            },
            isBulkAssign: {
              type: "boolean",
              title: "Assign Type",
              default: isOneByOne.indexOf(#severity#????) === -1,
              oneOf: [
                {
                  "const": true,
                  "title": "Bulk"
                },
                {
                  "const": false,
                  "title": "One By One"
                }
              ],
            },
            lowerThreshold: {
              type: "number",
              title: "Lower Threshold"
            },
            upperThreshold: {
              type: "number",
              title: "Upper Threshold"
            }
          }
        }
      }
```
Is there any way I can access the enum value of the field "severity" in the default property of "isBulkAssign"?
Thanks

Saulzi:

Implementation of If Then Else, will require some more work but this gives me enougth for the moment, note this is a partial implementation of if then else as does not support allOf (because of how this is currently merged) but does allow simple turning on and off of partial schema (see playground)

This is related to some tickets (need to hunt down) and also in relation to showing and hiding of fields.

Documentation will probably need updating down the line etc.

Lets get the ball rolling on this 

fixes #1636, fixes #850

### Checklist

### DIFF
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -17,7 +17,7 @@ import {
   getSchemaType,
 } from "../../utils";
 import UnsupportedField from "./UnsupportedField";
-import { isValid } from "../../validate";
+import { isValidAddDefaults } from "../../validate";
 
 const REQUIRED_FIELD_SYMBOL = "*";
 const COMPONENT_TYPES = {
@@ -246,7 +246,18 @@ function SchemaFieldRender(props) {
 
   if (schema.allOf) {
     for (var i = 0; i < schema.allOf.length; i++) {
-      schema = mergeSchemas(schema, schema.allOf[i]);
+      let allOfSchema = schema.allOf[i];
+
+      // if we see an if in our all of schema then evaluate the if schema and select the then / else, not sure if we should still merge without our if then else
+      if ("if" in allOfSchema) {
+        allOfSchema = isValidAddDefaults(allOfSchema.if, formData)
+          ? allOfSchema.then
+          : allOfSchema.else;
+      }
+
+      if (allOfSchema) {
+        schema = mergeSchemas(schema, allOfSchema);
+      }
     }
   }
 
@@ -366,7 +377,7 @@ function SchemaFieldRender(props) {
   let conditionalSchemaResult = null;
 
   if (schema.if) {
-    var conditionalSchema = isValid(schema.if, formData)
+    var conditionalSchema = isValidAddDefaults(schema.if, formData)
       ? schema.then
       : schema.else;
     if (conditionalSchema) {
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -1,4 +1,8 @@
-import { ADDITIONAL_PROPERTY_FLAG, mergeSchemas } from "../../utils";
+import {
+  ADDITIONAL_PROPERTY_FLAG,
+  mergeSchemas,
+  resolveSchema,
+} from "../../utils";
 import IconButton from "../IconButton";
 import React from "react";
 import PropTypes from "prop-types";
@@ -244,19 +248,29 @@ function SchemaFieldRender(props) {
   let idSchema = props.idSchema;
   var schema = retrieveSchema(props.schema, rootSchema, formData);
 
-  if (schema.allOf) {
-    for (var i = 0; i < schema.allOf.length; i++) {
-      let allOfSchema = schema.allOf[i];
+  let allOf = schema.allOf;
+
+  if (allOf) {
+    for (var i = 0; i < allOf.length; i++) {
+      let allOfSchema = allOf[i];
 
       // if we see an if in our all of schema then evaluate the if schema and select the then / else, not sure if we should still merge without our if then else
       if ("if" in allOfSchema) {
         allOfSchema = isValidAddDefaults(allOfSchema.if, formData)
           ? allOfSchema.then
           : allOfSchema.else;
+
+        console.log(allOfSchema);
+
+        if (allOfSchema) {
+          allOfSchema = resolveSchema(allOfSchema, rootSchema, formData); // resolve references etc.
+          console.log(allOfSchema);
+        }
       }
 
       if (allOfSchema) {
         schema = mergeSchemas(schema, allOfSchema);
+        delete schema.allOf;
       }
     }
   }
@@ -380,7 +394,14 @@ function SchemaFieldRender(props) {
     var conditionalSchema = isValidAddDefaults(schema.if, formData)
       ? schema.then
       : schema.else;
+
     if (conditionalSchema) {
+      conditionalSchema = resolveSchema(
+        conditionalSchema,
+        rootSchema,
+        formData
+      );
+
       var ConditionalSchemaComponent = getFieldComponent(
         conditionalSchema,
         uiSchema,
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -17,6 +17,7 @@ import {
   getSchemaType,
 } from "../../utils";
 import UnsupportedField from "./UnsupportedField";
+import { isValid } from "../../validate";
 
 const REQUIRED_FIELD_SYMBOL = "*";
 const COMPONENT_TYPES = {
@@ -355,6 +356,15 @@ function SchemaFieldRender(props) {
   const _AnyOfField = registry.fields.AnyOfField;
   const _OneOfField = registry.fields.OneOfField;
 
+  let conditionalSchema = null;
+  if (schema.if) {
+    if (isValid(schema.if, formData)) {
+      conditionalSchema = schema.then;
+    } else {
+      conditionalSchema = schema.else;
+    }
+  }
+
   return (
     <FieldTemplate {...fieldProps}>
       {field}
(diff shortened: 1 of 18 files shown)


packages/core/src/utils.js
```diff
@@ -655,18 +654,7 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
...
-        ...resolvedSchema,
-        allOf: resolvedSchema.allOf,
-      });
-    } catch (e) {
-      console.warn("could not merge subschemas in allOf:\n" + e);
-      const { allOf, ...resolvedSchemaWithoutAllOf } = resolvedSchema;
-      return resolvedSchemaWithoutAllOf;
-    }
-  }
+
```
epicfaace: Why remove the allOf logic?

Saulzi: @epicfaace  I have refactored allof and it now works more in line with https://json-schema.org/understanding-json-schema/reference/combining.html. before my changes the playground example does nothing, please take a look at updated playground example.

I still have to implement allof with if as per https://json-schema.org/understanding-json-schema/reference/conditionals.html?highlight=else.

another reasoning behind this removal is that current merge for allof does not work with if statement and giving it a resolver will not allow tie back of if / then

I aim to get allof working with if's to essentially do a case statement as I need fairly soon for the product which I am using this lib for.

thanks for your prompt response, hopefully we can get this nailed correctly in a prompt manner,



Likewise I still need to update the tests with regards to the allof stuff.

Saulzi:
@epicfaace I have now added allof support to if/then/else and updated tests and the playground

Saulzi:
Looking at this further, there is a slight issue which is happening with ajv when validating with unspecified value, when asking it to validate something against a constant when the value is undefined it is returning true when it should return false.

this can be seen in my playground example for allof ifthenelse when no value is selected.

Saulzi:
I have put in a workaround for this, shame current CI does not like referencing local package for this god only knows why your CI is failing. have you made changes?

Saulzi:
Please note netlify/rjsf/deploy-preview does not have latest playground content!
Is this perhaps because of references being against NPM rather than local,
I had real issues doing a local npm package in your mono repo, with npm you can reference local paths and git repos, for the mono-repo the playground should point at local content, something here broke your CI perhaps....

likewise real time has been wasted getting the playground and core packages to play nicely.

epicfaace:
> Please note netlify/rjsf/deploy-preview does not have latest playground content!
> Is this perhaps because of references being against NPM rather than local,

Yes, that's currently an issue with deploy previews. We can still see the latest code in this PR by checking it out locally and then running `npm start` from the `packages/core` directory though.

Perhaps a fix for this would be to configure netlify to run `npm link` in all the packages before building the playground deploy preview.

packages/core/src/utils.js
epicfaace: > before my changes the playground example does nothing, please take a look at updated playground example.

Thanks for pointing that out. The old allOf example was not the right example, but the allOf feature did work. I've added a working allOf example in this PR: https://github.com/rjsf-team/react-jsonschema-form/pull/1623/files#diff-4c77c9a609ff92a2b66c66edb321f45a

docs/src/index.md
```diff
@@ -194,7 +194,7 @@ This component follows [JSON Schema](http://json-schema.org/documentation.html)
 
     You can also use `oneOf` with [schema dependencies](dependencies.md#schema-dependencies) to dynamically add schema properties based on input data.
 
-    The `allOf` keyword is supported; it uses [json-schema-merge-allof](https://github.com/mokkabonna/json-schema-merge-allof) to merge subschemas to render the final combined schema in the form. When these subschemas are incompatible, though (or if the library has an error merging it), the `allOf` keyword is dropped from the schema. 
+    The `allOf` keyword is partually supported.
```
timkindberg: partially

packages/core/src/components/fields/SchemaField.js
```diff
@@ -241,7 +242,25 @@ function SchemaFieldRender(props) {
...
+
+      // if we see an if in our all of schema then evaluate the if schema and select the then / else, not sure if we should still merge without our if then else
+      if ("if" in allOfSchema) {
+        allOfSchema = isValidAddDefaults(allOfSchema.if, formData)
+          ? allOfSchema.then
+          : allOfSchema.else;
+      }
+
+      if (allOfSchema) {
+        schema = mergeSchemas(schema, allOfSchema);
```
Saulzi: @epicfaace please note that this merge is still incorrect as i believe the schema will take the last occurrence of an array, i.e. your example will take the last type, this is still incorrect as per https://json-schema.org/understanding-json-schema/reference/combining.html but this PR brings the allof functionality much closer to spec (in my mind)

docs/src/index.md
Saulzi: thanks ,

typo will be fixed in next commit, looking into this not working with a much more complicated schema, I aim to fix this today.

@epicfaace please see my comment on #1623, and take a look at the comment with regards to how schemas are now merged when rendering, while this is still not technically correct, it is much closer in my mind to the json schema specification.

Saulzi:
This now supports refs,

Looking at this perhaps the mutations to schema should not happen in render, i.e. should perhaps happen in resolveschema function in utils as this is given the data etc.

Saulzi:
@epicfaace @timkindberg 

Is it possible to review this, 

I will say that the current build process really needs some work.

Saulzi:
likewise I think netlify needs a longer timeout.

Saulzi:
realised I pushed without refs support.

packages/core/src/validate.js
```diff
@@ -275,3 +276,27 @@ export function isValid(schema, data) {
...
+        if (!("default" in property)) {
+          property["default"] = "";
+        }
+      }
+    }
+    return ajv.validate(schema, data);
+  } catch (e) {
+    return false;
+  }
+}
```
Saulzi: Looking at this, further I think that AJV could be correct here, 

https://github.com/epoberezkin/ajv/issues/913

this could be confusing but is basically a case of making the field required in schema,

I am looking at adding const support to this also as const is only partually supported, it should basically have default value and be disabled.
 

https://github.com/epoberezkin/ajv/issues/913

this

epicfaace:
@Saulzi , thanks for the PR.

However, if I try the following playground example:

```
{
  "type": "object",
  "allOf": [
    {
      "properties": {
        "lorem": {
          "type": [
            "string",
            "boolean"
          ],
          "default": true
        }
      }
    },
    {
      "properties": {
        "lorem": {
          "type": "boolean"
        },
        "ipsum": {
          "type": "string"
        }
      }
    }
  ]
}
```

I get nothing when running the code from this branch:

image

Whereas in the current playground (https://rjsf-team.github.io/react-jsonschema-form) I get:

image

Do we agree that what's rendered on the playground is correct in this case? It's great to add allOf support for if-then-else, but I'd like to add it in without removing the existing use cases for allOf that we already support.

Saulzi:
I don't agree with your playground image here.. were you looking at the last commit?

Saulzi:
likewise please see my other comments around relating to the playground not being latest after pulling / building etc.

norim13:
Any news on "if then else" support? It doesn't seem to be available yet... Is it expected to be available in stable versions any time soon?

PR #1617, opened 2020-02-28
Move docs to root-level directory, add to CI process

chanceaclark:

Attempts to fix comment from #1604.

### Checklist


epicfaace:
Just added a docs building step to the CI process so it won't break again.

PR #1623, opened 2020-03-01
Fix issues with new playground

### LINKED ISSUE #1620: Playground blurs after typing something
With the new playground update, the entire form blurs after typing something.

epicfaace:
Fixes #1620 

- Add options to select subthemes, so the old bootstrap subthemes (such as cerulean, etc.) are back
image

- Fix blurring when typing issue
- Use @rjsf/core instead of react-jsonschema-form package for the playground
- Fix allOf example

### DIFF
--- a/packages/playground/src/samples/allOf.js
+++ b/packages/playground/src/samples/allOf.js
@@ -5,7 +5,8 @@ module.exports = {
       {
         properties: {
           lorem: {
-            type: ["string", "number"],
+            type: ["string", "boolean"],
+            default: true,
           },
         },
       },
@@ -13,7 +14,6 @@ module.exports = {
         properties: {
           lorem: {
             type: "boolean",
-            minLength: 5,
           },
           ipsum: {
             type: "string",
--- a/packages/playground/package-lock.json
+++ b/packages/playground/package-lock.json
@@ -2965,6 +2965,22 @@
         "fastq": "^1.6.0"
       }
     },
+    "@rjsf/core": {
+      "version": "2.0.0-alpha.2",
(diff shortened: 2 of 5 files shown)


epicfaace:
@davidrenne can you take a look and review?

packages/playground/src/samples/allOf.js
```diff
@@ -5,15 +5,15 @@ module.exports = {
...
+            type: ["string", "boolean"],
+            default: true,
           },
         },
       },
       {
         properties: {
           lorem: {
             type: "boolean",
-            minLength: 5,
```
Saulzi: @epicfaace 

Like the addition of being able to select the bootstrap theme in the playground :P

with regards to my PR https://github.com/rjsf-team/react-jsonschema-form/pull/1610 and the allof, and my comments regarding merging of schema's

to quote

> To validate against allOf, the given data must be valid against all of the given subschemas.

> Note that it’s quite easy to create schemas that are logical impossibilities with allOf. The following example creates a schema that won’t validate against anything (since something may not be both a string and a number at the same time):

```
{
  "allOf": [
    { "type": "string" },
    { "type": "number" }
  ]
}
"No way"
```

https://json-schema.org/understanding-json-schema/reference/combining.html

here the type to validate against should be both a string and a boolean, this would essentially mean that we could never validate against the lorem field here as the entry could be both but i believe the merge will make it so that it picks the common one.

epicfaace: Yeah, exactly, so the schema in this file will, once merged, have `"lorem": {"type": "boolean", "default": true}`.

Saulzi: which I believe to be incorrect behavior

PR #1629, opened 2020-03-03
fix issue with large file uploads

erwinsmit:

Large files make the browser crash, with this code it will only parse files to base64 if they are smaller than roughly 5mb. If the file is larger a small text string will be parsed instead. This small text string is required to make the validation still pass when a large file is selected.

#1624 

### Checklist


packages/core/src/components/widgets/FileWidget.js
```diff
@@ -7,27 +7,6 @@ function addNameToDataURL(dataURL, name) {
   return dataURL.replace(";base64", `;name=${encodeURIComponent(name)};base64`);
 }
 
-function processFile(file) {
```
epicfaace, line 10: Can you re-convert this code to use the `processFile` and `processFiles` methods?

erwinsmit: @epicfaace Yes, no problem I have done this in the last commit.

erwinsmit: @epicfaace I'm having some issues getting the tests to pass since I pulled the latest changes into my branch.

The material-ui tests are failing with the following error:
ReferenceError: regeneratorRuntime is not defined

It seems that async/await is not supported with the current testing setup in the material-ui package.

epicfaace: Huh, you're right. Didn't realize that we didn't use async/await anywhere else in the codebase.

Will need to find a fix for this -- do you have any idea why it no longer works?

erwinsmit: I think it's babel related. I can investigate this later. I also saw that there were similar issues with tsdx but they should all be resolved in the latest version.

erwinsmit: I fixed the tests by adding the regenerator runtime to the setupTests file.

PR #1634, opened 2020-03-04
Do not regenerate state if no input data has changed.

RoboPhred:

RJSF regenerates state data on any property change and when its parent component changes, causing performance issues, data loss, and stack overflows, as documented in #1631.

This PR modifies getStateFromProps to return the same state if no new properties are detected.

To make this change, tests had to be updated to correctly call componentWillReceiveProps with all properties, as React normally calls it.  Previously, tests were only passing in one or two props, and getStateFromProps was set up to deal with this test-only behavior.

As usual, I am aiming to get this change into 1.x first, then port it to 2.x once it lands in 1.x

### Checklist

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -32,7 +32,7 @@ export default class Form extends Component {
 
   constructor(props) {
     super(props);
-    this.state = this.getStateFromProps(props, props.formData);
+    this.state = this.getStateFromProps(null, props.formData);
     if (
       this.props.onChange &&
       !deepEquals(this.state.formData, this.props.formData)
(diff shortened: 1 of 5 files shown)


RoboPhred:
I'm not sure why the build is failing.  It seems prettier is crashing, and playground cannot find a monaco package.  I changed nothing in playground, and prettier passes locally.  Additionally, these build steps work for all node versions != 6.  Is this a transient error?

epicfaace:
I've added an empty commit to try and re-run the Travis checks.

test/validate_test.js
```diff
@@ -689,7 +689,7 @@ describe("Validation", () => {
           validate,
           liveValidate: true,
         });
-        comp.componentWillReceiveProps({ formData });
+        comp.componentWillReceiveProps({ ...comp.props, formData });
 
         expect(comp.state.errorSchema).eql({
           0: { __errors: [] },
```
epicfaace: Can you add an actual test, if possible, that actually verifies this behavior?
- if the parent component changes, and none of the props change, the form should not rerender
- if the parent component changes, which causes one of the props to change, the form should rerender

src/components/Form.js
```diff
@@ -54,18 +54,43 @@ export default class Form extends Component {
...
+      nextProps.noValidate === this.props.noValidate &&
+      nextProps.idPrefix === this.props.idPrefix
+    ) {
+      // No data has changed, maintain state.
+      return state;
+    }
+
+    const schema = nextProps ? nextProps.schema : this.props.schema;
+    const uiSchema = nextProps ? nextProps.uiSchema : this.props.uiSchema;
+    const liveValidate = nextProps
```
epicfaace: Is there a particular reason why you changed the style here (from something like `props.liveValidate || this.props.liveValidate` to `nextProps ? nextProps.liveValidate : this.props.liveValidate`)?

src/components/Form.js
```diff
@@ -54,18 +54,43 @@ export default class Form extends Component {
...
+      ? nextProps.liveValidate
+      : this.props.liveValidate;
+    const customFormats = nextProps
```
epicfaace: Nice catch with `customFormats` and `additionalMetaSchemas` -- it didn't fall back to the props value previously. Do you know what kinds of bugs could have been caused by that omission previously? We may want to add tests that replicate those bugs and ensure that they don't regress.

epicfaace:
@RoboPhred thanks for the PR! By the way, it's probably best to get a fix into 2.x first and then backport it into 1.x (so we can make sure the latest version always has bugfixes - so there's never a case in which there is something  _better_ in a previous version), but this is fine too.

src/components/Form.js
RoboPhred: If liveValidate changed from true to false, props.liveValidate would pass over to this.props.liveValidate and liveValidate would remain on despite being turned off by the caller.

src/components/Form.js
RoboPhred: Shouldn't be any bugs, the only time partial props were being passed is from tests.  React itself will always give you every single property on componentWillReceiveProps, so in practice this check should be completely unnecessary.

test/validate_test.js
RoboPhred: I am not familiar with the test renderer you have set up.  I am not sure how to ask it to tell us if a rerender happened or not.  Possibly we can use a spy on the render function, or we could also test by passing in a validate function and ensuring it is not called if the data doesn't change.

Issue #701, opened 2017-09-19
Feature for embed UISchema into JSONSchema


qcho:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);

I'm proposing this feature before doing any coding or PR to see what the community and maintainers think about the change. I can create a PR with the changes (seems not a big feature) if it's not against any Design principles.

### Description

I want to propose a new feature where the UISchema is embedded within the JSONSchema.

Let's say we have these two definitions [from Simple example](https://mozilla-services.github.io/react-jsonschema-form)

**JSONSchema**:
``` json
{
  "title": "A registration form",
  "description": "A simple form example.",
  "type": "object",
  "required": [
    "firstName",
    "lastName"
  ],
  "properties": {
    "firstName": {
      "type": "string",
      "title": "First name"
    },
    "lastName": {
      "type": "string",
      "title": "Last name"
    },
    "age": {
      "type": "integer",
      "title": "Age"
    },
    "bio": {
      "type": "string",
      "title": "Bio"
    },
    "password": {
      "type": "string",
      "title": "Password",
      "minLength": 3
    },
    "telephone": {
      "type": "string",
      "title": "Telephone",
      "minLength": 10
    }
  }
}
```

**UISchema**:
```json
{
  "firstName": {
    "ui:autofocus": true,
    "ui:emptyValue": ""
  },
  "age": {
    "ui:widget": "updown",
    "ui:title": "Age of person",
    "ui:description": "(earthian year)"
  },
  "bio": {
    "ui:widget": "textarea"
  },
  "password": {
    "ui:widget": "password",
    "ui:help": "Hint: Make it strong!"
  },
  "date": {
    "ui:widget": "alt-datetime"
  },
  "telephone": {
    "ui:options": {
      "inputType": "tel"
    }
  }
}
```

Is there any reason why we can't just have one "extended" JSONSchema object (this is actually allowed by the JSONSchema spec) with the UI definitions instead of replicating the entire mapping?

The result would be **JSONSchema + UISchema**:
```json
{
  "title": "A registration form",
  "description": "A simple form example.",
  "type": "object",
  "required": [
    "firstName",
    "lastName"
  ],
  "properties": {
    "firstName": {
      "type": "string",
      "title": "First name",
      "ui:autofocus": true,
      "ui:emptyValue": ""
    },
    "lastName": {
      "type": "string",
      "title": "Last name"
[...]

PR #1638, opened 2020-03-09
#1635 #701 allow merging of UI Schema from JSON Schema

Saulzi:

As per #1635 and #701 

### Checklist

### DIFF
--- a/packages/core/test/validate_test.js
+++ b/packages/core/test/validate_test.js
@@ -49,6 +49,26 @@ describe("Validation", () => {
 
       expect(isValid(schema, { foo: "Test" })).to.be.true;
     });
+
+    it("should allow extra ui fields with ref", () => {
+      const schema = {
+        type: "object",
+        properties: {
+          foo: {
+            $ref: "#/definitions/foo",
+          },
+        },
+        definitions: {
+          foo: {
+            ui: {
+              widget: "hidden",
+            },
+          },
+        },
+      };
+
+      expect(isValid(schema, { foo: "Test" })).to.be.true;
+    });
   });
 
   describe("validate.validateFormData()", () => {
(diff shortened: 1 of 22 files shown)


Saulzi:
@epicfaace as per your comment, please note that we still need to update documentation and the decision needs to be made as to 'cut' away the current uiSchema in v2.0

packages/core/test/validate_test.js
```diff
@@ -35,6 +35,20 @@ describe("Validation", () => {
 
       expect(isValid(schema, { foo: "bar" })).to.be.false;
     });
+
+    it("should allow extra ui fields", () => {
```
epicfaace: Can you add some tests which involve using the `ui` property in more deeply nested examples, with references, etc.?

Saulzi: > Thanks for making this! I wonder if it might be better to make the `ui` property as below so that it would be easier to transition old uiSchemas to bring them to the new format:
> 
> ```
> ui: {
>     "ui:widget": "x",
> }
> ```

I am not sure about this, we could do and it would be a case of just changing the merge in SchemaField but as it is changing to under ui field having extra ui: item is extra typing.

> Also, perhaps we can support `uiSchema` (in addition to the `ui` property) in v2, but just deprecate it, and remove it in a later version. We could make it so that anything in the `ui` property overrides anything that was already in the `uiSchema`.

I believe this is the case with this functionality.

I will do something in the playground here and add some additional tests allthough this is working for me with some complex schemas under my if/then/else stuff

Saulzi:
@epicfaace I have added a couple of extra tests with regards to $ref

Saulzi:
@epicfaace this has somewhat fell behind. whats going on here?

epicfaace:
Haven't gotten a chance to take a look much further -- will give it a further review soon.

AayushB:
@epicfaace is there still a plan to address this?

PR #1639, opened 2020-03-09
Add docs for contributing

epicfaace:

Add docs for contributing on a separate page, and also explain the development / deployment process for the new multi-package setup.

### Checklist


epicfaace:
@juliankigwana @Saulzi @chanceaclark let me know if this is clearer or if the docs I've added might need more clarification?

Saulzi:
Yes definitely clarifies things, I will give this workflow a go now

Saulzi:
unfortunately this did not work on #1610

docs/contributing.md
```diff
@@ -0,0 +1,77 @@
...
+Note that when doing local development in a package such as `packages/core`, making changes to `packages/playground` will not affect the local playground that is run from `packages/core`. This should be fine in most cases for development, unless you need to update and test the default playground examples. In this case, you could use `npm link` to use the latest changes both from the playground and your package. As an example, to do this when developing locally on `packages/core`, you can run:
+
+```bash
+cd packages/playground
+npm run build:lib
+npm link
+cd ../core
+npm link @rjsf/playground
+npm start
+```
```
Saulzi: Perhaps this should be done when the user pulls the package rather than an additional step otherwise this works for me 👍

Saulzi:
> unfortunately this did not work on #1610

this did after install on bottom level and setting back index.js to be import Form, { withTheme } from "@rjsf/core";  in packages/playground on my branch

epicfaace:
Would it be simpler, rather than having each package have its own playground start scripts, to just have each package build their files (as @chanceaclark 's PR in #1606 does)? Then, in order to run the playground, one would just need to `npm link` whichever package they are working on, then run the playground from the `packages/playground` directory.

This thought came up because I was trying to add the playground to material-ui but realized that doing so would require a lot of overhead for each new package we add.

PR #1585, opened 2020-02-07
feature: add semantic ui theme form components

jacqueswho:
### Reasons for making this change

This is my first pull request and addition to the open-source community. Could you please let me know what else I need to do. My source of inspiration was what was done for rjsf-material-ui project. Because the project was done in typescript, I could not follow the build guide. Could you perhaps assist me with this?

#1537 

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [x] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #1641, opened 2020-03-09
Setup material-ui tests

### LINKED ISSUE #1655: Create separate @rjsf/utils package
Original comment:
> I say we break out the utils package since it'll be better to maintain that way. If we need to make a path or minor to the utils package, in theory it shouldn't break anyones installs if they are using a `^` (i.e. `"@rjsf/utils": "^X.X.X"`).
> 
> For the fields and widgets, they should probably just be exposed in the root index file still so you can import them like so:
> ```tsx
> import { ArrayField } from '@rjsf/core';`
> ```
> 
> We would just need to _ensure_ tree shaking is set up correctly, that way they don't import everything from `@rjsf/core`.

_Originally posted by @chanceaclark in https://github.com/rjsf-team/react-jsonschema-form/pull/1642_

epicfaace:

Add tests to the material-ui theme so that it can be released.

Also, update the docs to discuss the themes in more detail.

Finally, refactors utils to be exported from the default @rjsf/core package so that they are properly transpiled -- fixes #1655 (this had to be done in order to get material-ui tests to work, so that's why both are done in the same PR)

So, previously an import would look like this:

```js
import { getDefaultRegistry } from "@rjsf/core/lib/utils";
```

Now you would need to do this:

```js
import { utils } from "@rjsf/core";
import { getDefaultRegistry } from utils;
```

Related: #1561 #1585 -- we'd have to get the theme documentation in first before merging these two PRs.


heath-freenome:
Can we get a release with these fixes soon?

epicfaace:
Soon!

heath-freenome:
I really hope so, I'm starting to reach the point where this is almost completely blocking me

Morriz:
Yes please, so that we may override `ArrayFieldTemplate` properly to inject our custom `AddButton` to use color='primary' instead of secondary.

PR #1642, opened 2020-03-10
Playground and development environment fixes for Lerna

### LINKED ISSUE #1583: Export typings from @rjsf/core
@epicfaace was wondering if there was plans for a new alpha release soon? Hasn't been one since Nov of last year.

### LINKED ISSUE #1630: Troubles with the new Playground
I'm running into a host of problems with the new Playground setup that's made local development impossible.

Running `npm start` from within `packages/playground` greets me with compilation errors like:

```
ERROR in ./src/index.js
Module not found: Error: Can't resolve '@rjsf/core' in '/Users/jon/code/react-jsonschema-form/packages/playground/src'
 @ ./src/index.js 18:0-45 360:29-33 384:29-33 544:23-32 626:21-30 682:29-33
 @ ./playground/app.js
 @ multi webpack-hot-middleware/client?reload=true ./playground/app
```

Running the same from within `packages/core` I get:

```
ERROR in ./playground/app.js
Module not found: Error: Can't resolve '@rjsf/playground' in '/Users/jon/code/react-jsonschema-form/packages/core/playground'
 @ ./playground/app.js 3:0-42 10:27-37
 @ multi webpack-hot-middleware/client?reload=true ./playground/app
```

I've run `npm run bootstrap` from the root directory but still can't get it booted. I am running the most current `master`.

Am I doing something wrong?

epicfaace:

Building off of work done in #1606 and #1641.

- Update build and dev process and documentation (as discussed in https://github.com/rjsf-team/react-jsonschema-form/issues/1630#issuecomment-597163950). The developer will just need to run `lerna bootstrap`, run `npm start` from the package they are modifying, then run the playground from `packages/playground` to get a live-reloading playground.
- Fix deploy previews by configuring netlify so that it runs `lerna bootstrap`, thus linking the packages with the local changes in the PR, before building the playground.
- Export typings from `@rjsf/core` (fixes #1583) -- did this in this PR so that `@rjsf/material-ui` would work as well.
- Fix playground issues in which the `material-ui` playground did not use the right ObjectFieldTemplate or ArrayFieldTemplate.

Fixes #1630 .

### Checklist


packages/core/package.json
```diff
@@ -3,32 +3,33 @@
...
   "scripts": {
-    "build:lib": "rimraf lib && cross-env NODE_ENV=production babel -d lib/ src/",
-    "build:dist": "rimraf dist && cross-env NODE_ENV=production webpack --config webpack.config.dist.js",
-    "cs-check": "prettier -l \"{playground,src,test}/**/*.js\"",
-    "cs-format": "prettier \"{playground,src,test}/**/*.js\" --write",
-    "dist": "npm run build:lib && npm run build:dist",
-    "lint": "eslint src test playground",
+    "build": "npm run build:cjs && npm run build:es",
+    "build:cjs": "NODE_ENV=production BABEL_ENV=cjs babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./dist/cjs",
+    "build:es": "NODE_ENV=production BABEL_ENV=es babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./dist/es",
```
Saulzi: Why no umd, is there still a umd build? as this is what I am using! basically I include the whole lib and allow other plugins to use version pulled in via umd with requirejs, potentually umd works with both commonjs loaders, iife i.e. the browser straight and AMD loaders

packages/core/package.json
```diff
@@ -3,32 +3,34 @@
...
-    "build:lib": "rimraf lib && cross-env NODE_ENV=production babel -d lib/ src/",
-    "build:dist": "rimraf dist && cross-env NODE_ENV=production webpack --config webpack.config.dist.js",
-    "cs-check": "prettier -l \"{playground,src,test}/**/*.js\"",
-    "cs-format": "prettier \"{playground,src,test}/**/*.js\" --write",
-    "dist": "npm run build:lib && npm run build:dist",
-    "lint": "eslint src test playground",
+    "build": "npm run build:cjs && npm run build:es && npm run build:es:lib",
+    "build:cjs": "NODE_ENV=production BABEL_ENV=cjs babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./dist/cjs",
+    "build:es": "NODE_ENV=production BABEL_ENV=es babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./dist/es",
+    "build:es:lib": "NODE_ENV=production BABEL_ENV=es babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./lib",
```
epicfaace: @chanceaclark I had to add this line in because we needed to import things like `@rjsf/core/lib/utils`

packages/core/package.json
epicfaace: Not sure, any thoughts @chanceaclark ?

epicfaace:
this is now ready for review and CI is passing - @erunion @chanceaclark @Saulzi

netlify.toml
```diff
@@ -1,7 +1,7 @@
 [build]
-  base = "packages/playground"
-  publish = "build"
-  command = "npm run build:playground"
+  base = ""
+  publish = "packages/playground/build"
+  command = "npm run bootstrap && (npm run build || { sleep 120; false; })"
```
Saulzi: Is the sleep still needed here?

packages/core/package.json
Saulzi: I will check that I still get a UMD Module, this is vital for me and perhaps other users... essentially I pull this from the build output

packages/core/package.json
```diff
@@ -123,5 +128,6 @@
   "homepage": "https://github.com/mozilla-services/react-jsonschema-form#readme",
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "types": "index.d.ts"
```
Saulzi: this is definitely an improvement, not sure what needs to be done to get them to remove this from @types as essentially this means that this project has taken ownership

packages/playground/src/index.js
```diff
@@ -1,620 +1,18 @@
...
-  } else if (a instanceof Date && b instanceof Date) {
-    return a.getTime() === b.getTime();
-  } else if (a instanceof RegExp && b instanceof RegExp) {
-    return (
-      a.source === b.source &&
-      a.global === b.global &&
-      a.multiline === b.multiline &&
-      a.lastIndex === b.lastIndex &&
-      a.ignoreCase === b.ignoreCase
-    );
```
Saulzi: are some of these functions common i.e. do we have deep equality in the lib etc?

happy but DRY?

packages/core/package.json
Saulzi: The NODE_ENV stuff here seems to not work on windows

Saulzi: please see my other comments relating to UMD builds here, as this currently won't build for me I cannot check but basically need output as per UMD (this is commonjs iffe and amd all in one)

i.e.. 
``` js
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
        // AMD. Register as an anonymous module.
        define(['exports', 'b'], factory);
    } else if (typeof exports === 'object' && typeof exports.nodeName !== 'string') {
        // CommonJS
        factory(exports, require('b'));
    } else {
        // Browser globals
        factory((root.commonJsStrict = {}), root.b);
    }
}(this, function (exports, b) {
    //use b in some fashion.

    // attach properties to the exports object to define
    // the exported module properties.
    exports.action = function () {};
}));
```

Saulzi: this basically needs cross-env to be stuck in front in package.json's.. already there in other places.

packages/core/package.json
chanceaclark: I think we might have to change build tools then. `babel` has a plugin for converting to UMD but it's kind of a all or nothing approach.

@Saulzi, if you mention umd works with commonjs (cjs) loaders, couldn't you just load in the cjs modules?

In the meantime, play around with using rollup and try to push a commit.

packages/core/package.json
chanceaclark: @epicfaace I'm guessing that's to keep support for it. What's to stop people from using doing this now:
```tsx
import { deepEquals } from '@rjsf/core/dist/utils';`
```
IMO, it makes more sense to export the utils along with package or creating a utils package if you want to separate it.

Also, you are probably aware, but keep in mind this is still part of the alpha version so there is more freedom in introducing breaking changes.

packages/core/package.json
chanceaclark: We will probably want to keep the @types typings to support people using the old package `react-jsonschema-form`. Since we've update the package name to `@rjsf/core` the typings will be included, so there will be no need to import `@types/react-jsonschema-form`.

packages/playground/src/index.js
chanceaclark: I've pointed out that there is a bug in this: https://github.com/rjsf-team/react-jsonschema-form/pull/1438#issuecomment-526318486

We still need to support this function until we can really refactor using an external lib. My recommendation is to convert everything to TS before we can tackle this.

packages/core/package.json
Saulzi: Yes it works with both commonjs and AMD loaders please refer to https://www.davidbcalhoun.com/2014/what-is-amd-commonjs-and-umd/
https://riptutorial.com/javascript/example/16339/universal-module-definition--umd-
https://www.google.com/search?q=common+javascript+modules&oq=common+javascript+modules&aqs=chrome..69i57.11167j0j7&sourceid=chrome&ie=UTF-8

AMD has been around for some time before the ES6 (ESM) standard.. i.e. dojo framework had amd support in around since 2011, many javascript developers have used things such as jquery etc which have used this pattern for some time and not even realised.. for example react builds to a umd module other popular javascript frameworks such as leaflet, moment, knockout and chart.js are built to umd modules.

@chanceaclark from experience rollup will make porting to typescript much easier...
as in my mind webpack should be taken out back and shot (have you ever seen the intro to stealjs video), crazy amount of configuration etc and the whole node_modules black hole. I will say that steal is good for sites but not libs and another potential is parcel.

packages/core/package.json
Saulzi: Please note that this change in package.json to build both commonjs and esm really breaks things for me, basically there was one output before which was an amd module i.e. in the dist folder which I would dump with all my other js files

Saulzi: note this now dumps lots of small files and not a single minified / bundled file

packages/core/package.json
```diff
@@ -3,32 +3,34 @@
   "version": "2.0.0-alpha.2",
   "description": "A simple React component capable of building HTML forms out of a JSON schema.",
   "scripts": {
-    "build:lib": "rimraf lib && cross-env NODE_ENV=production babel -d lib/ src/",
-    "build:dist": "rimraf dist && cross-env NODE_ENV=production webpack --config webpack.config.dist.js",
```
Saulzi: This line used to cause a umd module

packages/core/package.json
epicfaace: @saulzi can you share some example code that relies on the UMD build? That way I can make sure that the way it outputs will work with that code.

netlify.toml
epicfaace: No

packages/core/package.json
epicfaace: Yeah, good point. It seems like `react-jsonschema-form/lib/utils` is already used, though, see material-ui (https://github.com/rjsf-team/react-jsonschema-form/pull/1642/files#diff-92ef0ee17db18857941b12442a459de0R6) or #1643

Additionally, the documentation even has an example of using an actual component from the `lib` directory (https://github.com/rjsf-team/react-jsonschema-form/pull/1642/files#diff-443373237a67e0fe395c3da27e3595ebR553).

```
import SchemaField from "@rjsf/core/lib/components/fields/SchemaField";
```

This is probably bad, as we probably don't want to constrain ourselves with file structure. However, it would also be useful for users to be able to access the default fields / widgets. Perhaps it would be best to just make a `@rjsf/utils` package and expose fields through `@rjsf/core/fields` and widgets through `@rjsf/core/widgets`?

packages/core/package.json
epicfaace: I've updated it to use the old webpack config to create the UMD build again. We can definitely switch to rollup but for now I'm just adding the old webpack config which works and won't affect local development.

packages/core/package.json
epicfaace: > this basically needs cross-env to be stuck in front in package.json's.. already there in other places.

Done

epicfaace:
ready for additional review @chanceaclark @saulzi

packages/core/package.json
chanceaclark: I say we break out the utils package since it'll be better to maintain that way. If we need to make a path or minor to the utils package, in theory it shouldn't break anyones installs if they are using a `^` (i.e. `"@rjsf/utils": "^X.X.X"`).

For the fields and widgets, they should probably just be exposed in the root index file still so you can import them like so:
```tsx
import { ArrayField } from '@rjsf/core';`
```

We would just need to _ensure_ tree shaking is set up correctly, that way they don't import everything from `@rjsf/core`.

epicfaace: Yeah, sounds good. Let's do that in another issue / PR.

packages/core/package.json
```diff
@@ -3,32 +3,35 @@
...
-    "{playground,src,test}/**/*.js": [
+    "{src,test}/**/*.js": [
       "eslint --fix",
       "prettier --write",
       "git add"
     ]
   },
-  "main": "lib/index.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/es/index.js",
```
chanceaclark: Probably want to add this instead of including it in the `files` array.
```suggestion
  "module": "dist/es/index.js",
  "typings": "index.d.ts"
```

packages/core/package.json
```diff
@@ -3,32 +3,35 @@
...
   "files": [
     "dist",
-    "lib"
+    "index.d.ts"
```
chanceaclark: Remove from above.

packages/core/package.json
```diff
@@ -3,32 +3,35 @@
...
-    "build:dist": "rimraf dist && cross-env NODE_ENV=production webpack --config webpack.config.dist.js",
-    "cs-check": "prettier -l \"{playground,src,test}/**/*.js\"",
-    "cs-format": "prettier \"{playground,src,test}/**/*.js\" --write",
-    "dist": "npm run build:lib && npm run build:dist",
-    "lint": "eslint src test playground",
+    "build": "npm run dist:build:umd && npm run build:cjs && npm run build:es && npm run dist:build:es:lib",
+    "build:cjs": "cross-env NODE_ENV=production BABEL_ENV=cjs babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./dist/cjs",
+    "build:es": "cross-env NODE_ENV=production BABEL_ENV=es babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./dist/es",
+    "dist:build:es:lib": "cross-env NODE_ENV=production BABEL_ENV=es babel --extensions \".js,.jsx,.ts,.tsx\" ./src --out-dir ./lib",
+    "dist:build:umd": "rimraf dist && cross-env NODE_ENV=production BABEL_ENV=umd webpack --config webpack.config.dist.js",
```
chanceaclark: Yikes. But we'll figure it out.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "react-jsonschema-form",
+  "name": "@rjsf/core",
```
chanceaclark: You probably don't want to change this. It should be the name of the repo if it's a monorepo.
```suggestion
  "name": "react-jsonschema-form",
```

packages/material-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -5,8 +5,8 @@ import MenuItem from '@material-ui/core/MenuItem';
 import Select from '@material-ui/core/Select';
 import InputLabel from '@material-ui/core/InputLabel';
 
-// import { WidgetProps } from 'react-jsonschema-form';
-import { asNumber, guessType } from 'react-jsonschema-form/lib/utils';
+// import { WidgetProps } from '@rjsf/core';
```
chanceaclark: May want to remove this.

packages/core/package.json
epicfaace: Done

packages/core/package.json
epicfaace: Done

package.json
epicfaace: Done

epicfaace: What do you think should be done with the `version` attribute? I don't think Lerna automatically bumped it upon release, and I'm not sure if `version` is even relevant in the top of the monorepo.

epicfaace:
I wonder whether it might be better to move `playground` from `packages/playground` to the root-level directory (`playground`) -- because I don't think we'll ever be publishing it as a package?

chanceaclark:
> I wonder whether it might be better to move `playground` from `packages/playground` to the root-level directory (`playground`) -- because I don't think we'll ever be publishing it as a package?

You won't be able to link `@rjsf/core` and the material-ui packages if we do. As long as there is a `"private": true` key/value in the package.json, it won't get published at all.

packages/core/package.json
Saulzi: > @Saulzi can you share some example code that relies on the UMD build? That way I can make sure that the way it outputs will work with that code.

@epicfaace @chanceaclark 

https://github.com/Saulzi/react-jsonschema-form/tree/amd-example
Here is an example of using react-jsonschema-form via UMD. using all three legacy module loader types across node, and the browser.
 
I found a few things i.e. synthetic defaults etc with the current build...

Why build commonjs when UMD does it all for legacy module loaders.. essentially newer build tools such as rollup / webpack use ESM, node is stuck using commonjs and browser can essentually do amd (via steal / require / dojo / curl?), commonjs (stealjs), umd script tags and esm modules.. 

things would be so much better if react was built to esm module because then it would play nicer with rollup etc and would allow me to load it via script type module rather than relying on AMD.

I am glad this commit is now in merged as its definitely an improvement but still needs a little bit of tweaking, @epicfaace does this also resolve issues with CI i.e. is the playground preview now up to date?

epicfaace: @saulzi
You said "I found a few things i.e. synthetic defaults etc with the current build..." -- did you mean that you are able to use the UMD version now or are there still issues?

Saulzi: @epicfaace it does work but really when referencing this in the browser and via amd you shouldn't have to use .default... it is basically because of synthetic exports.

PR #1644, opened 2020-03-11
validate: pass formData to transformErrors

romgrk:

`transformErrors()` might need access to the data.

### Checklist

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -109,7 +109,7 @@ render((
 Validation error messages are provided by the JSON Schema validation by default. If you need to change these messages or make any other modifications to the errors from the JSON Schema validation, you can define a transform function that receives the list of JSON Schema errors and returns a new list.
 
 ```js
-function transformErrors(errors) {
+function transformErrors(errors, formData) {
   return errors.map(error => {
     if (error.name === "pattern") {
       error.message = "Only digits are allowed"
--- a/packages/core/src/validate.js
+++ b/packages/core/src/validate.js
@@ -229,7 +229,7 @@ export default function validateFormData(
     ];
   }
   if (typeof transformErrors === "function") {
-    errors = transformErrors(errors);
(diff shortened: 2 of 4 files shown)


romgrk:
The example isn't super useful in this case as it only shows the current value, but this can be used to provide more helpful error message, eg by suggesting possible values depending on the current value.

epicfaace:
@romgrk thanks for the PR. How needed do you think this change is, actually? One could still access formData through transformErrors if they made `<Form>` a controlled component and accessed the formData through something like the state or used hooks.

Also, if we were to go this path to expose the formData in transformErrors, then that raises the question about why not expose schema, etc. -- it seems simpler just to let the user use the aforementioned solution to access the appropriate data needed.

romgrk:
I don't have strong opinions on that, but if I may argue for the way this PR does it:
 - Making `<Form>` controlled increases complexity for the user, there is more to manage.
 - Passing `formData` to `transformErrors` is way simpler overall for rjsf than to let it up to the users to access `formData` through their own state. The complexity increase is really low, and `transformErrors` has an arity of 2, which is still acceptable.
 - This allows to decouple `transformErrors` from the component in which `<Form>` is embedded. For example, in my use-case `transformErrors` is a function in a separate module, re-used by various components which have similar error patterns.

You can close the PR if you don't think it's a good addition, no hard feelings.

epicfaace:
@romgrk those are valid points, thanks.

> Also, if we were to go this path to expose the formData in transformErrors, then that raises the question about why not expose schema, etc.

What do you think about this concern?

romgrk:
I don't have a use case for it, but other arguments could be exposed indeed. Maybe for this PR I would simply add `schema`. I don't see `uiSchema` or other arguments being useful in this context. If you're ok with this proposition, I'll make the changes on Monday.

PR #1652, opened 2020-03-15
submitOnEnter prop = false will add event listener to prevent form submit when pressing enter

mirajp:

Issue #1597 

### Checklist

  - [ ] I've added and/or updated tests **(see below)**
image

---------

I tried adding a test, but got stuck on verifying the 'bug', i.e. when the enter key is pressed, the form gets submitted. No matter what I did, I couldn't programmatically fire the enter key via jsdom/react-dom TestUtils.Simulate. This is among the stuff I tried:

```
describe("Form submit", () => {
    it("by default pressing enter should invoke submit event which calls the submit function", () => {
        const schema = {
            type: "object",
            properties: {
                foo: { type: "string" },
            },
        };

        const { node, onSubmit } = createFormComponent({ schema });

        const input = node.querySelector("input[type=text]");

        input.addEventListener('keypress', (e) => console.log(e));
        input.addEventListener('keydown', (e) => console.log(e));

        Simulate.keyDown(input, {key: "Enter", keyCode: 13, which: 13, charCode: 13, bubbles: true});
        Simulate.keyPress(input, {key: "Enter", keyCode: 13, which: 13, charCode: 13, bubbles: true});
        Simulate.keyUp(input, {key: "Enter", keyCode: 13, which: 13, charCode: 13, bubbles: true});

        Simulate.submit(input);
        // Simulate.submit(node);

        console.log('onSubmit.called:', onSubmit.called);

        sinon.assert.calledOnce(onSubmit);
    });
});
```


docs/index.md
```diff
@@ -120,6 +120,20 @@ render((
 
 > Note: If there are fields in the `formData` that are not represented in the schema, they will be retained by default. If you would like to remove those extra values on form submission, then set the `omitExtraData` prop to `true`. Set the `liveOmit` prop to true in order to remove extra data upon form data change.
 
+#### Prevent early form submission (via enter key press)
```
epicfaace, line 123: thanks -- would it be possible to add a test?

mirajp: I couldn't figure out how to get the test to show the default behavior, i.e. the submit event fired when pressing the enter key. My PR description has a bit of the code I was tried that didn't work (submit function was never called).

PR #1657, opened 2020-03-16
Add CI with GitHub Actions, disable material-ui test running

epicfaace:
- Switch to CI for GitHub Actions
- Run build on Node 13 as well.
- It turns out the Azure Pipelines build was silently failing material-ui tests. Disabling the material-ui test running (there are no tests anyway as of now) so at least the build fails.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [8.x, 9.x, 10.x, 11.x, 12.x]
```
erunion: May as well!

```suggestion
        node-version: [8.x, 9.x, 10.x, 11.x, 12.x, 13.x]
```

PR #1660, opened 2020-03-18
feat: adding support for overriding UnsupportedField

erunion:

We're working on some error handling changes in our app to funnel errors through to an error tracker, and can't do that without being able to override the `UnsupportedField` component to throw an actual `Error`.

This adds support for being able to override `UnsupportedField` as you would any other.

### Checklist


packages/core/src/components/fields/SchemaField.js
```diff
@@ -49,6 +48,9 @@ function getFieldComponent(schema, uiSchema, idSchema, fields) {
   return componentName in fields
     ? fields[componentName]
     : () => {
+        const { fields } = registry;
```
epicfaace: Why use `registry.fields.UnsupportedField` instead of using `fields.UnsupportedField`? I'm trying to see why UnsupportedField requires pulling from the registry while the other fields don't.

packages/core/src/components/fields/SchemaField.js
```diff
@@ -16,7 +16,6 @@ import {
   deepEquals,
   getSchemaType,
 } from "../../utils";
-import UnsupportedField from "./UnsupportedField";
 
 const REQUIRED_FIELD_SYMBOL = "*";
 const COMPONENT_TYPES = {
```
epicfaace: Do you mind adding the list of fields you can override (which are defined here) to the docs, and add UnsupportedField to it? https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/#customizing-the-default-fields-and-widgets

packages/core/src/components/fields/SchemaField.js
erunion: Ha, well this is embarassing.

erunion:
@epicfaace Made those changes!

docs/advanced-customization.md
```diff
@@ -620,6 +620,20 @@ render((
 ), document.getElementById("app"));
 ```
 
+The default fields you can overwrite are:
```
dekelb: Pretty sure it should be "override"

PR #1666, opened 2020-03-20
Implementation of If Then Else / Const v2

Saulzi:
@epicfaace Please find updated if/then/else stuff to review

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1,6 +1,5 @@
 import React from "react";
 import * as ReactIs from "react-is";
-import mergeAllOf from "json-schema-merge-allof";
 import fill from "core-js/library/fn/array/fill";
 import validateFormData, { isValid } from "./validate";
 import union from "lodash/union";
@@ -171,6 +170,8 @@ function computeDefaults(
   } else if ("default" in schema) {
     // Use schema defaults for this node.
     defaults = schema.default;
+  } else if ("const" in schema) {
+    defaults = schema.const;
   } else if ("$ref" in schema) {
     // Use referenced schema defaults for this node.
     const refSchema = findSchemaDefinition(schema.$ref, rootSchema);
@@ -655,28 +656,63 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
     return {};
   }
   let resolvedSchema = resolveSchema(schema, rootSchema, formData);
-  if ("allOf" in schema) {
-    try {
-      resolvedSchema = mergeAllOf({
-        ...resolvedSchema,
-        allOf: resolvedSchema.allOf,
-      });
-    } catch (e) {
-      console.warn("could not merge subschemas in allOf:\n" + e);
-      const { allOf, ...resolvedSchemaWithoutAllOf } = resolvedSchema;
-      return resolvedSchemaWithoutAllOf;
-    }
-  }
+
   const hasAdditionalProperties =
     resolvedSchema.hasOwnProperty("additionalProperties") &&
     resolvedSchema.additionalProperties !== false;
+
   if (hasAdditionalProperties) {
-    return stubExistingAdditionalProperties(
+    resolvedSchema = stubExistingAdditionalProperties(
       resolvedSchema,
       rootSchema,
       formData
     );
   }
+
+  if ("if" in resolvedSchema) {
+    // Note that if and else are key words in javascript so extract to variable names which are allowed
+    var {
+      if: expression,
+      then,
+      else: otherwise,
+      ...resolvedSchemaLessConditional
+    } = resolvedSchema;
+    var conditionalSchema = isValid(expression, formData) ? then : otherwise;
+
+    if (conditionalSchema) {
+      conditionalSchema = resolveSchema(
+        conditionalSchema,
+        rootSchema,
+        formData
+      );
+      resolvedSchema = mergeSchemas(
+        resolvedSchemaLessConditional,
+        conditionalSchema
+      );
+    }
+  }
+
+  let allOf = resolvedSchema.allOf;
+
+  if (allOf) {
+    for (var i = 0; i < allOf.length; i++) {
+      let allOfSchema = allOf[i];
+
+      // if we see an if in our all of schema then evaluate the if schema and select the then / else, not sure if we should still merge without our if then else
+      if ("if" in allOfSchema) {
+        allOfSchema = isValid(allOfSchema.if, formData)
+          ? allOfSchema.then
+          : allOfSchema.else;
+      }
+
+      if (allOfSchema) {
+        allOfSchema = resolveSchema(allOfSchema, rootSchema, formData); // resolve references etc.
+        resolvedSchema = mergeSchemas(resolvedSchema, allOfSchema);
+        delete resolvedSchema.allOf;
+      }
+    }
+  }
+
   return resolvedSchema;
 }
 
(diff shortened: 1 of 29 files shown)


Saulzi:
@epicfaace this is a rebased #1610 to review

cypressious:
What's the status of this?

epicfaace:
I still need to review this.

packages/core/test/allOf_test.js
```diff
@@ -14,6 +14,7 @@ describe("allOf", () => {
   });
 
   it("should render a regular input element with a single type, when multiple types specified", () => {
+    // I think that this is wrong
```
epicfaace, line 17: What should be rendered instead?

packages/core/test/allOf_test.js
```diff
@@ -30,12 +31,44 @@ describe("allOf", () => {
...
+    expect(node.querySelectorAll("select")).to.have.length.of(1); // Schema 2
+  });
+
   it("should be able to handle incompatible types and not crash", () => {
     const schema = {
       type: "object",
       properties: {
         foo: {
-          allOf: [{ type: "string" }, { type: "boolean" }],
+          allOf: [{ type: "string" }, { type: "boolean" }], // this will basically pick up the last entry
```
epicfaace, line 71: What's the justification for picking up the last entry here? Isn't it better to show a validation error because technically the form data will never validate?

packages/core/test/allOf_test.js
```diff
@@ -30,12 +31,44 @@ describe("allOf", () => {
     expect(node.querySelectorAll("input")).to.have.length.of(1);
   });
 
+  it("should properly merge multiple schemas as per https://json-schema.org/understanding-json-schema/reference/combining.html", () => {
```
epicfaace, line 34:
```suggestion
  it("should properly merge multiple schemas with refs", () => {
```

packages/playground/src/samples/ifthenelse.js
```diff
@@ -0,0 +1,75 @@
+module.exports = {
```
epicfaace, line 1: I think ifthenelse.js is the same example as allofifthenelse.js. Did you mean to change this example?

packages/core/src/utils.js
```diff
@@ -655,28 +656,63 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
...
+      if ("if" in allOfSchema) {
+        allOfSchema = isValid(allOfSchema.if, formData)
+          ? allOfSchema.then
+          : allOfSchema.else;
+      }
+
+      if (allOfSchema) {
+        allOfSchema = resolveSchema(allOfSchema, rootSchema, formData); // resolve references etc.
+        resolvedSchema = mergeSchemas(resolvedSchema, allOfSchema);
+        delete resolvedSchema.allOf;
```
epicfaace, line 711: Is it ok to mutate here? or would it be safer to do something like

`let {allOf, ...resolvedSchema = mergeSchemas(resolvedSchema, allOfSchema);`

packages/core/src/validate.js
```diff
@@ -15,6 +15,7 @@ function createAjvInstance() {
     multipleOfPrecision: 8,
     schemaId: "auto",
     unknownFormats: "ignore",
+    useDefaults: true,
```
epicfaace, line 18: Why do this change? (it's a breaking change)

Saulzi: To be honest with you as this was quite some time ago, I cannot remember why I have a feeling that this was me looking into things as it was not quite behaving as I was expecting i.e. when no value specified all things where essentially valid so would give multiple options (perhaps)

epicfaace: Ok, let's try removing it and see if it still works.

DeadIine:
any news on this?

Saulzi:
@DeadIine 

Sorry, I am finding it extremely difficult to find time to make changes ATM because of current COVID-19 issue and juggling children etc.

I hope to get round to this soon as I would like to be using the official package rather than a fork myself , I have some other changes / fixes / issues which I have found which I will also be looking at when things get closer to normality.

Flamingo09:
Hey, is this feature support done?

packages/core/src/utils.js
```diff
@@ -655,28 +656,63 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
     return {};
   }
   let resolvedSchema = resolveSchema(schema, rootSchema, formData);
-  if ("allOf" in schema) {
-    try {
-      resolvedSchema = mergeAllOf({
```
erunion, line 660: What's the reasoning behind dropping `json-schema-merge-allof` in favor of a custom solution?

epicfaace: I don't think `json-schema-merge-allof` covers a lot of scenarios we need (such as allOf with if / then / else).

Also, the way `json-schema-merge-allof` gives errors is not great; it's hard for us to capture those errors and create react-jsonschema-form errors from them.

There's also this issue: https://github.com/rjsf-team/react-jsonschema-form/issues/1821

erunion: That's fair, I just wonder if maybe that work should be split out into either another library all together, or another PR. As it is, with our large deployment of RJSF across all of our customer sites this PR scares the hell out of me.

sar93:
Any news on this PR? I just need this functionality :(

packages/core/src/utils.js
epicfaace: Yeah, I don't think this PR is ready yet. It needs more work, first to minimize changes so that only the bare minimum needed for if/then/else remains, and then to verify functionality using tests.

epicfaace:
I'm going to label this a "breaking change" because there may be subtle differences between how allOf is handled between this approach and the json-schema-merge-allof approach (most notably, the way that json-schema-merge-allof handles errors is quite different -- I think it just logs to the console). It can go into v3.x, though.

packages/core/src/validate.js
raleng: > To be honest with you as this was quite some time ago, I cannot remember why I have a feeling that this was me looking into things as it was not quite behaving as I was expecting i.e. when no value specified all things where essentially valid so would give multiple options (perhaps)

The reason is that ajv returns `true` when validating against a constant when the value is undefined (see your comments [here](https://github.com/rjsf-team/react-jsonschema-form/pull/1610#issuecomment-592485064) and [here](https://github.com/rjsf-team/react-jsonschema-form/pull/1610#discussion_r387647808)).

norim13:
Any news on "if then else" support? Is it expected to be available as a stable version any time soon?

aidan-plenert-macdonald:
@epicfaace In your opinion, is it possible to support this without breaking changes? If you think this is fixable, I will fix it because I need this support.

epicfaace:
@aidan-plenert-macdonald I think merging this will necessarily require releasing a new breaking change, because we are switching away from json-schema-merge-allof and the behavior will most likely differ. However, I'm fine with cutting a new version 3.0 for this purpose -- so if you are able to work on this PR with the needed changes (minimize changes so that only the bare minimum needed for if/then/else remains, and then to verify functionality using tests) are done, we can get it merged. Thanks!

packages/core/src/utils.js
```diff
@@ -655,28 +656,63 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
...
-    }
-  }
+
   const hasAdditionalProperties =
     resolvedSchema.hasOwnProperty("additionalProperties") &&
     resolvedSchema.additionalProperties !== false;
+
   if (hasAdditionalProperties) {
-    return stubExistingAdditionalProperties(
+    resolvedSchema = stubExistingAdditionalProperties(
```
saswat3115, line 665: There is one issue here. If schema has additionProperties field then  having if-then its breaking. Becasue `stubExistingAdditionalProperties` thinks the added `then` schema is belongs to additionalProperties. Please check

epicfaace:
Closing in favor of https://github.com/rjsf-team/react-jsonschema-form/pull/2466. Thanks for your initial work @Saulzi and @nickgros for picking this PR up!

PR #1709, opened 2020-04-19
Fixed the Field templates' children from an array to a node by wrappi…

sramaswami11:
…ng it inside a div

[Please describe them here]

### Checklist

### DIFF
--- a/src/components/fields/SchemaField.js
+++ b/src/components/fields/SchemaField.js
@@ -352,50 +352,52 @@ function SchemaFieldRender(props) {
 
   return (
     <FieldTemplate {...fieldProps}>
-      {field}
+      <div>
+        {field}
 
-      {/*
+        {/*
         If the schema `anyOf` or 'oneOf' can be rendered as a select control, don't
         render the selection and let `StringField` component handle
         rendering
       */}
-      {schema.anyOf && !isSelect(schema) && (
-        <_AnyOfField
-          disabled={disabled}
-          errorSchema={errorSchema}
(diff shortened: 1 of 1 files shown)


epicfaace:
Can you use React.Fragment instead of a div? We don't want to change the HTML output of the page.

sramaswami11:
Done.

packages/core/test/FieldTemplate_test.js
```diff
@@ -14,6 +14,21 @@ describe("FieldTemplate", () => {
...
+        throw "Got wrong number of children";
+      }
+      return null;
+    }
+    createFormComponent({
+      schema: { type: "string" },
+      uiSchema: { "ui:disabled": true },
+      FieldTemplate,
+    });
+    expect();
```
epicfaace, line 29: Why do you have an empty `expect()` here?

sramaswami11: Removed.

packages/core/test/FieldTemplate_test.js
```diff
@@ -14,6 +14,20 @@ describe("FieldTemplate", () => {
     sandbox.restore();
   });
 
+  describe.only("FieldTemplate should only have one child", () => {
```
epicfaace, line 17:
```suggestion
  describe("FieldTemplate should only have one child", () => {
```

sramaswami11: Fixed.

PR #1714, opened 2020-04-20
Array buttons accessibility issues

dharizza:

There are some accessibility issues on the Move up/down/remove buttons on Array fields, it is failing also for the AddButton component.
image

According to standard WCAG AA:

```
This button element does not have a name available to an accessibility API. Valid names are: title undefined, element content, aria-label undefined, aria-labelledby undefined.
```

This is how it is being reported:
image

Here I'm adding a title in those places to comply with the standard.


dharizza:
Hi! Thanks for your feedback! Just pushed the changes.

epicfaace:
@dharizza I don't think there's been much of a focus on accessibility in this library so far -- if you're interested in setting up some kind of automated accessibility check for react-jsonschema-form (see https://web.dev/accessibility-auditing-react/), that would be really appreciated!

PR #1715, opened 2020-04-21
Utils.js - allow dependencies to use definitions

edparsons:
By adding definition to the validator definitions on the schema can be accessed inside dependency schemas.

At the moment if you try to reference definitions inside the oneOf part of a dependency it fails.

### Checklist


epicfaace:
@edparsons thanks for the PR! Can you add a test that uses a schema that caused an error before this change?

edparsons:
> @edparsons thanks for the PR! Can you add a test that uses a schema that caused an error before this change?

Hi @epicfaace I've added a test and without the fix the test fails with:
 1) utils
       retrieveSchema()
         schema dependencies
           conditional
             with $ref in oneOf
               should retrieve referenced properties from root schema definitions:
     Error: Could not find a definition for #/definitions/a.

ed-workbase:
Hey checking in on this PR, looks like it's all good with a test and just needs an approval

heath-freenome:
Closing this as we have removed `definitions` from the `@rjsf` version 5 release (see the `rjsf-v5` branch)

PR #1730, opened 2020-04-27
Fix #1314 event in onAddClick is no longer mandatory - with test

### LINKED ISSUE #1314: ArrayFieldTemplate - onAddClick should not require an argument
### Prerequisites

### Description

Right now calling this.props.onAddClick() from an ArrayFieldTemplate gives an error because it expects you to pass in an event as an argument. I don't see why passing in an event should be required so we should make it optional.

Parro:

Calling `onAddClick` from a custom `ArrayFieldTemplate` gives error if `event` is not passed as parameter, but there are several cases in which `event` is not set. This PR makes `event` optional fixing #1314. Code update taken from #1365 and #1324.

### Checklist


docs/advanced-customization.md
```diff
@@ -120,7 +120,7 @@ The following props are passed to each `ArrayFieldTemplate`:
 - `disabled`: A boolean value stating if the array is disabled.
 - `idSchema`: Object
 - `items`: An array of objects representing the items in the array. Each of the items represent a child with properties described below.
-- `onAddClick: (event) => void`: A function that adds a new item to the array.
+- `onAddClick: ([event]) => void`: A function that adds a new item to the array. `event` is optional.
```
epicfaace, line 123:
```suggestion
- `onAddClick: (event?) => void`: A function that adds a new item to the array. `event` is optional.
```

packages/core/test/ArrayField_test.js
```diff
@@ -249,6 +273,18 @@ describe("ArrayField", () => {
...
 
+    it("should add a field when clicking add button even if event is not passed to onAddClick", () => {
+      const { node } = createFormComponent({
+        schema,
+        ArrayFieldTemplate: CustomOnAddClickTemplate,
+      });
+
+      Simulate.click(node.querySelector(".array-item-add button"));
+
+      expect(node.querySelector(".array-item").hasAttribute(ArrayKeyDataAttr))
```
epicfaace, line 284: Can you instead make the test check just that `.array-item` exists? I don't think we need to check if the key is present, so we should safely be able to remove that check and the data-rjsf-itemkey attribute (which was only a hack needed to check the keys of array items for specific tests regarding array reordering)

Parro: I fixed the test but I've never used chai... Did i used the right syntax?

Parro:
> Thanks! Can you also update the types here: https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/index.d.ts#L184

Updated, I also removed the second event params, there is no currying for this function 😉

packages/core/test/ArrayField_test.js
```diff
@@ -62,6 +62,30 @@ const ExposedArrayKeyTemplate = function(props) {
...
 
+const CustomOnAddClickTemplate = function(props) {
+  return (
+    <div className="array">
+      {props.items &&
+        props.items.map(element => (
+          <div
+            key={element.key}
+            className="array-item"
+            data-rjsf-itemkey={element.key}>
```
epicfaace, line 73: Can you remove this attribute? then we should be good

Parro: Fixed

Issue #1209, opened 2019-03-06
Use react-is for widget type check


anlesk:
### Prerequisites

- [x] I have read the [documentation](https://react-jsonschema-form.readthedocs.io/);

### Description

Starting from **React 16.3.0** the component is not always a function, but could be an object if it was wrapped with `React.forwardRef`.

According to the guidelines, `react-is` should be used for type validation.
The proper way of doing that is simply call - `ReactIs.isValidElementType(Widget);`

The possible place to fix: [utils.js#L96](https://github.com/mozilla-services/react-jsonschema-form/blob/f9d4c63cac24146e522528d33eb04f465464e052/src/utils.js#L96)

### Steps to Reproduce

1. Define `DescriptionField` and wrap it with `React.forwardRef`
```
const DescriptionField = React.forwardRef((props, ref) => (
    <YourComponent {...props} ref={ref} />)
)
```
2. Define custom fields
```
const fields = { DescriptionField }
```
2. Pass fields to SchemaForm
```
<SchemaForm
   ...
   fields={fields}
   ...
/>
```
3. Observe in the console

> Uncaught Error: Unsupported widget definition: object

#### Expected behavior

Validation successfully passed.

#### Actual behavior

Validation is not passed, error is thrown.

### Version

    "react-jsonschema-form": "^1.0.3"

epicfaace:
@anlesk sounds good, can you make a PR?

anlesk:
Well, the proper fix would require dependencies to be switched to react>=16.3.

Until that happen - the proposed solution will cover ForwardedRef needs.

epicfaace:
@anlesk Sorry, I haven't really used React.ForwardedRef before. Do you mind linking to a jsfiddle in which you try to use React.ForwardedRef with react-jsonschema-form, so I can see what error is coming up and what exactly PR #1227 is fixing? Would appreciate it!

epicfaace:
@anlesk I've tried reproducing this in a jsfiddle but I don't get any issue: https://jsfiddle.net/f0dxpot9/3/

Am I missing something in the code?

anlesk:
@epicfaace aw, yeah. Seems that I've provided an incorrect input. This is only reproducible with widgets.

Check this one: https://jsfiddle.net/a4rqpj1d/9/

lyleunderwood:
[...]

PR #1737, opened 2020-04-30
uses ReactIs.isValidElementType to check widget types

jonathanhawleypeters:

Since React 16.3.0 components created by `React.forwardRef` may be an object. According to the guidelines, react-is should be used for type validation.

https://github.com/rjsf-team/react-jsonschema-form/issues/1209

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -108,11 +108,7 @@ export function getWidget(schema, widget, registeredWidgets = {}) {
     return Widget.MergedWidget;
   }
 
-  if (
-    typeof widget === "function" ||
-    ReactIs.isForwardRef(React.createElement(widget)) ||
-    ReactIs.isMemo(widget)
-  ) {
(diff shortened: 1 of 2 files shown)


packages/core/src/utils.js
```diff
@@ -108,11 +108,7 @@ export function getWidget(schema, widget, registeredWidgets = {}) {
     return Widget.MergedWidget;
   }
 
-  if (
-    typeof widget === "function" ||
-    ReactIs.isForwardRef(React.createElement(widget)) ||
-    ReactIs.isMemo(widget)
-  ) {
+  if (typeof widget !== "string" && ReactIs.isValidElementType(widget)) {
     return mergeOptions(widget);
```
epicfaace, line 112: Since our old tests didn't catch this issue, would it be possible to update them so that we don't regress again?

jonathanhawleypeters: Yes, I'll do that soon. Thanks for the reminder

PR #1401, opened 2019-08-12
Key changes on additionalProperties items no longer suffer data loss.

erunion:
### Reasons for making this change

If there is an `additionalProperties` object present that has a `title` contained within it, the input form data for that title is overridden whenever you click out of the input, resulting in user-entered data being lost.

This resolves https://github.com/mozilla-services/react-jsonschema-form/issues/1382, and here is a local dev playground link to see the fix in action:

http://localhost:8080/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0aXRsZSI6IkEgY3VzdG9taXphYmxlIHJlZ2lzdHJhdGlvbiBmb3JtIiwiZGVzY3JpcHRpb24iOiJBIHNpbXBsZSBmb3JtIHdpdGggYWRkaXRpb25hbCBwcm9wZXJ0aWVzIGV4YW1wbGUuIiwidHlwZSI6Im9iamVjdCIsImFkZGl0aW9uYWxQcm9wZXJ0aWVzIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7ImNvZGUiOnsidHlwZSI6ImludGVnZXIifSwidGV4dCI6eyJ0eXBlIjoic3RyaW5nIn19LCJ0aXRsZSI6IlRoaXMgaXMgYSBjdXN0b20gdGl0bGUifX0sInVpU2NoZW1hIjp7fX0=

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #1751, opened 2020-05-13
additionalProperties: retain object title when renaming key

### LINKED ISSUE #1749: When additionalProperties key names are changed, object title disappears
### Description

When additionalProperties are added with omitExtraData and liveOmit, object title disappears

### Steps to Reproduce

1. Go to [this playground link](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJmaXJzdE5hbWUiOiJDaHVjayIsImxhc3ROYW1lIjoiTm9ycmlzIiwiYXNkYXNkIjoiaW5maW5pdHkifSwic2NoZW1hIjp7InRpdGxlIjoiQSBjdXN0b21pemFibGUgcmVnaXN0cmF0aW9uIGZvcm0iLCJkZXNjcmlwdGlvbiI6IkEgc2ltcGxlIGZvcm0gd2l0aCBhZGRpdGlvbmFsIHByb3BlcnRpZXMgZXhhbXBsZS4iLCJ0eXBlIjoib2JqZWN0IiwicmVxdWlyZWQiOlsiZmlyc3ROYW1lIiwibGFzdE5hbWUiXSwiYWRkaXRpb25hbFByb3BlcnRpZXMiOnsidHlwZSI6InN0cmluZyJ9LCJwcm9wZXJ0aWVzIjp7ImZpcnN0TmFtZSI6eyJ0eXBlIjoic3RyaW5nIiwidGl0bGUiOiJGaXJzdCBuYW1lIn0sImxhc3ROYW1lIjp7InR5cGUiOiJzdHJpbmciLCJ0aXRsZSI6Ikxhc3QgbmFtZSJ9fX0sInVpU2NoZW1hIjp7ImZpcnN0TmFtZSI6eyJ1aTphdXRvZm9jdXMiOnRydWUsInVpOmVtcHR5VmFsdWUiOiIifX0sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6dHJ1ZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==)
1. Form looks like this
image
1. Type something in the additionalProperties key field, and focus outside the field
1. Title disappears
image

### Version

v2.0.0

epicfaace:

Fixes #1749. This reverts part of #1401 (the conditional setting of the title in the `wasPropertyKeyModified` logic in ObjectField.js) -- this piece of logic doesn't seem to contribute to the fix in #1401, and it's what caused #1749.

Additionally, the test that was added in #1401 didn't actually check to make sure the issue it was solving was resolved; that test passed even before #1401 was merged. The issue that #1401 was that `onChange` was called with the new key, but the new key did not show up in the DOM. I've updated that test to check for this behavior (updating the new key in the DOM) now.

@erunion can you review to make sure this all makes sense?


packages/core/src/components/fields/ObjectField.js
```diff
@@ -214,14 +214,7 @@ class ObjectField extends Component {
...
 
-    // If this schema has a title defined, but the user has set a new key/label, retain their input.
-    let title;
-    if (this.state.wasPropertyKeyModified) {
-      title = name;
-    } else {
-      title = schema.title === undefined ? name : schema.title;
-    }
-
+    const title = schema.title === undefined ? name : schema.title;
```
stanlemon, line 217: Not specific to these changes... You might want to make this falsey rather than specifically checking for undefined, in case a schema.title is set as false or null. I admit, that would be weird, but I think you'd still want to default to name there.

epicfaace: Hmm, not quite sure about that. Sometimes it's useful for me to set schema.title = "" in order to show _no title_ (I know there's also a uiSchema option for hiding titles, but perhaps it would make sense if the schema title is set to a falsey value, just to not show a title in that case)?

PR #1758, opened 2020-05-17
Set up playground visual regression tests with puppeteer and jest-image-snapshot

epicfaace:

Set up playground visual regression tests with puppeteer and jest-image-snapshot. This will help us catch regressions more easily.


epicfaace:
This diff is unexpected (note how the styling of the dropdowns change?)

index-test-js-test-default-schema-dependencies-1-diff

epicfaace:
> This diff is unexpected (note how the styling of the dropdowns change?)
> 
> index-test-js-test-default-schema-dependencies-1-diff

@stanlemon I was trying to set this up, as it might help for maintaining things like browser compatibility and quickly spotting regressions -- but I'm facing diffs like the above in which the images are off by a couple of pixels. I feel like I might be doing something slightly wrong here. Have you used these tools (jest + puppeteer + image snapshots) before -- any idea why this might be happening?

stanlemon:
@epicfaace I've used puppeteer quite a bit, but never with image snapshots. Looking at your screenshot it almost looks like the font-face is different between the two. Is that possible?

PR #1759, opened 2020-05-18
Make event optional for onDropIndexClick and onReorderClick, add onAd…

epicfaace:
…dIndexClick type definition

Better types. The event parameter is actually optional for all these methods.


epicfaace:
> Seems reasonable. Did this crop up in a theme or as part of custom templates?

I think I was just upgrading a project from rjsf 1 -> 2 (which includes types), and got errors because some of these methods have optional arguments.

PR #1760, opened 2020-05-18
Use React.Component instead of React.StatelessComponent type

epicfaace:

These props don't have to be stateless (functional) components; they can also be class-based components. React.Component seems more general.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -12,7 +12,7 @@ declare module '@rjsf/core' {
         acceptcharset?: string;
         action?: string;
         additionalMetaSchemas?: ReadonlyArray<object>;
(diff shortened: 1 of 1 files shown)


packages/core/index.d.ts
```diff
@@ -12,17 +12,17 @@ declare module '@rjsf/core' {
         acceptcharset?: string;
         action?: string;
         additionalMetaSchemas?: ReadonlyArray<object>;
-        ArrayFieldTemplate?: React.StatelessComponent<ArrayFieldTemplateProps>;
+        ArrayFieldTemplate?: React.Component<ArrayFieldTemplateProps>;
```
stanlemon, line 15: Maybe React.ComponentType would be better?

Issue #941, opened 2018-06-01
Enhancement: Allow rawErrors to be an Array of PropTypes.any


doncesarts:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description
I am facing the following scenario: 
I need to transform and translate error messages, this is done trough the function `transformErrors` [custom-error-messages](https://github.com/mozilla-services/react-jsonschema-form/blob/66ab17a72b9dec7fd2bc7bd21785f27557cc0b49/README.md#custom-error-messages)

1. To translate the message I am using  [react-intl](https://github.com/yahoo/react-intl).
2. The function `transformErrors` can not be wrapped with `react-intl` , so `transformErrors` returns an array of translatable messages (objects) , instead of strings. 
3. Error messages are translated in a custom `Error` component. 
3. There are errors in the console complaining about the type. 
4. This errors could be avoid if rawErrors  hare ` Array of PropTypes.any`

The documentation for [rawErrors](https://github.com/mozilla-services/react-jsonschema-form/blob/66ab17a72b9dec7fd2bc7bd21785f27557cc0b49/README.md#field-template) says as following:

> If you want to handle the rendering of each element yourself, you can use the props rawHelp, rawDescription and rawErrors.

Defining `rawErrors` to only accept strings limits the full potential of customisations.  I will be willing to do a pull request with this changes. 

#### Expected behavior
No warnings in the console. 

`   
 rawErrors: PropTypes.arrayOf(PropTypes. any),
`

#### Actual behavior

```
    rawErrors: PropTypes.arrayOf(PropTypes.string),

Warning: Failed prop type: Invalid prop `rawErrors[0]` of type `object` supplied to `StringField`, expected `string`.
```

epicfaace:
@doncesarts yes, makes sense, please feel free to make a PR with these changes

PR #1776, opened 2020-05-28
Enhancement: Allow rawErrors to be an Array of PropTypes.any #941

doncesarts:
On branch issue/941
 Changes to be committed:

- 	Updates documentation:   docs/advanced-customization/custom-templates.md
- 	Updates documentation:   docs/advanced-customization/custom-widgets-fields.md
- 	Updates prop types of errors:   packages/core/src/components/fields/SchemaField.js
- 	Updates prop types of errors:   packages/core/src/types.js
- 	Update a test using error message objects:  packages/core/test/validate_test.js

This PR related #941

### Checklist

### DIFF
--- a/packages/core/test/validate_test.js
+++ b/packages/core/test/validate_test.js
@@ -354,8 +354,19 @@ describe("Validation", () => {
       },
     };
     const newErrorMessage = "Better error message";
+    const newObjectErrorMessage = {
+      message: "Better error message with params",
+      params: {
+        type: "string",
+        property: ".foo",
+      },
+    };
+
     const transformErrors = errors => {
-      return [Object.assign({}, errors[0], { message: newErrorMessage })];
+      return [
+        Object.assign({}, errors[0], { message: newErrorMessage }),
+        { message: newObjectErrorMessage },
+      ];
     };
 
     let errors;
@@ -373,6 +384,11 @@ describe("Validation", () => {
     it("should use transformErrors function", () => {
       expect(errors).not.to.be.empty;
       expect(errors[0].message).to.equal(newErrorMessage);
+    });
+
+    it("should allow object error messages", () => {
+      expect(errors).not.to.be.empty;
+      expect(errors[1].message).to.equal(newObjectErrorMessage);
     });
   });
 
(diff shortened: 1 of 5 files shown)


doncesarts:
Hi @epicfaace ,  within `react-jsonschema-form` there is not an example of this. 
But as `react-jsonschema-form` exposes `transformErrors` [custom-error-messages](https://github.com/mozilla-services/react-jsonschema-form/blob/66ab17a72b9dec7fd2bc7bd21785f27557cc0b49/README.md#custom-error-messages) ,  then it is possible to return objects.  
The use case that I am facing is: 

1. To translate the message I am using  [react-intl](https://github.com/yahoo/react-intl).
2. The function `transformErrors` can not be wrapped with `react-intl` , so `transformErrors` returns an array of translatable messages (objects) , instead of strings. 
3. Error messages are translated in a custom `Error` component. 
3. There are errors in the console complaining about the type. 
4. This errors could be avoid if rawErrors  hare ` Array of PropTypes.any`
5. Then at render time, the object message is translated in any given language.

packages/core/test/validate_test.js
```diff
@@ -374,6 +385,11 @@ describe("Validation", () => {
       expect(errors).not.to.be.empty;
       expect(errors[0].message).to.equal(newErrorMessage);
     });
+
+    it("should allow object error messages", () => {
+      expect(errors).not.to.be.empty;
```
epicfaace, line 390: Would this test fail if the proptype was set to a string? Or would it merely emit a warning to the console?

doncesarts: None of those, after reviewing it , this test is not really enforcing the propTypes validation. I will update the test cases. It should be focus on SchemaField or any other field, as it is there where the propTypes validation occurs.

heath-freenome:
Now that RJSF is fully typescript, this PR no longer makes sense. One can use Typescript's [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) to modify the types within  ones project to avoid typing issues.

PR #1787, opened 2020-06-02
Add registry to custom FieldTemplate

stanlemon:

Certain customizations one might want to do in a FieldTemplate may require access to things that are only contained into the registry, such as the rootSchema. Note that this is already done in the ObjectFieldTemplate and so this merely brings the properties that are passed down into alignment with that template.

### Checklist

### DIFF
--- a/docs/advanced-customization/custom-templates.md
+++ b/docs/advanced-customization/custom-templates.md
@@ -145,6 +145,7 @@ The following props are passed to a custom field template component:
 - `schema`: The schema object for this field.
 - `uiSchema`: The uiSchema object for this field.
 - `formContext`: The `formContext` object that you passed to Form.
+- `registry`: The `registry` object.
 
 > Note: you can only define a single global field template for a form, but you can set individual field templates per property using `"ui:FieldTemplate"`.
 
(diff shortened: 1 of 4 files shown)


stanlemon:
@epicfaace I think I did that correctly, please let me know if I did not. I noticed that `ObjectFieldTemplate` does not include `registry` in the docs or type definition, though it does receive it. I'm happy to modify both of those, but wanted to check if you'd prefer that be in a separate PR?

epicfaace:
> @epicfaace I think I did that correctly, please let me know if I did not. I noticed that `ObjectFieldTemplate` does not include `registry` in the docs or type definition, though it does receive it. I'm happy to modify both of those, but wanted to check if you'd prefer that be in a separate PR?

Yeah, feel free to do it in this PR too. Thanks!

stanlemon:
Updated `ObjectFieldTemplate` props and updated `ObjectFieldTemplate` and `ArrayFieldTemplate` docs. Re-verified docs looked correct and re-ran tests, all looked 👍 .

stanlemon:
Thank you!

PR #1789, opened 2020-06-02
Add missing material theme string widgets

stanlemon:

Several string widgets are missing from the material theme, specifically color, date, date time, email and URI. Without these widgets using the various formats for fields with type 'string' will fall back to the bootstrap theme. This PR adds support for all of those widgets in material UI with the same reuse pattern as in the bootstrap theme (pass back the `TextWidget` with a more specific type, see https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/widgets/ColorWidget.js for an example).

_Note that I needed to add 'type' to the `WidgetProps` to support this. The bootstrap widgets do not have their props typed so I believe this is why that wasn't an issue before._

Here is an example schema that you can use in the playground to see what happens when you use the material theme.

```json
{
  "title": "String Widgets",
  "description": "String Widgets",
  "type": "object",
  "properties": {
    "date": {
      "type": "string",
      "title": "Date Widget",
      "format": "date"
    },
    "dateTime": {
      "type": "string",
      "title": "Date Time Widget",
      "format": "date-time"
    },
    "email": {
      "type": "string",
      "title": "Email Widget",
      "format": "email"
    },
    "uri": {
      "type": "string",
      "title": "URI Widget",
      "format": "uri"
    },
    "color": {
      "type": "string",
      "title": "Color Widget",
      "format": "color"
    }
  }
}

```

There is a similar PR just for the date widget in #1765 

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1048,6 +1048,36 @@ export function toDateString(
   return time ? datetime : datetime.slice(0, 10);
 }
 
+export function utcToLocal(jsonDate) {
+  if (!jsonDate) {
+    return "";
+  }
+
+  // required format of `"yyyy-MM-ddThh:mm" followed by optional ":ss" or ":ss.SSS"
+  // https://html.spec.whatwg.org/multipage/input.html#local-date-and-time-state-(type%3Ddatetime-local)
+  // > should be a _valid local date and time string_ (not GMT)
+
+  // Note - date constructor passed local ISO-8601 does not correctly
+  // change time to UTC in node pre-8
+  const date = new Date(jsonDate);
+
+  const yyyy = pad(date.getFullYear(), 4);
+  const MM = pad(date.getMonth() + 1, 2);
+  const dd = pad(date.getDate(), 2);
+  const hh = pad(date.getHours(), 2);
+  const mm = pad(date.getMinutes(), 2);
+  const ss = pad(date.getSeconds(), 2);
+  const SSS = pad(date.getMilliseconds(), 3);
+
+  return `${yyyy}-${MM}-${dd}T${hh}:${mm}:${ss}.${SSS}`;
+}
+
+export function localToUTC(dateString) {
+  if (dateString) {
+    return new Date(dateString).toJSON();
+  }
+}
+
 export function pad(num, size) {
   let s = String(num);
   while (s.length < size) {
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -390,6 +390,10 @@ declare module '@rjsf/core' {
 
         export function toDateString(dateObject: DateObject, time?: boolean): string;
 
+        export function utcToLocal(jsonDate: string): string;
+
+        export function localToUTC(dateString: string): Date;
+
         export function pad(num: number, size: number): string;
 
         export function setState(instance: React.Component, state: any, callback: Function): void;
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -109,6 +109,7 @@ declare module '@rjsf/core' {
         onBlur: (id: string, value: boolean | number | string | null) => void;
         onFocus: (id: string, value: boolean | number | string | null) => void;
         label: string;
+        type: string;
         multiple: boolean;
         rawErrors: string[];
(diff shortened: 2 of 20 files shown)


stanlemon:
I don't think this requires any documentation changes since this is just filling in gaps on the material theme, but I am happy to add anything if desired. As far as tests, I wasn't entirely sure what was best to add here. There are existing tests in `StringField_test.js` but they are not material-specific. Any guidance there would be appreciated.

stanlemon:
Snapshots were updated for email and url, I will add some schema for the other types.

I just noticed that my prettier extension in vscode reformatted some of the stuff I touched. I didn't find a prettier config, I did notice this comment https://react-jsonschema-form.readthedocs.io/en/latest/contributing/#coding-style and just wanted to make sure the defaults are being used and some of the formatting stuff included here is OK or if I should back it out.

stanlemon:
Tests added and snapshots have been updated @epicfaace.

stanlemon:
Interesting, looks like that happens in Chrome but not Safari. It appears to be a bug with material-ui. From what I gather I need to set this to make the empty render not look ugly:

```
InputLabelProps={{
  shrink: true
}}
```

On the `TextWidget` for that one. That will mean that label will not bounce up above the input upon focus, which is the current default behavior on rjsf. If you're OK with that I'll work that in.

Thinking out loud here... We could use these pickers https://material-ui.com/components/pickers/ for the date and time widgets, or we could use them in the alt-date alt-datetime widgets too. Any interest in either of those options?

stanlemon:
@epicfaace here's a set of changes that would allow for passing the shrink settings. https://github.com/stanlemon/react-jsonschema-form/pull/1/files

It also allows for further properties to be passed down to the underlying `TextWidget` which could come in handy. Let me know if this seems agreeable and I'll merge it into this branch.

epicfaace:
> Thinking out loud here... We could use these pickers https://material-ui.com/components/pickers/ for the date and time widgets, or we could use them in the alt-date alt-datetime widgets too. Any interest in either of those options?

Yeah, those would probably be ideal. We would need to add a peer dependency on `@material-ui/pickers`, though, so let's not do that in this PR. Let's save that for the next major version.

stanlemon:
> Yeah, those would probably be ideal. We would need to add a peer dependency on `@material-ui/pickers`, though, so let's not do that in this PR. Let's save that for the next major version.

Just to clarify: They would be ideal for the main widgets or the alt widgets?

epicfaace:
> > Yeah, those would probably be ideal. We would need to add a peer dependency on `@material-ui/pickers`, though, so let's not do that in this PR. Let's save that for the next major version.
> 
> Just to clarify: They would be ideal for the main widgets or the alt widgets?

Probably for the main widgets.

epicfaace:
> This will mean that label will not bounce up above the input upon focus, which is the current default behavior on rjsf. If you're OK with that I'll work that in.

Sounds good to me!

stanlemon:
Good deal. I merged in the change to support the shrink param, so that issue should be resolved.

stanlemon:
Clean build, this one might be ready. 🤞 If you have any guidance on that last commit with regards to the typing let me know. I think it is OK, but my typescript-foo is limited so I'm not entirely confident.

epicfaace:
@stanlemon can you also change the format in which the date widget gives its value?

Right now, the format doesn't seem to be valid for a "date-time" format in JSON Schema:

image


Compare this with the default date widget value:

image

stanlemon:
@epicfaace good catch, 81b20350329a2197a3ae1c607a5026a3cea88d25 should resolve that. Thanks.

packages/core/src/utils.js
```diff
@@ -1048,6 +1048,36 @@ export function toDateString(
   return time ? datetime : datetime.slice(0, 10);
 }
 
+export function utcToLocal(jsonDate) {
```
epicfaace, line 1051: What exactly is the point of this function?

stanlemon: This is what ensures the format is correct in the Bootstrap theme. I didn’t write this, just simply relocated it to utils.

PR #1790, opened 2020-06-03
Better consistency with Material UI patterns with usage of <TextField />

stanlemon:
**tldr; Use `<TextField />` as much as possible.**

This PR brings better consistency with Material UI patterns. It attempts to use `< TextField />` from Material UI whenever possible. There is an extra `<FormControl />` component in the current component tree and this fixes that (see images for before/after of `<TextWidget />`). It also moves the `required` property up to the main `<FieldTemplate />` and then sets it on the specific `<TextField />` components as well. This change makes the error behavior more consistent with the way `<TextField />` usually works out of the box. I also made minor adjustments to the `<List />` and `<ListItem />` used for errors to make the UI more consistent with what you see at https://material-ui.com/components/text-fields/#validation. Finally it adjusts widgets like `<SelectWidget />` to use `<TextField />` rather than reconstruct that component.

The extra `<FormControl />` node **BEFORE**:

MuiFormControl-root-OLD

...and **AFTER**:

MuiFormControl-root-NEW

The error highlighting **BEFORE**:

Screen Shot 2020-06-02 at 9 14 36 PM

...and **AFTER**

Screen Shot 2020-06-02 at 9 14 21 PM

Errors also show up on radio and range widgets properly:

Screen Shot 2020-06-08 at 8 00 37 PM
Screen Shot 2020-06-08 at 8 00 57 PM

To make rjsf behave more like the Material UI examples, and also to try and reduce unnecessary form construction by leveraging `TextField` as much as possible.

### Checklist


stanlemon:
@epicfaace `<TextField />` uses `<FormControl />` under the hood. `<FormControl />` is a lower level construct. In Material UI `<TextField />` is the go to unless you need to do specific custom work to construct your form. It covers a multitude of properties and behaviors and most of the time _just works_.

jdbence:
@stanlemon works great. The move over to ```<TextField/>``` was a good choice. Removing that extra ```<FormControl/>``` makes working with the Grid less painful.

packages/material-ui/test/Form.test.tsx
```diff
@@ -106,4 +105,96 @@ describe("single fields", () => {
...
+      "ui:widget": "radio",
+    };
+    const tree = renderer
+      .create(<Form schema={schema} uiSchema={uiSchema} />)
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+  // There is a bug in the Material UI <Slider /> component that prevents this from working.
+  // Error: `TypeError: Cannot read property 'addEventListener' of null`
+  // From: https://github.com/mui-org/material-ui/blob/v4.5.2/packages/material-ui/src/Slider/Slider.js#L622
```
stanlemon, line 185: Note that this does not appear to be a new bug in this branch as I am seeing the same thing if I copy this test over to master.

stanlemon:
@epicfaace Just wanted to check in on this one and see if there was anything I needed to do or clarify. Thanks!

epicfaace:
@stanlemon it looks great! The reason I'm still thinking about whether / when to merge this is that I'm wondering whether this should be considered a breaking change (and require a version bump to v3). What do you think?

stanlemon:
> @stanlemon it looks great! The reason I'm still thinking about when to merge this is that I'm wondering whether this should be considered a breaking change (and require a version bump to v3). What do you think?

I completely understand the concern around breaking changes. For the most part this PR just addresses consistency issues with how Material UI is used. I suppose the changing of where the error state is could be considered significant, but in my opinion the current behavior is perhaps a bug. The PR also fixes some bugs from props passing down (like#1769), but outside of that it should not functionally change anything. Even the changes to `<SelectWidget />` shouldn't be functional, because `<TextField />` actually uses the `<SelectWidget />` under the hood. I think the biggest compatibility break would be if someone had tests that wrote selectors expecting to form controls (eg. `.MuiFormControl-root .MuiFormControl-root`), which seems unlikely to me. 

Anyhow, I totally respect whatever you decide here. I'm happy to back out any piece(s) if that would help.  But I'd also advocate for this to not be considered a BC break on the Material theme.

packages/material-ui/src/FieldTemplate/FieldTemplate.tsx
```diff
@@ -1,35 +1,39 @@
...
+      required={required}>
       {children}
       {displayLabel && rawDescription ? (
         <Typography variant="caption" color="textSecondary">
           {rawDescription}
         </Typography>
       ) : null}
       {rawErrors.length > 0 && (
-        <List dense={true}>
+        <List dense={true} disablePadding={true}>
```
epicfaace, line 32: Why set disablePadding and disableGutters to true?

packages/material-ui/src/PasswordWidget/PasswordWidget.tsx
```diff
@@ -19,35 +17,32 @@ const PasswordWidget = ({
...
-        value={value ? value : ''}
-        onFocus={_onFocus}
-        onBlur={_onBlur}
-        onChange={_onChange}
-      />
-    </FormControl>
+      disabled={disabled || readonly}
+      type="password"
+      value={value ? value : ""}
+      error={rawErrors.length > 0 ? true : false}
```
epicfaace, line 41:
```suggestion
      error={rawErrors.length > 0}
```

epicfaace:
@stanlemon yes, that makes sense, thanks for explaining! It's good to consider this as a "bug fix" rather than functionally changing anything. If we do add something like ui:props (https://github.com/rjsf-team/react-jsonschema-form/pull/1672) though, that would probably provide better reason for changing the widget to be considered a breaking change.

packages/material-ui/src/FieldTemplate/FieldTemplate.tsx
stanlemon: I was trying to visually match the visual behavior here: https://material-ui.com/components/text-fields/#validation

With variants 'filled' and 'outlined' there's an indentation on the error display, but otherwise it's anchored to the edge with the input.

Issue #1786, opened 2020-06-01
Material UI theme problem with initial 0 number


DennisKuhn:
Kia Ora :-)
thanks for the awesome library :-)

### Description
If a number field is loaded, e.g. formData, with a 0, it is displayed blank when using Material UI. When using the standard theme the field is populated with 0.

### Steps to Reproduce
_schema.json:_
`{ "type": "number" }`

_formData.json:_
`0`

_uiSchema.json:_
`{}`

#### Expected behavior
populated with 0: [Playground default-theme](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6MCwic2NoZW1hIjp7InR5cGUiOiJudW1iZXIifSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsidmFsaWRhdGUiOnRydWUsImRpc2FibGUiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlfX0=)

#### Actual behavior
populated with blank/undefined [Playground material-ui-theme](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6MCwic2NoZW1hIjp7InR5cGUiOiJudW1iZXIifSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6Im1hdGVyaWFsLXVpIiwibGl2ZVNldHRpbmdzIjp7InZhbGlkYXRlIjp0cnVlLCJkaXNhYmxlIjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZX19)

## Version
Playground as of 2020-06-01
@rjsf/core@2.0.1
@rjsf/material-ui@2.0.1

@material-ui/core@4.9.10
react@16.13.1

Anything I can do?
Dennis

cstim:
Dear authors, the "initial 0 number" problem was fixed recently in #1795 except that the fix there didn't catch the UpDownWidget. That one still tests for truthy here: https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/material-ui/src/UpDownWidget/UpDownWidget.tsx#L43 and that line needs a similar change into `value={value || (value === 0) ? value : ''}` 
Could you fix that one, too, please? Thanks a lot!

PR #1795, opened 2020-06-03
Test value for null or undefined

DennisKuhn:

Previous test caused 0 to result in a blank field

[Please describe them here]

https://github.com/rjsf-team/react-jsonschema-form/issues/1786

### DIFF
--- a/packages/material-ui/src/TextWidget/TextWidget.tsx
+++ b/packages/material-ui/src/TextWidget/TextWidget.tsx
@@ -43,7 +43,7 @@ const TextWidget = ({
         disabled={disabled || readonly}
         name={name}
         type={schema.type as string}
-        value={value === undefined || value === null  ? '' : value}
+        value={value || (value === 0) ? value : ''}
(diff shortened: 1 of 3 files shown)


packages/material-ui/src/TextWidget/TextWidget.tsx
```diff
@@ -43,7 +43,7 @@ const TextWidget = ({
         disabled={disabled || readonly}
         name={name}
         type={schema.type as string}
-        value={value ? value : ''}
+        value={value === undefined || value === null  ? '' : value}
```
epicfaace, line 46: I believe we would also need to include NaN. Can we do this instead (because I believe the only falsy value we need to include is 0)?

```suggestion
        value={value || (value === 0)  ? '' : value}
```

DennisKuhn:
Not sure if I've done it right:
- added a test for the 0 case, committed test file and snap shot file. 
- updated the code to `value || (value === 0) ? value : ''`

PR #1798, opened 2020-06-04
Focus on first error field while submitting form (if any validation errors)

saswat3115:

[Please describe them here]

Focus on first error field while submitting form (if any validation erros)

### Checklist

### DIFF
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -391,11 +391,11 @@ export default class Form extends Component {
     const firstErrorField = error?.property?.replace(".", "");
     let { idPrefix } = this.props;
     if (idPrefix === undefined) {
-      idPrefix = "root_";
+      idPrefix = "root";
     }
 
     const firstErrorElement = this.formElement.elements[
-      `${idPrefix}${firstErrorField}`
+      `${idPrefix}_${firstErrorField}`
     ];
     return firstErrorElement;
   };
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -335,7 +335,7 @@ export default class Form extends Component {
         const firstErrorElement = this.getFirstErrorField(
           schemaValidation.errors[0]
         );
-        firstErrorElement.focus(); // Focus on first error field
+        firstErrorElement && firstErrorElement.focus(); // Focus on first error field
 
         if (this.props.extraErrors) {
           errorSchema = mergeObjects(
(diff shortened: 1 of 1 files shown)


saswat3115:
@epicfaace 

> Thanks, can you add a test?

ok

packages/core/src/components/Form.js
```diff
@@ -332,6 +332,11 @@ export default class Form extends Component {
       const schemaValidationErrors = errors;
       const schemaValidationErrorSchema = errorSchema;
       if (Object.keys(errors).length > 0) {
+        const firstErrorElement = this.getFirstErrorField(
+          schemaValidation.errors[0]
+        );
+        firstErrorElement && firstErrorElement.focus(); // Focus on first error field
```
mattcosta7, line 338: firstErrorElement here could be either an Element _or_ a RadioNodeList

this would cover both use cases, focusing the element if it was one, OR the first item in the radio node list if it wasn't

```
        if (firstErrorElement instanceof Element) {
          firstErrorElement.focus();
        } else if (firstErrorElement instanceof RadioNodeList) {
          const firstItem = firstErrorElement.item(0);
          if (firstItem instanceof Element) {
            firstItem.focus();
          }
        }
```

packages/core/src/components/Form.js
```diff
@@ -382,6 +387,19 @@ export default class Form extends Component {
     );
   };
 
+  getFirstErrorField = error => {
+    const firstErrorField = error?.property?.replace(".", "");
```
mattcosta7, line 391: this only works for single level object fields

```
error?.property?.replace(".", "")
          .replace(/\./g, "_")
          .replace(/\[/g, "_")
          .replace(/\]/g, "")
```

I think covers nested fields and array items.  There might be a better way to get the id from the schema fields - I haven't had a chance t look deep enough for that yet

kartikay101:
Any progress on this, wanted to use the library but we need this feature.

epicfaace:
> @epicfaace
> 
> > Thanks, can you add a test?
> 
> ok

We need a test for progress on this issue.

epicfaace:
Also, what's the rationale behind this change? Currently, if there's a validation error, how does the focus work (without this PR)?

kartikay101:
There is no focus, the fields become red and you have to find your way through the form to look for what is missing

heath-freenome:
Reimplemented by #3434, closing

PR #1800, opened 2020-06-04
Add note about `action` prop being just for markup

jnachtigall:
This really puzzled me for some time, until I read the source code. I would have expected that submit happens *unless* I do a `event.preventDefault()`. But turned out that this is done be RJSF already and it is the other way around: By default no network request unless I submit. Which is fine, just I wished I knew :)

[Please describe them here]

### Checklist

### DIFF
--- a/docs/api-reference/form-props.md
+++ b/docs/api-reference/form-props.md
@@ -6,7 +6,9 @@ The value of this prop will be passed to the `accept-charset` [HTML attribute on
 
 ## action
 
-The value of this prop will be passed to the `action` [HTML attribute on the form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-action).
+The value of this prop will be passed to the `action` [HTML attribute on the form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-action). 
+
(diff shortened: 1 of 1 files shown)


docs/api-reference/form-props.md
```diff
@@ -6,7 +6,9 @@ The value of this prop will be passed to the `accept-charset` [HTML attribute on
 
 ## action
 
-The value of this prop will be passed to the `action` [HTML attribute on the form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-action).
+The value of this prop will be passed to the `action` [HTML attribute on the form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-action). 
+
+Note that this just renders the `action` attribute in the HTML markup. There is no real network request being sent to this `action` on submit. Instead react-jsonschema-form catches the submit event with `event.preventDefault()`and then calls the [`onSubmit`](#onSubmit) function where you could send a request programmatically with `fetch` or similar.
```
epicfaace, line 11:
```suggestion
Note that this just renders the `action` attribute in the HTML markup. There is no real network request being sent to this `action` on submit. Instead, react-jsonschema-form catches the submit event with `event.preventDefault()` and then calls the [`onSubmit`](#onSubmit) function, where you could send a request programmatically with `fetch` or similar.
```

PR #1802, opened 2020-06-07
fix: label background with variant outlined

jdbence:

The label does not have the white background when using variant outlined TextFields.
This image shows the changes

image


stanlemon:
FWIW This would be fixed by the changes in #1790

PR #1807, opened 2020-06-09
Support passing ui:label = false for Material UI

stanlemon:

[Please describe them here]

### Checklist

### DIFF
--- a/docs/advanced-customization/custom-widgets-fields.md
+++ b/docs/advanced-customization/custom-widgets-fields.md
@@ -121,6 +121,7 @@ The following props are passed to custom widget components:
 
 - `id`: The generated id for this field;
 - `schema`: The JSONSchema subschema object for this field;
+- `uiSchema`: The [uiSchema](#the-uischema-object) for this field;
 - `value`: The current value for this field;
 - `placeholder`: the placeholder for the field, if any;
 - `required`: The required status of this field;
--- a/docs/advanced-customization/custom-widgets-fields.md
+++ b/docs/advanced-customization/custom-widgets-fields.md
@@ -121,7 +121,7 @@ The following props are passed to custom widget components:
 
 - `id`: The generated id for this field;
 - `schema`: The JSONSchema subschema object for this field;
-- `uiSchema`: The [uiSchema](#the-uischema-object) for this field;
+- `uiSchema`: The uiSchema for this field;
 - `value`: The current value for this field;
 - `placeholder`: the placeholder for the field, if any;
 - `required`: The required status of this field;
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -111,7 +111,6 @@ declare module '@rjsf/core' {
         onBlur: (id: string, value: boolean | number | string | null) => void;
         onFocus: (id: string, value: boolean | number | string | null) => void;
         label: string;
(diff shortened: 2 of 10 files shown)


epicfaace:
I guess this approach (passing `displayLabel` to widgets) is required because in the material-ui theme, the label is handled from the widgets themselves (rather than in the [FieldTemplate directly as in the core bootstrap theme](https://github.com/rjsf-team/react-jsonschema-form/blob/49321786f79e305e16b8887d1ecd88149e250ddf/packages/core/src/components/fields/SchemaField.js#L140)).

Any thoughts on how we could clarify the documentation to note this fact (that `displayLabel` for widgets in bootstrap is not really useful)?

stanlemon:
@epicfaace I've actually been wondering if maybe this code... https://github.com/rjsf-team/react-jsonschema-form/blob/49321786f79e305e16b8887d1ecd88149e250ddf/packages/core/src/components/fields/SchemaField.js#L265-L279 should be extracted into `utils` so it could be used on `uiSchema` from any template. That would make this a little cleaner in my opinion. You can set `label = false` in any of the Material UI widgets to get the same effect as in the Bootstrap theme, it's just as you note, the location of where this decision has to be made is fundamentally different in the Material UI theme.

stanlemon:
@epicfaace I tried something totally different, which I think I prefer. What do you think?

stanlemon:
@epicfaace just checking in on this one to see if you had any thoughts on proceeding with the latest approach?

packages/core/index.d.ts
```diff
@@ -99,6 +99,7 @@ declare module '@rjsf/core' {
         > {
         id: string;
         schema: JSONSchema7;
+        uiSchema: UiSchema;
```
epicfaace, line 102: Can you update documentation to include uiSchema as a prop passed to all widgets?

stanlemon: @epicfaace just pushed, I think I added it to the right place. 🤞

packages/core/src/utils.js
```diff
@@ -368,6 +368,26 @@ export function getUiOptions(uiSchema) {
     }, {});
 }
 
+export function getDisplayLabel(schema, uiSchema, rootSchema) {
```
stanlemon, line 371: @epicfaace something I thought of this morning... even though I suspect this is covered anecdotally in base tests, should I add a specific unit test for the new extracted method?

docs/advanced-customization/custom-widgets-fields.md
```diff
@@ -121,6 +121,7 @@ The following props are passed to custom widget components:
 
 - `id`: The generated id for this field;
 - `schema`: The JSONSchema subschema object for this field;
+- `uiSchema`: The [uiSchema](#the-uischema-object) for this field;
```
epicfaace, line 124: I think this is a dead link.

packages/core/src/utils.js
epicfaace: Yeah, that would also be good, especially because it is now exported in utils.

stanlemon: Check out the tests I added and let me know if you think I should add any others.

epicfaace: Looks good!

PR #1811, opened 2020-06-12
feat(playground): state persistent using url and localStorage

anikethsaha:

Saving the state of the changeable data like the schema , uischema etc, using the URL can let us share the link and get the same input and it also helps in triaging the issues as the issue author can simply share the link of their change in the playground in the issue description instead of sharing the snippet in issue.  

_It is already being done by eslint_

### Checklist


anikethsaha:
cc @jevakallio

anikethsaha:
Cool, I will change it 👍

PR #1814, opened 2020-06-15
Material UI: Fix warning on textarea when value is null

stanlemon:

If you view the Sandbox for Nullable with Material UI you'll see a warning about value being null. This prevents that warning (and the resulting problems of an uncontrolled component).

### Checklist


epicfaace:
@stanlemon I don't see the issue -- I can't notice any warnings that show up in the case you described?

Screen Shot 2020-06-14 at 9 28 21 PM

stanlemon:
Oops! I'm a fool, the issue is on bootstrap. 🤦 

Screen Shot 2020-06-14 at 9 33 34 PM

I need to wrap up for the night, but I can track that issue down tomorrow. Sorry for the confusion!

PR #1815, opened 2020-06-15
Bootstrap: Fix warning when textarea value is null

stanlemon:

If you view the Sandbox for Nullable with Bootstrap (default) you'll see a warning about value being null. This prevents that warning (and the resulting problems of an uncontrolled component).

Screen Shot 2020-06-14 at 9 33 34 PM

### Checklist


epicfaace:
Hmm -- I still don't see this issue on the regular playground though?

image

stanlemon:
Try running it locally, it’s a React warning that only gets emitted in non-production builds. I’m assuming the live playground uses the production version of React which has all of the dev warnings stripped out.

PR #1817, opened 2020-06-16
Updated name to show only where pulled in by props.

tisdadd:

Material Theme had the "name" variable that was undefined when I was using [next.js](https://nextjs.org/). I have added in if the props were expecting it, and removed it if not. As it ran on the playground and I wasn't thinking it was appropriate to change linting rules, I simply built and tested locally. In the Text Widget, it still is coming through as undefined but since it now is expected and matches the definition, I thought this would be ok.

### Checklist


stanlemon:
You'll need to update the snapshots, add --updateSnapshot to the test command in the package to do this.

packages/material-ui/src/UpDownWidget/UpDownWidget.tsx
```diff
@@ -40,7 +40,6 @@ const UpDownWidget = ({
         required={required}
         type="number"
         disabled={disabled || readonly}
-        name={name}
```
stanlemon, line 43: It's not clear to me why you would want to drop this?

packages/material-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -75,7 +75,6 @@ const SelectWidget = ({
     <TextField
       id={id}
       label={label || schema.title}
-      name={name}
```
stanlemon, line 78: It's not clear to me why you would want to drop this?

packages/material-ui/src/UpDownWidget/UpDownWidget.tsx
tisdadd: It is a non-declared variable and causes the server-side rendering in nextjs to reject it.

Take a look at the full file, do a search and name didn't appear anywhere else.

https://github.com/rjsf-team/react-jsonschema-form/blob/a1eb74792f0ae46335fc8f2845db03ee7e3f095b/packages/material-ui/src/UpDownWidget/UpDownWidget.tsx

packages/material-ui/src/SelectWidget/SelectWidget.tsx
tisdadd: It is a non-declared variable and causes the server-side rendering in nextjs to reject it.

Take a look at the full file, do a search and name didn't appear anywhere else.

https://github.com/rjsf-team/react-jsonschema-form/blob/a1eb74792f0ae46335fc8f2845db03ee7e3f095b/packages/material-ui/src/SelectWidget/SelectWidget.tsx

stanlemon: I'm wondering if it would be better to extract this property rather than drop it? Maybe I'm mistaken, but isn't it passed into the constructor for the widget? (Same applies for the other dropped reference.)

tisdadd: It would, I assume, be preferable to have name attached to every widget type that renders to a form of input at some point. However, I am not familiar enough with the code to know why that doesn't seem to be the case in this instance. It is possible my search missed something.

nameSearch

I checked on the playground, and the default theme is not passing through a name, and even with the name included in the material theme it is not actually coming through as a proper name for the input box. As such, I had assumed this was an architectural decision that someone had made, and without knowing more I didn't want to change the code more than is required to fix it.

defaultTheme
materialTheme

stanlemon: Thanks for being patient with me. :) It looks like the name is used in the default theme if a title is not set. As I poke around more I see what you're seeing and I concur with what you're doing here. I am actually wondering now though, if we should just drop it entirely from the `<TextWidget />` too, rather than add it to the params?

lerna.json
```diff
@@ -2,5 +2,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "2.1.0"
+  "version": "2.1.1"
```
stanlemon, line 5: I wouldn't include the version bumps here. I think @epicfaace will handle that once enough changes have amassed, but I'll let him correct me if I'm wrong there.

tisdadd: Sorry about that, I accidentally left some changes in when committing after trying the various scripts.

packages/material-ui/src/SelectWidget/SelectWidget.tsx
tisdadd: It is quite possible that this would be the thing to do. Sorry if I seem short with anything, life has been super busy. I was very happy these tools ready existed or I would have been building them from the ground up... I realized that I have spent ~ 4 years building proprietary "foundational" (in the company I was working at) components and am super thankful that others are doing things that I am now able to use in my own projects.

I realized with the TextWidget commit in the 2.0.0 branch there were some things that may make it difficult if someone is trying to theme some of the parts that I personally am not yet using. Should I be mentioning that somewhere? Thank you for your patience with me.

stanlemon: Not sure I follow? I think dropping the name in this one should be fine.

lerna.json
stanlemon: Looks like some of the package file changes are still in there.

tisdadd: Ok - I found locally and reverted the change, also did the name removal as requested from TextWidget.

packages/material-ui/src/SelectWidget/SelectWidget.tsx
tisdadd: I dropped as requested. My question was about this line `<List dense={true} disablePadding={true}>` (and others that directly override a MaterialUI preference) because then it would not be possible to do the overriding of `dense` and `disablePadding` [if someone wanted to with default props](https://material-ui.com/customization/globals/) in their own theme.

```
const theme = createMuiTheme({
  props: {
    List: {
      // The default props to change
      disablePadding: false, // This wouldn't work, as the prop is passed in manually
    },
  },
});
```

I was not requiring it changed but was not sure how generic this theme is meant to be. It wasn't specifically an issue (at least for me) and I am new to the open-source world so I wasn't sure where to put the question appropriately.

stanlemon: Ahhhh, gotcha, I'm tracking now. Great question. The use of <List/> here is an alternative to MUI's typical use of the helper text for errors. The addition of dense and disablePadding were an attempt to match the indentation and alignment of helper text.

stanlemon:
@tisdadd we're so close! Thanks for your patience. Looking at the build there's still a snapshot failure in `Form.text.tsx`. Once that's resolved I think this looks great and really appreciate your contribution!

stanlemon:
I'm at a loss here as to why those snapshot tests are failing. I checked out your branch locally and didn't see those failures.

@epicfaace do you have any ideas here? I'm stumped.

tisdadd:
@stanlemon @epicfaace 

I also do not know why it testing fine locally and not on the check-in. :-(

epicfaace:
I'm not quite sure, but could you first try merging the latest master from rjsf-team/react-jsonschema-form into your branch? It might have something to do with this line: https://github.com/tisdadd/react-jsonschema-form/blob/master/packages/material-ui/package.json#L19

tisdadd:
@epicfaace @stanlemon That I think did it as it made it past the tests? But I don't know about the netlify parts of the process used here. Also, I was unable to do a direct pull request to mine, and so had to use a rebase to merge the master in... that didn't seem right, would I be better off starting with a new fork since it was only the three files that I cared about?

stanlemon:
@tisdadd that might be worth doing, because right now the PR is shows 235 files changes 😬

tisdadd:
> @tisdadd that might be worth doing, because right now the PR is shows 235 files changes 😬

Sounds good, I was thinking the same thing.

PR #1823, opened 2020-06-22
Fixed the problem where the ui:widget=text is not taken into account …

sramaswami11:
…for an integer field

These are the changes made:

- Made the Updown widget as the DefaultWidget in the NumberField class and passed it as prop while calling StringField function.
- String field function modified to consider  the DefaultWidget  value passed from Numberfield if it is defined.
- TextWidget will send a type prop set to a value of "text" when it is passed to the BaseInput function.

This is how the screenshot looks like after making the change:

image

### Checklist


packages/core/src/components/fields/NumberField.js
```diff
@@ -83,7 +83,12 @@ class NumberField extends React.Component {
     }
 
     return (
-      <StringField {...props} formData={value} onChange={this.handleChange} />
+      <StringField
+        defWidget={"updown"}
```
epicfaace, line 87: Can we call this `__defaultWidget` instead of `defWidget`?

sramaswami11: Done

epicfaace:
@sramaswami11 __defaultWidget, not _defaultWidget

Issue #1796, opened 2020-06-03
Material UI theme problem with initial false boolean


DennisKuhn:
Kia Ora :-)
thanks for the awesome library :-)

### Description
If a required boolean property is loaded, e.g. formData with `test: false`, it is not recognised as having a value. When clicking submit, it shows the error: _Please tick this box if you want to proceed_.

_schema.json:_
```
{
  "type": "object",
  "properties": {
    "test": {
      "type": "boolean"    
    }
  },
  "required": ["test"]
}
```

_formData.json:_
```
{
  "test": false 
}
```

_uiSchema.json:_
`{}`

#### Expected behavior
Allows to submit with initial false-boolean[Playground default-theme](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJ0ZXN0IjpmYWxzZX0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJ0ZXN0Ijp7InR5cGUiOiJib29sZWFuIn19LCJyZXF1aXJlZCI6WyJ0ZXN0Il19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6dHJ1ZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==)

#### Actual behavior
[...]

PR #1825, opened 2020-06-24
Material UI Checkbox fix required from schema

stanlemon:

Addresses a bug in #1796 for the checkbox widget when it is required according to the schema.

### Checklist


packages/material-ui/src/CheckboxWidget/CheckboxWidget.tsx
```diff
@@ -19,6 +22,11 @@ const CheckboxWidget = (props: WidgetProps) => {
     onFocus,
   } = props;
 
+  // Because an unchecked checkbox will cause html5 validation to fail, only add
+  // the "required" attribute if the field value must be "true", due to the
+  // "const" or "enum" keywords
+  const required = schemaRequiresTrueValue(schema);
```
epicfaace, line 28: If HTML5 validation is turned off, should we still use `schemaRequiresTrueValue(schema)` or should we revert to `props.required`?

stanlemon: That's a fair question and I don't know the answer honestly. This change in particular brings the behavior into alignment with the bootstrap theme (https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/widgets/CheckboxWidget.js#L51).  I'd defer to your direction on the question though, so just let me know if you'd like for me to make changes here.

epicfaace: Ah, makes sense, let's just keep it the way you did it then.

packages/material-ui/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "start": "tsdx watch",
     "build": "rimraf dist && tsdx build",
-    "test": "tsdx test"
+    "test": "tsdx test --updateSnapshot"
```
epicfaace, line 14: I think you should revert this (so CI doesn't update snapshots too)

stanlemon: Good catch, didn’t mean to let that sneak in. Thanks!

stanlemon:
@epicfaace Bump ☺️

Issue #1211, opened 2019-03-06
Add a "credits" file?


glasserc:
The majority of the work maintaining rjsf lately has been done @epicfaace, @LucianBuzzo,  and others who are not formally recognized in the project metainformation. Right now we just have n1k0's name in the package.json "author" field. I feel like that's a little inadequate. Should we maybe add a section to the README listing current maintainers (and maybe the past ones such as myself as "contributors emeritus")? Or maybe a `CREDITS.md` in the project root?

epicfaace:
Sure, I think we can make an `AUTHORS` file in the project directory and then npm should add them to the contributors? https://docs.npmjs.com/files/package.json#default-values

At least, that's how React itself does it https://github.com/facebook/react

heath-freenome:
Github seems to have this built-in now. See the [contributors](https://github.com/rjsf-team/react-jsonschema-form/graphs/contributors) link on the side

PR #1829, opened 2020-06-25
Remove authors from package.json

epicfaace:

For #1211.

TODO: we should add a root-level AUTHORS file and figure out how to make it automatically / periodically sync with github contributors instead.


stanlemon:
I think you could also do:
```json
  "contributors": [
    "url": "https://github.com/rjsf-team/react-jsonschema-form/graphs/contributors"
  ]
```

PR #1831, opened 2020-06-25
Create dependabot.yml

epicfaace:

Gives us automatic dependency updates -- should make it easier to handle updates for all our subpackages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,36 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,36 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
```
stanlemon, line 3: This is my ignorance showing,  us what uses pip?

epicfaace: mkdocs

epicfaace:
@stanlemon I'm not actually clear whether we can just use `packages/` (and dependabot will then check all subdirectories within it) or we need to enumerate all package folders. I just went with the latter to be safe.

PR #1873, opened 2020-06-29
Use conventional-changelog for CHANGELOG.md generation

epicfaace:

Builds off of #1872. I think it might be easier if we use `conventional-changelog` for each release in the future, which automatically creates links to relevant commits / PRs for each release.

Do you think it's useful to have a root-level changelog @stanlemon ?


epicfaace:
Actually, I don't really like this approach because it's probably better to pick the right bullet points to add in a changelog, not just use a commit log diff. See https://keepachangelog.com/en/1.0.0/#bad-practices

PR #1876, opened 2020-06-29
Update antd dependencies

epicfaace:

Turns out that the `antd`, `dayjs`, and `@ant-design/icons` dependencies are needed to make antd compile (they're also peerDependencies of `@rjsf/antd`).

Also, add `@fluentui/react` to the playground so that it can work with the published `@rjsf/fluent-ui` package as well.

This wasn't caught by our tests because everything is linked to local packages through Lerna by default ... any ideas on how we can better test things like this in our CI?

### DIFF
--- a/packages/antd/README.md
+++ b/packages/antd/README.md
@@ -57,10 +57,12 @@ Ant Design theme, fields and widgets for `react-jsonschema-form`.
 ### Prerequisites
 
 - `antd >= 4.0.0`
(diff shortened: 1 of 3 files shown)


stanlemon:
Maybe this is silly... but I wonder if it would be possible to add a script or something to add all of the peer dependencies from the other packages as dependencies for the playground?

Issue #1875, opened 2020-06-29
@rjsf/antd is too big


epicfaace:
### Description

[@rjsf/antd](https://www.npmjs.com/package/@rjsf/antd) is too big -- it's around 6.7 MB. Need to analyze why it's so big and then reduce the size.

### Version

2.2.0

LouizFC:
isnt it fixed by #1900 ? probably can be closed

epicfaace:
That's correct, thanks!

rmannibucau:
Hi,

Seems antd still does not have a production.min nor "light" bundle - current bundle is around 1.1M - compared to material-ui or bootstrap-4 dist/. Is it intended? Any hope it gets enhanced?

Thanks.

PR #1900, opened 2020-07-07
antd - smaller bundle, fix classNames, date time issues

### LINKED ISSUE #1880: Antd Theme date-time has problems.
### Prerequisites

### Description

The antd theme has some minor problems.
When I use the date component. Selecting the value will cause an error. The form crashed directly

### Steps to Reproduce

1. Open the playground. https://rjsf-team.github.io/react-jsonschema-form/
2. select antd theme and  Data & time
3. select a time.

#### Expected behavior

I hope the value like 2020-06-13T04:39:00.000Z

#### Actual behavior

The form crashed directly

### Tips

Antd used the moment.js . So this value is moment object.You need to use it to convert values
image

### LINKED ISSUE #1889: antd doesn't support uiSchema's classNames
### Prerequisites

### Description

It's very straightforward. This uiSchema:
```
const uiSchema = {
  name: {
    classNames: "custom-class-name"
  }
}
```

won't affect my antd form where there's a `name` string property.

This is the documentation about this feature: https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/#classnames

I've looked at other issues, and couldn't find anyone complaining about this. I've seen the same problem being reported for material-ui, but I think that they don't share the same source (regarding this problem)...

@delyanr, I've seen you have some work on this project, and, first, congrats! If you just point out to me what would be necessary to fix the problem, I'm willing to make a PR. I've tried understanding it, but there's so much going on that I don't know what can be changed.

delyanr:

Addresses #1875 - The way less/styles work in ant-design means it will never be a small bundle. I managed to shave off 400Kb by using manual imports instead of relying on babel-plugin-import. Excluding the source-maps, the bundle stands at ~1Mb.

Fixes #1889 - There was a minor typo in the `WrapIfAdditional` component which prevented `classNames`. Tests were also affected.

Fixes #1880 - I don't have much time to check whether we can allow the users of `@rjsf/antd` to select which date library they use (e.g. moment or dayjs). Since dayjs is super small, I'm forcing its use here. Also deprecated the use of the antd-dayjs-webpack-plugin.

Finally, it would be great if there are a couple of volunteers who can add some additional tests. :)

### Checklist

### DIFF
--- a/packages/antd/src/components/DatePicker/index.js
+++ b/packages/antd/src/components/DatePicker/index.js
@@ -1,0 +1,6 @@
+import dayjsGenerateConfig from 'rc-picker/lib/generate/dayjs';
+import generatePicker from 'antd/lib/date-picker/generatePicker';
+
+const DatePicker = generatePicker(dayjsGenerateConfig);
+
+export default DatePicker;
--- a/packages/playground/webpack.config.dev.js
+++ b/packages/playground/webpack.config.dev.js
@@ -62,7 +62,7 @@ module.exports = {
       },
       {
         test: /\.less$/,
-        include: /node_modules\/antd/,
+        include: /node_modules[\\/]antd/,
         use: [
           {
             loader: "style-loader",
--- a/packages/antd/babel.config.js
+++ b/packages/antd/babel.config.js
@@ -15,13 +15,6 @@ module.exports = {
     "@babel/preset-react"
   ],
   "plugins": [
-    [
-      "import",
-      {
-        "libraryName": "antd",
-        "style": true
-      }
-    ],
     "@babel/plugin-proposal-object-rest-spread",
     "@babel/plugin-proposal-class-properties",
     [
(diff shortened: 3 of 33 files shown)


delyanr:
@epicfaace This might have a conflict with #1867 as I fixed some issues in the playground CI already.

packages/antd/src/components/DatePicker/index.js
```diff
@@ -0,0 +1,6 @@
+import dayjsGenerateConfig from 'rc-picker/lib/generate/dayjs';
```
epicfaace, line 1: This does seem to be a better solution than using the day-js-webpack-plugin.

Was day-js-webpack-plugin causing #1880? If so, why?

packages/playground/webpack.config.dev.js
```diff
@@ -62,7 +62,7 @@ module.exports = {
       },
       {
         test: /\.less$/,
-        include: /node_modules\/antd/,
+        include: /node_modules[\\/]antd/,
```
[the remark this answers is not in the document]
delyanr: This ensures webpack matches both `/` and `\` as path separators, and reduces issues between Unix and Windows.

packages/antd/src/components/DatePicker/index.js
delyanr: The `day-js-webpack-plugin` was not working as intended for some reason, meaning that the `DatePicker` component was still trying to access the default `moment` library instead of `dayjs` for date/time operations.

MacroBet:
Could you please version this?

PR #1930, opened 2020-07-20
Upgrade version from @babel/runtime-corejs2 to @babel/runtime-corejs3

ribeiromiranda:

Upgrade version from @babel/runtime-corejs2 to @babel/runtime-corejs3 to avoid conflicts with other libraries.

### Checklist

### DIFF
--- a/packages/antd/package-lock.json
+++ b/packages/antd/package-lock.json
@@ -3589,16 +3589,6 @@
 				"regenerator-runtime": "^0.13.4"
 			}
 		},
-		"@babel/runtime-corejs2": {
-			"version": "7.10.3",
-			"resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.10.3.tgz",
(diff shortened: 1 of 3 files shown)


ribeiromiranda:
> Seems good. Do you think this might cause any backwards-incompatible changes?

Yes. Because the documentation highlights the incompatibility between the two versions.
https://babeljs.io/blog/2019/03/19/7.4.0#migration-from-core-js-2

"Since versions 2 and 3 of core-js are incompatible with each other (we don't want to break your code!), it isn't enabled by default."

epicfaace:
Hmm, that means if it's backwards-incompatible, we'd have to wait until a major version bump for rjsf to merge this change in.

However, if all the imports that rjsf uses from @babel/runtime-corejs2 have backwards-compatible behavior in @babel/runtime-corejs3, then we should probably be able to upgrade without a major version bump, I think. Any idea if this is the case @ribeiromiranda ?

PR #1945, opened 2020-07-23
fix: don't recurse circular references (#14)

erunion:

When the form attempts to render a schema that has a circular `$ref`, it will currently do so recursively resulting in a maximum call stack fatality.

You can see this in action by loading up the playground and dropping in the following schema:

```json
{
  "definitions": {
    "node": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "children": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/node"
          }
        },
        "otherNode": {
          "$ref": "#/definitions/node"
        }
      }
    }
  },
  "type": "object",
  "properties": {
    "tree": {
      "title": "Recursive references",
      "$ref": "#/definitions/node"
    }
  }
}
```

This change attempts to resolve this issue by detecting circular refs during the `computeDefaults` stage, and backing out of the backing out of the recursive resolution step when a circular pointer is detected.

Resolves https://github.com/rjsf-team/react-jsonschema-form/issues/1160

### Checklist


erunion:
@epicfaace Will do. I'm downgrading this to a draft state for the moment as we've got some other fixes coming as right now this change rejects deep references that loop background on itself, but aren't actually technically circular.

epicfaace:
@erunion hey, what happened -- is this no longer an issue?

erunion:
We've deviated pretty hard from the current trunk and it's going to be a lot of work for us to get this up to date.

PR #1949, opened 2020-07-27
Fix type of parameter widget in getWidget and hasWidget functions

Klynger:

These two functions receive `widget` that can be `Widget | string`, but in typings, this is not expressed.

More details can be found here where I add the same changes in V1: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46374

### Extra information

I committed the changes made when I ran `npm install` for each package, but if there is any problem you can just comment in the PR and I can remove the commits

### Checklist

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -285,13 +285,13 @@ declare module '@rjsf/core' {
 
         export function getWidget(
             schema: JSONSchema7,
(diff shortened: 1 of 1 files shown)


Klynger:
Done @epicfaace! I made another in the return type of `getWidget` because it never returns `Error`, only throws it if there is a problem

PR #1982, opened 2020-08-08
Add support for additionalProperties for material-ui theme

ValYouW:

Adding support for `additionalProperties` for the material-ui theme, partially solves #1927

### Checklist

* [X ] **I'm adding or updating code**
  - [X ] I've added and/or updated tests

### DIFF
--- a/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
+++ b/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
@@ -1,15 +1,48 @@
 import React from 'react';
 
 import Grid from '@material-ui/core/Grid';
-import { makeStyles } from '@material-ui/styles';
+import Button from '@material-ui/core/Button';
+import AddIcon from '@material-ui/icons/Add';
 
-import { ObjectFieldTemplateProps } from '@rjsf/core';
+import { makeStyles } from '@material-ui/styles';
+import { UiSchema, ObjectFieldTemplateProps, utils } from '@rjsf/core';
+import { JSONSchema7 } from 'json-schema';
+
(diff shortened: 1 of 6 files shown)


packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
```diff
@@ -1,16 +1,49 @@
...
+  </Button>
+);
 
 const useStyles = makeStyles({
   root: {
     marginTop: 10,
   },
 });
 
+const canExpand = function canExpand(formData: object, schema: JSONSchema7, uiSchema: UiSchema) {
```
epicfaace, line 30: This function seems to be very similar to the [canExpand function in @rjsf/antd](https://github.com/rjsf-team/react-jsonschema-form/blob/f378583a27365bc2ae5cd99897d7385abd58949a/packages/antd/src/templates/ObjectFieldTemplate/index.js#L81). Can you instead move this function to @rjsf/core utils.js and then import it from there?

ValYouW:
It's also here in [core](https://github.com/rjsf-team/react-jsonschema-form/blob/f378583a27365bc2ae5cd99897d7385abd58949a/packages/core/src/components/fields/ObjectField.js#L14)

epicfaace:
> It's also here in [core](https://github.com/rjsf-team/react-jsonschema-form/blob/f378583a27365bc2ae5cd99897d7385abd58949a/packages/core/src/components/fields/ObjectField.js#L14)

That's fine, but can you refactor that copy of the function that's called from material-ui into utils for this PR? Later, we can do a bigger refactoring that makes sure all instances of the functions are called from utils.

epicfaace:
Resolved by #1993

PR #1990, opened 2020-08-12
Fix race condition for onChange when formData is controlled prop

olzraiti:

This bug was introduced by #1804

The race condition occurs when (1) Form receives the `formData` prop, and will call the `onChange` function, and (2) when `onChange` is called at the same time by some component rendered by the Form.

### Checklist


olzraiti:
The test is passing in the latest version, closing.

epicfaace:
Huh, I wonder what might have fixed it... good it's fixed though! :)

olzraiti:
Hmm, though after upgrading to the latest rjsf version, my tests broke - adding this patch again fixed the tests... There's something that the test doesn't catch.

epicfaace:
hmm, do you want to try adding a better test that catches this bug (or perhaps it's a new bug)?

PR #1553, opened 2019-12-18
Add null checks for ArrayField

gavin-haynes-grow:
### Reasons for making this change

When an array is set to "null" instead of empty, the page crashes instead of specifying an error or allowing it to be null.

Related to https://github.com/rjsf-team/react-jsonschema-form/issues/1346 but that may be a separate issue.

This change fixes the issue you can see at https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJyZXF1ZXN0IjpudWxsfSwic2NoZW1hIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7InJlcXVlc3QiOnsidHlwZSI6WyJhcnJheSIsIm51bGwiXSwiaXRlbXMiOnsidHlwZSI6InN0cmluZyJ9fX19LCJ1aVNjaGVtYSI6e319, where nothing is displayed because the component crashes.

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [x] **I'm adding or updating code**
  - [ ] I've added and/or updated tests (not sure if this can be tested with the current test framework. I tried a simple test using setProps but it didn't fail on the previous commit)
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #1992, opened 2020-08-13
Add null checks for ArrayField (New PR for #1553)

richmahn:
This is a new PR to replace PR #1553 since that PR's author is no longer responding. Resolves conflicts.


ri0ter:
Hi @richmahn is that PR still valid?
Does it require some more work (besides update/rebase), something I could help with?

@epicfaace do you think that this PR could be merged?

jsani-radar:
wanted to bump this PR, commented on the original issue: https://github.com/rjsf-team/react-jsonschema-form/issues/1346#issuecomment-769439171, let me know if I can assist in getting this through at all

epicfaace (commented):
Thanks for updating the PR! I'm a bit confused about what exactly the logic of the additions is -- when exactly would `keyedFormData` be equal to `null`? Wouldn't it always be equal to `[]` if the form data itself is equal to `null`? (see https://github.com/rjsf-team/react-jsonschema-form/blob/a4f6c1fc7ce61bbabd794ba528781b7e195d8ad0/packages/core/src/components/fields/ArrayField.js#L184-L193)?

richmahn:
@epicfaace The keyDataValue being null happens whenever the json file has a null, such as:

`{ "tags": null }`

Before it was only allowing it to be `{ "tags": [] }`

I have no idea why it is failing to build/deploy. Says that `cross-env` doesn't exist on netlify.

ri0ter:
@epicfaace I did investigate it further and as you are right that `generateKeyedFormData` provides default to `[]` but it is being called only in constructor to set the state. Now in `getDerivedStateFromProps` it receives the same props (in this case null) which has no fallback to `[]` which leads to the error.

Another thing (could be separate issue) is if you set the type to nullable it won't work, it will crash the same way. But in this case I would assume that nullable check should happen eariler and it should not even enter ArrayField component.

I wanted to mention that there are other issues opened that seems to be related to this one: #1957, #282, #2153

epicfaace:
> I did investigate it further and as you are right that `generateKeyedFormData` provides default to `[]` but it is being called only in constructor to set the state. Now in `getDerivedStateFromProps` it receives the same props (in this case null) which has no fallback to `[]` which leads to the error.

Ah, I see. So do you think a better solution might be to instead modify `getDerivedStateFromProps` in some way so that it gives us a default of `[]` for keyedFormData? That way, we won't need all these null checks. https://github.com/rjsf-team/react-jsonschema-form/blob/49e9184afa4e10906aa6b3cde9f324a1d26a2ed1/packages/core/src/components/fields/ArrayField.js#L222-L245

epicfaace:
Ah, wait, I see that this PR (https://github.com/rjsf-team/react-jsonschema-form/pull/2154) actually does exactly that! If you're good with it, I think we should merge that one instead of this one.

richmahn:
@epicfaace I would not be hurt in the least if you did that. Haha

ri0ter:
Yeah, just tested it and it seems to work.
It also kinda fixes #1346, kinda because in my opinion value `null` used with type `['array','null']` should display null instead of an empty array, but that might be more general issue with displaying nullables (or it's a feature).

Just as a side note, atm if I'll provide null value with type `['array']`, it will display me an empty ArrayField, but at the same time I will see an error saying that it "should be an array". The problem is that visually it's already an array. So just as a future UI improvement, maybe if the value has wrong type it shouldn't display empty ArrayField but a special element which could be saying "initialize array".

epicfaace:
Thanks for the explanation @ri0ter . I'll close this issue in favor of #2154, and close out some of these other related issues.

With the other PR, I did notice a minor issue with nullable arrays; I'm cataloguing it in https://github.com/rjsf-team/react-jsonschema-form/issues/2212.

> Just as a side note, atm if I'll provide null value with type `['array']`, it will display me an empty ArrayField, but at the same time I will see an error saying that it "should be an array". The problem is that visually it's already an array. So just as a future UI improvement, maybe if the value has wrong type it shouldn't display empty ArrayField but a special element which could be saying "initialize array".

Could you make this into a separate issue so we could address it later?

PR #2002, opened 2020-08-19
Maintain AJV properties when custom validator is in use.

RoboPhred:

This fixes the issue where AJV errors are stripped of important properties when a custom validator is in use.

This fixes issue #1596

### Checklist


packages/core/test/validate_test.js
```diff
@@ -239,20 +240,23 @@ describe("Validation", () => {
...
         errors = result.errors;
         errorSchema = result.errorSchema;
       });
 
       it("should return an error list", () => {
-        expect(errors).to.have.length.of(1);
-        expect(errors[0].stack).eql("pass2: passwords don't match.");
+        expect(errors).to.have.length.of(2);
+        expect(errors[0].stack).eql(".numberWithMinimum should be >= 5");
+        expect(errors[1].stack).eql("pass2: passwords don't match.");
```
RoboPhred, line 252: Note how the stack message is subtly different here:
RJSF custom errors use `fieldName: message`
AJV custom errors use `.field.path message`

I am slightly worried that this might be considered a breaking change, as previously our mangling of AJV errors would have changed the stack message to the RJSF format, so anything expecting the stack to remain the same will suffer.

Perhaps I should add an exception to rewrite AJV error stacks to the RJSF stack format?

packages/core/src/validate.js
```diff
@@ -252,10 +255,10 @@ export default function validateFormData(
   const errorHandler = customValidate(formData, createErrorHandler(formData));
   const userErrorSchema = unwrapErrorHandler(errorHandler);
   const newErrorSchema = mergeObjects(errorSchema, userErrorSchema, true);
-  // XXX: The errors list produced is not fully compliant with the format
-  // exposed by the jsonschema lib, which contains full field paths and other
-  // properties.
-  const newErrors = toErrorList(newErrorSchema);
+
+  // Append the user's errors to the current error list
+  //  Mantain the original error list to keep its AJV provided detail properties.
```
epicfaace, line 260:
```suggestion
  //  Maintain the original error list to keep its AJV-provided detail properties.
```

epicfaace, line 260: Should we change typings as well because of this change?

packages/core/test/validate_test.js
epicfaace: > My recommendation is changing RJSF to use AJV style messages in all cases

I like this approach. I think rjsf is due for a new breaking change soon (there are several PRs that would require breaking changes but are probably needed), so I think we should be fine if we just change this PR to do this.

RoboPhred: I switched custom errors to use AJV style error messages.

Note that errors on the root now show up with a single dot, as in ". This is a root error".  This appears a bit strange, but is indeed the way AJV shows root errors.

packages/core/src/validate.js
RoboPhred: It would be worth exposing the "property" property on the error object.  I do not know where the typings are stored though, and could not find them in the project.

packages/core/src/validate.js
```diff
@@ -80,21 +80,22 @@ function toErrorSchema(errors) {
   }, {});
 }
 
-export function toErrorList(errorSchema, fieldName = "root") {
-  // XXX: We should transform fieldName as a full field path string.
+export function toErrorList(errorSchema, fieldPath = []) {
```
epicfaace, line 83: Sorry, I'm a bit confused as to what the changes in `toErrorList` have to do with fixing #1596. Doesn't the change in line 259 already ensure that the original error list is untouched?

packages/core/src/validate.js
epicfaace: Sorry, I was mistaken earlier -- the typescript types already include the `property` property (https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/index.d.ts).

packages/core/src/validate.js
RoboPhred: This is for the comment above:
```
Note how the stack message is subtly different here:
RJSF custom errors use fieldName: message
AJV errors use .field.path message
```

toErrorList only tracked the current field's name, not the full path to that field.  I changed the function to receive the full field path so that it can use the dot-separated field path when generating the message.

epicfaace: Ah, thanks for the pointer.

epicfaace (commented):
@RoboPhred , since this is a breaking change, can you add an entry to the 3.x upgrade guide that explains to users what has changed and what they may need to change in their own code?

The guide isn't yet on the master branch (it's in #2211), but here's the version of the guide that's in that branch: https://github.com/rjsf-team/react-jsonschema-form/blob/26d34051bbc063777645aa825ed4192e260930ab/docs/3.x%20upgrade%20guide.md You can just create the `docs/3.x upgrade guide.md` file on this branch.

saadq:
Hi all, wanted to check on the status of this PR. What's still needed for this change to get merged in? Is it just updating the upgrade guide?

penx:
@saadq looks like it missed v3 so it's on the roadmap for v4: https://github.com/rjsf-team/react-jsonschema-form/issues/2437

heath-freenome:
@RoboPhred With the upcoming beta 5 that already has a bunch of breaking changes, are you willing to port this to the `rjsf-v5` branch into the `validator-ajv6` package?

heath-freenome:
Refactored into v5 beta via https://github.com/rjsf-team/react-jsonschema-form/pull/3047

PR #2010, opened 2020-08-22
Remove componentWillReceiveProps from Form.jsx

jimmycallin:

componentWillReceiveProps is deprecated, and is currently in use in two places in this project, which causes annoying warnings.

This PR removes the one in Form.jsx, including some refactoring (required due to getDerivedStateFromProps is a static method).

### Checklist


packages/core/src/components/Form.js
```diff
@@ -31,134 +36,38 @@ export default class Form extends Component {
...
-      additionalMetaSchemas,
-    };
-    if (schemaValidationErrors) {
-      nextState.schemaValidationErrors = schemaValidationErrors;
-      nextState.schemaValidationErrorSchema = schemaValidationErrorSchema;
+  static getDerivedStateFromProps(props, state) {
+    if (!deepEquals(state.lastProps, props)) {
+      return getStateFromProps(props, props.formData, state);
     }
-    return nextState;
```
sumanbh, line 136: Shouldn't this return `null` if we're not updating state as opposed to implicitly returning undefined?

From [docs](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops):
```
It should return an object to update the state, or null to update nothing.
```

jimmycallin: Yes, you're right! Good catch. I added null as default case.

RoboPhred:
Why the reliance on deep equals for comparing property references?  Does RJSF really want to support mutating object references?  Such an action will not cause react to re-render the component anyway, so supporting this seems to be just a big performance sink.

We are working with very large forms (several hundered fields), and this area has been a pain point for us.  I have an outstanding question about the issue [here](https://github.com/rjsf-team/react-jsonschema-form/issues/1631#issuecomment-677955094)

packages/core/src/components/Form.js
```diff
@@ -31,134 +36,39 @@ export default class Form extends Component {
...
-      schemaValidationErrors = errors;
-      schemaValidationErrorSchema = errorSchema;
-    } else {
-      const currentErrors = getCurrentErrors();
-      errors = currentErrors.errors;
-      errorSchema = currentErrors.errorSchema;
-      schemaValidationErrors = state.schemaValidationErrors;
-      schemaValidationErrorSchema = state.schemaValidationErrorSchema;
+  static getDerivedStateFromProps(props, state) {
+    if (!deepEquals(state.lastProps, props)) {
```
RoboPhred, line 62: Why the reliance on deep equals for comparing property references? Does RJSF really want to support mutating object references? Such an action will not cause react to re-render the component anyway, so supporting this seems to be just a big performance sink.

We are working with very large forms (several hundered fields), and this area has been a pain point for us. I have an outstanding question about the issue here

RoboPhred:
> 1. If props has formData set, _do not_ save formData state internally and let the parent component control formData entirely by listening to onChange events.

This would go a significant way to fixing [this issue](https://github.com/rjsf-team/react-jsonschema-form/issues/1631), although the continued use of deep equals still needs to be addressed.

packages/core/src/components/Form.js
jimmycallin: Hi @RoboPhred, from my view there is really no reason to use the deepEquals. I didn't spend any thought in reevaluating this, since I just wanted to have a PR with no breaking changes. I tried to replace deepEquals with shallow equals, and it didn't seem to break any tests, so I just pushed this change. We'll see if it passes review. :)

RoboPhred: Whoops, that change might not work, as you will always receive a new core props reference each time react re-renders, meaning it will try to make a new state every render.  You should compare each individual property to its previous value.  

You might be able to get away with something like:

```js
const keys = Object.keys(props);
if (keys.some(key => state.lastProps[key] != props[key]) {
  // regenerate state
}
```

I am a bit worried about adding or removing properties, but I /think/ that if you have an undefined property passed in react, react will actively create that property and set its value as undefined, meaning the property will still show up in Object.keys().  I don't think there are any cases where we might have a property defined in lastProps but not defined in props.

For reference, you can see my previous attempt at this [here](https://github.com/rjsf-team/react-jsonschema-form/pull/1634).  In that case, I individually checked the props by name, but I think we can get away with a bulk scan over the new properties.

jimmycallin: Ah, good catch. I added an additional check that checked the length of the number of properties, which I think will fix the case of any removed properties not triggering an onChange event. Not sure why the Netlify CI tests are breaking, but I think that's unrelated.

epicfaace:
@jimmycallin hey, I think this is a great change! Just haven't had a chance to review it yet. I'll put it on my list and try to get this in soon.

Zakhar-Kutsko:
Hi there! When you will merge these changes and release a new version?

epicfaace:
It's on the agenda for a sync review in a future meeting https://docs.google.com/document/d/12PjTvv21k6LIky6bNQVnsplMLLnmEuypTLQF8a-8Wss/edit

packages/core/src/utils.js
```diff
@@ -1265,3 +1265,115 @@ export function schemaRequiresTrueValue(schema) {
 
   return false;
 }
+
+export function getRegistry(props) {
```
epicfaace, line 1269: We should move all of these new functions in utils.js to just Form.js (because they're  quite Form-specific)

epicfaace (commented):
@jimmycallin can you join us at one of our synchronous meetings so you can explain the PR to us? You can add the PR to the agenda of a day in which you're able to make it. Thanks! https://docs.google.com/document/d/12PjTvv21k6LIky6bNQVnsplMLLnmEuypTLQF8a-8Wss/edit

packages/core/src/utils.js
```diff
@@ -1265,3 +1265,115 @@ export function schemaRequiresTrueValue(schema) {
...
+    resolvedSchema,
+    validate,
+    transformErrors,
+    additionalMetaSchemas,
+    customFormats
+  );
+}
+
+export function getStateFromProps(props, inputFormData, state = {}) {
+  const edit = typeof inputFormData !== "undefined";
```
epicfaace, line 1304: We should figure out -- why does this omit several lines from the original `getStateFromProps` function?

packages/core/src/components/Form.js
```diff
@@ -31,134 +36,42 @@ export default class Form extends Component {
...
-      errors,
-      errorSchema,
-      additionalMetaSchemas,
-    };
-    if (schemaValidationErrors) {
-      nextState.schemaValidationErrors = schemaValidationErrors;
-      nextState.schemaValidationErrorSchema = schemaValidationErrorSchema;
+  static getDerivedStateFromProps(props, state) {
+    if (
+      Object.keys(props).length !== Object.keys(state.lastProps).length ||
```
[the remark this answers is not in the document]
jimmycallin: Please see the discussion here about this change https://github.com/rjsf-team/react-jsonschema-form/pull/2010#discussion_r480266606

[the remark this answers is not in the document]
epicfaace: Hmmm, I'd prefer in this PR if we actually don't change to removing deepEquals and just focus on the refactoring -- we can remove deepEquals in a separate PR. This would make it easier to review that change / make this PR less likely to break things.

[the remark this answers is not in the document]
jimmycallin: Sure thing, I'll revert this.

packages/core/src/components/Form.js
```diff
@@ -6,18 +6,23 @@ import _isEmpty from "lodash/isEmpty";
...
+  getRegistry,
+  validate,
+  getStateFromProps,
 } from "../utils";
-import validateFormData, { toErrorList } from "../validate";
+import { toErrorList } from "../validate";
 import { mergeObjects } from "../utils";
 
+function handleChange(props, state) {
+  const { lastProps, ...formState } = state;
```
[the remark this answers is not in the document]
jimmycallin: On [line 62](https://github.com/rjsf-team/react-jsonschema-form/pull/2010/files#diff-a3830d0e95feb07e61c8e35cf02715c655766e69871111b67373658e5d98398bR62) we need to compare prop changes to determine if the consumer has updated their props.

If the consumer has updated the props, we update the internal state based on the props.

We could previously do this in componentWillReceiveProps since that method supports both viewing previous and new props. But since getDerivedStateFromProps only shows the new prop updates, we need to store the old props in the state.

packages/core/src/utils.js
jimmycallin: the lines omitted are mostly just variable declarations that I pick up directly from the prop object instead

jimmycallin:
@epicfaace i have now updated the PR with all your proposed changes and questions, and merged the lastest from master. so it's ready for another round of review :)

packages/core/src/components/Form.js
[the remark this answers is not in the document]
jimmycallin: And the reason why we cannot always derive state from props, is due to us supporting the mixed controlled/uncontrolled behavior we discussed on our last meeting this Friday.

epicfaace (commented):
Per our meeting -- `deepEquals(props, state.lastProps)` is probably too expensive / too much of a performance hit, so let's instead actually change some of the logic so that the component behaves more like a controlled / uncontrolled component.

Maybe we could let the user either
- pass in onChange and formData, which makes the Form component fully controlled, or
- not do so, which would make the Form component fully uncontrolled.

@jimmycallin , can you write up (or tell us in a future call) how the current behavior of the form works with formdata / onchange? We don't fully understand it.

ArtsiomAntropau:
Hi, @jimmycallin, do you have the time and desire to finish this PR? 
Our team can contribute to this PR if it is needed, and you with @epicfaace will highlight what needs to be improved.

jimmycallin:
@ArtsiomAntropau Hi! I have fairly limited time nowadays, and I'm not using rjsf in my work anymore, so if your team can take over this task I would greatly appreciate it.

As @epicfaace mentioned above, this PR in its current state is not a viable solution. We discussed options in a [weekly meeting](https://github.com/rjsf-team/react-jsonschema-form/issues/2677) a few weeks ago, and decided that a breaking change where we make the form component a proper controlled/uncontrolled component as described in these two comments makes more sense:

https://github.com/rjsf-team/react-jsonschema-form/pull/2010#issuecomment-678673604
https://github.com/rjsf-team/react-jsonschema-form/pull/2010#pullrequestreview-900430579

If you are willing to implement this, it would be very much appreciated.

cdolek:
Hi guys, is there a middle step where the warning goes away and `UNSAFE_componentWillReceiveProps` is not used anymore, and then the breaking change is introduced with refactoring?

jimmycallin:
Since this isn't a viable approach for removing componentWillReceiveProps, I'll close this one. Due to personal reasons I don't think I will be able to redo the PR according to [the previous comment](https://github.com/rjsf-team/react-jsonschema-form/pull/2010#issuecomment-1078356278), so if someone else wants to step in, please do so.

Kos:
Is there a relevant issue which we could follow for updates regarding this bit?

> We discussed options in a https://github.com/rjsf-team/react-jsonschema-form/issues/2677 a few weeks ago, and decided that a breaking change where we make the form component a proper controlled/uncontrolled component as described in these two comments makes more sense

Zakhar-Kutsko:
Hi there! Following the documentation, by default, <Form /> is already an uncontrolled component.@epicfaace, Or is your suggestion exactly to remove deepEquals(props, state.lastProps)?

cdolek:
@ArtsiomAntropau is your team working on this?

ArtsiomAntropau:
@cdolek Hi, 
@Zakhar-Kutsko still investigating on it

summer-cook:
Hi @ArtsiomAntropau , was just wondering if there is an open issue for this I could follow? Still getting this warning on our Form

PR #2012, opened 2020-08-23
Fix the `module` path for bootstrap-4, fluent-ui, material-ui

abdvl:

The `module` path in the `@rjsf/bootstrap-4` package.json is incorrect

### Checklist


epicfaace:
@abdvl any idea how we could add a test for this to prevent this from regressing in the future?

Issue #1893, opened 2020-07-02
require is not defined


jgroth:
I'm trying to upgrade from version 1.8.1 to version 2.2.0 but I get the following error when I run my code
```
Uncaught ReferenceError: require is not defined
```

It seems to originate from the following code.

https://github.com/rjsf-team/react-jsonschema-form/blob/7c18c751f65451dfc9d3ddbdb500b4ab707371a3/packages/core/src/utils.js#L63-L71

I've searched through the issues here and it seems that this might be solved by building with Webpack. I'm using Rollup and switching to Webpack is not an option. Rollup is also discouraging mixing `import` and `require` statements, so that might be why I get this problem.

Is there a workaround for this issue, or is it possible to refactor the above code to not rely on `require`? As of now upgrading seems impossible for me which is unfortunate.

epicfaace:
@jgroth would it be possible for you to make a repository with the minimal build config needed to reproduce this error? It would be easier to debug that way.

jgroth:
I have a branch where it is reproducable, I'm not sure how minimal it is though :smile: I tried to remove as much noise as possible from it. It is available here https://github.com/Lundalogik/lime-elements/tree/upgrade-rjsf
```
npm install
npm start
```
should to the trick. You should get a page where the form is used and throws an error in the console

Let me know if it does not work, I noticed we were using some private repos but I think I got rid of all of them for this build

ivarprudnikov:
Tried checking out and running. Bits of somewhat useful information in addition to the above:

*Console log*
```
Uncaught ReferenceError: require is not defined
    getDefaultRegistry http://localhost:3333/build/limel-form.entry.js:43048
    getRegistry http://localhost:3333/build/limel-form.entry.js:44508
    validate http://localhost:3333/build/limel-form.entry.js:44472
    getStateFromProps http://localhost:3333/build/limel-form.entry.js:44420
    Form http://localhost:3333/build/limel-form.entry.js:44362
```

Failing transpiled code:
```
function getDefaultRegistry() {
  return {
    fields: require("./components/fields")["default"],
    widgets: require("./components/widgets")["default"],
    definitions: {},
    rootSchema: {},
    formContext: {}
  };
}
```

I've noticed that the code is transpiled into what looks like `es5` except those `require` statements.

[...]

PR #2028, opened 2020-09-03
fix: avoid mixing CJS require in ESM modules

kuwoyuki:

Fixes https://github.com/rjsf-team/react-jsonschema-form/issues/1893
Mixing CommonJS `require` in ESM modules is incorrect. It works in webpack, however a lot of other bundlers do not support that.

### Checklist


epicfaace (approved):
Looks good, thanks! I'm not sure why the `require` statements were there earlier.

PR #2030, opened 2020-09-05
(Chore) Bumped prettier to 2.1.1

jimmycallin:

Steps to create this PR:

1. Upgraded Prettier to 2.1.1.
2. Removed Prettier from core.
3. Created .prettierignore with dist, build, and coverage.
4. Removed redundant prettier config in package.json, since they are default in Prettier 2. 
5. Ran `npx prettier --write .`

### Checklist


jimmycallin:
The issue with this PR is that Prettier does no longer support Node < 10.

epicfaace (commented):
Looks good -- it's especially good that we are able to format markdown too.

Could you also update the formatting checks and lint-staged commands to make sure they also include previously-ignored files, such as `.md` files and other config / js files that weren't included in these checks?

jimmycallin:
@epicfaace Fixed! If you don't mind I would also like to remove the last special option (jsxBracketSameLine: true), so it's all default behavior.

jimmycallin:
This one is stale and will need to be redone, closing.

PR #2031, opened 2020-09-06
Drop support for node 8 / 9

epicfaace:

Node 8 / 9 are no longer supported, and some of our dependencies no longer support these versions at all (see #2030, #2014).


jimmycallin (changes requested):
💯 💯 

But I think you also need to update the `engines` key value in the various package.json, and whatever it is that sets the required builds in the CI check.

stanlemon (approved):
Awesome! One thought, doesn't 10 end the maintenance window in April?

epicfaace:
That's true, though I think we can wait a bit before actually dropping support for it.

stanlemon:
> That's true, though I think we can wait a bit before actually dropping support for it.

Sure, I was just thinking in terms of v3 and breaking changes is all.

epicfaace:
Yeah, that's a good point. I guess we might as well drop official support for it (though it will likely still work on Node 10, we can just remove it from CI).

epicfaace:
Please take a look: https://github.com/rjsf-team/react-jsonschema-form/pull/2299

PR #2105, opened 2020-10-24
fix: submit event bubbling with React 17

### LINKED ISSUE #2104: React 17 submit event not bubbling
Because of the changes in way [react 17 attaches event listeners](https://reactjs.org/blog/2020/10/20/react-v17.html#changes-to-event-delegation), submitting the form programatically no longer works because submit event doesn't bubble.
Solution for this would be to add `bubbles: true` when dispatching CustomEvent submit.

dreadheadsic:
React 17 has changed the way event handlers are attached - [more info](https://reactjs.org/blog/2020/10/20/react-v17.html#changes-to-event-delegation) and because of that, submit event doesn't bubble up unless you explicitly tell it to bubble.
This issue affects submitting the form programmatically.

fixes #2104

### DIFF
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -407,6 +407,7 @@ export default class Form extends Component {
     if (this.formElement) {
       this.formElement.dispatchEvent(
         new CustomEvent("submit", {
+          bubbles: true,
           cancelable: true,
         })
       );


epicfaace (commented):
@dreadheadsic apologies for the delay. I'm having trouble understanding why the changes in the way [react 17 attaches event listeners](https://reactjs.org/blog/2020/10/20/react-v17.html#changes-to-event-delegation) makes the onSubmit event no longer bubble, as well as why onSubmit not bubbling prevents users from [submitting the form programmatically](https://github.com/rjsf-team/react-jsonschema-form/blob/ec5447130c6e3a365b14a7805613b9328650fd70/docs/advanced-customization/internals.md). Could you explain?

I'm also wondering if we can add a test to ensure we catch this issue with React 17.

epicfaace:
Specifically, I don't understand why we need to bubble the onSubmit event beyond `this.formElement`, when React 17 attaches form listeners on the element directly (`this.formElement`) anyway.

dreadheadsic:
@epicfaace no worries. You know what, when i gave it a second thought, it didn't made sense to me either, why sudden need to bubble the event up, when react just changed where it attaches handlers (used  to be on document, now it does on root node). So i went through that react blog post and figured, this change is targeted towards running two different react versions - `Thanks to this change, it is now safer to embed a React tree managed by one version inside a tree managed by a different React version.`
So what was the issue in my case? I've bumped react to 17.0 in my package, but @rjsf/core still depends on v16, so i had two react versions running. I've tested this locally by bumping react dep in @rjsf/core to v17 and voila, everything worked without changes in bubbling.
To conclude, @rjsf works fine with react 17 and this PR is irrelevant, you can throw it away :)

epicfaace:
Great, thanks for looking into this!

robertedjones:
I have forked the repo and changed the peer dependency to 17 but the issue is not fixed for me. This PR however does fix it.

PR #2123, opened 2020-11-09
Set schema errors to empty to clear schema errors when re-rendered

saimanoharnethi:

fixes https://github.com/rjsf-team/react-jsonschema-form/issues/2102

### Checklist

### DIFF
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -374,13 +374,14 @@ export default class Form extends Component {
       errors = [];
     }
 
-    this.setState({ 
-      formData: newFormData,
-      errors: errors,
-      errorSchema: errorSchema,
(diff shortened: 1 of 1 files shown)


epicfaace (changes requested):
Thanks for the change @saimanoharnethi, and thanks for reviewing, @newt10 -- one thing is that it would be great if you could add a test that tests the specific issue in https://github.com/rjsf-team/react-jsonschema-form/issues/2102 so we can make sure this issue doesn't regress in the future.

newt10:
@epicfaace , I am closing this in favor of #2536 . I have added tests to #2536 to prevent future regressions. Since the author has not gotten a chance to update the PR, we can go ahead with #2536 for addressing the issue.

Issue #1645, opened 2020-03-11
Is it possible to implement custom array field with multiple values?


ERRORthenBSOD:
Hello, first of all great thanks for your project. I already been working with rjsf a lot and read the docs from the beginning and to the the end several times and i already broke my brain trying to implement custom array with multiply values.
I want to set my array to be like [material-ui autocomplete Multiple values
](https://codesandbox.io/s/fcogc) [Material Design link](https://material-ui.com/components/autocomplete/#multiple-values) .
i have scheme like: 
```
const JSONSchema = {
  "type": "array",
  "items": {
      "type": "string"
    }
}
```
```
const formData = ["foo"]
```
it is array of strings.
Problem that if i set for example custom widget with material ui autocomplete on items, i can`t create new index value with onChange event because it always rewrites the first index in array when i add new value.

It looks like i should create new item on "target values" when onChange event fires there, and write new value in it immediately but it is impossible. Also i can`t set enum there because my select values are dynamic from api.

My current workaround is that i write in single string all values with space between (like ["foo bar"]) when material-ui autocomplte input onChange event fires and then when i submit my form i split them by space and write all values like `["foo", "bar"]`

Maybe you can give me an advice how i can implement multiply values on same array input without creating new items in GUI?
Thanks in advance

videni:
I also have this trouble.

I use a single string that has values seperated by comma, like `1,2`. it is implemented with material-ui autocomplete component.the problem is when you  do update, the formData we got from server might looks like following. 

```
{
    "formData": {
        "channels": '1,2'
    },
    "schema": {
        "title": "product",
        "type": "object",
        "properties": {
            "channels": {
                "title": "Chanels",
                "type": [
                    "string",
                    "null"
                ]
            }
        }
    },
    "uiSchema": {
        "channels": {
            "ui:widget": "autocomplete",
            "ui:options": {
                "placeholder": "acme.channel.choose_channel",
                "allowClear": true,
                "minimumInputLength": 0,
                "multiple": true,
                "route_parameters": {
                    "_org_id": 2
                },
                "autocomplete_alias": "channels",
                "properties": [
                    "name",
                    "hostname"
                ],
                "route_name": "acme_form_autocomplete_search",
                "component": "autocomplete",
                "option_label": "name",
                "option_value": "id"
            }
        }
    }
}
```
Assume we have following options. how can we show the selected label? 
```
[
    {
        "id": 1,
        "name": "test1"
    },
    {
[...]

PR #1196, opened 2019-02-26
Allow choosing a seperate widget for MultiSelect rather than just Select

bockelie93:
### Reasons for making this change

Allows you to override just the MultiSelectWidget without overriding SelectWidget. Use-case is that you would like to render all your multi-choice arrays as checkboxes rather than a multiselect without having to provide an UiSchema for each of your values.

i.e. 
const widgets = { MultiSelectWidget: CheckBoxesWidget };

Fixes https://github.com/mozilla-services/react-jsonschema-form/issues/1193

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [X] **I'm adding or updating code**
  - [X] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
  - [ ] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style
* [X] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #2125, opened 2020-11-11
Render a select widget for arrays if specified in the ui schema

sradevski:

Currently it is impossible to render a multiselect for an array that doesn't have an enum. This is useful when the array is open-ended and can contain any strings provided in a multiselect input. It also makes it more consistent as a behavior when compared with how it is decided whether it is a files widget or not.

Related tickets:
https://github.com/rjsf-team/react-jsonschema-form/pull/1196
https://github.com/rjsf-team/react-jsonschema-form/issues/774
https://github.com/rjsf-team/react-jsonschema-form/issues/939
https://github.com/rjsf-team/react-jsonschema-form/issues/1645
https://github.com/rjsf-team/react-jsonschema-form/issues/1570
https://github.com/rjsf-team/react-jsonschema-form/issues/1385

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -272,7 +272,8 @@ function computeDefaults(
         });
       }
       if (schema.minItems) {
-        if (!isMultiSelect(schema, rootSchema)) {
+        // The UI schema is not important when calculating the defaults, so we just pass a default empty object.
+        if (!isMultiSelect(schema, {}, rootSchema)) {
           const defaultsLength = defaults ? defaults.length : 0;
           if (schema.minItems > defaultsLength) {
             const defaultEntries = defaults || [];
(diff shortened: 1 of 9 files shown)


epicfaace (commented):
thanks -- can you add a test to Forms_test that tests this behavior as well?

packages/core/src/utils.js
```diff
@@ -528,7 +529,10 @@ export function isSelect(_schema, rootSchema = {}) {
   return false;
 }
 
-export function isMultiSelect(schema, rootSchema = {}) {
+export function isMultiSelect(schema, uiSchema, rootSchema = {}) {
```
AbigailFernandes, line 532: Why are we restricting this to only use the "select" widget? Can we remove that restriction and leave it up to the user. The reason being, it could be possible, that the customWidget for `select` handles only single select and one could have a separate widget for multi select.

heath-freenome:
@sradevski Sorry for the long delay in getting to the PR... Unfortunately, the v5 beta refactored the `utils.js` file into its own `@rjsf/utils` package, converted the `core` package to Typescript. On the plus side, the changes made in other themes beside `core` are now unnecessary as the `ArrayFieldTemplates` were simplified so only the change in core needs to happen. Are you able to update the v5 beta with these changes?

heath-freenome:
I believe that #2697 actually fixes this issue more elegantly and more broadly

PR #2127, opened 2020-11-12
fix: MultiSchemaField not getting correct idSchema

QingqiShi:
Fixes #2127

When MultiSchemaField is part of an array the idSchema is incorrectly generated. This commit fixes that by making sure the parent id is passed to `toIdSchema`.

### Checklist


packages/core/test/oneOf_test.js
```diff
@@ -456,7 +456,8 @@ describe("oneOf", () => {
 
     expect($select.value).eql("1");
 
-    Simulate.change(node.querySelector("input#root_bar"), {
+    console.log(node.querySelector("#root_items_bar").type);
```
Jlabroy, line 459: Remove console.log

QingqiShi: removed!

QingqiShi:
🙏  @epicfaace could you advice next steps, thanks!

PR #2156, opened 2020-12-03
Pass oneOf / anyOf subschemas to altSchema optionsList

### LINKED ISSUE #2147: Allow  extra properties on enumOptions
### Prerequisites

### Description

Enum options, from an anyOf schema populate a label and value, but other properties, like description are not preserved. For situations with checkbox/radio selects, descriptions might be useful for either a tooltip with a custom widget, or for extra description about an option

### Steps to Reproduce

1. Make a schema like

```js
schema: {
  type: 'string',
  anyOf: [
    { title: 'Title1', const: 'value1', description: 'a description 1' }, 
    { title: 'Title2', const: 'value2', description: 'a description 2' },
  ]
}

uiSchema: {
  'ui:widget': 'radio'
}
```

#### Expected behavior

Descriptions should be carried through to the widget

#### Actual behavior

in the widge

```
const { enumOptions } = options
=> [
    { label: 'Title1', value: 'value1' }, 
    { label: 'Title2', value: 'value2' },
  ]
```

this looks to be from 

```
export function optionsList(schema) {
  if (schema.enum) {
    return schema.enum.map((value, i) => {
      const label = (schema.enumNames && schema.enumNames[i]) || String(value);
      return { label, value };
    });
  } else {
    const altSchemas = schema.oneOf || schema.anyOf;
    return altSchemas.map((schema, i) => {
      const value = toConstant(schema);
      const label = schema.title || String(value);
      return { label, value };
    });
  }
}
```

which maps the altSchemas, and only maintains the value and labels, without other properties on th schemas.  I'm not sure what other properties might make sense, but at least description seems possibly warranted here.  Happy to make a PR for that that change, or one that spread the altSchema as well?

### Version 2.4.0

mattcosta7:
fix #2147

To pass an optional description through to the enum options

### Checklist


mattcosta7:
Anything necessary to move this along?

Looks like the checks here are just deployment steps?

epicfaace (commented):
Thanks for the PR -- would you be able to add a test to ensure this functionality doesn't regress in the future?

mattcosta7:
@epicfaace tests for optionsList added, let me know if there's anything else!

mattcosta7:
Just a friendly ping to see if there's anything else I can do to move this along? @epicfaace

packages/core/test/utils_test.js
```diff
@@ -3763,4 +3764,68 @@ describe("utils", () => {
...
+        enumSchema.enum.map(opt => ({ label: opt, value: opt }))
+      );
+      expect(optionsList(enumNameSchema)).eql(
+        enumNameSchema.enum.map((opt, index) => {
+          const label = enumNameSchema.enumNames[index] || opt;
+          return { label: label, value: opt };
+        })
+      );
+    });
+    it("shoulld generate options for a oneOf|anyOf schema", () => {
```
epicfaace, line 3789:
```suggestion
    it("should generate options for a oneOf|anyOf schema", () => {
```

epicfaace (commented):
Looks good. A few questions:
- I see that you suggested passing the entire altSchema instead of just `description` in the issue. I think this makes sense; could you make that change in this PR too?
- Can you also update the documentation to discuss the `enumOptions` property in more detail, along with this change? Currently we have some outdated description in the docs and a reference to v0.35.0 (which can probably be removed): https://github.com/rjsf-team/react-jsonschema-form/blob/bf6c99862afe9b381c9be4027ad97adbeabc1bd5/docs/advanced-customization/custom-widgets-fields.md#L2

packages/core/test/utils_test.js
mattcosta7: fixed this typo

packages/core/src/utils.js
```diff
@@ -573,7 +573,11 @@ export function optionsList(schema) {
     return altSchemas.map((schema, i) => {
       const value = toConstant(schema);
       const label = schema.title || String(value);
-      return { label, value };
+      return {
+        ...schema,
```
mattcosta7, line 577: spread optionsList

packages/core/index.d.ts
```diff
@@ -347,7 +347,10 @@ declare module '@rjsf/core' {
 
         export function allowAdditionalItems(schema: JSONSchema7): boolean;
 
-        export function optionsList(schema: JSONSchema7): { label: string; value: string }[];
+        export function optionsList(schema: JSONSchema7): (Partial<JSONSchema7Definition> & {
+            label: string;
+            value: string;
+        })[];
```
mattcosta7, line 353: added the Partial<JSONSchema7Definition
> here to ensure the types show other properties may be optionally passed here, but label and value are guaranteed

mattcosta7:
@epicfaace updated with the requested changes, thansk!

docs/advanced-customization/custom-widgets-fields.md
```diff
@@ -133,11 +133,10 @@ The following props are passed to custom widget components:
 - `onBlur`: The input blur event handler; call it with the the widget id and value;
 - `onFocus`: The input focus event handler; call it with the the widget id and value;
 - `options`: A map of options passed as a prop to the component (see [Custom widget options](#custom-widget-options)).
+- `options.enumOptions`: For enum fields, this property contains the list of options for the enum as an array of { label, value } objects. If the enum is defined using the oneOf/anyOf syntax, the entire schema object is spread onto the { ...schema.\[anyOf|oneOf\], label, value } object.
```
epicfaace, line 136:
```suggestion
- `options.enumOptions`: For enum fields, this property contains the list of options for the enum as an array of { label, value } objects. If the enum is defined using the oneOf/anyOf syntax, the entire schema object for each option is spread onto the { ...schema.\[anyOf|oneOf\], label, value } object.
```

docs/advanced-customization/custom-widgets-fields.md
```diff
@@ -133,11 +133,10 @@ The following props are passed to custom widget components:
 - `onBlur`: The input blur event handler; call it with the the widget id and value;
 - `onFocus`: The input focus event handler; call it with the the widget id and value;
 - `options`: A map of options passed as a prop to the component (see [Custom widget options](#custom-widget-options)).
+- `options.enumOptions`: For enum fields, this property contains the list of options for the enum as an array of { label, value } objects. If the enum is defined using the oneOf/anyOf syntax, the entire schema object for each option is spread onto the { ...schema.\[anyOf|oneOf\], label, value } object.
```
epicfaace: We should clarify that the schema is included regardless of whether it's defined using the oneOf/anyOf syntax.

packages/core/src/utils.js
epicfaace, line 577: Actually, I was just thinking it might be better to actually explicitly include the schema in a `schema` keyword. I know that `label` and `value` currently aren't reserved keywords in JSON Schema, but this way, we can prevent further conflicts in case it happens in the future, and don't have to worry about `label` and `value` potentially overriding something within the schema. What do you think?

```suggestion
        schema,
```

docs/advanced-customization/custom-widgets-fields.md
mattcosta7: When an enum is declared like

```
{ type: string, enum: [one, two, three] }
```

the options list currently doesn't have a schema defined, they are instead just { label, value } objects, where enumNames might override the label.   

only in cases where not using they enum key would the schema be added.  I think those 2 cases are oneOf/anyOf?   are there other times that's valid?

packages/core/src/utils.js
mattcosta7: agreed here, this makes sense to me, pushed a commit with this change made

docs/advanced-customization/custom-widgets-fields.md
mattcosta7: Maybe something along the lines of 

For enum fields defined using the enum property, this contains a list of options as an array of { label, value } objects.  For other enums, such as those defined with an anyOf or oneOf schema, this contains a list of options as an array of { schema, label, value } objects, where the schema is the schema defining each enum option individually

mattcosta7: On the other hand, we could also pass the base schema for the 

```
{ type: string, enum: [one, two, three] }
```

to the options themselves too, so this might generate

```
{
   schema: { type: string, enum: [one, two, three] },
   label: one,
   value: one
} 
```

for these items as well.  This might make sense since the enum options each inherit that schema?

mattcosta7: actually, nevermind that, I think it might cause more complexity than it would aim to solve, 

for instance, a 'description' on the parent schema would then automatically apply to the child enums, when it should not - so probably better to avoid adding the schema, or only copy specific schema properties in that case (ie, just type and _maybe_ format, but no others?)

epicfaace: Yeah, let's just not pass the base schema for simple (non-oneOf/anyOf enums) instead.

epicfaace (approved):

PR #2175, opened 2020-12-20
Properly hide rows when ui:widget is hidden (bootstrap-4)

pjb7687:

When `ui:widget` is `hidden`, add class `d-none` to `Row` so that the corresponding row is properly hided. This fix is for `bootstrap-4`, but basically the other themes can also be patched in the same way.

### Checklist

### DIFF
--- a/packages/bootstrap-4/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
+++ b/packages/bootstrap-4/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
@@ -33,7 +33,7 @@ const ObjectFieldTemplate = ({
       )}
       <Container fluid className="p-0">
         {properties.map((element: any, index: number) => (
-          <Row key={index} style={{ marginBottom: "10px" }}>
+          <Row key={index} style={{ marginBottom: "10px" }} className={(element.content.uiSchema["ui:widget"] === "hidden")?"d-none":undefined} >
             <Col xs={12}> {element.content}</Col>
           </Row>
         ))}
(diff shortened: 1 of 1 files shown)


pjb7687:
Related to #1683.

packages/bootstrap-4/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
```diff
@@ -33,7 +33,7 @@ const ObjectFieldTemplate = ({
       )}
       <Container fluid className="p-0">
         {properties.map((element: any, index: number) => (
-          <Row key={index} style={{ marginBottom: "10px" }}>
+          <Row key={index} style={{ marginBottom: "10px" }} className={(element.content.props.uiSchema["ui:widget"] === "hidden")?"d-none":undefined} >
```
epicfaace, line 36: Hmm, the analogous code in @rjsf/core just says:

https://github.com/rjsf-team/react-jsonschema-form/blob/a01b9e33d58b98cf7b10ce44c744cc3d420fb6e7/packages/core/src/components/fields/ObjectField.js#L32

When ui:widget is hidden, is `element.content` equal to null / undefined? I'm trying to see if we can achieve the same effect without having to explicitly check for `ui:widget` equal to `hidden`, such as something like the following:

```
{element.content && <Row>...</Row>}
```

pjb7687: Hi,
1. `element.content` exists for the hidden fields.
2. The row object contains the hidden input element, and removing the entire row means no hidden input element in the form.

Morriz:
bump

epicfaace:
Thanks for your PR @pjb7687 ! Merging this into https://github.com/rjsf-team/react-jsonschema-form/pull/2405.

PR #2181, opened 2020-12-28
Fix playground by upgrading react-bootstrap

epicfaace:
Otherwise, the playground doesn't run locally, since @rjsf/bootstrap-4 requires react bootstrap 1+ as a peer dep.


stanlemon (approved):

epicfaace:
I don't think this is working -- and the playground seems to be running fine for now, so I'll just close this for now.

PR #2195, opened 2021-01-16
Fix regular expression for number field cached formData value (#2157)

### LINKED ISSUE #2157: Changing a number field from 11 to 1 will fail to render the new value
In our project we were using a form that contained a number input, we noticed that if we changed the form value from 11 to 1 the field would continue to show 11 when the formData held 1

The pattern was confirmed with 420 and 20, or 17 and 7, 1234 and 234 ...

The error lies in the following line of code:
https://github.com/rjsf-team/react-jsonschema-form/blob/c9d0569a3d5bf527cefd16b0f3f3cda3d96b6f39/packages/core/src/components/fields/NumberField.js#L76

If you change it to:
```
const re = new RegExp(`^${value}`.replace(".", "\\.") + "\\.?0*$");
```
the field will render correctly when the value changes

(note the `"^"` added to the beginning of the regex so that it has to match the full string value, not just the right side of the value)

jeffothy:
Use a ^ to ensure regex matches the entire new value and not just the
end of the value.

Normalize the cache value with a leading zero when it starts with a
decimal character as is done in handleChange.

Co-authored-by: Landon Bradshaw <landon.bradshaw@gmail.com>

Fixes #2157

### Checklist

### DIFF
--- a/packages/core/src/components/fields/NumberField.js
+++ b/packages/core/src/components/fields/NumberField.js
@@ -73,11 +73,17 @@ class NumberField extends React.Component {
       // Construct a regular expression that checks for a string that consists
       // of the formData value suffixed with zero or one '.' characters and zero
       // or more '0' characters
-      const re = new RegExp(`${value}`.replace(".", "\\.") + "\\.?0*$");
+      const re = new RegExp("^" + `${value}`.replace(".", "\\.") + "\\.?0*$");
 
-      // If the cached "lastValue" is a match, use that instead of the formData
-      // value to prevent the input value from changing in the UI
-      if (lastValue.match(re)) {
+      let normalizedLastValue = `${lastValue}`;
+      if (`${lastValue}`.charAt(0) === ".") {
+        normalizedLastValue = `0${lastValue}`;
+      }
+
+      // If the normalized cached "lastValue" is a match, use the "lastValue"
+      // instead of the formData value to prevent the input value from changing
+      // in the UI
+      if (normalizedLastValue.match(re)) {
         value = lastValue;
       }
     }
--- a/packages/core/test/NumberField_test.js
+++ b/packages/core/test/NumberField_test.js
@@ -299,6 +299,26 @@ describe("NumberField", () => {
         expect($input.value).eql(".00");
       });
 
+      it("should not used cached value when substring matches", () => {
+        const schema = {
+          type: "number",
+        };
+        const { comp, node } = createFormComponent({
+          schema,
+          formData: 11,
+        });
+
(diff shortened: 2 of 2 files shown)


epicfaace (commented):
Thanks -- can you specifically add a test that replicates the scenario mentioned in #2157?

jeffothy:
@epicfaace As suggested, I have added a test case which replicates the scenario in #2157 and I have verified it passes with this fix and fails without. Please take a look.

jeffothy:
Hi @epicfaace sorry for the nag, but can I support in any way to get this one in?

epicfaace (commented):
@jeffothy thanks for bumping this. I understand your first change of adding `^` to the regexp, but could you explain a bit further what the normalizedLastValue code is for? Is there a specific scenario in which an error would occur without the normalizedLastValue fix? (in that case, it might be good to add a test to cover it)

jeffothy:
Thanks @epicfaace Right, I can't actually come up with a scenario where that error would occur, so I'll remove that part of the commit.

epicfaace (approved):
Great, this looks good then! Do you know why the tests are failing -- could you look into it? Once the tests are passing, we should be ready to merge.

jeffothy:
@epicfaace Oh goodness, my bad (only ran my new test on the recheck). These existing tests fail since they need the normalized cache changes that I just removed. Shall I reset back to https://github.com/rjsf-team/react-jsonschema-form/commit/04b1c1d1fd7b186bd393dd877d25add118c30f83 ?

epicfaace:
Ah, I see. Sure, let's revert, though could you explain a bit more what exactly normalizedLastValue is for? I'm having some trouble understanding.

jeffothy:
@epicfaace Yes. The line `if (`${value}`.charAt(0) === ".") {` within the handleChange function prepends a leading `0` in the case where the value starts with a decimal point `.`. Now, since the cached state (`lastValue`) is used to determine whether or not to update the UI, the same leading `0` is added to it ensure the cache is used.

The TC "should normalize values beginning with a decimal point" is the example to look at.
Target input is changed to ".00"
Value is 0 (due to the processing normalization in handlechange).
But .00 won't match /^0\.?0*$/ therefore the cache will miss.
So if we compare "0.00" ("0" + ".00") with the regex instead we will hit the cache.

I have updated the branch now.

Parnswir:
I would love to see this get released. This issue has created one of the most irritating bugs in my current project, where the reset button for a field did not work if you changed "1" to "11", but did work perfectly for "12" 🙃

PR #2198, opened 2021-01-18
fix: oneOf anyOf default value always using first option

### LINKED ISSUE #2183: oneOf with "object" types - impossible to replace Option 1
### Prerequisites

### Description
When using a json schema with `oneOf` where the possible types are `object`, Option 1 properties cannot be removed - when choosing a different option, Option 1 properties remain in the json output. See [Playground reproducible](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsb3JlbSI6e319LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsIm9uZU9mIjpbeyJhZGRpdGlvbmFsUHJvcGVydGllcyI6ZmFsc2UsInByb3BlcnRpZXMiOnsibG9yZW0iOnsidHlwZSI6Im9iamVjdCJ9fX0seyJhZGRpdGlvbmFsUHJvcGVydGllcyI6ZmFsc2UsInByb3BlcnRpZXMiOnsiaXBzdW0iOnsidHlwZSI6Im9iamVjdCJ9fX0seyJhZGRpdGlvbmFsUHJvcGVydGllcyI6ZmFsc2UsInByb3BlcnRpZXMiOnsicHlvdCI6eyJ0eXBlIjoib2JqZWN0In19fV19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6dHJ1ZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==)

### Steps to Reproduce / Actual behaviour

1. Create a json schema with oneOf with object types
2. Switch to Option 2
3. Option 2 properties added, Option 1 properties remain in the json output
4. Switch to Option 3
5. Option 3 properties added, Option 2 properties deleted (as they should), Option 1 properties remain

#### Expected behavior

1. Create a json schema with oneOf with object types
2. Switch to Option 2
3. Option 2 properties added, Option 1 properties deleted
4. Switch to Option 3
5. Option 3 properties added, Option 2 properties deleted

Correct behaviour can be found in the json-editor playground [here](https://json-editor.github.io/json-editor/?data=N4Ig9gDgLglmB2BnEAuUMDGCA2MBGqIAZglAIYDuApomALZUCsIANOHgFZUZQD62ZAJ5gArlELwwAJzplsrEIgwALKrNSgogiFUJhO3cWwRUA8kVQBtUGQAmtmLARyAClMhUpsGqiJzEVGwQ7jpeMD5oINjSahogWjp6BjwgAL7pLDb2jnDwriGe3sgoftgBQQVhEegQiCJ0cQm6KOxcKempmSB2Dk552G4eVcWl5SDBQ0VxEMLikU1JbeIdALqdispgFLye7lLFIDDwUJ5kPLkKUKoMhHhgYFCIUFJkEAAsaalAA===
)

### Version

Current playground versioon

QingqiShi:
Fixes #2183 

When generating default values, currently we're always assuming the first option is matched. This default value will then be merged with existing data - so if you have the second option selected, you can end up with both values from the first and the second options.

https://github.com/rjsf-team/react-jsonschema-form/issues/2183

### Checklist


Canas:
Whatever happened to this fix? This is still happening in the playground

QingqiShi:
Still waiting for a review and merge.

exevil:
+1 here. Would be great to merge this.

epicfaace (commented):
Can you add a test that specifically covers [this playground example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsb3JlbSI6e319LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsIm9uZU9mIjpbeyJhZGRpdGlvbmFsUHJvcGVydGllcyI6ZmFsc2UsInByb3BlcnRpZXMiOnsibG9yZW0iOnsidHlwZSI6Im9iamVjdCJ9fX0seyJhZGRpdGlvbmFsUHJvcGVydGllcyI6ZmFsc2UsInByb3BlcnRpZXMiOnsiaXBzdW0iOnsidHlwZSI6Im9iamVjdCJ9fX0seyJhZGRpdGlvbmFsUHJvcGVydGllcyI6ZmFsc2UsInByb3BlcnRpZXMiOnsicHlvdCI6eyJ0eXBlIjoib2JqZWN0In19fV19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6dHJ1ZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==) (from #2183)? Meaning, the case in which neither option has a default, and when you switch from option 1 to option 2.

chenghu3:
I think @QingqiShi addressed the comment, could we pls review and merge this? thanks

QingqiShi:
Yes indeed. I added a test to cover the example from #2183

tappleby:
Just been tracking down some issues with an upgrade and found this fix. I noticed that for any schemas that make use of refs, this fix may also depend on #2272

QingqiShi:
Need some help with this one. Build is failing after merging with `master`. It seems to be an unrelated test that's failing.

epicfaace:
@QingqiShi looks like [this is the line that is failing](https://github.com/rjsf-team/react-jsonschema-form/blob/5e52ae8dd627c4b5e9378e2d370ebfc812c3e4f0/packages/core/test/oneOf_test.js#L79). Maybe you could try debugging the test by printing out `onChange` and `onChange.lastCall`? I'm wondering if perhaps `onChange` isn't being called properly with the changes in this PR.

QingqiShi:
I realised that after the fix it no longer generate `formData` if you haven't selected any of the options in a `oneOf` or `anyOf` yet.

To make this fix backwards compatible, I've added a check to see if `formData` is an empty object, if it is then continue to pass `undefined` to `getMatchingOption` when generating defaults. This maintains the current behaviour before the user selects an option, and will start using `formData` after the user selects an option.

@epicfaace tests fixed, please instruct on how to proceed.

packages/core/src/utils.js
```diff
@@ -220,10 +220,22 @@ function computeDefaults(
...
+          schema.oneOf,
+          rootSchema
+        )
+      ];
   } else if ("anyOf" in schema) {
     schema =
-      schema.anyOf[getMatchingOption(undefined, schema.anyOf, rootSchema)];
+      schema.anyOf[
+        getMatchingOption(
+          JSON.stringify(formData) !== "{}" ? formData : undefined,
```
epicfaace, line 234: We should use the lodash function `isEmpty` instead, as it will make calculation faster (it could be slow to run JSON.stringify on the entire formData).

epicfaace, line 234: Also, should we move this logic out of this file and directly inside the `getMatchingOption` function instead?

Finally, `getMatchingOption` is also [used a few times in MultiSchemaField.js](https://github.com/rjsf-team/react-jsonschema-form/blob/3ec17f1c0ff40401b7a99c5e9891ac2834a1e73f/packages/core/src/components/fields/MultiSchemaField.js#L21-L33) -- should we apply this fix there as well, and if we do, perhaps we could add a test that covers that change as well?

epicfaace (commented):
Thanks for your changes! A few more comments.

packages/core/src/utils.js
QingqiShi: I'm a bit scared to widen this change.

I know that `getMatchingOption` is used for a couple of different purposes, such as generating defaults (as we're doing here), and for loading the initial selected multi-schema options (as you mentioned in `MultiSchemaField`), we also use it in `resolveDependencies`.

These use cases have subtle differences - in the `computeDefault` function, `formData` gets defaulted to `{}` (for reasons I don't understand), and that's why we need this check here so it goes back to generating defaults. Whereas in other places it may be necessary to keep the empty object as the object type may be matched to a particular sub-schema.

QingqiShi: If I may, I will restrain myself from potentially breaking the world. I'll leave it to your judgement as you are the expert on this.

QingqiShi: Replaced with lodash `isEmpty`

Stankk:
Hello, any news on the merge ?

Akifhan:
Hello, any updates on the merge?

davidpett:
Can we get this merged, please?

eliyaoo32:
Hi, are they plans to merge this?

packages/core/src/utils.js
logancool: @epicfaace is this a blocking comment? It would be great to get this pr merged.

logancool: cc @nickgros (I see you on an active pr)

heath-freenome:
@QingqiShi Do you have the bandwidth to refactor this code into the new `@rjsf/utils` package in the v5 beta?

heath-freenome:
@QingqiShi closing as I took your excellent work and refactored it into the new `@rjsf/utils` library code in https://github.com/rjsf-team/react-jsonschema-form/pull/3099

QingqiShi:
Thank you very much ❤️ @heath-freenome

PR #2199, opened 2021-01-18
fix: MultiSchema getting incorrect id in arrays

### LINKED ISSUE #2197: MultiSchemaField getting incorrect id when inside array
### Prerequisites

### Description

When MultiSchemaField is part of an array the idSchema is incorrectly generated. This can cause duplicated ids.

### Steps to Reproduce

1. Check `packages/core/test/anyOf_test.js`
2. Find the test `should correctly render form inputs for anyOf inside array items`
3. Currently we're expecting `input#root_foo` to exist, but that's not the right id

#### Expected behavior

The idSchema ids should maintain the full id even if they're inside arrays.

For example, given a schema 

```js
const schema = {
  type: "object",
  properties: {
    items: {
      type: "array",
      items: {
        type: "object",
        anyOf: [
          {
            properties: {
              foo: {
                type: "string",
              },
            },
          },
          {
            properties: {
              bar: {
                type: "string",
              },
            },
          },
        ],
      },
    },
  },
};
```

The id for the `foo` field within each array item should be `root_items_0_foo`, `root_items_1_foo` etc.

#### Actual behavior

For the example above, the generated id is `root_foo` for all array items.

### Version

2.4.1

QingqiShi:
Fixes #2197

When MultiSchemaField is part of an array the idSchema is incorrectly generated. This commit fixes that by making sure the parent id is passed to `toIdSchema`.

#2197

### Checklist

### DIFF
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -243,7 +243,7 @@ function SchemaFieldRender(props) {
   let idSchema = props.idSchema;
   const schema = retrieveSchema(props.schema, rootSchema, formData);
   idSchema = mergeObjects(
-    toIdSchema(schema, null, rootSchema, formData, idPrefix),
+    toIdSchema(schema, idSchema.$id, rootSchema, formData, idPrefix),
     idSchema
   );
   const FieldComponent = getFieldComponent(schema, uiSchema, idSchema, fields);
--- a/docs/3.x upgrade guide.md
+++ b/docs/3.x upgrade guide.md
@@ -26,29 +26,29 @@ Now any option with a reference will be resolved/dereferenced when given as prop
 
 ### Generate correct ids when arrays are combined with `anyOf`/`oneOf`
 
-Previously when you have arrays inside `anyOf` or `oneOf`, the parent isn't
-used to generate ids. For example, given a schema such as
+In v2, with arrays inside `anyOf` or `oneOf`, the parent name
+was not used to generate ids. For example, given a schema such as
 
-```
+```json
 {
-  type: "object",
-  properties: {
-    items: {
-      type: "array",
-      items: {
-        type: "object",
-        anyOf: [
+  "type": "object",
+  "properties": {
+    "items": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "anyOf": [
           {
-            properties: {
-              foo: {
-                type: "string",
+            "properties": {
+              "foo": {
+                "type": "string",
               },
             },
           },
           {
-            properties: {
-              bar: {
-                type: "string",
+            "properties": {
+              "bar": {
+                "type": "string",
               },
             },
           },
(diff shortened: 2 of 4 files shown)


epicfaace:
Hey @QingqiShi , sorry for the delay in reviewing. Is this PR essentially the same as the earlier PR #2199?

QingqiShi:
@epicfaace Yes! I accidentally force pushed the other branch which destroyed the PR, this one is identical.

QingqiShi:
Need some help with this one. Build is failing after rebasing to the latest `master` branch. It seems to be an unrelated test.

epicfaace:
This is technically a breaking change (though correct behavior), and so it can go into the 3.x release. @QingqiShi can you add a section to the "3.x upgrade guide" in the documentation that describes what changed and what users should watch out for?

epicfaace:
@QingqiShi It looks like [the failing test](https://github.com/rjsf-team/react-jsonschema-form/blob/3ec17f1c0ff40401b7a99c5e9891ac2834a1e73f/packages/core/test/oneOf_test.js#L770-L772) is happening because the old test is specifically depending on the previous behavior, that is, four dropdowns with the same id in arrays. Can you update the test so it works with this PR's changes with respect to ids?

QingqiShi:
@epicfaace Thank you for pointing me to the failing test!

The failing test is now fixed, and I've added a section to the upgrade guide as you've suggested. Thanks!

packages/core/src/components/fields/SchemaField.js
```diff
@@ -243,7 +243,7 @@ function SchemaFieldRender(props) {
   let idSchema = props.idSchema;
   const schema = retrieveSchema(props.schema, rootSchema, formData);
   idSchema = mergeObjects(
-    toIdSchema(schema, null, rootSchema, formData, idPrefix),
+    toIdSchema(schema, idSchema.$id, rootSchema, formData, idPrefix),
```
epicfaace, line 246: I'm a bit confused as to what exactly this change does. Do you mind explaining a bit as to how this change in the logic fixes the issue? @QingqiShi 

Depending on how complex it is, it might be good for us to add a comment so future users of this code can more easily understand, as well!

epicfaace (commented):
This looks all good to me, can you just explain one part of the PR to me? Thanks!

packages/core/src/components/fields/SchemaField.js
QingqiShi: Gladly.

However I'm not sure if we need much explaining here. The call signature for `toIdSchema` is `toIdSchema(schema, id, rootSchema, formData, idPrefix)`, so I believe passing `null` to the id argument is incorrect to begin with.

More specifically, in the `ArrayField` each item gets an `itemIdSchema` with the id `idSchema.$id + "_" + index`, this index then gets passed to the item through`SchemaField`. The logic after that is a bit fuzzy to me, but I imagine if we don't use it during the generation of new ids, it defaults to 'root'  prefix and this wrong id eventually get passed to the `MultiSchema` component. 

🙏 Thanks again for spending the time to maintain this package.

epicfaace: I see, thanks for the explanation! I'm still a bit confused as to how this change would have only affected the ids in arrays in MultiSchemaField. I would have expected that such a change to SchemaField.js would have had a significant effect on _all_ / many more ids generated through idSchema. Do you know why that's not the case?

heath-freenome:
@QingqiShi So I'd love to get this change into the upcoming v5 beta... do you have the bandwidth to port this to the `SchemaField.tsx` file in the `rjsf-v5` branch in the next week or so?

heath-freenome:
Fixed in v5 beta via https://github.com/rjsf-team/react-jsonschema-form/pull/3039

PR #2217, opened 2021-02-02
Docs: add instructions for manual npm packaging tests after release

epicfaace:

I began doing this after there was an issue once with the wrong files published to the npm package / package.json configuration errors. Documenting this so it's easy to add to the releasing workflow.


stanlemon:
Was the semantic package the only other one with updates in the lock file?

heath-freenome:
@epicfaace Want to keep any of this?

PR #2222, opened 2021-02-05
Fix: Support integer type in bootstrap TextWidget

JonathanWbn:
The bootstrap `TextWidget` currently does not support the `integer` type.
When an `integer` type is used in the schema, an `<input type="integer" />` is rendered, which is an invalid `type` and leads to the input appearing as a **text** input rather than a **number** input.


abeMedia:
Any chance getting this merged? I'm already using this in a local fork and it works great!

heath-freenome:
I believe that this fix was done incidentally as part of the `BaseInputTemplate` work in the `rjsf-v5` branch.

JonathanWbn:
@heath-freenome Looks the same to me. Closing because obsolete.

PR #2226, opened 2021-02-08
fix: check for uiSchema on getDisplayLabel

jacqueswho:

utils getDisplayLabel does not check if uiSchema prop exists

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -400,7 +400,7 @@ export function getDisplayLabel(schema, uiSchema, rootSchema) {
   if (schema.type === "boolean" && !uiSchema["ui:widget"]) {
     displayLabel = false;
   }
-  if (uiSchema["ui:field"]) {
+  if (uiSchema && uiSchema["ui:field"]) {
     displayLabel = false;
   }
   return displayLabel;


epicfaace (commented):
When would uiSchema be a non-object (null / undefined / etc.)? I just tried setting uiSchema to `null`, and the form crashed in several other parts of the code (such as https://github.com/rjsf-team/react-jsonschema-form/blob/c33204e5df7f33e63a76bab878cb52a9307bb011/packages/core/src/components/Form.js#L117) even with this code fix.

epicfaace:
Specifically, I'm trying to think of a test we could add that would cover this case (i.e. which would fail without this PR, but would pass with this PR), but I can't think of one.

epicfaace:
Actually, it seems like uiSchema is always set by default to `{}` in `SchemaField`, so I don't think there's ever a case where it's equal to null / undefined -- https://github.com/rjsf-team/react-jsonschema-form/blob/21a64704dd13b73ed1a3b71060331ca895244442/packages/core/src/components/fields/SchemaField.js#L403

ErQrYfkrju:
@jacqueswho ping

jacqueswho:
> Actually, it seems like uiSchema is always set by default to `{}` in `SchemaField`, so I don't think there's ever a case where it's equal to null / undefined --
> 
> https://github.com/rjsf-team/react-jsonschema-form/blob/21a64704dd13b73ed1a3b71060331ca895244442/packages/core/src/components/fields/SchemaField.js#L403

@epicfaace found out I had to change select widget and checkbox widget to not use this util. But my fix does prevent it from crashing.

epicfaace:
Hmm, so do you mean that there's another fix we should add to rjsf to fix the widgets?

jacqueswho:
> Hmm, so do you mean that there's another fix we should add to rjsf to fix the widgets?

@epicfaace as long as getDisplayLabel from core/util is not used on CheckboxWidget or SelectWidget. As there uiSchema is null on these for the tests when run

PR #2243, opened 2021-02-21
Pass the onChange handler to the FieldTemplate

### LINKED ISSUE #2238: Provide onChange to FieldTemplate
### Prerequisites

### Description

I'd like to be able to apply a "Reset" button to any arbitrary field in a form. Because it could be applied to any field type, I think it'd make sense to do this as part of the `FieldTemplate` level instead of as a part of each field or widget type.

This would be possible if the `onChange` function was passed to the `FieldTemplate`. Would that be an acceptable change? If so, I'm happy to work on a PR.

TimothyBJacobs:

This allows for a `FieldTemplate` to trigger a change event for its managed field.

Fixes #2238.

### Checklist


packages/core/src/components/fields/SchemaField.js
```diff
@@ -322,6 +322,7 @@ function SchemaFieldRender(props) {
     id,
     label,
     hidden,
+    onChange: props.onChange,
```
epicfaace, line 325: Small thing: for consistency with the other props, can we pass in a variable called `onChange` that is set equal to `props.onChange`?

epicfaace (changes requested):

packages/core/src/components/fields/SchemaField.js
TimothyBJacobs: Sure thing! Updated in b32c659.

epicfaace (approved):

Yashsharma1911:
@TimothyBJacobs does this allow us to access onChange inside the array field template?

PR #2251, opened 2021-02-25
Use `ui:emptyValue` for SelectWidget instead of forcing to `undefined`

### LINKED ISSUE #1041: ui:emptyValue does not work with selects
### Prerequisites

### Description

`ui:emptyValue` is ignored for select widgets, requiring messy workarounds to resolve due to the way that the empty value option is injected.

The workaround that works is to specify a non-falsy empty value (e.g. `"0"`) as the default for the form input (required in order to prevent the empty `<option>` from being injected), add your own empty value to the enum / enumNames (e.g. `"0"` as the value, "None" as the name), then reprocess to treat that empty value appropriately in the code dealing with the form (if `"0"` is OK then that's fine, but if you want a falsy default, e.g. an empty string, more work is required).

Obviously this is much more complex than just specifying `ui:emptyValue` with the desired value.

### Steps to Reproduce

1. Specify an enum field
1. Specify it as a select field
1. Give it a `ui:emptyValue` setting
1. Select the empty option - the `ui:emptyValue` is not used

[Simple playground reproduction](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJwcm9wZXJ0aWVzIjp7ImlkIjp7ImRlc2NyaXB0aW9uIjoiQW4gb3B0aW9uYWwgYWdlbnQgZm9yIHRoaXMgZG9jdW1lbnQuIiwiZW51bSI6WyIxIiwiMiJdLCJlbnVtTmFtZXMiOlsiSmFuZSBBZ2VudCIsIkpvZSBBZ2VudCJdLCJ0aXRsZSI6IkFnZW50IiwidHlwZSI6InN0cmluZyJ9fSwidHlwZSI6Im9iamVjdCJ9LCJ1aVNjaGVtYSI6eyJpZCI6eyJ1aTplbXB0eVZhbHVlIjoiIiwidWk6cGxhY2Vob2xkZXIiOiJObyBhZ2VudCIsInVpOndpZGdldCI6InNlbGVjdCJ9fX0=)

#### Expected behavior

The formData value should be the `ui:emptyValue` when the empty option is selected.

#### Actual behavior

There is no formData value for the field when the empty option is selected.

### Version

Current playground version.

thom4parisot:

This is related to #1041.

It is actually making `SelectWidget` as described in the documentation here: https://react-jsonschema-form.readthedocs.io/en/latest/usage/validation/#the-case-of-empty-strings

### Checklist

fixes #1041

cc @mogztter

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -584,7 +584,7 @@ const nums = new Set(["number", "integer"]);
 /**
  * Turn a DOM event form value into its formData counterpart.
  *
- * It is especially used in `SelectWidget` widgets.
+ * It is currently only used in `SelectWidget` widgets.
  *
  * @example turn "24" into 24 if the field is numeric
  * @example turn the user choice of an "empty value" (`""`) into the `ui:emptyValue` alternative, `undefined` otherwise
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -578,6 +578,44 @@ export function optionsList(schema) {
   }
 }
 
+// "enum" is a reserved word, so only "type" and "items" can be destructured
+const nums = new Set(["number", "integer"]);
+
+/**
+ * Turn a DOM event form value into its formData counterpart.
+ *
+ * It is especially used in `SelectWidget` widgets.
+ *
+ * @example turn "24" into 24 if the field is numeric
+ * @example turn the user choice of an "empty value" (`""`) into the `ui:emptyValue` alternative, `undefined` otherwise
+ * @see https://react-jsonschema-form.readthedocs.io/en/latest/usage/validation/#the-case-of-empty-strings
+ */
+export function processNewValue({ schema, uiSchema, newValue }) {
+  const { type, items } = schema;
+
+  if (newValue === "") {
+    return undefined;
+  } else if (type === "array" && items && nums.has(items.type)) {
+    return newValue.map(asNumber);
+  } else if (type === "boolean") {
+    return newValue === "true";
+  } else if (type === "number") {
+    return asNumber(newValue);
+  }
+
+  // If type is undefined, but an enum is present, try and infer the type from
+  // the enum values
+  if (schema.enum) {
+    if (schema.enum.every(x => guessType(x) === "number")) {
+      return asNumber(newValue);
+    } else if (schema.enum.every(x => guessType(x) === "boolean")) {
+      return newValue === "true";
+    }
+  }
+
+  return newValue;
+}
+
 export function findSchemaDefinition($ref, rootSchema = {}) {
   const origRef = $ref;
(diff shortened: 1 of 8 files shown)


epicfaace:
Is this issue happening with other themes as well? If so, do you think we need a similar kind of fix for the other themes?

oncletom:
@epicfaace I have not tried but I think the logic is pretty much the same. I'll reflect the change in the other themes, and eventuelly check if I can write a unit test for it.

epicfaace:
Cool, thanks. If you think of a way we could maybe reuse this code across multiple themes, that would probably be ideal.

packages/semantic-ui/src/SelectWidget/SelectWidget.js
```diff
@@ -6,9 +6,7 @@ import { utils } from '@rjsf/core';
 import PropTypes from "prop-types";
 import { getSemanticProps } from "../util";
 
-const { asNumber, guessType } = utils;
-
-const nums = new Set(["number", "integer"]);
+const { processNewValue } = utils;
```
oncletom, line 9: It happens the `processValue()` function was not even implemented in the `semantic-ui` widget. But maybe for a reason?

oncletom:
@epicfaace I've done my best to ensure the commits made sense. I started by factoring the function out of the widgets, writing a valid test. Then I DDD'ed the implementation change.

I hope it makes sense.

packages/semantic-ui/src/SelectWidget/SelectWidget.js
epicfaace: What do you mean? It seems like `processValue` was implemented in the semantic-ui widget -- see the diff for this file.

packages/core/src/utils.js
```diff
@@ -578,6 +578,44 @@ export function optionsList(schema) {
   }
 }
 
+// "enum" is a reserved word, so only "type" and "items" can be destructured
+const nums = new Set(["number", "integer"]);
+
+/**
+ * Turn a DOM event form value into its formData counterpart.
+ *
+ * It is especially used in `SelectWidget` widgets.
```
epicfaace, line 587:
```suggestion
 * It is currently only used in `SelectWidget` widgets.
```

packages/semantic-ui/src/SelectWidget/SelectWidget.js
oncletom: Sorry, I misread: the function was not originally part of `fluent-ui`.

jakob-ed: 👍

heath-freenome:
this change was replicated into the `rjsf-v5` branch via #3026

PR #2270, opened 2021-03-10
Dereference option schemas before giving it as props to anyOf and oneOf

mr-bjerre:
### Reasons for making this change

Fixes #2016 and #1819

If this is related to existing tickets, include links to them as well.

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests
  - [x] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #2245, opened 2021-02-22
fix: anyOf/oneOf not load options with $ref correctly

bachdx96:
### Reasons for making this change

This fixes oneOf/anyOf not updating default value and label correctly if $ref is used.

[...]

Issue #1433, opened 2019-08-27
Conflicting versions of ajv break the build


BurkovBA:
# Description

My project depends on both `react-jsonschema-form` (which depends on `ajv` version 5) and a separate instance of `ajv` library, version 6. 

When I build my project with Webpack, `react-jsonschema-form` is trying to use the global `ajv` version 6 instead of its own `ajv` version 5, resulting in an error in runtime:

```js
Uncaught Error: schema should be object or boolean
    at Ajv._addSchema (ajv.js:303)
    at Ajv.addSchema (ajv.js:141)
    at Ajv.addMetaSchema (ajv.js:156)
    at addDraft6MetaSchema (ajv.js:454)
    at new Ajv (ajv.js:75)
    at createAjvInstance (validate.js:38)
    at Object.<anonymous> (validate.js:33)
    at __webpack_require__ (bootstrap 5b418ac9dcd29920a85c:678)
    at fn (bootstrap 5b418ac9dcd29920a85c:88)
    at Object.<anonymous> (utils.js:83)
```

Here is my code example:

`index.jsx`
```js
import React from 'react';
import { withRouter } from 'react-router-dom';
import Form from "react-jsonschema-form";

import routes from 'services/routes.jsx';


class Index extends React.Component {
  constructor(props) {
    super(props);

    this.state = {
      schema: {
        title: "Todo",
        type: "object",
        required: ["title"],
        properties: {
          title: {type: "string", title: "Title", default: "A new task"},
          done: {type: "boolean", title: "Done?", default: false}
        }
      }
    };
  }

  render() {
    return (
      <div className="row">
        <div className="col-lg-12">
          <div className="hpanel">
            <div className="panel-heading">
              <h1>Ожидаемая эффективность клиентских менеджеров</h1>
            </div>
            <div className="panel-body">
              <Form schema={this.state.schema} onChange={console.log("changed")} onSubmit={console.log("submitted")} onError={console.log("errors")} />
            </div>
          </div>
        </div>
      </div>
    )
  }
}

export default withRouter(Index);
```

Problem can be solved by importing your bundle from `/dist` folder instead of using main module:

```js
import Form from "react-jsonschema-form/dist/react-jsonschema-form.js";
```

instead of

```js
import Form from react-jsonschema-form
```

### Version

Encountered in `react-jsonschema-form` 1.7 and 1.8.

epicfaace:
Thanks for reporting. #1286 would probably fix this.

PR #1286, opened 2019-05-17
Injection of custom ajv instance via prop (+ removal of global ajv)

sbusch:
### Reasons for making this change

DO NOT MERGE YET.

A custom ajv instance can be created and injected to the Form as "ajv" prop. This allows further customizations of behavior.

Notes:

* `createAjvInstance()` from `utils.js` can be used to generate a preconfigured instance, which can be further customized (by adding custom keywords, ....).
* change allows multiple Forms with differing ajv configuration, e.g. different `customFormats`
* If no "ajv" prop is provided, the Form generated an own instance that is cached (memoized) depending on the ajv-related configuration props (`additionalMetaSchemas` and `customFormats`)
* Many utility functions needed to be modified to pass ajv instance through. I decided to add ajv as 1st parameter, because this is no optional argument

This feature allows to deprecate some features, for example Form features `additionalMetaSchemas` and `customFormats` could be achieved by an customized ajv, further simplifying `react-jsonschema-form` implementation.

I decided to release this now to get some early feedback. ajv injection and removal of global ajv is finished and tests are passing (see status, below).

I'll try to rebase this for some time on master, but since there are many changes quick feedback would be great...

Comments very appreciated!

Status:

- [x] ajv injection and removal of global ajv finished
- [x] all tests are passing (despite the three ones that are always failing, see my other PR #1285 for a fix)
- [ ] provide tests (work in progress, I started with a suboptimal testcase which needs to be reworked)
- [ ] provide playground example (work in progress, here I also started with a suboptimal example)
- [ ] update docs
- [ ] document changed signatures of exported functions (❓ are there API docs somewhere?)
- [ ] (maybe) deprecate props/features like `additionalMetaSchemas` and `customFormats` (deprecation marker in source, updated docs)
- [ ] Link to existing issues that could be solved by this PR (🆘 Some help here would be great)

### Checklist

**TODO**

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
  - [ ] I've run `npm run cs-format` on my branch to conform my code to [prettier](https://github.com/prettier/prettier) coding style
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #2272, opened 2021-03-12
Fix default values and matching schemas when oneOf / anyOf subschemas contain references

### LINKED ISSUE #2269: Loading anyOf with initial data does not load the right option when using refs
### Prerequisites

### Description

I have an `anyOf` schema which I want to load with some initial data. Based on the data provided it can be inferred which options was selected but it is not working if I use `$ref` which is necessary due to recursive dependencies.

### Steps to Reproduce

Here is a [playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJpZCI6ImFycmF5IiwiY29tcG9uZW50cyI6W3siaWQiOiJiIiwibmFtZSI6IkNodWNrIn0seyJpZCI6ImEifV19LCJzY2hlbWEiOnsiJHJlZiI6IiMvZGVmaW5pdGlvbnMvQW55IiwiZGVmaW5pdGlvbnMiOnsiQSI6eyJ0aXRsZSI6IkEiLCJyZXF1aXJlZCI6WyJpZCJdLCJwcm9wZXJ0aWVzIjp7ImlkIjp7ImVudW0iOlsiYSJdfX19LCJCIjp7InRpdGxlIjoiQiIsInJlcXVpcmVkIjpbImlkIiwibmFtZSJdLCJwcm9wZXJ0aWVzIjp7ImlkIjp7ImVudW0iOlsiYiJdfSwibmFtZSI6eyJ0eXBlIjoic3RyaW5nIn19fSwiQyI6eyJ0aXRsZSI6IkMiLCJyZXF1aXJlZCI6WyJpZCJdLCJwcm9wZXJ0aWVzIjp7ImlkIjp7ImVudW0iOlsiYyJdfX19LCJBbnkiOnsiYW55T2YiOlt7IiRyZWYiOiIjL2RlZmluaXRpb25zL0EiLCJ0aXRsZSI6IkEifSx7IiRyZWYiOiIjL2RlZmluaXRpb25zL0IiLCJ0aXRsZSI6IkIifSx7IiRyZWYiOiIjL2RlZmluaXRpb25zL0MiLCJ0aXRsZSI6IkMifSx7IiRyZWYiOiIjL2RlZmluaXRpb25zL0FycmF5QW55IiwidGl0bGUiOiJBcnJheUFueSJ9XX0sIkFycmF5QW55Ijp7InRpdGxlIjoiQXJyYXkiLCJyZXF1aXJlZCI6WyJpZCJdLCJwcm9wZXJ0aWVzIjp7ImlkIjp7ImVudW0iOlsiYXJyYXkiXX0sImNvbXBvbmVudHMiOnsidHlwZSI6ImFycmF5IiwiaXRlbXMiOnsiJHJlZiI6IiMvZGVmaW5pdGlvbnMvQW55In19fX19fSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlIjpmYWxzZSwib21pdEV4dHJhRGF0YSI6dHJ1ZSwibGl2ZU9taXQiOmZhbHNlfX0=)

This example includes 4 schemas; `A`, `B`, `C` and `ArrayAny`. Each of these have an `id` property with a constant value to identify with schema is chosen. There is also a "reference schema" `Any` which is just an anyOf with reference to the four schemas. The root schema starts with `Any`.

The initial `formData` is 

```json
{
  "id": "array",
  "components": [
    {
      "id": "b",
      "name": "Chuck"
    },
    {
      "id": "a"
    }
  ]
}
```

#### Expected behavior

I'd expect my schema to load with open `ArrayAny` in root. I'd expect the property components to be a list of two schemas, i.e schema `B` followed by schema `A`.

#### Actual behavior

It simply loads the first schema, i.e. schema `A`.

#### Observations

If I submit when I load the playground then I get the following 6 validation although the initial data is actually valid

```
.id is a required property
.id is a required property
.name is a required property
.id is a required property
.id is a required property
should match some schema in anyOf
```

I'd expect that to be related to why it doesn't load the initial data properly.

I've also tried to make a similar playground without refs to the schemas with initial data - then it works as expected. Unfortunately that won't work with recursive models though.

### Version

Online playground

mr-bjerre:

Atm. schemas are not able to match with the correct option based on the formData if the option schema have any references.

Fixes #2269 
Branched out from PR #2270 since it is required

### Checklist


mr-bjerre:
@epicfaace thanks for this library first of all. I'm really digging it these days. 

I don't like to poke you like this, but I just noticed that there are a lot of pull requests pending so I'm wondering if there is anything I can do to ease the acceptance of my PRs? This and #2270 since they'll bring a lot of value for my use case and anyone using [FastAPI](https://github.com/tiangolo/fastapi) for their backend since it gives you auto/generated OpenAPI schema with references to each model schema (so references has to be supported).

epicfaace:
I think it's really just maintainer bandwidth -- I'm the only person available to review PRs, and I can only work for so much on a volunteer basis (the problem with open source in general!) If you'd like to be a maintainer and help review PRs to clear out the queue, let me know 😄

mr-bjerre:
I figured and totally understandable. I'm not sure what exactly it requires of me to be a maintainer. I'm quite new in the open-source world still but I might be able to help out (limited time for me as well).

On another note; I just noticed that this PR is highly related to #2245

mr-bjerre:
Integrated PR #2245 into this PR (copying some tests etc)

mr-bjerre:
Commit [67ef25d](https://github.com/rjsf-team/react-jsonschema-form/pull/2272/commits/67ef25d5b5a32db2f44cf165f08aa9f6d920be43) have caused some recursive issue. I'll address that later today

packages/core/test/oneOf_test.js
```diff
@@ -611,5 +614,45 @@ describe("oneOf", () => {
...
+      const { node } = createFormComponent({
+        schema,
+      });
+
+      const $select = node.querySelector("select");
+
+      expect($select.options[0].text).eql("Foo");
+      expect($select.options[1].text).eql("Option 2");
+      expect($select.options[2].text).eql("Baz");
+    });
```
epicfaace, line 656: Should we also add a `"should correctly infer the selected option based on value"` test case for oneOf?

packages/core/test/utils_test.js
```diff
@@ -3827,5 +3828,41 @@ describe("utils", () => {
         }))
       );
     });
+    it("should infer correct anyOf schema based on data", () => {
```
epicfaace, line 3831: Let's also add a test case for oneOf. We could just parameterize this test in a loop, if that's easier, since the test cases are nearly exactly the same.

packages/core/src/utils.js
```diff
@@ -1200,7 +1200,18 @@ export function getMatchingOption(formData, options, rootSchema) {
         })),
       };
 
-      let augmentedSchema;
+      // start augmentedSchema as an object to hold only references
+      let {
```
epicfaace, line 1204: I'm not comfortable with this approach, because I think the schema can have additional keys (either in the current spec or in future JSON Schema spec).

What exactly is this code doing, though -- what do you mean by "hold only references"? References can also refer to items within `properties`, too, I believe.

packages/core/src/components/fields/SchemaField.js
```diff
@@ -362,7 +362,9 @@ function SchemaFieldRender(props) {
             onBlur={props.onBlur}
             onChange={props.onChange}
             onFocus={props.onFocus}
-            options={schema.anyOf}
+            options={schema.anyOf.map(_schema =>
+              retrieveSchema(_schema, rootSchema, formData)
```
epicfaace, line 366: I guess this is technically a breaking change, since schemas passed to `AnyOfField` are dereferenced. This should be okay, though, given that otherwise, the non-dereferenced schemas would be useless because the `rootSchema` isn't passed to fields anyway.

packages/core/src/utils.js
mr-bjerre: yeah I'm not too fond of that either. Let's say we have a rootSchema like this

```json
{
  "components": {
    "a": {"type": "string"}
  },
  "definitions": {
    "a": {"enum": ["1", "2"]}
  },
  ...
}
```

When we have to validate some `schema` then we need to include `components` and `definitions`, i.e. those need to be included when we call [`isValid`](https://github.com/mr-bjerre/react-jsonschema-form/blob/bbf4a9468bf83b785f3fd00bb8f7aa28397112d2/packages/core/src/utils.js#L1239). So basically I'm trying to let the `augmentedSchema` to be an object only holding `components` and `definitions` by removing reserved JSON schema keywords - I'm not sure how to do that properly.

The issue with the latest release state is that `components` and `definitions` wouldn't be included in `augmentedSchema` so if the object being validated would have any `$ref` then `isValid` would return `false`.

Do you have any suggestions?

packages/core/src/components/fields/SchemaField.js
mr-bjerre: Yeah I agree and since `retrieveSchema` is idempotent then it should be okay so I figured it was best to do this as high up as possible (also based on a comment from you earlier).

Let me know if you think this can stay

packages/core/test/utils_test.js
mr-bjerre: Added almost identical test in `oneOf_test.js`

packages/core/src/components/fields/SchemaField.js
epicfaace: Yeah, this can stay

packages/core/src/utils.js
epicfaace: > When we have to validate some `schema` then we need to include `components` and `definitions`,

I think in that case, we could do:

```
let augmentedSchema = {components: rootSchema.components, definitions: rootSchema.definitions};
```

However, I'm still not sure if just including `components` and `definitions` are enough. Does the JSON Schema spec restrict references to items within `components` and `definitions`? I thought $ref's could refer to anywhere in the schema (not necessarily restricted to those keywords), but maybe I'm mistaken.

mr-bjerre: That's exactly the challenge and the reason that I "discarded" properties from the schema instead of handpicking. The logic being; I know what NOT to pick from the schema but I don't know what to pick. 

Alternatively we'd have to find all $ref in the specific schema being validated and make sure to include those from rootSchema

epicfaace: Hmm, but could a `$ref` point to an item within `properties`? If so, even the current approach in this PR wouldn't work.

mr-bjerre: I'm not sure if that's allowed. It's definitely bad practice and I don't think it's allowed. 

And yeah I guess that's basically what we do with Object.assign or destructering now. The hard part is retrieving the `defSchema` which is what I basically attempt

mr-bjerre: You can't just combine the root schema with the current subschema since the root schema can have a `type` for instance while the current subschema doesn't. Then the merged would be valid.

epicfaace: Actually, it looks like it's not forbidden by the json schema spec. Because of that, I don't know if it's really considered bad practice:

> The "$ref" keyword is an applicator that is used to reference a statically identified schema. Its results are the results of the referenced schema. [CREF5]

> The value of the "$ref" keyword MUST be a string which is a URI-Reference. Resolved against the current URI base, it produces the URI of the schema to apply. This resolution is safe to perform on schema load, as the process of evaluating an instance cannot change how the reference resolves.

https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.2.3.1

It's also supported by rjsf ([playground example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJhZGRyZXNzIjp7fSwic2hpcHBpbmdfYWRkcmVzcyI6eyJzdHJlZXRfYWRkcmVzcyI6IjIyMUIsIEJha2VyIFN0cmVldCIsImNpdHkiOiJMb25kb24iLCJzdGF0ZSI6Ik4vQSJ9LCJ0cmVlIjp7ImNoaWxkcmVuIjpbeyJuYW1lIjoibGVhZiJ9XSwibmFtZSI6InJvb3QifSwiYmlsbGluZ19hZGRyZXNzIjp7InN0cmVldF9hZGRyZXNzIjoiMjEsIEp1bXAgU3RyZWV0IiwiY2l0eSI6IkJhYmVsIiwic3RhdGUiOiJOZXZlcmxhbmQifX0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJhZGRyZXNzIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7InN0cmVldF9hZGRyZXNzIjp7InR5cGUiOiJzdHJpbmcifSwiY2l0eSI6eyJ0eXBlIjoic3RyaW5nIn0sInN0YXRlIjp7InR5cGUiOiJzdHJpbmcifX0sInJlcXVpcmVkIjpbInN0cmVldF9hZGRyZXNzIiwiY2l0eSIsInN0YXRlIl19LCJzdGF0ZSI6eyJ0aXRsZSI6IlN0YXRlIChyZWZlcmVuY2VkKSIsIiRyZWYiOiIjL3Byb3BlcnRpZXMvYWRkcmVzcy9wcm9wZXJ0aWVzL3N0YXRlIn19fSwidWlTY2hlbWEiOnsidWk6b3JkZXIiOlsiKiJdfSwidGhlbWUiOiJkZWZhdWx0IiwibGl2ZVNldHRpbmdzIjp7InZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==)).

epicfaace: > You can't just combine the root schema with the current subschema since the root schema can have a `type` for instance while the current subschema doesn't. Then the merged would suddenly have potential wrong type

Right, so the solution I have in mind with the [ajv link](https://ajv.js.org/guide/combining-schemas.html#combining-schemas-with-ref) is to not use an augmentedSchema at all, but rather:
- add the root schema using `addSchema()`, set the $id of the schema to to `rootSchema.json`
- rewrite all ref's in the subschema by prepending their values with something like `rootSchema.json`
- call ajv's `validate` function on the subschema

It's kind of messy, but it will work (but maybe there's a better alternative?)

mr-bjerre: Aha I'm catching your drift now :racing_car: 

I'll have a look later. I wonder if this should be done on a higher level then, like initially. Any thoughts on that?

mr-bjerre: Do you want to change `isValid` to receive the `rootSchema` as well? I'm not really sure where the `addSchema()` should happen since it returns an `ajv` instance which is needed for the validation. The issue is that `isValid` uses a global `ajv` instance.

epicfaace: > Do you want to change `isValid` to receive the `rootSchema` as well?

Yeah, I think that makes sense. The fact that the `ajv` instance is global is a separate problem (see https://github.com/rjsf-team/react-jsonschema-form/pull/1286/ and https://github.com/rjsf-team/react-jsonschema-form/issues/1433), but work has stalled on it. I think it's too much for this PR to also make ajv also non-global, though.

How about something like this (so we'd keep the global ajv, just reserve a key to be used for root schemas, then add / remove that schema whenever isValid is called):

```js
export function isValid(schema, data, rootSchema) {
  try {
    ajv.addSchema(rootSchema, "__rjsf_rootSchema");
    // transform $ref's in schema to add "__rjsf_rootSchema" as the prefix
    result = ajv.validate(schema, data);
    ajv.removeSchema("__rjsf_rootSchema");
    return result;
  } catch (e) {
    return false;
  }
}

Comment (mr-bjerre) [reply packages/core/src/utils.js]: I ended up creating a new ajv instance when using rootSchema. Alternatively we'd have to improve the error handling to make sure the `__rjsf_rootSchema` would always be removed (even though validate raises an error). We could easily do that as well of course (maybe just try catch within the existing catch).

I also had to keep the `augmentedSchema` (as in the latest release). Without it I got a `anyOf` test failing, i.e. "should correctly update inputs for anyOf inside array items after being moved down" (both items resolved to `foo` although the second should resolve to `bar` based on the data).

Comment (epicfaace) [reply packages/core/src/utils.js]: Actually, we can't use `createAjvInstance()` -- we need to use the global ajv instance because it includes any additional meta schemas or custom formats: https://github.com/rjsf-team/react-jsonschema-form/blob/64b89213061f1820c55a42d791446bc80ae8d94b/packages/core/src/validate.js#L182-L203

Yeah, makes sense why we need to keep the `augmentedSchema`, given the additional modifications the code does to it for `allOf` / `required` properties.

File: packages/core/src/validate.js [hunk 5]
```diff
@@ -263,14 +263,45 @@ export default function validateFormData(
   };
 }
 
+function isDict(v) {
+  return (
```
epicfaace, line 267: I believe we could make this a bit simpler by doing:

`return v.constructor === Object`

https://stackoverflow.com/questions/38304401/javascript-check-if-dictionary
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor

packages/core/src/validate.js
```diff
@@ -263,14 +263,45 @@ export default function validateFormData(
...
 /**
  * Validates data against a schema, returning true if the data is valid, or
  * false otherwise. If the schema is invalid, then this function will return
  * false.
  */
-export function isValid(schema, data) {
+export function isValid(schema, data, rootSchema) {
   try {
-    return ajv.validate(schema, data);
+    if (rootSchema) {
```
epicfaace, line 298: Let's add a comment here explaining what we are doing with ajv.

packages/core/src/validate.js
```diff
@@ -263,14 +263,45 @@ export default function validateFormData(
...
+  return (
+    typeof v === "object" &&
+    v !== null &&
+    !(v instanceof Array) &&
+    !(v instanceof Date)
+  );
+}
+
+/**
+ * Get a similar schema where ref's are prefixed with "__rjsf_rootSchema"
```
epicfaace, line 276: Add a comment here explaining that this is passed to ajv to represent the rootSchema in `isValid`.

mr-bjerre:
I should've accounted for your directions now and merged the branch with the latest changes

packages/core/src/validate.js
```diff
@@ -263,25 +263,17 @@ export default function validateFormData(
...
 /**
  * Get a similar schema where ref's are prefixed with "__rjsf_rootSchema"
+ * This is used in isValid to make references to the rootSchema
  */
 function withIdRefPrefix(schema) {
   const obj = { ...schema };
   for (let key of Object.keys(obj)) {
     const value = obj[key];
     if (key === "$ref") {
       obj[key] = "__rjsf_rootSchema" + value;
```
epicfaace, line 275: Let's extract __rjsf_rootSchema into a constant.

packages/core/src/validate.js
```diff
@@ -263,25 +263,17 @@ export default function validateFormData(
...
-    !(v instanceof Array) &&
-    !(v instanceof Date)
-  );
-}
-
 /**
  * Get a similar schema where ref's are prefixed with "__rjsf_rootSchema"
+ * This is used in isValid to make references to the rootSchema
  */
 function withIdRefPrefix(schema) {
```
epicfaace, line 270: Will this function work properly with a schema that has an array of objects, one of which has a $ref?

epicfaace:
> Should I squash the commits btw? It kinda blew up this progress

Nope, doesn't matter -- we'll squash and merge anyway.

packages/core/src/validate.js
mr-bjerre: I guess you're right. We should have a test for that

packages/core/src/validate.js
```diff
@@ -264,17 +265,25 @@ export default function validateFormData(
...
-function withIdRefPrefix(schema) {
-  const obj = { ...schema };
-  for (let key of Object.keys(obj)) {
-    const value = obj[key];
-    if (key === "$ref") {
-      obj[key] = "__rjsf_rootSchema" + value;
-    } else if (value.constructor === Object) {
-      obj[key] = withIdRefPrefix(value);
+function withIdRefPrefix(schemaNode) {
+  let obj = schemaNode;
```
epicfaace, line 272:
```suggestion
  let obj;
```

packages/core/src/validate.js
```diff
@@ -5,6 +5,7 @@ import { deepEquals, getDefaultFormState } from "./utils";
 
 let formerCustomFormats = null;
 let formerMetaSchema = null;
+const rootSchemaId = "__rjsf_rootSchema";
```
epicfaace, line 8: Rename to 
```suggestion
const ROOT_SCHEMA_PREFIX = "__rjsf_rootSchema_";
```

packages/core/src/validate.js
```diff
@@ -264,17 +265,25 @@ export default function validateFormData(
 }
 
 /**
- * Get a similar schema where ref's are prefixed with "__rjsf_rootSchema"
+ * Get a similar schema where ref's are prefixed with `rootSchemaId`
```
epicfaace, line 268:
```suggestion
 * Recursively prefixes all $ref's in a schema with `ROOT_SCHEMA_PREFIX`
```

packages/core/src/validate.js
```diff
@@ -264,17 +265,25 @@ export default function validateFormData(
...
-    const value = obj[key];
-    if (key === "$ref") {
-      obj[key] = "__rjsf_rootSchema" + value;
-    } else if (value.constructor === Object) {
-      obj[key] = withIdRefPrefix(value);
+function withIdRefPrefix(schemaNode) {
+  let obj = schemaNode;
+  if (schemaNode.constructor === Object) {
+    obj = { ...schemaNode };
+    for (let key of Object.keys(obj)) {
```
epicfaace, line 275:
```suggestion
    for (const key in obj) {
```

packages/core/src/validate.js
```diff
@@ -264,17 +265,25 @@ export default function validateFormData(
...
  */
-function withIdRefPrefix(schema) {
-  const obj = { ...schema };
-  for (let key of Object.keys(obj)) {
-    const value = obj[key];
-    if (key === "$ref") {
-      obj[key] = "__rjsf_rootSchema" + value;
-    } else if (value.constructor === Object) {
-      obj[key] = withIdRefPrefix(value);
+function withIdRefPrefix(schemaNode) {
```
epicfaace, line 271: Can we add a few unit tests for `withIdRefPrefix`? Just want to make sure it's able to handle complex schemas and it doesn't regress.

packages/core/src/validate.js
```diff
@@ -264,17 +265,25 @@ export default function validateFormData(
...
+      if (key === "$ref") {
+        obj[key] = rootSchemaId + value;
+      } else {
+        obj[key] = withIdRefPrefix(value);
+      }
+    }
+  } else if (Array.isArray(schemaNode)) {
+    obj = [...schemaNode];
+    for (var i = 0; i < obj.length; i++) {
+      obj[i] = withIdRefPrefix(obj[i]);
```
epicfaace, line 286: What about a schema like this:

```json
{
  "type": "object",
  "properties": {
    "$ref": {
      "type": "string",
      "title": "First name",
      "default": "Chuck"
    }
  }
}
```

It _is_ a valid schema (see [playground example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyIkcmVmIjoiQ2h1Y2siLCJmaXJzdE5hbWUiOiJDaHVjayIsImxhc3ROYW1lIjoiTm9ycmlzIiwiYWdlIjo3NSwiYmlvIjoiUm91bmRob3VzZSBraWNraW5nIGFzc2VzIHNpbmNlIDE5NDAiLCJwYXNzd29yZCI6Im5vbmVlZCJ9LCJzY2hlbWEiOnsidGl0bGUiOiJBIHJlZ2lzdHJhdGlvbiBmb3JtIiwiZGVzY3JpcHRpb24iOiJBIHNpbXBsZSBmb3JtIGV4YW1wbGUuIiwidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsiJHJlZiI6eyJ0eXBlIjoic3RyaW5nIiwidGl0bGUiOiJGaXJzdCBuYW1lIiwiZGVmYXVsdCI6IkNodWNrIn19fSwidWlTY2hlbWEiOnsiZmlyc3ROYW1lIjp7InVpOmF1dG9mb2N1cyI6dHJ1ZSwidWk6ZW1wdHlWYWx1ZSI6IiIsInVpOmF1dG9jb21wbGV0ZSI6ImZhbWlseS1uYW1lIn0sImxhc3ROYW1lIjp7InVpOmVtcHR5VmFsdWUiOiIiLCJ1aTphdXRvY29tcGxldGUiOiJnaXZlbi1uYW1lIn0sImFnZSI6eyJ1aTp3aWRnZXQiOiJ1cGRvd24iLCJ1aTp0aXRsZSI6IkFnZSBvZiBwZXJzb24iLCJ1aTpkZXNjcmlwdGlvbiI6IihlYXJ0aGlhbiB5ZWFyKSJ9LCJiaW8iOnsidWk6d2lkZ2V0IjoidGV4dGFyZWEifSwicGFzc3dvcmQiOnsidWk6d2lkZ2V0IjoicGFzc3dvcmQiLCJ1aTpoZWxwIjoiSGludDogTWFrZSBpdCBzdHJvbmchIn0sImRhdGUiOnsidWk6d2lkZ2V0IjoiYWx0LWRhdGV0aW1lIn0sInRlbGVwaG9uZSI6eyJ1aTpvcHRpb25zIjp7ImlucHV0VHlwZSI6InRlbCJ9fX0sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6ZmFsc2UsImRpc2FibGUiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlfX0=)).

I'm not sure if it would be easy to fix this use case, though. Let me know what you think.

epicfaace: Maybe we could just check to ensure that the ref's value is a string, and we rename the ref only in that case?

packages/core/src/validate.js
```diff
@@ -263,15 +264,60 @@ export default function validateFormData(
...
+    }
   } catch (e) {
+    try {
+      // make sure we also remove the rootSchema if an error occured before removing but after adding
+      ajv.removeSchema(rootSchemaId);
+    } catch (e) {
+      return false;
+    }
     return false;
   }
```
epicfaace, line 322: To make this cleaner, and consolidate the removeSchema logic above, let's actually just remove the schema in a finally block:

```suggestion
  }
  finally {
    try {
      ajv.removeSchema(rootSchemaId);
    } ... 
  }
```

Also, do you know if `removeSchema` returns an error if `rootSchemaId` hasn't already been added to ajv?

packages/core/src/validate.js
mr-bjerre: That would break stuff. The input name may be misleading. `schemaNode` can be primitive, Object or array. It should return accordingly

packages/core/src/validate.js
mr-bjerre: Oh it doesn't raise an error so no need for try around that part

mr-bjerre:
Might be ready now 😎

packages/core/src/validate.js
```diff
@@ -5,6 +5,7 @@ import { deepEquals, getDefaultFormState } from "./utils";
 
 let formerCustomFormats = null;
 let formerMetaSchema = null;
+const ROOT_SCHEMA_PREFIX = "__rjsf_rootSchema";
```
epicfaace, line 8: Let's add a trailing `_`, just so that the generated schema names will be more readable

```suggestion
const ROOT_SCHEMA_PREFIX = "__rjsf_rootSchema_";
```

epicfaace: Never mind, we'll have a `/` at the end anyway.

epicfaace, line 8: Also, just double-checking -- are we positive that all refs that are currently recognized by rjsf and/or valid according to the JSON Schema spec will start with `#`? If not, we might be breaking some functionality.

mr-bjerre: not sure what you're concern is exactly. Do you want to check that ref starts with `#` before adding the prefix?

packages/core/src/validate.js
```diff
@@ -263,15 +264,54 @@ export default function validateFormData(
...
  */
-export function isValid(schema, data) {
+export function isValid(schema, data, rootSchema) {
   try {
-    return ajv.validate(schema, data);
+    // if rootSchema is given then we add that schema with an id.
+    // then we rewrite the schema ref's to point to the rootSchema using the id
+    // this accounts for the case where schema have references to models
+    // that lives in the rootSchema but not in the schema in question.
+    if (rootSchema) {
```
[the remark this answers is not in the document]
mr-bjerre: You're right - except for some tests, but I've just updated those tests to add the schema as rootSchema as well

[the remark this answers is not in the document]
mr-bjerre: Alternatively I could leave the tests as is and have this instead

```javascript
return ajv
  .addSchema(rootSchema || schema, ROOT_SCHEMA_PREFIX)
  .validate(withIdRefPrefix(schema), data);
```

So if no rootSchema is provided then the schema itself will be used. Any preferences with one or the other?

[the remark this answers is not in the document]
epicfaace: I think it's better to just go ahead and update the tests rather than adding unnecessary functionality to this function.

packages/core/src/validate.js
epicfaace: Looking at this again, I don't think people will ever have ref's that don't start with `#` (because rjsf crashes otherwise!) So we should be okay with just the existing code.

packages/core/src/components/fields/SchemaField.js
```diff
@@ -362,7 +362,9 @@ function SchemaFieldRender(props) {
             onBlur={props.onBlur}
             onChange={props.onChange}
             onFocus={props.onFocus}
-            options={schema.anyOf}
+            options={schema.anyOf.map(_schema =>
```
epicfaace, line 365: I'm going to label this a breaking change because it technically is (so people can't rely on `$ref`s being present in the `options` parameter in custom `OneOfField` / `AnyOfField`s). We can still merge this in now, though, since 3.x will be the next release.

Can you add a section to the "3.x upgrade guide" file in the docs that describes what people should be aware about regarding the breaking change?

packages/core/src/validate.js
mr-bjerre: I actually went ahead and added 

```javascript
if (
        key === "$ref" &&
        typeof value === "string" &&
        value.startsWith("#")
      )
...
```

I assume its fine having that check anyways

packages/core/src/components/fields/SchemaField.js
mr-bjerre: Added - please verify if it makes sense to you

mr-bjerre:
Are we ready to merge this one?

SimonKlausLudwig:
Would love to see this as well. Thanks for your work.

epicfaace:
Haven't taken a look yet -- will do soon

gsstar9028:
please take a look soon

packages/core/test/validate_test.js
```diff
@@ -27,13 +31,71 @@ describe("Validation", () => {
...
+      };
+
+      withIdRefPrefix(schema);
+
+      expect(schema).to.eql({
+        anyOf: [{ $ref: "#/defs/foo" }],
+      });
+    });
+
+    it("should not change the ref since it's a property", () => {
```
epicfaace, line 88:
```suggestion
    it("should not change a property named '$ref'", () => {
```

packages/core/test/validate_test.js
```diff
@@ -27,13 +31,71 @@ describe("Validation", () => {
...
+        anyOf: [{ $ref: "#/defs/foo" }],
+      };
+      const expected = {
+        anyOf: [{ $ref: "__rjsf_rootSchema#/defs/foo" }],
+      };
+
+      expect(withIdRefPrefix(schema)).to.eql(expected);
+    });
+
+    it("shouldn't mutated the schema", () => {
```
epicfaace, line 76:
```suggestion
    it("shouldn't mutate the schema", () => {
```

packages/core/test/anyOf_test.js
```diff
@@ -639,6 +678,52 @@ describe("anyOf", () => {
     });
   });
 
+  it("should use title from refs schema before using fallback generated value as title", () => {
```
epicfaace, line 681: Can we just add one more test that includes a ref that refers to something within properties like (`{"$ref": "#/properties/test"}`) so we can ensure that functionality doesn't regress in the future?

gsstar9028: @mr-bjerre

packages/core/src/components/fields/SchemaField.js
epicfaace: This looks great, thanks!

mr-bjerre:
Ready to go (fingers crossed)

epicfaace (approved):
Great, thanks for your work on this!

mr-bjerre:
Wuhu thanks. My first actual code PR 🎊

PR #2276, opened 2021-03-16
fix docs: fields prop allows to override OneOfField and AnyOfField

skyboyer:

Docs reference to `MultiSchemaField` as overridable, but actually `OneOfField` and `AnyOfField` should be overriden

### Checklist


epicfaace (approved):
Thanks!

PR #2278, opened 2021-03-16
There shouldn;t be redundant name var

aexol:
Fluent UI forms doesn't compile in ts at all because of this.

### DIFF
--- a/packages/fluent-ui/src/TextWidget/TextWidget.tsx
+++ b/packages/fluent-ui/src/TextWidget/TextWidget.tsx
@@ -81,7 +81,6 @@ const TextWidget = ({
       required={required}
       disabled={disabled}
       readOnly={readonly}
-      name={name}
       type={inputType as string}
       value={value ? value : ""}
       onChange={_onChange as any}


epicfaace:
Instead of removing `name` completely, can you add `name` here (https://github.com/rjsf-team/react-jsonschema-form/blob/9244d8067ed1cc39181022b264e5f011d1bff157/packages/fluent-ui/src/TextWidget/TextWidget.tsx#L47-L60)?

aexol:
https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield
Hi again. It looks like `name` prop does not exist in fluent react anymore. Should we add it anyway?

epicfaace:
Hmm, this is problematic. I guess we can remove the `name` prop for now, though I'd really like it for fluent-ui to support the `name` prop in the future. Thanks for pointing this out!

epicfaace:
Can you fix CI, so then we can merge?

epicfaace:
Closing this in favor of https://github.com/rjsf-team/react-jsonschema-form/pull/2337, which already fixes the errors related to typescript compilation.

PR #2211, opened 2021-01-29
No core-js in core

wegry:
### Reasons for making this change

Two versions of core-js currently exist in my app because @rjsf/core pulls in @babel/runtime currently. There are two direct imports of core-js that I've opted to pull in pure imports for.

For the rationale of why library code shouldn't include polyfills, this comment
https://github.com/w3ctag/polyfills/issues/6#issuecomment-272647475 explains my rationale. This is a breaking change though, if downstream folks were depending on `@babel/runtime` being around.

If this is related to existing tickets, include links to them as well. fixes #1601

Recreates PR for #2049

### Checklist

* [x] **I'm updating documentation**
  - [x] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [X] **I'm adding or updating code**
  - [ ] I've added and/or updated tests
  - [x] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [x] **I'm adding a new feature**
  - [x] I've updated the playground with an example use of the feature

PR #2279, opened 2021-03-17
refactor: types for withTheme

### LINKED ISSUE #2135: Ref is not defined on FormProp type
### Description

The following documentation example produces an error when compiling with typescript :
https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/internals/#submit-form-programmatically

### Steps to Reproduce

1. Copy the example from https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/internals/#submit-form-programmatically:
```jsx
const onSubmit = ({formData}) => console.log("Data submitted: ",  formData);
let yourForm;

const schema = {
    type: "string"
};

render((
  <Form schema={schema}
        onSubmit={onSubmit} ref={(form) => {yourForm = form;}}/>

yourForm.submit();
```

#### Expected behavior

It should compile without problems

#### Actual behavior
When compiling with tsc

```
No overload matches this call.
  Overload 1 of 2, '(props: FormProps<any>, context?: any): ReactElement<any, any> | Component<FormProps<any>, any, any> | null', gave the following error.
    Type '{ children: Element; schema: JSONSchema7; liveValidate: true; noHtml5Validate: true; showErrorList: false; ref: (form: any) => void; onSubmit: (values: Record<string, any>)
e to type 'IntrinsicAttributes & FormProps<any>'.
      Property 'ref' does not exist on type 'IntrinsicAttributes & FormProps<any>'.
  Overload 2 of 2, '(props: PropsWithChildren<FormProps<any>>, context?: any): ReactElement<any, any> | Component<FormProps<any>, any, any> | null', gave the following error.
    Type '{ children: Element; schema: JSONSchema7; liveValidate: true; noHtml5Validate: true; showErrorList: false; ref: (form: any) => void; onSubmit: (values: Record<string, any>)
e to type 'IntrinsicAttributes & FormProps<any> & { children?: ReactNode; }'.
      Property 'ref' does not exist on type 'IntrinsicAttributes & FormProps<any> & { children?: ReactNode; }'.  TS2769

  > 47 |         ref={form => {
       |         ^
    48 |           yourForm = form;
    49 |         }}
    50 |         onSubmit={onSubmit}

```

### Version
```javascript
    "react": "^16.9.0",
    "typescript": "^3.8.3",
    "@rjsf/antd": "^2.4.0",
    "@rjsf/core": "^2.4.0",
    "@types/react": "^16.9.43",
```

mattcosta7:

fixes #2135 

The different themes were defining their main form (generated from the withTheme function from core) with type annotations that erased the 'ref' property. By treating withTheme as an enhancer that returns a `Form`, the Forms from the themes maintain their generic typings, as well as the ref properties they should have.

I tested a few alternatives, and found this to be the most ergonomic in usage with the best typings to least error ratio. 

- Returning a generic component function seemed wrong, and 'erases' that this ref is passed through to a class.
- Returning a ForwardRefExoticComponent or the like removes the generic nature of the types Form already uses in core, and this is something super powerful for ensuring the data and expected data across various methods stays in sync (I use this in projects relying on rjsf/core and material, so having that back in material would be great).

I added simple tests for this in the separate typescript projects here as a canary for regressions. If the typings break, these should break as well.  I didn't think it worth trying to test this in core, but I can look at that if desirable.

Additionally, I had prettier on without noticing in one of the files, i'm happy to undo those changes, but only noticed them after pushing, so i'll leave for now - likewise I can also run prettier on the other files if desired in a followup.

### Checklist


mattcosta7:
Given that this makes the typings for the exports to material, bootstrap and fluent more sound, it's possible they'd expose issues in current code for users.  I don't believe this should cause any churn, but it's possible it could

epicfaace (commented):
Thanks for the PR!

- I'm not too familiar with `ForwardRefExoticComponent`, but it looks like `forwardRef` does usually return a `ForwardRefExoticComponent`. Would it be misrepresenting the nature of `forwardRef` / could it cause any issues if we return a `typeof Form` instead?
- What kinds of backwards compatibility issues do you think users might face? If there are issues, we should probably add this as a breaking change.

packages/core/index.d.ts
```diff
@@ -437,7 +437,7 @@ declare module '@rjsf/core' {
 
 declare module '@rjsf/core/lib/components/fields/SchemaField' {
     import { JSONSchema7 } from 'json-schema';
-    import { FieldProps, UiSchema, IdSchema, FormValidation } from '@rjsf/core';
+    import Form, { FieldProps, UiSchema, IdSchema, FormValidation } from '@rjsf/core';
```
epicfaace, line 440: Why is this import needed?

packages/bootstrap-4/test/Form.test.tsx
```diff
@@ -198,3 +198,38 @@ describe("single fields", () => {
...
+        CoreForm<{
+          field: string;
+        }>
+      >(null);
+    
+      return (
+        <Form<{
+          field: string;
+        }>
+          ref={ref}
```
epicfaace, line 215: How exactly does this snapshot ensure that the `ref` property is properly passed down? What exactly are we testing here -- just to make sure that the rendering doesn't crash? (if so, we wouldn't need a snapshot)

Also, I'm having trouble seeing why the old typings would break this test. What would happen with the old types?

mattcosta7:
@epicfaace the reason for not using Forward RefExoticComponent is that the typings for it (in @types/react) prevent a generic

In core the class Form<T> allows end users to properly type things based on the shape of form data. 

The react typings for ForwardRefExoticComponent will swallow that generic.  I think that maintaining it is more important than returning that type as is (this way users can pass the generic to the component as necessary), and provide more accurate intellisense

With theme is really just an factory returning a Form component, with some defaults. I think this typing reflects that intent, albeit it hides the implementation a little

mattcosta7:
As far as comparability - 

In core components generated from withTheme would have only allowed a Formdata type of any, so the internal handlers referencing those types would have always been any (as would refs related).

It's possible that a user could have passed a formData value that gets inferred elsewhere - in a submit or change handler which might reference properties that don't exist in formData, and therefore cause a compilation error.  Since the old typing would have been any for these, by default, that would not have occured. 

For the material/fluent/etc themes, no generic was passed and so it's unlikely any issue could have existed there. 


I agree these tests aren't necessarily important, but they ensure that theres a ref passable (otherwise no value could be set as a ref) and that the component accepts a generic.  Both of these would error without that possibility - although I would like to find a good way to test this in core as well (although I'm not familiar yet with how/if we can easily test tax with the mocha setup there)

epicfaace:
> I agree these tests aren't necessarily important, but they ensure that theres a ref passable (otherwise no value could be set as a ref) and that the component accepts a generic. Both of these would error without that possibility - although I would like to find a good way to test this in core as well (although I'm not familiar yet with how/if we can easily test tax with the mocha setup there)

Ok, great -- in that case, would you be fine removing just the snapshots (so there's less snapshots for us to maintain)? Or do you think it's important to keep the snapshots themselves for the canary functionality?

mattcosta7:
Not at all, I'll drop those to just make sure something passes, and add a
bit more description to those tests so it's clearer what they're being used
for

packages/bootstrap-4/test/Form.test.tsx
```diff
@@ -199,37 +199,42 @@ describe("single fields", () => {
...
+      expect(ref.current.submit).toBeDefined();
+      expect(ref.current.formElement).toBeDefined();
+      expect(ref.current.validate).toBeDefined();
+    } else {
+      /**
+       * If we hit this, then the ref wasn't properly
+       * forwarded
+       */ 
+      expect(false).toBe(true)
+    }
```
mattcosta7, line 238: removed those extra snapshots, and using these we can also validate a few properties that should be defined on the ref

mattcosta7:
@epicfaace updated tests here.

packages/material-ui/test/Form.test.tsx
```diff
@@ -212,3 +212,43 @@ describe("single fields", () => {
...
+            field: {
+              type: "string",
+            },
+          },
+        }}
+      />
+    );
+
+    expect(ref.current).toBeDefined();
+    if (ref.current) {
```
epicfaace, line 242: I think it would be simpler just to do, instead of the else statement,

`expect(ref.current).toBeTruthy() // Expect ref to be properly forwarded`

mattcosta7:
Agreed, it would be, however the following lines reference a property on
that object, which typescript won't infer from the expect().toBeTruthy
statement

We could use non null assertions instead of wrapping in the if statement

Ala

Expect(ref.current!.submit).toBeDefined()

mattcosta7:
I pulled in the type tests from @zepatrik 's branch #1915

packages/core/index.d.ts
epicfaace: Checking on this @mattcosta7 ?

mattcosta7: it's not. i'll drop that update.  I actually don't think most of these imports are needed

epicfaace:
@mattcosta7 , since this is technically a breaking change, can you add an entry to the 3.x upgrade guide that explains to users what has changed and what they may need to change in their own code?

The guide isn't yet on the master branch (it's in #2211), but here's the version of the guide that's in that branch: https://github.com/rjsf-team/react-jsonschema-form/blob/26d34051bbc063777645aa825ed4192e260930ab/docs/3.x%20upgrade%20guide.md You can just create the `docs/3.x upgrade guide.md` file on this branch.

packages/bootstrap-4/test/Form.test.tsx
```diff
@@ -198,3 +198,35 @@ describe("single fields", () => {
     expect(tree).toMatchSnapshot();
   });
 });
+
+describe("refs", () => {
```
epicfaace, line 202: Is the test here essentially doing the same thing as the tests in the material-ui / fluent-ui tests? I wonder if they're redundant and we can just create the single test in core perhaps (where we use `withTheme` to create a simple theme). This would simplify maintenance as we wouldn't have to add this test to every new theme that uses TypeScript.

mattcosta7: yep - i'll drop this now that we have the type tests here

packages/core/index.d.ts
```diff
@@ -61,6 +61,7 @@ declare module '@rjsf/core' {
         onChange: (formData: T, newErrorSchema: ErrorSchema) => void;
         onBlur: (id: string, value: boolean | number | string | null) => void;
         submit: () => void;
+        formElement: HTMLFormElement | null;
```
epicfaace, line 64: Why add this? I thought this was meant to be private -- we don't have it documented -- but perhaps I'm wrong, is there a reason why someone would need to access this property?

mattcosta7: ahh, if it should be private, i'll hide that.  I had been using it undocumented in a project,  but can easily remove that

epicfaace: Oh -- how were you using it? If people are already using it then we might as well just document it

mattcosta7: I'm probably doing something a bit non-ideal with it. 

in an onError handler i'm looking up the property path of the field with the first error, then converting that to the name string, and finding the formElement child that matches to focus on that field.  it's non-ideal, but I'm not aware of another way to do something similar yet

medfreeman: I was using it previously to attach event handlers directly to the form for specific purposes and to avoid composing callbacks.
Are you sure that `formElement` can be null?

mattcosta7: @medfreeman I believe, but will double check, that it could be null while mounting (prior to render) and right after unmount, so it's safer to make the null check, but haven't looked in a while, so I could be wrong

medfreeman: it makes sense. thx @mattcosta7

packages/core/index.d.ts
```diff
@@ -274,7 +275,7 @@ declare module '@rjsf/core' {
 
     export function withTheme<T = any>(
         themeProps: ThemeProps<T>,
-    ): React.ComponentClass<FormProps<T>> | React.StatelessComponent<FormProps<T>>;
+    ): typeof Form;
```
medfreeman, line 278: Here's my take on this, in use in a company project for a year.
It solves the generics problem too.

For reference, it is then used in a module that has form factory filling specific needs in our project (additional default behavior, external styling additions, data type references in event handlers, error message translation, etc..).

_withTheme.tsx_
```ts
import React, { ReactElement, Ref, RefAttributes, forwardRef } from 'react';

import JsonSchemaForm, { FormProps, ThemeProps } from '@rjsf/core';

type TFormWithTheme = <T = {}>(
  p: FormProps<T> & RefAttributes<JsonSchemaForm<T>>
) => ReactElement | null;

export function withTheme<T = {}>(themeProps: ThemeProps<T>): TFormWithTheme {
  return forwardRef((props: FormProps<T>, ref: Ref<JsonSchemaForm<T>>) => {
    const { fields, widgets, ...directProps } = props;

    return (
      <JsonSchemaForm<T>
        {...themeProps}
        {...directProps}
        fields={{ ...themeProps.fields, ...fields }}
        widgets={{ ...themeProps.widgets, ...widgets }}
        ref={ref}
      />
    );
  }) as TFormWithTheme;
}
```

_usage.tsx_
```ts
const FormWithMUITheme = withTheme(MuiTheme);

type TForm = <T = {}>(
  p: FormProps<T> & RefAttributes<JsonSchemaForm<T>>
) => ReactElement | null;

export const Form: TForm = forwardRef(
    <T extends unknown = {}>(props: FormProps<T>, ref: Ref<JsonSchemaForm<T>>) => {
      const { whatever, ...otherProps } = props;
    
      /* various operations */

      return (
        <FormWithMUITheme<T>
          ref={ref}
          {...otherProps}
        />
      )}
) as TForm;
```

packages/core/typing_test.tsx
```diff
@@ -0,0 +1,321 @@
+// Originally from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/@rjsf/core/react-jsonschema-form-tests.tsx
```
epicfaace, line 1:
```suggestion
// Originally from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/93797a6d6f67508c7ab7d60b89fa2a6c2681a219/types/react-jsonschema-form/react-jsonschema-form-tests.tsx
```

docs/3.x upgrade guide.md
```diff
@@ -15,3 +15,25 @@ For a slightly more elaborate setup, [@babel/preset-env](https://babeljs.io/docs
...
 
 > We leverage [`browserslist`, `compat-table`, and `electron-to-chromium`] to maintain mappings of which version of our supported target environments gained support of a JavaScript syntax or browser feature, as well as a mapping of those syntaxes and features to Babel transform plugins and core-js polyfills.
+
+### Typescript usage of withTheme
+
+The typings for `withTheme` have been updated to more accurately reflect its nature as a factory function that returns a `Form` component and to properly type the ref forwarded to this component. See more about [the changes here](https://github.com/rjsf-team/react-jsonschema-form/pull/2279)
+
+If you're currently using `withTheme`, the typing for the formData was always `any`, as the generated Form component was not able to accept a generic argument.
+
+In v3.x these typings will properly infer formData if no type is given, or validate it if one is, similar to how the exported Form from core would work.
```
epicfaace, line 25: How is a type "given"? Should we show an example code of how to do so?

jorisw: Ping @mattcosta7

Issue #2271, opened 2021-03-11
Placeholders still not visible in v2.5.0


HannesOlszewski:
Hello there,

I installed the latest version (2.5.0) of `@rjsf/core` and `@rjsf/bootstrap-4`.

As I can see in the releases section version 2.5.0 fixed the placeholder issue. But in the package installed through yarn the issue is still present (also when using npm instead of yarn).

I checked the files in `node_modules` and the placeholder prop is not present for the TextWidget, although the `package.json` says it's version 2.5.0.

Is there an issue with the published package or is it an issue on my end?

epicfaace:
Hmm, is it working on the playground (https://rjsf-team.github.io/react-jsonschema-form/)?

HannesOlszewski:
Funny thing, in the playground with the `bootstrap-4` theme, a normal text field has a placeholder, but when adding `"ui:widget": "password"` it has no placeholder (same when selecting `semantic-ui`, the other themes work completely in the playground though).

But when using the (working) schemas from the playground in my local installation there are no placeholders.

My form schema:
```json
{
  "title": "Login",
  "type": "object",
  "properties": {
    "username": {
      "type": "string",
      "title": "Username",
      "propertyOrder": 1
    },
    "password": {
      "type": "string",
      "title": "Password",
      "widget": "password",
      "propertyOrder": 2
    }
  },
  "required": [
    "username",
    "password"
  ]
}
```

My ui schema:
```json
{
  "username": {
    "ui:placeholder": "Username"
  },
  "password": {
    "ui:placeholder": "Password",
    "ui:widget": "password"
  }
}
```

When using `material-ui` locally, the placeholders work though, but not with `bootstrap-4`.

epicfaace:
Looks like when I built 2.5.0, the build failed but I didn't notice, so I ended up publishing an older version to npm. Working on a fix...

epicfaace:
@HannesOlszewski this should now be fixed in [v2.5.1](https://github.com/rjsf-team/react-jsonschema-form/releases/tag/v2.5.1). We've added automated publishing via GitHub Actions so this shouldn't happen again. Thanks for your patience!

PR #2283, opened 2021-03-19
Release to npm using CI

epicfaace:

Release to npm using CI. This will ensure that releases are more reproducible and is better than me having to build code on my computer manually (related: https://github.com/rjsf-team/react-jsonschema-form/issues/2271).

I've also cleaned up the ci.yml file a bit. I've added additional node versions (node 14, since it's used for release) and dropped some intermediate node versions on the CI builds.

The npm credentials are from the rjsf-bot user (info in https://github.com/rjsf-team/secrets/blob/main/README.md).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,12 +8,10 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     strategy:
       matrix:
-        node-version: [8.x, 9.x, 10.x, 11.x, 12.x, 13.x]
+        node-version: [8.x, 9.x, 10.x, 12.x, 14.x]
 
     steps:
     - uses: actions/checkout@v2
@@ -26,8 +24,6 @@ jobs:
     - run: npm test
     - run: npm run lint
     - run: npm run cs-check
-      env:
-        CI: true
 
   deploy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v2.1.2
+        with:
+          node-version: 14.x
+      - run: npm install
+      - run: npm run build
+      - run: npm test
+      - run: npm run lint
+      - run: npm run cs-check
+      - run: npx lerna publish from-package --yes
```
epicfaace, line 21: I use `from-package` instead of `from-git` so that it's possible to tag non-version-bump commits as new versions and still have the pipeline run properly.

.github/workflows/ci.yml
```diff
@@ -8,12 +8,10 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     strategy:
       matrix:
-        node-version: [8.x, 9.x, 10.x, 11.x, 12.x, 13.x]
+        node-version: [8.x, 9.x, 10.x, 12.x, 14.x]
```
stanlemon, line 14: What is the reason for keeping 8 & 9 around?

stanlemon (approved):

.github/workflows/ci.yml
epicfaace: Because we technically still have node 8 and 9 compatibility, so I don't want to remove them until we maybe add a breaking change and say we don't support those versions anymore.

stanlemon: Got it. Fwiw both of those versions of node no longer receive any support.

PR #2285, opened 2021-03-20
Change dependabot frequency to "monthly"

epicfaace:

Dependabot is too annoying 😆  

By removing this, we'll continue to get security updates, but just not other updates for each dependency patch release.


.github/dependabot.yml
```diff
@@ -1,16 +0,0 @@
...
-updates:
-  - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
-      interval: "daily"
-
-  - package-ecosystem: "npm"
-    directory: "/"
-    schedule:
-      interval: "daily"
```
erunion, line 11: Why not set it to `monthly` instead?

epicfaace: Good point, `monthly` seems to strike a good balance. My major concern was that we have too many dependabot PRs that are just patch updates and don't add anything valuable. However, checking monthly will probably alleviate this problem.

jimmycallin (approved):

PR #2310, opened 2021-03-27
Fix local build errors from @types/json-schema

paintedbicycle:
Updates `@types/json-schema` dependency to latest in order to fix local build errors.

More details: https://github.com/rjsf-team/react-jsonschema-form/issues/2259#issuecomment-806309555

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [12.x, 14.x, 15.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 15.x]
+        node-version: [12.x, 14.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "postinstall": "lerna bootstrap",
     "publish": "lerna publish",
     "changed": "lerna changed",
-    "test": "lerna run test",
+    "test": "lerna run --stream test",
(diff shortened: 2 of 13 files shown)


epicfaace:
Thanks -- I think you need to update `package-lock.json` in order to fix CI.

Also, I'm curious why this didn't break CI earlier -- ideally, we'd be able to catch issues like this through CI. Any thoughts on how we can change CI to replicate your circumstances so it can fail on the `@types/json-schema` errors?

paintedbicycle:
I'm happy to push up the lock files. I omitted them because I noticed they were huge diffs, so got concerned that maybe you weren't updating them on your end until a new release is cut or something like that. But I actually think the large lock file diffs part of the new npm 7 changes which introduced [`v2` of their lock file](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile). LMK if you'd prefer I downgrade to npm@6 or if it's ok to push npm@7 lock files.

I also wonder if this entire issue is around npm 7 changes which includes [peer dependency changes](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies)

Could be the answer to both my issue and also your CI stuff?

paintedbicycle:
I downgraded to npm@6 for now and pushed the lockfiles. Hopefully that fixes the checks for this PR

epicfaace:
Yes, thanks for adding npm v6 lock files. Ah, perhaps the issue is caused by npm v7, as I did notice that node 15 was failing on CI earlier. I'll try adding that build here.

paintedbicycle:
@epicfaace Feels like the checks are stuck. Might have to kick it with another build

epicfaace:
It looks like this PR still fails on Node 15, so I just removed it from CI for now.

I'm fine with merging this PR as-is -- the only drawback here is that we can't guarantee through CI that this kind of error won't happen again. On the other hand, if we are able to replicate this issue through CI and add a CI check for it, we can more reliably prevent it from regressing. Lmk if you have any thoughts on how else we could try replicating this on CI -- otherwise let's merge it.

epicfaace (approved):

paintedbicycle:
@epicfaace Cool, let's merge. Thank you!

PR #2312, opened 2021-03-27
Update peer dependency to support React 17

paintedbicycle:
This PR updates the peer dependency to support React 16 and 17 at the same time. 
Discussion [here](https://github.com/rjsf-team/react-jsonschema-form/issues/2104#issuecomment-806213821).

Tests: I've also introduced a _very_ basic starting point for testing with React 16 and React 17 at the same time, inspired by [this article](https://medium.com/welldone-software/two-ways-to-run-tests-on-different-versions-of-the-same-library-f-e-react-17-react-16-afb7f861d1e9). Note that we are using mocha, not jest and so `moduleNameMapper` was swapped for an external library, `rewiremock`.

Discussion around this change [here](https://github.com/rjsf-team/react-jsonschema-form/issues/2104#issuecomment-806653044).

You can run `npm run test` and `npm run test-react-17`. To see it working, you can `console.log(REACT_VERSION, React.version)` in the `Form_test.js` file when running each test and it'll tell you which React it imported. 

Current limitations: I'm not swapping `react-dom` and I'm not testing React 17 for tests not in this file. This is a POC for discussion or hopefully someone else can take it and expand upon it. I'd assume we'd want to decide if we want to test both React 16 and 17 across the entire sweet, or just a few tests.

I'm totally happy to just scrap this approach - just putting it up for discussion.

Lastly: this PR also relies on [this PR](https://github.com/rjsf-team/react-jsonschema-form/pull/2310). Once it lands, I'll merge upstream into this which should make this PR a lot smaller - just a couple files.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,7 +40,7 @@
     "react": ">=16"
   },
   "dependencies": {
-    "@types/json-schema": "^7.0.4",
+    "@types/json-schema": "^7.0.7",
     "ajv": "^6.7.0",
     "core-js-pure": "^3.6.5",
     "json-schema-merge-allof": "^0.6.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -16,6 +16,7 @@
     "start": "concurrently \"npm:build:* -- --watch\"",
     "tdd": "cross-env NODE_ENV=test mocha --require @babel/register --watch --require ./test/setup-jsdom.js test/**/*_test.js",
     "test": "cross-env BABEL_ENV=test NODE_ENV=test mocha --require @babel/register --require ./test/setup-jsdom.js test/**/*_test.js",
+    "test-react-17": "cross-env BABEL_ENV=test NODE_ENV=test USE_REACT_17=true mocha --require @babel/register --require ./test/setup-jsdom.js test/**/*_test.js",
     "test-coverage": "cross-env NODE_ENV=test nyc --reporter=lcov mocha --require @babel/register --require ./test/setup-jsdom.js test/**/*_test.js",
     "test-debug": "cross-env NODE_ENV=test mocha --require @babel/register --require ./test/setup-jsdom.js --debug-brk --inspect test/Form_test.js"
   },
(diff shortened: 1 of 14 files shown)


packages/core/package.json
```diff
@@ -90,13 +91,15 @@
...
     "react": "^16.9.0",
+    "react-17": "npm:react@^17.0.2",
     "react-codemirror2": "^4.1.0",
     "react-dom": "^16.9.0",
     "react-monaco-editor": "^0.26.2",
     "react-portal": "^4.2.0",
     "react-transform-catch-errors": "^1.0.0",
     "react-transform-hmr": "^1.0.1",
     "redbox-react": "^1.3.3",
+    "rewiremock": "^3.14.3",
```
epicfaace, line 102: I think we need to install `rewiremock` in all packages.

paintedbicycle:
Oh interesting. Yeah, I'm not too familiar with monorepos & learna. We're only using it in Core right now, but I can easily do that.

I am curious about your thoughts on overall approach - is it worthwhile going down this path?

paintedbicycle:
@epicfaace Wanted to ping you again on this.

In this PR, I started down a path where we run two sets of tests (current set of React 16 tests and a subset of tests run in React 17). It was based on your comment in the original issue. I realized that it will be a bit of trial and error for me to go through and get it fully working so I had paused and pinged you to make sure this is the right approach. What are your thoughts? If I finish this up will you then take it on for the rest of the repo?

Alternatively, just moving all tests to React 17 could be an easy alternative. A third path could be to drop the tests from this PR and just merge the original fix.

What would you like to see?

epicfaace:
@paintedbicycle apologies for the delay! I think the approach taken in this PR may be a bit too complicated, given that you would have to modify each test file to allow it to run in React 16 vs 17. I'm trying an alternate, perhaps simpler approach in this PR -- not sure if it works, but maybe if it works, we can combine it with this PR: https://github.com/rjsf-team/react-jsonschema-form/pull/2332

If we can't find a simple approach to do it, then I'm just fine with changing the dev dependencies to use React 17.

paintedbicycle:
@epicfaace Yup, agreed! I like the approach in https://github.com/rjsf-team/react-jsonschema-form/pull/2332 - seems a lot cleaner. Closing this one in favour of that

PR #2324, opened 2021-04-03
Jsdom bump

jimmycallin:

Noticed you were running quite an old version of jsdom, so thought I would help bumping it to latest :)


packages/core/test/Form_test.js
```diff
@@ -375,7 +375,10 @@ describeRepeated("Form common", createFormComponent => {
   });
 
   describe("Custom submit buttons", () => {
-    it("should submit the form when clicked", done => {
+    // Skipping test due to regression in JSDOM  that causes it to not handle
+    //  form subsmissions from click events properly
+    // https://github.com/jsdom/jsdom/issues/3117
+    it.skip("should submit the form when clicked", done => {
```
jimmycallin, line 381: I spent some time trying to work around this regressions, but it wasn't easy. I chose to skip this test instead, in hope that it will be fixed soon. https://github.com/jsdom/jsdom/issues/3117

packages/core/test/NumberField_test.js
```diff
@@ -274,7 +278,7 @@ describe("NumberField", () => {
             sinon.assert.calledWithMatch(onChange.lastCall, {
               formData: test.output,
             });
-            expect($input.value).eql(test.input);
+            expect($input.getAttribute("value")).eql(test.input);
```
jimmycallin, line 281: Not entirely sure why the behaviour of `input.value` has changed, but using `input.getAttribute("value")` and `input.defaultValue` still seems to work as expected. Chose to use `input.getAttribute("value")` instead of digging deeper of what's up with the input.value regression.

packages/core/test/test_utils.js
```diff
@@ -56,6 +56,6 @@ export function describeRepeated(title, fn) {
 
 export function submitForm(node) {
   act(() => {
-    node.querySelector("button[type=submit]").click();
+    Simulate.submit(node);
```
jimmycallin, line 59: This is due to the same regression with buttons no longer sending submit events as mentioned above https://github.com/jsdom/jsdom/issues/3117

packages/core/test/uiSchema_test.js
```diff
@@ -1201,11 +1204,11 @@ describe("uiSchema", () => {
           schema,
           uiSchema,
           formData: {
-            foo: 3.14,
+            foo: 13.14,
```
jimmycallin, line 1207: The input range has a minimum of 10. The fact that JSDOM let you set 3.14 before was faulty, so I increased the value to within the valid input range.

packages/core/test/Form_test.js
epicfaace: Hmm, I'm not comfortable with disabling tests unless we really need to do so. Maybe we could try upgrading to a version of jsdom before the regression?

jimmycallin: Spent some time on this again and actually figured out the issue - apparently submit events are not fired from buttons if they are disconnected from the document tree. Mind reviewing again? :)

packages/core/test/uiSchema_test.js
```diff
@@ -533,7 +533,10 @@ describe("uiSchema", () => {
     expect(node.querySelector("div.help-block").textContent).eql("plop");
   });
 
-  describe("ui:focus", () => {
+  // Skipping due to JSDOM regression introduced in 15.2.0
+  // <input autofocus> no longer sets activeElement correctly
+  // https://github.com/jsdom/jsdom/issues/3041
+  describe.skip("ui:focus", () => {
```
epicfaace, line 539: Can we instead just use a JSDOM version less than 15.2.0? It's fine to have workarounds to work around JSDOM regressions, but I don't really want to skip tests because of regressions like https://github.com/jsdom/jsdom/issues/3041.

packages/core/test/Form_test.js
epicfaace: Great!

packages/core/test/uiSchema_test.js
jimmycallin: Sure, I reverted to jsdom 15.1.1. Would be nice to upgrade to latest jsdom, since I was thinking about migrating core to jest which uses jsdom 16, so we use the same test runner in all packages, but I understand your point.

jimmycallin: @epicfaace nevermind, i figured it out :) similar issue as to the other test

packages/core/test/NumberField_test.js
```diff
@@ -209,7 +211,9 @@ describe("NumberField", () => {
           formData: 2,
         });
 
-        expect(node.querySelector(".field input").value).eql("2");
+        expect(node.querySelector(".field input").getAttribute("value")).eql(
```
epicfaace, line 214: It's a bit strange, because if you try doing `.value` for this line of code, everything works. Any idea why it works for this test case, but not for the other test cases in this file?

jimmycallin: honestly no clue, if you want me to dig deeper before merge I can try to figure it out

jimmycallin: figured it out - apparently "2." is not a valid value in input number fields, but for some reason we still support it which causes some issues. that jsdom supported this before was faulty behavior.

image

jimmycallin: i would honestly prefer to remove support for "2." since it is not compatible with the HTML standard, but I just added a comment for now.

epicfaace: Great, thanks for looking into this!

epicfaace (approved):

PR #123, opened 2016-04-05
Fixes #122: Better DateTimeWidget.

n1k0:
Refs #122.

r=? @magopian @leplatrem

PR #2330, opened 2021-04-09
allow booleans to have descriptions

### LINKED ISSUE #2203: Checkbox doesn't support descriptions for material-ui / bootstrap-4
### Prerequisites

### Description

The description of the the boolean field widget is missing when the theme is material-ui and bootstrap-4

### Steps to Reproduce

1. Go to the playground https://rjsf-team.github.io/react-jsonschema-form/
2. Select the `Widgets` from the top row and observe the boolean field on the right when the default theme is selected. The top widget (checkbox) has a visible description of `This is the checkbox-description`. 
3. If you select the `material-ui` or `bootstrap-4` theme the description disappears for the checkbox widget. The description is still there for the radio widget and the select widget.
4. Is there a way to bring it back?

#### Expected behavior

I expect to see (somewhere) the description of the boolean field widget

#### Actual behavior

The description disappears for the checkbox widget

### Version

These are the packages that I have installed:
    "@rjsf/core": "2.4.1",
    "react": "16.14.0",
    "react-icons": "4.1.0",
    "react-dom": "16.14.0",
    "react-bootstrap": "1.4.3",
    "@rjsf/bootstrap-4": "2.4.1",
    "@material-ui/core": "4.11.2",
    "@material-ui/icons": "4.11.2",
    "@rjsf/material-ui": "2.4.1",

cfluke-twilio:

Descriptions for boolean checkboxes were being ignored.

fixes #2203 

### Checklist

### Deploy preview

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -397,9 +397,6 @@ export function getDisplayLabel(schema, uiSchema, rootSchema) {
   if (schema.type === "object") {
     displayLabel = false;
   }
-  if (schema.type === "boolean" && !uiSchema["ui:widget"]) {
-    displayLabel = false;
-  }
   if (uiSchema["ui:field"]) {
     displayLabel = false;
   }


cfluke-twilio:
Is the correct fix to this issue, to add the schema.description check in [CheckboxWidget](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/material-ui/src/CheckboxWidget/CheckboxWidget.tsx)?
to make it similar to [core](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/widgets/CheckboxWidget.js#L27)?

this appears to be a problem in material-ui, antd, and sematic-ui as well. 

The current solution does currently make it double appear in default theme. This seems to be caused by the WrapIfAdditional and [DefaultTemplate](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/fields/SchemaField.js#L139) adding the description as well.

epicfaace (commented):
Are you sure this is the right approach? This same function appears in utils.js -- https://github.com/rjsf-team/react-jsonschema-form/blob/1ae4f39c9ca1c3538a935c1afdab92ca0abe0b77/packages/core/src/utils.js#L389-L407 -- is there another way to fix this?

cfluke-twilio:
Hi @epicfaace, see my comment above.
The easiest, simplest fix for material ui is to add:
`{schema.description && (
        <DescriptionField description={schema.description} />
      )}`
      
into the CheckBox widget itself. but the problem is in 2 other theme libraries as well. So I was looking at a more global solution to keep the pattern consistent across the various themes.

Should I move ahead with only the material-ui fix to move this forward?

epicfaace:
> From a library design perspective, do you want widgets to be responsible for their own descriptions? or allow WrapIfAdditional pattern?

Yeah, let's move ahead with the material-ui fix to move this forward for now. Unfortunately, as you note, there is an inconsistency between themes in which some themes have widgets responsible for their own descriptions, while others don't. Maybe we can consider improving that to make it consistent in a future major version release.

epicfaace (commented):
Thanks, can you add a snapshot test for this as well?

epicfaace (approved):
Looks good -- thanks!

epicfaace:
> From a library design perspective, do you want widgets to be responsible for their own descriptions? or allow WrapIfAdditional pattern?

Sorry, @cfluke-twilio I'm a bit confused. Why do descriptions work for the TextWidget in material-ui, even though there's no code like the one in this PR for TextWidget?

cfluke-twilio:
TextWidget is calling the [getDisplayLabel](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/utils.js#L389) from the core utils [here](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/material-ui/src/TextWidget/TextWidget.tsx#L43)
The FieldTemplate uses that getDisplayLabel to determine if it should add a Typography and rawDescription [here](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/material-ui/src/FieldTemplate/FieldTemplate.tsx#L45)

The core utils has this return false for checkboxes.

Morriz:
bump

Morriz:
We have created a lot of hacky workarounds to overcome all of these weird exceptions. Maybe we should consider adding to your core, as we depend on it sooooo much @epicfaace. Thanks for all the hard work. I do get the feeling that the maintainers are all paid volunteers, and it would be awesome if this project could get adopted by a major corp to start funding development.

PR #2332, opened 2021-04-14
Test react 16 and 17, update peer dependency to support React 17

epicfaace:
Test react 16 and 17 -- potential alternative to the approach in https://github.com/rjsf-team/react-jsonschema-form/pull/2312.

Here, we use `lerna add` to replace React 16 with 17 in some workflow runs.

Feedback appreciated.


epicfaace:
@paintedbicycle

paintedbicycle:
@epicfaace Cool! Looks like they passed after that?

epicfaace:
Yeah, but it's a bit strange that by undoing the `"^16.x || ^17.x"` change (see this commit: https://github.com/rjsf-team/react-jsonschema-form/pull/2332/commits/464392299d5f93d54aa9ed43b89e6b1e292a4df7), all the checks still pass. Ideally, they should fail because submitting the form programmatically should fail without that change on React 17. Do you know why they passed? Maybe we need to update the test for submitting the form programmatically?

paintedbicycle:
That might be.

When I looked through, I had assumed that [this](https://github.com/rjsf-team/react-jsonschema-form/blob/16db739eeb0902413cbf0fccb73c1ef8ef1708e9/packages/core/test/Form_test.js#L812) was going to test it. However it may be that we need to create a form with no submit button, add a custom submit button that then use the `Simulate.click` event on that button to actually test this.

epicfaace:
@paintedbicycle do you think you'd have the availability to work off this PR to create one that adds that test in?

paintedbicycle:
Yeah, I'll try to spend some time on it next week. I'm not super familiar with the testing, so I find it a bit difficult. Biggest issue I can see is attaching the button to the form via a ref. Easy enough with the library, but the way the test uses createFormComponent it looks like all you can pass is a prop, not attach a ref. I'll have to investigate if that's different or not and now to do it

paintedbicycle:
Having trouble getting this branch to build

```
@rjsf/playground: Module not found: Error: Can't resolve './fluent-ui.cjs.development.js'
```

Tried fresh install several times. Can you merge master into this branch and try to build?

paintedbicycle:
Oh I didn't see that you did the merge. Thank you! It fixed the build for me.

I've created a test for programatically submitting the form [here](https://github.com/paintedbicycle/react-jsonschema-form/commit/3640896cb4af58cdc65f5fb3afa692406bd57369). It's failing so I have to investigate a bit more - I think there's something about the `.submit()` as it's complaining that `ReferenceError: CustomEvent is not defined` which I believe is my test, not the feature, but I'll need to dig in. LMK if you spot anything I'm missing.

epicfaace:
> as it's complaining that `ReferenceError: CustomEvent is not defined` which I believe is my test, not the feature, but I'll need to dig in. LMK if you spot anything I'm missing.

Hmm, CustomEvent is used here (https://github.com/rjsf-team/react-jsonschema-form/blob/3ec17f1c0ff40401b7a99c5e9891ac2834a1e73f/packages/core/src/components/Form.js#L409), and it's a JavaScript API that usually should be available (https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) so perhaps that helps debug.

paintedbicycle:
Yeah when I found the MDN doc, that's when I gave up, haha. Feels like something that's going to need some debugging. I'll dig in this weekend.

epicfaace:
Maybe https://libertyseeds.ca/2015/07/17/React-Jest-and-CustomEvent-testing/ might fix this?

paintedbicycle:
@epicfaace The fact that it doesn't exist in the node environment was a good tip. I looked at the post you sent, but they're using jest and I don't think we are in code, so I grabbed a customEvent polyfill. The test passes now.

paintedbicycle:
@epicfaace When I set peerDependencies back:

```
  "peerDependencies": {
    "react": "^16.x"
  },
```

the tests still pass. Am I right in assuming that this is an issue? We should see a failure, correct?

paintedbicycle:
Oh wait, no, I'm not running 17 locally anymore. We should try in the CI

paintedbicycle:
Hmm, ok I did that test and it still passes. Running out of ideas here

paintedbicycle:
@epicfaace Can you confirm that the steps I did in https://github.com/rjsf-team/react-jsonschema-form/pull/2351 should have resulted in a fail? I added the examples of submission via a form ref from the docs and then tried with the old peer dependencies. I'm not experienced with GitHub actions, nor mocha testing, but I believe the way you had it set up, adding that test should have passed in react 17 but failed in react 16 GitHub action environments, correct?

epicfaace:
> but I believe the way you had it set up, adding that test should have passed in react 17 but failed in react 16 GitHub action environments,

Sorry, a bit confused. I thought the submission with a form ref should have passed in react 16 but failed in react 17? That was the original issue in https://github.com/rjsf-team/react-jsonschema-form/issues/2104. Is that what you meant?

paintedbicycle:
Yep! I was wondering if you spotted anything weird in the setup - the tests are passing but shouldn't be, given that this issue is showing up in a real environment. Trying to sort out what's going on

PR #2339, opened 2021-04-17
remove componentWillReceiveProps from altdatewidget

jimmycallin:

Another step towards making RJSF strict mode compatible, this time without functional components :)

### Checklist

### Deploy preview

### DIFF
--- a/packages/core/src/components/widgets/AltDateWidget.js
+++ b/packages/core/src/components/widgets/AltDateWidget.js
@@ -63,8 +63,12 @@ class AltDateWidget extends Component {
     this.state = parseDateString(props.value, props.time);
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    this.setState(parseDateString(nextProps.value, nextProps.time));
+  componentDidUpdate(prevProps, prevState) {
+    if (
+      prevProps.value !== parseDateString(this.props.value, this.props.time)
+    ) {
+      this.setState(parseDateString(this.props.value, this.props.time));
+    }
   }
 
   shouldComponentUpdate(nextProps, nextState) {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/widgets/AltDateWidget.js
```diff
@@ -63,8 +63,12 @@ class AltDateWidget extends Component {
     this.state = parseDateString(props.value, props.time);
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    this.setState(parseDateString(nextProps.value, nextProps.time));
+  componentDidUpdate(prevProps, prevState) {
```
epicfaace, line 66: I think it'd be better to instead use getDerivedStateFromProps (see https://stackoverflow.com/questions/49449527/why-use-getderivedstatefromprops-instead-of-componentdidupdate)

epicfaace (commented):
Great, thanks! 🙂

packages/core/src/components/widgets/AltDateWidget.js
jimmycallin: the component needs to be able to compare previous props to current props to determine if it should update its state. with getDerivedStateFromProps, it only has its current props available. I could add the previous props as a "lastProps" property in the state and compare it that way, if you prefer that, but this felt better for me.

(the fact that we need to compare previous to current props to determine whether state should update feels like faulty design, see my comment here, https://github.com/rjsf-team/react-jsonschema-form/pull/2010#issuecomment-678673604, but I didn't want to do any drastic changes in this pr)

epicfaace: Sorry, yeah, you're completely right! `componentDidUpdate` is definitely better.

I agree with you about faulty design; as you said, though, let's keep that issue for another PR / future breaking change. 🙂

epicfaace (approved):

PR #2360, opened 2021-05-04
Help elements need a predictable ID

bradhugh:

When a ui:help is added to uiSchema for a property, the generated p or div tag needs to have an id attribute so it addressable.  This same thing already exists for description fields. This is needed for association with the corresponding form fields for accessibility (like adding aria-describedby="helpId") on the associated input fields.

### Checklist

### Deploy preview

### DIFF
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -85,14 +85,22 @@ function LabelInput(props) {
 }
 
 function Help(props) {
-  const { help } = props;
+  const { id, help } = props;
   if (!help) {
     return null;
   }
   if (typeof help === "string") {
-    return <p className="help-block">{help}</p>;
+    return (
+      <p id={id} className="help-block">
+        {help}
+      </p>
+    );
   }
-  return <div className="help-block">{help}</div>;
+  return (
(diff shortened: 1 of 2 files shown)


epicfaace (commented):
Thanks -- are there any similar changes that could / should be done for other themes' help elements as well?

bradhugh:
> Thanks -- are there any similar changes that could / should be done for other themes' help elements as well?

It doesn't look like many of the templates actually override the Help field.  I don't know anything about semantic-ui, but take a look at HelpField.js in semantic-ui.  Right now it's got the following:
`<HelpField helpText={rawHelp} id={id} />`

I'll try to dig a bit and understand where it's getting "id" from in this context.

bradhugh:
Yep, that's what I thought. It's actually duplicating the id here which is a bug. I'll make a change to fix this but while I'm here.

image

bradhugh:
@epicfaace anything else I need to do here?

epicfaace (approved):
Thanks! This looks great.

epicfaace (commented):
Actually, maybe we should add some snapshot tests for the themes with help fields? Just so we make sure that in the future, help fields always have ids that end with "__help" and this doesn't regress.

PR #2366, opened 2021-05-07
Make `onBlur` and the `onFocus` accept `any` types for value

that-jpg:

I have some custom fields that have a array of string as value (I'm thinking about multiselect or a checkbox group type of component), I can pass its value to the onChange without any issue, but the same is not true for the `onBlur` event, because their typescript definitions are not the same.

I can think some scenarios where the field could pass a object on those events so my suggestion would be to make them all equal to the `onChange` with a `any`.

### Checklist

* [-] **I'm updating documentation**
* [-] **I'm adding a new feature**
  - [-] I've updated the playground with an example use of the feature

### Deploy preview


epicfaace (approved):
This looks good, thanks!

PR #2386, opened 2021-05-19
Add predictable ID for help fields, aria describedby attributes

### LINKED ISSUE #959: Provide accessible error messages for AT users
### Prerequisites

### Description

In order to providy more accessibility to our screen reaser users, we would like to add an `aria-describedby`-attribute to input fields that reference the inline error message (e.g. "is a required property") with an id of this error meesage when validating the form. This would cause the AT to read out the error message under the input field when the user tabs through the form elements.

Is there a way to configure `aria-describedby`-attributes and to enable ids to error messages?

Any other ideas to solve this problem are appreciated. :)

For example, instead of this markup from the [live playground](https://mozilla-services.github.io/react-jsonschema-form/)

```html
<div class="form-group field field-string field-error has-error has-danger">
  <label class="control-label" for="root_lastName">Last name*</label>
  <div></div>
  <input type="text" 
     class="form-control" 
     value="" 
     id="root_lastName" 
     label="Last name" 
     required="" 
     placeholder="">
  <div>
    <p></p>
    <ul class="error-detail bs-callout bs-callout-info">
      <li class="text-danger">is a required property</li>
    </ul>
  </div>
  <div></div>
</div>
```

it should look like this:
```html
<div class="form-group field field-string field-error has-error has-danger">
  <label class="control-label" for="root_lastName">Last name*</label>
  <div></div>
  <input type="text" 
     class="form-control" 
     value="" 
     id="root_lastName" 
     label="Last name" 
     required="" 
     placeholder="" 
<!-- CHANGE -->
     aria-describedby="error-message-123456789" 
<!-- /CHANGE -->
     >
  <div>
    <p></p>
    <ul class="error-detail bs-callout bs-callout-info">
      <li class="text-danger" 
<!-- CHANGE -->
          id="error-message-123456789"
<!-- /CHANGE -->
          >is a required property</li>
    </ul>
  </div>
  <div></div>
</div>
```

### Steps to Reproduce

1. Install a screen reader like [NVDA](https://www.nvaccess.org/download/)
2. Launch the screen reader
3. Open the [live playground](https://www.nvaccess.org/download/) in your browser (The first input field ('First Name') should be focused)
4. Use the <kbd>TAB</kbd> key to navigate to the required 'Last Name'-input-field.
5. Delete the value --> The field gets red and a message appears below the input field ('is a required property')
6. Listen to the screen reader.

#### Expected behavior

The screen reader should read out also the message that is configured here ("is a required property").

#### Actual behavior

The screen reader tells you that this field is required and should be filled, since it recognizes the HTML5 validation functionality of the browser. But the message is not being read out.

ArnoSaine:

Make screen reader read field errors, description and help text.

Fixes #959

### Checklist

### Deploy preview

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -433,6 +433,12 @@ declare module '@rjsf/core' {
         ): number;
 
         export function schemaRequiresTrueValue(schema: JSONSchema7): boolean;
+
+        export function descriptionId(id: string): string;
+
+        export function helpId(id: string): string;
+
+        export function ariaDescribedBy(id: string): { 'aria-describedby': string };
     }
 }
 
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -440,7 +440,7 @@ declare module '@rjsf/core' {
 
         export function errorsId(id: string): string;
 
-        export function ariaDescribedBy(id: string): { 'aria-describedby': string };
+        export function ariaDescribedBy(id: string): string;
     }
 }
 
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -438,6 +438,8 @@ declare module '@rjsf/core' {
 
         export function helpId(id: string): string;
 
+        export function errorsId(id: string): string;
+
         export function ariaDescribedBy(id: string): { 'aria-describedby': string };
     }
 }
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -116,7 +116,7 @@ function ErrorList(props) {
 
   return (
     <div>
-      <ul className="error-detail bs-callout bs-callout-info" id={errorsId(id)}>
+      <ul className="error-detail bs-callout bs-callout-info" id={id}>
         {errors
           .filter(elem => !!elem)
           .map((error, index) => {
@@ -331,7 +331,7 @@ function SchemaFieldRender(props) {
     rawDescription: description,
     help: <Help id={helpId(id)} help={help} />,
     rawHelp: typeof help === "string" ? help : undefined,
-    errors: <ErrorList errors={errors} id={id} />,
+    errors: <ErrorList id={errorsId(id)} errors={errors} />,
     rawErrors: errors,
     id,
     label,
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -13,6 +13,9 @@ import {
   deepEquals,
   getSchemaType,
   getDisplayLabel,
+  ariaDescribedBy,
+  helpId,
+  descriptionId,
 } from "../../utils";
 
 const REQUIRED_FIELD_SYMBOL = "*";
@@ -64,7 +67,7 @@ function Label(props) {
     return null;
   }
   return (
-    <label className="control-label" htmlFor={id}>
+    <label className="control-label" htmlFor={id} {...ariaDescribedBy(id)}>
       {label}
       {required && <span className="required">{REQUIRED_FIELD_SYMBOL}</span>}
     </label>
@@ -80,6 +83,7 @@ function LabelInput(props) {
       id={id}
       onBlur={event => onChange(event.target.value)}
       defaultValue={label}
+      {...ariaDescribedBy(id)}
     />
   );
 }
@@ -91,7 +95,7 @@ function Help(props) {
   }
   if (typeof help === "string") {
     return (
-      <p id={id} className="help-block">
+      <p className="help-block" id={helpId(id)}>
         {help}
       </p>
     );
@@ -318,13 +322,13 @@ function SchemaFieldRender(props) {
   const fieldProps = {
     description: (
       <DescriptionField
-        id={id + "__description"}
+        id={descriptionId(id)}
         description={description}
         formContext={formContext}
       />
     ),
     rawDescription: description,
-    help: <Help id={id + "__help"} help={help} />,
+    help: <Help help={help} id={id} />,
     rawHelp: typeof help === "string" ? help : undefined,
     errors: <ErrorList errors={errors} />,
     rawErrors: errors,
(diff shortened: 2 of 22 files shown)


epicfaace:
Thanks for the PR. I wonder if we can merge the changes done in https://github.com/rjsf-team/react-jsonschema-form/pull/2360 into this. Specifically, can you make sure that all the help / description / title field ids are using the id utility functions you defined in utils.js? Specifically, we might need to change `packages/semantic-ui/src/FieldTemplate/FieldTemplate.js`, for example.

Can we also add snapshot tests as needed for the affected themes, because now the help fields have different ids?

ArnoSaine:
Thank you for the suggestions. I searched for other mentions of `__description`, `__help` and `__errors`. `packages/semantic-ui/src/FieldTemplate/FieldTemplate.js` was the only one not using the utility function. it's now fixed.

I updated the affected snapshots and added more material-ui updates.

packages/core/src/utils.js
```diff
@@ -1265,3 +1265,13 @@ export function schemaRequiresTrueValue(schema) {
...
 }
+
+export const descriptionId = id => `${id}__description`;
+
+export const helpId = id => `${id}__help`;
+
+export const errorsId = id => `${id}__errors`;
+
+export const ariaDescribedBy = id => ({
+  "aria-describedby": `${errorsId(id)} ${descriptionId(id)} ${helpId(id)}`,
```
epicfaace, line 1276: What if, for example, the help text is empty? Would it be correct in that case to still include helpId within aria-describedby, or should it be removed in that case?

epicfaace, line 1276: For consistency, I wonder if it's better to just return a string here instead of a dict.

packages/core/src/components/fields/SchemaField.js
```diff
@@ -68,10 +68,7 @@ function Label(props) {
     return null;
   }
   return (
-    <label
-      className="control-label"
-      htmlFor={id}
-      aria-describedby={ariaDescribedBy(id)}>
+    <label className="control-label" htmlFor={id}>
```
epicfaace: Why was the original aria-describedby here incorrect?

packages/core/src/utils.js
ArnoSaine: It is an issue. As far as I know, the specification does not say anything about missing element references. Still, https://validator.w3.org/ gives an error, if for example help element is missing.

VoiceOver reads only the existing elements without a problem when the input gets focus, as expected. I guess other tools work similarly. Order of the IDs also matter: I think it's best to have errors first. If field has errors, user gets that infomation first, without need to go through description and help every time.

I'd suggest we add all IDs, because it still improves accessibility.

ArnoSaine: Done! You are right. It's better that way.

packages/core/src/components/fields/SchemaField.js
ArnoSaine: I added it by mistake here: https://github.com/rjsf-team/react-jsonschema-form/pull/2386/commits/d332de122ed74dd534334d0c2835a23ce1577811

Original is correct: `<input>` has `id`, `<label>` has `for`. `aria-describedby` goes also to the `<input>`

packages/core/src/utils.js
epicfaace: I see, thanks!

packages/material-ui/src/FieldTemplate/FieldTemplate.tsx
```diff
@@ -43,22 +43,27 @@ const FieldTemplate = ({
...
           </Typography>
         ) : null}
         {rawErrors.length > 0 && (
-          <List dense={true} disablePadding={true}>
+          <List id={utils.errorsId(id)} dense={true} disablePadding={true}>
             {rawErrors.map((error, i: number) => {
               return (
                 <ListItem key={i} disableGutters={true}>
-                  <FormHelperText id={id}>{error}</FormHelperText>
+                  <FormHelperText>{error}</FormHelperText>
```
epicfaace, line 58: Why remove the id here?

packages/core/index.d.ts
```diff
@@ -433,6 +433,14 @@ declare module '@rjsf/core' {
         ): number;
 
         export function schemaRequiresTrueValue(schema: JSONSchema7): boolean;
+
+        export function descriptionId(id: string): string;
```
epicfaace, line 437: Let's add one-line comments before each new function explaining what they do.

packages/material-ui/src/FieldTemplate/FieldTemplate.tsx
ArnoSaine: It was inside a `.map`. If there were multiple errors, it was not unique. Same id was also used with the `rawHelp` text. I moved the id to the parent `<List>` element and added suffix to both help and error list ids.

packages/core/index.d.ts
ArnoSaine: Sure! I commented `packages/core/src/utils.js` as most of the comments are in that file.

packages/material-ui/src/FieldTemplate/FieldTemplate.tsx
epicfaace: Thanks! Can you add a short section in the "v3 upgrade guide" doc (within the documentation) that describes some of these breaking changes in id's? Just a quick list of what exactly were changed, so that someone who upgrades to v3 will know how the ids will change in the form.

ArnoSaine: Could we consider this id change to be an internal fix? Element ids must be unique within the document and it is an undocumented feature. Here's a screenshot of the issue from the playground:
Screenshot 2021-05-24 at 15 50 29

In this PR we let the input use the generated `id` (no change). To fix the case of non-unique ids and make `aria-describedby` behave correctly, we suffix help and errors ids (in material-ui FieldTemplate).

epicfaace:
Hi @ArnoSaine it's been a while, do you remember where we are with this PR? Is this ready except for making it a breaking change / adding it to a v5.x upgrade guide?

heath-freenome:
Reimplemented in #3380

PR #2398, opened 2021-05-26
Fix onSubmit type

saadq:
Currently, the onSubmit only allows one parameter. The docs mention that
onSubmit should receive a second parameter for the native form submit
event, so the type has been updated to include it.

The current type for `onSubmit` is inaccurate and leads to type errors if trying to use the second parameter in your submit handler inside a TypeScript project.

### Checklist

I believe the below checklist isn't applicable for this change, so will not be updating it.

### Deploy preview

Once CI finishes, you should be able to view a deploy preview of the playground from this PR at this link: https://deploy-preview-2398--rjsf.netlify.app/

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -40,7 +40,7 @@ declare module '@rjsf/core' {
         onChange?: (e: IChangeEvent<T>, es?: ErrorSchema) => any;
         onError?: (e: any) => any;
         onFocus?: (id: string, value: any) => void;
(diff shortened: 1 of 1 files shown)


epicfaace (approved):
Thanks!

saadq:
Np, thanks for the merge. Just wanted to check with the regular release cycle is like? Do you merge in type changes as a patch release? Or do you only do releases with several changes at once?

epicfaace:
Usually periodic released with several changes at once.

PR #2405, opened 2021-05-27
Hide grid for hidden fields (material-ui, bootstrap-4, fluent-ui); pass hidden prop to ObjectFieldTemplate

### LINKED ISSUE #1974: Hidden field with Material-ui display large Spacing
### Prerequisites

### Description

Hidden field with Material-ui theme : field is hidden but a spacing is display between field

### Steps to Reproduce

1. Create 3 fields 
2. Hide the second field  : "mySecondField": { "ui:widget": "hidden"  }
3. Use Material-UI has theme

You can see that on the playground https://rjsf-team.github.io/react-jsonschema-form/
Select "widget" and use Material-ui as theme, see space where the Secret hidden field is.

#### Expected behavior

Field is hide without extra Blank

#### Actual behavior

 Field is hide but it remains large spacing between first and third field , this is ok with other Theme

### Version

1.8.1

Fox32:

`hidden` fields aren't hidden for material-ui as the grid around them is still displayed. This hides the grid for hidden fields.
Implementations follows the one that already exists for ant design.

**Before** (note the space before the disabled example)

Bildschirmfoto 2021-05-27 um 11 17 55

**After**

Bildschirmfoto 2021-05-27 um 11 20 59

Closes #1974
Fixes https://github.com/rjsf-team/react-jsonschema-form/issues/1683

### Checklist

### Deploy preview

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -215,6 +215,7 @@ declare module '@rjsf/core' {
             name: string;
             disabled: boolean;
             readonly: boolean;
+            hidden: boolean;
         }[];
         onAddClick: (schema: JSONSchema7) => () => void;
         readonly: boolean;
--- a/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
+++ b/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
@@ -53,6 +53,8 @@ const ObjectFieldTemplate = ({
       )}
       <Grid container={true} spacing={2} className={classes.root}>
         {properties.map((element: any, index: number) =>
+          // Remove the <Grid> if the inner element is hidden as the <Grid>
+          // itself would otherwise still take up space.
           isHidden(element) ? (
             element.content
           ) : (
--- a/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
+++ b/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
@@ -33,6 +33,9 @@ const ObjectFieldTemplate = ({
 }: ObjectFieldTemplateProps) => {
   const classes = useStyles();
 
+  const isHidden = (element: any): boolean =>
+    element.content.props.uiSchema["ui:widget"] === "hidden";
+
   return (
     <>
       {(uiSchema['ui:title'] || title) && (
@@ -49,16 +52,19 @@ const ObjectFieldTemplate = ({
         />
       )}
       <Grid container={true} spacing={2} className={classes.root}>
-        {properties.map((element: any, index: number) => (
-          <Grid
-            item={true}
-            xs={12}
-            key={index}
-            style={{ marginBottom: '10px' }}
-          >
-            {element.content}
-          </Grid>
-        ))}
+        {properties.map((element: any, index: number) =>
+          isHidden(element) ? (
+            element.content
+          ) : (
+            <Grid
+              item={true}
+              xs={12}
+              key={index}
+              style={{ marginBottom: "10px" }}>
+              {element.content}
+            </Grid>
+          )
+        )}
         {canExpand(schema, uiSchema, formData) && (
           <Grid container justify='flex-end'>
             <Grid item={true}>
--- a/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
+++ b/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
@@ -33,9 +33,6 @@ const ObjectFieldTemplate = ({
 }: ObjectFieldTemplateProps) => {
   const classes = useStyles();
 
-  const isHidden = (element: any): boolean =>
-    element.content.props.uiSchema["ui:widget"] === "hidden";
-
   return (
     <>
       {(uiSchema['ui:title'] || title) && (
@@ -52,10 +49,10 @@ const ObjectFieldTemplate = ({
         />
       )}
       <Grid container={true} spacing={2} className={classes.root}>
-        {properties.map((element: any, index: number) =>
+        {properties.map((element, index) =>
           // Remove the <Grid> if the inner element is hidden as the <Grid>
           // itself would otherwise still take up space.
-          isHidden(element) ? (
+          element.hidden ? (
             element.content
           ) : (
             <Grid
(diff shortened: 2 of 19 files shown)


Fox32:
hey @agustin107, can you take a look?

Morriz:
politely bumping @agustin107 ;)

or can other ppl review @Fox32 ?

Fox32:
Not sure, @agustin107 is set as a codeowner, maybe @epicfaace can review it too

packages/material-ui/test/Form.test.tsx
```diff
@@ -197,6 +197,18 @@ describe("single fields", () => {
...
   });
+  test("hidden field", () => {
+    const schema: JSONSchema7 = {
+      type: "string",
+    };
+    const uiSchema: UiSchema = {
+      "ui:widget": "hidden",
+    };
+    const tree = renderer
+      .create(<Form schema={schema} uiSchema={uiSchema} />)
```
epicfaace, line 208: Should we add tests for a hidden object and hidden array as well?

packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
```diff
@@ -49,16 +52,19 @@ const ObjectFieldTemplate = ({
...
-            item={true}
-            xs={12}
-            key={index}
-            style={{ marginBottom: '10px' }}
-          >
-            {element.content}
-          </Grid>
-        ))}
+        {properties.map((element: any, index: number) =>
+          isHidden(element) ? (
```
epicfaace, line 56: Why do we need this change in ObjectFieldTemplate -- what does it do? Isn't the FieldTemplate change enough?

Fox32: Here (`ObjectFieldTemplate`) a `<Grid>` with a margin is added, which means that a hidden element would still take up space. We have to remove the `<Grid>` if the field is hidden.

Fox32: Do you think the code needs a comment to explain this?

epicfaace: Oh, I see. Instead of doing this, could we check if `element.content === null` and then remove the grid? I'm asking because it's currently possible to specify widgets in different ways than `ui:widget` (we can do `{'ui:options': {'widget': ...`) so it's good not to tie the implementation specifically to `ui:widget`.

Can you also add a comment describing why we're hiding the grid in this case?

packages/material-ui/test/Form.test.tsx
Fox32: Played around with both cases and they aren't fixed by this PR, so a test wouldn't help. But I could offer to include a fix for them too.

packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
Fox32: hmm `element.content === null` is not `true` in that case. There is still a field create, just a hidden one. Is there any other way I can use to support both cases without having to hard code it in that spot?

epicfaace: What is element.content equal to in that case?

packages/material-ui/test/Form.test.tsx
epicfaace: Okay, let's just not add it, and keep this PR limited in scope / easier to merge.

packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
Fox32:
```typescript
{
            content: React.ReactElement;
            name: string;
            disabled: boolean;
            readonly: boolean;
}
```
https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/index.d.ts#L213-L218 

If `@rjsf/core` would add a `hidden` property there too it would be an easy.

Fox32: Oh, just noticed that you ask for `.content`...  Not 100% sure, but I think it is a `SchemaField` from `core`.

Fox32: But in that case they would both evaluate the same condition: https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/fields/SchemaField.js#L307

epicfaace: Yeah, good point.

I wonder if we could still pass the `hidden` prop over to ObjectFieldTemplate from `@rjsf/core`, though, as this would reduce code duplication with checking ui:widget. How feasible / easy do you think that change would be?

epicfaace: Doing this would also help out with https://github.com/rjsf-team/react-jsonschema-form/pull/2175! (similar issue with bootstrap-4 theme)

Fox32: Had some trouble with `nanoid`, but I think I got it working now. Still have to postpone it till tomorrow.

Fox32: PR is updated. I couldn't find a good place in `core` to test it, so I skipped that. I replaced the old implementation in `antd`. I also went through all themes and fixed #2175 for bootstrap-4 and fluid-ui.

packages/core/index.d.ts
```diff
@@ -215,6 +215,7 @@ declare module '@rjsf/core' {
             name: string;
             disabled: boolean;
             readonly: boolean;
+            hidden: boolean;
```
epicfaace, line 218: Can you update the docs as well where we document the props passed over to ObjectFieldTemplate?

epicfaace (approved):
This looks great!

packages/core/index.d.ts
epicfaace: Done

Fox32: Thanks!

Issue #1964, opened 2020-08-03
Docs - "themes" link broken


epicfaace:
When I click on the "themes" title in the left navbar, it jumps to https://react-jsonschema-form.readthedocs.io/en/latest/usage/themes/ instead. It's the right page, but feels really jumpy because the highlighted menu item has just suddenly changed position.

I believe this was introduced by #1907.

epicfaace:
This actually happens to any link in the menu.

PR #2422, opened 2021-06-06
Docs: Fix jumpy menu

epicfaace:
Docs: fix jumpy menu issue, fixes https://github.com/rjsf-team/react-jsonschema-form/issues/1964; it turns out that this is because our theme doesn't like too nested menu items.

It now looks like this:

image


stanlemon (approved):
👍🏻

PR #2424, opened 2021-06-08
Replace shortid with nanoid

### LINKED ISSUE #2407: Remove dependency to deprecated `shortid`
### Prerequisites

### Description

There are dependencies to `shortid`, which is deprecated according to their [author](https://github.com/dylang/shortid#shortid-is-deprecated-because-the-architecture-is-unsafe-we-instead-recommend-nano-id-which-has-the-advantage-of-also-being-significantly-faster-than-shortid):

> shortid is deprecated, because the architecture is unsafe. we instead recommend Nano ID, which has the advantage of also being significantly faster than shortid

`shortid` also has dependecy to an old version of `nanoid`, which is incompatible with newer versions of `nanoid`. This causes issues when using Webpack 5 (which depends on a recent version of `nanoid`).

Replacing the few usages of `shortid` with `nanoid` should be quite straightforward, but there are possibly better ways to resolve this.

### Version

2.5.1

damianfrizzi:

Fixes #2407.
Remove dependencies to deprecated shortid and replace it with nanoid.
I have run `lerna clean` and then `learn bootstrap`. 

### Checklist


packages/semantic-ui/package.json
```diff
@@ -68,6 +67,7 @@
     "mini-css-extract-plugin": "^0.4.3",
     "mocha": "^8.3.2",
     "monaco-editor-webpack-plugin": "^1.7.0",
+    "nanoid": "^3.1.23",
```
epicfaace, line 70: nanoid should be added as a dependency, not a devDependency, for all packages.

epicfaace (changes requested):

packages/semantic-ui/package.json
damianfrizzi: Done

packages/playground/package.json
```diff
@@ -54,12 +54,12 @@
     "less": "^3.11.3",
     "less-loader": "^5.0.0",
     "lodash": "^4.17.21",
+    "nanoid": "^3.1.23",
```
epicfaace, line 57: We should be able to remove this dependency (because nanoid is already a dependency pulled in by @rjsf/core, etc.). So we can revert package.json and package-lock.json just for the playground directory.

packages/semantic-ui/package.json
```diff
@@ -115,6 +113,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "nanoid": "^3.1.23",
```
epicfaace, line 116: We should keep nanoid as a peerDependency for the semantic-ui package (just like what we had originally for nanoid).

epicfaace (changes requested):

packages/playground/package.json
damianfrizzi: I reverted the package-lock changes and removed the shortid dependency

epicfaace (approved):
Great, thanks!

damianfrizzi:
@epicfaace could you give me some guidance why it's failing here? Running `lerna clean` and `lerna bootstrap` on my local machine gives me a lot of changes to the lock files. I think the lockfile for semantic-ui should be updated but it would contain a lot of unrelated changes. Can you advice me what to do here (as I'm not very familiar with lerna)?

wegry:
@damianfrizzi which version npm are you using?

epicfaace (commented):
Looks like the problem might be fixed by adding nanoid as a dev dependency in semantic-ui -- as the tests are failing since nanoid doesn't seem to be found.

damianfrizzi:
@wegry I'm using 6.14.8
@epicfaace I added it to the dev dependencies now as well. Could you approve the workflow again?

wegry:
I was thinking it might be an older version of npm causing the issue, but that sounds new enough to me. https://github.com/npm/cli/compare/v6.14.8...v6.14.13#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

epicfaace (approved):

epicfaace:
Thanks @damianfrizzi and @wegry!

damianfrizzi:
@epicfaace thanks for merging! Would you mind creating a minor release for this? We need it in order to upgrade to Webpack 5.

epicfaace:
We'll have a new release soon, it will probably be a major release though.

--
Ashwin Ramaswami

uwolfer:
> We'll have a new release soon, it will probably be a major release though.

@epicfaace Do you expect to have such a release ready in the next days / weeks? You feedback would help us to decide whether we need to depend on a fork (which includes this fix) or we can wait for the official release.

epicfaace:
Just released.

PR #2436, opened 2021-06-25
Update 3.x upgrade guide.md

epicfaace:
Update upgrade guide to encompass changes in https://github.com/rjsf-team/react-jsonschema-form/pull/2360. @bradhugh can you also take a look to ensure this is accurate?


jimmycallin (approved):

bradhugh:
Yes, this is basically right. The normal theme didn't have an id for the help field at all previously but the semantic-ui theme had a bug where the id for the help was the id of the field itself.

epicfaace:
Thanks for the clarification!

Issue #2440, opened 2021-06-27
alt-date select boxes do not change value


paintedbicycle:
What I believe to be an issue with the v3 of rjsf, alt-date module select box values cannot be changed.

[...]

PR #2441, opened 2021-06-27
Add null check to componentDidUpdate

jimmycallin:

Fixes https://github.com/rjsf-team/react-jsonschema-form/issues/2440#issue-930943828

setState was mistakenly called in componentDidUpdate when no value was passed to the prop.

### Checklist

### DIFF
--- a/packages/core/src/components/widgets/AltDateWidget.js
+++ b/packages/core/src/components/widgets/AltDateWidget.js
@@ -65,6 +65,7 @@ class AltDateWidget extends Component {
 
   componentDidUpdate(prevProps, prevState) {
     if (
+      prevProps.value &&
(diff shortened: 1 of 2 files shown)


epicfaace (commented):
Looks good. Could we add a quick test so we can ensure this doesn't regress again?

Also, maybe we should revert the lock files changes in this PR?

jimmycallin:
> Looks good. Could we add a quick test so we can ensure this doesn't regress again?

Good idea

> Also, maybe we should revert the lock files changes in this PR?

I added the lock file changes due to the CI failing without them. I'll try to remove them, see if it fails, and then create a new PR with only the lock file changes.

jimmycallin:
@epicfaace seems like it works now, not sure what happened. : )

epicfaace (approved):

PR #2456, opened 2021-07-01
Update react peer dependency to be 16.8.0 so hooks can be used in this repo.

ssbyoung:
### Reasons for making this change

I am planning on making a change to fix #2449. Want to be able to use hooks in the feature I'm writing.

### Checklist

* [X] **I'm adding or updating code**
  - [ ] I've added and/or updated tests
  - [X] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed

PR #2457, opened 2021-07-01
fix: ui:emptyValue behavior

### LINKED ISSUE #2449: ui:emptyValue does not work when initial data has an empty string
### Prerequisites

- [X] I have read the [documentation](https://react-jsonschema-form.readthedocs.io/);

### Description

The method of providing a default value, putting a `ui:emptyValue` in the uiSchema does not fill out a default value.

SSCCE 
[rjsf-empty-value-issue.zip](https://github.com/rjsf-team/react-jsonschema-form/files/6735809/rjsf-empty-value-issue.zip)

empty value does not work

### Steps to Reproduce

1. Use ui:emptyValue

#### Expected behavior

The form should have the value of ui:emptyValue if the property is falsey.

#### Actual behavior

Form is still empty

### Version

3.0.0

ssbyoung:
**Please note that this PR relies on https://github.com/rjsf-team/react-jsonschema-form/pull/2456 being merged and released**

Fixes #2449

By default, if a form has a field with an empty string and `ui:emptyValue` is filled out, the empty value will not appear until the user interacts with the form. I believe that this is incorrect. This change makes it so that `ui:emptyValue` will be populated on the first form render.

I also refactored everywhere this logic lived to use a hook inside of @rjsf/core. Because it is a hook it requires the user to be on at least react 16.8.0, hence the need for https://github.com/rjsf-team/react-jsonschema-form/pull/2456 . I also updated all the peer dependencies to 3.0.0. If PR #2456 goes through as a major version. All of these files will need to be updated.

### Important Changes

* `ui:emptyValue` will be populated for empty form fields
* If `ui:emptyValue` is not populated AND the user enters the empty string `onChange` will be called with the empty string instead of undefined.

### Checklist


packages/core/index.d.ts
```diff
@@ -286,6 +288,22 @@ declare module '@rjsf/core' {
 
     export module utils {
 
+        type UseEmptyValueOnChangeProps = {
```
ssbyoung, line 291: I'm definitely not super strong at typescript. If someone has a suggestion how to clean this up, it would be much appreciated.

packages/core/package.json
```diff
@@ -14,7 +14,7 @@
     "precommit": "lint-staged",
     "publish-to-npm": "npm run build && npm publish",
     "start": "concurrently \"npm:build:* -- --watch\"",
-    "tdd": "cross-env NODE_ENV=test mocha --require @babel/register --watch --require ./test/setup-jsdom.js test/**/*_test.js",
+    "tdd": "cross-env BABEL_ENV=test NODE_ENV=test mocha --require @babel/register --watch --require ./test/setup-jsdom.js test/**/*_test.js",
```
ssbyoung, line 17: This command does not work without the `BABEL_ENV` being set.

packages/core/test/StringField_test.js
```diff
@@ -208,6 +208,21 @@ describe("StringField", () => {
       });
     });
 
+    it("should render ui:EmptyValue when formData is emptyString and ui:EmptyValue is defined", () => {
```
ssbyoung, line 211: This is the most important test in this PR. The `act(() => {})` gives the chance for the hook to run.

packages/core/test/StringField_test.js
```diff
@@ -221,7 +236,7 @@ describe("StringField", () => {
       });
 
       sinon.assert.calledWithMatch(onChange.lastCall, {
-        formData: undefined,
+        formData: "",
```
ssbyoung, line 239: These were previously undefined because `options.emptyValue` was undefined. If the user enters the empty string, I believe we should call `onChange` with the empty string.

packages/core/src/utils.js
```diff
@@ -7,6 +7,11 @@ import jsonpointer from "jsonpointer";
 import fields from "./components/fields";
 import widgets from "./components/widgets";
 import validateFormData, { isValid } from "./validate";
+import useEmptyValueOnChange from './hooks/useEmptyValueOnChange';
+
+export const hooks = {
```
ssbyoung, line 12: I just tacked the hooks onto the utils. If there is a better place, let me know.

packages/core/src/hooks/useEmptyValueOnChange.js
```diff
@@ -0,0 +1,30 @@
...
+  const rawOnChange = useCallback((newValue) => {
+    return onChange(mergeValues(newValue));
+  }, [onChange, mergeValues]);
+
+  const eventOnChange = useCallback((event) => {
+    const newValue = event?.target?.value;
+    return rawOnChange(newValue);
+  }, [rawOnChange]);
+
+  useEffect(() => {
```
ssbyoung, line 17: This useEffect takes care of the inital render state.

packages/core/src/hooks/useEmptyValueOnChange.js
```diff
@@ -0,0 +1,30 @@
...
+    if (value === mergeValues(value)){
+      return;
```
ssbyoung, line 19: Added in this guard clause because I was seeing tests fail for `onChange` getting called multiple times if the value wasn't empty.

epicfaace:
> By default, if a form has a field with an empty string and `ui:emptyValue` is filled out, the empty value will not appear until the user interacts with the form. I believe that this is incorrect. This change makes it so that `ui:emptyValue` will be populated on the first form render.

Trying to understand why we need to change this though -- isn't populating data on the first form render what `default` in the schema is for? Why couldn't someone just set something equal to `ui:emptyValue` and `default` if they want this behavior?

I think that when `ui:emptyValue` was originally added to this library, the intention was for it to only override the default behavior of setting a field to `undefined` after the user interacts with the form -- "When a text input is empty, the field in form data is set to undefined" (see https://github.com/rjsf-team/react-jsonschema-form/blob/3ec17f1c0ff40401b7a99c5e9891ac2834a1e73f/docs/usage/validation.md#the-case-of-empty-strings).

Maybe could you further explain your use case so we can see if this PR is the best solution or there is another alternative that may work better?

ssbyoung:
@epicfaace default value does not work properly. Here is a a small example demonstrating it: 
[rjsf-default-issue.zip](https://github.com/rjsf-team/react-jsonschema-form/files/6809881/rjsf-default-issue.zip)

I think default is the behavior I want. It wasn't documented and I only found the syntax in the playground. 

My use case is that when I initially render the form with empty data, I'd like to have certain fields have pre-populated values as defined in the schema.

epicfaace:
@ssbyoung can you share the issue with `default` using a playground example? https://rjsf-team.github.io/react-jsonschema-form/

ssbyoung:
@epicfaace Can you take a look at the example I sent? I think there might be an issue with the consumer of this library using react 17 and this library using react 16.  I'm also unsure how to modify the playground with the issue I'm seeing.

epicfaace:
I think you could make a codepen here (https://codepen.io/epicfaace/pen/WNjvBrb) and put in your desired versions of React / rjsf

ssbyoung:
@epicfaace I'm having a hard time getting a codepen with rjsf@3.0.0. Do you have an example with one that uses the latest version?

ssbyoung:
@epicfaace  is the playground https://rjsf-team.github.io/react-jsonschema-form/ running on version 3.0.0?

epicfaace:
Oh, I think codepens with v3 an issue because of https://github.com/rjsf-team/react-jsonschema-form/issues/2447. Could you try using a codepen with the latest version of v2 instead? I'm guessing the issue will also appear in that version.

ssbyoung:
@ssbyoung https://codepen.io/ssbyoung/pen/NWjdNjv?editors=1111 is a codepen showing this same issue.

epicfaace:
Hmm -- why is the original formData set to {  prop1: undefined }? If you
just set formData to {}, then the default properly gets set.

--
Ashwin Ramaswami

ssbyoung:
As a consumer I'd expect all falsey values to to get replaced. Does it explicitly not have to be set in the object for the default to appear?

ssbyoung:
@epicfaace after trying default with my code, the current implementation works as is. I will be closing this PR as well as the other one related to this change.

PR #2459, opened 2021-07-02
Properly assign label prop for MultiSelect ArrayField

### LINKED ISSUE #2378: Label undefined for Multiselect ArrayField
### Prerequisites

- [X] I have read the [documentation](https://react-jsonschema-form.readthedocs.io/);
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net/n1k0/f2y3fq7L/6/) or a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

JSFiddle: https://jsfiddle.net/f7xjq935/

### Description

While implementing a highly customized rjsf in my current project, I use the `label` prop within the custom widgets to do some i18n / metadata lookup to render the widgets correctly (In essence, all forms are user-generated and the translated labels are provided by a third party). This works surprisingly fine.

However, with an array field with unique items (aka Multiselect), the `label` prop of the custom widget is `undefined` and no other prop can be used as a fallback (as far as I can see) to determine the name of the property in the schema.

I have looked into the code and noticed, that in https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/fields/ArrayField.js#L542, `label` is in fact undefined (it's not part of the ArrayField props). However, there's the `name` prop, which is not being used (and contains exactly what I need).

Is this a bug or expected behavior? Would it be appropriate to set `label = this.props.name`?

JSFiddle: https://jsfiddle.net/f7xjq935/

### Steps to Reproduce

1. Override `SelectWidget` with a custom component that renders the `label` prop
2. Render a form with an enum field (successfully renders the label) and an array field (label = undefined)
3. Be confused

### Version

2.5.1

Parnswir:

Fixes #2378

@epicfaace @fsteger do you think this would be an appropriate solution?

### Checklist


epicfaace:
Why set it to "name" but not "title"?

AdelRashed97 (approved):

Parnswir:
@epicfaace `title` is not yet defined here. In other places, it is set to `const title = schema.title || name;`, so if you think that would be the better solution here, I would add this line as well.

packages/core/src/components/fields/ArrayField.js
```diff
@@ -570,7 +571,7 @@ class ArrayField extends Component {
         disabled={disabled}
         readonly={readonly}
         required={required}
-        label={label}
+        label={title}
```
epicfaace, line 574: Will this give us the same value that `label` had earlier for non-multiselect ArrayFields?

Parnswir: For all other widgets in `ArrayField.js`, `label` is not set and thus undefined. Other fields set it similarly to title ([StringField](https://github.com/Parnswir/react-jsonschema-form/blob/fix/arrayfield_label/packages/core/src/components/fields/StringField.js#L47), [NumberField](https://github.com/Parnswir/react-jsonschema-form/blob/fix/arrayfield_label/packages/core/src/components/fields/BooleanField.js#L61)).

Again, I cannot tell if this is the way it should behave, but for use-cases like i18n (if done client-side), one will need the name or title of the schema property in custom widgets.

epicfaace: > Again, I cannot tell if this is the way it should behave, but for use-cases like i18n (if done client-side), one will need the name or title of the schema property in custom widgets.

Sorry, I don't think I understood what you meant here. Could you explain further?

Parnswir: I was just trying to motivate my use-case again 😄

epicfaace: I'm still a bit confused, do you mind explaining that comment again? "Again, I cannot tell if this is the way it should behave, but for use-cases like i18n (if done client-side), one will need the name or title of the schema property in custom widgets."

Parnswir: ok, sorry for the misunderstanding. All I mean is I don't know how exactly the `label` prop is defined in general.
For my use-case, basically a dynamic translation of select options in the frontend, all I need is the value of the property name (or title), so that I know which translation key to choose. Of course it's not that simple and I can't show code, but that's the point where the missing name/title/label value in custom multiselects is a problem, hence the issue/PR 😄

Parnswir: @epicfaace What can I do to get this merged?

Parnswir: Sorry for being pushy, but we're working around this problem for three months now and this PR would help a lot. Can I do anything to help get this merged?

epicfaace (approved):
Thanks! I took another look and it's helpful to know that other LOCs in the same ArrayField.js have similar logic as `const title = schema.title || name;` -- this makes it easier to understand how it works in context (and it's not adding anything new, rather just adding something that was missing earlier).

PR #2460, opened 2021-07-05
Fix race condition for onChange when formData is controlled prop

olzraiti:
(I had an older PR for this, but I closed that one since I though the issue was fixed. That was a mistake - the test wasn't run because it wasn't inside a `describe` block. The old PR was #1990, sorry for making this a bit messy.)

This bug was introduced by #1804

The race condition occurs when (1) Form receives the formData prop, and will call the onChange function, and (2) when onChange is called at the same time by some component rendered by the Form.

### Checklist


heath-freenome:
Reimplemented on v5 beta due to typescript conversion of the file via https://github.com/rjsf-team/react-jsonschema-form/pull/3052

PR #2465, opened 2021-07-12
Prevent "submit" event propagation

Mottie:

I'm adding profile editing modals to our form's contact info page. The problem is that the modal renders a nested form. And clicking "update" inside the modal causes the "submit" event to propagate to the outer form causing the form to navigate to the next page.

I know nested form aren't allowed within the [HTML spec](https://html.spec.whatwg.org/multipage/forms.html#the-form-element):

> Content model:
> &nbsp;&nbsp;Flow content, but with no form element descendants.

But I could not find a way around it (ref the conversation from the first Slack link)

The only solution I found was to modify this library. I only call the `event.stopPropagation()` if a prop is passed to ensure I don't break something unintentionally.

Related:
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/27137
- https://github.com/department-of-veterans-affairs/vets-website/pull/17831
- https://dsva.slack.com/archives/CBU0KDSB1/p1625846517423300
- https://dsva.slack.com/archives/CBU0KDSB1/p1626107406455800

### Checklist


Mottie:
oops, meant to open this in our fork

PR #2466, opened 2021-07-12
Implementation of If Then Else

nickgros:
I've picked up from where #1666 has left off, hoping that the work I've done so far will be enough to add this feature.

In addition to the changes outlined in #1666, I've tried to address changes requested in that review, and fix a few other issues as well.

### Checklist


packages/core/src/validate.js
```diff
@@ -300,17 +301,27 @@ export function withIdRefPrefix(schemaNode) {
...
+        ajv.addSchema(rootSchema, ROOT_SCHEMA_PREFIX);
+      }
+    }
+    return ajv.validate(withIdRefPrefix(schema), data);
   } catch (e) {
+    console.warn("Encountered error while validating schema", e);
     return false;
-  } finally {
-    // make sure we remove the rootSchema from the global ajv instance
-    ajv.removeSchema(ROOT_SCHEMA_PREFIX);
```
nickgros, line 314: Why is the root schema being removed (before my change)?

Since we now need to run the validator frequently to evaluate conditional subschemas, keeping the root schema in the cache is essential for performance, especially for large schemas.

packages/core/test/ifthenelse_test.js
```diff
@@ -0,0 +1,424 @@
...
+      const { node } = createFormComponent({
+        schema: schemaWithAllOfRef,
+        formData,
+      });
+
+      expect(node.querySelector("input[label=postcode]")).not.eql(null);
+    });
+  });
+
+  it("handles nested if then else", () => {
```
nickgros, line 291: #1666 didn't handle this case properly. A small change in the `resolveSchema` function fixed it (changing `if "if" in schema` to `while "if" in schema`), and here's a test to prove it.

packages/core/test/ifthenelse_test.js
```diff
@@ -0,0 +1,424 @@
...
+    const { node } = createFormComponent({
+      schema: schemaWithNested,
+      formData,
+    });
+    expect(node.querySelector("select[id=root_country]")).not.eql(null);
+    expect(node.querySelector("select[id=root_state]")).not.eql(null);
+    expect(node.querySelector("select[id=root_city]")).not.eql(null);
+  });
+
+  it("handles additionalProperties with if then else", () => {
```
nickgros, line 394: Test we fixed the issue in this comment: https://github.com/rjsf-team/react-jsonschema-form/pull/1666#discussion_r640334787

packages/core/src/validate.js
nickgros: I think I've figured it out: this function uses the same ajv instance as `validateFormData`, which has some advantages and some drawbacks. I think with this new implementation (not removing the root schema), validation breaks for certain schemas. I'm not happy with the performance of the current implementation, though, so I'll have to figure something else out.

nickgros: One cause of performance issues here is that we can't take advantage of AJV's cache because `withIdRefPrefix` creates a new object every time. We see a substantial improvement in performance if we memoize `withIdRefPrefix`, because `validate` sees the same schema and can use the ajv cache instead of recompiling. I'm not yet sure that memoization is 100% safe here, but it seems like a possible path 


Another thing I've noticed: we get different behavior with the current implementation if the root schema has an $id. When an $id is present, I _think_ AJV totally ignores the root schema prefix when we call `addSchema(rootSchema, ROOT_SCHEMA_PREFIX)`, so the `removeSchema` call never works. Not 100% on this either, though.

nickgros: As for `$id` issues, I've filed #2471, ideally would like to address it here (at the risk of increasing scope of this PR) since how RJSF uses AJV is closely related with adding conditional support

nickgros:
Converting this PR to a draft because while I think my changes in [`78fc62c` (#2466)](https://github.com/rjsf-team/react-jsonschema-form/pull/2466/commits/78fc62c971b8e5b19abcf86513ebc868ce39c614) are a step forward, I think it's going to break some things and I don't want to indicate that this should be merged as-is without more tests (in addition to fixing the ones that fail).

packages/core/src/utils.js
```diff
@@ -1,13 +1,16 @@
 import React from "react";
 import * as ReactIs from "react-is";
-import mergeAllOf from "json-schema-merge-allof";
 import fill from "core-js-pure/features/array/fill";
 import union from "lodash/union";
 import jsonpointer from "jsonpointer";
 import fields from "./components/fields";
 import widgets from "./components/widgets";
 import validateFormData, { isValid } from "./validate";
+import _ from "lodash";
```
jimmycallin, line 9: To keep the build size down, it's preferable to import just the necessary function similarly to line 4.

```suggestion
import memoize from "lodash/memoize";
```

packages/core/src/utils.js
```diff
@@ -662,21 +672,28 @@ export function stubExistingAdditionalProperties(
...
+  rootSchema = DEFAULT_ROOT_SCHEMA,
+  formData = DEFAULT_FORM_DATA
+) {
   if (schema.hasOwnProperty("$ref")) {
     return resolveReference(schema, rootSchema, formData);
   } else if (schema.hasOwnProperty("dependencies")) {
     const resolvedSchema = resolveDependencies(schema, rootSchema, formData);
     return retrieveSchema(resolvedSchema, rootSchema, formData);
-  } else if (schema.hasOwnProperty("allOf")) {
+  } else if (schema["allOf"]) {
```
jimmycallin, line 687: what's the reason for this change?

jimmycallin:
I tried it out a bit and I think this looks great! If we can get it in a mergeable state where only the if-then-else changes are included in this PR, I think we are pretty close.

A few comments:
- Currently if you type a value in a conditional field, and then change the parent such that the field is removed, the value is still included in formData and will be submitted, unless `additionalProperties: false` is included. Is this the preferred behaviour? I can see pros and cons of both resetting the value of the formData and including it, but I'm leaning towards resetting the values.
- The memoization changes seem to cause issues when omitExtraData is set to true - I'm getting "maximum call size exceeded". See attached screenshot. 
image

docs/index.md
```diff
@@ -81,7 +81,85 @@ For more information on what themes we support, see [Using Themes](usage/themes)
...
+- When the value is an object in the form data, the defaults are deeply merged into the form data, using the rules defined here for the deep merge.
+- Then the value is an array in the form data, defaults are only injected in existing array items. No new array items will be created, even if the schema has minItems or additional items defined.
+
+### Merging of defaults within the schema
+
+In the schema itself, defaults of parent elements are propagated into children. So when you have a schema which defines a deeply nested object as default, these defaults will be applied to children of the current node. This also merges objects defined at different levels together with the "deeper" not having precedence. If the parent node defines properties, which are not defined in the child, they will be merged so that the default for the child will be the merged defaults of parent and child.
+
+For arrays this is not the case. Defining an array, when a parent also defines an array, will be overwritten. This is only true when arrays are used in the same level, for objects within these arrays, they will be deeply merged again.
+
+## Tips and tricks
```
jimmycallin, line 162: Not sure what's happening here with these docs changes, they are unrelated to the PR and are also out commented. I guess this should be reverted?

packages/core/src/components/fields/SchemaField.js
```diff
@@ -249,7 +249,8 @@ function SchemaFieldRender(props) {
   const FieldTemplate =
     uiSchema["ui:FieldTemplate"] || registry.FieldTemplate || DefaultTemplate;
   let idSchema = props.idSchema;
-  const schema = retrieveSchema(props.schema, rootSchema, formData);
+  var schema = retrieveSchema(props.schema, rootSchema, formData);
```
jimmycallin, line 252: I'm assuming all `const -> var` related changes are due to the original PR being a bit old. In any case, const is highly preferable to var, so please revert all of these changes.

nickgros:
Thanks for taking a look @jimmycallin!

> Currently if you type a value in a conditional field, and then change the parent such that the field is removed, the value is still included in formData and will be submitted, unless `additionalProperties: false` is included. Is this the preferred behaviour? I can see pros and cons of both resetting the value of the formData and including it, but I'm leaning towards resetting the values.

For my use case, I need the option to keep data as an additional property if its field is removed. As our users fill out the form, we want to be as nondestructive as possible with their data, and give them the opportunity to 'backtrack' without having to re-fill fields that they may have already completed.

Shouldn't this concern be handled by `omitExtraData`? It seems like `liveOmit` is working properly, but nothing is being omitted on submit if `liveOmit` is false, so that will need to be fixed (I'm actually seeing this issue in the current playground version 😬). I'm also not actually seeing a difference in behavior when I change `additionalProperties` like you note.

> The memoization changes seem to cause issues when omitExtraData is set to true - I'm getting "maximum call size exceeded". See attached screenshot.

Good catch, and thanks for attaching the stack trace! I'm having trouble reproducing this, but you guided me to find that I can't actually change the data in the "All Of If Then Else" example if `liveOmit` is true. Possible that these are related issues.

packages/core/src/utils.js
nickgros: `allOf` may be set to undefined here: https://github.com/nickgros/react-jsonschema-form/blob/5e7a08dce80c0c1c85293ac3465c63d65693c6d4/packages/core/src/utils.js#L767

I think I changed that per a comment on the previous PR. Could use `delete` instead and revert this.

nickgros: Switching that line to `delete` actually fixed the `liveOmit` issue that I mentioned [here](https://github.com/rjsf-team/react-jsonschema-form/pull/2466#issuecomment-884929527). I'm not sure if it fixed the stack overflow error that you saw.

jeremypele:
Thanks for that PR @nickgros 🎉 

Is there anything I can help you with to be able to complete this feature? Would love to this merged sonner than later

dieseldjango:
I need support for if-then-else and it's great to see there's a PR, but appears stalled. Anything I can help with?

heath-freenome:
@nickgros Does it make sense to rethink these changes on top of the v5 beta's `@rjsf/validator-ajv8`?

nickgros:
This is probably not the right way to optimize validation. Closing

PR #2467, opened 2021-07-12
Add `ui:DuplicateKeySuffixSeparator`

nickgros:

For my own use case, I needed to change the separator from `-` to `_` between a duplicate key and its unique suffix when using `additionalProperties`.

To accomplish this, I added a new field to the uiSchema called `ui:DuplicateKeySuffixSeparator`, which can be used to override the separator with any string.

If the property is undefined, `-` is used, so the default behavior is unchanged.

### Checklist

### DIFF
--- a/docs/api-reference/uiSchema.md
+++ b/docs/api-reference/uiSchema.md
@@ -218,5 +218,18 @@ const uiSchema = {
   "ui:title": "Your password"
 };
 ```
+
+## Duplicate Key Suffix Separator
+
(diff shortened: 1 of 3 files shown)


epicfaace (commented):
Does this need to be in the uiSchema (i.e. being controlled by a per-field basis) or do you think this would also work if you just made it a form prop (so being controlled on an entire-form basis)?

nickgros:
Hey @epicfaace, thanks for the quick response.

This could definitely be a prop on the form component. Should that be in addition to, or instead of exposing it via uiSchema?

nickgros:
Obsoleted by #3048

PR #2470, opened 2021-07-14
Load 3.0 codepen examples in documentation once more

### LINKED ISSUE #2447: NONE of the codepen examples/snippets in the documentation work.

wegry:

The React module's Component property wasn't defined in the 3.0 UMD bundle for some reason in codepen. Use ESM @rjsf/core bundle instead.

fixes #2447 

### Checklist

  Screen Shot 2021-07-14 at 18 29 29

### DIFF
--- a/docs/main.js
+++ b/docs/main.js
@@ -12,7 +12,7 @@ $(".language-jsx").each(function() {
     var prefill = {
         "title":"RJSF example",
         "tags":[],
-        "scripts":["https://cdnjs.cloudflare.com/ajax/libs/react/17.0.2/umd/react.production.min.js","https://cdnjs.cloudflare.com/ajax/libs/react-dom/17.0.2/umd/react-dom.production.min.js","https://unpkg.com/@rjsf/core@3.0.1/dist/react-jsonschema-form.js"],
+        "scripts":["https://cdnjs.cloudflare.com/ajax/libs/react/17.0.2/umd/react.production.min.js","https://cdnjs.cloudflare.com/ajax/libs/react-dom/17.0.2/umd/react-dom.production.min.js","https://unpkg.com/@rjsf/core/dist/react-jsonschema-form.js"],
         "stylesheets":["//maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css"]
     };
     var div = $("<div><div class=\"codepen\" data-height=\"400\" data-theme-id=\"light\" data-default-tab=\"js,result\" data-prefill='" + JSON.stringify(prefill) + " '>\"<pre data-lang=\"html\">\n&lt;div id=\"app\">&lt;/div></pre><pre data-lang=\"babel\">" + text + "</pre></div></div>").css("margin", "30px 0px");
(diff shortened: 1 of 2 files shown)


epicfaace:
Thanks for fixing this. But does this mean that the UMD bundle no longer works? Is there a potential fix for that issue as well?

wegry:
@epicfaace I don't know. I haven't used UMD consciously in years if ESM is available. I'd expect if UMD was broken, it'd get raised as an issue otherwise.

wegry:
Alright, looks like the `external` change I had made to the webpack.config.js was assuming the UMD bundle would be used from a bundler/module aware system. It did _not_ account for React/ReactDOM being an ambient global (like the non-esm CDN approach).

docs/main.js
```diff
@@ -12,7 +12,7 @@ $(".language-jsx").each(function() {
     var prefill = {
         "title":"RJSF example",
         "tags":[],
-        "scripts":["https://cdnjs.cloudflare.com/ajax/libs/react/16.13.1/umd/react.production.min.js","https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.13.1/umd/react-dom.production.min.js","https://unpkg.com/@rjsf/core@2.5.1/dist/react-jsonschema-form.js"],
+        "scripts":["https://cdnjs.cloudflare.com/ajax/libs/react/17.0.2/umd/react.production.min.js","https://cdnjs.cloudflare.com/ajax/libs/react-dom/17.0.2/umd/react-dom.production.min.js","https://unpkg.com/@rjsf/core@3.0.1/dist/react-jsonschema-form.js"],
```
epicfaace, line 15: Let's always get the latest version

```suggestion
        "scripts":["https://cdnjs.cloudflare.com/ajax/libs/react/17.0.2/umd/react.production.min.js","https://cdnjs.cloudflare.com/ajax/libs/react-dom/17.0.2/umd/react-dom.production.min.js","https://unpkg.com/@rjsf/core/dist/react-jsonschema-form.js"],
```

epicfaace (approved):
Great, thanks for finding the fix! :)

packages/core/webpack.config.dist.js
```diff
@@ -14,6 +15,9 @@ module.exports = {
     libraryTarget: "umd"
   },
   plugins: [
+    new MonacoWebpackPlugin({
+      languages: ['json']
+    }),
```
wegry: I don't think this plugin is usable within a library. I'm almost certain that we're just incurring a build time cost with no usable change in output.

epicfaace: Yeah, that's right. Perhaps we could split it so it's only available for the playground and docs in a separate PR though?

PR #2475, opened 2021-07-18
Fix canary builds

epicfaace:
Fix canary builds by not including `from-package` (earlier, it wouldn't publish canaries at all due to the presence of this argument).


heath-freenome:
@epicfaace is this PR still needed? IF not maybe close.

Issue #2439, opened 2021-06-25
Update peerDependencies to include @rjsf/core v3


epicfaace:
Update peerDependencies to include @rjsf/core v3

epicfaace:
fixed by #2479

PR #2479, opened 2021-07-18
#2439 bump core dep in package peer deps

wegry:

Re-add the ability to use @rjsf/core version within lerna.

### Checklist

### DIFF
--- a/packages/antd/package.json
+++ b/packages/antd/package.json
@@ -28,7 +28,7 @@
   },
   "peerDependencies": {
     "@ant-design/icons": "^4.0.0",
-    "@rjsf/core": "^2.0.0",
+    "@rjsf/core": "^3.0.0",
     "antd": "^4.0.0",
     "antd-dayjs-webpack-plugin": "1.0.0",
     "dayjs": "^1.8.0",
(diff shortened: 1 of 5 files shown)


jimmycallin (changes requested):
Thanks! I believe you also should run `npm install` to update the package-lock.json file the new dependency requirements.

epicfaace:
> Thanks! I believe you also should run `npm install` to update the package-lock.json file the new dependency requirements.

Actually, I don't think this is required. This is because peer dependencies won't be installed automatically when using node 14, so the package-lock.json files will be unchanged. In fact, we can see that that is the case because CI runs `npm ci`, which would have errored if the package-lock.json files are out of date.

epicfaace (approved):

jimmycallin (approved):

jimmycallin:
> > Thanks! I believe you also should run `npm install` to update the package-lock.json file the new dependency requirements.
> 
> Actually, I don't think this is required. This is because peer dependencies won't be installed automatically when using node 14, so the package-lock.json files will be unchanged. In fact, we can see that that is the case because CI runs `npm ci`, which would have errored if the package-lock.json files are out of date.

ah, interesting, good to know and apologies for the wrong change request

epicfaace:
No worries!

kerematam:
https://github.com/rjsf-team/react-jsonschema-form/issues/2486#issuecomment-888314289

PR #2497, opened 2021-08-03
fluent-ui: Allow value of 0 in TextWidget

pksorensen:
Fixes proper undefined/null check and not falsy check.  when value is 0 its changed to "" instead of "0"

[Please describe them here]

### Checklist

### DIFF
--- a/packages/fluent-ui/src/TextWidget/TextWidget.tsx
+++ b/packages/fluent-ui/src/TextWidget/TextWidget.tsx
@@ -84,7 +84,7 @@ const TextWidget = ({
       // TODO: once fluent-ui supports the name prop, we can add it back in here.
       // name={name}
       type={inputType as string}
-      value={typeof value === "undefined" || value === null ?  "" : value}
+      value={value || value === 0 ? value : ""}
       onChange={_onChange as any}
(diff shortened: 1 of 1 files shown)


epicfaace (commented):
Thanks, can you add a test?

pksorensen:
I can try, not sure where to start.

packages/fluent-ui/src/TextWidget/TextWidget.tsx
```diff
@@ -84,7 +84,7 @@ const TextWidget = ({
       // TODO: once fluent-ui supports the name prop, we can add it back in here.
       // name={name}
       type={inputType as string}
-      value={value ? value : ""}
+      value={typeof value === "undefined" || value === null ?  "" : value}
```
epicfaace, line 87: Actually, let's just do the same thing as in https://github.com/rjsf-team/react-jsonschema-form/blob/110adbe1b1d46edbae512c22dfd1f8d5183ceba1/packages/material-ui/src/TextWidget/TextWidget.tsx#L59:

```suggestion
      value={value || value === 0 ? value : ""}
```

epicfaace (approved):

PR #2524, opened 2021-08-25
Documenting readonlyAsDisabled functionality.

mdornseif:

There is a `readonlyAsDisabled` which is not mentioned in the Documentation.

### Checklist


epicfaace:
Interesting, I didn't realize `readonlyAsDisabled` existed. Is this a useful feature to have around -- what's the main use case? I wonder if we should just remove it in the next major version.

heath-freenome (approved):
@epicfaace also a fairly straight forward documentation fix

docs/api-reference/form-props.md
```diff
@@ -95,6 +95,7 @@ React component used to customize each field of the form. See [Custom Templates]
 ## formContext
 
 You can provide a `formContext` object to the Form, which is passed down to all fields and widgets. Useful for implementing context aware fields and widgets.
+Using `{readonlyAsDisabled: false}` will make most themes use the toolkit native functionality offer readOnly fields instead of using the HTML-disabled mechanism.
```
epicfaace, line 98: I think this is a more accurate description of what it does, what do you think @mdornseif @heath-freenome  ?
```suggestion
Setting `{readonlyAsDisabled: false}` on the formContext will make the antd theme treat readOnly fields as disabled.
```

epicfaace (approved):

PR #2536, opened 2021-09-02
Fix for clearing errors after updating and submitting form

newt10:
There is a bug in the core form (and other variations) where by the form continues to render errors even after the user has fixed them and submitted the form.

The reason this happens is as follows, when the user submits the form, [onSubmit handler](https://github.com/rjsf-team/react-jsonschema-form/blob/d889f071aa9823bae92d8ea048ce1db0c59e75e9/packages/core/src/components/Form.js#L306) is called. If validation is enabled then, the handler [runs validation functions to check for errors](https://github.com/rjsf-team/react-jsonschema-form/blob/d889f071aa9823bae92d8ea048ce1db0c59e75e9/packages/core/src/components/Form.js#L333), 
```
      let schemaValidation = this.validate(newFormData);
      let errors = schemaValidation.errors;
      let errorSchema = schemaValidation.errorSchema;
      const schemaValidationErrors = errors;
      const schemaValidationErrorSchema = errorSchema;
```
If errors are generated then the form State is [updated with the new errors](https://github.com/rjsf-team/react-jsonschema-form/blob/d889f071aa9823bae92d8ea048ce1db0c59e75e9/packages/core/src/components/Form.js#L348)
```
        this.setState(
          {
            errors,
            errorSchema,
            schemaValidationErrors,
            schemaValidationErrorSchema,
          },
```

If errors are not generated then the [state is updated as follows](https://github.com/rjsf-team/react-jsonschema-form/blob/d889f071aa9823bae92d8ea048ce1db0c59e75e9/packages/core/src/components/Form.js#L377)
```
    this.setState(
      { formData: newFormData, errors: errors, errorSchema: errorSchema },
```
**As we see in the later state update `schemaValidationErrors` and `schemaValidationErrorSchema` are not being updated.**

This creates a problem because the state for the form is derived using the above values as [seen here](https://github.com/rjsf-team/react-jsonschema-form/blob/d889f071aa9823bae92d8ea048ce1db0c59e75e9/packages/core/src/components/Form.js#L73)

```
    const getCurrentErrors = () => {
      if (props.noValidate) {
        return { errors: [], errorSchema: {} };
      } else if (!props.liveValidate) {
        return {
          errors: state.schemaValidationErrors || [],
          errorSchema: state.schemaValidationErrorSchema || {},
        };
      }
      return {
        errors: state.errors || [],
        errorSchema: state.errorSchema || {},
      };
    };
```

What this does is that when the last error is fixed on re-submission of the form, the state is calculated from the stale values of schemaValidationErrors & schemaValidationErrorSchema

We fix this by updating the schemaValidationErrors & schemaValidationErrorSchema in the state when form is submitted with errors fixed.

### Checklist

#### Failing test due to this bug
Test Fail

#### Passing test after fix
Tests Passing

### DIFF
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -364,6 +364,8 @@ export default class Form extends Component {
       }
     }
 
+    // There are no errors generated through schema validation.
+    // Check for user provided errors and update state accordingly.
     let errorSchema;
     let errors;
     if (this.props.extraErrors) {
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -375,7 +375,13 @@ export default class Form extends Component {
     }
 
     this.setState(
-      { formData: newFormData, errors: errors, errorSchema: errorSchema },
+      {
+        formData: newFormData,
+        errors: errors,
+        errorSchema: errorSchema,
+        schemaValidationErrors: [],
+        schemaValidationErrorSchema: {},
+      },
       () => {
         if (this.props.onSubmit) {
           this.props.onSubmit(
--- a/packages/core/test/Form_test.js
+++ b/packages/core/test/Form_test.js
@@ -1531,6 +1531,31 @@ describeRepeated("Form common", createFormComponent => {
           schemaValidationErrorSchema: {},
         });
       });
+
+      it("should reset errors from UI after correction and resubmission", () => {
+        const { node } = createFormComponent({
+          schema,
+        });
+
+        Simulate.change(node.querySelector("input[type=text]"), {
+          target: { value: "short" },
+        });
+        Simulate.submit(node);
+
+        const errorListHTML =
+          '<li class="text-danger">should NOT be shorter than 8 characters</li>';
+        const errors = node.querySelectorAll(".error-detail");
+        // Check for errors attached to the field
+        expect(errors).to.have.lengthOf(1);
+        expect(errors[0]).to.have.property("innerHTML");
+        expect(errors[0].innerHTML).to.be.eql(errorListHTML);
+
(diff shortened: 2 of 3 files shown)


newt10:
This also addresses the PR #2123 with addition of tests.

packages/core/test/Form_test.js
```diff
@@ -1524,6 +1524,12 @@ describeRepeated("Form common", createFormComponent => {
         });
         Simulate.submit(node);
         sinon.assert.notCalled(onError);
+        sinon.assert.calledWithMatch(onSubmit.lastCall, {
+          errors: [],
+          errorSchema: {},
+          schemaValidationErrors: [],
+          schemaValidationErrorSchema: {},
+        });
       });
```
epicfaace, line 1533: Let's add a UI test like https://github.com/rjsf-team/react-jsonschema-form/blob/739a303d5ac59c7ef50b149c56ad777a6044116a/packages/core/test/Form_test.js#L630-L634 that tests the exact scenario for the user that this PR fixes.

packages/core/src/components/Form.js
```diff
@@ -375,7 +375,13 @@ export default class Form extends Component {
     }
```
epicfaace, line 375: Let's add a brief comment at line 367 saying "there are no errors"

epicfaace (approved):

epicfaace (approved):
Address these changes and then feel free to merge

packages/core/src/components/Form.js
newt10: Added comment as requested

packages/core/test/Form_test.js
newt10: Added new UI tests as requested checking for generated HTML

PR #2537, opened 2021-09-03
Fix for ui:options label & description

### LINKED ISSUE #2535: ui:options not behaving as expected from the documentation
### Prerequisites

### Description

Hello ! I'm sorry if I missed something, but in the [documentation](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/) it seems to state that `{"ui:options": {[property]: [value]}}` and `{"ui:[property]": value}` are equivalent. When I try this locally or on your playground though, I'm having a different behaviour where the `ui:options` does not seem to be read.

### Steps to Reproduce

https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0aXRsZSI6IkEgcmVnaXN0cmF0aW9uIGZvcm0iLCJkZXNjcmlwdGlvbiI6IkEgc2ltcGxlIGZvcm0gZXhhbXBsZS4iLCJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJuYW1lIjp7InR5cGUiOiJzdHJpbmcifX19LCJ1aVNjaGVtYSI6eyJ1aTpvcHRpb25zIjp7InRpdGxlIjoiVGl0bGUiLCJkZXNjcmlwdGlvbiI6IkRlc2NyaXB0aW9uIn0sIm5hbWUiOnsidWk6b3B0aW9ucyI6eyJ0aXRsZSI6IkN1c3RvbSB0aXRsZSIsImRlc2NyaXB0aW9uIjoiY3VzdG9tIGRlc2MifX19LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlIjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZX19

#### Expected behavior

`uiSchema["ui:options].title` and  `uiSchema["ui:options].description` should be taken into account and override the field title that is in the json schema

#### Actual behavior

`uiSchema["ui:options"]` seem to be ignored for these ui options.

### Version

3.1.0

newt10:

There are many bugs describing the problem with the current code ignoring some settings from uiSchema; namely
 - ui:title,
 - ui:description
 - ui:options
   - label: <boolean>
   - title: <string>
   - description: <string>

This changeset fixes above problems for all fields, array fields and object fields. Labels are hidden when the ui:options {label: false} is provided as well as title and description fields from ui:options are applied as per the documentation.

If this is related to existing tickets, include links to them as well. Use the syntax fixes #2535 .

### Checklist

### Note

- No documentation updates are required as this is a bug fix and aligns the expected behavior with documented behavior.

### Screenshots

uiOptions

uiTitle

#### Tests passing after the code change
uiSchema Tests Passing

#### Tests failing before code change
https://pastebin.com/QE2a9jqv

### DIFF
--- a/packages/core/src/components/fields/ObjectField.js
+++ b/packages/core/src/components/fields/ObjectField.js
@@ -8,13 +8,17 @@ import {
   getDefaultRegistry,
   canExpand,
   ADDITIONAL_PROPERTY_FLAG,
+  getUiOptions,
 } from "../../utils";
 
 function DefaultObjectFieldTemplate(props) {
   const { TitleField, DescriptionField } = props;
+  const uiOptions = getUiOptions(props.uiSchema);
+  let { label: displayLabel = true } = uiOptions;
+
   return (
     <fieldset id={props.idSchema.$id}>
-      {(props.uiSchema["ui:title"] || props.title) && (
+      {displayLabel && (props.uiSchema["ui:title"] || props.title) && (
         <TitleField
           id={`${props.idSchema.$id}__title`}
           title={props.title || props.uiSchema["ui:title"]}
@@ -22,7 +26,7 @@ function DefaultObjectFieldTemplate(props) {
           formContext={props.formContext}
         />
       )}
-      {props.description && (
+      {displayLabel && props.description && (
         <DescriptionField
           id={`${props.idSchema.$id}__description`}
           description={props.description}
--- a/packages/core/src/components/fields/ObjectField.js
+++ b/packages/core/src/components/fields/ObjectField.js
@@ -18,10 +18,10 @@ function DefaultObjectFieldTemplate(props) {
 
   return (
     <fieldset id={props.idSchema.$id}>
-      {displayLabel && (props.uiSchema["ui:title"] || props.title) && (
+      {displayLabel && props.title && (
         <TitleField
           id={`${props.idSchema.$id}__title`}
-          title={props.title || props.uiSchema["ui:title"]}
+          title={props.title}
           required={props.required}
           formContext={props.formContext}
         />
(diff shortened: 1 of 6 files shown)


packages/core/src/components/fields/ObjectField.js
```diff
@@ -8,24 +8,31 @@ import {
...
   ADDITIONAL_PROPERTY_FLAG,
+  getUiOptions,
 } from "../../utils";
 
 function DefaultObjectFieldTemplate(props) {
   const { TitleField, DescriptionField } = props;
+  const uiOptions = getUiOptions(props.uiSchema);
+  let { label: displayLabel = true } = uiOptions;
+
+  const title = uiOptions.title || props.title;
```
epicfaace, line 19: I noticed that we're changing the order of precedence here (uioptions.title first and then props.title). This is OK (and good so that we're consistent), but I think we should probably label this a breaking change? What do you think?

if so, can you make a "4.x upgrade guide.md" in the docs and describe this change to users?

newt10:
@epicfaace I have made the changes based on your review. Let me know if you need anything else in order to merge this.

packages/core/src/components/fields/ObjectField.js
bejito: @epicfaace I'm not sure this is the case.
If  you look [below](https://github.com/rjsf-team/react-jsonschema-form/blob/93c278b11111eaa031eb5ccd3e8045c84ce7e344/packages/core/src/components/fields/ObjectField.js#L231) on this file, you can see that `props.title` is actually `uiSchema["ui:title"] || title`
This seems to be a little convoluted I agree, we are looking at the uiSchema both in the field definition and in the template. But in any case, the behaviour should be consistent with the current version

epicfaace: Oh, then that's great! What do you think @newt10 ? I think we can remove the 4.x upgrade guide changes then?

newt10: @bejito thats a good catch, this fix is not inverting the order. @epicfaace  I confirmed it on the playground with before and after the fix

#Before - ui:title is taking priority
Screen Shot 2021-09-06 at 11 55 04 AM

#After - ui:options takes priority if provided
Screen Shot 2021-09-06 at 11 58 17 AM

bejito:
LGTM

MilesNorton:
@newt10 @bejito What is the status with this ticket?

heath-freenome:
@newt10 With the conversion to typescript in v5, I was wondering if you had time to refactor this onto the latest changes in master?

PR #2539, opened 2021-09-06
Update SelectWidget.tsx

pksorensen:
Dropdown has its own Label Support and if providing onRenderLabel, the label is provided twice given the old implementation. This fixes such it works as today but allow to use onRenderLabel


heath-freenome:
Fixed in #3411

PR #2543, opened 2021-09-10
prevent reference from losing local properties on nested references, …

luisrafaelmartins13:
prevent reference from losing local properties on nested references, using the same behaviour as `resolveReference()`

using the following schema: 

```
{
  definitions: {
    basePhone: {
      type: "string",
      minLength: 10,
    },
    cellphone: {
      $ref: "#/definitions/basePhone",
      title: "Cellphone number",
    },
    alternativeCellphone: {
      $ref: "#/definitions/basePhone",
      title: "Alternative Cellphone number",
      description: "Not applied to all",
    }
  },
  maximum: 12,
  $ref: "#/definitions/cellphone"
}

```
I'm pointing to the definition #/definitions/cellphone that points to #/definitions/basePhone.

the expected result was to have the properties from each reference resulting in:

```
{
  maximum: 12,
  minLength: 10,
  type: "string",
  title: "Cellphone number"
}
```

the result retrieving the schema was:
```
{
  maximum: 12,
  minLength: 10,
  type: "string",
}

```
ignoring the properties from the 1st reference.
The function `resolveReference()` is prepared get the local properties by merging the result, but on `findSchemaDefinition()` it isn't, resulting in the schema above.

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -597,7 +597,8 @@ export function findSchemaDefinition($ref, rootSchema = {}) {
     throw new Error(`Could not find a definition for ${origRef}.`);
   }
   if (current.hasOwnProperty("$ref")) {
-    return findSchemaDefinition(current.$ref, rootSchema);
+    const { $ref, ...currentSchema } = current;
+    return { ...findSchemaDefinition($ref, rootSchema), ...currentSchema };
   }
   return current;
 }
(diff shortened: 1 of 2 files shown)


heath-freenome:
@luisrafaelmartins13 This is a good fix, unfortunately the v5 beta refactored the `utils.js` into its own `@rjsf/utils` package and thus these changes will need to be refactored there.

heath-freenome:
Closing as this was fixed in https://github.com/rjsf-team/react-jsonschema-form/blob/eb577da7da1fabaa43c8954256c6c51cea6eea69/packages/utils/src/findSchemaDefinition.ts#L52

PR #1442, opened 2019-08-29
Add propertyNames enum and pattern validation (fixes #1250)

v1vendi:
### Reasons for making this change

JSON Schema Draft 6 allows to validate propertyNames, I added this feature
Also fixes #1250 

If this is related to existing tickets, include links to them as well.

### Checklist

* [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests
  - [x] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
* [x] **I'm adding a new feature**
  - [x] I've updated the playground with an example use of the feature

PR #2544, opened 2021-09-12
Add propertyNames enum and pattern validation

watilde:

Follow up PR to https://github.com/rjsf-team/react-jsonschema-form/pull/1442. Rebase onto master and address the comments.
Fixes https://github.com/rjsf-team/react-jsonschema-form/issues/1250.

### Checklist

### DIFF
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -98,8 +98,11 @@ class LabelInput extends React.Component {
         schema={schema}
         formData={this.state.value}
         registry={registry}
-        onChange={value => this.setState({ value })}
-        onBlur={(id, value) => onChange(value)}
+        onChange={value => {
+          onChange(value);
+          this.setState({ value });
+        }}
+        onBlur={(_, value) => onChange(value)}
       />
     );
   }
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -71,17 +71,38 @@ function Label(props) {
   );
 }
 
-function LabelInput(props) {
-  const { id, label, onChange } = props;
-  return (
-    <input
-      className="form-control"
-      type="text"
-      id={id}
-      onBlur={event => onChange(event.target.value)}
-      defaultValue={label}
-    />
-  );
+class LabelInput extends React.Component {
+  constructor(props) {
+    super();
+    this.state = {
+      value: props.label,
+    };
+  }
+  render() {
+    const { id, label, onChange, schema, registry } = this.props;
+
+    if (!schema) {
+      return (
+        <input
+          className="form-control"
+          type="text"
+          id={id}
+          onBlur={event => onChange(event.target.value)}
+          defaultValue={label}
+        />
+      );
+    }
+
+    return (
+      <SchemaField
+        schema={schema}
+        formData={this.state.value}
+        registry={registry}
+        onChange={value => this.setState({ value })}
+        onBlur={(id, value) => onChange(value)}
+      />
+    );
+  }
 }
 
 function Help(props) {
@@ -190,6 +211,7 @@ function WrapIfAdditional(props) {
     readonly,
     required,
     schema,
+    propertyNamesSchema,
   } = props;
   const keyLabel = `${label} Key`; // i18n ?
   const additional = schema.hasOwnProperty(ADDITIONAL_PROPERTY_FLAG);
@@ -206,7 +228,8 @@ function WrapIfAdditional(props) {
             <Label label={keyLabel} required={required} id={`${id}-key`} />
             <LabelInput
               label={label}
-              required={required}
+              schema={propertyNamesSchema}
+              registry={props.registry}
               id={`${id}-key`}
               onChange={onKeyChange}
             />
@@ -244,6 +267,7 @@ function SchemaFieldRender(props) {
     required,
     registry = getDefaultRegistry(),
     wasPropertyKeyModified = false,
+    propertyNamesSchema,
   } = props;
   const { rootSchema, fields, formContext } = registry;
   const FieldTemplate =
@@ -344,6 +368,7 @@ function SchemaFieldRender(props) {
     fields,
     schema,
     uiSchema,
+    propertyNamesSchema,
     registry,
   };
 
@@ -428,6 +453,7 @@ if (process.env.NODE_ENV !== "production") {
     schema: PropTypes.object.isRequired,
     uiSchema: PropTypes.object,
     idSchema: PropTypes.object,
+    propertyNamesSchema: PropTypes.object,
     formData: PropTypes.any,
     errorSchema: PropTypes.object,
     registry: types.registry.isRequired,
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -71,41 +71,28 @@ function Label(props) {
   );
 }
 
-class LabelInput extends React.Component {
-  constructor(props) {
-    super();
-    this.state = {
-      value: props.label,
-    };
-  }
-  render() {
-    const { id, label, onChange, schema, registry } = this.props;
-
-    if (!schema) {
-      return (
-        <input
-          className="form-control"
-          type="text"
-          id={id}
-          onBlur={event => onChange(event.target.value)}
-          defaultValue={label}
-        />
-      );
-    }
-
+function LabelInput(props) {
+  const { id, label, onChange, schema, registry } = props;
+  if (!schema) {
     return (
-      <SchemaField
-        schema={schema}
-        formData={this.state.value}
-        registry={registry}
-        onChange={value => {
-          onChange(value);
-          this.setState({ value });
-        }}
-        onBlur={(_, value) => onChange(value)}
+      <input
+        className="form-control"
+        type="text"
+        id={id}
+        onBlur={event => onChange(event.target.value)}
+        defaultValue={label}
       />
     );
   }
+  return (
+    <SchemaField
+      schema={schema}
+      formData={label}
+      registry={registry}
+      onChange={onChange}
+      onBlur={(_, value) => onChange(value)}
+    />
+  );
 }
 
 function Help(props) {
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -74,11 +74,12 @@ function Label(props) {
 function LabelInput(props) {
   return (
     <SchemaField
-      {...props}
       idSchema={props.idSchema || { $id: props.id }}
       schema={props.schema || { type: "string" }}
       defaultValue={props.defaultValue || props.label}
       formData={props.formData || props.label}
+      registry={props.registry}
+      onChange={props.onChange}
       onBlur={(event, value) => props.onChange(value || event.target.value)}
     />
   );
--- a/packages/core/src/components/fields/SchemaField.js
+++ b/packages/core/src/components/fields/SchemaField.js
@@ -231,6 +231,7 @@ function WrapIfAdditional(props) {
             <Label label={keyLabel} required={required} id={`${id}-key`} />
             <LabelInput
               label={label}
+              required={required}
               schema={propertyNamesSchema}
               registry={props.registry}
               id={`${id}-key`}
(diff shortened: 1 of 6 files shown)


packages/core/src/components/fields/SchemaField.js
```diff
@@ -71,17 +71,41 @@ function Label(props) {
...
+    this.state = {
+      value: props.label,
+    };
+  }
+  render() {
+    const { id, label, onChange, schema, registry } = this.props;
+
+    if (!schema) {
+      return (
+        <input
```
watilde, line 86: I did not update this `<input />` as it seems to be a cause of breaking change if we replace that with `{ type: "string" }`.

watilde:
@epicfaace Kindly asking for your review on this PR :)

watilde:
@newt10 @heath-freenome @epicfaace Kindly asking for your review on this PR :)

packages/core/src/components/fields/SchemaField.js
```diff
@@ -71,17 +71,41 @@ function Label(props) {
...
-      onBlur={event => onChange(event.target.value)}
-      defaultValue={label}
-    />
-  );
+class LabelInput extends React.Component {
+  constructor(props) {
+    super();
+    this.state = {
+      value: props.label,
+    };
```
heath-freenome, line 79: I'm curious why you switched to a full React compnent when you could have just used the `useState()` hook for this? There is a lot more overhead for full components vs a functional component

packages/core/src/components/fields/SchemaField.js
```diff
@@ -206,7 +231,8 @@ function WrapIfAdditional(props) {
             <Label label={keyLabel} required={required} id={`${id}-key`} />
             <LabelInput
               label={label}
-              required={required}
```
heath-freenome, line 209: Why remove `required`?

packages/core/src/components/fields/SchemaField.js
watilde: @heath-freenome Better consistency as all components in the same directory use class when using `this.state` and there is no `useState` hook. I agree with you and we might want to refactor :) For this PR, it'd be better to keep the scope small.

packages/core/src/components/fields/SchemaField.js
watilde: That seems to be needed. I just took `required` back, thank you!

packages/core/src/components/fields/SchemaField.js
heath-freenome: I'm not sure what you mean by "there is not `useState` hook"... See the documentation: https://reactjs.org/docs/hooks-state.html

watilde: @heath-freenome I forgot to say `in the same directory`. I meant the below:
> $ grep -r useState . react-jsonschema-form/packages/core/src/components
> # Not found

heath-freenome (approved):
LGTM, but I would recommend that @epicfaace do a sanity check since I've only contributed two small PRs

packages/core/src/components/fields/SchemaField.js
epicfaace: Sounds good. I'm curious, what would the breaking change have looked like? If we have replaced it with a SchemaField with type: string, I would have expected it to look the same, though perhaps there's something I'm missing.

packages/core/src/components/fields/SchemaField.js
```diff
@@ -71,17 +71,41 @@ function Label(props) {
...
+        />
+      );
+    }
+
+    return (
+      <SchemaField
+        schema={schema}
+        formData={this.state.value}
+        registry={registry}
+        onChange={value => {
```
epicfaace, line 101: Since we're always calling `onChange(value);` in the `onChange` prop, is there even a need to store `this.state.value` now? Could we just use something like `this.props.value` to get the current value of the field?

watilde: `this.props.value` did not work as the props does not seem to have value, but I could use label as `formData`. Let me know if the update addresses your comment.

packages/core/src/components/fields/SchemaField.js
watilde: What I tried is the below SchemaField and faced 33 fails([detail](https://gist.github.com/watilde/9d731026a6f3edd5ba3df6b71c0b3fc0)).

```jsx
 <SchemaField
  schema={{type: "string"}}
  value={label}
  registry={registry}
  onChange={onChange}
  onBlur={(_, value) => onChange(value)}
/>;
```

watilde: @epicfaace Could we replace the input with SchemaField separately from this PR? This PR uses the exactly same `<input>` tag with the current implementation.

epicfaace: That seems OK. However, I'm just worried now that the `SchemaField` we _do_ implement in this PR (in the screenshot below) might also have similar bugs that we're not aware of:

image

Can we try to get to the bottom of this? At least if we can understand why the <SchemaField> for the type: string was failing (and demonstrate that it's a different situation from the SchemaField we added, which wouldn't fail), then I'd be more comfortable merging this PR>

watilde: @epicfaace Sure! I'll dig into that a bit more, but it'd be great if we do not need to refactor `SchemaField` when we find any. My ask was to keep the current behaviour for no schema case so that we can minimize breaking change risk as less as we can. As the screenshot you put shows no schema case is using the exactly same `<input />` tag with the current master, I'm wondering if we could discuss in a separated issue as the original issue is not related to that.

I'll investigate anyway.

watilde: Turned out SchemaField puts wrong id. In the below example, it should be `id="root_first-key", but it uses `id="root"`. 
https://gist.github.com/watilde/bbd85528c7ba684e366ea451145c5f72#file-after-html-L10

watilde: @epicfaace Ok, I fixed the issue by passing id as `idSchema`. We use only SchemaField now :)
https://github.com/rjsf-team/react-jsonschema-form/pull/2544/commits/72dcbf75d313e84bb776ec05238f30f7a8c8b771

packages/core/src/components/fields/SchemaField.js
```diff
@@ -72,14 +72,14 @@ function Label(props) {
...
-  const { id, label, onChange } = props;
   return (
-    <input
-      className="form-control"
-      type="text"
-      id={id}
-      onBlur={event => onChange(event.target.value)}
-      defaultValue={label}
+    <SchemaField
+      {...props}
```
epicfaace, line 77: Nice, great you figured it out!

Since props can have other attributes like the id attribute, which we don't need to pass to SchemaField, could we instead explicitly name / pass down all props instead of doing {...props}?

watilde: Thank you for your advice! I just updated the lines to explicitly use attributes from props :)

packages/core/test/ObjectField_test.js
```diff
@@ -414,7 +414,7 @@ describe("ObjectField", () => {
         formData: { first: 1 },
       });
 
-      expect(node.querySelectorAll(".field-string")).to.have.length.of(1);
+      expect(node.querySelectorAll(".field-string")).to.have.length.of(2);
```
epicfaace, line 417: Why does this change from 1 to 2?

packages/core/test/SchemaField_test.js
```diff
@@ -292,6 +292,36 @@ describe("SchemaField", () => {
...
+        1
+      );
+    });
+
+    it("should render SchemaField as input for object key if schema has propertyNames schema", () => {
+      const schemaWithPropertyNames = {
+        ...schema,
+        propertyNames: {
+          type: "string",
+          enum: ["1", "2"],
```
epicfaace, line 315: Can you add a test to see if you just set propertyNames equal to something like `"pattern": "^[A-Za-z_][A-Za-z0-9_]*$"` (without specifying a `type` explicitly), it still works?

packages/core/test/ObjectField_test.js
watilde: This change adds a field-string wrapper for a single input. I'll investigate how to remove the wrapper shortly

packages/core/test/SchemaField_test.js
watilde: Sounds good, I'll add them!

packages/core/test/ObjectField_test.js
watilde: Turned out the wrapper is added here. We also cannot use `{ type: "string", [ADDITIONAL_PROPERTY_FLAG]: {} }}` as the flag requires many additional attributes which are not in arguments.
https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/fields/SchemaField.js#L197-L199

watilde:
@epicfaace Kindly asking for your review again to push the thing forward together :)

packages/core/test/ObjectField_test.js
epicfaace: Sorry, I'm still a bit confused as to why we are seeing this behavior only with this PR. This schema doesn't have anything to do with additionalProperties, right?

Can you explain a bit further -- or if you want to quickly grab some time to talk over Zoom, you can also do that (https://calendly.com/aramaswamis/meeting?month=2021-09)?

heyzeusalmighty:
Hello!  Are there any plans on merging this PR in for release?  Happy to help in anyway if assistance is needed.

epicfaace:
Thanks -- do you mind adding it to the weekly meeting agenda? I think a synchronous review would be easiest -- https://github.com/rjsf-team/react-jsonschema-form/issues/2677

heyzeusalmighty:
Added.

packages/core/src/components/fields/SchemaField.js
```diff
@@ -244,6 +248,7 @@ function SchemaFieldRender(props) {
     required,
     registry = getDefaultRegistry(),
     wasPropertyKeyModified = false,
+    propertyNamesSchema,
```
epicfaace, line 251:
```suggestion
    propertyNamesSchema = { type: "string" },
```

packages/core/src/components/fields/SchemaField.js
```diff
@@ -207,6 +209,8 @@ function WrapIfAdditional(props) {
             <LabelInput
```
epicfaace, line 209: Remove LabelInput, and replace with

```jsx
<SchemaField
      schema={propertyNamesSchema}
      idSchema={{ $id: "${id}-key" }}
      defaultValue={label}
      registry={props.registry}
      onBlur={(event, value) => onKeyChange(value)}
/>
```

Add a comment saying something like "this lets us render an arbitrary field based on property names (such as dropdown for `enum`)"

packages/core/test/SchemaField_test.js
```diff
@@ -292,6 +292,50 @@ describe("SchemaField", () => {
...
+        propertyNames: {
+          enum: ["1", "2"],
+          pattern: "^[A-Za-z_][A-Za-z0-9_]*$",
+        },
+      };
+      const { node } = createFormComponent({ schema: schemaWithPropertyNames });
+      Simulate.click(node.querySelector(".glyphicon-plus"));
+
+      expect(node.querySelectorAll("select")).to.have.length.of(1);
+    });
```
epicfaace, line 336: Let's add more tests for robustness:

```
{propertyNames: {oneOf: [{const: "a"}, const: "b"}]}}

{propertyNames: {$ref: "#/definitions/test"}}

{propertyNames: {type: "boolean"}}

{propertyNames: {const: "test"}}
```

and make sure the appropriate widget renders properly.

epicfaace:
@heyzeusalmighty Let's branch off this PR and finish it with these changes. Thanks for your help!

epicfaace:
```
git pull https://github.com/watilde/react-jsonschema-form.git propertyNames
```

heath-freenome:
@watilde @heyzeusalmighty It seems that the delay in adding tests along with the conversion of core to Typescript in the v5 beta has made this PR need more work, sorry

heath-freenome:
Due to the v5 version converting these files to typescript, this PR needs to be refactored along with the additional tests requested

watilde:
Let me close this PR and wait for someone to try

PR #2556, opened 2021-09-18
add event bubbling to submit event (fix for React 17 and up)

elad-ivanir:

Programmatic submittng does not work for React 17+, due to changes in event delegation in React 17.

linked issue - #2551 

### Checklist


epicfaace:
Thanks for the PR. I'm going to close this for now -- see https://github.com/rjsf-team/react-jsonschema-form/issues/2104#issuecomment-854158482 -- we need someone to look into what is the root cause of the issue before we can accept a fix. Not sure if `bubbles: true` is the right fix that gets to the root of the problem.

PR #2589, opened 2021-10-24
fix(bootstrap-4): double margin above select widget help text

abemedia:

The `SelectWidget` has a double margin above the help text. Have added `mb-0` like on the `TextWidget`.


heath-freenome:
@abemedia This PR has conflicts, is missing updates to the test snapshot and, I'm also interested in seeing a screenshot of the before and after

heath-freenome:
@abemedia This most likely was fixed when the `Form.Group` was moved to the `FieldTemplate` in v5. If this is still an issue, please provide a new fix.

PR #2595, opened 2021-11-04
Don't crash when non-object formData is passed in to a schema item with additionalProperties

olensmar:

Fixed a crash when null formData was passed for property with additionalProperties defined in the schema

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -634,6 +634,9 @@ export function stubExistingAdditionalProperties(
     properties: { ...schema.properties },
   };
 
+  // handle null formData
+  formData = formData || {};
+
   Object.keys(formData).forEach(key => {
     if (schema.properties.hasOwnProperty(key)) {
       // No need to stub, our schema already has the property
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -634,8 +634,8 @@ export function stubExistingAdditionalProperties(
     properties: { ...schema.properties },
   };
 
-  // handle null formData
-  formData = formData || {};
+  // make sure formData is an object
+  formData = isObject(formData) ? formData : {};
 
   Object.keys(formData).forEach(key => {
     if (schema.properties.hasOwnProperty(key)) {
--- a/packages/core/test/Form_test.js
+++ b/packages/core/test/Form_test.js
@@ -46,7 +46,7 @@ describeRepeated("Form common", createFormComponent => {
         <Form {...props}>
           <button type="submit">Submit</button>
           <button type="submit">Another submit</button>
-        </Form>,
+        </Form>
       );
       const node = findDOMNode(comp);
       expect(node.querySelectorAll("button[type=submit]")).to.have.length.of(2);
@@ -73,11 +73,11 @@ describeRepeated("Form common", createFormComponent => {
       const comp = renderIntoDocument(
         <Form {...props}>
           <button type="submit">Submit</button>
-        </Form>,
+        </Form>
       );
       const node = findDOMNode(comp);
       expect(node.querySelector(".unsupported-field").textContent).to.contain(
-        "Unknown field type undefined",
+        "Unknown field type undefined"
       );
     });
   });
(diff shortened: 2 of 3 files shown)


packages/core/src/utils.js
```diff
@@ -634,6 +634,9 @@ export function stubExistingAdditionalProperties(
     properties: { ...schema.properties },
   };
 
+  // handle null formData
+  formData = formData || {};
```
epicfaace, line 638: Shouldn't `formData = {}` already fix this on line 629? Why doesn't that work?

epicfaace: I'm trying to understand when `formData` would ever be `null`? Won't it always be an object or `undefined`?

olensmar: Hi @epicfaace - thanks for looking into this! 

I bumped into this where I had a nested object property that is explicitly set to null (not undefined) which caused the error you will get if you comment my fix and run the provided test..

The actual code in the npm package for `stubExistingAdditionalProperties` contains

```javascript
var formData = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
```

which you can see doesn't handle null - only undefined. I think better would be 

```javascript
var formData = arguments.length > 2 && arguments[2] ? arguments[2] : {};
```

since null evaluates to false - but I don't know where this "transpilation" is coming from!? 

Happy to provide more context/info if needed!

epicfaace: The transpilation happens because the original code says

```
export function stubExistingAdditionalProperties(
  schema,
  rootSchema = {},
  formData = {}
) 
```

Specifying a default argument of `{}` means that the default will be reverted to only if `formData` is undefined (but not if it is null).



> I bumped into this where I had a nested object property that is explicitly set to `null` (not `undefined`) which caused the error you will get if you comment my fix and run the provided test..

Interesting! Can you add a test that explicitly checks for this condition from the end user's point of view? i.e. a test in Form_test.js that sets the formData according to what you mentioned

olensmar: Hi @epicfaace - sorry for the delay - i've added the corresponding Form test - if you uncomment my fix you should get something like the following error when running `npm run test` 
```
@rjsf/core:   5) utils
@rjsf/core:        retrieveSchema()
@rjsf/core:          should handle null formData for schema which contains additionalProperties:
@rjsf/core:      TypeError: Cannot convert undefined or null to object
@rjsf/core:       at Function.keys (<anonymous>)
@rjsf/core:       at stubExistingAdditionalProperties (src/utils.js:674:10)
@rjsf/core:       at retrieveSchema (src/utils.js:758:12)
```

Hope this helps!

epicfaace: Thanks for the test, I get what you're doing now. I think we want to handle not only null formData, but any non-object formData.

How about we do this instead:

```suggestion
formData = isObject(formData) ? formData : {};
```

And then let's also add an additional test where the formData is set to a non-null non-object (for example, when formData is equal to `2` or an array)?

packages/core/test/Form_test.js
```diff
@@ -506,6 +506,29 @@ describeRepeated("Form common", createFormComponent => {
       expect(node.querySelectorAll("input[type=text]")).to.have.length.of(1);
     });
 
+    it("should handle null values for property with additionalProperties", () => {
```
epicfaace, line 509:
```suggestion
    it("should not crash with null values for property with additionalProperties", () => {
```

olensmar: done!

packages/core/test/utils_test.js
```diff
@@ -1770,6 +1770,21 @@ describe("utils", () => {
       });
     });
 
+    it("should handle null formData for schema which contains additionalProperties", () => {
```
olensmar, line 1773: @epicfaace do you still want to keep this test?

olensmar:
thanks @epicfaace - i've made some more updates in line with your suggestions

packages/core/test/utils_test.js
epicfaace: Sure

epicfaace (approved):
Thanks! This is great.

olensmar:
my pleasure - thanks for a great project!

epicfaace:
Can you fix formatting (see CI)?

olensmar:
@epicfaace done! (I think...)

olensmar:
thanks again @epicfaace - when do you expect to do a release including this fix?

epicfaace:
doing one now

PR #2597, opened 2021-11-08
MUI rendered is missing any control over the fields render. Other ren…

mikezerosix:
Other renders have 'className' variable to control the style of the forms fields. Toward this end I added 'fieldFlexWidth' uiSchema varaible to control the flex grid withs in MUI ObjectFieldTemplate.tsx instead of the hard coded '12'. If variable is not the the code defaults to '12' as before and code behaves exactly as before.

Other renders have 'className' variable to control the style of the forms fields. MUI rendered has no way to control the style. And in order to be able to render form in multiple columns #2108 and have some responsiveness to UI. Change is minimal and non volatile. When not used it has no effect on the behavior and code defaults to previously hard coded value. 

### Checklist

### DIFF
--- a/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
+++ b/packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
@@ -57,7 +57,7 @@ const ObjectFieldTemplate = ({
           ) : (
             <Grid
               item={true}
-              xs={12}
+              xs={uiSchema['ui:fieldFlexWidth'] || 12}
               key={index}
               style={{ marginBottom: "10px" }}>
               {element.content}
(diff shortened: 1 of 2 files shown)


mikezerosix:
Linked to issue
:https://github.com/rjsf-team/react-jsonschema-form/issues/2108

packages/material-ui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx
```diff
@@ -57,7 +57,7 @@ const ObjectFieldTemplate = ({
           ) : (
             <Grid
               item={true}
-              xs={12}
+              xs={uiSchema['ui:fieldFlexWidth'] || 12}
```
heath-freenome, line 60: As there is more than this one grid in the theme, so I'm not sure your fix is complete. Also, it is missing an update to tests to verify the change. Moreover, in v5 beta, an additional theme for material-ui 5 was added so you might want replicate this there. And finally, other themes also use the flex grid concept, so maybe making the appropriate changes there is a good thing?

Issue #531, opened 2017-03-30
Make custom widgets aware of their location in formData


lebek:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

It'd be useful if custom widgets were passed a prop describing their location in the formData (probably as an array, representing a key path). For example:
```javascript
const schema = {
  type: "object",
  properties: {
    foo: {
      type: "object",
      properties: {
        bar: {type: "string"},
        baz: {type: "string"}
      }
    }
  }
};

const uiSchema = {
  foo: {
    bar: {
      "ui:widget": (props) => {
        console.log(props.schemaPath) // outputs: ["foo", "bar"]
      }
    }
  }
}
```

Use case:
I have a custom widget that lets the use select from a list of entities _or_ create a new entity. If the user chooses to create a new entity they're sent to a new page with a different form. When they submit this second form, they're directed back to the first form where the newly created entity should be populated in the original field.

At the end of the flow we need to know where to put the new entity in the formData, which is difficult because the original custom widget never knew where it was.

Let me know if there's a way to implement this I haven't thought of (apart from passing a key path in `ui:options`). I'm happy to work on a PR for this if it makes sense.

eeinowski:
idSchema can be useful for this in some scenarios.  For instance, if you want to know which array item you are working from, you can extract it from the idSchema using regex. Example: root_specimens_0 tells me that I'm working in the array Specimens at the index of 0.

epicfaace:
@lebek Are you still having this problem? Did the idSchema solution work for you?

suriya-p-rivet:
Hi @epicfaace ,
I am having a custom field called buttonField, in the buttonfield I can able to get the idSchema from props.
however, it is separated by `"_"`.
But in my object many property name also having `"_"` like `"external_link"`.
In this case how can I get the path for this field. Any suggestion, ?

epicfaace:
> Hi @epicfaace ,
> I am having a custom field called buttonField, in the buttonfield I can able to get the idSchema from props.
> however, it is separated by `"_"`.
> But in my object many property name also having `"_"` like `"external_link"`.
> In this case how can I get the path for this field. Any suggestion, ?

[...]

PR #2628, opened 2021-12-08
Can now specify idSeparator to control how id is created

green7ea:
By default, '_' is used as a separator for the different parts that
make up a field's id. This can be problematic when '_' is part of a
field's name.

This commit adds the ability to specify idSeparator when building a
form which allows users use a character that isn't used in field
names.

With the idSeparator option set to '.', we would get:

'root.field_a.field_123' instead of 'root_field_a_field_123'.

We use `_` as part of our field names and need to be able to access the complete path of nested fields in our internal application.

- relates to #531
- fixes #688

### Checklist

### DIFF
--- a/packages/core/test/utils_test.js
+++ b/packages/core/test/utils_test.js
@@ -2716,6 +2716,27 @@ describe("utils", () => {
       });
     });
 
+    it("should handle idSeparator parameter", () => {
+      const schema = {
+        definitions: {
+          testdef: {
+            type: "object",
+            properties: {
+              foo: { type: "string" },
+              bar: { type: "string" },
+            },
+          },
+        },
+        $ref: "#/definitions/testdef",
+      };
+
+      expect(toIdSchema(schema, undefined, schema, {}, "rjsf", "/")).eql({
+        $id: "rjsf",
+        foo: { $id: "rjsf/foo" },
+        bar: { $id: "rjsf/bar" },
+      });
+    });
+
     it("should handle null form data for object schemas", () => {
       const schema = {
         type: "object",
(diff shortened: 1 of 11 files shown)


green7ea:
This commit follows `idPrefix` closely, it didn't update the docs since `idPrefix` isn't mentioned there either.

epicfaace (commented):
Thanks, this is great! idPrefix is mentioned in the docs (see https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/form-props/#idprefix), so can you please add a section for idSeparator as well?

green7ea:
Thanks for pointing out the documentation, I don't know how I missed it. I updated the commit:

- added documentation,
- added a line in the changelog for `v3.3.1 (upcoming)` pointing here.

I'm not sure if v3.3.1 was the right place to add the change so please let me know if I should move it.

heath-freenome:
This is awesome! I can use this fix as well

packages/core/test/utils_test.js
```diff
@@ -2716,6 +2716,27 @@ describe("utils", () => {
...
+        $ref: "#/definitions/testdef",
+      };
+
+      expect(toIdSchema(schema, undefined, schema, {}, "rjsf", "/")).eql({
+        $id: "rjsf",
+        foo: { $id: "rjsf/foo" },
+        bar: { $id: "rjsf/bar" },
+      });
+    });
+
```
epicfaace, line 2739: Thanks! One more thing. Can you add a test within Form_test.js that sets idSeparator and makes sure that the fields are rendered with the right idSeparator?

heath-freenome: @green7ea do you have time for adding this? Just one small change away from merging... if not, I can try

heath-freenome (approved):

epicfaace (approved):
lgtm

PR #2676, opened 2022-01-27
To improve performance, skip validating subschemas in oneOf / anyOf if formData is undefined

KittyJose:
Fixing getMatchingOption in the case of oneOf/anyOf by validating schema only when formData is available

When there is a property with oneOf/anyOf - schema gets validated when there is not formData provided which leads to lot of load time. This fix will improves speed and performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 # v4.0.2 (upcoming)
 
+## @rjsf/core
+
+- For performance, skip validating subschemas if formData is undefined (#2676)
+
 # v4.0.1
 
 - Bumped the peer dependencies of `@rjsf/core` to `^4.0.0` for all of themes in `package.json`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/core
 
-- For performance, skip validating subschemas if formData is undefined (#2676)
+- To improve performance, skip validating subschemas in oneOf / anyOf if formData is undefined (#2676)
 
 # v4.0.1
 
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1296,7 +1296,6 @@ export function getMatchingOption(formData, options, rootSchema) {
       if (isValid(augmentedSchema, formData, rootSchema)) {
         return i;
       }
-    
     } else if (isValid(option, formData, rootSchema)) {
       return i;
     }
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1291,7 +1291,7 @@ export function getMatchingOption(formData, options, rootSchema) {
       if (formData && isValid(augmentedSchema, formData, rootSchema)) {
         return i;
       }
-      return i;
+    
     } else if (formData && isValid(option, formData, rootSchema)) {
       return i;
     }
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1288,10 +1288,11 @@ export function getMatchingOption(formData, options, rootSchema) {
       // been filled in yet, which will mean that the schema is not valid
       delete augmentedSchema.required;
 
-      if (isValid(augmentedSchema, formData, rootSchema)) {
+      if (formData && isValid(augmentedSchema, formData, rootSchema)) {
         return i;
       }
-    } else if (isValid(option, formData, rootSchema)) {
+      return i;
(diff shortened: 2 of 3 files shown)


packages/core/src/utils.js
```diff
@@ -1230,10 +1230,11 @@ export function getMatchingOption(formData, options, rootSchema) {
       // been filled in yet, which will mean that the schema is not valid
       delete augmentedSchema.required;
 
-      if (isValid(augmentedSchema, formData, rootSchema)) {
+      if (formData && isValid(augmentedSchema, formData, rootSchema)) {
```
epicfaace, line 1233: Thanks. Can you add a test that covers this case, so it's clearer what issue this is fixing?

epicfaace (changes requested):

KittyJose:
@epicfaace , I have added a test to utils_test.js, in which I pass undefined as formData to getMatchingOption(), and it returns 0. My point is, if we dont pass formData to getMatchingOption => then there is no need to validate data against the whole schema.

In my schema, I use anyOfs (100 to 200 anyOfs) , and the check in utils.js passes undefined as formData as shown in the below pic. I also have a very large schema and multiple checks to validate formData against schema is very slow.

image

Maybe this is not the right solution, I can pass you my huge schema and you can paste it in playground and see it takes lot of time to load, eventually it loads after 3 minutes.

https://github.com/terminusdb/terminusdb-documents-ui/blob/main/seshatSchema.json

Any help would be really appreciated, we can set up a call as well, but do let me know what has to be done.

Thank you so much :)

KittyJose:
Ok, thanks @epicfaace, I shall join the call on discord this Friday 11 am ET  :)

packages/core/src/utils.js
epicfaace, line 1233: Change these in both places to `formData !== undefined`

epicfaace: Add a comment

```
// For performance, skip validating subschemas if formData is undefined (as they would return false).
```

epicfaace, line 1233: At the beginning of the function -- let's add:

```
  // For performance, skip validating subschemas if formData is undefined. We just
  // want to get the first option in that case.
  if (formData === undefined) {
    return 0;
  }
```

epicfaace: And remove the other changes in this file.

packages/core/test/utils_test.js
```diff
@@ -3843,6 +3843,33 @@ describe("utils", () => {
         }))
       );
     });
+    it("should infer correct anyOf schema based on data if passing undefined", () => {
```
epicfaace, line 3846: Let's add a test for getMatchingOption
- where formData is `null`
- Where option 2 is `{type: null}`
- We expect `getMatchingOption` to be equal to 2



```
options = [{type: "string"}, {type: "string"}, {type: "null"}];
expect(getMatchingOption(null, options, rootSchema)).eql(2);
```

epicfaace (commented):
Thanks! Can you fix the lint so the CI passes?

KittyJose:
@epicfaace ok I have fixed the lint issues :)

epicfaace:
Still failing -- see https://github.com/rjsf-team/react-jsonschema-form/runs/5189456113?check_suite_focus=true

Dhakkain:
I have a question about that - is it fix that issue too? #1564

rrooij:
The linter issues should be fixed. :smile:

KittyJose:
@epicfaace, Hello, Can you contact me on discord (Kitty Jose). I would like to discuss a matter with you. I cant seem to find ur username on discord.Thanks :D

epicfaace:
I'm not sure what your discord username is -- mine is Ashwin#0168.

epicfaace (approved):

epicfaace (changes requested):
Can you update this branch with the latest from master and add an entry to CHANGELOG.md?

rrooij:
We added an entry to the changelog and rebased to master. :smile:

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 # v4.0.2 (upcoming)
 
+## @rjsf/core
+
+- For performance, skip validating subschemas if formData is undefined (#2676)
```
epicfaace, line 24:
```suggestion
- To improve performance, skip validating subschemas in oneOf / anyOf if formData is undefined (#2676)
```

epicfaace: Can you make this change? I don't have access to edit. Thanks!

epicfaace (approved):

rrooij:
Did the change :+1:

epicfaace (approved):

PR #2759, opened 2022-03-11
v4.1.0

heath-freenome:
New Release


epicfaace (approved):
I don't think it's necessary to update the devDependencies for all the packages to 4.1.0 (keeping it at 4.0.1 will give us the same behavior), but it's fine nevertheless.

epicfaace:
> I don't think it's necessary to update the devDependencies for all the packages to 4.1.0 (keeping it at 4.0.1 will give us the same behavior), but it's fine nevertheless.

Or did lerna do that automatically?

heath-freenome:
> > I don't think it's necessary to update the devDependencies for all the packages to 4.1.0 (keeping it at 4.0.1 will give us the same behavior), but it's fine nevertheless.
> 
> Or did lerna do that automatically?

Yes it did

PR #2778, opened 2022-03-24
fix(@rjsf/antd): make FieldTemplate response correctly to empty description

mzvast:

As is shown [below,](https://rjsf-team.github.io/react-jsonschema-form/) antd theme template has a bug, when description is empty, it will still recogonize it as truthy so it add a block there, which is not as we expected.
image

The bottom logic in antd `Form.item` is as shown below:
image

Then we go a level up in the theme antd's`FieldTemplate'
image

Then our core package in `SchemaField`, there are two data source of description,`description` is a React.FC, while `rawDescription` is a string.
image

The solution to fix this is to tell from rawDescription whether it is empty, then if it's not empty we pass the description(React component) to it.

### Checklist

### DIFF
--- a/packages/antd/src/templates/FieldTemplate/index.js
+++ b/packages/antd/src/templates/FieldTemplate/index.js
@@ -27,6 +27,7 @@ const FieldTemplate = ({
   readonly,
   required,
   schema,
+  rawDescription,
   // uiSchema,
 }) => {
   const {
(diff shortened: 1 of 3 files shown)


heath-freenome:
@mzvast If you deal with the conflicts and update the `CHANGELOG.md` with information about the fix and what issue it closes, then I will approve

heath-freenome:
Closing this as it has been fixed in the v5 beta via #3112

PR #2795, opened 2022-04-06
Use idSeparator prop for array item ids (#2792)

### LINKED ISSUE #2792: ArrayField not taking in account the idSeparator
### Prerequisites

### Description

When id of an array are generated, I would expect that the index is also separated with the idSeparator. 

In the below example, I expect to have for first input (with hello value): 
- root:::0:::name (instead of currently root_0:::name)

### Steps to Reproduce

https://codesandbox.io/s/elastic-marco-izoptf?file=/src/index.js

### Version

I am using latest version (4.1.1)

### Code source
https://github.com/rjsf-team/react-jsonschema-form/blob/7023efa72d76c9017b3fa8aaea83f2b61373c51c/packages/core/src/components/fields/ArrayField.js#L492

lpillonel:

To prefix ids for items' array, underscore (ie "_0") was used in the id prefix.

This fixes #2792 

### Checklist

### DIFF
--- a/packages/core/src/components/fields/ArrayField.js
+++ b/packages/core/src/components/fields/ArrayField.js
@@ -475,7 +475,7 @@ class ArrayField extends Component {
       onBlur,
       onFocus,
       idPrefix,
-      idSeparator,
+      idSeparator = "_",
       rawErrors,
     } = this.props;
     const title = schema.title === undefined ? name : schema.title;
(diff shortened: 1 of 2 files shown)


heath-freenome:
> lgtm, but I think we should probably make this a breaking change! what do you think @nickgros @heath-freenome ?

The original change wasn't breaking, I don't see why this would be

heath-freenome:
> This looks like a breaking change, since the default separator for array elements will now be '/' (or the custom idSeparator) rather than '_'.
> 
> Are there any downsides to making the array separator identical by default to the object field separator, rather than potentially a separate character?

Nick, what you are seeing is that the '/' was a test change where the hard-coded `_` separator is now the proper one specified by the `idSeparator` param

nickgros:
@heath-freenome I see, so this will truly only "break" for someone that depended on the bug in the first place. Makes sense, thanks!

epicfaace:
> @heath-freenome I see, so this will truly only "break" for someone that depended on the bug in the first place. Makes sense, thanks!

I still think this might be a breaking change. Added to the agenda to discuss: https://docs.google.com/document/d/12PjTvv21k6LIky6bNQVnsplMLLnmEuypTLQF8a-8Wss/edit

epicfaace (approved):

PR #2796, opened 2022-04-07
Calling onChange on the Form's parent due to props no longer throws an error

danielbrown96:

Previously, the 'onChange' method called from inside the Form component's UNSAFE_componentWillReceiveProps method would cause React to throw a "Cannot update a component from inside the function body of a different component" error. 

This PR updates the code based on the suggestions in [this react documentation](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) to use componentDidUpdate instead, fixing the issue to allow the parent to be able to control the form's data while still benefitting from the logic in the form itself, with minimal impact to the rest of the codebase.

### Checklist

### DIFF
--- a/packages/core/src/components/Form.js
+++ b/packages/core/src/components/Form.js
@@ -44,14 +44,17 @@ export default class Form extends Component {
 
   UNSAFE_componentWillReceiveProps(nextProps) {
     const nextState = this.getStateFromProps(nextProps, nextProps.formData);
-    if (
-      !deepEquals(nextState.formData, nextProps.formData) &&
-      !deepEquals(nextState.formData, this.state.formData) &&
-      this.props.onChange
-    ) {
-      this.props.onChange(nextState);
-    }
     this.setState(nextState);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (
+      this.props.onChange &&
+      !deepEquals(this.props.formData, this.state.formData) &&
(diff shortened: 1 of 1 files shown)


epicfaace:
Hey @jimmycallin , do you think this PR would be a smaller version to eventually get to https://github.com/rjsf-team/react-jsonschema-form/pull/2010? or do the changes in this PR constitute a breaking change?

jimmycallin:
I haven't tested it, but it looks like a non-breaking approach to me! The unit tests surrounding this are fairly covering, so i think it should be fine.

nickgros:
Seems like an improvement to me. I'm having a really hard time wrapping my head around this case where a caller is treating Form as a "partially-controlled" component. 

Is it true that `onChange` in `componentDidUpdate` will only be invoked in this weird case? I'm assuming so because I can't come up with a scenario in which the formData prop updated, but is still out-of-sync with the formData state. If so, should we log a warning that we're going to break this in the future, and that the user should treat Form as fully controlled or fully uncontrolled?

heath-freenome:
The call to `setState()` in `UNSAFE_componentWillReceiveProps` will, I believe, cause the `componentDidUpdate()` to get called... I don't see an issue with this

heath-freenome:
Actually, the issue is that the `this.props.onChange()` call should actually be `nextProps.onChange()`, this is being fixed in rjsf v5 via #3052

PR #2834, opened 2022-04-27
made statelessComponent to FunctionComponent cause stateless is depre…

kokoyroy:
made statelessComponent to FunctionComponent cause stateless is deprecated

### made statelessComponent to FunctionComponent cause stateless is deprecated

### Checklist

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -12,7 +12,7 @@ declare module '@rjsf/core' {
         acceptcharset?: string;
         action?: string;
         additionalMetaSchemas?: ReadonlyArray<object>;
-        ArrayFieldTemplate?: React.StatelessComponent<ArrayFieldTemplateProps>;
+        ArrayFieldTemplate?: React.FunctionComponent<ArrayFieldTemplateProps>;
         autoComplete?: string;
         autocomplete?: string; // deprecated
         children?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


packages/core/index.d.ts
```diff
@@ -37,7 +37,7 @@ declare module '@rjsf/core' {
         name?: string;
         noHtml5Validate?: boolean;
         noValidate?: boolean;
-        ObjectFieldTemplate?: React.StatelessComponent<ObjectFieldTemplateProps>;
+        ObjectFieldTemplate?: React.FunctionComponent<ObjectFieldTemplateProps>;
```
epicfaace, line 40: Are there issues with FunctionComponent (see https://github.com/facebook/create-react-app/pull/8177)? Or do those not apply here?

kokoyroy: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/34e20b805043483182fb77a96008244a70874dd2/types/react/v17/index.d.ts#L542

children are now optional as i understand better in line 822
```type PropsWithChildren<P> = P & { children?: ReactNode | undefined };```

epicfaace (approved):

packages/core/index.d.ts
heath-freenome: > Also it would be worth to extract TitleFieldProps and DescriptionFieldProps types.

I am doing that as part of a v5 change coming in about 1.5 months

epicfaace: > So it would be nice to add optional children prop for backward compatibility + more powerful extensebility.

Thanks. Can you add a PR that does this?

PR #2839, opened 2022-04-29
a

stuartwilkes:

As described [here](https://github.com/rjsf-team/react-jsonschema-form/issues/2752), if an allOf block is present at multiple levels within a schema, and the nested allOf block implements multiple if/then/else statements, the nested dependency is not correctly rendered.

This PR takes the patch described in the linked thread and implements it within the library to fix the issue as described. It also provides a test for the issue.

The current behaviour can be seen as failing in [this playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJBbmltYWwiOiJDYXQiLCJUYWlsIjoiTG9uZyJ9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsiQW5pbWFsIjp7ImRlZmF1bHQiOiJDYXQiLCJlbnVtIjpbIkNhdCIsIkRvZyJdLCJ0aXRsZSI6IkFuaW1hbCIsInR5cGUiOiJzdHJpbmcifX0sImFsbE9mIjpbeyJpZiI6eyJyZXF1aXJlZCI6WyJBbmltYWwiXSwicHJvcGVydGllcyI6eyJBbmltYWwiOnsiY29uc3QiOiJDYXQifX19LCJ0aGVuIjp7InByb3BlcnRpZXMiOnsiVGFpbCI6eyJkZWZhdWx0IjoiTG9uZyIsImVudW0iOlsiTG9uZyIsIlNob3J0IiwiTm9uZSJdLCJ0aXRsZSI6IlRhaWwgbGVuZ3RoIiwidHlwZSI6InN0cmluZyJ9fSwicmVxdWlyZWQiOlsiVGFpbCJdfX0seyJpZiI6eyJyZXF1aXJlZCI6WyJBbmltYWwiXSwicHJvcGVydGllcyI6eyJBbmltYWwiOnsiY29uc3QiOiJEb2cifX19LCJ0aGVuIjp7InByb3BlcnRpZXMiOnsiQnJlZWQiOnsidGl0bGUiOiJCcmVlZCIsInByb3BlcnRpZXMiOnsiQnJlZWROYW1lIjp7ImRlZmF1bHQiOiJBbHNhdGlhbiIsImVudW0iOlsiQWxzYXRpYW4iLCJEYWxtYXRpb24iXSwidGl0bGUiOiJCcmVlZCBuYW1lIiwidHlwZSI6InN0cmluZyJ9fSwiYWxsT2YiOlt7ImlmIjp7InJlcXVpcmVkIjpbIkJyZWVkTmFtZSJdLCJwcm9wZXJ0aWVzIjp7IkJyZWVkTmFtZSI6eyJjb25zdCI6IkFsc2F0aWFuIn19fSwidGhlbiI6eyJwcm9wZXJ0aWVzIjp7IkZ1ciI6eyJkZWZhdWx0IjoiYnJvd24iLCJlbnVtIjpbImJsYWNrIiwiYnJvd24iXSwidGl0bGUiOiJGdXIiLCJ0eXBlIjoic3RyaW5nIn19LCJyZXF1aXJlZCI6WyJGdXIiXX19LHsiaWYiOnsicmVxdWlyZWQiOlsiQnJlZWROYW1lIl0sInByb3BlcnRpZXMiOnsiQnJlZWROYW1lIjp7ImNvbnN0IjoiRGFsbWF0aW9uIn19fSwidGhlbiI6eyJwcm9wZXJ0aWVzIjp7IlNwb3RzIjp7ImRlZmF1bHQiOiJzbWFsbCIsImVudW0iOlsibGFyZ2UiLCJzbWFsbCJdLCJ0aXRsZSI6IlNwb3RzIiwidHlwZSI6InN0cmluZyJ9fSwicmVxdWlyZWQiOlsiU3BvdHMiXX19XSwicmVxdWlyZWQiOlsiQnJlZWROYW1lIl19fX19XSwicmVxdWlyZWQiOlsiQW5pbWFsIl19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJ2YWxpZGF0ZSI6ZmFsc2UsImRpc2FibGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==)

### Checklist

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -767,6 +767,26 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
     return resolveCondition(schema, rootSchema, formData);
   }
 
+  if (resolvedSchema.properties) {
+    var properties = {};
+
+    Object.entries(resolvedSchema.properties).forEach(entries => {
+      var propName = entries[0];
+      var propSchema = entries[1];
+      var propData = formData && formData[propName];
+      var resolvedPropSchema = retrieveSchema(propSchema, rootSchema, propData);
+
+      properties[propName] = resolvedPropSchema;
+
+      if (
+        propSchema !== resolvedPropSchema &&
+        resolvedSchema.properties !== properties
+      ) {
+        resolvedSchema = { ...resolvedSchema, properties };
+      }
+    });
+  }
+
   if ("allOf" in schema) {
     try {
       resolvedSchema = mergeAllOf({
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -768,19 +768,25 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
   }
 
   if (resolvedSchema.properties) {
-    var properties = {};
+    const properties = {};
 
     Object.entries(resolvedSchema.properties).forEach(entries => {
-      var propName = entries[0];
-      var propSchema = entries[1];
-      var propData = formData && formData[propName];
-      var resolvedPropSchema = retrieveSchema(propSchema, rootSchema, propData);
+      const propName = entries[0];
+      const propSchema = entries[1];
+      const rawPropData = formData && formData[propName];
+      const propData = isObject(rawPropData) ? rawPropData : {};
+      const resolvedPropSchema = retrieveSchema(
+        propSchema,
+        rootSchema,
+        propData
+      );
 
       properties[propName] = resolvedPropSchema;
 
       if (
         propSchema !== resolvedPropSchema &&
-        resolvedSchema.properties !== properties
+        resolvedSchema.properties !== properties &&
+        properties !== null
       ) {
         resolvedSchema = { ...resolvedSchema, properties };
       }
--- a/packages/core/test/utils_test.js
+++ b/packages/core/test/utils_test.js
@@ -2540,6 +2540,282 @@ describe("utils", () => {
           required: ["animal", "food"],
         });
       });
+      it("should resolve multiple conditions in nested allOf blocks", () => {
+        const schema = {
+          type: "object",
+          properties: {
+            Animal: {
+              default: "Cat",
+              enum: ["Cat", "Dog"],
+              title: "Animal",
+              type: "string",
+            },
+          },
+          allOf: [
+            {
+              if: {
+                required: ["Animal"],
+                properties: {
+                  Animal: {
+                    const: "Cat",
+                  },
+                },
+              },
+              then: {
+                properties: {
+                  Tail: {
+                    default: "Long",
+                    enum: ["Long", "Short", "None"],
+                    title: "Tail length",
+                    type: "string",
+                  },
+                },
+                required: ["Tail"],
+              },
+            },
+            {
+              if: {
+                required: ["Animal"],
+                properties: {
+                  Animal: {
+                    const: "Dog",
+                  },
+                },
+              },
+              then: {
+                properties: {
+                  Breed: {
+                    title: "Breed",
+                    properties: {
+                      BreedName: {
+                        default: "Alsatian",
+                        enum: ["Alsatian", "Dalmation"],
+                        title: "Breed name",
+                        type: "string",
+                      },
+                    },
+                    allOf: [
+                      {
+                        if: {
+                          required: ["BreedName"],
+                          properties: {
+                            BreedName: {
+                              const: "Alsatian",
+                            },
+                          },
+                        },
+                        then: {
+                          properties: {
+                            Fur: {
+                              default: "brown",
+                              enum: ["black", "brown"],
+                              title: "Fur",
+                              type: "string",
+                            },
+                          },
+                          required: ["Fur"],
+                        },
+                      },
+                      {
+                        if: {
+                          required: ["BreedName"],
+                          properties: {
+                            BreedName: {
+                              const: "Dalmation",
+                            },
+                          },
+                        },
+                        then: {
+                          properties: {
+                            Spots: {
+                              default: "small",
+                              enum: ["large", "small"],
+                              title: "Spots",
+                              type: "string",
+                            },
+                          },
+                          required: ["Spots"],
+                        },
+                      },
+                    ],
+                    required: ["BreedName"],
+                  },
+                },
+              },
+            },
+          ],
+          required: ["Animal"],
+        };
+        const definitions = {};
+        const formData = {
+          Animal: "Cat",
+          Tail: "Long",
+        };
+
+        expect(retrieveSchema(schema, { definitions }, formData)).eql({
+          type: "object",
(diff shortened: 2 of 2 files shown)


packages/core/src/utils.js
```diff
@@ -767,6 +767,32 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
...
+        rootSchema,
+        propData
+      );
+
+      properties[propName] = resolvedPropSchema;
+
+      if (
+        propSchema !== resolvedPropSchema &&
+        resolvedSchema.properties !== properties &&
+        properties !== null
```
AlimovSV, line 789: I think `properties !== null` check is redundand, isn't?

stuartwilkes: Thanks, you're right - this was a holdover from when I was addressing an issue that caused the patch to fail [this test](https://github.com/rjsf-team/react-jsonschema-form/blob/d87a88b99c1e9f793b414cb32d412ea9b59293ec/packages/core/test/utils_test.js#L1096)

packages/core/test/utils_test.js
```diff
@@ -2540,6 +2540,282 @@ describe("utils", () => {
...
+              default: "Long",
+              enum: ["Long", "Short", "None"],
+              title: "Tail length",
+              type: "string",
+            },
+          },
+          required: ["Animal", "Tail"],
+        });
+      });
+      it("should resolve multiple conditions in nested allOf blocks", () => {
```
nickgros, line 2674: Can we change the test names to capture what the difference is between the tests, and more importantly so we don't have two tests with the same name?

packages/core/src/utils.js
```diff
@@ -767,6 +767,31 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
     return resolveCondition(schema, rootSchema, formData);
   }
 
+  if (resolvedSchema.properties) {
+    const properties = {};
+
+    Object.entries(resolvedSchema.properties).forEach(entries => {
+      const propName = entries[0];
+      const propSchema = entries[1];
```
nickgros, line 775: Is it possible that a resolved schema will have additional properties such that this ordering doesn't hold? Seems like it would make more sense to filter

packages/core/src/utils.js
```diff
@@ -767,6 +767,31 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
     return resolveCondition(schema, rootSchema, formData);
   }
 
+  if (resolvedSchema.properties) {
```
nickgros, line 770: Could you add a comment clarifying why we have to do this new step, and what exactly we're doing? It's hard to follow exactly what's going on.

packages/core/src/utils.js
```diff
@@ -767,6 +767,31 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
...
+        propSchema,
+        rootSchema,
+        propData
+      );
+
+      properties[propName] = resolvedPropSchema;
+
+      if (
+        propSchema !== resolvedPropSchema &&
+        resolvedSchema.properties !== properties
```
nickgros, line 788: It looks like `properties` will always be a different object than `resolvedSchema.properties`, since you declared a new `properties` object on line 771. So I think this would always be true.

packages/core/test/utils_test.js
nickgros: Follow up question: are both tests necessary? Do they capture separate edge cases that we effectively handle differently?

packages/core/src/utils.js
AlimovSV: No, this is possible ONLY when the recursive call of retrieveSchema returns a modified schema for the specific property. Otherwise resolvedSchema will be returned unchanged.

packages/core/src/utils.js
AlimovSV: Could you provide the schema which is not handled correctly by this code? For better understanding the case you are talking about. I use this code in the project with large and complex schemas with nested if/then/else so if you see the case which will not work - tell me please.

packages/core/test/utils_test.js
stuartwilkes: Thanks, you're right, we don't need both these tests, only the second one, which captures the failing behaviour in the current live code and where the new code addresses this issue. Removed.

packages/core/src/utils.js
stuartwilkes: I believe as we're using `entries` here, and the properties will always be objects representing the fields, index 0 will always be `propName` and index 1 the subschema ?

AlimovSV: Sure, but I guess @nickgros is talking about something like `.additionalProperties`.

nickgros: I see now, I think I was misunderstanding the code. This makes sense to me. Thanks for the clarification

packages/core/src/utils.js
AlimovSV: The idea of this check is to reduce creation of shallow copy for the retrieved schemas. So if a schema isn't computed schema (doesn't have any if/then/else) it will be returned unchanged (and reference equality check will be enough for this condition)

packages/core/src/utils.js
AlimovSV: finding the right words is not my best side :) I like your description

packages/core/src/utils.js
nickgros: Sorry if I'm being unclear, but essentially I don't think you will ever have referential equality between `resolvedSchema.properties` and `properties` because `properties` is always a new object.

AlimovSV: No, let me explain. Before loop we initialize `properties` variable as a new object, you right. And during the iteration (first time when we meet `propSchema !== resolvedPropSchema`) the condition `resolvedSchema.properties !== properties` will be truthy, you right too. But after we make a shallow copy of the schema with new properties container `resolvedSchema = { ...resolvedSchema, properties };` during any next iteration `resolvedSchema.properties !== properties` will be falsy and we will not make shallow copy. The bonuse is: when resolvedSchema doesn't contain properties with computed schema we do not make shallow copy at all and return the unmodified object.

nickgros: 🤦‍♂️ I finally see it now! Thank you for this explanation (and for bearing with me)!

nickgros (approved):
This looks good to me if tests are passing. I'd like another set of eyes since our schema resolution utils are so complicated

jacqueswho (approved):

AlimovSV (approved):

PR #2631, opened 2021-12-10
clear errors from "additionalProperties: false"

Lausselloic:
### Reasons for making this change

if additionalProperties is set to `false` omit extra data doesn't work as expected

Add a test to validate it. Before fixes test failed, after fix test success

Fixes #1507

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
  - [x] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/master/CHANGELOG.md) with a description of the PR
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #2853, opened 2022-05-06
fix(@rjsf/core): correctly omit extra data when additionalProperty is an object (#2667)

### LINKED ISSUE #2667: Passing additionalProperties does not work with omitExtraData
https://github.com/rjsf-team/react-jsonschema-form/blob/63636b7af2189bc151dfaa261318f167d6bb36e6/packages/core/src/utils.js#L1041

Passing additionalProperties = false does not work because in the line above. It should check if it is true instead of checking the precence:

http://json-schema.org/understanding-json-schema/reference/object.html#additional-properties

lpillonel:

fixes #2667 

When additionalProperties are an object, the data omission triggered with `liveOmit` and `omitExtraData` was skipped due to the flag set here :
https://github.com/rjsf-team/react-jsonschema-form/blob/36ee173e35f4859a624b7620426e9a009dafe811/packages/core/src/utils.js#L1142

Solution is to ensure that additionalProperties value is `true` in order to set the flag

### Checklist


packages/core/src/utils.js
```diff
@@ -1139,7 +1139,10 @@ export function toPathSchema(schema, name = "", rootSchema, formData = {}) {
     return toPathSchema(_schema, name, rootSchema, formData);
   }
 
-  if (schema.hasOwnProperty("additionalProperties")) {
+  if (
+    schema.hasOwnProperty("additionalProperties") &&
+    schema.additionalProperties === true
+  ) {
```
epicfaace, line 1145: Thanks can you add a test?

lpillonel: Test added; I amended the previous commit

lpillonel:
@epicfaace 
Do you still expect anything to get this merged ?

heath-freenome:
@lpillonel There is another [fix](https://github.com/rjsf-team/react-jsonschema-form/pull/2631/files) that is checking if `additionalProperties` is not false. It sounds like the more correct fix is yours?

heath-freenome:
Fixed in version 5 beta via https://github.com/rjsf-team/react-jsonschema-form/pull/2940

PR #2876, opened 2022-06-07
fix: allow UISchemaSubmitButtonOptions properties to be undefined

maxpou:

How to reproduce:

In a TS context, when you create a form like the following:

```ts
const uiSchema: UiSchema = {
  'ui:submitButtonOptions': {
    submitText: 'Save and next',
    // norender: false,
    // props: {},
  },
};

function Part2(): JSX.Element {

  return (
    <Form
      className="Ssa-Part2"
      schema={whatever}
      uiSchema={uiSchema}
    />
  );
}
``` 

TS will complain because `props` and `norender` are needed. We have to add them with the default values. Since these 3 props don't seem to be interconnected, I'd suggest making the 3 optional.

Regarding the checklog, do I need to update the changelog?

Have a nice day :)

### Checklist


epicfaace (approved):
Thanks!

PR #2879, opened 2022-06-10
[updated] trim empty objects to allow correct validation of optional objects with mandatory fields

azertyfun:

This is a rebase and fix of #1228, which is stale. See also #675, #682

I've kept the change exactly the same (trusting @numical's work there), beyond rebasing I merely added b9f462e2176b4471ca24e60b92f43c455069ed6e which fixes the tests. I don't know if it makes sense to also edit the documentation beyond the changelog, or if this should be implicit behavior?

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 4.2.1 (upcoming)
 
+## @rjsf/core
+- Optional object with empty values are now trimmed to avoid "required" errors (#675)
+
 # 4.2.0
 
 ## @rjsf/core
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1365,9 +1365,9 @@ export function schemaRequiresTrueValue(schema) {
   return false;
 }
 
-function trimObject(object) {
-  return Object.entries(object).reduce((acc, [key, value]) => {
-    const trimmed = trimEmptyValues(value);
+function trimObject(schema, object) {
+  return Object.entries(schema.properties).reduce((acc, [key, value]) => {
+    const trimmed = trimEmptyValues(value, object[key]);
     if (trimmed || trimmed === false) {
       if (!acc) {
         acc = {};
@@ -1378,9 +1378,9 @@ function trimObject(object) {
   }, undefined);
 }
 
-function trimArray(array) {
+function trimArray(schema, array) {
   return array.reduce((acc, value) => {
-    const trimmed = trimEmptyValues(value);
+    const trimmed = trimEmptyValues(schema.items, value);
     if (trimmed || trimmed === false) {
       acc.push(trimmed);
     }
@@ -1388,10 +1388,10 @@ function trimArray(array) {
   }, []);
 }
 
-export function trimEmptyValues(value) {
-  return Array.isArray(value)
-    ? trimArray(value) || []
-    : typeof value === "object"
-    ? trimObject(value) || {}
+export function trimEmptyValues(schema, value) {
+  return schema.type === "array"
+    ? trimArray(schema, value) || []
+    : schema.type === "object"
+    ? trimObject(schema, value) || {}
     : value;
 }
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1390,8 +1390,8 @@ function trimArray(array) {
 
 export function trimEmptyValues(value) {
   return Array.isArray(value)
-    ? trimArray(value)
+    ? trimArray(value) || []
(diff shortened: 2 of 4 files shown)


azertyfun:
I just realized that I accidentally brought back a test file. Currently fixing that.

azertyfun:
I just realized that I accidentally brought back a test file. Currently fixing that.

azertyfun:
Actually the more I look at this the less I understand.. The original PR only trims extra data on validation, but doesn't do anything for the submitted data.

With this schema:

```json
{
    "type": "object",
    "properties": {
        "test": {
            "type": "object",
            "properties": {"nested": {"type": "string", "required": "true"}}
        }
    }
}
```

If no field is filled then the form will validate successfully but the submitted data will be `{test: { nested: undefined }}` when I would expect `{}`. Fine, whatever, this looks like intended behavior (at the time?) and may be desirable and is easily dealt with by the user in the `onSubmit` handler.

---

My second issue is that I don't know what the intended behavior is when dealing with the tests for `oneOf`/`allOf`. The original implementation didn't consider the schema at all, but the trimmed data gets rid of an empty string which is actually meaningful in those cases (and I assume everywhere else that isn't a text field!). Essentially we're just _guessing_ that `""` is the user not filling in a field, but the test suite uses `""` as a meaningful "other" choice instead (which isn't semantically wrong IMO).

```json
{
  "type": "object",
  "properties": {
    "street_address": {
      "type": "string"
    },
    "country": {
      "enum": [
        "United States of America",
        "Canada",
        "United Kingdom",
        "France"
      ]
    }
  },
  "allOf": [
    {
      "if": {
        "properties": {
          "country": {
            "const": "United States of America"
          }
        }
      },
      "then": {
        "properties": {
          "zipcode": {
            "type": "string"
          }
        }
      }
    },
    {
      "if": {
        "properties": {
          "country": {
            "const": "United Kingdom"
          }
        }
      },
      "then": {
        "properties": {
          "postcode": {
            "type": "string"
          }
        }
      }
    },
    {
      "if": {
        "properties": {
          "country": {
            "const": "France"
          }
        }
      },
      "then": {
        "properties": {
          "telephone": {
            "type": "string"
          }
        }
      }
    }
  ]
}
```

This leads me to believe that maybe **the right approach is to detect empty filed _in the UI_** and set the to `undefined` when empty? Then we can change `trimEmptyValues` to `trimUndefinedValues` which is more meaningful.

azertyfun:
Actually the more I look at this the less I understand.. The original PR only trims extra data on validation, but doesn't do anything for the submitted data.

With this schema:

```json
{
    "type": "object",
    "properties": {
        "test": {
            "type": "object",
            "properties": {"nested": {"type": "string", "required": "true"}}
        }
    }
}
```

If no field is filled then the form will validate successfully but the submitted data will be `{test: { nested: undefined }}` when I would expect `{}`. Fine, whatever, this looks like intended behavior (at the time?) and may be desirable and is easily dealt with by the user in the `onSubmit` handler.

---

My second issue is that I don't know what the intended behavior is when dealing with the tests for `oneOf`/`allOf`. The original implementation didn't consider the schema at all, but the trimmed data gets rid of an empty string which is actually meaningful in those cases (and I assume everywhere else that isn't a text field!). Essentially we're just _guessing_ that `""` is the user not filling in a field, but the test suite uses `""` as a meaningful "other" choice instead (which isn't semantically wrong IMO).

```json
{
  "type": "object",
  "properties": {
    "street_address": {
      "type": "string"
    },
    "country": {
      "enum": [
        "United States of America",
        "Canada",
        "United Kingdom",
        "France"
      ]
    }
  },
  "allOf": [
    {
      "if": {
        "properties": {
          "country": {
            "const": "United States of America"
          }
        }
      },
      "then": {
        "properties": {
          "zipcode": {
            "type": "string"
          }
        }
      }
    },
    {
      "if": {
        "properties": {
          "country": {
            "const": "United Kingdom"
          }
        }
      },
      "then": {
        "properties": {
          "postcode": {
            "type": "string"
          }
        }
      }
    },
    {
      "if": {
        "properties": {
          "country": {
            "const": "France"
          }
        }
      },
      "then": {
        "properties": {
          "telephone": {
            "type": "string"
          }
        }
      }
    }
  ]
}
```

This leads me to believe that maybe **the right approach is to detect empty filed _in the UI_** and set the to `undefined` when empty? Then we can change `trimEmptyValues` to `trimUndefinedValues` which is more meaningful.

heath-freenome:
@azertyfun As part of the pending v5 release we are moving utils and validation into separate packages. It also sounds like you aren't confident that this fix is doing what is intended?

PR #2912, opened 2022-07-08
Fix incorrect react syntax inside MUI ArrayFieldTemplate (broken hook order)

cerpins:

Misuse of functional react is causing following console error in my application (which I'm trying to avoid). 
image
After investigation, I notice that `DefaultArrayItem` is invoked as a function, not as a JSX.Element. Inside of `DefaultArrayItem` a hook is called, and since the component is called conditionally (as callback to `.map`) and not as an JSX.Element but a function, this plays out as broken hook order.

Could not find if this is related to any existing tickets

Please double check PR and let me know how I can get this merged, thanks.

### Checklist

  - [X] Fix error by invoking local component as a JSX component, instead of function
  - [X] Pass key to JSX component so that there is no console.error about not using key

### DIFF
--- a/packages/material-ui/src/ArrayFieldTemplate/ArrayFieldTemplate.tsx
+++ b/packages/material-ui/src/ArrayFieldTemplate/ArrayFieldTemplate.tsx
@@ -49,7 +49,7 @@ const ArrayFieldDescription = ({ DescriptionField, idSchema, description }: Arra
 };
 
 // Used in the two templates
-const DefaultArrayItem = (props: any) => {
+const DefaultArrayItem = (props: { extra: any }) => {
   const { ArrowDownwardIcon, ArrowUpwardIcon, Box, Grid, IconButton, Paper, RemoveIcon } = useMuiComponent();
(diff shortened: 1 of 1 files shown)


heath-freenome:
While this is a good fix, in August we will be releasing rjsf v5 beta that removes the entire hook issue entirely. Unfortunately this will be a breaking release. How seriously is this causing issues for you?

cerpins:
> While this is a good fix, in August we will be releasing rjsf v5 beta that removes the entire hook issue entirely. Unfortunately this will be a breaking release. How seriously is this causing issues for you?

Thanks for the reply. It's mildly bothersome, but nothing urgent that I couldn't wait on. Primary issue is that I don't want my test suite polluted with library errors, since I have tests configured to fail on (unexpected) console errors. As long as you are working on it one way or another, I am happy :)

heath-freenome:
This was fixed by removing hooks entirely from the `@rjsf/material-ui` and new `@rjsf/mui` themes

PR #2931, opened 2022-07-15
Optimized and updated imports in core package.json

heath-freenome:

- Bumped `prettier` in the main `package.json` file to pickup support for the latest typescript
  - Also changed to using `"bracketSameLine": false,` since `jsxBracketSameLine` is deprecated
- Updated `tsdx` build to also run umd, eliminating the need for the one-off webpack version build
- Built `package-lock.json` with node 16
- Deleted the `test-old` and `build-umd` scripts
- Added `@rjsf/utils` to `peerDependencies`
- Updated the `cs-check`, `cs-format` and `lint-staged` globs to include typescript
- Removed all of the webpack and cross-env related packages no longer needed for the one-off build
- Also removed `express`, `gh-pages`, `estraverse*`, `prettier` and `husky`
- Updated a bunch of the `peerDependencies` package versions to the latest
- Deleted the now unnecessary `webpack.config.dist.js`
- Ran `cs-format` on core and playground to fix issues failing the build

### Checklist


packages/core/package.json
```diff
@@ -3,22 +3,20 @@
...
     "precommit": "lint-staged",
     "publish-to-npm": "npm run build && npm publish",
     "start": "tsdx watch",
     "test": "tsdx test",
     "test:watch": "tsdx test --watch",
     "test:update": "tsdx test --u",
-    "test-old": "cross-env BABEL_ENV=test NODE_ENV=test mocha --require @babel/register --require ./test/setup-jsdom.js test/**/*_test.js",
     "test-coverage": "tsdx test --coverage"
   },
   "lint-staged": {
```
nickgros, line 18: Does the lint-staged rule have to be here? Can we move it to the base package.json and remove it from all sub-packages for consistency/single point of control?

heath-freenome: I'm not sure... I do know that different packages have different globs for the files to check so, for the moment I think we need to keep them per package

PR #2958, opened 2022-07-22
Switched playground to webpack 5 and other small fixes

heath-freenome:

- Updated `.github/workflows/release.yml` to use node 16
- Updated the main `package.json` to remove `webpack` and `webpack-cli` as they are only needed by `playground`
- Updated all of the `package.json` files for the themes to fix the `build` script to change `--format cjs,es,umd` to `--format cjs,esm,umd`
  - This is because `dts-cli` changed `es` to `esm`
  - Also updated `material-ui` to add that missing flag to `build`
- Updated `.babelrc` to get it working with webpack 5
- Updated `package.json` for playground to update (or remove) `webpack` and all of its plugins and utilities
  - Explicitly added `ajv8` since `webpack-dev-server` uses it and the `validator-ajv6` has an earlier version
- Updated the `webpack.config.*.js` files to support webpack 5
- Updated the `index.js` for playground to organize the themes in alphabetical order (after default)
- Bumped `antd` to latest

### Checklist


packages/playground/package.json
```diff
@@ -84,40 +84,38 @@
     "@material-ui/icons": "^4.11.3",
     "@mui/icons-material": "^5.8.4",
     "atob": "^2.1.2",
+    "ajv": "^8.11.0",
```
nickgros, line 87: v8? Do we even need this?

nickgros (approved):
Not a babel/webpack expert but LGTM. See note about ajv on package.json

packages/playground/package.json
heath-freenome: We do... see this remark in the commit description:?
```
Explicitly added ajv8 since webpack-dev-server uses it and the validator-ajv6 has an earlier version
```

PR #2986, opened 2022-08-04
Prepare utils and tests for upcoming BaseInputTemplate work

heath-freenome:

Adding support for a new `BaseInputTemplate` with a new utility. Also made all of the tests in the themes have the full superset of different tests that were scattered across each theme.

- Added the `BaseInputTemplate` type into the `TemplatesType`
    - Refactored code from `BaseInput` in `@rjsf/core` as `getInputProps()`
      - Added 100% unit tests
    - Consolidated all the various tests in different themes into all of the other themes
      - Also uncommented the broken slider tests in material-ui themes after figuring out how to make them work
    - Updated the snapshots for all the new tests in all themes
 
### Checklist


packages/utils/src/getInputProps.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (inputProps.step === undefined) {
+        // Since this is integer, you always want to step up or down in multiples of 1
+        inputProps.step = 1;
+      }
+    }
+  }
+
+  if (options.autocomplete) {
+    inputProps.autoComplete = options.autocomplete;
+  }
```
nickgros, line 47: unnecessary conditional?


```suggestion
inputProps.autoComplete = options.autocomplete;
```

nickgros (approved):

packages/utils/src/getInputProps.ts
heath-freenome: Maybe not. If `options.autocomplete` were set to a falsy value, that value it would get set onto the `inputProps`, which I'm not sure we want

PR #2989, opened 2022-08-05
Implement BaseInputTemplate in antd

heath-freenome:

- Refactored `TextWidget` as `BaseInputTemplate`, using the `getInputProps()` to implement `core` feature
- Deleted the following widgets because they are now just duplicate code to what is implemented with the template in `core`
  - `ColorWidget`, `EmailWidget`, `URLWidget`, `UpDownWidget`
  - Updated the `widgets` overrides in `index.js` to remove the deleted class references
- Updated the test snapshots to match the subtle little changes caused by the `getInputProps()` feature and increased use of the `BaseInputTemplate`

### Checklist


packages/antd/test/__snapshots__/Form.test.js.snap
```diff
@@ -522,37 +522,23 @@ exports[`single fields format color 1`] = `
   <div
     className="form-group field field-string"
   >
-    <label
-      className="ant-checkbox-wrapper"
+    <input
```
nickgros, line 525: I don't really understand why this changed. Seems like it might be a problem that we lost the `label` tag?

heath-freenome: I think it has something to do with the old `Color` widget importing a different input, which in my mind seems wrong.
```
import Input from "antd/lib/checkbox";
```

I'll test it in the old and new playground and see...

heath-freenome: The old playground just looks weird with a checkbox and plain-old doesn't actually show the selected color at all:
Screen Shot 2022-08-07 at 12 58 35 AM

Where as this implementation does:
Screen Shot 2022-08-07 at 1 12 19 AM

nickgros: Still looks weird, but better I guess, so let's go with it

nickgros (approved):

PR #2997, opened 2022-08-06
Remove unnecessary lodash uses

jimmycallin:

Just removing some lodash uses that is more efficiently served by built-in functions, to reduce bundle size.

### Checklist

### DIFF
--- a/packages/antd/src/templates/ObjectFieldTemplate/index.js
+++ b/packages/antd/src/templates/ObjectFieldTemplate/index.js
@@ -1,5 +1,7 @@
 import React from "react";
 import classNames from "classnames";
+import isObject from "lodash/isObject";
+import isNumber from "lodash/isNumber";
 
 import { canExpand } from "@rjsf/utils";
 import Button from "antd/lib/button";
@@ -61,12 +63,12 @@ const ObjectFieldTemplate = ({
         ? 24
         : 12;
 
-    if (colSpan instanceof Object) {
+    if (isObject(colSpan)) {
       return (
         colSpan[widget] || colSpan[field] || colSpan[type] || defaultColSpan
       );
     }
-    if (typeof colSpan === "number") {
+    if (isNumber(colSpan)) {
       return colSpan;
     }
     return defaultColSpan;
--- a/packages/antd/src/templates/ObjectFieldTemplate/index.js
+++ b/packages/antd/src/templates/ObjectFieldTemplate/index.js
@@ -1,6 +1,5 @@
 import React from "react";
 import classNames from "classnames";
-import _ from "lodash";
 
 import { canExpand } from "@rjsf/utils";
 import Button from "antd/lib/button";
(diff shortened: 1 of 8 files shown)


packages/antd/src/templates/ObjectFieldTemplate/index.js
```diff
@@ -62,12 +61,12 @@ const ObjectFieldTemplate = ({
         ? 24
         : 12;
 
-    if (_.isObject(colSpan)) {
+    if (colSpan instanceof Object) {
```
heath-freenome, line 64: I'm pretty sure this isn't a 1:1 change given that `isObject` is implemented as:
```
function isObject(value) {
  var type = typeof value;
  return value != null && (type == 'object' || type == 'function');
}
```

packages/validator-ajv6/src/createAjvInstance.ts
```diff
@@ -42,7 +41,7 @@ export default function createAjvInstance(
   }
 
   // add more custom formats to validate against
-  if (isObject(customFormats)) {
+  if (customFormats instanceof Object) {
```
heath-freenome, line 44: I'm pretty sure this isn't a 1:1 change given that `isObject` is implemented as:
```
function isObject(value) {
  var type = typeof value;
  return value != null && (type == 'object' || type == 'function');
}
```

packages/antd/src/templates/ObjectFieldTemplate/index.js
```diff
@@ -62,12 +61,12 @@ const ObjectFieldTemplate = ({
...
       return (
         colSpan[widget] || colSpan[field] || colSpan[type] || defaultColSpan
       );
     }
-    if (_.isNumber(colSpan)) {
+    if (typeof colSpan === "number") {
```
heath-freenome, line 69: I'm pretty sure this isn't a 1:1 change given that `isNumber` is implemented as:
```
function isNumber(value) {
  return typeof value == 'number' ||
    (isObjectLike(value) && baseGetTag(value) == numberTag);
}
```

packages/utils/src/mergeSchemas.ts
```diff
@@ -32,7 +30,7 @@ export default function mergeSchemas(
       Array.isArray(right)
     ) {
       // Don't include duplicate values when merging 'required' fields.
-      acc[key] = union(left, right);
```
heath-freenome, line 35: I'm pretty sure this isn't a 1:1 as the `lodash` `union()` handles more use cases

heath-freenome:
@jimmycallin How did you want to deal with the differences I pointed out above?

jimmycallin:
@heath-freenome thanks for the comments, I reverted those changes to reduce risk of missing use cases.

packages/utils/src/mergeSchemas.ts
```diff
@@ -1,5 +1,4 @@
 import union from "lodash/union";
-
```
heath-freenome, line 2: Revert this one line change so that then entire file is no longer in the PR

heath-freenome (approved):

PR #2998, opened 2022-08-06
add nx to monorepo

ahkhanjani:
Added Nx: Install `Nx Console` extension on VS Code to use the tools.
Added Nx Cloud distributed caching. See [here](https://nx.app/?utm_source=nx.dev).
Updated Lerna to v5.3.0.
Moved Prettier config to `.prettierrc.yaml`.


nx.json
```diff
@@ -0,0 +1,37 @@
...
+          "test:watch",
+          "precommit",
+          "publish-to-npm",
+          "test-coverage",
+          "build:lib",
+          "prepublishOnly",
+          "build:dist",
+          "publish-to-gh-pages"
+        ],
+        "accessToken": "ZmFmMjBmY2QtZTc1Mi00MGViLThkMGItODA5OTU4MDc5MWEyfHJlYWQtd3JpdGU="
```
heath-freenome, line 23: Is this a secret that needs to be passed in as a variable? Also, I'm not sure we want to use the cloud... and if we did, we really don't want this secret sitting in a public file (we have a place for them). Can we not use the cloud services by default?

ahkhanjani: The cloud caches the build and in later builds, only the parts that are affected are built. Since it's free for open-source I think we can make use of it.
Though you're correct. See [here](https://nx.dev/nx-cloud/account/access-tokens) for access tokens. To pass the access token as an environment variable for CI workflow, you can set NX_CLOUD_ACCESS_TOKEN in secrets. But it doesn't matter at the moment because only a few people have write access to the source code. But please consider creating an Nx Cloud account and passing the token so that we can make use of it.
If you're still not sure, you can remove the token and it will disable the cloud for the time being.
Best regards

heath-freenome: @ahkhanjani can you please remove the token for now. I've created an issue #3005 to make sure we get a proper access token setup for our builds!

heath-freenome:
@ahkhanjani Whoops, I've managed to create a minor conflict when I upgraded `husky` and ` lint-staged`, sorry.

ahkhanjani:
> @ahkhanjani Whoops, I've managed to create a minor conflict when I upgraded `husky` and ` lint-staged`, sorry.

No worries I've got you

nx.json
heath-freenome, line 23:
```suggestion
        ]
```

heath-freenome (approved):

heath-freenome:
@ahkhanjani Did you want me to do the merge of this PR for you?

ahkhanjani:
> @ahkhanjani Did you want me to do the merge of this PR for you?

That would be great!

PR #3010, opened 2022-08-11
fix: able to override className for submit button on core package

jacqueswho:
### Checklist

### DIFF
--- a/packages/core/src/components/widgets/SubmitButton.tsx
+++ b/packages/core/src/components/widgets/SubmitButton.tsx
@@ -18,7 +18,7 @@ export default function SubmitButton<T, F>({
       <button
         type="submit"
         {...submitButtonProps}
-        className={`btn btn-info ${submitButtonProps.className}`}
+        className={submitButtonProps.className || `btn btn-info`}
       >
         {submitText}
       </button>


packages/core/src/components/widgets/SubmitButton.tsx
```diff
@@ -18,7 +18,7 @@ export default function SubmitButton<T, F>({
       <button
         type="submit"
         {...submitButtonProps}
-        className={`btn btn-info ${submitButtonProps.className}`}
+        className={submitButtonProps.className || `btn btn-info`}
```
heath-freenome, line 21: @jacqueswho Your change will replace the classNames entirely. Was that the expected behavior, rather than adding additional formatting to the className?

jacqueswho: Yes, so that is the default for core but if a user puts his own, there will be no clash or confusion. What do you think? They can still include then the default with their own

heath-freenome: Most of the other libraries that I've used (like Material UI) the classNames are added in addition to the default classes required to properly render the defaults. I'm worried that without the `btn btn-info` classes the resulting button would look very different. Most people will probably do things like move the button, change its color stuff like that.

jacqueswho: Ok will close pull request

PR #3018, opened 2022-08-13
Add getTemplate() function to utils

heath-freenome:

Fixes an issue I identified while working on updating the documentation for templates where not all templates were able to be overridden by the `uiSchema`

- Added a new `getTemplate()` function used to fetch a template from either the `uiSchema.options` or the `registry`
  - Exported it from the library and added 100% unit tests
  - Also added an optional `uiSchema` to the `ArrayFieldDescriptionTemplate`
- Updated all uses of getting a template from `registry.templates` to use the `getTemplate()` function instead with the following additions:
  - Updated all `ArrayFieldTemplate`s to use `getTemplate()` for the three templates it uses as well as passing `uiSchema` to the `ArrayFieldDescriptionTemplate`
  - Updated all `ObjectFieldTemplate`s to use `getTemplate()` for the two templates is uses
  - Updated all `Widget` implementations that to use `getTemplate()` for getting the `BaseInputTemplate` (or on occasion the `DescriptionFieldTemplate` or `TitleFieldTemplate`)
  - Updated the `ArrayFieldDescriptionTemplate` and `ArrayFieldTitleTemplate` in `core` to use `getTemplate()` to get `DescriptionFieldTemplate` and `TitleFieldTemplate` respectively
  - Updated the `ArrayField`, `ObjectField` and `SchemaField` in `core` to use `getTemplate()` to fetch all necessary templates
    - Removed the need for the `Defaultxxx` template imports as the registry is guaranteed to have them now
  - Updated the `Form` in `core` to get the `ErrorListTemplate` using `getTemplate()`
  - Updated the `FieldTemplate` in `semantic-ui` to use `getTemplate()` to get its one necessary template

### Checklist

### DIFF
--- a/packages/bootstrap-4/src/ArrayFieldTemplate/ArrayFieldTemplate.tsx
+++ b/packages/bootstrap-4/src/ArrayFieldTemplate/ArrayFieldTemplate.tsx
@@ -5,6 +5,7 @@ import Container from "react-bootstrap/Container";
 import {
   ArrayFieldTemplateItemType,
   ArrayFieldTemplateProps,
+  getTemplate,
   getUiOptions,
 } from "@rjsf/utils";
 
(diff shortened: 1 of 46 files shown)


packages/bootstrap-4/src/ArrayFieldTemplate/ArrayFieldTemplate.tsx
```diff
@@ -24,12 +25,23 @@ const ArrayFieldTemplate = (props: ArrayFieldTemplateProps) => {
...
     title,
   } = props;
-  const {
-    ArrayFieldDescriptionTemplate,
-    ArrayFieldItemTemplate,
-    ArrayFieldTitleTemplate,
-  } = registry.templates;
   const uiOptions = getUiOptions(uiSchema);
+  const ArrayFieldDescriptionTemplate =
+    getTemplate<"ArrayFieldDescriptionTemplate">(
```
nickgros, line 30: Do we need the generic argument in these cases? Seems like in some cases it's inferred from the first arg, in other places like this it's explicit.

nickgros (approved):

packages/bootstrap-4/src/ArrayFieldTemplate/ArrayFieldTemplate.tsx
heath-freenome: The places where it is inferred are JS files which do not support generics. The <Name> generic causes the appropriately casted return value to be exactly the right function. Super helpful... now if only I could have figured out how to use the generic as a variable to access the array

Issue #2616, opened 2021-11-19
[Theme][Bootstrap4] Readonly attribute is not valid on select element


Lausselloic:
Hello,

Looks like there's a mistake on bootstrap4 select Widget that have been fixed in core. Readonly attribute have no effect on select element. So maybe like in core https://github.com/rjsf-team/react-jsonschema-form/blob/79a61ff70c25f84ec3e989899ac7f92d1dd3e443/packages/core/src/components/widgets/SelectWidget.js#L74 use disabled on select element if readonly or disabled value is set to true?

https://github.com/rjsf-team/react-jsonschema-form/blob/79a61ff70c25f84ec3e989899ac7f92d1dd3e443/packages/bootstrap-4/src/SelectWidget/SelectWidget.tsx#L91

Lausselloic:
Issue also present on 
semanticUI : https://github.com/rjsf-team/react-jsonschema-form/blob/79a61ff70c25f84ec3e989899ac7f92d1dd3e443/packages/semantic-ui/src/SelectWidget/SelectWidget.js#L113

epicfaace:
Interesting, can you make a PR to fix it?

--
Ashwin Ramaswami

heath-freenome:
Fixed in v5 beta via https://github.com/rjsf-team/react-jsonschema-form/pull/3036

PR #3036, opened 2022-08-23
Fix hidden widget in several themes and SelectWidget in bootstrap-4

heath-freenome:

Fixes (#2571) and (#2616)

- Updated the `FieldTemplate` in `chakra-ui`, `material-ui`, `mui` and `semantic-ui` to properly implement the hidden field
- Updated `SelectWidget` in `bootstrap-4` to fix missing `htmlFor` and the `disabled` state when `readonly`
- Updated snapshots to verify fixes
- Also removed the `test:update` script from `core`, `utils` and `validator-ajv6` because they are useless since those three packages do NOT have snapshots

### Checklist

### DIFF
--- a/packages/bootstrap-4/src/SelectWidget/SelectWidget.tsx
+++ b/packages/bootstrap-4/src/SelectWidget/SelectWidget.tsx
@@ -55,7 +55,7 @@ const SelectWidget = ({
         value={typeof value === "undefined" ? emptyValue : value}
         required={required}
         multiple={multiple}
-        disabled={disabled && readonly}
+        disabled={disabled || readonly}
         autoFocus={autofocus}
         className={rawErrors.length > 0 ? "is-invalid" : ""}
         onBlur={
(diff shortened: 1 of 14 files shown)


packages/bootstrap-4/src/SelectWidget/SelectWidget.tsx
```diff
@@ -52,8 +55,7 @@ const SelectWidget = ({
         value={typeof value === "undefined" ? emptyValue : value}
         required={required}
         multiple={multiple}
-        disabled={disabled}
-        readOnly={readonly}
+        disabled={disabled && readonly}
```
nickgros, line 58: Why were these consolidated?

nickgros: If they have the same effect, then I think it shouldn't it be `||`?

epicfaace: Yeah, I think they are actually different? Not 100% sure about how this widget works on bootstrap

heath-freenome: They were consolidated to fix and issue because readOnly is not supported on the bootstrap select widget... and yeah it should be an ||

jacqueswho (approved):

PR #3041, opened 2022-08-24
Restored uiSchema.classNames with deprecation warning

heath-freenome:

- Updated `SchemaField` to restore support for `uiSchema.classNames` with big fat deprecation warning
- Updated the existing `classNames` test to verify restored code and deprecation warning
- NOTE: Will fix up the documentation in the other PR

### Checklist

### DIFF
--- a/packages/core/src/components/fields/SchemaField.tsx
+++ b/packages/core/src/components/fields/SchemaField.tsx
@@ -258,7 +258,7 @@ function SchemaFieldRender<T, F>(props: FieldProps<T, F>) {
   if (uiSchema?.classNames) {
     if (process.env.NODE_ENV !== "production") {
       console.warn(
-        "WARNING: 'uiSchema.classNames' is deprecated and will be removed in the next major release; Use 'ui:classNames' instead."
+        "WARNING: 'uiSchema.classNames' is deprecated and may be removed in a major release; Use 'ui:classNames' instead."
       );
     }
     classNames.push(uiSchema.classNames);
--- a/packages/core/src/components/fields/SchemaField.tsx
+++ b/packages/core/src/components/fields/SchemaField.tsx
@@ -209,7 +209,7 @@ function SchemaFieldRender<T, F>(props: FieldProps<T, F>) {
 
   const { __errors, ...fieldErrorSchema } = errorSchema || {};
   // See #439: uiSchema: Don't pass consumed class names to child components
-  const fieldUiSchema = omit(uiSchema, ["ui:classNames"]);
+  const fieldUiSchema = omit(uiSchema, ["ui:classNames", "classNames"]);
   if ("ui:options" in fieldUiSchema) {
     fieldUiSchema["ui:options"] = omit(fieldUiSchema["ui:options"], [
       "classNames",
@@ -254,6 +254,12 @@ function SchemaFieldRender<T, F>(props: FieldProps<T, F>) {
   const classNames = ["form-group", "field", `field-${schema.type}`];
   if (!hideError && errors && errors.length > 0) {
     classNames.push("field-error has-error has-danger");
+  }
+  if (uiSchema?.classNames) {
+    console.warn(
+      "WARNING: 'uiSchema.classNames' is deprecated and will be removed in the next major release; Use 'ui:classNames' instead."
+    );
+    classNames.push(uiSchema.classNames);
   }
   if (uiOptions.classNames) {
     classNames.push(uiOptions.classNames);
--- a/packages/core/src/components/fields/SchemaField.tsx
+++ b/packages/core/src/components/fields/SchemaField.tsx
@@ -256,9 +256,11 @@ function SchemaFieldRender<T, F>(props: FieldProps<T, F>) {
     classNames.push("field-error has-error has-danger");
   }
   if (uiSchema?.classNames) {
-    console.warn(
-      "WARNING: 'uiSchema.classNames' is deprecated and will be removed in the next major release; Use 'ui:classNames' instead."
-    );
+    if (process.env.NODE_ENV !== "production") {
+      console.warn(
+        "WARNING: 'uiSchema.classNames' is deprecated and will be removed in the next major release; Use 'ui:classNames' instead."
+      );
+    }
     classNames.push(uiSchema.classNames);
   }
   if (uiOptions.classNames) {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/fields/SchemaField.tsx
```diff
@@ -249,6 +249,12 @@ function SchemaFieldRender<T, F>(props: FieldProps<T, F>) {
   if (!hideError && errors && errors.length > 0) {
     classNames.push("field-error has-error has-danger");
   }
+  if (uiSchema?.classNames) {
+    console.warn(
+      "WARNING: 'uiSchema.classNames' is deprecated and will be removed in the next major release; Use 'ui:classNames' instead."
```
epicfaace, line 254: maybe we can discuss on Friday, but I'm not quite sure if we should ever remove this at all. uiSchema is rjsf's own creation, it's not like we're trying to conform to some spec like JSON Schema.

I don't see the point in ever removing classNames -- we can of course change all examples to use ui:classNames, but I see no harm in continuing to support it indefinitely.

packages/core/src/components/fields/SchemaField.tsx
```diff
@@ -249,6 +249,12 @@ function SchemaFieldRender<T, F>(props: FieldProps<T, F>) {
   if (!hideError && errors && errors.length > 0) {
     classNames.push("field-error has-error has-danger");
   }
+  if (uiSchema?.classNames) {
+    console.warn(
```
epicfaace, line 253: I think `console.warn` might be called hundreds of times , due to each re-render? This has caused issues with deprecation warnings in the past as they slow down things immensely. We might also want to only call this warning if `process.env.NODE_ENV === 'development'`

packages/core/src/components/fields/SchemaField.tsx
heath-freenome: @epicfaace I was adding a consistency to the `uiSchema` for options. `classNames` was literally the only `uiSchema` option that did not come with the `ui:` prefix so I was making it consistent with all the others.

packages/core/src/components/fields/SchemaField.tsx
heath-freenome: I'll add the `process.env.NODE_ENV`... as for slowing things down, well there is a clear fix to making them go away ;)

packages/core/src/components/fields/SchemaField.tsx
epicfaace: Right, consistency is good, and is a good reason to make "ui:classNames" the "official" way to do it. But I don't think consistency dictates that we also have to remove support for "classNames".

heath-freenome: I understand... I also wonder if you didn't have your uiSchema files in a DB making it harder to just globally change them, would you feel so strongly about it keeping this inconsistency?

heath-freenome: Got it... Well it is deprecated for the moment, with a warning it may be removed in a future release

jacqueswho (commented):
Is there an update to the playground example, if this is used?

heath-freenome:
> Is there an update to the playground example, if this is used?

There are no examples in the `playground` that uses it

packages/core/src/components/fields/SchemaField.tsx
```diff
@@ -255,6 +255,14 @@ function SchemaFieldRender<T, F>(props: FieldProps<T, F>) {
   if (!hideError && errors && errors.length > 0) {
     classNames.push("field-error has-error has-danger");
   }
+  if (uiSchema?.classNames) {
+    if (process.env.NODE_ENV !== "production") {
+      console.warn(
+        "WARNING: 'uiSchema.classNames' is deprecated and may be removed in a major release; Use 'ui:classNames' instead."
```
heath-freenome, line 261:
```suggestion
        "'uiSchema.classNames' is deprecated and may be removed in a major release; Use 'ui:classNames' instead."
```

heath-freenome: Will commit this before merging if no other feedback is provided, otherwise will update on another commit

packages/core/test/uiSchema_test.js
```diff
@@ -43,15 +46,26 @@ describe("uiSchema", () => {
...
-      const [foo, bar] = node.querySelectorAll(".field-string");
+      const [foo, bar, baz] = node.querySelectorAll(".field-string");
 
       expect(foo.classList.contains("class-for-foo")).eql(true);
       expect(bar.classList.contains("class-for-bar")).eql(true);
       expect(bar.classList.contains("another-for-bar")).eql(true);
+      expect(baz.classList.contains("class-for-baz")).eql(true);
+      expect(
+        console.warn.calledWithMatch(
+          /WARNING: 'uiSchema.classNames' is deprecated/
```
heath-freenome, line 66:
```suggestion
          /'uiSchema.classNames' is deprecated/
```

heath-freenome: Will commit this before merging if no other feedback is provided, otherwise will update on another commit

packages/core/test/uiSchema_test.js
```diff
@@ -43,15 +46,26 @@ describe("uiSchema", () => {
...
+      const [foo, bar, baz] = node.querySelectorAll(".field-string");
 
       expect(foo.classList.contains("class-for-foo")).eql(true);
       expect(bar.classList.contains("class-for-bar")).eql(true);
       expect(bar.classList.contains("another-for-bar")).eql(true);
+      expect(baz.classList.contains("class-for-baz")).eql(true);
+      expect(
+        console.warn.calledWithMatch(
+          /'uiSchema.classNames' is deprecated/
+        )
```
heath-freenome, line 67:
```suggestion
        console.warn.calledWithMatch(/'uiSchema.classNames' is deprecated/);
```

packages/core/test/uiSchema_test.js
```diff
@@ -43,15 +46,24 @@ describe("uiSchema", () => {
...
       const { node } = createFormComponent({ schema, uiSchema });
-      const [foo, bar] = node.querySelectorAll(".field-string");
+      const [foo, bar, baz] = node.querySelectorAll(".field-string");
 
       expect(foo.classList.contains("class-for-foo")).eql(true);
       expect(bar.classList.contains("class-for-bar")).eql(true);
       expect(bar.classList.contains("another-for-bar")).eql(true);
+      expect(baz.classList.contains("class-for-baz")).eql(true);
+      expect(
+        console.warn.calledWithMatch(/'uiSchema.classNames' is deprecated/);
```
heath-freenome, line 65:
```suggestion
        console.warn.calledWithMatch(/'uiSchema.classNames' is deprecated/)
```

heath-freenome:
@nickgros @epicfaace @jacqueswho I believe this PR is ready for approva

nickgros (approved):

jacqueswho (approved):

PR #3048, opened 2022-08-26
Feature: Add new prop duplicateKeySuffixSeparator for additionalProperties

nickgros:

This was originally done in #2467, porting to v5 here.

When using `additionalProperties` and a duplicate property key is added, RJSF automatically changes the duplicate key from e.g. `collision` to `collision-1`.

I have a use case where we use `additionalProperties`, but `-` is an illegal character for our property keys. I added a FormProp `duplicateKeySuffixSeparator` to override `-` with any string.

Couple of questions
- Is there a better/less verbose prop name?
- This could be added to the `uiSchema` to change the separator per object field. I do not need this for my use case, but it's a potential change that could be made.

### Checklist

### DIFF
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -62,10 +62,6 @@ export interface FormProps<T = any, F = any> {
    * to each field of the form. If you just want to disable some fields, see the `ui:disabled` parameter in `uiSchema`
    */
   disabled?: boolean;
-  /** When using `additionalProperties`, key collision is prevented by appending a unique integer to the duplicate key.
-   * This prop allows you to change the separator between the original key name and the integer. Default is "-"
-   */
-  duplicateKeySuffixSeparator?: string;
   /** It's possible to make the whole form read-only by setting the `readonly` prop. The `readonly` prop is then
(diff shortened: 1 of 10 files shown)


packages/core/src/components/Form.tsx
```diff
@@ -62,6 +62,10 @@ export interface FormProps<T = any, F = any> {
    * to each field of the form. If you just want to disable some fields, see the `ui:disabled` parameter in `uiSchema`
    */
   disabled?: boolean;
+  /** When using `additionalProperties`, key collision is prevented by appending a unique integer to the duplicate key.
+   * This prop allows you to change the separator between the original key name and the integer. Default is "-"
+   */
+  duplicateKeySuffixSeparator?: string;
```
heath-freenome, line 68: Could we use the `idSeparator` instead? Less to pass around and has a similar use case

heath-freenome: Oh wait, update the `CHANGELOG.md` too...

nickgros:
Review: will move to uiSchema, keeping separate from idSeparator because idSeparator implies hierarchy and duplicateKeySuffixSeparator does not).

heath-freenome (approved):

heath-freenome:
@nickgros go ahead and merge this!

PR #3067, opened 2022-08-29
deps: update semantic-ui-react dependency

jacqueswho:

- update semantic-ui-react dependency to 2.1.3


packages/semantic-ui/package.json
```diff
@@ -27,8 +27,8 @@
     ]
   },
   "peerDependencies": {
-    "@rjsf/core": "^5.0.0-beta.1",
-    "@rjsf/utils": "^5.0.0-beta.1",
+    "@rjsf/core": "^5.0.0-beta.2",
+    "@rjsf/utils": "^5.0.0-beta.2",
```
heath-freenome, line 31: Jacques, I don't believe you need to change these since they have the `^` on them.

heath-freenome (approved):
Revert the peerDependency changes maybe?

PR #3127, opened 2022-09-19
fix: radio widget for boolean not working for multiple as intended

johannesjo:
When there are multiple bootstrap radio widgets with the same label (e.g. for boolean fields with "Yes" and "No" clicking any label will select the corresponding input for the first input on the page rather than the clicked one.

### DIFF
--- a/packages/bootstrap-4/src/RadioWidget/RadioWidget.tsx
+++ b/packages/bootstrap-4/src/RadioWidget/RadioWidget.tsx
@@ -49,7 +49,7 @@ const RadioWidget = ({
           <Form.Check
             inline={inline}
             label={option.label}
-            id={option.label}
+            id={id + "_" + option.label}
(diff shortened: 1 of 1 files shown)


heath-freenome:
@johannesjo Can you update the `CHANGELOG.md` and the test snapshots

johannesjo:
@heath-freenome could you maybe tell me which tests I need to adjust? I'm having problems to understand the test setup tbh.

heath-freenome:
@johannesjo after doing an `npm install` and npm run build` from the root directory, try doing `npm run test` from the bootstrap directory. That will fail. Doing `npm run test:update` will update the snapshots. After that, also update `CHANGELOG.md` to reference your fix

heath-freenome:
@johannesjo I'm going to close this as I've incorporated your change into #3145

PR #3150, opened 2022-09-26
chore: convert semantic-ui to typescript

heath-freenome:

- Updated `.eslintrc` to use the typescript version
- Update `package.json` to switch from `js` to `ts?(x)`, adding necessary type imports and `typings` block
- Refactored all of the `.js` file to Typescript
  - Reworked a few components to deal with unavailable props in types
- Converted the tests, renaming the snapshots accordingly, updating those templates that differed by just some whitespace rendering

### Checklist


packages/semantic-ui/src/ArrayFieldItemTemplate/ArrayFieldItemTemplate.tsx
```diff
@@ -1,45 +1,45 @@
...
 
 import { MaybeWrap } from "../util";
 
-const gridStyle = (vertical) => ({
+const gridStyle = (vertical: boolean) => ({
   display: "grid",
   gridTemplateColumns: `1fr ${vertical ? 65 : 110}px`,
 });
 
-// checks if it's the first array item
```
heath-freenome, line 11: @jacqueswho do you remember why you used this odd logic to determine the first array item rather than simply using the `index` like I did below? This always seems to return true from what the tests indicate

jacqueswho (approved):
looks good

packages/semantic-ui/src/ArrayFieldItemTemplate/ArrayFieldItemTemplate.tsx
jacqueswho: I don't remember doing this code, could have been a contribution

PR #3292, opened 2022-12-12
isObject should return false for Date as well as File, null, and Array

BenBeattieHood:

Without this, assigning dates to the form's value instead merges them with a default value for objects of `{}` (in mergeDefaultsWithFormData), which stops them being dates.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Updated `computedDefaults` (used by `getDefaultFormState`) to skip saving the computed default if it's an empty object unless `includeUndefinedValues` is truthy, fixing [#2150](https://github.com/rjsf-team/react-jsonschema-form/issues/2150) and [#2708](https://github.com/rjsf-team/react-jsonschema-form/issues/2708)
 - Expanded the `getDefaultFormState` util's `includeUndefinedValues` prop to accept a boolean or `"excludeObjectChildren"` if it does not want to include undefined values in nested objects
 - Updated `mergeObjects` to add new `preventDuplicates` mode when concatenating arrays so that only unique values from the source object array are copied to the destination object array
-- Fix isObject to correctly identify 'Date' as not mergable with Object default value.
+- Fix `isObject` to correctly identify 'Date' as not an object, similar to 'File', thus preventing them from being merged with Object default values.
 
 ## Dev / docs / playground
 - Removed extraneous leading space on the examples in the validation documentation, fixing [#3282](https://github.com/rjsf-team/react-jsonschema-form/issues/3282)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,6 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/core
 - Added `ref` definition to `ThemeProps` fixing [#2135](https://github.com/rjsf-team/react-jsonschema-form/issues/2135)
 - Updated the `onChange` handler in `Form` to use the new `preventDuplicates` mode of `mergeObjects()` when merging `extraErrors` when live validation is off, fixing [#3169](https://github.com/rjsf-team/react-jsonschema-form/issues/3169)
-- Fix isObject to correctly identify 'Date' as not mergable with Object default value.
 
 ## @rjsf/material-ui
 - Fix RangeWidget missing htmlFor and schema.title [#3281](https://github.com/rjsf-team/react-jsonschema-form/pull/3281)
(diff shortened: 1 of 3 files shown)


packages/utils/src/isObject.ts
```diff
@@ -8,5 +8,8 @@ export default function isObject(thing: any) {
   if (typeof File !== "undefined" && thing instanceof File) {
     return false;
   }
+  if (typeof Date !== "undefined" && thing instanceof Date) {
+    return false;
+  }
```
heath-freenome, line 13: You will need to add a test for this use case

heath-freenome:
Also, update the `CHANGELOG.md` file please

packages/utils/src/isObject.ts
BenBeattieHood: I'm sorry to ask, but would you be able to give me a pointer as to how? I'm not familiar with the test configuration used on this project. 🙏

heath-freenome: @BenBeattieHood so it looks like we never added unit tests for `isObject()` whoops. Go ahead an add a new file `packages/utils/test/isObject.ts` with the following code:

```
import { isObject } from '../src';

const NON_OBJECTS = [
  'string',
  10,
  NaN,
  true,
  null,
  undefined,
];

const OBJECTS = [
  { plain: 'object' },
  new Object(),
  new Map(),
];

describe('isObject()', () => {
  it('returns false when a non-object is provided', () => {
    NON_OBJECTS.forEach((nonObject: string | number | boolean | null | undefined) => {
      expect(isObject(nonObject)).toBe(false);
    });
  });
  it('returns false when a File is provided', () => {
    const file = new File(["test"], "test.txt");
    expect(isObject(file)).toBe(false);
  });
  it('returns false when an array is provided', () => {
    expect(isObject(['foo'])).toBe(false);
  });
  it('returns true when an object is provided', () => {
    OBJECTS.forEach((object: any) => {
      expect(isObject(object)).toBe(true);
    });
  });
});

```

This will test all of the current code. I leave testing for your new code to you, should be fairly straight-forward

BenBeattieHood:
Thanks for your guidance @heath-freenome :+1: I've updated the tests and the changelog as required.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/core
 - Added `ref` definition to `ThemeProps` fixing [#2135](https://github.com/rjsf-team/react-jsonschema-form/issues/2135)
 - Updated the `onChange` handler in `Form` to use the new `preventDuplicates` mode of `mergeObjects()` when merging `extraErrors` when live validation is off, fixing [#3169](https://github.com/rjsf-team/react-jsonschema-form/issues/3169)
+- Fix isObject to correctly identify 'Date' as not mergable with Object default value.
```
heath-freenome, line 37: This actually belongs in the `@rjsf/utils` section below. I'll approve once it is moved.

BenBeattieHood: 🤦‍♂️sorry - I should've seen that.

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Updated `computedDefaults` (used by `getDefaultFormState`) to skip saving the computed default if it's an empty object unless `includeUndefinedValues` is truthy, fixing [#2150](https://github.com/rjsf-team/react-jsonschema-form/issues/2150) and [#2708](https://github.com/rjsf-team/react-jsonschema-form/issues/2708)
 - Expanded the `getDefaultFormState` util's `includeUndefinedValues` prop to accept a boolean or `"excludeObjectChildren"` if it does not want to include undefined values in nested objects
 - Updated `mergeObjects` to add new `preventDuplicates` mode when concatenating arrays so that only unique values from the source object array are copied to the destination object array
+- Fix isObject to correctly identify 'Date' as not mergable with Object default value.
```
heath-freenome, line 48:
```suggestion
- Fix `isObject` to correctly identify 'Date' as not an object, similar to 'File', thus preventing them from being merged with Object default values.
```

heath-freenome (approved):

heath-freenome:
@BenBeattieHood Also, I'm guessing that you didn't have the git precommit install hook properly installed (see the [docs](https://react-jsonschema-form.readthedocs.io/en/latest/contributing/#first-time-step) on how to do that). Therefore, you will need to run `npm run cs-format` in the `packages/utils` directory and push the resulting formatting changes to allow the build to succeed.

BenBeattieHood:
Aah! I see that now. Thanks for the guidance & docs on this. Ok, have applied the lint formatting and rebased for a clean diff.

PR #3311, opened 2022-12-23
Keep old umd file path

epicfaace:
Keep old umd file path so old umd links don't break.


epicfaace:
Actually, maybe we shouldn't do this. This will basically break all legacy sites that link to the latest version of rjsf. I'm fine with just not keeping this anymore. cc @nickgros / @heath-freenome

PR #2634, opened 2021-12-14
explicit ajv stack error

Lausselloic:
### Reasons for making this change

Improve the error format from AJV. Currently it return a dotted path from schema like `.foo.bar.mycustomProp is a required property` into the stack error. What I ropose is to fetch "display" name of the property if it exists `title` or the property himself otherwise. It's more human readable and it's whats is needed for a user interface I think.

Related to https://github.com/rjsf-team/react-jsonschema-form/issues/1337 need to enbale verbose mode of AJV

User could still create the old message by using Property and message returned into the error field

If this is related to existing tickets, include links to them as well. Use the syntax `fixes #[issue number]` (ex: `fixes #123`).

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/master/CHANGELOG.md) with a description of the PR
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #3337, opened 2023-01-03
Fix: Better validation messages using schema/uiSchema title if it exists

### LINKED ISSUE #3246: Is it possible to display field titles in validation error messages instead of real field names?
### Prerequisites

### What theme are you using?

core

### What is your question?

Is it possible to display field titles in validation error messages instead of real property names? Field titles may be clearer to users.
If I have to replace the property name in `transformErrors`, then how to get the field title from the validation error object?

image

dagda1:

Fixes #3246 by reimplementing #2634

- The ajv8 `ErrorObject` message is enhanced by replacing the error message field with either the `uiSchema`'s `ui:title` field if one exists or the `parentSchema` title if one exists.

### Checklist


heath-freenome (commented):
Whoops, these have been sitting pending for days as I forgot to click submit, sorry

packages/validator-ajv8/src/validator.ts
```diff
@@ -19,10 +19,12 @@ import {
   StrictRJSFSchema,
   ValidationData,
   ValidatorType,
+  PROPERTIES_KEY,
 } from "@rjsf/utils";
 
 import { CustomValidatorOptionsType, Localizer } from "./types";
 import createAjvInstance from "./createAjvInstance";
+import { get } from "lodash";
```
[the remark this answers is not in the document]
dagda1: oops, that would have been catastrophic for bundle size.  That was a VSCode added reference

packages/validator-ajv8/src/validator.ts
```diff
@@ -199,16 +201,35 @@ export default class AJV8Validator<
...
+          )
+          .join("");
+
+        const title = get(schema, `${PROPERTIES_KEY}.${titlePath}`);
+
+        if (title) {
+          const existing = property.split(".").slice(-1)[0];
+          message = message?.replace(existing, title);
+        }
+
```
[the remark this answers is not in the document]
dagda1: @heath-freenome having the `parentSchema` in the error made things more straightforward.  I have used the title in all validation messages.

I take your point about nested error messages all looking the same.

What do you think a suitable error message would look like if we had a `schemaPath` like `.nested.lastName` and a title like 'Last Name'?

packages/validator-ajv8/test/validator.test.ts
```diff
@@ -1612,21 +1706,27 @@ describe("AJV8Validator", () => {
...
+              type: "string",
+            },
+            schema: "\\d+",
             schemaPath: "#/definitions/Dataset/properties/datasetId/pattern",
           },
         ]);
       });
     });
-    describe("validating using several custom meta schemas", () => {
+    describe("validating using several custom met©©a schemas", () => {
```
heath-freenome, line 1729: Revert this?
```suggestion
    describe("validating using several custom meta schemas", () => {
```

packages/validator-ajv8/src/validator.ts
```diff
@@ -202,14 +203,45 @@ export default class AJV8Validator<
...
+        const parentSchemaTitle =
+          parentSchema?.[PROPERTIES_KEY]?.[currentProperty]?.title;
+
+        if (parentSchemaTitle) {
+          message = message?.replace(currentProperty, parentSchemaTitle);
+        }
+
         stack = message!;
+      } else {
+        let title = property.split(".").pop() as string;
```
heath-freenome, line 236: I find it interesting that you needed to do a cast here since property is a string, split should create an array of strings, so popping should still be a string.

packages/validator-ajv8/src/validator.ts
```diff
@@ -202,14 +203,45 @@ export default class AJV8Validator<
...
-      let stack = `${property} ${message}`.trim();
+
+      let stack: string | undefined = undefined;
+
       if ("missingProperty" in params) {
         property = property
           ? `${property}.${params.missingProperty}`
           : params.missingProperty;
+
+        const currentProperty = property.split(".").slice(-1)[0];
```
heath-freenome, line 225: Couldn't you just use `params.missingProperty` instead of all this jiggling since it will always be the last one?

packages/validator-ajv8/src/validator.ts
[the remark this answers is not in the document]
heath-freenome: Ideally the UI would want to expand the path with all of the names. And there is also the consideration of name overrides in the `uiSchema`.

packages/validator-ajv8/src/validator.ts
dagda1: @heath-freenome if you look at the types for `pop`, it returns `T | undefined`

```ts
    /**
     * Removes the last element from an array and returns it.
     * If the array is empty, undefined is returned and the array is not modified.
     */
    pop(): T | undefined;
```

I would not call it a `cast` as nothing happens at runtime, this is purely at the type level, and I think a [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) is the right call here.

heath-freenome: Ah, thanks for clarifying!

packages/validator-ajv8/src/validator.ts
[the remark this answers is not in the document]
dagda1: > Ideally the UI would want to expand the path with all of the names. And there is also the consideration of name overrides in the `uiSchema`.

I honestly did not know this was possible, do you have an example of what a uiSchema might look like and how l would get a hold of the `uiSchema` in `transformRJSFValidationErrors`?

If you have time, could you write a failing test?  Thanks for taking the time to review the PR BTW.  I'm working on upgrading backstage to the latest version of `@rjsf` and this PR would be great.

packages/validator-ajv8/src/validator.ts
dagda1: @heath-freenome in the case of a nested property, property would be like this `.nested.lastName` so I do think I need to get the last item. 

I could possibly use a regex if you think that is better?

dagda1: Great shout, that is much better.

packages/validator-ajv8/src/validator.ts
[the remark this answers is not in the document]
heath-freenome: Here's the [uiSchema](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/#title) documentation. Right now there isn't a way to get the `uiSchema` so it would require updating the `ValidatorType` interface in `@rjsf/utils` to add an additional (optional) property on `validateFormData()`. Which means that the `F extends FormContextType = any` generic would have to be added to the type as well. This ripples through to the `Form` in `@rjsf/core` since you would now need to update typing and pass `uiSchema` in as well.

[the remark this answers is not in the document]
dagda1: Is this something that could wait for a future PR, or is it non-negotiable?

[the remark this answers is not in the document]
heath-freenome: @dagda1 So, I decided to do the work of passing in `uiSchema` to `validateFormData()` in #3358. Once it merges you can simply use it.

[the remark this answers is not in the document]
dagda1: @heath-freenome could we complete this PR and do the `uiSchema`  on another PR to keep the size of the PR down?

Totally your shout and if you feel they are dependent then I will add it here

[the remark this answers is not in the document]
heath-freenome: @dagda1 Do you feel like you can get the `uiSchema` piece done a few days after we merge this one? If so, I'd be ok with merging this before we make another release. Otherwise, if someone has a `uiSchema` override of the title and the error message shows something different, it would be a bug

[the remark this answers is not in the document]
heath-freenome: There is a simple example in the docs [here](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/#title) and you can see it in action in the [playground](https://rjsf-team.github.io/react-jsonschema-form/) by clicking on the `UI Options` example.

Screen Shot 2023-01-11 at 1 33 58 PM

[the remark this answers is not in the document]
dagda1: @heath-freenome thanks, I'll just add it to this PR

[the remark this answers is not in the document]
dagda1: @heath-freenome I have had a go at the `uiSchema` change

heath-freenome:
@dagda1 Would you please revert the `package-lock.json` changes

packages/validator-ajv8/src/validator.ts
```diff
@@ -202,17 +204,66 @@ export default class AJV8Validator<
...
         property = property
           ? `${property}.${params.missingProperty}`
           : params.missingProperty;
+
+        const currentProperty: string = params.missingProperty;
+
+        const uiSchemaTitle = get(
+          uiSchema,
+          `${property.replace(/^\./, "")}.['ui:title']`
+        );
```
heath-freenome, line 234: Given that `title` could also be in the `ui:options` object, Do this instead:
```suggestion
        const uiSchemaTitle = getUiOptions<T, S, F>(get(uiSchema,`${property.replace(/^\./, "")}`)).title;
```

packages/validator-ajv8/src/validator.ts
```diff
@@ -202,17 +204,66 @@ export default class AJV8Validator<
...
+        }
+
         stack = message!;
+      } else {
+        let title = property.split(".").pop() as string;
+
+        const uiSchemaTitle = get(
+          uiSchema,
+          `${property.replace(/^\./, "")}.['ui:title']`
+        );
```
heath-freenome, line 254:
```suggestion
        const uiSchemaTitle = getUiOptions<T, S, F>(get(uiSchema,`${property.replace(/^\./, "")}`)).title;
```

heath-freenome: Also, since you are doing this in both branches of the `if` statement, consider moving the variable assignment above and out of the `if`

heath-freenome (changes requested):

heath-freenome:
Finally, once this is merged, we will release 5.0.0

packages/validator-ajv8/src/validator.ts
dagda1: @heath-freenome the property paths are different in each case so the best way to generalise it would be to pull it into a function, I'm not sure that makes it better but I would do that if you agree.

Also, you don't need to pass the `<T, S, F>` type arguments when calling `getUiOptions`, typescript will infer those.

heath-freenome: Ah, yes, you are right in that property is updated in the first `if` block and isn't in the second.

heath-freenome:
> > @dagda1 Would you please revert the `package-lock.json` changes
> 
> oops, I think I might have a different version of npm than you. What version are you using.
> 
> I have reverted the files.

I'm using node 18 although node 16 also uses the same formats. I'm pushing a PR #3372 to bump the `package-lock.json` files using clean `node_modules` directories which is how new developers like you start out

packages/validator-ajv8/src/validator.ts
```diff
@@ -202,17 +205,64 @@ export default class AJV8Validator<
...
+        instancePath,
+        keyword,
+        params,
+        schemaPath,
+        parentSchema,
+        ...rest
+      } = e;
+      let { message } = rest;
       let property = instancePath.replace(/\//g, ".");
-      let stack = `${property} ${message}`.trim();
```
heath-freenome, line 210: Given how many tests in `core` are failing, it may make sense to either keep this and only replace when title is provided in the `schema` or `uiSchema` OR you'll have to update the core tests to remove the leading `.` that is currently there. I'd lean towards the former since the latter IS a breaking change in behavior vs what the `validator-ajv6` is doing. At the end of the day it might be better to get the opinion of the owner of RJSF, @epicfaace

dagda1: I totally agree.  It is great that there is a full test suite to work against.   I have reverted the default stack.

dagda1:
@heath-freenome I'm getting yellow lines in the coverage report because of the optional chaining around both message and parentSchema.

e.g. 

```ts
message = message?.replace(currentProperty, uiSchemaTitle);
```

and

```ts
parentSchema?.[PROPERTIES_KEY]?.[currentProperty]?.title;
```

dagda1:
I notice the existing code did this with message

```ts
stack = message!;
```

This tells the typescript compiler that message will not be null or undefined.  If this is the case would it not make sense to make `message` in `ErrorObject` not optional?

And for `parentSchema`, do you know how I could be able to test it being undefined?  The only way I can think is if `transformRJSFValidationErrors` was not private.

heath-freenome:
> @heath-freenome I am getting yellow lines in the coverage report around the optional chaining on `message` and `parentSchema`.
> 
> I notice the existing code does this with message
> 
> ```ts
> stack = message!;
> ```
> 
> This tells the typescript compiler that `message` will not be null or undefined. If this is the case, would it not make sense to make `message` in `ErrorObject` not optional?

`ErrorObject` comes from `ajv` and from all the testing it normally has a value so that's the reason for the `!`

> 
> And for `parentSchema`, do you know how I could be able to test it is undefined? The only way I can think is if `transformRJSFValidationErrors` was not private.

If you look up the top around line 16 you'll see how you can "expose" that function in a manner similar to how we exposed `withIdRefPrefix`

heath-freenome:
One last thing, don't forget to update the `CHANGELOG.md` similar to how it is done [here](https://github.com/rjsf-team/react-jsonschema-form/pull/3371/files)

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ should change the heading of the (upcoming) version to include a major version b
 -->
 # 5.0.0-beta-17
 
+## @rjsf/validator-ajv8
+- Updated `AJV8Validator#transformRJSFValidationErrors` to return more human readable error messages.  The ajv8 `ErrorObject` message is enhanced by replacing the error message field with either the `uiSchema`'s `ui:title` field if one exists or the `parentSchema` title if one exists.
```
heath-freenome, line 21: Can you add this to the `@rjsf/validator-ajv8` section below instead?

packages/validator-ajv8/src/validator.ts
```diff
@@ -234,17 +235,19 @@ export default class AJV8Validator<
...
         if (uiSchemaTitle) {
-          message = message?.replace(currentProperty, uiSchemaTitle);
+          message = message.replace(currentProperty, uiSchemaTitle);
         } else {
-          const parentSchemaTitle =
-            parentSchema?.[PROPERTIES_KEY]?.[currentProperty]?.title;
+          const parentSchemaTitle = get(
+            parentSchema,
+            `${PROPERTIES_KEY}.${currentProperty}.title`
+          );
```
heath-freenome, line 243: You could also use the array notation for the props instead saving the need to construct a string, i.e.:
```
const parentSchemaTitle = get(
            parentSchema,
            [PROPERTIES_KEY, currentProperty, "title"]
          );
```

heath-freenome (approved):

heath-freenome:
One required and one optional change and then we can merge it!

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/validator-ajv8
 - Remove alias for ajv -> ajv8 in package.json. This fixes [#3215](https://github.com/rjsf-team/react-jsonschema-form/issues/3215).
 
+- Updated `AJV8Validator#transformRJSFValidationErrors` to return more human readable error messages.  The ajv8 `ErrorObject` message is enhanced by replacing the error message field with either the `uiSchema`'s `ui:title` field if one exists or the `parentSchema` title if one exists.
```
heath-freenome, line 26: Let's keep the changes tight together
```suggestion
- Updated `AJV8Validator#transformRJSFValidationErrors` to return more human readable error messages.  The ajv8 `ErrorObject` message is enhanced by replacing the error message field with either the `uiSchema`'s `ui:title` field if one exists or the `parentSchema` title if one exists.
```

packages/validator-ajv8/package.json
```diff
@@ -33,7 +33,7 @@
     "lodash-es": "^4.17.15"
   },
   "peerDependencies": {
-    "@rjsf/utils": "^5.0.0-beta.12"
+    "@rjsf/utils": "^5.0.0-beta.15"
```
heath-freenome, line 36: It's now beta.16
```suggestion
    "@rjsf/utils": "^5.0.0-beta.16"
```

packages/validator-ajv8/src/validator.ts
```diff
@@ -199,20 +202,67 @@ export default class AJV8Validator<
...
+        params,
+        schemaPath,
+        parentSchema,
+        ...rest
+      } = e;
+      let { message = "" } = rest;
+
       let property = instancePath.replace(/\//g, ".");
+
       let stack = `${property} ${message}`.trim();
```
heath-freenome, line 224: Let's keep the variables all together
```suggestion
      let { message = "" } = rest;
      let property = instancePath.replace(/\//g, ".");
      let stack = `${property} ${message}`.trim();
```

packages/validator-ajv8/test/validator.test.ts
```diff
@@ -12,11 +12,19 @@ import {
 
 import AJV8Validator from "../src/validator";
 import { Localizer } from "../src";
+import { type ErrorObject } from "ajv";
```
heath-freenome, line 15: Move this to the top of the file please, and I don't believe you need the `type` either

packages/validator-ajv8/src/validator.ts
```diff
@@ -199,20 +202,67 @@ export default class AJV8Validator<
...
+
       if ("missingProperty" in params) {
         property = property
           ? `${property}.${params.missingProperty}`
           : params.missingProperty;
-        stack = message!;
+
+        const currentProperty: string = params.missingProperty;
+
+        const uiSchemaTitle = getUiOptions(
```
heath-freenome, line 233: Let's keep the variables together
```suggestion
        const currentProperty: string = params.missingProperty;
        const uiSchemaTitle = getUiOptions(
```

heath-freenome (commented):
Ah... Sorry for being so picky, but a few more little things. I'd make them myself except I can't easily update the location of the `import`

packages/validator-ajv8/test/validator.test.ts
dagda1: I'll commit this but I think the type is important because that imports only the types and none of the bundled javascript

heath-freenome: In the scope of the test it is probably not that big of a deal

heath-freenome:
@dagda1 Thanks for all your hard work on this PR and dealing with the feedback. I'm just awaiting the final move of the import in the test then I will merge

packages/validator-ajv8/test/validator.test.ts
```diff
@@ -1,4 +1,5 @@
 import Ajv2019 from "ajv/dist/2019";
+import { ErrorObject } from "ajv";
```
heath-freenome, line 2:
```suggestion
import { ErrorObject } from "ajv";
import Ajv2019 from "ajv/dist/2019";
```

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/validator-ajv8
 - Remove alias for ajv -> ajv8 in package.json. This fixes [#3215](https://github.com/rjsf-team/react-jsonschema-form/issues/3215).
+- Updated `AJV8Validator#transformRJSFValidationErrors` to return more human readable error messages.  The ajv8 `ErrorObject` message is enhanced by replacing the error message field with either the `uiSchema`'s `ui:title` field if one exists or the `parentSchema` title if one exists.
```
heath-freenome, line 25:
```suggestion
- Updated `AJV8Validator#transformRJSFValidationErrors` to return more human readable error messages.  The ajv8 `ErrorObject` message is enhanced by replacing the error message field with either the `uiSchema`'s `ui:title` field if one exists or the `parentSchema` title if one exists. Fixes [#3246](https://github.com/rjsf-team/react-jsonschema-form/issues/3246)
```

dagda1:
> @dagda1 Thanks for all your hard work on this PR and dealing with the feedback. I'm just awaiting the final move of the import in the test then I will merge

the feedback was constructive so thank you for the good review

Issue #3215, opened 2022-10-29
codepen examples in docs broken


epicfaace:
See https://react-jsonschema-form.readthedocs.io/en/latest/usage/single/

Do you know what might have happened @heath-freenome @nickgros ?

heath-freenome:
@epicfaace Can you be more specific about how they are broken? It's pretty clear we forgot to import the `Form` from `@rjsf/core` but the old links didn't either.

epicfaace:
Running into this issue (https://github.com/skypackjs/skypack-cdn/issues/337) now.

epicfaace:
> @epicfaace Can you be more specific about how they are broken? It's pretty clear we forgot to import the `Form` from `@rjsf/core` but the old links didn't either.

By the way @heath-freenome , the reason they're broken now is that previously, we just used the UMD version and didn't need the import statements, as JSONSchemaForm was a global variable available. Now the code examples all use imports from different packages so we'll have to switch from using UMD to configuring codepen to also support js `import` statements.

PR #3359, opened 2023-01-08
Fix umd builds

epicfaace:
Fix umd builds by adding a default JSONSchemaForm export. This addresses part of the issue of https://github.com/rjsf-team/react-jsonschema-form/issues/3215.

### DIFF
--- a/packages/core/dts.config.js
+++ b/packages/core/dts.config.js
@@ -1,0 +1,11 @@
+// Not transpiled with TypeScript or Babel, so use plain Es6/Node.js!
+/**
+ * @type {import('dts-cli').DtsConfig}
+ */
+module.exports = {
+    // This function will run for each entry/format/env combination
+    rollup(config, options) {
+      config.output.name = "JSONSchemaForm";
+      return config; // always return a config.
+    },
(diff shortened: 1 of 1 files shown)


epicfaace:
Note that this will only partially fix the problem, as you still can't import the validator through a UMD build...

packages/core/dts.config.js
```diff
@@ -0,0 +1,11 @@
...
+/**
+ * @type {import('dts-cli').DtsConfig}
+ */
+module.exports = {
+    // This function will run for each entry/format/env combination
+    rollup(config, options) {
+      config.output.name = "JSONSchemaForm";
+      return config; // always return a config.
+    },
+  };
```
heath-freenome, line 11: Adding a new line at the end to avoid the little 
Screen Shot 2023-01-08 at 1 48 55 PM

```suggestion
  };
  
```

heath-freenome:
> Note that this will only partially fix the problem, as you still can't import the validator through a UMD build...

Would doing something similar for those two libraries help?

heath-freenome (approved):

PR #3362, opened 2023-01-08
Remove ajv8 alias

epicfaace:

Remove ajv8 alias as it just complicates things, I don't see a benefit to keep it in. This will also address https://github.com/skypackjs/skypack-cdn/issues/337 (which needs to resolved to fix https://github.com/rjsf-team/react-jsonschema-form/issues/3215).

### DIFF
--- a/packages/validator-ajv8/package-lock.json
+++ b/packages/validator-ajv8/package-lock.json
@@ -9,8 +9,8 @@
       "version": "5.0.0-beta.15",
       "license": "Apache-2.0",
       "dependencies": {
+        "ajv": "^8.11.0",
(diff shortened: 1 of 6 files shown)


heath-freenome:
@epicfaace This will break the `playground` deployment. Work with @nickgros to figure out how to solve the issue that was fixed with this

nickgros:
To clarify, it doesn't break the deployment, but it does break the devmode server

nickgros:
Trying to do this and fix the dev server issue that this change will introduce in #3371

heath-freenome:
Closing in favor of #3371

PR #3377, opened 2023-01-14
fix: reimplement PR 2356 to enable searching SelectWidget by label

heath-freenome:

Antd Select widget allows to search in the option array. The current code makes the Select search by the key, not the value - the label that is presented to the user.

Reimplements #2356

- Updated the `SelectWidget` to implement a `filterOption` function that compares the string against the select option label

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ should change the heading of the (upcoming) version to include a major version b
 -->
 # 5.0.0-beta-17
 
+## @rjsf/antd
+- Enable searching in the `SelectWidget` by the label that the user sees rather than by the value
+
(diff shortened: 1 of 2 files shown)


nickgros (approved):

nickgros:
Change looks good but the antd widget is totally busted for me in the current playground (can't make selections), so it's hard to determine how it works before this change.

heath-freenome:
> Change looks good but the antd widget is totally busted for me in the current playground (can't make selections), so it's hard to determine how it works before this change.

@nickgros I can select things in the current playground, but the selection box just won't close. I've created #3381 to get this fixed

PR #2142, opened 2020-11-25
Cast the value as an Array in CheckboxesWidget

Mogztter:
### Reasons for making this change

Fixes #2141

PR #3379, opened 2023-01-15
fix: Cast the value as an Array in CheckboxesWidget

### LINKED ISSUE #2141: CheckboxesWidget should ensure that the value is an Array
### Prerequisites

### Description

`CheckboxesWidget` assumes that the `value` is an Array but does not check the type of the value.
At runtime, the widget will throw an exception when a value is selected or deselected.

### Steps to Reproduce

See the codesandbox example: https://codesandbox.io/s/checkboxeswidget-with-non-array-value-react-jsonschema-form-qi96b?file=/src/App.js

Important parts:

```js
var formData = {
  typeArticle: "Essai"
};
```

#### Expected behavior

The widget should check that the value is an Array. If not, the widget could either create an Array from the value or ignore the value.

#### Actual behavior

Throws an exception when the user is trying to select or deselect a value.
Please note that the value passed as a String will be correctly selected.

In the above example, the value "Essai" is selected (even if the value was `Essai` and not `["Essai"]`):

selected

### Version

`2.4.0`

heath-freenome:

Fixes #2141 by reimplementing #2142 for all themes but `antd`

When the value passed to the `CheckboxesWidget` was a single value rather than an array, things would break in the control
- Updated the `CheckboxesWidget` in all themes but `antd` (which uses simpler logic) to fix ensure the value used in the helper functions is an array
- Added a test in `ArrayField_test.ts` that verifies the fix
- Updated the snapshot in `bootstrap-4` which showed an issue when the value string contained an element of the enumeration value rather than the whole value
- Refactored the common `selectValue` and `deselectValue` functions used in `CheckboxesWidget` into `@rjsf/utils` and used them in the `CheckboxesWidget`

### Checklist

### DIFF
--- a/packages/fluent-ui/src/CheckboxesWidget/CheckboxesWidget.tsx
+++ b/packages/fluent-ui/src/CheckboxesWidget/CheckboxesWidget.tsx
@@ -17,16 +17,16 @@ const styles_red = {
   fontFamily: `"Segoe UI", "Segoe UI Web (West European)", "Segoe UI", -apple-system, BlinkMacSystemFont, Roboto, "Helvetica Neue", sans-serif;`,
 };
 
-const selectValue = (value: any, selected: any, all: any) => {
+const selectValue = (value: any, selected: any[], all: any) => {
   const at = all.indexOf(value);
   const updated = selected.slice(0, at).concat(value, selected.slice(at));
 
   // As inserting values at predefined index positions doesn't work with empty
   // arrays, we need to reorder the updated selection to match the initial order
-  return updated.sort((a: any, b: any) => all.indexOf(a) > all.indexOf(b));
+  return updated.sort((a, b) => Number(all.indexOf(a) > all.indexOf(b)));
 };
 
-const deselectValue = (value: any, selected: any) => {
+const deselectValue = (value: any, selected: any[]) => {
   return selected.filter((v: any) => v !== value);
 };
 
@@ -50,6 +50,7 @@ export default function CheckboxesWidget<
   rawErrors = [],
 }: WidgetProps<T, S, F>) {
   const { enumOptions, enumDisabled } = options;
+  const checkboxesValues = Array.isArray(value) ? value : [value];
 
   const _onChange =
     (option: any) =>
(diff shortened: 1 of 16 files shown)


packages/fluent-ui/src/CheckboxesWidget/CheckboxesWidget.tsx
```diff
@@ -17,16 +17,16 @@ const styles_red = {
...
   // As inserting values at predefined index positions doesn't work with empty
   // arrays, we need to reorder the updated selection to match the initial order
-  return updated.sort((a: any, b: any) => all.indexOf(a) > all.indexOf(b));
+  return updated.sort((a, b) => Number(all.indexOf(a) > all.indexOf(b)));
 };
 
-const deselectValue = (value: any, selected: any) => {
+const deselectValue = (value: any, selected: any[]) => {
   return selected.filter((v: any) => v !== value);
 };
```
nickgros, line 31: Used in multiple packages, consider moving to utils

packages/semantic-ui/src/CheckboxesWidget/CheckboxesWidget.tsx
```diff
@@ -12,19 +12,19 @@ import { getSemanticProps } from "../util";
 
 function selectValue<S extends StrictRJSFSchema = RJSFSchema>(
   value: EnumOptionsType<S>["value"],
-  selected: any,
+  selected: any[],
   all: any[]
```
nickgros, line 16: Can we use these more accurate types for the other declarations of this function (or if moving to utils, use these types there)? Same for `deselectValue`.

nickgros (changes requested):

packages/fluent-ui/src/CheckboxesWidget/CheckboxesWidget.tsx
heath-freenome: Will do!

packages/semantic-ui/src/CheckboxesWidget/CheckboxesWidget.tsx
heath-freenome: Yeah, I'll pull this one into utils

nickgros (approved):

PR #3382, opened 2023-01-16
Use pnpm + workspaces

nickgros:

pnpm + workspaces + Lerna deduplicates dependencies between packages. Only one copy of a dependency version will be installed across the entire repo. Only one lockfile is needed for the entire project. This should greatly reduce the amount of time needed to install dependencies, and improve overall developer experience.

### Checklist


package.json
```diff
@@ -42,5 +43,11 @@
     "ts-jest": "^27.1.5",
     "tslib": "^2.4.1",
     "typescript": "^4.9.4"
-  }
+  },
+  "resolutions": {
+    "@types/react": "^17.0.39"
+  },
```
nickgros, line 49: TODO: I think I can remove this and just have this in utils/package.json

packages/chakra-ui/src/AltDateWidget/AltDateWidget.tsx
```diff
@@ -129,7 +129,7 @@ function AltDateWidget<
 
   return (
     <Box>
-      <Box display="flex" flexWrap="wrap" alignItems="center" justify="center">
+      <Flex flexWrap="wrap" alignItems="center">
```
nickgros, line 132: I think I happened to pull in a more recent version of chakra-ui and discovered `justify` is not a valid prop on `Box`. We can use the `Flex` component, instead. I think we actually want to keep this left-aligned, though, so I ended up removing `justify="center"` anyways.

packages/playground/vite.config.ts
```diff
@@ -15,13 +15,13 @@ export default defineConfig({
...
+      "@mui/styles": path.resolve("../../node_modules", "@mui/styles"),
       "@material-ui/styles": path.resolve(
-        "./node_modules",
+        "../../node_modules",
         "@material-ui/styles"
       ),
-      "@emotion/react": path.resolve("./node_modules/@emotion/react"),
-      "@emotion/styled": path.resolve("./node_modules/@emotion/styled"),
+      "@emotion/react": path.resolve("../../node_modules/@emotion/react"),
+      "@emotion/styled": path.resolve("../../node_modules/@emotion/styled"),
```
nickgros, line 24: Since the dependencies are now deduplicated and stored in the root node_modules folder, we need to update these references.

packages/utils/package.json
```diff
@@ -60,6 +60,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "resolutions": {
+    "@types/react": "^17.0.48"
+  },
```
nickgros, line 65: Some dependency is pulling in `@types/react@*` which causes us to pull an incompatible version of `@types/react`, so let's force using v17.

cdaringe:
strong encouragement to check out pnpm workspaces. pnpm workspaces, based on my personal research, trump yarn workspaces by features/stability/performance. my yarn workspace experience is out-of-date by probably 8 months, but pnpm experience is fresh, and it's delightful.

bollwyvl:
:+1: to yarn and (modern) lerna. I have no experience with `pnpm`, but also have very few complaints about the yarn 1.x experience, while 2.x+ seems to have gone wandering in the woods a bit.

One can go a little beyond the built-in yarn deduplication with [yarn-deduplicate](https://www.npmjs.com/package/yarn-deduplicate), which can squeeze a few more things out (especially when paired with a few tactical `resolutions`).

Also, on the typescript side: having a metapackage which contains references to all the other packages allows for having a single, incremental `tsc -b` entrypoint, which makes a big difference in both improving performance of type-checking at scale, and complexity.

nickgros:
@cdaringe @bollwyvl thanks for the perspectives! While I have more experience with yarn, pnpm ended up being a breeze to configure comparatively, so we'll go with pnpm for now.

If incremental builds end up being pretty easy to switch to, then I'll include that in this change, as well.

ahkhanjani:
Why not use Nx for distributed caching? Has the same benefits with way better tooling and will reduce lint, test and build time to potentially under 10 seconds

ahkhanjani:
> I don't think that this change would preclude adding Nx cloud caching, which has no impact on dependency install time anyways.

@nickgros But why use pnpm workspaces over Nx itself?

cdaringe:
As an nx user and a pnpm user and a pnpm+nx user, there are many reasons why would or wouldnt. Reasons to not would be high maintenance costs, poor debugging story, high noise:signal ratio of boilerplate, stringly typed references to workspace/project entities. There are good reasons as well (affected). Not sure nx is worth the effort for a project if this size. Were i the maintainer, I opt out until something demanded it

ahkhanjani:
> As a nx user and a pnpm user and a pnpm+nx user, there are many reasons why would or wouldnt. Reasons to not would be high maintenance costs, poor debugging story, high noise:signal ratio of boilerplate, stringly typed references to workspace/project entities. There are good reasons as well (affected). Not sure nx is worth the effort for a project of this size. Were i the maintainer, Id opt out until something demanded it

@cdaringe I think having internal libraries (validator-ajv6/8, utils, etc.) as a regular import and not a package, and most importantly "affected" will make the local and deployment process a hundred times faster at the very least. running all the tests and linting and then building every package for the change of only one package is something worth considering in my opinion. 

Best regards

ahkhanjani:
Any updates on this?

nickgros:
@ahkhanjani I keep getting errors in CI that are hard to reproduce locally. Haven't had much time to try to find the root cause.

ahkhanjani:
@nickgros I will try to do the same thing and fix some issues. The problems seem to be poor typing issues in some places

PR #2890, opened 2022-06-23
fix(@rjsf/core): custom field that is any of displays twice

truh:
### Reasons for making this change

If a custom field is provided for an anyOf schema, it should be rendered in place of the `AnyOfField` component rather than be wrapped in the `AnyOfField` component. Same for oneOf.

https://github.com/rjsf-team/react-jsonschema-form/issues/2691

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests. I've run `npm run test:update` to update snapshots, if needed.
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
  - [x] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/master/CHANGELOG.md) with a description of the PR
* [ ] **I'm adding a new feature**
  - [x] I've updated the playground with an example use of the feature

PR #3400, opened 2023-01-25
Fix: 2691 with an opt-out for custom fields for anyOf/oneOf

heath-freenome:

Fixes: #2691 better by providing an opt-out behavior for the automatic wrapping of fields by anyOf/oneOf components

The original fix (#2890) to never wrap components is a breaking change from 4.x and will cause existing users problems if they were counting on the old behavior (which my implementation did). This new fix provides an opt-out instead thereby reversing the breaking change

- In `@rjsf/utils`, updated just the `UiSchema` type to add the new `ui:fieldReplacesAnyOrOneOf` flag
- In `@rjsf/core`, updated `SchemaField` to only skip wrapping a custom field in the `anyOf`/`oneOf` if the `ui:fieldReplacesAnyOrOneOf` is true along with there being a custom field
  - Updated/added tests to verify the fix
- Updated the `uiSchema` documentation to include `ui:fieldReplacesAnyOrOneOf`

### Checklist

### DIFF
--- a/docs/api-reference/uiSchema.md
+++ b/docs/api-reference/uiSchema.md
@@ -71,6 +71,13 @@ Specify either the name of a field that is used to look up an implementation fro
 
 See [Custom Widgets and Fields](https://react-jsonschema-form.readthedocs.io/en/stable/api-reference/custom-widgets-fields#custom-field-components) for more information about how to use this property.
 
+### ui:fieldReplacesAnyOrOneOf
+
+By default, any field that is rendered for an `anyOf`/`oneOf` schema will be wrapped inside the `AnyOfField` or `OneOfField` component.
+By providing a `true` value for this flag in association with a custom `ui:field` the default behavior is skipped.
+Instead, it will be up to the custom field implementation to deal with rendering the `anyOf`/`oneOf` behavior.
+Providing a `false` value, will maintain the wrapping behavior rather than the replacing behavior.
+
 ### ui:options
 
 The `ui:options` property cannot be nested inside itself and thus is the last exception.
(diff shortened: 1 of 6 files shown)


heath-freenome:
@truh Here's the update to your fix in case you are interested

docs/api-reference/uiSchema.md
```diff
@@ -71,6 +71,13 @@ Specify either the name of a field that is used to look up an implementation fro
 
 See [Custom Widgets and Fields](https://react-jsonschema-form.readthedocs.io/en/stable/api-reference/custom-widgets-fields#custom-field-components) for more information about how to use this property.
 
+### ui:fieldReplacesAnyOrOneOf
+
+By default, any field that is rendered for an `anyOf`/`oneOf` schema will be wrapped inside the `AnyOfField` or `OneOfField` component.
+By providing a `true` value for this flag in association with a custom `ui:field` the default behavior is skipped.
+Instead, it will be up to the custom field implementation to deal with rendering the `anyOf`/`oneOf` behavior.
+Providing a `false` value, will maintain the wrapping behavior rather than the replacing behavior.
```
heath-freenome, line 79: @nickgros / @epicfaace, This sounds a bit awkward even to me, so feel free to improve it if you can.

truh:
> @truh Here's the update to your fix in case you are interested

Thanks for the heads-up.

docs/api-reference/uiSchema.md
nickgros, line 79:
```suggestion
By default, any field that is rendered for an `anyOf`/`oneOf` schema will be wrapped inside the `AnyOfField` or `OneOfField` component. This default behavior may be undesirable if your custom field already handles behavior related to choosing one or more subschemas contained in the `anyOf`/`oneOf` schema.
By providing a `true` value for this flag in association with a custom `ui:field`, the wrapped components will be omitted, so just one instance of the custom field will be rendered.
If the flag is omitted or set to `false`, your custom field will be wrapped by `AnyOfField`/`OneOfField`.
```

nickgros: @heath-freenome does this make more sense?

PR #3406, opened 2023-01-27
fix: Pass undefined value when MultiSchemaField has nothing selected

heath-freenome:

Some implementations of `SelectWidget` may allow clearing the value selected. In those circumstances, the `value` being passed back to the widget should be `undefined` not `-1`
- Updated the `MultiSelectField` render function to send `undefined` as the value when the selectedOption is `-1`.
  - Also make sure the selected `option` is also null in that situation
  - Added tests to detect that undefined is passed rather than -1
- Update the `Form` to receive an optional `ref` prop, removing the need to declare it on the `ThemeProps`

### Checklist

### DIFF
--- a/packages/core/src/components/fields/MultiSchemaField.tsx
+++ b/packages/core/src/components/fields/MultiSchemaField.tsx
@@ -201,7 +201,8 @@ class AnyOfField<
     const rawErrors = get(errorSchema, ERRORS_KEY, []);
     const fieldErrorSchema = omit(errorSchema, [ERRORS_KEY]);
 
-    const option = retrievedOptions[selectedOption] || null;
+    const option =
+      selectedOption >= 0 ? retrievedOptions[selectedOption] || null : null;
     let optionSchema: S;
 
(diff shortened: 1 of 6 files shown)


packages/core/src/components/fields/MultiSchemaField.tsx
```diff
@@ -201,7 +201,8 @@ class AnyOfField<
     const rawErrors = get(errorSchema, ERRORS_KEY, []);
     const fieldErrorSchema = omit(errorSchema, [ERRORS_KEY]);
 
-    const option = retrievedOptions[selectedOption] || null;
+    const option =
+      selectedOption >= 0 ? retrievedOptions[selectedOption] || null : null;
```
nickgros, line 205: Why not just do this and revert L238?
```suggestion
    const option =
      selectedOption >= 0 ? retrievedOptions[selectedOption] || null : undefined;
```

nickgros (approved):

PR #3202, opened 2022-10-18
Add "list-inline-item" as a class name on LI elements in the AltDateWidget

danfunk:
…rap 4 docs: https://getbootstrap.com/docs/4.0/content/typography/#inline

### Reasons for making this change

use list-inline-item for li items in the AltDateWidget, as per bootstrap 4 docs: https://getbootstrap.com/docs/4.0/content/typography/#inline

It is a minor change.  I did run the tests to assure this does not break anything. 

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests. I've run `npm run test:update` to update snapshots, if needed.
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

Before:
Screenshot 2023-02-03 at 4 19 44 PM

After:
Screenshot 2023-02-03 at 4 19 33 PM

PR #3430, opened 2023-02-04
Fix: Prevent undefined properties at the root level

heath-freenome:

Fixes: #3424 by preventing the inclusion of undefined properties at the root level
- Updated `@rjsf/utils`, making `computeDefaults()` helper in `getDefaultFormState()` to skip adding undefined values when `allowEmptyObject` is set.
  - Updated the `getDefaultFormState()` definition in the `ValidatorType` to add this new option
  - Updated the tests accordingly
- Updated `@rjsf/core`, switching the `excludeObjectChildren` with `allowEmptyObject` in `Form`
- Updated the documentation for `getDefaultFormState()` to add this new option
  - Also added information the changelog for PR #3202

### Checklist

### DIFF
--- a/packages/utils/src/schema/getDefaultFormState.ts
+++ b/packages/utils/src/schema/getDefaultFormState.ts
@@ -34,13 +34,6 @@ export enum AdditionalItemsHandling {
   Fallback,
 }
 
-/** List of choices for `includeUndefinedValues` that should pass false to the recursion
- */
-const passAsFalseForIncludeUndefinedValues: Array<string | boolean> = [
-  "excludeObjectChildren",
-  "allowEmptyObject",
-];
-
 /** List of choices for `includeUndefinedValues` that should allow undefined values to be saved
  */
 const allowUndefinedForIncludeUndefinedValues: Array<string | boolean> = [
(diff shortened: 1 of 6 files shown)


packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -194,12 +213,22 @@ export function computeDefaults<
             get(defaults, [key]),
             rootSchema,
             get(formData, [key]),
-            includeUndefinedValues === "excludeObjectChildren"
+            passAsFalseForIncludeUndefinedValues.includes(
+              includeUndefinedValues
+            )
               ? false
               : includeUndefinedValues
```
nickgros, line 220: Instead of inlining this can we set it before L210?
```ts
          const nestedIncludeUndefinedValues = passAsFalseForIncludeUndefinedValues.includes(
              includeUndefinedValues
            )
              ? false
              : includeUndefinedValues
          // Compute the defaults for this node, with the parent defaults we might
          // have from a previous run: defaults[key].
          const computedDefault = computeDefaults<T, S, F>(
            validator,
            get(schema, [PROPERTIES_KEY, key]),
            get(defaults, [key]),
            rootSchema,
            get(formData, [key]),
            nestedIncludeUndefinedValues
          );
```

Also, couldn't
```ts
passAsFalseForIncludeUndefinedValues.includes(includeUndefinedValues) ? false : includeUndefinedValues
```
be simplified to this?
```ts
includeUndefinedValues === true
```

heath-freenome: Yes, yes it could. Great simplification

nickgros (approved):

PR #3455, opened 2023-02-25
Rm docs workflow

epicfaace:
Because we already build docs in another workflow.


heath-freenome:
@epicfaace Can you also update github so that it doesn't have the `docs` workflow as being required?

heath-freenome:
@epicfaace Seems like you'll need to deal with the conflicts and also update Github so that it doesn't make the `docs` step required.

epicfaace:
> Screenshot 2023-02-28 at 9 42 08 AM
> 
> @epicfaace Can you also update github so that it doesn't have the `docs` workflow as being required?

Updated!

heath-freenome (approved):

PR #3472, opened 2023-03-05
fix: Fixed FileWidget to work across all themes

### LINKED ISSUE #2095: [bootstrap-4] File uploads result in error ".file should match format "data-url""
### Prerequisites

### Description

Using the `bootstrap-4` theme, submitting a file triggers an error:

```
.file should match format "data-url"
```

### Steps to Reproduce

1. [Playground link](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJmaWxlIjoiQzpcXGZha2VwYXRoXFxXZWJTdG9ybS0yMDIwLjIuMy5leGUifSwic2NoZW1hIjp7InRpdGxlIjoiRmlsZXMiLCJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJmaWxlIjp7InR5cGUiOiJzdHJpbmciLCJmb3JtYXQiOiJkYXRhLXVybCIsInRpdGxlIjoiU2luZ2xlIGZpbGUifSwiZmlsZXMiOnsidHlwZSI6ImFycmF5IiwidGl0bGUiOiJNdWx0aXBsZSBmaWxlcyIsIml0ZW1zIjp7InR5cGUiOiJzdHJpbmciLCJmb3JtYXQiOiJkYXRhLXVybCJ9fSwiZmlsZXNBY2NlcHQiOnsidHlwZSI6InN0cmluZyIsImZvcm1hdCI6ImRhdGEtdXJsIiwidGl0bGUiOiJTaW5nbGUgRmlsZSB3aXRoIEFjY2VwdCBhdHRyaWJ1dGUifX19LCJ1aVNjaGVtYSI6eyJmaWxlc0FjY2VwdCI6eyJ1aTpvcHRpb25zIjp7ImFjY2VwdCI6Ii5wZGYifX19LCJ0aGVtZSI6ImJvb3RzdHJhcC00IiwibGl2ZVNldHRpbmdzIjp7InZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2V9fQ==)
2. Click "Choose file" for the first file upload
3. Select any file from your computer
4. Click "Submit"

#### Expected behavior

No errors are displayed.

#### Actual behavior

The following error is displayed:

```
.file should match format "data-url"
```

### Version

`master`

heath-freenome:

Fixes #2095 by making `FileWidget` use the `BaseInputTemplate` for rendering the file input
- Updated `@rjsf/utils` to fix two small bugs found while fixing the issue:
  - `dataURItoBlob()` was updated to deal with parsing errors in `atob()` with malformed URL and to decode the name to a human-readable string - Added additional tests to verify the fixes
  - `replaceStringParameters()` was updated to split using a regex to deal with situations where the params themselves contain replaceable character identifiers
    - Added an additional test to verify the bug was fixed
- Updated `@rjsf/core` to switch `FileWidget` so that it uses the `BaseInputTemplate` to render the input, passing a new `onChangeOverride` function to deal with file events
  - Updated `BaseInputTemplate` to favor `onChangeOverride` rather than the default `onChange` handler it builds
- Updated the `BaseInputTemplate` in all of the themes to also favor `onChangeOverride` if specified over the default `onChange` handler
  - Updated the snapshots in all themes to reflect the improved `FileWidget` use of `BaseInputTemplate` in all themes
- Updated `@rjsf/bootstrap-4` to delete the `FileWidget` in favor of the improved `core` one
- Updated `@rjsf/material-ui` and `@rjsf/mui` to have the `BaseInputTemplate` support the automatic `shrink` of the label when the `type` of the input is `date`, `datetime-local` and `file`
  - Also removed the `DateWidget` and `DateTimeWidget` in both themes since they were only providing the `shrink` behavior which is now encapsulated in `BaseInputTemplate`
- Updated the `CHANGELOG.md` file accordingly
 
### Checklist


packages/core/src/components/widgets/FileWidget.tsx
```diff
@@ -150,20 +147,14 @@ function FileWidget<
...
-          multiple={multiple}
-          accept={options.accept ? String(options.accept) : undefined}
-          aria-describedby={ariaDescribedByIds<T>(id)}
-        />
-      </p>
+      <BaseInputTemplate
+        {...props}
+        disabled={disabled || readonly}
+        type="file"
+        onChangeOverride={handleChange}
```
nickgros, line 154: How were you able to supply this `onChangeOverride` prop without adding it to the `WidgetProps` type?

packages/antd/src/templates/BaseInputTemplate/index.tsx
```diff
@@ -46,8 +47,10 @@ export default function BaseInputTemplate<
 
   const handleNumberChange = (nextValue: number | null) => onChange(nextValue);
 
-  const handleTextChange = ({ target }: React.ChangeEvent<HTMLInputElement>) =>
-    onChange(target.value === "" ? options.emptyValue : target.value);
+  const handleTextChange = onChangeOverride
+    ? onChangeOverride
+    : ({ target }: React.ChangeEvent<HTMLInputElement>) =>
+        onChange(target.value === "" ? options.emptyValue : target.value);
```
nickgros, line 53: This default `onChange` behavior is duplicated across all themes. Might make more sense to refactor to utils, e.g.

```ts
// Is there a better name?
export default function useOnEventTargetStringChange<
  T = any,
>(
  options: WidgetProps<T, S, F>["options"],
  onChange: WidgetProps<T, S, F>["onChange"]
) {
  return useCallback(
    ({ target: { value } }: React.ChangeEvent<HTMLInputElement>) =>
      onChange(value === "" ? options.emptyValue : value),
    [onChange, options.emptyValue]
  );
}
```

I am leaning towards the idea that this logic should be extracted from `BaseInputTemplate` altogether, since it doesn't work for all cases. Then we wouldn't need an `onChangeOverrides` prop. The downside is that we would have to pass this in almost every time we use `BaseInputTemplate`:


```tsx
export default function CustomWidget() {
  const { onChange: _onChange, value, options, registry } = props;
  const BaseInputTemplate = getTemplate<"BaseInputTemplate", T, S, F>(
    "BaseInputTemplate",
    registry,
    options
  );
  const onChange = useOnEventTargetStringChange(options, _onChange)
  return (
    <BaseInputTemplate
      {...props}
      onChange={onChange}
    />
  );
}
```

I'm not sure if this verbosity + inversion is worth it for one (known) edge case...WDYT?

packages/core/src/components/widgets/FileWidget.tsx
nickgros: Is file upload the only case that would benefit from this style?

Having both `onChange` and `onChangeOverride` seems like a confusing API, but I'm not sure if it should be clarified with better prop names or if it should have a different design.

packages/material-ui/src/BaseInputTemplate/BaseInputTemplate.tsx
```diff
@@ -66,6 +70,12 @@ export default function BaseInputTemplate<
 
   const { schemaUtils } = registry;
   const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema);
+  const DisplayInputLabelProps = TYPES_THAT_SHRINK_LABEL.includes(type)
+    ? {
+        ...InputLabelProps,
+        shrink: true,
+      }
+    : InputLabelProps;
```
nickgros, line 78: Nice simplification

packages/antd/src/templates/BaseInputTemplate/index.tsx
nickgros: A lower-lift option is to extend the default logic to also handle files (possibly being aware of the input type).

```ts
export default function useOnChangeEvent<
  T = any,
>(
  options: WidgetProps<T, S, F>["options"],
  onChange: WidgetProps<T, S, F>["onChange"],
  isFileInput?: boolean
) {
  return useCallback(
    (event: React.ChangeEvent<HTMLInputElement>) => {
      if (isFileInput) {
	// ...
      } else {
        onChange(value === "" ? options.emptyValue : value)
      }
    },
    [onChange, options.emptyValue, isFileInput]
  );
}
```

nickgros (commented):
Like the fix but not sure about the approach, have some suggestions but not sure if they are what we want either

nickgros (approved):

PR #2581, opened 2021-10-14
fix: Ensure errorSchema is a plain object before going deep in toErrorList

lpillonel:
### Reasons for making this change

In our application, we also pass some "__warnings" properties within the errorSchema (warnings which are then extracted and used in a custom SchemaField). Without this change, the toErrorList function just crash when it encounter a __warnings array, as it currently assumes that anything but __errors are objects.

If this is related to existing tickets, include links to them as well. Use the syntax `fixes #[issue number]` (ex: `fixes #123`).

### Checklist

* [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://react-jsonschema-form.readthedocs.io/en/latest/#contributing) of the Markdown text I've added
* [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests
  - [ ] I've updated [docs](https://react-jsonschema-form.readthedocs.io/) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/master/CHANGELOG.md) with a description of the PR
* [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #3527, opened 2023-03-20
fix: reimplement stronger type checking from #2581

heath-freenome:

Reimplemented the stronger type checking from #2581
- Updated `@rjsf/validator-ajv8` to check for plain object types before recursing in `toErrorList()` and `unwrapErrorHandler()`
  - Switched from using `isObject` to `isPlainObject` since the other places it was used were also checking plain objects

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Updated the `UiSchema` to support the optional `filePreview?: boolean` option and to add a new `TranslatableString.PreviewLabel` to the `enums`
 
+## @rjsf/validator-ajv8
+
+- Improve `toErrorList()` and `unwrapErrorHandler()` by ensuring objects before recursing
+
 ## Dev / docs / playground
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 5.4.0
```
nickgros, line 19: If just releasing this change, it could probably be 5.3.2

packages/validator-ajv8/src/validator.ts
```diff
@@ -148,7 +151,7 @@ export default class AJV8Validator<T = any, S extends StrictRJSFSchema = RJSFSch
         return { ...acc, [key]: this.createErrorHandler(value) };
       }, handler);
     }
-    if (isObject(formData)) {
+    if (isPlainObject(formData)) {
```
nickgros, line 154: Not clear to me why we need to be more restrictive with `isPlainObject` instead of `isObject`, but I don't think it will be an issue

PR #3537, opened 2023-03-24
Bugfix disable submit button

### LINKED ISSUE #3264: Disabling the form should disable the submit button as well.
### Prerequisites

- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net/n1k0/f2y3fq7L/6/) or a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

4.2.3

### Current Behavior

When I pass the `disabled` props, all inputs are disabled, but the form can still be submitted by pressing the submit button, which is still enabled. 

### Expected Behavior

The submit button should be disabled. 

### Steps To Reproduce

https://codesandbox.io/s/lucid-dawn-k56bdt?file=/src/App.js (4.2.3)

 

### Environment

```markdown
See sandbox
```

### Anything else?

Similar issues: 

https://github.com/rjsf-team/react-jsonschema-form/issues/854 - suggestion - pass in empty div as children  to remove the submit buttton 

Another option is to pass the `uiSchema` prop in with the `ui:submitButtonOptions.props.disabled` when you are trying to disable the form. 

P.S Fantastic work with this library. This thing is great.

dwjohnston:
Fixes #3264 
Also fixes an issue with the Playground where the 'Disable whole form' button does not work. 

The disabled button  can be controlled via the `ui:submitButtonProps` properties, or the top level `disabled` property.  The logic I've gone for is a simple OR operator between these values. You could argue that the `ui:submitButtonProps` should always take precedence, even if set to false, but I can't think of a scenario where that makes sense.  

### Checklist


packages/core/src/components/Form.tsx
```diff
@@ -811,7 +812,7 @@ export default class Form<
           disabled={disabled}
           readonly={readonly}
         />
-        {children ? children : <SubmitButton uiSchema={uiSchema} registry={registry} />}
+        {children ? children : <SubmitButton uiSchema={uiSchema} registry={registry} disabled={disabled} />}
```
heath-freenome, line 815: Rather than going through all of this, why not just update the `uiSchema` that gets passed in directly to the `SubmitButton`? Something like:
```
// Add SUBMIT_BTN_OPTIONS_KEY and UI_OPTIONS_KEY to the import for @rjsf/utils
let { [SUBMIT_BTN_OPTIONS_KEY]: submitOptions = {} } = getUiOptions<T, S, F>(uiSchema);
if (disabled) {
  submitOptions = { ...submitOptions, submitButtonProps: { ...submitOptions.submitButtonProps, disable: true } };
}
const submitUiSchema = { [UI_OPTIONS_KEY]: { [SUBMIT_BTN_OPTIONS_KEY]: submitOptions } };
```

Then pass 
```suggestion
        {children ? children : <SubmitButton uiSchema={submitUiSchema} registry={registry} />}
```

It saves a whole bunch of coding and doesn't require updating all of the `SubmitButton` implementations in all themes and changing any types

heath-freenome (changes requested):

heath-freenome:
@dwjohnston Have you had a chance to look over my feedback?

packages/core/src/components/Form.tsx
dwjohnston: Yes, I did consider this, and I think you're right. 

I was looking at doing something here: https://github.com/rjsf-team/react-jsonschema-form/blob/8ee7d472fa7ef9fdd79113aeff9e012e6bdc6078/packages/core/src/components/Form.tsx#L299

What put me off was around the handling of 'what if the uiSchema exists, what if doesn't, what if nested properties exist....' and then also you've got that state.schemaUtils stuff, which I wasn't sure about. 

But yeah, I think you're right, much nicer than some kind of cascading change. I'll make that change.

heath-freenome: Using the `getUiOptions()` function will mitigate lack of a `uiSchema` and the two possibilities of `ui:submitButtonOptions` vs the `"ui:options": { submitButtonOptions }`. SchemaUtils do not deal with the `submitButtonOptions`. Also in the suggestion I gave, it should handle the nested options or not case as well

packages/playground/src/app.jsx
```diff
@@ -17,7 +17,7 @@ const liveSettingsSchema = {
   type: 'object',
   properties: {
     liveValidate: { type: 'boolean', title: 'Live validation' },
-    disable: { type: 'boolean', title: 'Disable whole form' },
+    disabled: { type: 'boolean', title: 'Disable whole form' },
```
heath-freenome, line 20: This file was recently converted to typescript and these contents moved into a different file

packages/playground/src/app.jsx
```diff
@@ -480,6 +480,8 @@ class Playground extends Component {
 
     return (
       <div className='container-fluid'>
+
+        {JSON.stringify(liveSettings)}
```
heath-freenome, line 484: I'm not sure you meant to leave this in here? It seems odd to render them as a string

packages/core/src/components/Form.tsx
```diff
@@ -296,7 +296,20 @@ export default class Form<
...
+          ...rawUiSchema,
+          'ui:submitButtonOptions': {
+            ...rawUiSchema['ui:submitButtonOptions'],
+            props: {
+              ...rawUiSchema['ui:submitButtonOptions']?.props,
+              disabled: true,
+            },
+          },
+        }
+      : rawUiSchema;
```
heath-freenome, line 312: There is one problem with this approach. `ui:submitButtonOptions` could also be defined as `ui:options: { submitButtonOptions } `; The approach I suggested of providing a separately constructed `uiSchema` just to the `SubmitButton` handles either situation.

I would revert this whole change since the only use for `submitButtonOptions` really is the `SubmitButton` and move the code down in the the render and pass a one-off set of props just to the `SubmitButton`

packages/core/src/components/templates/ButtonTemplates/SubmitButton.tsx
```diff
@@ -11,6 +11,7 @@ export default function SubmitButton<
   if (norender) {
     return null;
   }
+
```
heath-freenome, line 14: I would suggest just reverting this change since it really isn't necessary to fixing the bug

packages/core/src/components/Form.tsx
```diff
@@ -815,6 +829,7 @@ export default class Form<
           disabled={disabled}
           readonly={readonly}
         />
+
```
heath-freenome, line 832: Something like:
```suggestion
// Add SUBMIT_BTN_OPTIONS_KEY and UI_OPTIONS_KEY to the import for @rjsf/utils
let { [SUBMIT_BTN_OPTIONS_KEY]: submitOptions = {} } = getUiOptions<T, S, F>(uiSchema);
if (disabled) {
  submitOptions = { ...submitOptions, submitButtonProps: { ...submitOptions.submitButtonProps, disable: true } };
}
const submitUiSchema = { [UI_OPTIONS_KEY]: { [SUBMIT_BTN_OPTIONS_KEY]: submitOptions } };
```

Then pass 
```
        {children ? children : <SubmitButton uiSchema={submitUiSchema} registry={registry} />}
```

packages/playground/src/components/Playground.tsx
```diff
@@ -32,7 +32,7 @@ const liveSettingsSchema: RJSFSchema = {
   type: 'object',
   properties: {
     liveValidate: { type: 'boolean', title: 'Live validation' },
-    disable: { type: 'boolean', title: 'Disable whole form' },
+    disabled: { type: 'boolean', title: 'Disable whole form' },
```
heath-freenome, line 35: This got fixed in another PR

CHANGELOG.md
```diff
@@ -60,7 +61,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Converted the `playground` to use Typescript, including some refactoring of code to make that job easier
 - Updated the `custom-templates` documentation to add `errorSchema` to the props for `ObjectFieldTemplate`
-
+- Fixes bug where the 'Disable whole form' button did not work.
```
heath-freenome, line 64: You can also revert this as it got fixed in another PR

packages/core/src/components/Form.tsx
```diff
@@ -350,6 +352,7 @@ export default class Form<
       props.idPrefix,
       props.idSeparator
     );
+
     const nextState: FormState<T, S, F> = {
```
heath-freenome, line 356: Let's just restore this back to what it was
```suggestion
    );
    const nextState: FormState<T, S, F> = {
```

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Updated `FileWidget` to pass false for `required` once a value has been specified, fixing [#3504](https://github.com/rjsf-team/react-jsonschema-form/issues/3504)
 - Updated `ObjectField` to pass the `errorSchema` to the `ObjectFieldTemplate` to allow custom templates access to the errors
 - Centralized snapshot tests from each theme into `core`, adding snapshots tests for `core` as well
+- Updated `Form` to include the top `disabled` property in the `ui:submitButtonOptions` so the submit button will be disabled when the whole form is disabled. Fixes [#3264](https://github.com/rjsf-team/react-jsonschema-form/issues/3264).
```
heath-freenome, line 53: Can you move this to be under the 5.5.1 section now?

heath-freenome (approved):

PR #3545, opened 2023-03-27
fix: Got playground working after some issue were introduced

heath-freenome:

The playground had some issues with not properly loading  urls or changing options
- Updated most of the `React.FC` components to stateless functional components
- Also removed the `code` state in the Editor as it was not updating based on outside prop changes
- Added more types where necessary
- Ensured that the load also saves the schema and formdata and only runs once by introducing a new state variable

### Checklist

### DIFF
--- a/packages/playground/src/components/Editor.tsx
+++ b/packages/playground/src/components/Editor.tsx
@@ -1,4 +1,4 @@
-import { memo, useCallback, useState } from 'react';
+import { useCallback, useState } from 'react';
 import MonacoEditor from '@monaco-editor/react';
 
 const monacoEditorOptions = {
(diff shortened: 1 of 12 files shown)


packages/playground/src/components/Editor.tsx
```diff
@@ -8,50 +8,49 @@ const monacoEditorOptions = {
...
 
-    const icon = valid ? 'ok' : 'remove';
-    const cls = valid ? 'valid' : 'invalid';
+      try {
+        const parsedCode = JSON.parse(code);
+        setValid(true);
+        onChange(parsedCode);
+      } catch (err) {
+        setValid(false);
+      }
```
nickgros, line 32: I would think you would want to set state on any change (as before), or else code won't update while the JSON is invalid? 

Or does the editor have a hidden/local state value that we don't need to set?

```suggestion
      try {
        const parsedCode = JSON.parse(code);
        setValid(true);
      } catch (err) {
        setValid(false);
      } finally {
        onChange(parsedCode);
      }
```

nickgros (approved):
Just one question about how this works

PR #3566, opened 2023-04-01
chore(playground): Extract Header and Editors and small improvements

ahkhanjani:

Buildup for future Next.js migration.

### Checklist


packages/playground/src/components/Header.tsx
```diff
@@ -0,0 +1,194 @@
+import { useCallback } from 'react';
+
+import Form, { IChangeEvent } from '@rjsf/core';
```
heath-freenome, line 3:
```suggestion
import { useCallback } from 'react';
import Form, { IChangeEvent } from '@rjsf/core';
```

packages/playground/src/components/Editors.tsx
```diff
@@ -0,0 +1,147 @@
+import { useCallback, useState } from 'react';
+import MonacoEditor from '@monaco-editor/react';
+import isEqualWith from 'lodash/isEqualWith';
+
+import type { ErrorSchema, RJSFSchema, UiSchema } from '@rjsf/utils';
```
heath-freenome, line 5: You don't really need the `type` for importing also this puts all of the package-level imports in alphabetical order
```suggestion
import MonacoEditor from '@monaco-editor/react';
import { ErrorSchema, RJSFSchema, UiSchema } from '@rjsf/utils';
import isEqualWith from 'lodash/isEqualWith';
```

packages/playground/src/components/Playground.tsx
```diff
@@ -1,7 +1,7 @@
 import { useCallback, useState, useRef, useEffect, ComponentType, FormEvent } from 'react';
 import 'react-app-polyfill/ie11';
-import isEqualWith from 'lodash/isEqualWith';
-import Form, { withTheme, IChangeEvent, FormProps } from '@rjsf/core';
+
+import { withTheme, IChangeEvent, FormProps } from '@rjsf/core';
```
heath-freenome, line 4:
```suggestion
import { withTheme, IChangeEvent, FormProps } from '@rjsf/core';
```

packages/playground/src/components/Playground.tsx
```diff
@@ -11,129 +11,18 @@ import {
...
-      )}
-    </div>
-  );
-}
-
-interface LiveSettings {
+export interface LiveSettings {
   showErrorList: false | 'top' | 'bottom';
   [key: string]: any;
 }
```
heath-freenome, line 28: I would strongly suggest moving this into the `Header` rather than leaving it here because you are creating a circular import reference between the two files that might cause future issues.

```suggestion
```

packages/playground/src/components/Header.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { RJSFSchema, UiSchema, ValidatorType } from '@rjsf/utils';
+import localValidator from '@rjsf/validator-ajv8';
+
+import CopyLink from './CopyLink';
+import ThemeSelector, { ThemesType } from './ThemeSelector';
+import Selector from './Selector';
+import ValidatorSelector from './ValidatorSelector';
+import SubthemeSelector from './SubthemeSelector';
+import RawValidatorTest from './RawValidatorTest';
+import { LiveSettings } from './Playground';
```
heath-freenome, line 13: Rather than importing `LiveSettings` export it from here instead. It prevents a circular import reference between `Playground` and `Header`
```suggestion

export interface LiveSettings {
  showErrorList: false | 'top' | 'bottom';
  [key: string]: any;
}
```

packages/playground/src/components/Playground.tsx
```diff
@@ -11,129 +11,18 @@ import {
   UiSchema,
   ValidatorType,
 } from '@rjsf/utils';
-import localValidator from '@rjsf/validator-ajv8';
 
 import { samples } from '../samples';
+import Header from './Header';
```
heath-freenome, line 16: Importing `LiveSettings` when exported from `Header` prevents a circular import
```suggestion
import Header, { LiveSettings } from './Header';
```

heath-freenome (changes requested):
Overall a nice cleanup! Just a few recommendations to avoid circular imports

ahkhanjani:
@heath-freenome The pattern of multiple components per page that we're using here is: declaring children first and parent last. Even though this is how it should be the order in declaring normal functions, in my personal opinion since in React the props flow like parent -> child, it makes more sense to declare the parent first. For example in Editors.tsx, as the file name indicates, the main component here is Editors, so we would expect to see it in the beginning of the file; and as we read and reach the Editor child component, we would see them right after Editors. What's your comment on this?

packages/playground/src/components/Header.tsx
```diff
@@ -0,0 +1,197 @@
...
+import ThemeSelector, { ThemesType } from './ThemeSelector';
+import Selector from './Selector';
+import ValidatorSelector from './ValidatorSelector';
+import SubthemeSelector from './SubthemeSelector';
+import RawValidatorTest from './RawValidatorTest';
+
+export interface LiveSettings {
+  showErrorList: false | 'top' | 'bottom';
+  [key: string]: any;
+}
```
ahkhanjani, line 16: Should we move this right before HeaderProps where it's being used?

heath-freenome: That is a personal preference. I just put it where I put it for the convenience of making a single suggestion

heath-freenome:
> @heath-freenome The pattern of multiple components per page that we're using here is: declaring children first and parent last. Even though this is how it should be the order in declaring normal functions, in my personal opinion since in React the props flow like parent -> child, it makes more sense to declare the parent first. For example in Editors.tsx, as the file name indicates, the main component here is Editors, so we would expect to see it in the beginning of the file; and as we read and reach the Editor child component, we would see them right after Editors. What's your comment on this?

While it is a minor thing, javascript tends to like defining things before they are used. The general pattern in most of the rest of the RJSF library is to do child first.

heath-freenome (approved):
Overall a nice cleanup! Just a few recommendations to avoid circular imports

ahkhanjani:
@heath-freenome I see you've removed most memoizations. May I know the reason?

heath-freenome:
> @heath-freenome I see you've removed most memoizations. May I know the reason?

There were issues related to the playground not updating when clicking on different links and theme changes. So I just removed all of the ones that were potentially causing issues. Once I got things working I didn't bother adding them back in since, honestly, the performance benefits didn't seem worth the extra effort of trying to figure out which ones were the problem. You are welcome to restore the ones that don't cause the UI to become stuck like that again.

packages/playground/src/components/Playground.tsx
```diff
@@ -264,94 +125,33 @@ export default function Playground({ themes, validators }: PlaygroundProps) {
...
-    } catch (error) {
-      setShareURL(null);
-      console.error(error);
-    }
-  }, [formData, liveSettings, schema, theme, uiSchema]);
-
   const templateProps: Partial<TemplatesType> = {
     ArrayFieldTemplate,
     ObjectFieldTemplate,
   };
```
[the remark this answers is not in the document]
heath-freenome: Some of the examples have their own template implementations and will likely break if you remove them

packages/playground/src/components/Header.tsx
```diff
@@ -67,6 +48,25 @@ function HeaderButtons({ playGroundFormRef }: { playGroundFormRef: React.Mutable
   );
 }
 
+const liveSettingsSchema: RJSFSchema = {
+  type: 'object',
+  properties: {
+    liveValidate: { type: 'boolean', title: 'Live validation' },
+    disable: { type: 'boolean', title: 'Disable whole form' },
```
heath-freenome, line 55: Can you change this to `disabled` since that is the property value in the `Form`

heath-freenome (changes requested):

packages/playground/src/components/Playground.tsx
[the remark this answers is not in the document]
ahkhanjani: How is that different from passing them directly as props? And if it will break then should we at least wrap it in `useMemo`?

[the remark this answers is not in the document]
heath-freenome: Oh whoops, I misunderstood your question. Sorry. Doing what you suggest and passing them in directly makes sense

heath-freenome:
@ahkhanjani When you are done making changes and are ready for the final review, then re-request a review

packages/playground/src/components/Playground.tsx
```diff
@@ -1,7 +1,6 @@
 import { useCallback, useState, useRef, useEffect, ComponentType, FormEvent } from 'react';
 import 'react-app-polyfill/ie11';
```
ahkhanjani, line 2:
```suggestion
```

Should we remove IE11 polyfill as it's reached its end of life?

packages/playground/src/components/Playground.tsx
```diff
@@ -389,8 +183,9 @@ export default function Playground({ themes, validators }: PlaygroundProps) {
               theme={theme}
             >
               <FormComponent
-                {...templateProps}
                 {...liveSettings}
+                ArrayFieldTemplate={ArrayFieldTemplate}
+                ObjectFieldTemplate={ObjectFieldTemplate}
```
heath-freenome, line 188: Dang, I totally missed this, twice. Once while I was doing v5 upgrades and again just a bit earlier. The `ArrayFieldTemplate` and `ObjectFieldTemplate` are no longer direct props on `Form` instead they need to be passed via `templates`
```suggestion
                templates={{ ArrayFieldTemplate, ObjectFieldTemplate }}
```

heath-freenome (changes requested):

packages/playground/src/components/Playground.tsx
ahkhanjani: Haha yes. Then I will take responsibility and make this sacrifice ✊

ahkhanjani:
@heath-freenome If the build is successful feel free to merge. I will try to upgrade to React 18 in another PR

heath-freenome (approved):

PR #3575, opened 2023-04-04
fix: Switch to using isValid() for oneOf schema detection in retrieveSchema()

heath-freenome:

Added an optimization in `retrieveSchema()` to use the `isValid()` function rather than `validateFormData()`

- In `@rjsf/utils`, switched the `withExactlyOneSubschema()` to use `isValid()`
  - Updated the test mocks accordingly

### Checklist

### DIFF
--- a/packages/core/test/Form.test.jsx
+++ b/packages/core/test/Form.test.jsx
@@ -209,7 +209,6 @@ describeRepeated('Form common', (createFormComponent) => {
 
     it('should work with oneOf', function () {
       const schema = {
-        $schema: 'http://json-schema.org/draft-06/schema#',
         type: 'object',
         properties: {
           connector: {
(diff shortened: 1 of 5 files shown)


packages/core/test/Form.test.jsx
```diff
@@ -209,7 +209,6 @@ describeRepeated('Form common', (createFormComponent) => {
 
     it('should work with oneOf', function () {
       const schema = {
-        $schema: 'http://json-schema.org/draft-06/schema#',
```
heath-freenome, line 212: This caused compilation errors that broke the `isValid()` checks but didn't affect the old `validateFormData()` checks because that handled the error differently

nickgros (approved):

PR #3586, opened 2023-04-06
feature: Refactored common utility functions from validators into utils

heath-freenome:

There were a bunch of utility functions in the two validator implementations that were duplicated
- In `@rjsf/utils`, refactored the common functions as follows:
  - Refactored the `createErrorHandler()`, `toErrorList()`, `toErrorSchema()` and `unwrapErrorHandler()` functions from the `@rjsf/validator-ajv6` and `@rjsf/validator-ajv8` implementations since they were identical - As a result, the `mergeValidationData()` function was deprecated in favor of the new `validationDataMerge()` function that uses the refactored `toErrorList()` function - Refactored the `ROOT_SCHEMA_PREFIX` constant as well
  - Updated `ValidatorType` and `SchemaUtilsType` to deprecate the `toErrorList()` and `mergeValidationData()` functions, respectively
- In `@rjsf/validator-ajv6` and `@rjsf/validator-ajv8`, updated the validator implementations to use the new refactored functions
- In `@rjsf/core` updated `Form` to use the `validationDataMerge()` function instead of the `schemaUtils.mergeValidationData()` function
- In the `utility-functions` documentation, described the new refactored functions and deprecated `mergeValidationData()`

### Checklist

### DIFF
--- a/packages/utils/src/createErrorHandler.ts
+++ b/packages/utils/src/createErrorHandler.ts
@@ -1,0 +1,33 @@
+import isPlainObject from 'lodash/isPlainObject';
+
+import { ERRORS_KEY } from './constants';
+import { FieldValidation, FormValidation, GenericObjectType } from './types';
+
+/** Given a `formData` object, recursively creates a `FormValidation` error handling structure around it
+ *
+ * @param formData - The form data around which the error handler is created
+ * @returns - A `FormValidation` object based on the `formData` structure
+ */
+export default function createErrorHandler<T = any>(formData: T): FormValidation<T> {
+  const handler: FieldValidation = {
+    // We store the list of errors for this node in a property named __errors
+    // to avoid name collision with a possible sub schema field named
+    // 'errors' (see `utils.toErrorSchema`).
+    [ERRORS_KEY]: [],
+    addError(message: string) {
+      this[ERRORS_KEY]!.push(message);
+    },
+  };
+  if (Array.isArray(formData)) {
+    return formData.reduce((acc, value, key) => {
+      return { ...acc, [key]: createErrorHandler(value) };
+    }, handler);
+  }
+  if (isPlainObject(formData)) {
+    const formObject: GenericObjectType = formData as GenericObjectType;
+    return Object.keys(formObject).reduce((acc, key) => {
+      return { ...acc, [key]: createErrorHandler(formObject[key]) };
+    }, handler as FormValidation<T>);
+  }
+  return handler as FormValidation<T>;
+}
--- a/packages/utils/src/types.ts
+++ b/packages/utils/src/types.ts
@@ -917,6 +917,8 @@ export interface ValidatorType<T = any, S extends StrictRJSFSchema = RJSFSchema,
    *
    * @param errorSchema - The `ErrorSchema` instance to convert
    * @param [fieldPath=[]] - The current field path, defaults to [] if not specified
+   * @deprecated - Use the `toErrorList()` function provided by `@rjsf/utils` instead. This function will be removed in
+   *        the next major release.
    */
   toErrorList(errorSchema?: ErrorSchema<T>, fieldPath?: string[]): RJSFValidationError[];
   /** Validates data against a schema, returning true if the data is valid, or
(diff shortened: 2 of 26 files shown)


packages/utils/src/createErrorHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { ERRORS_KEY } from './constants';
+import { FieldValidation, FormValidation, GenericObjectType } from './types';
+
+/** Given a `formData` object, recursively creates a `FormValidation` error handling structure around it
+ *
+ * @param formData - The form data around which the error handler is created
+ * @returns - A `FormValidation` object based on the `formData` structure
+ */
+export default function createErrorHandler<T = any>(formData: T): FormValidation<T> {
```
heath-freenome, line 11: @nickgros I am debating whether I rename this to something like `wrapWithErrorHandler()` so that it is an analog to the `unwrap` function.

packages/utils/src/unwrapErrorHandler.ts
```diff
@@ -0,0 +1,25 @@
+import isPlainObject from 'lodash/isPlainObject';
+
+import { ErrorSchema, FormValidation, GenericObjectType } from './types';
+
+/** Unwraps the `errorHandler` structure into the associated `ErrorSchema`, stripping the `addError()` functions from it
+ *
+ * @param errorHandler - The `FormValidation` error handling structure
+ * @returns - The `ErrorSchema` resulting from the stripping of the `addError()` function
+ */
+export default function unwrapErrorHandler<T = any>(errorHandler: FormValidation<T>): ErrorSchema<T> {
```
heath-freenome, line 10: If I were to rename the `createErrorHandler()` to `wrapWithErrorHandler()` then I wonder if calling this `unwrapFromErrorHandler()` is a bit better? or perhaps this could want to be renamed to be the analog of `create`... like `destroyErrorHandler()`? Hmmm, naming things is sometimes so challenging.

packages/utils/src/validationDataMerge.ts
```diff
@@ -0,0 +1,31 @@
...
+/** Merges the errors in `additionalErrorSchema` into the existing `validationData` by combining the hierarchies in the
+ * two `ErrorSchema`s and then appending the error list from the `additionalErrorSchema` obtained by calling
+ * `toErrorList()` on the `errors` in the `validationData`. If no `additionalErrorSchema` is passed, then
+ * `validationData` is returned.
+ *
+ * @param validationData - The current `ValidationData` into which to merge the additional errors
+ * @param [additionalErrorSchema] - The optional additional set of errors in an `ErrorSchema`
+ * @returns - The `validationData` with the additional errors from `additionalErrorSchema` merged into it, if provided.
+ */
+export default function validationDataMerge<T = any>(
```
heath-freenome, line 16: I couldn't just remove the `ValidatorType` parameter from `mergeValidationData()` so I figured a little reshuffle of words. Maybe `validationDataMerging()` or `validationDataMerger()` is better?

packages/utils/src/unwrapErrorHandler.ts
nickgros: I don't think this function needs to have an "opposite" name since it's not a true inverse (create is `TFormData -> FormValidation<T>` and unwrap is `FormValidation<T> -> ErrorSchema<T>`). I'm actually ok with the current names, but not opposed to changing them if you think it's easier to understand.

packages/utils/test/toErrorList.test.ts
```diff
@@ -0,0 +1,11 @@
+import { toErrorList } from '../src';
+import { TEST_ERROR_LIST, TEST_ERROR_SCHEMA } from './testUtils/testData';
+
+describe('toErrorList()', () => {
+  it('returns empty array when nothing is passed', () => {
+    expect(toErrorList()).toEqual([]);
+  });
+  it('', () => {
```
nickgros, line 8: Should have a test name even if it's basic "validate examples"

nickgros (dismissed):

packages/utils/test/toErrorList.test.ts
heath-freenome: Oops! Good catch

packages/utils/src/types.ts
```diff
@@ -1046,6 +1048,8 @@ export interface SchemaUtilsType<T = any, S extends StrictRJSFSchema = RJSFSchem
    * @param schema - The schema for which retrieving a schema is desired
    * @param [formData] - The current formData, if any, to assist retrieving a schema
    * @returns - The schema having its conditions, additional properties, references and dependencies resolved
+   * @deprecated - Use the `validationDataMerge()` function exported from `@rjsf/utils` instead. This function will be
+   *        removed in the next major release.
```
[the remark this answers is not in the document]
heath-freenome: Another good catch!

nickgros (approved):

PR #3590, opened 2023-04-08
Add option to provide a callback function to focusOnFirstError

### LINKED ISSUE #3580: Allow custom focusOnError
### Prerequisites

### What theme are you using?

core

Our forms sometimes have disabled fields that are automatically filled by doing some type of action.
Example: Address fields that are readonly but are filled if you search and select the adress, this will then automatically fill in the fields. The adress fields are required because it wouldn't make sense to have the search field required.

The problem here is that when i use `focusOnFirstError` it wont trigger on fields that are disabled or readonly.
https://github.com/rjsf-team/react-jsonschema-form/blob/ff46301dbbeb951758c5ee1b2ec0f95a459b4dfe/packages/core/src/components/Form.tsx#L691-L703

This might be a very niche issue but it's quite important from a UX standpoint.

There are different solutions that would work.

1. Let us provide our own callback for `focusOnError`, this would give the most freedom.
2. Add a check if the element is disabled/readonly and instead call `element.scrollIntoViewIfNeeded(true)` might have to use `element.scrollIntoView()` since support for IfNeeded is not complete.

_No response_

Fronix:
fixes #3580 

Focusing the first error using `focusOnFirstError` requires an actual element to be focusable. This is not always possible, you might have disabled fields that are filled by some type of action that the user has missed.

This change provides the option to add a custom callback function to `focusOnFirstError`, the function recieves a `RJSFValidationError` in which you can handle what happens then `focusOnFirstError` is called.

Example:
```tsx
import { RJSFSchema, RJSFValidationError } from '@rjsf/utils';

const schema: RJSFSchema = {
  type: 'string',
};

const focusOnError = (error: RJSFValidationError) => {
  console.log('I need to handle focusing this error');
};

render(<Form schema={schema} focusOnFirstError={focusOnError} />, document.getElementById('app'));
```

### Checklist

- [X] **I'm updating documentation**
  - [X] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
  - [X] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed

### DIFF
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -173,7 +173,7 @@ export interface FormProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
   transformErrors?: ErrorTransformer<T, S, F>;
   /** If set to true, then the first field with an error will receive the focus when the form is submitted with errors
    */
-  focusOnFirstError?: boolean;
+  focusOnFirstError?: boolean | ((error: RJSFValidationError) => void);
   /** Optional string translation function, if provided, allows users to change the translation of the RJSF internal
    * strings. Some strings contain replaceable parameter values as indicated by `%1`, `%2`, etc. The number after the
    * `%` indicates the order of the parameter. The ordering of parameters is important because some languages may choose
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Form.tsx
```diff
@@ -725,7 +725,11 @@ export default class Form<
         errors = merged.errors;
       }
       if (focusOnFirstError) {
-        this.focusOnError(schemaValidation.errors[0]);
+        if (typeof focusOnFirstError === 'function') {
+          focusOnFirstError(schemaValidation.errors[0]);
```
Fronix, line 729: Not sure but it could be better to just provide the entire error array here. Though this would change the definition of `onFirstError`.

heath-freenome: yes, that would change the definition.

heath-freenome (approved):

PR #3604, opened 2023-04-13
Implement feature-flagged support for ignoring of optional array fields with `minItems` set

### LINKED ISSUE #3363: Conflict when using an optional array with minItems set
### Prerequisites

### What theme are you using?

material-ui

### Version

5.x

### Current Behavior

Hi and thanks for this great lib 🙌

When setting an optional property in an `"object"`, which is of type `"array"` with a `"minItem"` set to 1 or more, it's not possible to remove the property from the said object.

By default, the array property widget appears with items that can be empty, which can cause bugs.

### Expected Behavior

I don't have any design proposal, I know it's tricky to represent, but the default items should only appear when deciding that the property is not undefined somehow.

### Steps To Reproduce

```
const mySchema = {
  type: "object,
  properties: {
    optionalArray: {
      type: "array",
      items: { type: "string" },
      minItems: 2
    }
  }
}
```

### Environment

```markdown
- OS: Any
- Node: Any
- npm: Any
```

### Anything else?

_No response_

RPiAwesomeness:

Currently, if `minItems` is set on an optional field that field will show up as an empty array in the form data if no values have been entered by the user. This is an area of ambiguity in the spec and some users want this behavior while others don't. Thus, it should be possible to choose whether you want the field to be populated as an empty array or to be ignored and thus have the key non-existent in the form output if no items have been added.

- fixes #3363
- fixes unreported bug where live settings in the Playground aren't being saved/loaded

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,11 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
-# 5.7.0
+# 5.8.0
 
 ## @rjsf/core
 
-- Added new `defaultFormStateBehavior` prop to `Form` to specify alternate behavior when dealing with the rendering of array fields where `minItems` is set but field is not `required` (fixes [3363](https://github.com/rjsf-team/react-jsonschema-form/issues/3363))
+- Added new `experimental_defaultFormStateBehavior` prop to `Form` to specify alternate behavior when dealing with the rendering of array fields where `minItems` is set but field is not `required` (fixes [3363](https://github.com/rjsf-team/react-jsonschema-form/issues/3363)) (PR [3604](https://github.com/rjsf-team/react-jsonschema-form/pull/3604))
 
 ## @rjsf/utils
 
@@ -27,7 +27,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## Dev / docs / playground
 
-- Added **Default Form State Behavior** live setting
+- Added **minItems behavior for array field** live setting
 
 # 5.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,19 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
+# 5.7.0
+
+## @rjsf/core
+
+- Added new `defaultFormStateBehavior` prop to `Form` to specify alternate behavior when dealing with the rendering of array fields where `minItems` is set but field is not `required` (fixes [3363](https://github.com/rjsf-team/react-jsonschema-form/issues/3363))
+
+## @rjsf/utils
+
+- Refactored optional parameters for `computeDefaults()` into destructured props object to reduce clutter when only specifying later of the optional argument
+
+## Dev / docs / playground
+
+- Added **Default Form State Behavior** live setting
 
 # 5.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,19 +15,6 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
-# 5.8.0
-
-## @rjsf/core
-
-- Added new `experimental_defaultFormStateBehavior` prop to `Form` to specify alternate behavior when dealing with the rendering of array fields where `minItems` is set but field is not `required` (fixes [3363](https://github.com/rjsf-team/react-jsonschema-form/issues/3363)) (PR [3604](https://github.com/rjsf-team/react-jsonschema-form/pull/3604))
-
-## @rjsf/utils
-
-- Refactored optional parameters for `computeDefaults()` into destructured props object to reduce clutter when only specifying later of the optional argument
-
-## Dev / docs / playground
-
-- Added **minItems behavior for array field** live setting
 
 # 5.7.0
 
@@ -39,6 +26,8 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Updated the `MultiSchemaField` to use the new `getDiscriminatorFieldFromSchema()` API
 
+- Added new `experimental_defaultFormStateBehavior` prop to `Form` to specify alternate behavior when dealing with the rendering of array fields where `minItems` is set but field is not `required` (fixes [3363](https://github.com/rjsf-team/react-jsonschema-form/issues/3363)) ([3604](https://github.com/rjsf-team/react-jsonschema-form/pull/3604))
+
 ## @rjsf/fluent-ui
 
 - Added support for `additionalProperties` to fluent-ui theme, fixing [#2777](https://github.com/rjsf-team/react-jsonschema-form/issues/2777).
@@ -48,11 +37,13 @@ should change the heading of the (upcoming) version to include a major version b
 - Added two new APIs `getDiscriminatorFieldFromSchema()` (a refactor of code from `MultiSchemaField`) and `hashForSchema()`
   - Updated `getDefaultFormState()` and `toPathSchema()` to use `getDiscriminatorFieldFromSchema()` to provide a discriminator field to `getClosestMatchingOption()` calls.
 - Refactored the `retrieveSchema()` internal API functions to support implementing an internal `schemaParser()` API for use in precompiling schemas, in support of [#3543](https://github.com/rjsf-team/react-jsonschema-form/issues/3543)
+- Refactored optional parameters for `computeDefaults()` into destructured props object to reduce clutter when only specifying later of the optional argument
 - Fixed `toPathSchema()` to handle `properties` in an object along with `anyOf`/`oneOf`, fixing [#3628](https://github.com/rjsf-team/react-jsonschema-form/issues/3628) and [#1628](https://github.com/rjsf-team/react-jsonschema-form/issues/1628)
 
 ## Dev / docs / playground
 
 - Added documentation to `custom-templates` describing how to extend the `BaseInputTemplate`
+- Added **minItems behavior for array field** live setting ([3604](https://github.com/rjsf-team/react-jsonschema-form/pull/3604))
 
 # 5.6.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,8 +37,8 @@ should change the heading of the (upcoming) version to include a major version b
 - Added two new APIs `getDiscriminatorFieldFromSchema()` (a refactor of code from `MultiSchemaField`) and `hashForSchema()`
   - Updated `getDefaultFormState()` and `toPathSchema()` to use `getDiscriminatorFieldFromSchema()` to provide a discriminator field to `getClosestMatchingOption()` calls.
 - Refactored the `retrieveSchema()` internal API functions to support implementing an internal `schemaParser()` API for use in precompiling schemas, in support of [#3543](https://github.com/rjsf-team/react-jsonschema-form/issues/3543)
-- Refactored optional parameters for `computeDefaults()` into destructured props object to reduce clutter when only specifying later of the optional argument
 - Fixed `toPathSchema()` to handle `properties` in an object along with `anyOf`/`oneOf`, fixing [#3628](https://github.com/rjsf-team/react-jsonschema-form/issues/3628) and [#1628](https://github.com/rjsf-team/react-jsonschema-form/issues/1628)
+- Refactored optional parameters for `computeDefaults()` into destructured props object to reduce clutter when only specifying later of the optional argument ([3604](https://github.com/rjsf-team/react-jsonschema-form/pull/3604))
 
 ## Dev / docs / playground
 
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -182,10 +182,9 @@ export interface FormProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
    * to put the second parameter before the first in its translation.
    */
   translateString?: Registry['translateString'];
-  /** Optional feature flags, if provided, specify different behavior that can be enabled. This enables differing behavior within
-   * the library for differing interpreations of edge cases without breaking existing behavior.
-   */
-  featureFlags?: { defaultFormStateBehavior?: number };
+  /** Optional bit flag, if provided, allows users to override default form state behavior.
+   * Currently only affecting minItems on array fields */
+  defaultFormStateBehavior?: number;
   // Private
   /**
    * _internalFormWrapper is currently used by the semantic-ui theme to provide a custom wrapper around `<Form />`
@@ -311,9 +310,8 @@ export default class Form<
     const mustValidate = edit && !props.noValidate && liveValidate;
     const rootSchema = schema;
     const behaviorBitFlags =
-      ('behaviorBitFlags' in props
-        ? props.featureFlags?.defaultFormStateBehavior
-        : this.props.featureFlags?.defaultFormStateBehavior) || DefaultFormStateBehavior.Legacy_PopulateMinItems;
+      ('behaviorBitFlags' in props ? props.defaultFormStateBehavior : this.props.defaultFormStateBehavior) ||
+      DefaultFormStateBehavior.Legacy_PopulateMinItems;
     let schemaUtils: SchemaUtilsType<T, S, F> = state.schemaUtils;
     if (!schemaUtils || schemaUtils.doesSchemaUtilsDiffer(props.validator, rootSchema, behaviorBitFlags)) {
       schemaUtils = createSchemaUtils<T, S, F>(props.validator, rootSchema, behaviorBitFlags);
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -315,7 +315,7 @@ export default class Form<
         ? props.featureFlags?.defaultFormStateBehavior
         : this.props.featureFlags?.defaultFormStateBehavior) || DefaultFormStateBehavior.Legacy_PopulateMinItems;
     let schemaUtils: SchemaUtilsType<T, S, F> = state.schemaUtils;
-    if (!schemaUtils || schemaUtils.doesSchemaUtilsDiffer(props.validator, rootSchema)) {
+    if (!schemaUtils || schemaUtils.doesSchemaUtilsDiffer(props.validator, rootSchema, behaviorBitFlags)) {
       schemaUtils = createSchemaUtils<T, S, F>(props.validator, rootSchema, behaviorBitFlags);
     }
     const formData: T = schemaUtils.getDefaultFormState(schema, inputFormData) as T;
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -309,12 +309,12 @@ export default class Form<
     const liveValidate = 'liveValidate' in props ? props.liveValidate : this.props.liveValidate;
     const mustValidate = edit && !props.noValidate && liveValidate;
     const rootSchema = schema;
-    const behaviorBitFlags =
-      ('behaviorBitFlags' in props ? props.defaultFormStateBehavior : this.props.defaultFormStateBehavior) ||
+    const defaultFormStateBehavior =
+      ('defaultFormStateBehavior' in props ? props.defaultFormStateBehavior : this.props.defaultFormStateBehavior) ||
       DefaultFormStateBehavior.Legacy_PopulateMinItems;
     let schemaUtils: SchemaUtilsType<T, S, F> = state.schemaUtils;
-    if (!schemaUtils || schemaUtils.doesSchemaUtilsDiffer(props.validator, rootSchema, behaviorBitFlags)) {
-      schemaUtils = createSchemaUtils<T, S, F>(props.validator, rootSchema, behaviorBitFlags);
+    if (!schemaUtils || schemaUtils.doesSchemaUtilsDiffer(props.validator, rootSchema, defaultFormStateBehavior)) {
+      schemaUtils = createSchemaUtils<T, S, F>(props.validator, rootSchema, defaultFormStateBehavior);
     }
     const formData: T = schemaUtils.getDefaultFormState(schema, inputFormData) as T;
     const retrievedSchema = schemaUtils.retrieveSchema(schema, formData);
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -3,7 +3,6 @@ import {
   createSchemaUtils,
   CustomValidator,
   deepEquals,
-  DefaultFormStateBehavior,
   ErrorSchema,
   ErrorTransformer,
   FormContextType,
@@ -33,6 +32,7 @@ import {
   ValidationData,
   validationDataMerge,
   ValidatorType,
+  Experimental_DefaultFormStateBehavior,
 } from '@rjsf/utils';
 import _get from 'lodash/get';
 import _isEmpty from 'lodash/isEmpty';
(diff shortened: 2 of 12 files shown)


RPiAwesomeness:
Sorry about the commit mess. Wanted to make sure my commits were being verified properly.

CHANGELOG.md
```diff
@@ -16,6 +16,20 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 5.6.1
```
heath-freenome, line 19: You are adding a new feature so let's bump the version
```suggestion
# 5.7.0
```

CHANGELOG.md
```diff
@@ -16,6 +16,20 @@ should change the heading of the (upcoming) version to include a major version b
...
+
+## @rjsf/core & @rjsf/utils
```
heath-freenome, line 21: You already mention `@rjsf/util` below
```suggestion
## @rjsf/core
```

packages/core/src/components/Form.tsx
```diff
@@ -305,9 +309,12 @@ export default class Form<
     const liveValidate = 'liveValidate' in props ? props.liveValidate : this.props.liveValidate;
     const mustValidate = edit && !props.noValidate && liveValidate;
     const rootSchema = schema;
+    const behaviorBitFlags =
+      ('behaviorBitFlags' in props ? props.defaultFormStateBehavior : this.props.defaultFormStateBehavior) ||
```
heath-freenome, line 313: I think you missed this when you renamed the prop
```suggestion
      ('defaultFormStateBehavior' in props ? props.defaultFormStateBehavior : this.props.defaultFormStateBehavior) ||
```

heath-freenome (commented):
Sorry I've been busy. This is a partial review. Will dig deeper tomorrow

packages/utils/src/createSchemaUtils.ts
```diff
@@ -27,26 +27,30 @@ import {
...
+ * that one does not have to explicitly pass the `validator`, `rootSchema`, or `defaultFormStateBehavior` to each method.
+ * Since these generally do not change across a `Form`, this allows for providing a simplified set of APIs to the
  * `@rjsf/core` components and the various themes as well. This class implements the `SchemaUtilsType` interface.
  */
 class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>
   implements SchemaUtilsType<T, S, F>
 {
   rootSchema: S;
   validator: ValidatorType<T, S, F>;
+  behaviorBitFlags: number;
```
heath-freenome, line 42:
```suggestion
  defaultFormStateBehavior: number;
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -27,26 +27,30 @@ import {
...
 
   /** Constructs the `SchemaUtils` instance with the given `validator` and `rootSchema` stored as instance variables
    *
    * @param validator - An implementation of the `ValidatorType` interface that will be forwarded to all the APIs
    * @param rootSchema - The root schema that will be forwarded to all the APIs
+   * @param behaviorBitFlags Bitwise flags to set which behavior is chosen for certain edge cases.
```
heath-freenome, line 48:
```suggestion
   * @param [defaultFormStateBehavior] - Optional bit flag, if provided, allows users to override default form state behavior
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -27,26 +27,30 @@ import {
...
    */
-  constructor(validator: ValidatorType<T, S, F>, rootSchema: S) {
+  constructor(validator: ValidatorType<T, S, F>, rootSchema: S, behaviorBitFlags: number) {
     this.rootSchema = rootSchema;
     this.validator = validator;
+    this.behaviorBitFlags = behaviorBitFlags;
```
heath-freenome, line 53:
```suggestion
  constructor(validator: ValidatorType<T, S, F>, rootSchema: S, defaultFormStateBehavior: number) {
    this.rootSchema = rootSchema;
    this.validator = validator;
    this. defaultFormStateBehavior = defaultFormStateBehavior;
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -63,13 +67,18 @@ class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
    *
    * @param validator - An implementation of the `ValidatorType` interface that will be compared against the current one
    * @param rootSchema - The root schema that will be compared against the current one
+   * @param behaviorBitFlags Bitwise flags to set which behavior is chosen for certain edge cases
```
heath-freenome, line 70:
```suggestion
   * @param [defaultFormStateBehavior] - Optional bit flag, if provided, allows users to override default form state behavior
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -63,13 +67,18 @@ class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
...
    * @returns - True if the `SchemaUtilsType` differs from the given `validator` or `rootSchema`
    */
-  doesSchemaUtilsDiffer(validator: ValidatorType<T, S, F>, rootSchema: S): boolean {
+  doesSchemaUtilsDiffer(validator: ValidatorType<T, S, F>, rootSchema: S, behaviorBitFlags: number): boolean {
```
heath-freenome, line 73:
```suggestion
  doesSchemaUtilsDiffer(validator: ValidatorType<T, S, F>, rootSchema: S, defaultFormStateBehavior: number): boolean {
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -63,13 +67,18 @@ class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
...
     if (!validator || !rootSchema) {
       return false;
     }
-    return this.validator !== validator || !deepEquals(this.rootSchema, rootSchema);
+    return (
+      this.validator !== validator ||
+      !deepEquals(this.rootSchema, rootSchema) ||
+      this.behaviorBitFlags !== behaviorBitFlags
```
heath-freenome, line 80:
```suggestion
      this. defaultFormStateBehavior !== defaultFormStateBehavior
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -87,7 +96,14 @@ class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
...
     includeUndefinedValues: boolean | 'excludeObjectChildren' = false
   ): T | T[] | undefined {
-    return getDefaultFormState<T, S, F>(this.validator, schema, formData, this.rootSchema, includeUndefinedValues);
+    return getDefaultFormState<T, S, F>(
+      this.validator,
+      schema,
+      formData,
+      this.rootSchema,
+      includeUndefinedValues,
+      this.behaviorBitFlags
```
heath-freenome, line 105:
```suggestion
      this. defaultFormStateBehavior
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -258,12 +274,17 @@ class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
  *
  * @param validator - an implementation of the `ValidatorType` interface that will be forwarded to all the APIs
  * @param rootSchema - The root schema that will be forwarded to all the APIs
+ * @param [behaviorBitFlags=0] - Optional bitwise flags to set which behavior is chosen for certain edge cases.
```
heath-freenome, line 277:
```suggestion
 * @param [defaultFormStateBehavior=0] - Optional bit flag, if provided, allows users to override default form state behavior
```

packages/utils/src/createSchemaUtils.ts
```diff
@@ -258,12 +274,17 @@ class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
...
   S extends StrictRJSFSchema = RJSFSchema,
   F extends FormContextType = any
->(validator: ValidatorType<T, S, F>, rootSchema: S): SchemaUtilsType<T, S, F> {
-  return new SchemaUtils<T, S, F>(validator, rootSchema);
+>(
+  validator: ValidatorType<T, S, F>,
+  rootSchema: S,
+  behaviorBitFlags = DefaultFormStateBehavior.Legacy_PopulateMinItems
+): SchemaUtilsType<T, S, F> {
+  return new SchemaUtils<T, S, F>(validator, rootSchema, behaviorBitFlags);
```
heath-freenome, line 289:
```suggestion
  defaultFormStateBehavior = DefaultFormStateBehavior.Legacy_PopulateMinItems
): SchemaUtilsType<T, S, F> {
  return new SchemaUtils<T, S, F>(validator, rootSchema, defaultFormStateBehavior);
```

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -92,28 +100,45 @@ function maybeAddDefaultToObject<T = any>(
   }
 }
 
+interface computeDefaultsProps<T = any, S extends StrictRJSFSchema = RJSFSchema> {
+  parentDefaults?: T;
+  rootSchema?: S;
+  rawFormData?: T;
+  includeUndefinedValues?: boolean | 'excludeObjectChildren';
+  _recurseList?: string[];
+  behaviorBitFlags?: number;
```
heath-freenome, line 109:
```suggestion
  defaultFormStateBehavior?: number;
```

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -306,6 +343,7 @@ export function computeDefaults<T = any, S extends StrictRJSFSchema = RJSFSchema
  * @param [includeUndefinedValues=false] - Optional flag, if true, cause undefined values to be added as defaults.
  *          If "excludeObjectChildren", cause undefined values for this object and pass `includeUndefinedValues` as
  *          false when computing defaults for any nested object properties.
+ * @param [behaviorBitFlags=0] Optional bitwise flags to set which behavior is chosen for certain edge cases.
```
heath-freenome, line 346:
```suggestion
 * @param [defaultFormStateBehavior=0] - Optional bit flag, if provided, allows users to override default form state behavior
```

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -317,14 +355,20 @@ export default function getDefaultFormState<
   theSchema: S,
   formData?: T,
   rootSchema?: S,
-  includeUndefinedValues: boolean | 'excludeObjectChildren' = false
+  includeUndefinedValues: boolean | 'excludeObjectChildren' = false,
+  behaviorBitFlags = 0
```
heath-freenome, line 359:
```suggestion
  defaultFormStateBehavior = 0
```

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -317,14 +355,20 @@ export default function getDefaultFormState<
...
   if (!isObject(theSchema)) {
     throw new Error('Invalid schema: ' + theSchema);
   }
   const schema = retrieveSchema<T, S, F>(validator, theSchema, rootSchema, formData);
-  const defaults = computeDefaults<T, S, F>(validator, schema, undefined, rootSchema, formData, includeUndefinedValues);
-  if (typeof formData === 'undefined' || formData === null || (typeof formData === 'number' && isNaN(formData))) {
+  const defaults = computeDefaults<T, S, F>(validator, schema, {
+    rootSchema,
+    includeUndefinedValues,
+    behaviorBitFlags,
```
heath-freenome, line 368:
```suggestion
    behaviorBitFlags: defaultFormStateBehavior,
```

packages/utils/src/types.ts
```diff
@@ -956,9 +956,10 @@ export interface SchemaUtilsType<T = any, S extends StrictRJSFSchema = RJSFSchem
    *
    * @param validator - An implementation of the `ValidatorType` interface that will be compared against the current one
    * @param rootSchema - The root schema that will be compared against the current one
+   * @param behaviorBitFlags Bitwise flags to set which behavior is chosen for certain edge cases
```
heath-freenome, line 959:
```suggestion
   * @param [defaultFormStateBehavior=0] - Optional bit flag, if provided, allows users to override default form state behavior
```

packages/utils/src/types.ts
```diff
@@ -956,9 +956,10 @@ export interface SchemaUtilsType<T = any, S extends StrictRJSFSchema = RJSFSchem
...
    * @returns - True if the `SchemaUtilsType` differs from the given `validator` or `rootSchema`
    */
-  doesSchemaUtilsDiffer(validator: ValidatorType<T, S, F>, rootSchema: S): boolean;
+  doesSchemaUtilsDiffer(validator: ValidatorType<T, S, F>, rootSchema: S, behaviorBitFlags: number): boolean;
```
heath-freenome, line 962:
```suggestion
  doesSchemaUtilsDiffer(validator: ValidatorType<T, S, F>, rootSchema: S, defaultFormStateBehavior: number): boolean;
```

CHANGELOG.md
RPiAwesomeness: Done

CHANGELOG.md
RPiAwesomeness: Ah, that's leftover from a previous set of CHANGELOG changes, my bad

packages/utils/src/createSchemaUtils.ts
RPiAwesomeness: I'm open to discuss this but I had intentionally kept this parameter as required for the constructor here. Making the parameter for the `SchemaUtils` constructor makes it clear that this is important to the usage of `SchemaUtils`. Also, when I made it optional, it caused the unit tests to become more complex without providing much, if any, benefit. If a user wants to create a `SchemaUtils` instance I think it makes sense for them to explicitly specify which behavior they want. 

The only place this object is constructed is `createSchemaUtils()` in `createSchemaUtils.ts:289` and the parameter for that function is optional, which I believe accomplishes the same result from the perspective of the library's users.

heath-freenome: Oh, yeah I agree that making it required is necessary, I was just copy/pasting the name and documentation from other lines and missed this one.

heath-freenome:
@RPiAwesomeness It looks like when you accepted my feedback I wrote it in a way that broke the formatting rules. You can run `npm run cs-format` at the root level to fix it and push a new commit

RPiAwesomeness:
I tried running the tests locally after rebasing onto main and correcting the format and it seems like there's a significant number failing all for `antd` and `chakra-ui`. Any ideas as to what might be causing this?

heath-freenome:
> I tried running the tests locally after rebasing onto main and correcting the format and it seems like there's a significant number failing all for `antd` and `chakra-ui`. Any ideas as to what might be causing this?

There was a bug fix for antd just pushed over the weekend. Not sure about chakra-ui.

RPiAwesomeness:
> There was a bug fix for antd just pushed over the weekend. Not sure about chakra-ui.

Weird. I deleted and re-cloned the repository and now it's passing. Probably some weird caching issue.

heath-freenome:
> > There was a bug fix for antd just pushed over the weekend. Not sure about chakra-ui.
> 
> Weird. I deleted and re-cloned the repository and now it's passing. Probably some weird caching issue.

@RPiAwesomeness I recently added a new script to the root to blow away all of the node_modules directories and reinstall them. Next time give that a try?
```
npm run refresh-node-modules
```

nickgros:
Remind me why we decided on a bitmask? In my opinion, a bitmask is only useful if you're either
* Using bitwise operators that don't have simple boolean equivalents or
* Are resource constrained, e.g. bandwidth

I don't think that either of these apply here, and I'm worried that some behavior might not decompose to binary options, which would complicate the API even further. I think we would be better off with an object containing options with enum or boolean values. This is basically equivalent to the bitwise but has a more developer-friendly API.

Is the "legacy" behavior going to warrant its own option? Would it make sense for that to just be the `undefined` case for an object prop and make that the default value?

If we're not yet certain about how this new prop API should look, then I think a good option would be to name it something like `experimental_defaultFormStateBehavior` and in the docs state that we may break it between minor/patch versions. That would give us the freedom to wait and see which (if any--there are many other options) of these potential APIs would be best:

```ts
type FormProps = {
  // bitmask
  experimental_defaultFormStateBehavior?: number
}

type FormProps = {
  // enumerated behavior
  experimental_defaultFormStateBehavior?: 'legacy' | 'ignoreArrayMinItemsUnlessRequired'
}

type FormProps = {
  // many individual options
  experimental_defaultFormStateBehavior: {
    arrayMinItems?: 'populate' | 'requiredOnly' // | 'ignore' // if other options like 'ignore' make sense
  }
}

type FormProps = {
  // enums broken down by type
  experimental_defaultFormStateBehavior: {
    arrays?: 'populateMinItems' | 'ignoreMinItemsUnlessRequired'
	objects?: unknown // TBD
  }
}

type FormProps = {
  // boolean fields split by type
  experimental_defaultFormStateBehavior: {
    arrays?: {
      ignoreMinItemsUnlessRequired?: boolean
      // ...other array properties
    }
	objects?: unknown // TBD
  }
}

type FormProps = {
  // enumeration fields split by type
  experimental_defaultFormStateBehavior: {
    arrays?: {
      minItems?: "ignore" | "requiredOnly" | "always"
      // ...other array properties
    }
	objects?: unknown // TBD
  }
}
```

RPiAwesomeness:
> Remind me why we decided on a bitmask?
 
@nickgros I believe it was so this could be more flexible in the future because we had initially decided to have a single "features" or "behavior" prop and you could turn individual features on and off something like this: `behavior={DefaultFormStateBehavior.IgnoreMinArrayItemsUnlessRequired | Make.BaconPancakes | Require.AdditionalPylons`. 

I'm personally a bigger fan of switching to a specific type union a la your third example, especially since the configuration of this has already been moved into an individual `defaultFormStateBehavior` prop.

> ```ts
> type FormProps = {
>   // many individual options
>   experimental_defaultFormStateBehavior: {
>     arrayMinItems?: 'populate' | 'requiredOnly' // | 'ignore' // if other options like 'ignore' make sense
>   }
> }
> ```

I remember something about adding at least one more mode but I don't remember what exactly at this moment. I'll wait for input from @heath-freenome before finally making the changes but I think that implementation makes a lot more sense in terms of developer ergonomics.

heath-freenome:
> > Remind me why we decided on a bitmask?
> 
> @nickgros I believe it was so this could be more flexible in the future because we had initially decided to have a single "features" or "behavior" prop and you could turn individual features on and off something like this: `behavior={DefaultFormStateBehavior.IgnoreMinArrayItemsUnlessRequired | Make.BaconPancakes | Require.AdditionalPylons`.
> 
> I'm personally a bigger fan of switching to a specific type union a la your third example, especially since the configuration of this has already been moved into an individual `defaultFormStateBehavior` prop.
> 
> > ```ts
> > type FormProps = {
> >   // many individual options
> >   experimental_defaultFormStateBehavior: {
> >     arrayMinItems?: 'populate' | 'requiredOnly' // | 'ignore' // if other options like 'ignore' make sense
> >   }
> > }
> > ```
> 
> I remember something about adding at least one more mode but I don't remember what exactly at this moment. I'll wait for input from @heath-freenome before finally making the changes but I think that implementation makes a lot more sense in terms of developer ergonomics.

Let's go with the `experimental_defaultFormStateBehavior` solution

RPiAwesomeness:
I've made the feature flag refactor and rebased onto main, should be working the same as expected. Minor changes to the playground to support the new feature flag format and updated the documentation.

heath-freenome:
@RPiAwesomeness it looks like you'll need to run `npm run cs-format` for your PR

CHANGELOG.md
```diff
@@ -15,6 +15,19 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
+# 5.8.0
```
heath-freenome, line 18: Since we haven't published `5.7.0` yet this could be released as part of that release. I have one more PR to complete for it and won't be done with it until early next week.

RPiAwesomeness:
Re-ran `cs-format` and updated the CHANGELOG

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -93,28 +100,45 @@ function maybeAddDefaultToObject<T = any>(
   }
 }
 
+interface computeDefaultsProps<T = any, S extends StrictRJSFSchema = RJSFSchema> {
```
nickgros, line 103: For consistency, all interfaces and types use PascalCase.

```suggestion
interface ComputeDefaultsProps<T = any, S extends StrictRJSFSchema = RJSFSchema> {
```

packages/utils/src/types.ts
```diff
@@ -31,6 +31,10 @@ export type RJSFSchema = StrictRJSFSchema & GenericObjectType;
  */
 export type FormContextType = GenericObjectType;
 
+export type Experimental_DefaultFormStateBehavior = {
+  arrayMinItems?: 'populate' | 'requiredOnly';
+};
```
nickgros, line 36: Would like this types documentation to match what's in the docs project, not a big deal while this is experimental though

packages/utils/test/schema/getDefaultFormStateTest.ts
```diff
@@ -34,7 +34,7 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
           },
           $ref: '#/definitions/testdef',
         };
-        expect(computeDefaults(testValidator, schema, undefined, schema)).toEqual({
+        expect(computeDefaults(testValidator, schema, { rootSchema: schema })).toEqual({
```
nickgros, line 37: I wasn't sold the new function signature for `computeDefaults` until I saw these tests. This is much cleaner!

packages/utils/test/schema/getDefaultFormStateTest.ts
```diff
@@ -218,20 +228,124 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
...
+        };
+        expect(
+          computeDefaults(testValidator, schema, {
+            rootSchema: schema,
+            experimental_defaultFormStateBehavior: { arrayMinItems: 'requiredOnly' },
+          })
+        ).toEqual({ requiredArray: ['default0', 'default1'] });
+      });
+      // BUG: https://github.com/rjsf-team/react-jsonschema-form/issues/3602
+      it.skip('should combine defaults with raw form data for a required array property with minItems', () => {
```
nickgros, line 329: Thanks for keeping the skipped test so this is easy to come back to

packages/utils/test/schema/getDefaultFormStateTest.ts
```diff
@@ -218,20 +228,124 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
...
+          },
+        };
+        expect(
+          computeDefaults(testValidator, schema, {
+            rootSchema: schema,
+            experimental_defaultFormStateBehavior: { arrayMinItems: 'requiredOnly' },
+          })
+        ).toEqual({});
+      });
+      it('should return empty array when given an empty array as form data for an optional array property with minItems ', () => {
```
nickgros, line 271:
```suggestion
      it('should return empty array when given an empty array as form data for an optional array property with minItems', () => {
```

packages/docs/docs/api-reference/form-props.md
```diff
@@ -59,6 +59,41 @@ Formerly the `validate` prop.
...
+  minItems: 3,
+};
+
+render(
+  <Form
+    schema={schema}
+    validator={validator}
+    experimental_defaultFormStateBehavior={
+      arrayMinItems: 'requiredOnly'
+    }
```
nickgros, line 91:
```suggestion
    experimental_defaultFormStateBehavior={{
      arrayMinItems: 'requiredOnly'
    }}
```

nickgros (approved):
Couple of suggestions but none of them should block us from merging this feature. This looks great!

packages/utils/test/schema/getDefaultFormStateTest.ts
RPiAwesomeness: With all the new optional stuff being added between my MR and the others going in ahead of me I figured it just made sense. Glad you like it!

packages/utils/src/types.ts
nickgros: I just mean having the supplemental text here as well so IDEs can provide the documentation info at a glance, so one does not have to open the docs page to know what each option does

PR #3605, opened 2023-04-13
chore: Added new scripts for automating releasing steps

heath-freenome:

There are a lot of little steps that I do when releasing that I'm automating
- Added a new `scripts/bump-peer-deps.js` automation that will update all package.json files in the packages subdirectories to bump any `@rjsf/*` peer dependencies
  - It sets the peer dependencies to the `Major.minor.x` version of the devDependency `^Major.minor.patch`
  - This is needed due to this use case:
```
 A new API is added in `@rjsf/utils@5.4.0` that is used by `@rjsf/core@5.4.0`. If someone were to only update
 `@rjsf/core` to 5.4.0 but keep using `@rjsf/utils@5.3.0` then we will have a problem. By bumping the
 peerDependencies to `5.4.x` we support any patch versions of `5.4` but will prevent the user from forgetting to
 bump `@rjsf/utils` to 5.4
```
- Added new `bump-packages` script to all `package.json` files that calls `npm update --save --lockfile-version 2`
  - We need to lock down to lockfile version 2 because version 3 is only supported on the latest versions of Node 18
- Added new scripts to the main `package.json`, `bump-all-packages`, `bump-peer-deps`, `refresh-node-modules` and `post-versioning` that do the following:
  - `bump-all-packages` calls `bump-packages` for all `package.json` files and then runs `npm install` to restore sym links
    - It also runs the build and test scripts to make sure things work
  - `bump-peer-deps` calls the new script using node
  - `refresh-node-modules` removes the main `node_modules` and `packages/*/node_modules` directory and runs `npm install` to refresh all `node_modules` directories
    - This also can have the side-effect of bumping `package-lock.json` files
  - `post-versioning` calls the `bump-peer-deps` and `refresh-node-module` and then adds any changes made by those two scripts and creates and pushes a commit with them
- Updated the `contributing` docs to instruct people to use the `post-versioning` command as part of the process

### Checklist


package.json
```diff
@@ -12,6 +12,9 @@
     "test": "lerna run --concurrency 2 --stream test",
     "test:update": "lerna run --concurrency 2 --stream test:update",
     "lint": "lerna run lint",
+    "bump-peer-deps": "node scripts/bump-peer-deps.js",
+    "refresh-node-modules": "rm -rf node_modules && rm -rf packages/*/node_modules && npm install",
```
nickgros, line 16: Would prefer using `rimraf` for platform independence, but do we really need to blow away node_modules for a new release? Couldn't we just run `npm update --save`?

nickgros (approved):

Issue #2734, opened 2022-02-25
fluent-ui theme is no longer rendering properly in the playground


epicfaace:
![image](https://user-images.githubusercontent.com/1689183/155753135-ced2b72f-3b08-4dd1-8857-8ef21b5302d0.png)

epicfaace:
Still an issue

gnalvesteffer:
Still happening.

heath-freenome:
Since we deprecated and removed this theme, this is no longer an issue

PR #3626, opened 2023-04-26
Add additionalProperties support to fluent-ui theme

### LINKED ISSUE #2777: AdditionalProperties are not supported in FluentUI template
### Prerequisites

### Description

Hello,
AdditionalProperties do not work in the Fluent UI template, I've followed this issue tracked here for Semantic-ui Bootstrap4 and fluentUI in #1927 plus for MaterialUI #1724 and unfortunately the Fluent version never happened.

So I'm just going to log this here as an open item. I'd love to contribute to this, but time won't be on my side.. :(
 

### Steps to Reproduce

1. Goto https://rjsf-team.github.io/react-jsonschema-form/
2. Select additionalProperties example
3. Select theme fluent-ui

#### Expected behavior

image

#### Actual behavior

Additional properties section has no add new row or remove row.

### Version

^4.1.0

tjasz:

Additional properties should be handled in the fluent-ui theme equivalently to how they are handled in the other themes.

- Adds "Add Item" button to objects that support additional properties.
- Adds editable key and a remove button to additional properties.
- Fixes #2777.

### Checklist

### Screenshots

The playground still does not render fluent-ui styles per #2734. Regardless, here is a before-and-after of the additionalProperties example:

Before
image

After
image

### DIFF
--- a/packages/fluent-ui/src/FieldTemplate/FieldTemplate.tsx
+++ b/packages/fluent-ui/src/FieldTemplate/FieldTemplate.tsx
@@ -13,24 +13,22 @@ export default function FieldTemplate<
   S extends StrictRJSFSchema = RJSFSchema,
   F extends FormContextType = any
 >(props: FieldTemplateProps<T, S, F>) {
-  const { id, children, errors, help, displayLabel, rawDescription, hidden, uiSchema, registry } = props;
+  const { children, errors, help, displayLabel, rawDescription, hidden, uiSchema, registry } = props;
   const uiOptions = getUiOptions<T, S, F>(uiSchema);
   const WrapIfAdditionalTemplate = getTemplate<'WrapIfAdditionalTemplate', T, S, F>(
     'WrapIfAdditionalTemplate',
     registry,
     uiOptions
   );
-  // TODO: do this better by not returning the form-group class from master.
-  let { classNames = '' } = props;
-  classNames = 'ms-Grid-col ms-sm12 ' + classNames.replace('form-group', '');
+  if (hidden) {
+    return <div style={{ display: 'none' }}>{children}</div>;
+  }
   return (
     <WrapIfAdditionalTemplate {...props}>
-      <div id={id} className={classNames} style={{ marginBottom: 15, display: hidden ? 'none' : undefined }}>
-        {children}
-        {displayLabel && rawDescription && <Text>{rawDescription}</Text>}
-        {errors}
-        {help}
-      </div>
+      {children}
+      {displayLabel && rawDescription && <Text>{rawDescription}</Text>}
+      {errors}
+      {help}
     </WrapIfAdditionalTemplate>
   );
 }
--- a/packages/fluent-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
+++ b/packages/fluent-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
@@ -41,7 +41,7 @@ export default function WrapIfAdditionalTemplate<
 
   if (!additional) {
     return (
-      <div id={id} className={classNames} style={{ ...style, marginBottom: 15 }}>
+      <div className={classNames} style={{ ...style, marginBottom: 15 }}>
         {children}
       </div>
     );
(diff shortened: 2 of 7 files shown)


packages/fluent-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
```diff
@@ -1,11 +1,75 @@
...
+  } = props;
+  const { templates, translateString } = registry;
+  // Button templates are not overridden in the uiSchema
+  const { RemoveButton } = templates.ButtonTemplates;
+  const keyLabel = translateString(TranslatableString.KeyLabel, [label]);
+  const additional = ADDITIONAL_PROPERTY_FLAG in schema;
+
+  if (!additional) {
+    return (
+      <div className={classNames} style={style}>
```
tjasz, line 40: Most of the snapshot changes come from this additional `div`. This was added so that the `classNames` and `style` properties have an effect. This matches what is done in all other themes.

heath-freenome:
@tjasz Thanks for fixing this issue!

Please add a few screen shots of the changes inside of the playground so that I can make sure it looks ok.

packages/fluent-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
heath-freenome: @tjasz Can you replace this with the `div` from the `FieldTemplate` and remove it from there? This will make `fluent-ui` consistent with the others. Also you may want to either pass in the `classNames` defined in `FieldTemplate` to `WrapIfAdditionalTemplate` or move that code here as well.

tjasz: Thanks for pointing me to that. That was the right thing to do. The tests look a lot better. The Array and Form tests still have some diffs where `display: undefined` was removed, but no more extra `<div>`s.

packages/fluent-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
```diff
@@ -1,11 +1,79 @@
...
+      <div id={id} className={classNames} style={{ ...style, marginBottom: 15 }}>
+        {children}
+      </div>
+    );
+  }
+
+  const handleBlur = ({ target }: FocusEvent<HTMLInputElement>) => onKeyChange(target.value);
+
+  return (
+    <div className={classNames} style={{ ...style, marginBottom: 15 }} dir='ltr'>
```
heath-freenome, line 53: I believe you also want the id here too?
```suggestion
    <div id={id} className={classNames} style={{ ...style, marginBottom: 15 }} dir='ltr'>
```

packages/fluent-ui/src/FieldTemplate/FieldTemplate.tsx
```diff
@@ -13,24 +13,22 @@ export default function FieldTemplate<
...
     'WrapIfAdditionalTemplate',
     registry,
     uiOptions
   );
-  // TODO: do this better by not returning the form-group class from master.
-  let { classNames = '' } = props;
-  classNames = 'ms-Grid-col ms-sm12 ' + classNames.replace('form-group', '');
+  if (hidden) {
+    return <div style={{ display: 'none' }}>{children}</div>;
+  }
```
heath-freenome, line 25: Good catch on this!

packages/fluent-ui/src/FieldTemplate/FieldTemplate.tsx
```diff
@@ -13,24 +13,22 @@ export default function FieldTemplate<
...
   return (
     <WrapIfAdditionalTemplate {...props}>
-      <div id={id} className={classNames} style={{ marginBottom: 15, display: hidden ? 'none' : undefined }}>
```
heath-freenome, line 28: Ugh, that could have been written cleaner. Sigh, nice cleanup there too

heath-freenome (changes requested):
One small change that will reduce the diffs in the snapshots

packages/fluent-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
tjasz: Looking at the other themes again, it doesn't look like the `WrapIfAdditionalTemplate`s in mui or antd us the `id` this way. They only use it on the text field (`id={`${id}-key`}`).

I agree that I should have `id={id}` in both `return`s or neither `return`. Preference?

heath-freenome: Neither return actually. I was looking in the snapshots and there are a ton of duplicate IDs because of this, which is bad HTML. Another good catch!

tjasz: Sounds good. Thanks!

heath-freenome (approved):

PR #3673, opened 2023-05-15
fix(core): prevent from breaking when element to focus cannot be found

anthonycaron:

If for some reason the element to focus cannot be found when the validation fails and the `focusOnFirstError` prop is set, then it will throw an error.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 5.7.1
+
+## @rjsf/core
+
+- Added a check to prevent from breaking when the element to focus - in case `focusOnFirstError` is set - cannot be found
+
 # 5.7.0
 
 ## @rjsf/antd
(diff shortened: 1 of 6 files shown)


anthonycaron:
Hey @heath-freenome, I've created this really small PR because I had an issue with this in my project. I know it might be related to the fact that this happens in a custom widget and that somehow it's not using the right id in that widget to focus my element. But still if we can prevent it from breaking I guess it's good, and it might allow me to see the errors on my field anyway.

heath-freenome:
@anthonycaron 

> Hey @heath-freenome, I've created this really small PR because I had an issue with this in my project. I know it might be related to the fact that this happens in a custom widget and that somehow it's not using the right id in that widget to focus my element. But still if we can prevent it from breaking I guess it's good, and it might allow me to see the errors on my field anyway.

I'm not sure why the snapshots needed to be updated, but they seem to be breaking the tests

heath-freenome:
@anthonycaron I added the fix into my PR #3674 since I needed to do an emergency release anyway

anthonycaron:
@heath-freenome Thank you !

PR #3674, opened 2023-05-15
fix: only output createSchemaValidators function for the cjs development mode

### LINKED ISSUE #3668: Latest @rjsf/validator-ajv8 (with precompiled validator) breaks web import (it imports 'fs')
### Prerequisites

### What theme are you using?

validator-ajv8

### Version

5.7.0

### Current Behavior

Importing `@rjsf/validator-ajv8@5.7.0` causes client-side to break:
```
.../node_modules/@rjsf/validator-ajv8/dist/validator-ajv8.esm.js:6:0
Module not found: Can't resolve 'fs'
```
It seems that some nodejs-specific javascript code paths should be separated from default import!

### Expected Behavior

Downgrading to `5.6.2` fixes this, works like expected! No error.

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

heath-freenome:

fixes #3668 (hopefully) by only outputting the `createSchemaValidators()` function for the CJS development mode
- Copied `index.ts` to `index-dev.ts` and then removed `createSchemaValidators()` (the function using `fs`) from the `index.ts`
- Added a `dts.config.js` file that uses the `index-dev.ts` file as the compile source ONLY for the cjs development build
  - This should hopefully prevent browsers and deployed production software from encountering the `fs` issue
  - Updated the documentation to describe how to avoid the issue for development deployments in webpack and nextjs

### Checklist


packages/validator-ajv8/dts.config.js
```diff
@@ -0,0 +1,12 @@
...
+
+module.exports = {
+  // This function will run for each entry/format/env combination
+  rollup(config, options) {
+    //
+    if (options.format === 'cjs' && options.env === 'development') {
+      config.input['validator-ajv8'] = config.input['validator-ajv8'].replace('index.ts', 'index-dev.ts');
+    }
+    return config; // always return a config.
+  },
```
nickgros, line 11: I think this approach is going to cause more confusion, since one of the bundles will just be different. Could we just pass two files ([see docs](https://github.com/weiran-zsd/dts-cli#multiple-entry-files)) where one file is `index.ts` with `compileSchemaValidators` removed, as you have done, and then the other entrypoint could just be `compileSchemaValidators.ts`? Then in the docs I think we could then instruct users to precompile in Node by importing `@rjsf/validator-ajv8/compileSchemaValidators` (or maybe this would be `@rjsf/validator-ajv8/dist/esm/compileSchemaValidators` or `@rjsf/validator-ajv8/dist/cjs/compileSchemaValidators`, but whatever the end path is isn't that important)

heath-freenome: Putting the `compileSchemaValidators` into the esm build is part of the problem since that build is used by browsers because it is defined by `module` in the package.json. Since only cjs development ever needs the compiler then keeping it separate seems to be the best idea. Another option could be to provide a whole new package, `validator-ajv8-cli`, but then we'd have to do a better job of providing a CLI like `ajv-cli` does

nickgros (approved):

PR #3721, opened 2023-06-05
Cache schemas by explicit ID in AJV8Validator.isValid

### LINKED ISSUE #3692: Terrible performance with big anyOf field
### Prerequisites

### What theme are you using?

core

### Version

5.7.0

### Current Behavior

Since upgrade from v2 to v5, we're having a lot of trouble with performance on one of our bigger forms. The whole schema is a single `oneOf` field, with 86 different object options, and about 6000 lines total.

The form is basically unusable when typing, because `validator.isValid()` calls from inside `Form->getDefaultFormState()->getClosestMatchingOption()` and `MuiSchemaField->componentDidUpdate()->getClosestMatchingOption()` are clogging up the main thread completely:

image

I was hoping to solve those problems by using the discriminator property, as hinted [in this comment](https://github.com/rjsf-team/react-jsonschema-form/issues/3517#issuecomment-1536485098), but it isn't making to be making any noticable difference - because `isValid` still gets called, even when using discriminant: 
``` typescript 
if (discriminatorField && has(option, [PROPERTIES_KEY, discriminatorField])) {
  const value = get(formData, discriminatorField);
  const discriminator = get(option, [PROPERTIES_KEY, discriminatorField], {});

  if (validator.isValid(discriminator, value, rootSchema)) {
    return i;
  }
}
``` 

I'm not sure why this is clogging the thread so badly even on strong computers, but it does. Not sure if its a bug, but it may very well be, since `liveValidate` runs completely OK on the form - and I don't see a reason why pre-selecting an option should be so much more demanding than liveValidating the whole form. 

This modification removes my performance problems completely, although it's certainly not a suitable complete solution: https://github.com/michal-kurz/react-jsonschema-form/pull/2/files

Is there a bug with ajv validators? If not, is it appropriate to run this demanding validation in this manner? What do you think should be done about it, if anything? 

### Expected Behavior

_No response_

### Steps To Reproduce

Paste [this schema](https://pastebin.com/J9Btpkn6) into playground  - this is our real schema, with values replaced by hashes (to strip potentially sensitive information)

### Environment

_No response_

### Anything else?

_No response_

evshi:

fixes #3692

For allOf/anyOf schemas, subschemas are compiled each time they are encountered if they lack an $id field. This can create poor performance for large lists or recursive refs. Leveraging the ajv cache to cache subschemas that have been encountered before greatly improves performance.

### Checklist

### DIFF
--- a/packages/validator-ajv8/src/validator.ts
+++ b/packages/validator-ajv8/src/validator.ts
@@ -138,16 +138,15 @@ export default class AJV8Validator<T = any, S extends StrictRJSFSchema = RJSFSch
         this.ajv.addSchema(rootSchema, rootSchemaId);
       }
       const schemaWithIdRefPrefix = withIdRefPrefix<S>(schema) as S;
-      const schemaHash = hashForSchema(schemaWithIdRefPrefix);
+      const schemaId = schemaWithIdRefPrefix[ID_KEY] ?? hashForSchema(schemaWithIdRefPrefix);
       let compiledValidator: ValidateFunction | undefined;
-      if (schemaWithIdRefPrefix[ID_KEY]) {
-        compiledValidator = this.ajv.getSchema(schemaWithIdRefPrefix[ID_KEY]);
-      } else {
-        compiledValidator = this.ajv.getSchema(schemaHash);
-      }
+      compiledValidator = this.ajv.getSchema(schemaId);
       if (compiledValidator === undefined) {
+        // Add schema by an explicit ID so it can be fetched later
+        // Fall back to using compile if necessary
+        // https://ajv.js.org/guide/managing-schemas.html#pre-adding-all-schemas-vs-adding-on-demand
         compiledValidator =
-          this.ajv.addSchema(schemaWithIdRefPrefix, schemaHash).getSchema(schemaHash) ||
+          this.ajv.addSchema(schemaWithIdRefPrefix, schemaId).getSchema(schemaId) ||
           this.ajv.compile(schemaWithIdRefPrefix);
       }
       const result = compiledValidator(formData);
--- a/packages/validator-ajv8/src/validator.ts
+++ b/packages/validator-ajv8/src/validator.ts
@@ -13,6 +13,7 @@ import {
   ValidationData,
   ValidatorType,
   withIdRefPrefix,
+  hashForSchema,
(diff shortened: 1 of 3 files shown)


packages/validator-ajv8/src/validator.ts
```diff
@@ -137,12 +138,17 @@ export default class AJV8Validator<T = any, S extends StrictRJSFSchema = RJSFSch
...
       if (schemaWithIdRefPrefix[ID_KEY]) {
         compiledValidator = this.ajv.getSchema(schemaWithIdRefPrefix[ID_KEY]);
+      } else {
+        compiledValidator = this.ajv.getSchema(schemaHash);
       }
       if (compiledValidator === undefined) {
-        compiledValidator = this.ajv.compile(schemaWithIdRefPrefix);
+        compiledValidator =
+          this.ajv.addSchema(schemaWithIdRefPrefix, schemaHash).getSchema(schemaHash) ||
+          this.ajv.compile(schemaWithIdRefPrefix);
```
heath-freenome, line 151: I'm curious about the `||` of the two statements. Why not do this instead?
```suggestion
        const addedSchema = this.ajv.addSchema(schemaWithIdRefPrefix, schemaHash).getSchema(schemaHash);
        compiledValidator = this.ajv.compile(addedSchema);
```

evshi: This is recommended by the ajv docs https://ajv.js.org/guide/managing-schemas.html#pre-adding-all-schemas-vs-adding-on-demand.

`compile` will cache by the subschema `$id` property, or by the object hash. It doesn't allow specifically storing and fetching with a specific hash. So the `||` allows us to store and fetch by the hash, and then fallback to the default behavior if necessary.

heath-freenome: Can you add the link to the documentation above this line then?

evshi:
@heath-freenome updated, PTAL

heath-freenome (commented):
@evshi This looks pretty good! Can you update the `CHANGELOG.md` and add an `5.8.2` section for this change?

evshi:
@heath-freenome updated CHANGELOG and confirmed up to date with main

nickgros (approved):

PR #3746, opened 2023-06-27
fix: bug fix and new feature for experimental_defaultFormStateBehavior

heath-freenome:

Fixed a bug with `experimental_defaultFormStateBehavior.emptyObjectFields='populateRequiredDefaults'` and added new feature for merging `minItems` array defaults into `formData`
- In `@rjsf/utils`, fixed the bug and added the new feature as follows:
  - Updated the `mergeDefaultsWithFormData()` function to take a new optional `mergeExtraArrayDefaults=false` parameter that causes array defaults to append onto `formData` values when true
  - Added a new `Experimental_ArrayMinItems` type containing `populate` and `mergeExtraDefaults`
  - Updated the `Experimental_DefaultFormStateBehavior` to switch `arrayMinItems` to be of the `Experimental_ArrayMinItems` type
    - Updated all tests to switch to use this new object structure
  - Updated the `computeDefaults()` function to not default `required` and to pass it to all of the recursive calls to itself where it was missing
  - Updated the `maybeAddDefaultToObject()` function to use the required state of the field itself when the (now optional) `isParentRequired` is not specified
  - Updated the `getDefaultFormState()` function to pass the value of `experimental_defaultFormStateBehavior.arrayMinItems.mergeExtraDefaults` to `mergeDefaultsWithFormData()`
    - Added additional tests that verify the bug fix and new behavior, while uncommenting a skipped test after fixing the expected result
- Updated the documentation for the new features in `mergeDefaultsWithFormData()` and `experimental_defaultFormStateBehavior.arrayMinItems`

### Checklist

Screenshot 2023-06-27 at 11 57 57 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
-# 5.8.3
+# 5.9.0
 
 ## @rjsf/utils
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,23 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
+# 5.8.3
+
+## @rjsf/utils
+
+- Updated `getDefaultFormState()` to fix a bug where `experimental_defaultFormStateBehavior: { emptyObjectFields: 'populateRequiredDefaults' }` wasn't working for object properties with `$ref`s
+- Experimental feature **breaking change**:
+  - Updated the `experimental_defaultFormStateBehavior.arrayMinItems` from simple flag to an object containing two optional fields, `populate` and `mergeExtraDefaults`
+    - The new `arrayMinItems.mergeExtraDefaults` flag, when "true", allows users to merge defaults onto the end of `formData` arrays when `minItems` is specified
+  - If you were previously passing `experimental_defaultFormStateBehavior` as `{ arrayMinItems = 'requiredOnly }` on the `Form`, now you would pass `{ arrayMinItems: { populate: 'requiredOnly' } }`
+- Added a new, optional `mergeExtraArrayDefaults=false` flag to the `mergeDefaultWithFormData()` utility function to support the new `arrayMinItems.mergeExtraDefaults` experimental feature
+
+## Dev / docs / playground
+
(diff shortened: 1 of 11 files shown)


packages/docs/docs/api-reference/form-props.md
```diff
@@ -61,24 +61,41 @@ See [Validation](../usage/validation.md) for more information.
...
+| Flag Value     | Description                                                                                                                         |
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
+| `all`          | Legacy behavior - populate minItems entries with default values initially and include empty array when no values have been defined. |
+| `requiredOnly` | Ignore `minItems` on a field when calculating defaults unless the field is required.                                                |
+
+#### `arrayMinItems.mergeExtraDefaults`
+
+Optional boolean flag, defaulting to `false` when not specified.
+When `formData` is provided and does not contain `minItems` worth of data, this flag controls whether the extra data provided by the defaults is appended onto the existing `formData` items to ensure the `minItems` condition is met.
+When `false`, only the `formData` provided is merged into the default form state, even if there are fewer than the `minItems`.
```
heath-freenome, line 87:
```suggestion
When `false` (legacy behavior), only the `formData` provided is merged into the default form state, even if there are fewer than the `minItems`.
```

CHANGELOG.md
```diff
@@ -15,6 +15,23 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
+# 5.8.3
```
nickgros, line 18: Should probably go to 5.9.0 since we added a new flag (and maybe because we're breaking an experimental feature?)

packages/docs/docs/api-reference/form-props.md
```diff
@@ -61,24 +61,41 @@ See [Validation](../usage/validation.md) for more information.
 
 ## experimental_defaultFormStateBehavior
 
-Experimental features to specify different form state behavior. Currently, this only affects the handling of optional array fields where `minItems` is set.
+Experimental features to specify different form state behavior.
```
nickgros, line 64: Unrelated to your changes, I think we should add this for clarity.

```suggestion
Experimental features to specify different form state behavior. **This API is experimental and unstable, therefore breaking changes may be shipped in minor or patch releases.** If you want to use this feature, we recommend pinning exact versions of @rjsf/\* packages in your package.json file.
```

nickgros (approved):

CHANGELOG.md
heath-freenome:
```suggestion
# 5.9.0
```

PR #3765, opened 2023-07-13
Enable npm workspaces

zxbodya:

Enabling workspaces, to improve dependency installation time and IDE performance/usability (having much less code to index making things faster)

Similar to #3382 - which btw would be great when ready, and probably can be considered intermediate step on the way to it.

By using npm workspaces, it is possible to get good improvements, while not having to fix symlink compatibility issues (which I guess is what is yet to be looked into for using pnpm)


packages/antd/jest.config.js
```diff
@@ -6,7 +6,5 @@ module.exports = {
   },
   moduleNameMapper: {
     '\\.(css|less)$': '<rootDir>/__mocks__/styleMock.js',
-    '^react$': '<rootDir>/node_modules/react',
-    '^react-dom$': '<rootDir>/node_modules/react-dom',
```
zxbodya, line 10: this is no longer correct or needed, instead there is single react version which is resolved correctly

packages/mui/test/__snapshots__/Form.test.tsx.snap
```diff
@@ -14176,7 +14176,8 @@ label[data-shrink=false]+.MuiInputBase-formControl .emotion-3:focus::-ms-input-p
                 "height": 0,
                 "left": 0,
                 "overflow": "hidden",
-                "padding": 0,
+                "paddingBottom": 0,
+                "paddingTop": 0,
```
zxbodya, line 14180: this is because not all package-lock files were used, and effectively an update for some of the dependencies in repository (change is coming from the new mui version)

packages/playground/vite.config.ts
```diff
@@ -13,12 +12,6 @@ export default defineConfig({
...
   resolve: {
     preserveSymlinks: true, // Fixes https://github.com/rjsf-team/react-jsonschema-form/issues/3228
-    alias: {
-      // The following is needed to allow the material ui v4 and v5 themes to properly load the css
-      '@mui/styles': path.resolve('./node_modules', '@mui/styles'),
-      '@material-ui/styles': path.resolve('./node_modules', '@material-ui/styles'),
-      '@emotion/react': path.resolve('./node_modules/@emotion/react'),
-      '@emotion/styled': path.resolve('./node_modules/@emotion/styled'),
-    },
+    alias: {},
```
zxbodya, line 15: this is to be incorrect path (this dependencies happen to be installed in top level node_modules), also everything still works as expected without this

packages/utils/package.json
```diff
@@ -46,7 +46,6 @@
     "@babel/plugin-transform-react-jsx": "^7.22.5",
     "@babel/preset-env": "^7.22.5",
     "@babel/preset-react": "^7.22.5",
-    "@types/jest-expect-message": "^1.1.0",
```
zxbodya, line 49: types are loaded from `jest-expect-message` directly

package.json
```diff
@@ -34,6 +33,7 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7.22.5",
     "@nrwl/nx-cloud": "^15.3.5",
+    "@types/estree": "^1.0.1",
```
zxbodya, line 36: this is to avoid typescript errors from `@types/eslint` package for which outdated `@types/estree` peer dependency will be installed otherwise

better fix, maybe as a next step would be to update `dts-cli`

zxbodya:
> @zxbodya I'm not sure why these changes are not working on the github actions. Can you investigate?
yes, looking into it - it looks to be because of `cache-dependency-path` being misconfigured (just pushed a fix)

heath-freenome:
Also, I'm releasing 5.10.0 right now which includes bumping packages so once I merge it I expect you will have to resolve conflicts

heath-freenome:
Ok, resolve those merge commits...

heath-freenome (approved):

PR #3771, opened 2023-07-14
use jest directly instead of dts-cli

zxbodya:
Updating project configurations to use jest directly instead of dts-cli

This allows better jest IDE integrations to work out of the box.

Also updating to current jest version.


packages/validator-ajv8/jest.config.js
```diff
@@ -11,7 +11,8 @@ module.exports = {
   coveragePathIgnorePatterns: ['<rootDir>/node_modules/', '<rootDir>/test'],
   coverageThreshold: {
     global: {
-      branches: 100,
+      // todo: dropped from 100 after jest 29
+      branches: 98,
```
zxbodya, line 15: there is improvement in coverage reporting in new jest causing drop in branch coverage

nickgros (approved):

packages/validator-ajv8/jest.config.js
heath-freenome: Please file a new issue to restore this back to 100%

heath-freenome:
@zxbodya I'm not sure what's going on but your tests seem to be running without completion for both node 16 and 18. Is that the case locally? I'm guessing not. Can you investigate?

zxbodya:
yes, it works locally. will look into it

zxbodya:
found the issue:

```
@rjsf/core: Jest did not exit one second after the test run has completed.
@rjsf/core: 
@rjsf/core: 'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

this was caused by something in jsdom environment setup, switching to use `testEnvironment: 'jsdom'` in jest config - fixed it

heath-freenome (approved):

PR #3779, opened 2023-07-21
update playground vite config to use sources directly

zxbodya:
Updating vite configuration to use source of packages inside monorepo directly, avoiding the compilation step

This is to improve developer experience using playground, allowing instant reload for changes inside monorepo bypassing need of the build step


nickgros (approved):
Looks good, pending local check of the built playground

PR #3781, opened 2023-07-21
update build configuration to not use dts-cli

zxbodya:
Updating build configuration to directly use tsc, esbuild and rollup instead of dts-cli.

Also updating publish configuration to include source files, improving debug experience for consumers.

continuation of #3771

- improving build time
- allowing to run typescript compiler separately from build (and so making it easier to quickly check for type errors)
- by having typescript build output, it is possible to have correct navigation to source improving DX
- prerequisite for updating typescript version

### Changes

- using `tsc` to generate esm modules and typescript definitions, among other benefits this allows to navigation to sources inside the monorepo
- using `esbuild` for creating commonjs and esm bundle
- using `rollup` for creating umd bundle from esm bundle
- adding src folder to published files, so it can be used together with provided source maps

### Changes in exported artefacts

- no more split to production/development bundles
- no overrides to use ems import path (for example imports from `antd/lib` are not overritten to use `antd/es`)

Both of this, if needed - are expected to be configured in a bundler on the library consumer side


heath-freenome:
@zxbodya Go ahead and get this working now that the jest PR is merged

packages/docs/main.js
```diff
@@ -1,41 +0,0 @@
-$(function () {
```
zxbodya, line 1: the file is not used anymore, and also has broken link to `unpkg`

packages/docs/docs/00-introduction.mdx
```diff
@@ -39,24 +39,6 @@ import Form from '@rjsf/core';
 
 Our latest version requires React 16+. You can also install `react-jsonschema-form` (the 1.x version) which works with React 15+.
 
-### As a script served from a CDN
```
zxbodya, line 42: `core.cjs.production.min.js` is to be no longer correct after the changes (new link should be to `index.js` and it is also not minified)

However, for both urls - it is a commonjs bundle, which will not work in browser environment because it will try to "require" all the dependencies (react, lodash, etc…) 

This can be changes to umd bundle, but that also would not be easy - requiring bunch of script tags with all the dependencies. And so, I believe, it is better to just remove this section entirely to avoid confusion and also to consider umd bundle as deprecated and to remove that in next major version(in case anyone still using it)

packages/semantic-ui/tsconfig.json
```diff
@@ -1,12 +1,13 @@
...
     "rootDir": "./src",
+    "outDir": "./lib",
     "baseUrl": "./",
-    "paths": {
-      "*": ["src/*", "node_modules/*"]
-    },
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    // todo:
+    // ../../node_modules/semantic-ui-react/dist/commonjs/generic.d.ts(73,73): error TS2344: Type 'TProps' does not satisfy the constraint 'Record<string, any>'.
```
heath-freenome, line 10: More explanation? It looks like semantic ui has a type issue?

packages/chakra-ui/tsconfig.json
```diff
@@ -1,12 +1,13 @@
...
     "rootDir": "./src",
+    "outDir": "./lib",
     "baseUrl": "./",
-    "paths": {
-      "*": ["src/*", "node_modules/*"]
-    },
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    // todo:
+    // ../../node_modules/@chakra-ui/menu/dist/declarations/src/use-menu.d.ts:986:61 - error TS2694: Namespace '"node_modules/csstype/index".Property' has no exported member 'ColorAdjust'.
```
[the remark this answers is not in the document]
heath-freenome: replace todo with a comment about `skipLibCheck`

packages/fluent-ui/tsconfig.json
```diff
@@ -1,12 +1,13 @@
...
+    "outDir": "./lib",
     "baseUrl": "./",
-    "paths": {
-      "*": ["src/*", "node_modules/*"]
-    },
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    // todo
+    // ../../node_modules/@chakra-ui/menu/dist/declarations/src/use-menu.d.ts:986:61 - error TS2694: Namespace '"react-jsonschema-form/node_modules/csstype/index".Property' has no exported member 'ColorAdjust'.
+    "skipLibCheck": true
```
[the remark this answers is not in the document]
heath-freenome: replace todo with a comment about `skipLibCheck`?

packages/semantic-ui/tsconfig.json
heath-freenome: replace todo with a comment about `skipLibCheck`

heath-freenome:
@zxbodya Since I just released 5.11.1 you'll need to resolve your conflicts. Sorry

CHANGELOG.md
```diff
@@ -25,6 +25,14 @@ should change the heading of the (upcoming) version to include a major version b
...
 - Updated the `resolveAllReferences()` function to use object spreading to update properties and items in a schema rather than directly modifying the schema to avoid issues with frozen object, fixing [#3805](https://github.com/rjsf-team/react-jsonschema-form/issues/3805)
 
+## Dev / playground
+
+- update playground vite config to use sources directly, allowing to reload changes in it without additional build step
+- moving from `dts-cli` to use individual dev tools directly, updating package publish config
+  - tsc for generating type definitions and esm modules
+  - esbuild for CJS bundle
+  - rollup for UMD bundle
+
```
heath-freenome, line 35: Leave 5.11.2 and move this into 5.12.0. I needed to release the fix for #3805 to unblock my company

zxbodya: done

heath-freenome (approved):

nickgros (approved):

PR #3845, opened 2023-08-29
fix: fixes #3692 - simple discriminator now bypasses isValid() calls in getMatchingOption and getClosestMatchingOption for improved performance

### LINKED ISSUE #3692: Terrible performance with big anyOf field

michal-kurz:
fixes #3692

### Checklist


packages/docs/docs/api-reference/utility-functions.md
```diff
@@ -296,6 +296,22 @@ Using the `schema`, `defaultType` and `options`, extract out the props for the `
 
 - InputPropsType: The extracted `InputPropsType` object
 
+### getOptionMatchingSimpleDiscriminator()
+
+Compares the value of `discriminatorField` within `formData` against the value of `discriminatorField` within schema for each `option`. Returns index of first `option` whose discriminator matches formData. Returns `undefined` if there is no match.
+
+This function does not work with discriminators of `"type": "object"` and `"type": "array"`
```
michal-kurz, line 303: I quite struggled trying to make this description as understandable and straght-forwards as possible.

packages/utils/src/getOptionMatchingSimpleDiscriminator.ts
```diff
@@ -0,0 +1,36 @@
...
+        continue;
+      }
+
+      if (discriminator.const === value) {
+        return i;
+      }
+
+      if (discriminator.enum?.includes(value)) {
+        return i;
+      }
```
michal-kurz, line 31: As somebody who is not all too familiar with the full `jsonschema` specifiacation, it's hard for me to clearly understand what cases are left after excluding `discriminator.type === 'object' || discriminator.type === 'array'`.

It would make it much more understandable for me, if the function were conceived as `handles consts and enums of primitive js types` rather than `handles all discriminators except discriminator.type === 'object' || discriminator.type === 'array'`. This ties into the naming of the function, and also its docs description.

michal-kurz: Also, looking at the code, it seems it would produce exactly the same results if the 
``` ts
if (discriminator.type === 'object' || discriminator.type === 'array') {
   continue;
}
```
were to be omitted (for valid schemas), which to me suggests that the function's role should be further refined.

packages/utils/test/getOptionMatchingSimpleDiscriminator.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    test('enum discriminator', () => {
+      expect(
+        getOptionMatchingSimpleDiscriminator(
+          { foo: 'foo' },
+          [{}, { type: 'object', properties: { foo: { enum: ['bar', 'foo'] } } }],
+          'foo'
+        )
+      ).toEqual(1);
+    });
+  });
```
michal-kurz, line 58: Perhaps there are some other relevant cases that I'm missing

packages/utils/src/getOptionMatchingSimpleDiscriminator.ts
```diff
@@ -0,0 +1,36 @@
+import get from 'lodash/get';
+import { PROPERTIES_KEY } from './constants';
+import { RJSFSchema, StrictRJSFSchema } from './types';
+
+export default function getOptionMatchingSimpleDiscriminator<T = any, S extends StrictRJSFSchema = RJSFSchema>(
```
heath-freenome, line 5: Please add the same documentation you put into the docs above here.

packages/utils/test/getOptionMatchingSimpleDiscriminator.test.ts
heath-freenome: You just need to figure out the test coverage

packages/utils/src/getOptionMatchingSimpleDiscriminator.ts
michal-kurz: done

packages/utils/src/getOptionMatchingSimpleDiscriminator.ts
michal-kurz: After digging a bit deeper, I now understand that discriminator is just a reference to an arbitrary a part of any valid schema, and not a completely independent mechanism of matching options against data. Now it makes sense to me why you wanted to define and name the function this way :)

packages/utils/test/getOptionMatchingSimpleDiscriminator.test.ts
michal-kurz: Done

heath-freenome (approved):

Issue #3639, opened 2023-05-05
Bump antd theme to latest major version


heath-freenome:
Antd theme is one major version behind, we are on 4, the latest is 5

heath-freenome:
Do you know whether this can be done with backwards compatibility to 4.x? If not, we'll have to begin the 6.x plan and have to do it in a beta branch

heath-freenome:
Actually someone has taken some time to do the [work](https://github.com/fran122157/react-jsonschema-form/pull/1) in response to this issue https://github.com/rjsf-team/react-jsonschema-form/issues/3507

koszti:
wondering if something has changed in the last two month. Has anyone made some progress to support antd 5?

heath-freenome:
We haven't started the v6 work. Someone has already come up with what seems to be the [changes necessary](https://github.com/fran122157/react-jsonschema-form/pull/1) to get it to work with 5.x. If someone wanted to ensure those changes were backwards compatible with 4.x, then we would consider pulling it into the 5.x release of RJSF

cr8or1:
@heath-freenome It will be backwards compatible with 4.x We really need the antd 5 to work with RJSF. Can you approve this PR?

heath-freenome:
@cr8or1 The changes are in fork that we don't control.  You are welcome to copy that code into your own fork and push up a PR for us to review and merge. Can you do that?

PR #3849, opened 2023-08-31
antd theme related changes, bump v4 to v5 

Manish3323:

[AntD version upgrade](https://github.com/rjsf-team/react-jsonschema-form/issues/3639)

Somehow my fork's main branch got all its commits discarded, that resulted in Automatic closing of the earlier PR.


packages/antd/package.json
```diff
@@ -37,7 +37,7 @@
     "@ant-design/icons": "^4.0.0",
     "@rjsf/core": "^5.12.x",
     "@rjsf/utils": "^5.12.x",
-    "antd": "^4.0.0",
+    "antd": "^4.24.0",
```
nickgros, line 40: Shouldn't this be `"^4.24.0 || ^5.8.5" ? Otherwise antd v5 users will get a peer dependency warning/error

Manish3323: the caret symbol will handle it right?.. npm installs latest version of v4..

Manish3323: Ah okay.. Understood..

Manish3323: Fixed

PR #3862, opened 2023-09-08
[DRAFT] Custom widgets for arrays

### LINKED ISSUE #573: Custom array widget.

aliyss:

Allows to have custom Widgets for arrays. I have not tested it apart from my own implementations. It worked for me without any issues currently. I'll make sure to check back in in case I notice some issues.

I would like to mark this as a DRAFT, since I'm not sure of any issues. I also didn't have time to check the tests yet. Any help testing this would be appreciated which is why I'm already making the pull request.

Here are some logical tests I did:
- `addedByAdditionalProperties` is `true`, since the `name` is something like root_[number].
- `uiSchema.additionalProperties` must be empty, but like this it still won't break
- `get(uiSchema, [name], {})` works if the name doesn't match anything but falls back to default behaviour 

fixes #573 

### Checklist

### DIFF
--- a/packages/core/src/components/fields/ObjectField.tsx
+++ b/packages/core/src/components/fields/ObjectField.tsx
@@ -276,7 +276,7 @@ class ObjectField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
       description: uiOptions.label === false ? undefined : description,
       properties: orderedProperties.map((name) => {
         const addedByAdditionalProperties = has(schema, [PROPERTIES_KEY, name, ADDITIONAL_PROPERTY_FLAG]);
-        const fieldUiSchema = addedByAdditionalProperties ? uiSchema.additionalProperties : uiSchema[name];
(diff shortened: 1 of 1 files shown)


heath-freenome:
@aliyss I'm not sure how this is enabling your use case. Can you provide a codesandbox.io with more code so that we can understand what it is you are attempting to do

aliyss:
@heath-freenome I think I'm going crazy. I could have sworn something the widget stuff didn't work beforehand and after changing that part it worked.

I am using a json-schema to create a json-schema (like a form builder) so I'm sure I messed up at some point in the recursiveness. Apologies for the nonsensical code.

Keeping this to reproduce in the future if I stumble across it again:
https://codesandbox.io/p/sandbox/peaceful-noether-g635td

I'll close this Pull Request.

PR #3881, opened 2023-09-26
feat(templates): Allow customization per-field by using Registry string key

### LINKED ISSUE #3695: current per-field customization break JSON rules
### Prerequisites

### What theme are you using?

core

### Version

5.x

### Current Behavior

Currently, as described in the documentation https://rjsf-team.github.io/react-jsonschema-form/docs/advanced-customization/custom-templates, it is possible to customize per-field Field, Template and Widget user interfaces. The problem with the current model is that assigning the function/class directly in the uiSchema causes it to no longer be valid application/json.

```json
{
  "ui:field": MyCustomField,
  "ui:ArrayFieldTemplate": MyArrayTemplate
}
```

### Expected Behavior

A much consistent and superior alternative, in my humble opinion, would be to assign the name of the key associated with the templates transmitted during the initialization of the Form, as is done with widgets and fields.

```js
import LayoutFieldTemplate from '~/lib/rjsf/templates/LayoutFieldTemplate.js'
import Form from '@rjsf/core'

const templates = {
  "layout": LayoutFieldTemplate
}

function App () {
  return <Form templates={templates} />
}
```

and when necessary to change the template, use it like:

```json
{
  "ui:template": "layout"
}
```

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Windows 22H2 (19045.2965)
- Node: 16.20.0
- npm: 8.19.4
```

### Anything else?

_No response_

nagaozen:
Fixes #3695

Currently, to per-field customize the template, it is necessary to send the `function`/`class` of the component in the uiSchema, breaking the JSON nature of the `uiSchema`. It's not possible to provide just the registration key.

This PR fixes this issue by allowing the consumer of the library (the one with access to changing only schema, uiSchema and formData) to choose a different template provided to the form by the developer.

@ahkhanjani, could you please review?

### Checklist

### DIFF
--- a/packages/docs/docs/advanced-customization/custom-templates.md
+++ b/packages/docs/docs/advanced-customization/custom-templates.md
@@ -76,7 +76,7 @@ render(
 );
 ```
 
-You also can provide your own field template to a uiSchema by specifying a `ui:ArrayFieldTemplate` property.
+You also can provide your own field template to a uiSchema by specifying a `ui:ArrayFieldTemplate` property with your Component or a string value from the `Registry`.
 
 ```tsx
 import { UiSchema } from '@rjsf/utils';
@@ -162,7 +162,7 @@ render(
 );
 ```
 
-You also can provide your own template to a uiSchema by specifying a `ui:ArrayFieldDescriptionTemplate` property.
+You also can provide your own template to a uiSchema by specifying a `ui:ArrayFieldDescriptionTemplate` property with your Component or a string value from the `Registry`.
 
 ```tsx
 import { UiSchema } from '@rjsf/utils';
@@ -260,7 +260,7 @@ render(
 );
 ```
 
-You also can provide your own template to a uiSchema by specifying a `ui:ArrayFieldDescriptionTemplate` property.
+You also can provide your own template to a uiSchema by specifying a `ui:ArrayFieldDescriptionTemplate` property with your Component or a string value from the `Registry`.
 
 ```tsx
 import { UiSchema } from '@rjsf/utils';
@@ -612,7 +612,7 @@ render(
 );
 ```
 
-You also can provide your own field template to a uiSchema by specifying a `ui:FieldTemplate` property.
+You also can provide your own field template to a uiSchema by specifying a `ui:FieldTemplate` property with your Component or a string value from the `Registry`.
 
 ```tsx
 import { UiSchema } from '@rjsf/utils';
@@ -691,7 +691,7 @@ render(
 );
 ```
 
-You also can provide your own field template to a uiSchema by specifying a `ui:ObjectFieldTemplate` property.
+You also can provide your own field template to a uiSchema by specifying a `ui:ObjectFieldTemplate` property with your Component or a string value from the `Registry`.
 
 ```tsx
 import { UiSchema } from '@rjsf/utils';
--- a/packages/utils/src/getTemplate.ts
+++ b/packages/utils/src/getTemplate.ts
@@ -18,6 +18,10 @@ export default function getTemplate<
   if (name === 'ButtonTemplates') {
     return templates[name];
   }
+  // Allow templates to be customized per-field by using string keys from the registry
+  if (Object.hasOwn(uiOptions, name) && Object.hasOwn(templates, uiOptions[name])) {
+    return templates[uiOptions[name]]
+  }
   return (
     // Evaluating uiOptions[name] results in TS2590: Expression produces a union type that is too complex to represent
     // To avoid that, we cast uiOptions to `any` before accessing the name field
--- a/packages/utils/src/getTemplate.ts
+++ b/packages/utils/src/getTemplate.ts
@@ -24,10 +24,10 @@ export default function getTemplate<
     typeof uiOptions[name] === 'string' &&
     Object.hasOwn(templates, uiOptions[name] as string)
   ) {
-    const key = uiOptions[name]
+    const key = uiOptions[name];
     // Evaluating templates[key] results in TS2590: Expression produces a union type that is too complex to represent
     // To avoid that, we cast templates to `any` before accessing the key field
-    return (templates as any)[key]
+    return (templates as any)[key];
   }
   return (
     // Evaluating uiOptions[name] results in TS2590: Expression produces a union type that is too complex to represent
--- a/packages/utils/src/getTemplate.ts
+++ b/packages/utils/src/getTemplate.ts
@@ -19,7 +19,11 @@ export default function getTemplate<
     return templates[name];
   }
   // Allow templates to be customized per-field by using string keys from the registry
-  if (Object.hasOwn(uiOptions, name) && Object.hasOwn(templates, uiOptions[name])) {
+  if (
+    Object.hasOwn(uiOptions, name) &&
+    typeof uiOptions[name] === 'string' &&
+    Object.hasOwn(templates, uiOptions[name])
+  ) {
     return templates[uiOptions[name]]
   }
   return (
--- a/packages/utils/src/getTemplate.ts
+++ b/packages/utils/src/getTemplate.ts
@@ -22,9 +22,12 @@ export default function getTemplate<
   if (
     Object.hasOwn(uiOptions, name) &&
     typeof uiOptions[name] === 'string' &&
-    Object.hasOwn(templates, uiOptions[name])
+    Object.hasOwn(templates, uiOptions[name] as string)
   ) {
-    return templates[uiOptions[name]]
+    const key = uiOptions[name]
+    // Evaluating uiOptions[name] results in TS2590: Expression produces a union type that is too complex to represent
+    // To avoid that, we cast templates to `any` before accessing the key field
+    return (templates as any)[key]
   }
   return (
     // Evaluating uiOptions[name] results in TS2590: Expression produces a union type that is too complex to represent
--- a/packages/utils/src/getTemplate.ts
+++ b/packages/utils/src/getTemplate.ts
@@ -25,7 +25,7 @@ export default function getTemplate<
     Object.hasOwn(templates, uiOptions[name] as string)
   ) {
     const key = uiOptions[name]
-    // Evaluating uiOptions[name] results in TS2590: Expression produces a union type that is too complex to represent
+    // Evaluating templates[key] results in TS2590: Expression produces a union type that is too complex to represent
     // To avoid that, we cast templates to `any` before accessing the key field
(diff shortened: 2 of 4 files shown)


nagaozen:
Anyone to help be review this PR? @heath-freenome?

packages/utils/test/getTemplate.test.ts
```diff
@@ -86,4 +86,19 @@ describe('getTemplate', () => {
       expect(getTemplate<typeof name>(name, registry, uiOptions)).toBe(CustomTemplate);
     });
   });
+  it('returns the template from registry using uiOptions key when available', () => {
```
heath-freenome, line 89: Consider adding an additional test that uses a name that is not one of the standard names? I.e. if a user adds their own named template, then they should be able to get it out via that name, correct?

heath-freenome:
@nagaozen This fix makes sense in the basic form, only it doesn't really help the situation where someone wants to build multiple templates to override. Because, currently, the `TemplateTypes` interface only supports the known template names, by simply adding a new (say) `ObjectFieldTemplate` to the `template` prop in the `Form` ALL `ObjectFieldTemplate`s will now use this implementation. If what you are hoping to support is something like how `fields` and `widgets` work (i.e. allowing a user to build several `ObjectFieldTemplate` implementations and override the `uiSchema` to select from one of the implementations), then more work is necessary. For one, you will have to extend the `TemplateTypes` interface to support custom named `templates` for a specific type (note `fields` and `widgets` all have the same Typescript type where as templates have unique Typescript types). This is no easy task and requires an enhancement to the design of the templates scheme in `RJSF`. Come to the next weekly meeting if you'd like to talk about it.

nagaozen:
> @nagaozen This fix makes sense in the basic form, only it doesn't really help the situation where someone wants to build multiple templates to override. Because, currently, the `TemplateTypes` interface only supports the known template names, by simply adding a new (say) `ObjectFieldTemplate` to the `template` prop in the `Form` ALL `ObjectFieldTemplate`s will now use this implementation. So basically, by simply adding the updated implemention as the `ObjectFieldTemplate` in `templates` it will replace the OOTB implementation provided by `core` or the theme you are using. So picking the name inside of a `uiSchema` is effectively the same as not providing it at all.
> 
> If what you are hoping to support is something like how `fields` and `widgets` work (i.e. allowing a user to build several `ObjectFieldTemplate` implementations and override the `uiSchema` to select from one of the implementations), then more work is necessary.
> 
> For one, you will have to extend the `TemplateTypes` interface to support custom named `templates` for a specific type (note `fields` and `widgets` all have the same Typescript type where as templates have unique Typescript types). This is no easy task and requires an enhancement to the design of the templates scheme in `RJSF`. Come to the next weekly meeting if you'd like to talk about it.

@heath-freenome, thank you for your response. My idea with this patch is to use the existing mechanism for customizing templates per field as described in <https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/docs/docs/advanced-customization/custom-templates.md>. The difference is that, instead of receiving the function directly in the uiSchema body, the developer can choose to insert another key in the template registry that can be used instead of manually sending the entire component manually (think about using the same customized template in different fields). I'm sure the patch works as we are already running it on production. For example, see how we are using the layout template at <https://actions.looplex.com/code/custom-widgets-showcase> for both `Shipping Address` object and `Strings`. There's also an example of our autofill field working: try to type "01302001" on the CEP field and TAB for the next field :-) 

image

I'll be glad to expose the reasoning behind the patch in the next weekly. Thank you for the invitation.

packages/utils/test/getTemplate.test.ts
heath-freenome, line 89: Add another set of tests for custom names

packages/utils/src/getTemplate.ts
```diff
@@ -18,6 +18,17 @@ export default function getTemplate<
   if (name === 'ButtonTemplates') {
     return templates[name];
   }
+  // Allow templates to be customized per-field by using string keys from the registry
+  if (
+    Object.hasOwn(uiOptions, name) &&
+    typeof uiOptions[name] === 'string' &&
+    Object.hasOwn(templates, uiOptions[name] as string)
+  ) {
+    const key = uiOptions[name];
```
heath-freenome, line 27: To truly support things in a way that matches how `widgets` and `fields` updating the `TemplateTypes` type in `@rjsf/utils/src/types.ts` to add this one additional line is all the extra work you will need to do:

```
  /** Allow this to support any named `ComponentType` or an object of named `ComponentType`s */
  [name: string]: ComponentType<any> | { [name: string]: ComponentType<any> };
```

packages/docs/docs/advanced-customization/custom-templates.md
```diff
@@ -76,7 +76,7 @@ render(
 );
 ```
 
-You also can provide your own field template to a uiSchema by specifying a `ui:ArrayFieldTemplate` property.
+You also can provide your own field template to a uiSchema by specifying a `ui:ArrayFieldTemplate` property with your Component or a string value from the `Registry`.
```
heath-freenome, line 79: Consider adding at least one example of how to use an alternate template with a different name in the `uiSchema` here and possibly all the sections you've updated below

heath-freenome (changes requested):

heath-freenome:
> > @nagaozen This fix makes sense in the basic form, only it doesn't really help the situation where someone wants to build multiple templates to override. Because, currently, the `TemplateTypes` interface only supports the known template names, by simply adding a new (say) `ObjectFieldTemplate` to the `template` prop in the `Form` ALL `ObjectFieldTemplate`s will now use this implementation. So basically, by simply adding the updated implemention as the `ObjectFieldTemplate` in `templates` it will replace the OOTB implementation provided by `core` or the theme you are using. So picking the name inside of a `uiSchema` is effectively the same as not providing it at all.
> > If what you are hoping to support is something like how `fields` and `widgets` work (i.e. allowing a user to build several `ObjectFieldTemplate` implementations and override the `uiSchema` to select from one of the implementations), then more work is necessary.
> > For one, you will have to extend the `TemplateTypes` interface to support custom named `templates` for a specific type (note `fields` and `widgets` all have the same Typescript type where as templates have unique Typescript types). This is no easy task and requires an enhancement to the design of the templates scheme in `RJSF`. Come to the next weekly meeting if you'd like to talk about it.
> 
> @heath-freenome, thank you for your response. My idea with this patch is to use the existing mechanism for customizing templates per field as described in https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/docs/docs/advanced-customization/custom-templates.md. The difference is that, instead of receiving the function directly in the uiSchema body, the developer can choose to insert another key in the template registry that can be used instead of manually sending the entire component manually (think about using the same customized template in different fields). I'm sure the patch works as we are already running it on production. For example, see how we are using the layout template at https://actions.looplex.com/code/custom-widgets-showcase for both `Shipping Address` object and `Strings`. There's also an example of our autofill field working: try to type "01302001" on the CEP field and TAB for the next field :-)
> 
> image
> 
> I'll be glad to expose the reasoning behind the patch in the next weekly. Thank you for the invitation.

After thinking more about it, I believe that changing the `TemplateType` and the associated types like I've noted above will be enough

packages/utils/src/getTemplate.ts
heath-freenome: Oops, I was wrong. It will require creating a new interface called `RJSFBaseProps` with the definition:
```
export interface RJSFBaseProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any> {
  /** The schema object for the field being described */
  schema: S;
  /** The uiSchema object for this description field */
  uiSchema?: UiSchema<T, S, F>;
  /** The `registry` object */
  registry: Registry<T, S, F>;
}
```

Then you will want to update all of the interfaces and types that the `TemplateTypes` inner props that have `ComponentType<XXX>` to make the `XXX` interface/type extend `RJSFBaseProps`

Then add the following to `TemplateTypes` instead of what I mentioned above:

/** Allow this to support any named `ComponentType` or an object of named `ComponentType`s */
  [name: string]: ComponentType<RJSFBaseProps<T, S, F> | { [name: string]: ComponentType<RJSFBaseProps<T, S, F> };

heath-freenome:
@nagaozen Any progress on this?

nagaozen:
> @nagaozen Any progress on this?

Hi @heath-freenome ! I am deep diving in a vital feature in my business, but asked if someone in my team could implement the static typing requirements for us. If nobody cares to conclude the ticket I'll come to finish it asap. Thanks for the insights

nickgros:
@nagaozen there's a similar request in #3962 with a slightly different API than you've proposed. What do you think of this?

nagaozen:
> @nagaozen there's a similar request in #3962 with a slightly different API than you've proposed. What do you think of this?

Hi @nickgros,

Ultimately, the goal is the same: "Customize the ObjectFieldTemplate through the `uiSchema`." The use case presented by @elevesque-nexapp is similar to what we are facing here: "The backend is produced by an engineering team and the frontend developer provides only the `schema, uiSchema, and formData`." This segregation of concern is what makes this feature necessary. Because backend developers can document these alternatives so that frontend engineers can use them.  As for the suggested approach (nesting the options in objects), it differs from how the library already handles custom widgets and fields. Therefore I believe this current approach superior in terms of DX.

elevesque-nexapp:
I agree the solution proposed here is better aligned with how customization works for other fields. I'd be perfectly happy to see this solution merged!

heath-freenome:
@nagaozen Seems like your solution is the winner. Let's get those typing changes made and then we complete your PR

nagaozen:
> @nagaozen Do you think you will have time for this?
> 
> Maybe I can play with this to see if I can get it to a mergeable state

Oh, please be my guest. Its on my todo list for a considerable time. I believe this will greatly improve the library.

Oddwerth:
Hello, any news about this PR ?

Oddwerth:
> > @nagaozen Do you think you will have time for this?
> > Maybe I can play with this to see if I can get it to a mergeable state
> 
> Oh, please be my guest. Its on my todo list for a considerable time. I believe this will greatly improve the library.

I would gladly make the required changes but I don't have the rights to update your branch

heath-freenome:
@Oddwerth You could create your own fork and then cherry pick these changes from theirs

lucasmcht:
@heath-freenome I have created a new PR with the changes here : https://github.com/rjsf-team/react-jsonschema-form/pull/4352

Issue #3909, opened 2023-10-17
Form becomes unresponsive when "liveOmit" and "omitExtraData" props are turned on for schemas containing fixed array of objects


Rozamo:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

5.13.2

### Current Behavior

I am using the following schema in the playground with `liveOmit` and `omitExtraData` turned on

```json
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "properties": {
    "array": {
      "type": "array",
      "items": [
        {
          "properties": {
            "title": {
              "type": "string"
            }
          }
        }
      ]
    }
  }
}
```

Upon typing anything in the `title` input box, the focus shifts away from the input box and there is no change in the input box or the form data

### Expected Behavior

The data entered should be reflected in the `title` input box and the form data.

### Steps To Reproduce

[...]

PR #3916, opened 2023-10-21
fix: add condition to pick correct schemas for fixed array items in topathschemainternal util

Rozamo:

Fixes https://github.com/rjsf-team/react-jsonschema-form/issues/3909.

#### Changes
- Adds condition to generate correct path schemas for schemas containing fixed array of items in the `toPathSchemaInternal` util. The `items` property for these schemas are of `array` type instead of `object`. Because the condition for the array type `items` was not present, the array type schema was being sent into the recursive function as the schema for the form data, which was not matching. Hence empty form data was being generated each time Form's `onChange` was being called
- Adds tests for the above change ensuring 100% coverage for the `toPathSchema` util

### Checklist

### DIFF
--- a/packages/utils/src/schema/toPathSchema.ts
+++ b/packages/utils/src/schema/toPathSchema.ts
@@ -95,6 +95,8 @@ function toPathSchemaInternal<T = any, S extends StrictRJSFSchema = RJSFSchema, 
             element,
             _recurseList
           );
+        } else {
+          console.warn(`Unable to generate path schema for "${name}.${i}". No schema defined for it`);
         }
       });
     } else {
--- a/packages/utils/src/schema/toPathSchema.ts
+++ b/packages/utils/src/schema/toPathSchema.ts
@@ -6,7 +6,6 @@ import {
   ALL_OF_KEY,
   ANY_OF_KEY,
   ADDITIONAL_PROPERTIES_KEY,
-  ADDITIONAL_ITEMS_KEY,
   DEPENDENCIES_KEY,
   ITEMS_KEY,
   NAME_KEY,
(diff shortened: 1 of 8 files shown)


packages/utils/test/schema/toPathSchemaTest.ts
```diff
@@ -702,5 +702,179 @@ export default function toPathSchemaTest(testValidator: TestValidatorType) {
...
+              },
+            ],
+            additionalItems: false,
+          },
+        },
+      };
+
+      const formData = {
+        str: 'string',
+        arr: [{ name: 'name1' }, 'name2'],
```
heath-freenome, line 860: I'm wondering whether there should be a `console.warn` or `console.error` for this condition?

Rozamo: I can see that `toPathSchema` util is mostly used as a step to omit extra data. In that case should we warn the users about the omitted data?

packages/utils/src/schema/toPathSchema.ts
```diff
@@ -73,16 +73,42 @@ function toPathSchemaInternal<T = any, S extends StrictRJSFSchema = RJSFSchema,
...
+        } else if (schemaAdditionalItems) {
+          pathSchema[i] = toPathSchemaInternal<T, S, F>(
+            validator,
+            schemaAdditionalItems as S,
+            `${name}.${i}`,
+            rootSchema,
+            element,
+            _recurseList
+          );
+        }
```
heath-freenome, line 98: For the moment, let's add a `console.warn()` here to let users know that they have unhandled `formData`. Which means we want to update the test below with a check that `console.warn()` was called. This can be done with a `jest.spyOn()`

Rozamo: Sure. Added a message. Let me know if there is a change required.

heath-freenome (approved):

packages/utils/src/schema/toPathSchema.ts
Rozamo: @heath-freenome Should this log be moved up a level outside the for loop? We don't know how many items there will be in the array and if there are too many items, it might fill up the console.

Issue #3972, opened 2023-11-21
How to implement `ajv-keywords` with rjsf `@rjsf/validator-ajvX`?


michal-kurz:
### Prerequisites

- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)

### What theme are you using?

core

### What is your question?

I want to use the [uniqItemProperties](https://ajv.js.org/packages/ajv-keywords.html#uniqueitemproperties) keyword from `ajv-keyword` library - how can I make this work with `@rjsf/validator-ajvX`?

I am using ajv-6 validator invoked like this: 

``` typescript
import { customizeValidator } from '@rjsf/validator-ajv6'
const validator = customizeValidator({ customFormats: CUSTOM_FORMAT_CRITERIA as any })
```

The custom keyword should be applied to the `ajv` instance itself like so:
``` typescript
require('ajv-keywords')(ajv, ['uniqueItemProperties'])
```
`ajv-keyword` then calls `ajv.addKeyword()` on the `ajv` instance - is there a clean way of applying it with `@rjsf/validator-ajv6`?

heath-freenome:
@michal-kurz at the moment we don't expose the internal `ajv` instance anywhere as we are trying to make validators implementation agnostic. That said, it probably makes sense for us to update our `customizeValidator()` to expose the underlying `ajv` instance so that you can do what you need. I'll push a PR with the changes soon

michal-kurz:
Thank you, @heath-freenome !

PR #3991, opened 2023-12-05
Fix: Expose the internal `ajv` variable in the validator implementation classes

heath-freenome:

Fixes: #3972 indirectly by exposing the `ajv` variable for use in the issue
- In `@rjsf/ajv6`, updated `AJV6Validator` to make the `ajv` variable public and changed the return of the `customizeValidator()` function to remove the interface return
- In `@rjsf/ajv8`, updated `AJV8Validator` to make the `ajv` variable public and changed the return of the `customizeValidator()` function to remove the interface return
  - Also removed some `@ts-expect-error` tags that are no longer needed due to the exposed variable

### Checklist


packages/docs/docs/usage/validation.md
```diff
@@ -585,6 +585,21 @@ const validator = customizeValidator({ ajvOptionsOverrides });
 render(<Form schema={schema} validator={validator} />, document.getElementById('app'));
 ```
 
+## Using the raw Ajv instance
+
+The `customizeValidator()` function returns the `AJV8Validator` (or `AJV6Validator` depending on the library you use) implementation class, which has an internal raw `ajv` instance within it.
+If you need to do some deep customization of the instance using any of the `ajv` libraries (like `ajv-keywords`), you can do so using this raw instanse as follows:
```
nickgros, line 591:
```suggestion
If you need to do some deep customization of the instance using any of the `ajv` libraries (like `ajv-keywords`), you can do so using this raw instance as follows:
```

nickgros (approved):
Looks good, one typo

PR #4065, opened 2024-01-25
Fix potential XSS in the preview button of FileWidget

### LINKED ISSUE #4057: Validate dataURL to prevent XSS attacks
Hi, I've found a Cross-Site Scripting (XSS) vulnerability in `@rjsf/core`

**Vulnerability Details:**
- **Severity**: High/Critical
- **Description**: There's a risk of malicious script execution when the value of the FileWidget is controlled by an adversary.

**Steps to Reproduce:**
```
import React from "react";
import ReactDOM from "react-dom/client";
import { getDefaultRegistry } from "@rjsf/core";

const schema = {
  title: "Files",
  type: "object",
  properties: {
    file: {
      type: "string",
      format: "data-url",
      title: "Single file",
    },
  },
};

function App() {
  const FileWidget = getDefaultRegistry().widgets.FileWidget;

  return (
    <FileWidget
      value={["javascript:alert(1)"]}
      registry={getDefaultRegistry()}
      options={{ id: 1, filePreview: true }}
      id="1"
      schema={schema}
      onChange={() => {}}
      onFocus={() => {}}
    />
  );
}

const root = ReactDOM.createRoot(document.getElementById("root"));
root.render(<App />);

```
In this case, when a user clicks the preview button, the malicious script `alert(1)` will be executed. Note that the specified `format: "data-url"` does not successfully sanitize the value.

**Suggested Fix or Mitigation:**
The root cause is due to: https://github.com/rjsf-team/react-jsonschema-form/blob/f31bef19c1420aec74384fc4f4152bed579f0902/packages/core/src/components/widgets/FileWidget.tsx#L79
where a malicious dataURL may be passed to the `<a />`, causing an XSS attack. 

It is best practice for a package to ensure security. Even though the dataURL has been validated elsewhere, there's still a risk of malicious dataURL being passed to `<a />` as shown in the example. Please consider patching this issue by validating the dataURL to prevent the XSS attack. Thanks!

yuki-js:

fixes #4057

The value prop in FileWidget parses the data URL inappropriately, which also permits the use of `javascript:` schema. 
When the `javascript:` schema is set, and `ui:filePreview` is `true`, and when the user clicks `Preview` button, the JavaScript code is evaluated, and then allows XSS attack.

I fixed it by: 
- parsing the data URL correctly
- throwing `Error` when the URL is invalid
- restricting image types from any image to JPEG, PNG, GIF

### Checklist

### DIFF
--- a/packages/core/src/components/widgets/FileWidget.tsx
+++ b/packages/core/src/components/widgets/FileWidget.tsx
@@ -126,17 +126,26 @@ function FilesInfo<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends F
 }
 
 function extractFileInfo(dataURLs: string[]): FileInfoType[] {
-  return dataURLs
-    .filter((dataURL) => dataURL)
-    .map((dataURL) => {
+  return dataURLs.reduce((acc, dataURL) => {
+    if (!dataURL) {
+      return acc;
+    }
+    try {
       const { blob, name } = dataURItoBlob(dataURL);
-      return {
-        dataURL,
-        name: name,
-        size: blob.size,
-        type: blob.type,
-      };
-    });
+      return [
+        ...acc,
+        {
+          dataURL,
+          name: name,
+          size: blob.size,
+          type: blob.type,
+        },
+      ];
+    } catch (e) {
+      // Invalid dataURI, so just ignore it.
+      return acc;
+    }
+  }, [] as FileInfoType[]);
 }
 
 /**
--- a/packages/core/src/components/widgets/FileWidget.tsx
+++ b/packages/core/src/components/widgets/FileWidget.tsx
@@ -72,7 +72,7 @@ function FileInfoPreview<T = any, S extends StrictRJSFSchema = RJSFSchema, F ext
   // If type is JPEG, PNG, or GIF, then show image preview.
   // Originally, any type of image was supported, but this was changed into a whitelist
   // since SVGs are also images, which is generally considered a security risk.
-  if (['image/jpeg', 'image/png', 'image/gif'].includes(type)) {
+  if (['image/jpeg', 'image/png'].includes(type)) {
     return <img src={dataURL} style={{ maxWidth: '100%' }} className='file-preview' />;
   }
 
--- a/packages/core/src/components/widgets/FileWidget.tsx
+++ b/packages/core/src/components/widgets/FileWidget.tsx
@@ -69,9 +69,9 @@ function FileInfoPreview<T = any, S extends StrictRJSFSchema = RJSFSchema, F ext
     return null;
   }
 
-  // If type is JPEG, PNG, or GIF, then show image preview.
+  // If type is JPEG or PNG then show image preview.
   // Originally, any type of image was supported, but this was changed into a whitelist
-  // since SVGs are also images, which is generally considered a security risk.
+  // since SVGs and animated GIFs are also images, which are generally considered a security risk.
   if (['image/jpeg', 'image/png'].includes(type)) {
     return <img src={dataURL} style={{ maxWidth: '100%' }} className='file-preview' />;
   }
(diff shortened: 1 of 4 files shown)


packages/core/src/components/widgets/FileWidget.tsx
```diff
@@ -69,10 +69,15 @@ function FileInfoPreview<T = any, S extends StrictRJSFSchema = RJSFSchema, F ext
     return null;
   }
 
-  if (type.indexOf('image') !== -1) {
+  // If type is JPEG, PNG, or GIF, then show image preview.
+  // Originally, any type of image was supported, but this was changed into a whitelist
+  // since SVGs are also images, which is generally considered a security risk.
+  if (['image/jpeg', 'image/png', 'image/gif'].includes(type)) {
```
heath-freenome, line 75: Animated GIFs are also a security risk... Maybe we only show the first two

packages/utils/src/dataURItoBlob.ts
```diff
@@ -4,39 +4,36 @@
...
+  // check if the dataURI is base64
+  if (splitted.length !== 2) {
+    throw new Error('File is invalid: dataURI must be base64');
+  }
+  const [media, base64] = splitted;
+  const [mime, ...mediaparams] = media.split(';');
+  const type = mime || '';
+  const name = decodeURI(
+    mediaparams.map((param) => param.split('=')).find(([key]) => key === 'name')?.[1] || 'unknown'
+  );
```
heath-freenome, line 23: Can you add documentation around what this is doing

packages/utils/src/dataURItoBlob.ts
```diff
@@ -4,39 +4,36 @@
...
-      array.push(binary.charCodeAt(i));
+      array[i] = binary.charCodeAt(i);
     }
     // Create the blob object
     const blob = new window.Blob([new Uint8Array(array)], { type });
 
     return { blob, name };
   } catch (error) {
-    return { blob: { size: 0, type: (error as Error).message }, name: dataURI };
+    throw new Error('File is invalid: failed to decode base64');
```
heath-freenome, line 37: Why not just treat all of the errors to return a blob like these?

heath-freenome:
@yuki-js your build in `@rjsf/utils` is failing.

packages/utils/src/dataURItoBlob.ts
yuki-js: The original code treated the invalid file as zero-sized file. Of course, the invalid file is different from the zero-sized valid file. Invalid files should have been taken care of by the callee, but it is not.
This behavior also allowed `javascript:` schema.

yuki-js:
@heath-freenome Thank you for your review. 
This failure is the same as the one we faced previously. Could you rerun it?
https://github.com/rjsf-team/react-jsonschema-form/actions/runs/7474848380/attempts/3

heath-freenome (commented):
Sorry, I forgot to submit these comments over the weekend

packages/core/src/components/widgets/FileWidget.tsx
```diff
@@ -121,17 +126,26 @@ function FilesInfo<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends F
...
+        ...acc,
+        {
+          dataURL,
+          name: name,
+          size: blob.size,
+          type: blob.type,
+        },
+      ];
+    } catch (e) {
+      // Invalid dataURI, so just ignore it.
```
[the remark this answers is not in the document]
yuki-js: **Possibility of Exception:** When the ordinary user uploads files in ordinary way, the exception should never happen, since the data URL is always returned, and Base64 encoding/decoding always be successful. Therefore the data should not disappear in an unintuitive way.

**Error notification:** If the caller injects the malformed string into `formData`, validator rejects it with `.files.0 must match format "data-url"`, but it would be an issue on the side of the caller.


Although I could also change it by adding error flag in `FileInfoType`, I decided not to notify the explicit error for this reason.

heath-freenome (approved):

PR #4103, opened 2024-02-23
Fix: Error state not resetting when schema changes (#4079)

### LINKED ISSUE #4079: Is it possible to clear form errors programatically?
### Prerequisites

### What theme are you using?

core

### What is your question?

I use `rjsf` in form wizard and depending on step I change schema in the form. 
Wizard allows to move to different steps even if previous forms was not filled or with errors. If step has some errors, values with errors are not populated to the form data.

When

1. I run validation on wizard step **A** and some errors are shown
2. I go to another step **B**, schema is changed and different form without errors is displayed
3. I return back to the step **A**

the errors are still visible

But I need clear errors before (or after) moving to another step

So is it possible to clear errors programatically?

abdalla-rko:

Fixes #4079

### Checklist

### DIFF
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -382,7 +382,9 @@ export default class Form<
     const _retrievedSchema = retrievedSchema ?? schemaUtils.retrieveSchema(schema, formData);
 
     const getCurrentErrors = (): ValidationData<T> => {
-      if (props.noValidate) {
+      // f the `props.novalidate` option is set, we reset the error state.
+      // Or if the `retrievedSchema` param is undefined, this indicates that the schema has changed. Therefore, we also reset the error state.
+      if (props.noValidate || typeof retrievedSchema === 'undefined') {
         return { errors: [], errorSchema: {} };
       } else if (!props.liveValidate) {
         return {
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -317,7 +317,6 @@ export default class Form<
           : this.state.retrievedSchema,
         isSchemaChanged
       );
-      console.log('nextState _retrieved', nextState.retrievedSchema);
       const shouldUpdate = !deepEquals(nextState, prevState);
       return { nextState, shouldUpdate };
     }
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -307,14 +307,14 @@ export default class Form<
   ): { nextState: FormState<T, S, F>; shouldUpdate: true } | { shouldUpdate: false } {
     if (!deepEquals(this.props, prevProps)) {
       const isSchemaChanged = !deepEquals(prevProps.schema, this.props.schema);
+      const isFormDataChanged = !deepEquals(prevProps.formData, this.props.formData);
       const nextState = this.getStateFromProps(
         this.props,
         this.props.formData,
         // If the `schema` has changed, we need to update the retrieved schema.
-        // Or if the `formData` changes, for example in the case of a schema with dependencies that need to match one of the subSchemas, the retrieved schema must be updated.
-        isSchemaChanged || !deepEquals(prevProps.formData, this.props.formData)
-          ? undefined
-          : this.state.retrievedSchema,
+        // Or if the `formData` changes, for example in the case of a schema with dependencies that need to
+        //  match one of the subSchemas, the retrieved schema must be updated.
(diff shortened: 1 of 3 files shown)


packages/core/test/Form.test.jsx
```diff
@@ -4013,6 +4013,41 @@ describe('Form omitExtraData and liveOmit', () => {
...
+        properties: {
+          foo: { type: 'string' },
+        },
+      };
+
+      const extraErrors = {
+        foo: {
+          __errors: ['foo'],
+        },
+      };
```
heath-freenome, line 4029: Rather than setting extra errors, might I suggest making foo required, not providing a form value, clicking submit, seeing that there is an error, and changing it to be optional and then verifying the error was cleared?

packages/core/src/components/Form.tsx
```diff
@@ -306,10 +306,16 @@ export default class Form<
     prevState: FormState<T, S, F>
   ): { nextState: FormState<T, S, F>; shouldUpdate: true } | { shouldUpdate: false } {
     if (!deepEquals(this.props, prevProps)) {
+      const isSchemaChanged = !deepEquals(prevProps.schema, this.props.schema);
       const nextState = this.getStateFromProps(
         this.props,
         this.props.formData,
-        prevProps.schema !== this.props.schema ? undefined : this.state.retrievedSchema
+        // If the `schema` has changed, we need to update the retrieved schema.
+        // Or if the `formData` changes, for example in the case of a schema with dependencies that need to match one of the subSchemas, the retrieved schema must be updated.
```
heath-freenome, line 314:
```suggestion
        // Or if the `formData` changes, for example in the case of a schema with dependencies that need to
        // match one of the subSchemas, the retrieved schema must be updated.
```

packages/core/src/components/Form.tsx
```diff
@@ -306,10 +306,16 @@ export default class Form<
     prevState: FormState<T, S, F>
   ): { nextState: FormState<T, S, F>; shouldUpdate: true } | { shouldUpdate: false } {
     if (!deepEquals(this.props, prevProps)) {
+      const isSchemaChanged = !deepEquals(prevProps.schema, this.props.schema);
```
heath-freenome, line 309:
```suggestion
    const isSchemaChanged = !deepEquals(this.props.schema, prevProps.schema);
    const isPropsChanged = !deepEquals(this.props, prevProps)
    if (isSchemaChanged || isPropsChanged) {
```

packages/core/src/components/Form.tsx
```diff
@@ -382,7 +395,8 @@ export default class Form<
     const _retrievedSchema = retrievedSchema ?? schemaUtils.retrieveSchema(schema, formData);
 
     const getCurrentErrors = (): ValidationData<T> => {
-      if (props.noValidate) {
+      // If the `props.noValidate` option is set or the schema has changed, we reset the error state.
+      if (props.noValidate || isSchemaChanged) {
```
nickgros, line 399: Should we validate against the new schema instead of clearing the errors if `mustValidate` is true?
```suggestion
      if (props.noValidate || (isSchemaChanged && !mustValidate)) {
```

packages/core/src/components/Form.tsx
heath-freenome, line 309:
```suggestion
    if (!deepEquals(this.props, prevProps)) {
      const isSchemaChanged = !deepEquals(prevProps.schema, this.props.schema);
      const isPropsChanged = !deepEquals(prevProps.formData, this.props.formData);
```

packages/core/src/components/Form.tsx
```diff
@@ -306,10 +306,16 @@ export default class Form<
...
       const nextState = this.getStateFromProps(
         this.props,
         this.props.formData,
-        prevProps.schema !== this.props.schema ? undefined : this.state.retrievedSchema
+        // If the `schema` has changed, we need to update the retrieved schema.
+        // Or if the `formData` changes, for example in the case of a schema with dependencies that need to match one of the subSchemas, the retrieved schema must be updated.
+        isSchemaChanged || !deepEquals(prevProps.formData, this.props.formData)
+          ? undefined
+          : this.state.retrievedSchema,
+        isSchemaChanged
```
heath-freenome, line 318:
```suggestion
        // If the `schema` has changed, we need to update the retrieved schema.
        // Or if the `formData` changes, for example in the case of a schema with dependencies that need to
        //  match one of the subSchemas, the retrieved schema must be updated.
        isSchemaChanged || isFormDataChanged ? undefined  : this.state.retrievedSchema,
        isSchemaChanged
```

nickgros (approved):

PR #4117, opened 2024-03-06
Fixes: Warning: validateDOMNesting(...): <p> cannot appear as a desce…

orenf:
Fixes: Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.

### DIFF
--- a/packages/chakra-ui/src/DescriptionField/DescriptionField.tsx
+++ b/packages/chakra-ui/src/DescriptionField/DescriptionField.tsx
@@ -12,7 +12,7 @@ export default function DescriptionField<
 
   if (typeof description === 'string') {
     return (
-      <Text id={id} mt={2} mb={4}>
+      <Text as='sup' fontSize='md' id={id}>
         {description}
       </Text>
     );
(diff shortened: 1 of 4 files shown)


packages/chakra-ui/src/DescriptionField/DescriptionField.tsx
```diff
@@ -12,7 +12,7 @@ export default function DescriptionField<
 
   if (typeof description === 'string') {
     return (
-      <Text id={id} mt={2} mb={4}>
+      <Text as='sup' fontSize='md' id={id}>
```
heath-freenome, line 15: Why remove the margins? Does the `sup` change things that much?

heath-freenome:
@orenf Can you provide a before/after screenshots?

packages/chakra-ui/src/DescriptionField/DescriptionField.tsx
orenf: Sup with Added Margin
Sup No Margin


Keeping the margins did not seem to change much. Let me know if I should add back in.

heath-freenome (approved):

PR #2754, opened 2022-03-09
Use require for every mui component and streamline context usage

rollercodester:
### Reasons for making this change

Material-UI tree-shaking does not work when requiring an entire package. Each MUI component should be required separately to ensure that the entire MUI library is not unnecessarily bundled (i.e. only require what is actually used).

Also abstracted context usage into a custom hook, which is a popular pattern in commercial React dev.

fixes #2748

PR #4132, opened 2024-03-16
Fix: use correct ConfigProvider context by replacing deep antd imports with named imports

### LINKED ISSUE #4129: Widgets unusable under Antd v5 using a dark theme (wrong ConfigProvider context passed down)
### Prerequisites

### What theme are you using?

antd

### Version

5.x

### Current Behavior

When using Antd v5 in combination with a dark theme, the widgets provided by rjsfs/antd are unusuable:
image

### Expected Behavior

Colors, background colors etc. should match the selected antd theme.

### Steps To Reproduce
Just use the RJFS/antd widgets in combination with a dark theme. For convenience you'll find different ways to easily reproduce the error.

#### Using stackblitz:
https://stackblitz.com/~/github.com/momesana/rjsf-darktheme-bug. You may have to use `pnpm run dev` from StackBlitz's terminal environment to launch it.

#### Using a github repository (using a webpack based custom starter using `pnpm`)
Here is a repository to highlight the issue: https://github.com/momesana/rjsf-darktheme-bug. Clone the repo and then issue `pnpm install` followed by `pnpm run dev` which should open the developoment environment on localhost:8080.

#### Using vite and npm
Alternatively, if you don't like the opinionated custom webpack-based starter above, you can achieve the same with vite by:
1. running `npm create vite@latest rjsf-darktheme-bug -- --template react-ts`
2. change into rjsf-darktheme-bug and install deps: `cd rjsf-darktheme-bug && npm install`
2. now install rjsf, antd and other required packages: `npm install --save @rjsf/core @rjsf/utils @rjsf/validator-ajv8 @rjsf/antd antd dayjs @ant-design/icons`
3. changing the content of `src/App.tsx` to match that of https://github.com/momesana/rjsf-darktheme-bug/blob/main/src/App.tsx either by manually replacing the content or running `curl https://raw.githubusercontent.com/momesana/rjsf-darktheme-bug/main/src/App.tsx -o src/App.tsx` from within the project root folder
4. lastly run `npm run dev`

In short (with `npm` and `curl` being available):
```bash
cd /tmp && \
npm create vite@latest rjsf-darktheme-bug -- --template react-ts && \
cd rjsf-darktheme-bug && \
npm install && \
npm install --save @rjsf/core @rjsf/utils @rjsf/validator-ajv8 @rjsf/antd antd dayjs @ant-design/icons && \
curl https://raw.githubusercontent.com/momesana/rjsf-darktheme-bug/main/src/App.tsx -o src/App.tsx && \
npm run dev
```
 
### Environment

```markdown
- OS: linux
- Node: v20LTS
- npm: actually pnpm v8
```

### Anything else?
When I clone the rjsf repository and change the way the antd form components are imported from deep imports to named imports as follows, then build rjsfs and link it with `npm link <path-to-the-self-built-rjfs-antd-package>` the issue goes away:

```diff
...
-import Button from 'antd/lib/button';
-import Col from 'antd/lib/col';
-import Row from 'antd/lib/row';
+import { Button, Col, Row } from 'antd';
...
```
The result being this (with the primaryColor set to crimson red):
image

I don't have much experience with building libraries and the requirements it entails and I assume the imports where the way they were for a particular reason (facilitating tree-shaking etc.) though [according to this post](https://github.com/ant-design/ant-design/discussions/42660#discussioncomment-6011427) on the antd github repo using the named exports i. e. importing directly from 'antd' is the preferred way to handle this. [This answer here also seems to corroborate it](https://github.com/ant-design/ant-design/issues/44446) and is from one of the active antd developers.

In the case that the deep imports weren't that intentional (and also for anyone interested in testing out the proposed fix as a means to get to the bottom of things) [here is a PR](
https://github.com/rjsf-team/react-jsonschema-form/pull/4132) with the above mentioned changes.

Something peculiar I noticed that I can't yet wrap my head around is that using stackblitz without webpack/vite etc. seems to work fine:
https://stackblitz.com/edit/react-cnrcej?file=demo.tsx,index.html
I assume this is due to Stackblitz directly importing the es modules instead of bundling things as webpack and vite do. It's just conjecture though.

**Edit:**
I've just created a very simple vite project with just antd in it using a dark theme and a single Antd Button. As soon as I switch to a default deep import via `import Button from 'antd/lib/button';` as opposed to `import { Button } from 'antd';` the theming stops working, so either it's an antd bug or these deep imports aren't meant to be used.

**2nd Edit:**
After reading [this](https://github.com/ant-design/ant-design/discussions/42660#discussioncomment-6011427) and [this](https://github.com/ant-design/ant-design/issues/44446), I assume the named exports are supposed to be used instead of the deep default imports. To clarify this, I've also started [this discussion thread](https://github.com/ant-design/ant-design/discussions/47927) on the antd github project.

**3rd Edit:**
Ok, the reason why this doesn't work is that the used components must have a `useContext(C)` or a ContextConsumer that must be identical to the one used to create the context. The only way to make the inputs receive the correct theming data from the context is to also import the Context via a deep import i.e. `import ConfigProvider from "antd/lib/config-provider";`.
This evidently isn't practical as client code will most likely have used named imports for importing the antd components and thus the code really needs to be fixed in react-json-schema-forms.

I've updated the above PR and removed all deep imports and rewritten the part that used the `ConfigConsumer` to use the `ConfigProvider.ConfigContext` so that there is no more dependence on any unexposed antd internals. Needless to say, the theming now would break for client code that uses deep imports but that shouldn't exist and thus rather constitutes a bug in the client code that we don't need to account for here.

momesana:

fixes #4129

### Checklist

added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Added a new `skipEmptyDefault` option in `emptyObjectFields`, fixing [#3880](https://github.com/rjsf-team/react-jsonschema-form/issues/3880)
 
+## @rjsf/antd
+- Fix issue where the theme provided by the ConfigProvider under antd v5 wasn't respected thereby rendering the form items unusable under dark themes [#4129](https://github.com/rjsf-team/react-jsonschema-form/issues/4129)
+
 ## Dev / docs / playground
 
 - Updated the documentation to describe how to use the `skipEmptyDefault` option.
--- a/packages/antd/src/templates/ArrayFieldItemTemplate/index.tsx
+++ b/packages/antd/src/templates/ArrayFieldItemTemplate/index.tsx
@@ -1,6 +1,4 @@
-import Button from 'antd/lib/button';
-import Col from 'antd/lib/col';
-import Row from 'antd/lib/row';
+import { Button, Col, Row } from 'antd';
 import { ArrayFieldTemplateItemType, FormContextType, RJSFSchema, StrictRJSFSchema } from '@rjsf/utils';
 
 const BTN_GRP_STYLE = {
--- a/packages/antd/src/templates/ArrayFieldTemplate/index.tsx
+++ b/packages/antd/src/templates/ArrayFieldTemplate/index.tsx
@@ -9,9 +9,8 @@ import {
   StrictRJSFSchema,
 } from '@rjsf/utils';
 import classNames from 'classnames';
-import Col from 'antd/lib/col';
-import Row from 'antd/lib/row';
-import { ConfigConsumer, ConfigConsumerProps } from 'antd/lib/config-provider/context';
+import { Col, Row, ConfigProvider } from 'antd';
+import { useContext } from 'react';
 
 const DESCRIPTION_COL_STYLE = {
   paddingBottom: '8px',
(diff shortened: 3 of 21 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Added a new `skipEmptyDefault` option in `emptyObjectFields`, fixing [#3880](https://github.com/rjsf-team/react-jsonschema-form/issues/3880)
 
+## @rjsf/antd
+- Fix issue where the theme provided by the ConfigProvider under antd v5 wasn't respected thereby rendering the form items unusable under dark themes [#4129](https://github.com/rjsf-team/react-jsonschema-form/issues/4129)
+
```
heath-freenome, line 31: Can you add this above `@rjsf/core` above to keep it alphabetical?

heath-freenome:
@momesana Also, can you verify these changes don't break things with `antd` v4

heath-freenome (approved):

momesana:
Thanks @heath-freenome for merging the changes. I noticed your message too late.

sebastibe:
I was about to create a PR to revert part of these changes as it seems reverting to named import has broken tree shaking and the whole ant design lib is now imported into the bundles unsing `@antd/rjsf`.

For reference:
- ant design without tree shaking is not recommended as it add 1.2MB to your bundle (compressed)
- internally ant design is also using deep imports to enable tree shaking
- other rjsf integrated libraries like `mui` are also following the same pattern as per #2754 
- although there are work-arounds for developers who want to import the whole antd library to use the deep imports form as described in the original issue, the opposite is not true, so the deep import approach is more open.

momesana:
> I was about to create a PR to revert part of these changes as it seems reverting to named import has broken tree shaking and the whole ant design lib is now imported into the bundles unsing `@antd/rjsf`.

antd should actually support treeshaking with modern named imports as it both exposes an [es module export](https://github.com/ant-design/ant-design/blob/23b6737ad7fb9b3930e28c3eb5496d89d56a1ef4/package.json#L33) and also explicitly states that only css files have [`sideEffects`](https://github.com/ant-design/ant-design/blob/23b6737ad7fb9b3930e28c3eb5496d89d56a1ef4/package.json#L29).

We should first investigate why treeshaking doesn't work with named exports before reverting the change and forcing everyone to do deep imports and employ weird workarounds to fix the issue (i.e. #4129) that initially prompted the above change.

sebastibe:
It does not work because not all bundlers supports ES module analysis. 

For example we are using a bundler based on Google Closure Compiler and it does not support it. Reverting to deep imports increase bundler compatibility.

momesana:
> It does not work because not all bundlers supports ES module analysis.
> 
> For example we are using a bundler based on Google Closure Compiler and it does not support it. Reverting to deep imports increase bundler compatibility.

There is a trade-off here. If you revert the change to decrease the bundle-size (which is a legitimate goal) you'll end up causing the initial issue. Every instance where rjfs makes use of context values passed down from the application will cause problems unless the application uses deep imports for the provider and all subsequent access to said context as well. That for me is a far graver issue than a larger bundle size which may very well be temporary as tree-shaking es modules -- at least in my humble opinion -- is a feature no modern bundler can afford to lack.

I'd rather make an issue for this in the upstream project (i.e. the closure compiler) and wait for it to be resolved. Prioritizing bundle size at the cost of making the library brittle and error-prone certainly doesn’t feel like the right trade-off to me.

sebastibe:
> That for me is a far graver issue than a larger bundle size which may very well be temporary as tree-shaking es modules -- at least in my humble opinion -- is a feature no modern bundler can afford to lack.

I don’t have leverage over bundler roadmaps—much like browser quirks, they’re outside our control. What we can control is how painful we make things for users.

With your PR, anyone who wants smaller bundles would effectively need to fork the library to get there. By contrast, for your use case, you can keep using this repo today by deep-importing the ConfigProvider (and any related context consumers). That’s a one-line change on your side versus a fork for everyone else.

Given that trade-off, asking the app to deep-import the provider is far simpler—and avoids pushing the maintenance burden onto all users.

heath-freenome:
@sebastibe Consider pushing a PR to implement you suggestions and allow @momesana a chance to try it

momesana:
> > That for me is a far graver issue than a larger bundle size which may very well be temporary as tree-shaking es modules -- at least in my humble opinion -- is a feature no modern bundler can afford to lack.
> 
> I don’t have leverage over bundler roadmaps—much like browser quirks, they’re outside our control. What we can control is how painful we make things for users.
> 
> With your PR, anyone who wants smaller bundles would effectively need to fork the library to get there. By contrast, for your use case, you can keep using this repo today by deep-importing the ConfigProvider (and any related context consumers). That’s a one-line change on your side versus a fork for everyone else.
> 
> Given that trade-off, asking the app to deep-import the provider is far simpler—and avoids pushing the maintenance burden onto all users.

I am surprised antd is bundled at all. It shouldn't be, as it is explicitly defined as a peerDependency. Building it into the production bundle defeats the purpose of having it as a peerDependency and can cause issues on the client side anyway, even if only a few js files are bundled along with the lib.

heath-freenome:
> > > That for me is a far graver issue than a larger bundle size which may very well be temporary as tree-shaking es modules -- at least in my humble opinion -- is a feature no modern bundler can afford to lack.
> > 
> > 
> > I don’t have leverage over bundler roadmaps—much like browser quirks, they’re outside our control. What we can control is how painful we make things for users.
> > With your PR, anyone who wants smaller bundles would effectively need to fork the library to get there. By contrast, for your use case, you can keep using this repo today by deep-importing the ConfigProvider (and any related context consumers). That’s a one-line change on your side versus a fork for everyone else.
> > Given that trade-off, asking the app to deep-import the provider is far simpler—and avoids pushing the maintenance burden onto all users.
> 
> I am surprised antd is bundled at all. It shouldn't be, as it is explicitly defined as a peerDependency. Building it into the production bundle defeats the purpose of having it as a peerDependency and can cause issues on the client side anyway, even if only a few js files are bundled along with the lib.

@momesana As far as I can tell by looking at the distribution libraries from @rjsf/antd, there is no inclusion of `antd` directly in the library. I'm guessing the bundler being used is doing that. If I had to guess, it may require configuring the bundler differently.

Issue #4202, opened 2024-05-25
Issue when resetting certain numbers with a default value


ayush-shta:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

5.x

### Current Behavior

Numbers not resetting when the number entered in that field ends with the default value provided for that field in the schema

### Expected Behavior

Values should reset as expected and should not depend upon what the default value is

### Steps To Reproduce

[...]

PR #4220, opened 2024-06-14
[fix] Resetting number fields should check the entire string when deciding to leave the input text alone (#4202)

nickgros:
The regular expression should check that the value matches at the beginning of the string

### DIFF
--- a/packages/core/src/components/fields/NumberField.tsx
+++ b/packages/core/src/components/fields/NumberField.tsx
@@ -71,7 +71,8 @@ function NumberField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends
     // Construct a regular expression that checks for a string that consists
     // of the formData value suffixed with zero or one '.' characters and zero
     // or more '0' characters
-    const re = new RegExp(`${value}`.replace('.', '\\.') + '\\.?0*$');
+    // const re = new RegExp(`${value}`.replace('.', '\\.') + '\\.?0*$');
+    const re = new RegExp(`^(${String(value).replace('.', '\\.')})?\\.?0*$`);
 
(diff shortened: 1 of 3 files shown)


packages/core/src/components/fields/NumberField.tsx
```diff
@@ -71,7 +71,8 @@ function NumberField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends
     // Construct a regular expression that checks for a string that consists
     // of the formData value suffixed with zero or one '.' characters and zero
     // or more '0' characters
-    const re = new RegExp(`${value}`.replace('.', '\\.') + '\\.?0*$');
+    // const re = new RegExp(`${value}`.replace('.', '\\.') + '\\.?0*$');
```
nickgros, line 74:
```suggestion
```

packages/core/src/components/fields/NumberField.tsx
```diff
@@ -71,7 +71,7 @@ function NumberField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends
     // Construct a regular expression that checks for a string that consists
     // of the formData value suffixed with zero or one '.' characters and zero
     // or more '0' characters
-    const re = new RegExp(`${value}`.replace('.', '\\.') + '\\.?0*$');
+    const re = new RegExp(`^(${String(value).replace('.', '\\.')})?\\.?0*$`);
```
nickgros, line 74: * Add `^` so the regular expression checks the entire string. In this bug, the value `1` matched `'231'` because this was missing.
* Wrap value in capture group and add `?` since the string representation of the value may not be present for strings like `'.001'`. This prevents the above change from breaking other cases.

heath-freenome (approved):

PR #4227, opened 2024-06-21
Handle undefined target in onBlur and onFocus handlers

heath-freenome:
There are some instances where the `onBlur` and `onFocus` handlers are receiving an undefined `target`, causing the following type error:
```
Cannot read properties of undefined (reading 'value')
```
- Updated all of the `onBlur()` and `onFocus()` handlers in all libraries to fix this

[Please describe them here]

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,43 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 5.18.5
 
+## @rfsf/antd
+
+- Updated widgets to handle undefined `target` in `onFocus` and `onBlur` handlers
+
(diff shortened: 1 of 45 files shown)


nickgros (approved):
It would be great to know what causes this to happen. I am a bit concerned about passing `false` to these functions when `target` is null, but it will probably be ok.

heath-freenome:
> It would be great to know what causes this to happen. I am a bit concerned about passing `false` to these functions when `target` is null, but it will probably be ok.

It really only seems to be an issue with the `SelectWidget` on `semantic-ui` but I'm not taking any changes

PR #4282, opened 2024-08-23
Bug: issue with dependencies computeDefaults (#4271)

### LINKED ISSUE #4271: How to get `rawErrors` for parent schema or `errorSchema` in custom BaseInputTemplate
### Prerequisites

### What theme are you using?

core

### What is your question?

I am specifying a custom `BaseInputTemplate` and `ObjectFieldTemplate` in my project. My schema is complex enough that I have a nested properties inside of a `allOf/if/then` schema. See example: https://stackblitz.com/edit/vitejs-vite-srg8et?file=src%2FApp.tsx

In the example above, when you submit without entering any information you will see validation failed. But the inputs will not be in an error state since we do not have anything in the `rawErrors` prop being passed to the `BaseInputTemplate`. 

How do I get the `rawErrors` from the parent schema (in this case it would be `usernameAndPassword`) in this field? Alternatively is it possible to pass the full `errorSchema` to the `BaseInputTemplate` instead? 

**note:** I understand I can get the `errorSchema` from `ErrorListTemplate` but this means I would need to pass it down via context which I don't think is ideal. 

If you switch the auth type to "SSO" and submit without entering information you will observe the input in a error state.

abdalla-rko:

fixes #4271 

Fix is to first get the defaults from the properties if set, to ensure the dependencies conditions are resolved based on it. Then we can get the dependencies defaults.

The issue was caused by the `**load**` method in the playground and not by the formData `editor` as I explained in #4271. The `**load**` method updates the `formData` state because it is initially set to `samples.Simple.formData` after the component has mounted. This triggers the form `componentDidUpdate` which will try to get `computedDefaults` again. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Updated the peer dependencies to `5.20.x` due to types and API changes in `@rjsf/utils`
 
+## @rjsf/utils
+
+- Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)
+
 # 5.20.0
 
 ## @rjsf/core 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,13 +18,15 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 5.20.2
 
-## Dev / docs / playground
-
-- Updated the peer dependencies to `5.20.x` due to types and API changes in `@rjsf/utils`
-
 ## @rjsf/utils
 
 - Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)
+
+# 5.20.1
+
+## Dev / docs / playground
+
+- Updated the peer dependencies to `5.20.x` due to types and API changes in `@rjsf/utils`
 
 # 5.20.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
-# 5.20.1
+# 5.20.2
 
 ## Dev / docs / playground
 
--- a/packages/utils/src/schema/getDefaultFormState.ts
+++ b/packages/utils/src/schema/getDefaultFormState.ts
@@ -141,12 +141,22 @@ function maybeAddDefaultToObject<T = any>(
 }
 
 interface ComputeDefaultsProps<T = any, S extends StrictRJSFSchema = RJSFSchema> {
+  /** Any defaults provided by the parent field in the schema */
   parentDefaults?: T;
+  /** The options root schema, used to primarily to look up `$ref`s */
   rootSchema?: S;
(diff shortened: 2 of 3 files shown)


packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -188,6 +188,166 @@ export function computeDefaults<T = any, S extends StrictRJSFSchema = RJSFSchema
   let schemaToCompute: S | null = null;
   let updatedRecurseList = _recurseList;
 
+  // Return the defaults if they are defined for object or array schemas.
+  const getDefaultBasedOnSchemaType = (): T | T[] | void => {
```
heath-freenome, line 192: Can we break this down into 3 functions?

```
getDefaultBasedOnSchemaType()
getObjectDefaults()
getArrayDefaults()
```

alfonsoar (approved):
Thanks for taking the time to work on this bug!

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Updated the peer dependencies to `5.20.x` due to types and API changes in `@rjsf/utils`
 
+## @rjsf/utils
+
+- Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)
+
```
heath-freenome, line 28: @abdalla-rko Can you make this be in  `# 5.20.2` section instead?

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -269,164 +276,238 @@ export function computeDefaults<T = any, S extends StrictRJSFSchema = RJSFSchema
...
+
+/** Computes the default value based on the schema type.
+ *
+ * @param validator - an implementation of the `ValidatorType` interface that will be used when necessary
+ * @param rawSchema - The schema for which the default state is desired
+ * @param {ComputeDefaultsProps} computeDefaultsProps - Optional props for this function
+ * @param defaults - Optional props for this function
+ * @returns - The default value based on the schema type if they are defined for object or array schemas.
+ */
+export function getDefaultBasedOnSchemaType<
```
heath-freenome, line 292: Can you move the `getObjectDefaults()` and `getArrayDefaults()` to be defined before this function in the file.

heath-freenome (commented):
Just a few little organizational changes and you are good.

CHANGELOG.md
```diff
@@ -16,12 +16,16 @@ should change the heading of the (upcoming) version to include a major version b
...
+# 5.20.2
 
 ## Dev / docs / playground
 
 - Updated the peer dependencies to `5.20.x` due to types and API changes in `@rjsf/utils`
 
+## @rjsf/utils
+
+- Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)
+
```
heath-freenome, line 28:
```suggestion
# 5.20.2

# 5.20.1

## @rjsf/utils

- Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)

## Dev / docs / playground

- Updated the peer dependencies to `5.20.x` due to types and API changes in `@rjsf/utils`

```

heath-freenome (approved):

Issue #4291, opened 2024-09-05
performance issue related to lodash/isEqual function


igorbrasileiro:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [ ] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

5.x

### Steps To Reproduce

1. Run the gist https://gist.github.com/igorbrasileiro/754a1a0fe7bb4f8a080ca2c0ed3426e0

the Gist was made with deno bench, but can be done with other js runtime.

### Environment

```markdown
- OS: macOS Version 14.6
- deno: 1.46.2
```

### Anything else?

The results of the benchmark, in all three cases, the fast-deep-equal is better than lodash/isEqual
![image](https://github.com/user-attachments/assets/81f44e04-1f90-4e7b-bb3c-433167a66c77)

igorbrasileiro:
I already have changed all isEqual occurrences in the RJSF https://github.com/igorbrasileiro/react-jsonschema-form/commit/274a0a5f419517f2e151e7335b9fd0d1af046e58. Can I apply this fix?

igorbrasileiro:
I did a test running the playground locally with a bigger schema and big formdata and had 20% improvement. Others tests could be done

Steps to test:
1. open the playground
2. clean schema, formdata and uischema
3. click on live validation
4. fill the [Schema](https://gist.github.com/igorbrasileiro/52383714fb0f4dbc15ee161360f2de06) and [formData](https://gist.github.com/igorbrasileiro/74e4befe9ea3b555e4c8581360450387)
5. wait form renders
6. click to start the browser profiling
7. click on input name (root_name)
8. press one key

version with fast-deep-equal: https://github.com/igorbrasileiro/react-jsonschema-form. This problem starts to be more serious when you have realtime forms with schema or data changes.

Results:
- at RJSF main: 549ms
![Pasted Graphic](https://github.com/user-attachments/assets/4f4ce5f1-0c24-4cc9-8b0f-6ecc3298d76e)

- with fast-deep-equal: 419ms
[...]

PR #4292, opened 2024-09-09
feat: improve deepEquals performance

igorbrasileiro:
This PR improves the RJSF performance when comparing objects deeply.
fixes https://github.com/rjsf-team/react-jsonschema-form/issues/4291

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)
+- use `fast-equals.createCustomEqual` instead of `lodash.isEqualWith`
+
+## @validator-ajv8
+
+- use `@rjsf/utils` deepEquals instead of `lodash.isEqual`
 
 # 5.20.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,13 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)
-- use `fast-equals.createCustomEqual` instead of `lodash.isEqualWith`
+- Updated `deepEquals()` to use `fast-equals.createCustomEqual()` instead of `lodash.isEqualWith()`, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291)
+  - Switched uses of `lodash.isEqual()` to `deepEquals()` in many of the utility functions as well
+
 
 ## @validator-ajv8
 
-- use `@rjsf/utils` deepEquals instead of `lodash.isEqual`
+- Use `@rjsf/utils` `deepEquals()` instead of `lodash.isEqual()` to improve performance, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291)
 
 # 5.20.1
 
--- a/packages/utils/src/deepEquals.ts
+++ b/packages/utils/src/deepEquals.ts
@@ -1,4 +1,22 @@
-import isEqualWith from 'lodash/isEqualWith';
+import { createCustomEqual, State } from 'fast-equals';
+
+function isFunctions(a: any, b: any) {
+  return typeof a === 'function' && typeof b === 'function';
+}
+
+const customDeepEqual = createCustomEqual({
+  createInternalComparator: (comparator: (a: any, b: any, state: State<any>) => boolean) => {
+    return (a: any, b: any, _idxA: any, _idxB: any, _parentA: any, _parentB: any, state: State<any>) => {
+      if (isFunctions(a, b)) {
+        // Assume all functions are equivalent
+        // see https://github.com/rjsf-team/react-jsonschema-form/issues/255
+        return true;
+      }
+
+      return comparator(a, b, state);
+    };
+  },
+});
 
 /** Implements a deep equals using the `lodash.isEqualWith` function, that provides a customized comparator that
  * assumes all functions are equivalent.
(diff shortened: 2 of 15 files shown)


packages/utils/src/deepEquals.ts
```diff
@@ -1,19 +1,33 @@
-import isEqualWith from 'lodash/isEqualWith';
+import { createCustomEqual, State } from 'fast-equals';
 
-/** Implements a deep equals using the `lodash.isEqualWith` function, that provides a customized comparator that
+function isFunctions(a: any, b: any) {
```
heath-freenome, line 3: Can you provide JSDocs for this function?

packages/utils/src/deepEquals.ts
```diff
@@ -1,19 +1,33 @@
...
+  return typeof a === 'function' && typeof b === 'function';
+}
+
+const customDeepEqual = createCustomEqual({
```
heath-freenome, line 7: Can you provide JSDocs for this function?

heath-freenome (commented):
Just some documentation and then I'll approve

igorbrasileiro:
Added the JSDoc for new functions.

heath-freenome:
Can you also fix the merge conflict in the `CHANGELOG.md`? Otherwise looks good! How is the performance of `fast-equals` vs `fast-deep-equals`?

igorbrasileiro:
I pointed in the [issue](https://github.com/rjsf-team/react-jsonschema-form/issues/4291#issuecomment-2338336546) the performance of the customDeepEqual. But the results
image

The deepEqualsFunction
```ts
const customDeepEqual = createCustomEqual({
  createInternalComparator: (
    comparator: (a: any, b: any, state: State<any>) => boolean,
  ) => {
    return (
      a: any,
      b: any,
      _idxA: any,
      _idxB: any,
      _parentA: any,
      _parentB: any,
      state: State<any>,
    ) => {
      if (typeof a === "function" && typeof b === "function") {
        // Assume all functions are equivalent
        // see https://github.com/rjsf-team/react-jsonschema-form/issues/255
        return true;
      }

      return comparator(a, b, state);
    };
  },
});
```

CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Fixes an issue with dependencies computeDefaults to ensure we can get the dependencies defaults [#4271](https://github.com/rjsf-team/react-jsonschema-form/issues/4271)
+- use `fast-equals.createCustomEqual` instead of `lodash.isEqualWith`
+
+## @validator-ajv8
+
+- use `@rjsf/utils` deepEquals instead of `lodash.isEqual`
```
heath-freenome, line 28:
```suggestion
- Updated `deepEquals()` to use `fast-equals.createCustomEqual()` instead of `lodash.isEqualWith()`, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291)
  - Switched uses of `lodash.isEqual()` to `deepEquals()` in many of the utility functions as well


## @validator-ajv8

- Use `@rjsf/utils` `deepEquals()` instead of `lodash.isEqual()` to improve performance, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291)
```

heath-freenome (approved):

benjdlambert:
We seem to be getting test failures for this upgrade with `maximum call stack exceeded` https://github.com/backstage/backstage/pull/26639 don't have time to put together a reproduction repo just now, but will try and get to it this week. Just a heads up for now.

igorbrasileiro:
@benjdlambert thank you for let me know about it. I will try to reproduce this error and fix it.

Does your props have circularity?

igorbrasileiro:
I discovered how to reproduce. I reverted the `deepEquals` fn and the problem happened, so the issue isn't the way the fast-equals compare, it was the refactor.

heath-freenome:
@igorbrasileiro Is it possible to programmatically fix it such that the `null` option is unnecessary?

igorbrasileiro:
@heath-freenome I don't have enough context about the JSON schema spec and where to start investigating this issue, but I could implement a fix. I think my issue is different from that of @benjdlambert . 

Regarding the issue related to this PR, I haven't found a way to reproduce it. Do you know or have a hint?

heath-freenome:
@igorbrasileiro I know that @nickgros is working on an issue related to this where certain schemas cause an infinite looping. I wonder if your deep comparison changes are affecting a situation where `isEqual()` returns a value that differs from `deepEquals()` thereby causing infinite looping. For your situation at least, if you could debug the before/after results to see how `isEqual()` vs `deepEquals()` causes the algorithms around infinite recursion to change, and then either fixing it (if you clearly introduced a bug) or reaching back out if it doesn't make sense to you...

igorbrasileiro:
Sure, that sounds good to me. What do you think about reverting this PR, then I try to find the issue and reimplement it with the fix?

heath-freenome:
> Sure, that sounds good to me. What do you think about reverting this PR, then I try to find the issue and reimplement it with the fix?
> 
> Edit: Would be good to know how to reproduce the issue.

@igorbrasileiro That makes sense... Can you push the PR and then also, work on fixing the problem you experienced. I agree understanding the problem would be super helpful. It's clear that you already have one use case to start with.

@benjdlambert Can you provide us with your schema as well to help us debug.

@nickgros If you  get a chance to figure out the infinite recursion issue you are working on, feel free to chime in.

benjdlambert:
The schemas in the tests were pretty trivial to be honest, I tried to replicate this in a small codesandbox, and think it might have been to do with the fact that the `formData` might have had some cyclical references in there.

There's an example here: https://codesandbox.io/p/sandbox/n6txcy so if you try and edit the form data it would break.

igorbrasileiro:
@benjdlambert thank you for the codesandbox!

Changing the onChange function from your example to follow the docs, the error stops.
```diff
        onChange={(e) => {
-         setFormState(e);
+         setFormState(e.formData);
        }}
```

igorbrasileiro:
@heath-freenome as it seems to be a wrong usage of the form, can I revert the revert #4300 ? I'm considering that is not allowed pass props with cyclic reference. If is allowed I can setup the customcomparator from fast-equals to check if cyclic reference.

benjdlambert:
@igorbrasileiro I'm not convinced that crashing because of circular references in any of the props is great to be honest. I wonder if there's a way to drop the circular references ahead of the diffing or something.

I mentioned in my reproduction `Granted this example is not really what you would actually do`. We use just `e.formData` 
but was still happening to us in our test cases, was just a quick way I could put together the error. 

Let me see if I can get some time today to dig into what is actually happening in the broken version, as when I was debugging our tests the props that we were using looked pretty simple and we're not using cyclical references to what I could see.

igorbrasileiro:
> We use just `e.formData` but this was still happening to us in our test cases, was just a quick way I could put together the error.

@benjdlambert can you share the schema and formData from your test? I only get `maximum call stack size exceeded` with a schema with loop, but it also crashed using the form in the previous version. 

formData: `{}`, schema:
```JSON
{
  "definitions": {
    "first": {
      "type": "object",
      "properties": {
        "name": { "type": "string" },
        "children": { "$ref": "#/definitions/second" }
      }
    },
    "second": {
      "type": "object",
      "properties": {
        "name2": { "type": "string" },
        "children": { "$ref": "#/definitions/first" }
      }
    }
  },
  "$ref": "#/definitions/first"
}
```

Issue #3258, opened 2022-11-22
"allOf" Support for the custom JSON schema keywords merging 


Dzuming:
### Prerequisites

- [X] I have read the [documentation](https://react-jsonschema-form.readthedocs.io/)

### What theme are you using?

core

### Is your feature request related to a problem? Please describe.

In a JSON schema, there is a possibility to use [if else then](https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else). I want to define a JSON schema with a custom keyword

```JSON
...
        "properties": {
          "hasDescription": {
            "type": "boolean"

          },
          "description": {
            "type": "string",
            "customProperty": {
              "isImportant": false
            }
          },
        }
...

"allOf": [
          {
            "if": {
              "properties": {
                "hasDescription": {
                  "const": true
                }
              }
            },
            "then": {
              "properties": {
                "description": {
                  "customProperty": {
                    "isImportant": true
                  }
                }
              }
            },
            "else": {}
          }
        ]
```

If all of the conditions are met in the form, there is a warning in the console, and values are not merged:

```
could not merge subschemas in allOf:
Error: No resolver found for key customProperty. You can provide a resolver for this keyword in the options, or provide a default resolver
```

### Describe the solution you'd like

RJSF library uses [json schema merge all](https://github.com/mokkabonna/json-schema-merge-allof) library under the hood to merge subschemas. There is an option to [add custom resolvers via options property ](https://github.com/mokkabonna/json-schema-merge-allof#options).

Now, `packages/utils/src/schema/retrieveSchema.ts`  code where merging happens looks like

```
 if (ALL_OF_KEY in schema) {
    try {
      resolvedSchema = mergeAllOf({
        ...resolvedSchema,
        allOf: resolvedSchema.allOf,
      }) as S;
    } catch (e) {
      console.warn("could not merge subschemas in allOf:\n" + e);
      const { allOf, ...resolvedSchemaWithoutAllOf } = resolvedSchema;
      return resolvedSchemaWithoutAllOf as S;
    }
  }
```

The idea is to be able to provide custom mergeAllOf resolvers

```
resolvedSchema = mergeAllOf({
        ...resolvedSchema,
        allOf: resolvedSchema.allOf,
      }, 
      {
          resolvers: customResolvers
) as S;
```

So the end result will be like

```
resolvedSchema = mergeAllOf({
        ...resolvedSchema,
        allOf: resolvedSchema.allOf,
      }, 
      {
[...]

PR #4308, opened 2024-09-20
experimental_customMergeAllOf

MarekBodingerBA:

**NOTE: This is just a draft PR; details are subject to discussion.**

In many cases, [json-schema-merge-allof](https://github.com/mokkabonna/json-schema-merge-allof) is a performance bottleneck for this library:
- There are a few related issues: https://github.com/rjsf-team/react-jsonschema-form/issues/2811. https://github.com/rjsf-team/react-jsonschema-form/issues/3258 (this change allows to use custom keywords for the original library).
- The library is unmaintained, with the last [release](https://www.npmjs.com/package/json-schema-merge-allof?activeTab=versions) on 22/01/2021 (4 years ago).
- The library [deep clones](https://github.com/mokkabonna/json-schema-merge-allof/blob/ea2e48ee34415022de5a50c236eb4793a943ad11/src/index.js#L312) every schema, and this cloning is called recursively for each call. In large production schemas, this is detrimental to performance.
- There are "new players on the block" such as [allof-merge](https://github.com/udamir/allof-merge) that [promise](https://github.com/udamir/allof-merge/tree/master?tab=readme-ov-file#benchmark) much better performance. 
- For our use case, as we only use limited features of JSON Schema, we can use a very fast [custom implementation](https://gist.github.com/MarekBodingerBA/5bb3aeda02d2255c7e1d023eac22168c) of merging. This largely solves the library's performance issues for us; `getDefaultFormState` (not just the merging itself) performs 7x faster in our implementation:
```
getDefaultFormState_original x 141 ops/sec ±2.65% (80 runs sampled)
getDefaultFormState_fastMergeAllOf x 987 ops/sec ±2.72% (77 runs sampled)
```

I have a couple of questions:
- Is this something that would be desirable to implement for others?
- I would like feedback on the naming and types of `experimental_customMergeAllOf`. Maybe it's better to provide the function wrapped in an object rather than the function itself.
- `schemaParser` (which is only used in `compileSchemaValidatorsCode`) relies internally on `retrieveSchemaInternal`. I chose not to add this option here; is it necessary to include it?

The functional change is located [here](https://github.com/rjsf-team/react-jsonschema-form/pull/4308/files#diff-03661af48370908512b3758c30bf0d7fb27a1bca19e5de5442f8ec9332c0e37aR476), other changes are just adding the argument where needed.

### Checklist


MarekBodingerBA:
When I think about it, I would implement it now as an object (that would be easier for future compatibility). I imagine something like this:
```
export type Experimental_MergeAllOfOptions<S extends StrictRJSFSchema = RJSFSchema> =
  | {
      merger: "default";
      overideOptions: Options; // json-schema-merge-allof options
    }
  | {
      merger: "custom";
      mergeFn: (schema: S) => S;
    }
  | {
      merger: "xyz"; // other library if added in the future
    };
```

WDYT?

heath-freenome:
@MarekBodinger I totally understand why you want to add this. We feel like exposing the `Options` for `json-schema-merge-allof` is a bad idea. Honestly, if [allof-merge](https://github.com/udamir/allof-merge) has better performance and doesn't break our exists tests, then we welcome a PR to swap it in. That said, making it an object just to provide a function seems like overkill. So making it be a merge function like you started is fine. Not sure whether we can/want to expose this in the playground.

nickgros:
Also linking this bug: https://github.com/rjsf-team/react-jsonschema-form/issues/2927 which may be fixed by an alternative/improved merge implementation

nickgros:
I tried to just plug in `allof-merge` and the only meaningful test failure was [this one in retrieveSchemaTest](https://github.com/rjsf-team/react-jsonschema-form/blob/4cb834266187d95a0421292c7bdb7746466a9d8c/packages/utils/test/schema/retrieveSchemaTest.ts#L940). It looks like `allof-merge` has a lot of customization options, so it may be worth spending a bit of time to see if the behavior can be tweaked to get it to work properly, and we could make it the default implementation (if you're interested in looking into it @MarekBodingerBA)

MarekBodingerBA:
I don't think it is a good to replace j`son-schema-merge-allof` with `allof-merge` without thorough testing. As a user of this library, I would be concerned about such a change.

The approach already in progress, which introduces an `experimental_customMergeAllOf` option, is more appropriate. This would allow interested users who understand the implications to provide their own custom implementation, where they could easily incorporate `allof-merge` if desired.

I've completed the work on this and appreciate the feedback received.

nickgros (approved):
LGTM!

nickgros:
@MarekBodingerBA can you resolve the conflicts? Likely because we merged #4312

MarekBodingerBA:
@nickgros Unfortunately, I don't have access to a computer for next three weeks. Is it something that you are able to do, or should we wait until I come back?

nickgros:
@MarekBodingerBA I don't have permission to update your branch. Could you give me permission (I think in this PR you can give me permission to modify only this branch). Otherwise I could either create a new PR or wait for you to have time. Up to you!

MarekBodingerBA:
@nickgros I resolved the conflicts.

Issue #3975, opened 2023-11-27
enableMarkdownInDescription does not work in parent elements


quazgar:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

material-ui

### Version

5.x / current playground

### Current Behavior

[...]

PR #3665, opened 2023-05-12
Markdown in description

dejanbelusic:
Added an option to use markdown in the field description fields. Used the library for markdown that was already in the project. Also had to update the FieldTemplate components for some of the themes because some were showing the `description` and others the `rawDescription` prop. Made it consistent, and now it is using the `description` and rendered the markdown along all the themes.

### Checklist

- [x] **I'm updating documentation**
  - [x] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests. I've run `npm run test:update` to update snapshots, if needed.
  - [x] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [x] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [x] **I'm adding a new feature**
  - [x] I've updated the playground with an example use of the feature

PR #4405, opened 2024-12-01
Markdown in descriptions

db-legacy:
### Update the feature to render markdown in the field description. 

In the fields of type `object` and type `boolean`, markdown will be rendered in the description when `enableMarkdownInDescription` is set to `true`.

Issue reporting this bug: https://github.com/rjsf-team/react-jsonschema-form/issues/3975
Previous PR to add the feature: https://github.com/rjsf-team/react-jsonschema-form/pull/3665

The result: 
Screenshot 2024-12-01 at 18 44 54

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 5.23.2
+
+## @rjsf/core
+
+- Fix default value population when switching between options in `MultiSchemaField` [#4375](https://github.com/rjsf-team/react-jsonschema-form/pull/4375). Fixes [#4367](https://github.com/rjsf-team/react-jsonschema-form/issues/4367)
+
+## @rjsf/validator-ajv8
+
+- Fixed issue where `ui:title` in anyOf/oneOf is not shown in error messages. Fixes [#4368](https://github.com/rjsf-team/react-jsonschema-form/issues/4368)
+
 # 5.23.1
 
 ## @rjsf/chakra-ui
@@ -63,7 +73,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Fix issue 'Maximum call stack size exceeded' with playground share with large content.
 
-# 5.22.3 
+# 5.22.3
 
 ## @rjsf/utils
 
@@ -139,7 +149,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/core
 
-- Updated `Form` to fix `focusOnError()` to support the ids that include dots, fixing [#4279](https://github.com/rjsf-team/react-jsonschema-form/issues/4279) 
+- Updated `Form` to fix `focusOnError()` to support the ids that include dots, fixing [#4279](https://github.com/rjsf-team/react-jsonschema-form/issues/4279)
 
 ## @rjsf/mui
 
@@ -167,7 +177,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 5.20.0
 
-## @rjsf/core 
+## @rjsf/core
 
 - Support allowing raising errors from within a custom Widget [#2718](https://github.com/rjsf-team/react-jsonschema-form/issues/2718)
 - Updated `ArrayField`, `BooleanField` and `StringField` to call `optionsList()` with the additional `UiSchema` parameter, fixing [#4215](https://github.com/rjsf-team/react-jsonschema-form/issues/4215) and  [#4260](https://github.com/rjsf-team/react-jsonschema-form/issues/4260)
@@ -185,7 +195,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 5.19.4
 
-## @rjsf/core 
+## @rjsf/core
 
 - Fix XSS when rendering schema validation errors [#4254](https://github.com/rjsf-team/react-jsonschema-form/issues/2718)
   - NOTE: This will have potential consequences if you are using the [translateString](https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/form-props/#translatestring) feature and are trying to render HTML. Switching to [Markdown](https://www.markdownguide.org/) will solve your problems.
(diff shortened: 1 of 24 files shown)


packages/utils/src/types.ts
```diff
@@ -683,7 +683,7 @@ export type ObjectFieldTemplateProps<
   /** A string value containing the title for the object */
   title: string;
   /** A string value containing the description for the object */
-  description?: string;
+  description?: string | JSX.Element;
```
heath-freenome, line 686: `ReactNode` is the better choice for this
```suggestion
  description?: ReactNode;
```

packages/core/src/components/widgets/CheckboxWidget.tsx
```diff
@@ -58,12 +63,18 @@ function CheckboxWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F exte
   );
   const description = options.description ?? schema.description;
 
+  const richDescription = uiOptions.enableMarkdownInDescription ? (
+    <Markdown options={{ disableParsingRawHTML: true }}>{description || ''}</Markdown>
+  ) : (
+    description || ''
+  );
+
```
heath-freenome, line 71: @dejanbelusic after looking around it turns out that every theme's `CheckboxWidget` has a similar description rendering logic in it. We suggest creating a new component in the `@rjsf/utils` package that wraps this logic as follows and then updating every `FieldDescriptionTemplate` to use it:

```tsx
export interface RichDescriptionProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any> {
  /** The description of the field being rendered */
  description: ReactNode;
  /** The uiSchema that was passed to field */
  uiSchema?: UiSchema<T, S, F>;
  /** The `registry` object */
  registry: Registry<T, S, F>;
}
export default function RichDescription<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>({ description, registry, uiSchema = {}}: RichDescriptionProps<T,S,F>) {
  const { globalUiOptions } = registry;
  const uiOptions = getUiOptions<T, S, F>(uiSchema, globalUiOptions);
  if (uiOptions.enableMarkdownInDescription) {
    return <Markdown options={{ disableParsingRawHTML: true }}>{description || ''}</Markdown>
 }
 return description;
}
```

With all the `FieldDescriptionTemplates` calling it similarly to the change that would be made in `@rjsf/core`'s implementation:
```
export default function DescriptionField<
  T = any,
>(props: DescriptionFieldProps<T, S, F>) {
  const { id, description, uiSchema, registry } = props;
  if (!description) {
    return null;
  }
  return (
   <div id={id} className='field-description'>
      <RichDescription description={description} uiSchema={uiSchema} registry={registry} />
    </div>
  );
}
```

dejanbelusic: Hi @heath-freenome, I see your point and I agree. I can create a new component `RichDescription` as suggested, place it in the `@rjsf/utils`, and then use it in every `DescriptionField` component. 

When looking into `DescriptionField` components, I can see that one in `@rjsf/antd` packages is wrapped in a `span`, the one in `@rjsf/bootstrap-4` is wrapped in two `div`s, the one in  `@rjsf/chakra-ui` package is wrapped in `Text` component... What I'm saying is every theme has similar, but different implementation. Do I leave them as is?

heath-freenome: @dejanbelusic I would leave the different implementations of `DescriptionField` alone except render the `RichDescription` in place of `description` directly.

dejanbelusic: Hi @heath-freenome, I commited the changes, please take a look if I understood you correctly. 

Also, was looking into updating the tests, but I'm getting some errors, can you give me some pointers on how to troubleshoot them?

packages/utils/src/components/RichDescription.tsx
heath-freenome, line 1: You'll need to add unit tests for this in `rjsf` utils to pass the 100% coverage requirement

CHANGELOG.md
```diff
@@ -21,11 +21,16 @@ should change the heading of the (upcoming) version to include a major version b
...
+- Updated `ObjectField` and `CheckboxWidget` to render markdown in the description when `enableMarkdownInDescription` is set to `true`, fixing [#3975](https://github.com/rjsf-team/react-jsonschema-form/issues/3975)
 
 ## @rjsf/validator-ajv8
 
 - Fixed issue where `ui:title` in anyOf/oneOf is not shown in error messages. Fixes [#4368](https://github.com/rjsf-team/react-jsonschema-form/issues/4368)
 
+## Dev / docs / playground
+
+- Fixed typo in `package.json`
+
```
heath-freenome, line 33: The package change is so small as to not require mentioning. What does want to be mentioned is that it is a new feature so requires `5.25.0` AND you want to mention the changes in each package.

heath-freenome (commented):
Looks good, some suggestions

packages/utils/src/types.ts
```diff
@@ -541,7 +541,7 @@ export type DescriptionFieldProps<T = any, S extends StrictRJSFSchema = RJSFSche
   /** The uiSchema object for this description field */
   uiSchema?: UiSchema<T, S, F>;
   /** The description of the field being rendered */
-  description: string | ReactElement;
+  description: string;
```
heath-freenome, line 544: I believe this wants to be `ReactNode` as well. Which is what is causing the `@rjsf/core` tests to fail

heath-freenome:
@dejanbelusic Thanks for getting this closer to the finish line. I've left you some suggestions and comments

heath-freenome:
@dejanbelusic How close are you to finishing this?

dejanbelusic:
Hi @heath-freenome, I didn't have much free time to work in this. I think I could find some time in April.

heath-freenome:
@db-legacy So I just reimplemented your fix in v6 with #4597

Issue #3838, opened 2023-08-22
FormData Does Not Update on Change for Dependent Properties


relero90:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

material-ui

### Version

5.12.0

### Current Behavior

I am struggling with a bug that seems inherent to the use of conditional logic to render different enum options for one property based on the value of a different property. In my current setup, essentially, property 1 = value 1 allows property 2 to have enum 1 or enum 2 while property 1 = value 2 only allows property 2 to have enum 2 .  My default values are property 1 = value 1 and property 2 = enum 1. 

The problem that I'm running into is that changing property 1's value to value 2 does not update property 2's value to enum 2 in the resulting formData (the form renders appropriately - it just doesn't match it's formData) and subsequently the field for property 2 acts basically as a read only field (i.e. selecting enum 2 - the only choice - from that dropdown doesn't update the formData value either).

I have created the schema using both if...then syntax and dependencies - both methods run into this same formData bug. 

### Expected Behavior

When updating property 1's value (da) to enum 2 (DA 50) in the form, the field for property 2 (t) should update its corresponding formData value to the only available enum 2.

### Steps To Reproduce

[...]

PR #4439, opened 2025-01-08
fix: incorrect formData with racing onChange invokes

DevTakao:
### Story

I was trying to do a workaround for issue of dependent values not being reset in `formData` when dependee changed (#3838). In my case, it was dependent enums with dynamic options. The workaround worked by triggering `onChange` from dependent widgets and setting the first option as default whenever `enumOptions` changed. During that, I noticed an issue where `formData` only gets the update from last widget in the rendering order.

The state update in `onChange` is not using the callback form which can handle racing setStates in order. This should allow workarounds to work properly while the root issue is being fixed. It is also a good practice to write setState in callback form generally.

### Checklist

### DIFF
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -705,7 +705,13 @@ export default class Form<
         errors: toErrorList(errorSchema),
       };
     }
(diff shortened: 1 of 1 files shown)


DevTakao:
Sorry, it is still not working so I close. Will try again.

Issue #3759, opened 2023-07-10
Object defaults for "additionalProperties" result in rows that cannot be deleted and have side effects when modified


ghost:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [x] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

mui

### Version

5.9.0

### Current Behavior

For an object type, default values are seemingly assumed to map to "properties" rather than "additionalProperties," or at the very least, mapping them to "additionalProperties" does not seem to work as expected. Take the following `JSONSchema` example:

```
{
  "title": "Test Defaults",
  "default": {
    "test-key": "test-value"
  },
  "description": "A simple form with additional properties example.",
  "type": "object",
  "additionalProperties": {
    "type": "string"
  }
}
```

[...]

PR #4440, opened 2025-01-08
fix 3759 - uneditable & permanent defaults with additional properties

Piotr-Debicki:

Fixes [#3759](https://github.com/rjsf-team/react-jsonschema-form/issues/3759)

Additional properties with defaults set currently creates a permanent key-value pair that is essentially uneditable.
Trying to edit the key creates an entirely new key value entry on blur while still keeping the default key-value pair. The default key-value pair is also not deletable. Solution here generates defaults for additional properties once on form initialization, and then skips regeneration on any form update

Something to keep in mind with this solution is that defaults for additional properties will not generate if custom formData is provided to the form - a way around that would be to keep track of something like a defaults generated state instead of relying on 'edit'

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ should change the heading of the (upcoming) version to include a major version b
 -->
 # 5.24.3
 
+## @rjsf/utils
+
+- Rollback [4446](https://github.com/rjsf-team/react-jsonschema-form/pull/4446) due to regression
+
 ## Dev / docs / playground
 - Fixed issue with selector, where validator was getting refreshed on clicking on anything in selector. [#4472](https://github.com/rjsf-team/react-jsonschema-form/pull/4472)
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -16648,15 +16648,6 @@
       "integrity": "sha512-VxPP4NqbUjj6MaAOafWeUn2cXWLcCtljklUtZf0Ind4XQ+QPtmA0b18zZy0jIQx+ExRVCR/ZQpBmik5lXshNsw==",
       "dev": true
     },
-    "node_modules/fast-equals": {
-      "version": "5.2.2",
-      "resolved": "https://registry.npmjs.org/fast-equals/-/fast-equals-5.2.2.tgz",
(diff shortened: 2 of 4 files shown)


heath-freenome:
@Piotr-Debicki what happens if you add additional properties to an object nested within another object instead?

```
schema = {
  type: 'object',
  nested: {
  properties: {
    baz: {
      type: 'string';
    },
    blah: {
      type: 'object',
      additionalProperties: {
        type: 'string',
      },
      default: {
        foo: 'bar',
      },
    }
  }
};
```

Piotr-Debicki:
> @Piotr-Debicki what happens if you add additional properties to an object nested within another object instead? Do your changes fix this? If not, you may need to improve the `getDefaultFormState()` function instead.
> 
> ```
> schema = {
>   type: 'object',
>   nested: {
>   properties: {
>     baz: {
>       type: 'string';
>     },
>     blah: {
>       type: 'object',
>       additionalProperties: {
>         type: 'string',
>       },
>       default: {
>         foo: 'bar',
>       },
>     }
>   }
> };
> ```

@heath-freenome From what I can tell these seem to be two separate issues. The issue this solution fixes is just the regeneration of defaults for additional properties on form change, (e.g. pressing the delete button on the default key will just regenerate the default key since it 'changed', making it a permanent fixture) - not the generation of the defaults themselves. I can take a look at the bug you presented when I get a bit more time, but these changes didn't have that specific bug in mind

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
+# 5.24.1
```
heath-freenome, line 18:
```suggestion
# 5.24.2
```

heath-freenome:
@Piotr-Debicki The main reason why I suggested the above test was that, I believe, the code fix belongs in the `getDefaultFormState()` function rather than in `Form` since the former deals with the recursion deeper into the schema (and would handle the outermost case as well). Does that make sense?

Piotr-Debicki:
@heath-freenome Gotcha, I'll take another look at it

PR #4509, opened 2025-03-04
Add playground examples for LayoutGrid for each theme

heath-freenome:

Updated the `playground` so that each theme has an example of the `LayoutGridField` usage
- In `@rjsf/utils` added a `browser` section to the `package.json` to support `getTestIds()` working properly in a browser
  - Updated the `getTestIds()` to refactor out `getTestIdsBrowser()` and export it from the new `indexBrowser.ts`
  - Updated `package.json` to build `indexBrowser.js` for each export type
- In `@rjsf/core` made a few changes to support the playground examples:
  - In `LayoutMultiSchemaField` added support for the hiding of labels as well as extracting the `placeholder` from the UiSchema
  - In `GridTemplate` removed the explicit adding of the `row` class, leaving that up to the users of the `UiSchema`
- In `@rjsf/fluentui-rc` made a few changes to support the playground examples:
  - In `GridTemplate` switched to using the `GridShim` displaying it only when `columns` or `rows` are provided, otherwise simply exporting a `div` due to the CSS Grid
  - In `RadioWidget` implemented support for `inline` styling of the radio elements
- In `playground` implemented the `Layout Grid` example with the following changes:
  - Updated the `vite.config.ts` to pull the `utils` classes from the `indexBrowser`
  - Added the new `UiSchemaForTheme` type, overriding the `uiSchema` for `Sample` to be `UiSchema` or `UiSchemaForTheme`
  - Renamed `Selector` to `SampleSelector`, adding the `selectedSample` as a prop, removing the use of the `current` state in favor of `selectedSample`
    - Also changed the `onSelected()` callback to take a `sampleName: string` rather than `data: any`
  - Updated `Header` to replace `load` with `onSampleSelected` adding `sampleName: string`, serializing `sampleName` on share
    - Also, passed `sampleName` as `selectedSample` and `onSampleSelected` to the `SampleSelector`
  - Updated `EditorProps` to take `data: any` for `onChange`
  - Updated `EditorsProps` to take a new `hasUiSchemaGenerator: boolean`, using it to provide a different title for the `UiSchema` section when true
  - Updated `Playground` add two new state variables, `uiSchemaGenerator` and `sampleName`
    - Updated `onThemeSelected()` to update the `uiSchema` to the one generated from the `uiSchemaGenerator(theme)` if that function is defined in state
    - Updated `load()` to take an additional `sampleName` prop, using it to set the currently selected sample and to set the `uiSchemaGenerator` if `samples[sampleName]` has one
      - Also, set the `uiSchema` to a generated one if the incoming `uiSchema` is a function and adding a undefined check for the `validator`
    - Added a new `onSampleSelected()` which calls `load()` with the selected sample and `sampleName`
  - Updated `DemoFrame` and `app.ts` to rename `material-ui-5` to `mui`
  - Added a new `layoutGrid.ts` and exported it from the `samples/index.ts`
### Checklist

### DIFF
--- a/CHANGELOG_v6.md
+++ b/CHANGELOG_v6.md
@@ -79,7 +79,7 @@ should change the heading of the (upcoming) version to include a major version b
   - `findFieldInSchema<T = undefined, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, path: string | string[], schema: S, formData?: T, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): FoundFieldType<S>`: Finds the field specified by the `path` within the root or recursed `schema`
   - `findSelectedOptionInXxxOf<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, schema: S, fallbackField: string,xxx: 'anyOf' | 'oneOf', formData?: T, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): S | undefined`: Finds the option that matches the selector field in the `schema` or undefined if nothing is selected
   - `getFromSchema<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, schema: S, path: string | string[], defaultValue: T | S, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): T | S`: Helper that acts like lodash's `get` but additionally retrieves `$ref`s as needed to get the path for schemas
-
+- Exported a `browser` version of the libraries with a different `getTestIds()`
 
 ## @rjsf/validator-ajv6
 
@@ -91,6 +91,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Updated the `custom-templates.md` documentation for the changes to the `ArrayFieldTemplateItem` and add the two new templates
 - Updated the `utility-functions.md` documentation to add the `buttonId()` function
 - Added the `v6.x upgrade guide.md` documentation
+- Updated the `playground` to add a `Layout Grid` example and made the selected example now be part of the shared export
 
 # 6.0.0-alpha.0
 
--- a/CHANGELOG_v6.md
+++ b/CHANGELOG_v6.md
@@ -79,7 +79,7 @@ should change the heading of the (upcoming) version to include a major version b
   - `findFieldInSchema<T = undefined, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, path: string | string[], schema: S, formData?: T, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): FoundFieldType<S>`: Finds the field specified by the `path` within the root or recursed `schema`
   - `findSelectedOptionInXxxOf<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, schema: S, fallbackField: string,xxx: 'anyOf' | 'oneOf', formData?: T, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): S | undefined`: Finds the option that matches the selector field in the `schema` or undefined if nothing is selected
   - `getFromSchema<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, schema: S, path: string | string[], defaultValue: T | S, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): T | S`: Helper that acts like lodash's `get` but additionally retrieves `$ref`s as needed to get the path for schemas
-- Exported a `browser` version of the libraries with a different `getTestIds()`
+- Exported a `browser` version of the libraries with a browser-safe version of `getTestIds()`
 
 ## @rjsf/validator-ajv6
 
--- a/packages/playground/src/app.tsx
+++ b/packages/playground/src/app.tsx
@@ -26,7 +26,6 @@ const validators: PlaygroundProps['validators'] = {
   AJV8_es: esV8Validator,
   AJV8_2019,
   AJV8_2020,
-  'AJV6 (deprecated)': v6Validator,
 };
 
 const themes: PlaygroundProps['themes'] = {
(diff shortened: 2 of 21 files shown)


CHANGELOG_v6.md
```diff
@@ -79,7 +79,7 @@ should change the heading of the (upcoming) version to include a major version b
...
   - `findSelectedOptionInXxxOf<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, schema: S, fallbackField: string,xxx: 'anyOf' | 'oneOf', formData?: T, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): S | undefined`: Finds the option that matches the selector field in the `schema` or undefined if nothing is selected
   - `getFromSchema<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(validator: ValidatorType<T, S, F>, rootSchema: S, schema: S, path: string | string[], defaultValue: T | S, experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>): T | S`: Helper that acts like lodash's `get` but additionally retrieves `$ref`s as needed to get the path for schemas
-
+- Exported a `browser` version of the libraries with a different `getTestIds()`
```
heath-freenome, line 82:
```suggestion
- Exported a `browser` version of the libraries with a browser-safe version of `getTestIds()`
```

jasny:
I dislike this implementation for the grid. For v5 the UI schema works for all themes. Custom ui options are ignored for other themes. That's not the case for this grid implementation. Additionally, it doesn't provide proper options for different screen sizes for themes , which is a must for a grid layout.

I would suggest assuming a 1/12th grid, and only allow setting the span to `1` to `12`. But allow setting it for `small`, `medium` and `large` screens. The fields can just fill up to 12 columns.

Limiting the options makes more sense. The grid should look decent, instead of focussing on finetuned customized styling. For that people can overwrite the template.

```
{
  "firstName": {
    "ui:autofocus": true,
    "ui:emptyValue": "",
    "ui:placeholder": "ui:emptyValue causes this field to always be valid despite being required",
    "ui:autocomplete": "family-name",
    "ui:enableMarkdownInDescription": true,
    "ui:description": "Make text **bold** or *italic*. Take a look at other options [here](https://markdown-to-jsx.quantizor.dev/).",
    "ui:grid": {
      "medium": 6,
      "large": 4
    }
  },
  "lastName": {
    "ui:autocomplete": "given-name",
    "ui:enableMarkdownInDescription": true,
    "ui:description": "Make things **bold** or *italic*. Embed snippets of `code`. <small>And this is a small texts.</small> ",
    "ui:grid": {
      "medium": 6,
      "large": 4
    }
  },
  "age": {
    "ui:widget": "updown",
    "ui:title": "Age of person",
    "ui:description": "(earth year)"
    "ui:grid": {
      "large": 4
    }
  },
  "bio": {
    "ui:widget": "textarea"
  },
  "password": {
    "ui:widget": "password",
    "ui:help": "Hint: Make it strong!",
    "ui:grid": {
      "medium": 6,
      "large": 6
    }
  },
  "telephone": {
    "ui:options": {
      "inputType": "tel"
    },
    "ui:grid": {
      "medium": 6,
      "large": 6
    }
  }
}
```

heath-freenome:
> I dislike this implementation for the grid. For v5 the UI schema works for all themes. Custom ui options are ignored for other themes. That's not the case for this grid implementation. Additionally, it doesn't provide proper options for different screen sizes for all themes, which is a must for a grid layout.
> 
> I would suggest assuming a 1/12th grid, and only allow setting the span to `1` to `12`. But allow setting it for `small`, `medium` and `large` screens. The fields can just fill up to 12 columns.
> 

@jasny Here's the issue I see with your approach; the Fluent and Chakra-UI are both using native CSS grids which do NOT have a 1-12 approach. And `antd` uses a 24 column grid. Granted we could use math in that theme's `GridTemplate` if we did go with the 1-12 approach.

At the end of the day, the UiSchema is about modifying the standard UI approach for RJSF to fit the specifics of the UI the developer wants. I imagine that 99% of people using RJSF do so with a single theme. So providing a way for users to fine tune the gridding to exactly what they want, I believe, is the best approach. Making it a least-common-denominator solution like you are suggesting won't help people make exactly the UIs they want. It likely will even hinder it. With this grid approach I've developed one can put any field into any position. Using my example... If you wanted to move the `address` information after the name and before the `birth_date` and `race`, you could.

In terms of providing "proper options for different screen sizes for all themes", again, that is up to the developer to build. All of the themes that use the 1-12 grid system support the screen breakpoints, which the developer can easily add to their UI schema. Yes, it likely will take some playing around with the fine grain details of the grid in the UiSchema for a while, but once the developer has it nailed they likely won't change it much.

All that said, I definitely want to chat more about your ideas around how you see it work. Perhaps we do add support for the simplest LCD approach with the new `ui:grid` which basically sets up the layout in terms of gridding the properties in the `schema` in the order they appear. You want everything in a 2 column grid in small and 4 column in large then just set the outer grid to be:

```
{
    'ui:field': 'LayoutGridField',
    'ui:layoutGrid': {
        large: 4,
        small: 2
    }
}
```

packages/playground/src/app.tsx
```diff
@@ -102,7 +101,7 @@ const themes: PlaygroundProps['themes'] = {
     stylesheet: '',
     theme: FluentUIRCTheme,
   },
-  'material-ui-5': {
+  mui: {
     stylesheet: '',
     theme: MuiV5Theme,
```
nickgros, line 106: could be renamed to just MuiTheme now

packages/utils/src/indexBrowser.ts
```diff
@@ -0,0 +1,144 @@
...
+import enumOptionsValueForIndex from './enumOptionsValueForIndex';
+import ErrorSchemaBuilder from './ErrorSchemaBuilder';
+import findSchemaDefinition from './findSchemaDefinition';
+import getDateElementProps, { type DateElementFormat } from './getDateElementProps';
+import getDiscriminatorFieldFromSchema from './getDiscriminatorFieldFromSchema';
+import getInputProps from './getInputProps';
+import getSchemaType from './getSchemaType';
+import getSubmitButtonOptions from './getSubmitButtonOptions';
+import getTemplate from './getTemplate';
+import { getTestIdsBrowser as getTestIds } from './getTestIds';
```
nickgros, line 23: Why doesn't `getTestIds` work in the browser?

heath-freenome: Browsers don't have `process` defined and caused a TypeError

packages/playground/src/app.tsx
heath-freenome: Good point, I will consider making a change to add `MuiTheme` and deprecating `MuiV5Theme`

nickgros: I just mean to change the import 

```
import { Theme as MuiV5Theme } from '@rjsf/mui';
```

to

```
import { Theme as MuiTheme } from '@rjsf/mui';
```

nickgros (approved):
LGTM, I agree with @jasny though that it's a good idea to explore augmenting this with a more simple/conventional API.

packages/playground/src/app.tsx
heath-freenome: Done

PR #4576, opened 2025-04-23
Added documentation for the Layout Grid feature and fixed a few issues

heath-freenome:

Added docs for the Layout Grid feature and fixed some issues
- Updated the `package.json` files to move the `types` definition to the start of each `exports` block to fix some npm warnings
- Updated the `LayoutMultiSchemaField` to wrap the `Widget` in a `FieldTemplate` to get the field title displayed properly in several themes
  - Updated the tests accordingly
  - This required regenerating snapshots in some themes
- Updated the documentation in the `LayoutGridField` to fix some issues
- Added the `LayoutGridField.md` file to the `api-reference` section
- Added the `layout-grid.md` file to the `usage` section to describe in detail how to use the `LayoutGridField`
- Updated the `v6` upgrade guide to add some small changes

### Checklist

### DIFF
--- a/packages/docs/docs/api-reference/LayoutGridField.md
+++ b/packages/docs/docs/api-reference/LayoutGridField.md
@@ -1,0 +1,267 @@
+# LayoutGridField
+
+The `LayoutGridField` will render a schema, uiSchema and formData combination out into a `GridTemplate` in the shape
+described in the uiSchema. To define the grid to use to render the elements within a field in the schema, provide in
+the uiSchema for that field the object contained under a `ui:ayoutGrid` element. E.g. (as a JSON object):
+
+```json
+{
+  "field1" : {
+    "ui:field": "LayoutGridField",
+    "ui:layoutGrid": {
+      "ui:row": { ... }
+    }
+  }
+}
+```
+
+See the [LayoutGridField usage](../usage/layout-grid.md) for more detailed walk-through on how to use the component. This page is a more
+dense version of the `uiSchema` APIs for faster reference.
+
+## ui:row
+
+The outermost level of a `LayoutGridField` is the `ui:row` that defines the nested rows, columns, and/or condition
+elements (i.e. "grid elements") in the grid. This definition is either a simple "grid elements" OR an object with
+native `GridTemplate` implementation specific props and a `children` array of "grid elements". E.g. (as JSON objects):
+
+Simple `ui:row` definition, without additional `GridTemplate` props:
+
+```json
+{
+  "ui:row": [
+    { "ui:row"|"ui:col"|"ui:columns"|"ui:condition": ... },
+    ...
+  ]
+}
+```
+
+Complex `ui:row` definition, with additional `GridTemplate` (this example uses `@mui/material/Grid2` native props):
+
+```json
+{
+  "ui:row": {
+    "spacing": 2,
+    "size": {
+      "md": 4
+    },
+    "alignContent": "flex-start",
+    "className": "GridRow",
+    "children": [
+        {
+          "ui:row"|"ui:col"|"ui:columns"|"ui:condition": ...
+        },
+        ...
+      ]
+    }
+  }
+```
+
+> NOTE: Special note about the native `className` prop values. All className values will automatically be looked up in
+> the `formContext.lookupMap` in case they have been defined using a CSS-in-JS approach. In other words, from the
+> example above, if the `Form` was constructed with a `layoutGridLookupMap` set to `{ GridRow: cssInJs.GridRowClass }`
+> then when rendered, the native `GridTemplate` will get the `className` with the value from
+> `cssInJs.GridRowClass`. This automatic lookup will happen for any of the "grid elements" when rendering with
+> `GridTemplate` props. If multiple className values are present, for example:
+> `{ className: 'GridRow GridColumn' }`, the classNames are split apart, looked up individually, and joined
+> together to form one className with the values from `cssInJs.GridRowClass` and `cssInJs.GridColumnClass`.
+
+## ui:col
+
+The `ui:col` grid element is used to specify the list of columns within a grid row. A `ui:col` element can take on
+several forms:
+
+1. a simple list of dotted-path field names within the root field;
+2. a list of objects containing the dotted-path field `name` any other props that are gathered into `ui:options` for the field;
+3. a list with a one-off `render` functional component with or without a non-field `name` identifier and any other to-be-spread props; and
+4. an object with native `GridTemplate` implementation specific props and a `children` array with 1) or 2) or even a nested `ui:row` or a `ui:condition` containing a `ui:row` (although this should be used carefully). E.g. (as JSON objects):
+
+Simple `ui:col` definition, without additional `GridTemplate` props and form 1 only children:
+
+```json
+{
+  "ui:col": ["innerField", "inner.grandChild", ...]
+}
+```
+
+Complicated `ui:col` definition, without additional `GridTemplate` props and form 2 only children:
+
+```json
+{
+  "ui:col": [
+    { "name": "innerField", "fullWidth": true },
+    { "name": "inner.grandChild", "convertOther": true },
+    ...
+  ]
+}
+```
+
+More complicated `ui:col` definition, without additional `GridTemplate` props and form 2 children, one being a
+one-off `render` functional component without a non-field `name` identifier
+
+```json
+{
+  "ui:col": [
+    "innerField",
+    {
+      "render": "WizardNavButton",
+      "isNext": true,
+      "size": "large"
+    }
+  ]
+}
+```
+
+Most complicated `ui:col` definition, additional `GridTemplate` props and form 1, 2 and 3 children (this example
+uses @mui/material/Grid2 native props):
+
+```json
+{
+  "ui:col": {
+    "size": { "md": 4 },
+    "className": "GridColumn",
+    "children": [
+      "innerField",
+      { "name": "inner.grandChild", "convertOther": true },
+      { "name": "customRender", "render": "CustomRender", "toSpread": "prop-value" },
+      { "ui:row|ui:condition": ... },
+      ...
+    ]
+  }
+}
+```
+
+> NOTE: If a `name` prop does not exist or its value does not match any field in a schema, then it is assumed to be a
+> custom `render` component. If the `render` prop does not exist, a null render will occur. If `render` is a
+> string, its value will be looked up in the `formContext.lookupMap` first before defaulting to a null render. If the
+> resulting looked up `render` is not a function (React components are functions), then it will be a null render.
+
+## ui:columns
+
+The `ui:columns` grid element is syntactic sugar to specify a set of `ui:col` columns that all share the same set of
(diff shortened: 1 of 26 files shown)


packages/docs/docs/api-reference/LayoutGridField.md
```diff
@@ -0,0 +1,267 @@
+# LayoutGridField
+
+The `LayoutGridField` will render a schema, uiSchema and formData combination out into a `GridTemplate` in the shape
+described in the uiSchema. To define the grid to use to render the elements within a field in the schema, provide in
+the uiSchema for that field the object contained under a `ui:ayoutGrid` element. E.g. (as a JSON object):
```
nickgros, line 5:
```suggestion
the uiSchema for that field the object contained under a `ui:layoutGrid` element. E.g. (as a JSON object):
```

packages/docs/docs/api-reference/LayoutGridField.md
```diff
@@ -0,0 +1,267 @@
...
+    "ui:field": "LayoutGridField",
+    "ui:layoutGrid": {
+      "ui:row": { ... }
+    }
+  }
+}
+```
+
+See the [LayoutGridField usage](../usage/layout-grid.md) for more detailed walk-through on how to use the component. This page is a more
+dense version of the `uiSchema` APIs for faster reference.
```
nickgros, line 19: nitpick: I think this is more clear
```suggestion
See the [LayoutGridField usage](../usage/layout-grid.md) for more detailed walk-through on how to use the component. This page is a
condensed version of the `uiSchema` APIs for faster reference.
```

packages/docs/docs/api-reference/LayoutGridField.md
```diff
@@ -0,0 +1,267 @@
...
+
+## ui:row
+
+The outermost level of a `LayoutGridField` is the `ui:row` that defines the nested rows, columns, and/or condition
+elements (i.e. "grid elements") in the grid. This definition is either a simple "grid elements" OR an object with
+native `GridTemplate` implementation specific props and a `children` array of "grid elements". E.g. (as JSON objects):
```
nickgros, line 25:
```suggestion
native `GridTemplate` implementation-specific props and a `children` array of "grid elements". E.g. (as JSON objects):
```

packages/docs/docs/api-reference/LayoutGridField.md
```diff
@@ -0,0 +1,267 @@
...
+One can write this instead:
+
+```json
+{
+  "ui:row": {
+    "children": [
+      {
+        "ui:columns": {
+          "className": "GridColumn col-md-4",
+          "children": ["innerField", "inner.grandChild", { "name": "inner.grandChild2", "convertOther": true }]
```
nickgros, line 194: what is this `convertOther` property? I thought it was some custom/arbitrary property but I don't see it in the previous example?

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+
+Version 6 of RJSF introduced a new way of generating forms using the `UiSchema`, via the Layout Grid.
+Unlike the original method of showing each field in the JSON Schema, in order, as they appear hierarchically within the
+nested object hierarchy, the Layout Grid lets you define where each field goes on a grid in any place/order you choose.
+It even adds the ability to pass additional props to the rendering widget for each element being displayed.
+
+The Layout Grid takes full advantage of the `UiSchema`'s ability to customize the `ui:field` that is used to render a field
+or collection of fields by providing a new out-of-the-box field implementation called `LayoutGridField` that you can
+assign to the whole of the `schema` or just a piece of it. It also comes with its own "syntax" that lets you define
+a `row`, a `col`umn, and a set of `columns`. It also provides `condition`al logic that describe how a subset of the grid
```
nickgros, line 11:
```suggestion
a `row`, a `col`umn, and a set of `columns`. It also provides `condition`al logic that describes how a subset of the grid
```

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+  +------------------------------------------------------------------------------+
+
+  Last Name*
+  +------------------------------------------------------------------------------+
+  |                                                                              |
+  +------------------------------------------------------------------------------+
+```
+
+If you wanted to switch the order of the fields or remove middle name entirely, you would have to change the JSON schema.
+At least you would have to if you didn't have the `LayoutGridField`.
```
nickgros, line 85: This is not true, the uiSchema already has `ui:order` and `"ui:widget": "hidden"`. That said, I like this very simple example, so I don't want you to make it more complicated for the sake of it. Just update the doc so it doesn't suggest that it's impossible without `LayoutGridField`

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+case the `"className": "row"` will be spread onto the underlying theme's grid implementation. You may also pass other
+properties, such as `style` blocks or, in the case of other themes, properties that will affect the underlying grid system
+components.
+
+### ui:col
+
+The `ui:col` "syntax" represents a single column of data in the grid. Like `ui:row` this syntax will spread the
+`"className": "col-xs-12"` and `"className": "col-xs-4"` properties onto the underlying theme's grid implementation, as
+well as any other properties, such as `style` blocks, etc. In its most compact form, the `ui:col` can be written as
+`{ "ui:col": ["person"] }`, in a manner similar `ui:row` where the array is implicitly assumed to be the `children`
```
nickgros, line 325:
```suggestion
`{ "ui:col": ["person"] }`, in a manner similar to `ui:row` where the array is implicitly assumed to be the `children`
```

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+  }
+}
+```
+
+Here is how the `LayoutGridField` "syntax" above is used.
+
+### ui:row, part 2
+
+As the above examples show, the `ui:row` object can be nested inside of the `children` of a `ui:row`. This signifies the
+nesting of one row inside another row. It also show that any additional properties added to the `ui:row` object, in this
```
nickgros, line 315:
```suggestion
nesting of one row inside another row. It also shows that any additional properties added to the `ui:row` object, in this
```

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+case the `"className": "row"` will be spread onto the underlying theme's grid implementation. You may also pass other
```
nickgros, line 316:
```suggestion
case the `"className": "row"`, will be spread onto the underlying theme's grid implementation. You may also pass other
```

packages/docs/docs/api-reference/LayoutGridField.md
heath-freenome: Ah, its a custom prop that we have in our world. I'll switch it to something like `"fullWidth": true`

packages/docs/docs/usage/layout-grid.md
heath-freenome: Oh wow, I forgot about those :) I'll reword

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+  +--------------------------------+  +---------+  +--------------------------------+
+```
+
+### ui:columns
+
+The `ui:columns` "syntax" represents a group of columns in the grid. Like `ui:row` and `ui:col` this syntax will spread
+the `"className": "col-xs-4"` properties onto the underlying theme's grid implementation, as well as any other properties,
+such as `style` blocks, etc. It is short-hand notation for rendering a group of columns all with the same grid system
+properties. As shown in the above examples, the set of three `ui:col` definitions for first, middle and last names are
+the same as the single `ui:columns` definition with all three listed as children.
```
nickgros, line 407: Clarify that the passed props are arbitrary/theme-based.  The example you're referencing is a couple of examples ago.
```suggestion
The `ui:columns` "syntax" represents a group of columns in the grid. Like `ui:row` and `ui:col` this syntax will spread
the properties onto the underlying theme's grid implementation (in an earlier example, `"className": "col-xs-4"`), as well as any other properties,
such as `style` blocks, etc. It is short-hand notation for rendering a group of columns all with the same grid system
properties. As shown in the above examples, the set of three `ui:col` definitions for first, middle and last names are
the same as the single `ui:columns` definition with all three listed as children.
```

packages/docs/docs/api-reference/LayoutGridField.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+> NOTE: Special note about the native `className` prop values. All className values will automatically be looked up in
+> the `formContext.lookupMap` in case they have been defined using a CSS-in-JS approach. In other words, from the
+> example above, if the `Form` was constructed with a `layoutGridLookupMap` set to `{ GridRow: cssInJs.GridRowClass }`
+> then when rendered, the native `GridTemplate` will get the `className` with the value from
+> `cssInJs.GridRowClass`. This automatic lookup will happen for any of the "grid elements" when rendering with
+> `GridTemplate` props. If multiple className values are present, for example:
+> `{ className: 'GridRow GridColumn' }`, the classNames are split apart, looked up individually, and joined
+> together to form one className with the values from `cssInJs.GridRowClass` and `cssInJs.GridColumnClass`.
```
nickgros, line 66: Probably link to that section in the other doc here

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+// Import the UI Schema from the file above
+import uiSchema from './uiSchema.json';
+
+const css = {
+  STYLED_GRID: 'StyledGrid',
+};
+
+// Define the style
+const StyledGrid = styled(Grid)`
+  ${`& .${css.StyledGrid}`} {
```
nickgros, line 779:
```suggestion
  ${`& .${css.STYLED_GRID}`} {
```

packages/docs/docs/usage/layout-grid.md
```diff
@@ -0,0 +1,1142 @@
...
+Since some themes, like `mui`, support CSS-in-JS class names, one cannot hard-code the actual `className` in the
+`uiSchema` easily. That is where the name-mapped information in the `FormContext` comes into play. Let's assume that
+you are using a theme like `mui` and you need to style a `Grid` using CSS-in-JS you could do something like the following:
+
+```json uiSchema.json
+{
+  "ui:field": "LayoutGridField",
+  "ui:layoutGrid": {
+    "ui:row": {
+      "className": "GridStyle anotherClass",
```
nickgros, line 755: As a convention, I wonder if we should suggest giving the 'placeholder' class name to be replaced a prefix or something like that, so it's clear that it's not a real classname

nickgros (approved):
Plenty of suggestions, but this looks great.

packages/docs/docs/usage/layout-grid.md
heath-freenome: In the examples? So you mean `GridStyle-REPLACE_ME`?

PR #4583, opened 2025-04-30
Chore: Update 3rd Party libraries to the latest

heath-freenome:

Updated as many of the 3rd Party libraries as possible
- Updated the root `package.json` to consolidate the dev dependencies from all of the sub-packages like:
  - Typescript, rollup, jest, react, types and babel
  - Also bumped all of the libraries that were out-of-date
- Updated the individual `package.json` files to update libraries and to remove duplicated dev dependencies that are in the root `package.json`
- Updated types in `@rjsf/utils` and `@rjsf/core` due to typescript and lodash type updates
- Updated `DemoFrame` in `playground` to work with the latest version of `react-frame-component`
- Updated all the snapshots for themes that needed them due to library updates
- Updated `docs` libraries, including fixing up .md files that caused problems with the latest version of Docusaurus
- Made `@rjsf/utils` `isObject()` a type guard

### Checklist


packages/antd/package.json
```diff
@@ -76,38 +76,17 @@
...
-    "@ant-design/icons": "^5.6.1",
-    "@babel/cli": "^7.23.9",
-    "@babel/core": "^7.23.9",
-    "@babel/eslint-parser": "^7.23.10",
-    "@babel/plugin-proposal-class-properties": "^7.18.6",
-    "@babel/plugin-proposal-optional-chaining": "^7.21.0",
-    "@babel/preset-env": "^7.23.9",
-    "@babel/preset-react": "^7.23.3",
-    "@babel/preset-typescript": "^7.23.3",
+    "@ant-design/icons": "6.0.0",
```
nickgros, line 79: Is the intent behind `6.0.0` instead of `^6.0.0` to stabilize things like snapshots and typechecks so they only change when we explicitly upgrade?

packages/playground/src/components/DemoFrame.tsx
```diff
@@ -74,7 +64,7 @@ export default function DemoFrame(props: DemoFrameProps) {
     );
     setContainer(instanceRef.current.contentDocument.body);
     setWindow(() => instanceRef.current.contentWindow);
-  }, []);
+  }, [instanceRef]);
```
nickgros, line 67: I don't think this will do anything because the `ref` will never change

packages/playground/src/components/DemoFrame.tsx
```diff
@@ -53,16 +53,6 @@ export default function DemoFrame(props: DemoFrameProps) {
...
 
-  const handleRef = useCallback(
-    (ref: any) => {
-      instanceRef.current = {
-        contentDocument: ref ? ref.node.contentDocument : null,
-        contentWindow: ref ? ref.node.contentWindow : null,
-      };
-    },
-    [instanceRef],
-  );
```
nickgros, line 64: not needed anymore?

packages/utils/src/types.ts
```diff
@@ -1013,10 +1017,6 @@ export type UiSchema<
     'ui:globalOptions'?: GlobalUISchemaOptions;
     /** Allows the form to generate a unique prefix for the `Form`'s root prefix  */
     'ui:rootFieldId'?: string;
-    /** Allows RJSF to override the default field implementation by specifying either the name of a field that is used
-     * to look up an implementation from the `fields` list or an actual one-off `Field` component implementation itself
-     */
-    'ui:field'?: Field<T, S, F> | string;
```
nickgros, line 1019: Got it, `MakeUIType` will handle this with `field` defined in `UIOptionsBaseType`

packages/antd/package.json
```diff
@@ -76,38 +76,17 @@
     "rc-picker": "2.7.6"
   },
   "devDependencies": {
-    "@ant-design/icons": "^5.6.1",
-    "@babel/cli": "^7.23.9",
```
nickgros, line 80: fyi, if we switch to pnpm then we may still need to explicitly list the dependencies we are using in each package

nickgros (approved):
had some questions, not blocking approval

packages/antd/package.json
heath-freenome: Ah, no I just used an alias that does things exactly since that is the way we do it at work. I can add back the `^`

packages/playground/src/components/DemoFrame.tsx
heath-freenome: No, it the `instanceRef` is now directly set and does the same thing

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -440,12 +440,12 @@ export function getObjectDefaults<T = any, S extends StrictRJSFSchema = RJSFSche
...
     const objectDefaults = Object.keys(retrievedSchema.properties || {}).reduce(
       (acc: GenericObjectType, key: string) => {
-        const propertySchema = get(retrievedSchema, [PROPERTIES_KEY, key]);
+        const propertySchema: S = get(retrievedSchema, [PROPERTIES_KEY, key], {}) as S;
         // Check if the parent schema has a const property defined AND we are supporting const as defaults, then we
         // should always return the computedDefault since it's coming from the const.
         const hasParentConst = isObject(parentConst) && (parentConst as JSONSchema7Object)[key] !== undefined;
         const hasConst =
-          ((isObject(propertySchema) && CONST_KEY in propertySchema) || hasParentConst) &&
+          ((typeof propertySchema === 'object' && CONST_KEY in propertySchema) || hasParentConst) &&
```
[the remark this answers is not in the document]
heath-freenome: type checking. Our `isObject()` function did not work as a type guard

packages/playground/src/components/DemoFrame.tsx
heath-freenome: technically the `instanceRef.current` can be changed during the mount/unmount... it is technically correct to have it there

nickgros: But even when `current` changes, `instanceRef` will be the same object

packages/utils/src/schema/getDefaultFormState.ts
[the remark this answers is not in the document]
nickgros: Why not update `isObject` to be a typeguard? Think it would be as simple as

```
export default function isObject(thing: any): thing is object {
```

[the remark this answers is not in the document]
heath-freenome: That works!

packages/playground/src/components/DemoFrame.tsx
heath-freenome: True... And having it as a dependency doesn't hurt and is technically correct

Issue #4389, opened 2024-11-20
update chakra to v3


julianhille:
### Prerequisites

- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)

### What theme are you using?

chakra-ui

### Is your feature request related to a problem? Please describe.

Chakra ui has released its version 3

Here is the migration guide:
https://www.chakra-ui.com/docs/get-started/migration

First test shows that a lot of errors are thrown when combining chakra v3 and latest rjsf

### Describe the solution you'd like

Be compatible to chakra v3

heath-freenome:
@julianhille Is this something you would be willing to do for us... Perhaps in our v6 beta (that way we have support for it in the new release)

julianhille:
I could give it a try.

Syynth:
@MilanKrupa I have started on a version of this here:
https://github.com/Syynth/react-jsonschema-form/tree/main

there is definitely more work to do, some tests are crashing for reasons I don't understand, and i'm getting a strange error in the playground on some of the examples (`A validator is required for Form functionality to work`), but if anyone wants to take it further or assist me, i'd love to collaborate/etc.

MilanKrupa:
@Syynth So I guess I've unnecessarily started to migrate it here: [https://github.com/MilanKrupa/react-jsonschema-form/tree/chakra-ui-v2-to-v3-migration](https://github.com/MilanKrupa/react-jsonschema-form/tree/chakra-ui-v2-to-v3-migration). I wonder which one of us made more progress so that we can continue to work on this in a single repo. Based on your comment I assume you were able to get the playground up and running? I also did, and I was able to fix majority of the type errors. In my case some of the components are not yet working properly in the playground though. There are some styling adjustments to be made too (because of the CSS reset i guess).

About the validator - I wouldn't worry about it too much for now. When you get this error, just select the AJV8 and refresh the form using the button in error message. It seems like this is an issue with the playground itself. Same thing happens here: [https://rjsf-team.github.io/react-jsonschema-form/](https://rjsf-team.github.io/react-jsonschema-form/).

I propably won't be able to do too much this week, so let me know if you'd like to continue. If you don't, then I'll try to get it done next week

heath-freenome:
[...]

PR #4620, opened 2025-05-13
restore getChakra calls

antpaw:

I've added the uiSchema props via `getChakra` call back in

https://github.com/rjsf-team/react-jsonschema-form/issues/4389#issuecomment-2873342138

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Fixed form data propagation with `patternProperties` [#4617](https://github.com/rjsf-team/react-jsonschema-form/pull/4617)
 
+## @rjsf/chakra-ui
+
+- Added `getChakra` to package exports
+- Restored the `ui:options` customization
+
 # 6.0.0-beta.7
 
 ## @rjsf/core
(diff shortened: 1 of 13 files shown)


antpaw:
I can not test it, there is an issue in the playground for all the providers 
image

thesnups (commented):
Thanks! I don't have full context but want to check if the chakra props should be spread last so that they can override the existing props? For example, I might want to change the margin-bottom on BaseInputTemplate

antpaw:
i've changed it

heath-freenome:
@antpaw does it make sense to also add it to the `AltDateWidget` on the `Box` with the flex? Also, you have some Typescript issues

thesnups:
Can we export the `getChakra` util function for package consumers so that they can use it in their custom templates?

heath-freenome:
> Can we export the `getChakra` util function for package consumers so that they can use it in their custom templates?

That is a fair request. @antpaw can you export it along with the `ChakraUiSchema` type that is already being exported?

antpaw:
@heath-freenome please review, i had to cast its as `any` as its not possible to use the CheckboxesWidget wrapped in `<Field>`

antpaw:
@heath-freenome done

heath-freenome (approved):

PR #4490, opened 2025-02-07
Fix 3759 - uneditable & permanent defaults with additional properties

Piotr-Debicki:
### Reasons for making this change

Fixes [#3759](https://github.com/rjsf-team/react-jsonschema-form/issues/3759)

Additional properties with defaults set currently creates a permanent key-value pair that is essentially uneditable.
Trying to edit the key creates an entirely new key value entry on blur while still keeping the default key-value pair. The default key-value pair is also not deletable. Solution here generates defaults for additional properties once on form initialization, and then skips regeneration on any form update

This solution works for root & nested additionalProperties (see tests)
Allows for 'reset' form to regenerate defaults
Also fixes the issue I mentioned in the [Closed PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4440) with custom formData not being applied when generating additionalProperties defaults

I wanted to avoid prop drilling, but there wasn't an obvious way to get root form state into getDefaultFormState

### Checklist

- [x] **I'm updating documentation**
  - [x] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.
  - [x] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [x] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #4625, opened 2025-05-15
feat: allow override form values with defaults

Lonli-Lokli:

Add the ability to override existing form values (especially arrays) with defaults

### Checklist


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+## @rjsf/util
```
heath-freenome, line 19: This belongs in the `beta-8` section below

packages/docs/docs/api-reference/form-props.md
```diff
@@ -269,6 +269,9 @@ NOTE: If there is a default for a field and the `formData` is unspecified, the d
 | ------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
 | `useFormDataIfPresent`          | Legacy behavior - Do not merge defaults if there is a value for a field in `formData` even if that value is explicitly set to `undefined` |
 | `useDefaultIfFormDataUndefined` | If the value of a field within the `formData` is `undefined`, then use the default value instead                                          |
+| `useDefault`                    | Always use the default value instead of form data                                                                                         |
```
heath-freenome, line 272: Consider `useDefaultAlways` to be super clear

packages/playground/src/components/Header.tsx
```diff
@@ -201,6 +201,11 @@ const liveSettingsSelectSchema: RJSFSchema = {
               title: 'Use default for undefined field value',
               enum: ['useDefaultIfFormDataUndefined'],
             },
+            {
+              type: 'string',
+              title: 'Always use default for field value',
+              enum: ['useDefault'],
```
heath-freenome, line 207: `useDefaultAlways`?

packages/utils/src/mergeDefaultsWithFormData.ts
```diff
@@ -20,7 +20,7 @@ import isNil from 'lodash/isNil';
  * @param [formData] - The form data into which the defaults will be merged
  * @param [mergeExtraArrayDefaults=false] - If true, any additional default array entries are appended onto the formData
  * @param [defaultSupercedesUndefined=false] - If true, an explicit undefined value will be overwritten by the default value
- * @param [overrideFormDataWithDefaults=false] - If true, the default value will overwrite the form data value. If the value
+ * @param [overrideFormDataWithDefaultsStrategy='noop'] - If not 'noop', the default value will overwrite the form data value. Values can be either replaced or merged, if the value
```
heath-freenome, line 23: This seems like a line that is longer than 120 characters, consider reflowing this with the lines below

packages/utils/src/mergeDefaultsWithFormData.ts
```diff
@@ -48,33 +48,43 @@ export default function mergeDefaultsWithFormData<T = any>(
...
+          overrideFormDataWithDefaultsStrategy,
         );
       }
       return value;
     });
 
     // Merge any extra defaults when mergeExtraArrayDefaults is true
-    // Or when overrideFormDataWithDefaults is true and the default array is shorter than the formData array
-    if ((mergeExtraArrayDefaults || overrideFormDataWithDefaults) && mapped.length < overrideOppositeArray.length) {
+    // Or when overrideFormDataWithDefaults is not noop and the default array is shorter than the formData array
```
heath-freenome, line 58: There is a difference between not `noop` as you've mentioned here and the `merge` value below. Consider updating comment if the code it correct or the code if the comment is correct

packages/utils/src/mergeDefaultsWithFormData.ts
```diff
@@ -31,14 +31,14 @@ export default function mergeDefaultsWithFormData<T = any>(
   formData?: T,
   mergeExtraArrayDefaults = false,
   defaultSupercedesUndefined = false,
-  overrideFormDataWithDefaults = false,
+  overrideFormDataWithDefaultsStrategy: 'noop' | 'replace' | 'merge' = 'noop',
```
heath-freenome, line 34: Consider making an enum for this, or at least a type.

packages/utils/src/mergeDefaultsWithFormData.ts
```diff
@@ -48,33 +48,43 @@ export default function mergeDefaultsWithFormData<T = any>(
...
       const keyExistsInDefaults = isObject(defaults) && key in (defaults as GenericObjectType);
       const keyExistsInFormData = key in (formData as GenericObjectType);
+      // overrideFormDataWithDefaultsStrategy can be 'merge' only when the key value exists in defaults
+      // Or if the key value doesn't exist in formData
+      const keyOverrideDefaultStrategy =
+        overrideFormDataWithDefaultsStrategy === 'replace'
+          ? 'replace'
+          : keyExistsInDefaults || !keyExistsInFormData
+            ? overrideFormDataWithDefaultsStrategy
+            : 'noop';
```
heath-freenome, line 81: Nested ternary expressions can be a bit hard to parse sometimes. Consider a `let` with chained if statements?

packages/utils/src/types.ts
```diff
@@ -98,8 +98,9 @@ export type Experimental_DefaultFormStateBehavior = {
    *        even if that value is explicitly set to `undefined`
    * - `useDefaultIfFormDataUndefined`: - If the value of a field within the `formData` is `undefined`, then use the
    *        default value instead
+   * - `useDefault`: - Always use the default value
```
heath-freenome, line 101: `useDefaultAlways`

packages/utils/test/schema/getDefaultFormStateTest.ts
```diff
@@ -1340,6 +1340,65 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
             });
           });
         });
+
+        describe('mergeDefaultsIntoFormData set to "useDefault"', () => {
```
heath-freenome, line 1344: `useDefaultAlways`

heath-freenome (changes requested):

packages/utils/test/mergeDefaultsWithFormData.test.ts
```diff
@@ -307,4 +307,94 @@ describe('mergeDefaultsWithFormData()', () => {
       expect(mergeDefaultsWithFormData(obj1, obj2)?.a).toBeInstanceOf(File);
     });
   });
+
+  describe('test with overrideFormDataWithDefaults set to `replace`', () => {
+    it('should return empty array even when no defaults', () => {
+      expect(mergeDefaultsWithFormData(undefined, [2], undefined, undefined, 'replace')).toEqual([]);
+    });
```
nickgros, line 314: Maybe I'm misunderstanding the feature, why does this return an empty array?

packages/utils/test/mergeDefaultsWithFormData.test.ts
```diff
@@ -307,4 +307,94 @@ describe('mergeDefaultsWithFormData()', () => {
...
+      expect(mergeDefaultsWithFormData({ a: 1 }, { a: 2 }, undefined, undefined, 'replace')).toEqual({ a: 1 });
+    });
+
+    it('should not return undefined from defaults', () => {
+      expect(
+        mergeDefaultsWithFormData({ a: { b: undefined } }, { a: { b: { c: 1 } } }, undefined, undefined, 'replace'),
+      ).toEqual({
+        a: { b: {} },
+      });
+    });
```
nickgros, line 344: `undefined` is not a valid JSON value,. so what JSON Schema results in a `default` value that looks like `{ a: { b: undefined } }`? Are we doing something special in RJSF that I am ignorant of?

packages/utils/test/mergeDefaultsWithFormData.test.ts
Lonli-Lokli: That's because undefined will be treated as missing value for this, so replaced with proper default for arrays (empty). Do you think it's incorrect?

packages/utils/test/mergeDefaultsWithFormData.test.ts
Lonli-Lokli: I do not know actually how it might happen in JSON schema, I just tried to repeat existing tests. Here we wiill replace missing value (undefined from default arg) with proper default for object (empty one). Do you think it's incorect?

nickgros: It looks like our `computeDefaults` utility may return `undefined`, so I guess we do need to handle it.

packages/utils/test/mergeDefaultsWithFormData.test.ts
nickgros: My assumption was that `mergeDefaultsWithFormData` treats `undefined` computed defaults as there being no defaults, so I would think in this scenario we keep user data. But if it's the case that it's only called when there is some default value (even if it's computed to be `undefined`), then it makes sense to replace user data with some value like `[]`. So I think what you have is fine. We can refine the implementation over time too if necessary.

nickgros (approved):

Lonli-Lokli:
@heath-freenome I can wait until https://github.com/rjsf-team/react-jsonschema-form/pull/4637 be merged, rebase and recheck.

I do not think this PR will solve my issue, actually.

Lonli-Lokli:
@heath-freenome I've merged latest changes, but maybe need your help with additional test.

I want to test several scenarios (some test data already in the code)
1) User on empty form selects country UK -> we have ['London'] as values in the form ✔️ **done**
2) User on form where UK is selected as country and some city array is populated eg ['London']  selects 'France' as country - country changing to France as expected as list of cities changed to defaults ['Paris']
3) User on form where UK is selected as country and some city array is populated eg ['London']  adding new city eg Manchester - country stays as UK and city list becomes ['London', 'Manchester']

Can you please either add tests yourself or five some example?

Lonli-Lokli:
Basically I want to have different behaviour for two cases - when main field changing I want to clear dependant and apply defaults instead, if dependant is changing I want to keep (aka add/remove) old value. I cannot find a way as of now to do it

Lonli-Lokli:
I spent some time today and think my current code will not work in a way I want. Instead I am thinking of moving useAlwaysDefault logic to Form onChange handler 
1) determine changing fields
2) if field has dependencies and formBehavior is `useDefaultAlways` - remove/nullish all fields before calling schemaUtils.getDefaultFormState
3) revert changes in getDefaultFormState

@heath-freenome what doo you think?

heath-freenome:
@Lonli-Lokli 

> I spent some time today and think my current code will not work in a way I want. Instead I am thinking of moving useAlwaysDefault logic to Form onChange handler
> 
> 1. determine changing fields
> 2. if field has dependencies and formBehavior is `useDefaultAlways` - remove/nullish all fields before calling schemaUtils.getDefaultFormState
> 3. revert changes in getDefaultFormState
> 
> @heath-freenome what doo you think?

Can you explain what you are considering? Also, there is another [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4490) that maybe helpful for you?

Lonli-Lokli:
So the problem which my current PR is not solving - I want kind of cascading lookups, ie secondary must be changed when primary changing.

Problem I see with current PR is that it will always restore defaults from primary , as getDefaultFormState do not know which field has been changed - primary (so we should replace secondary with defaults) or secondary (and we should keep this new multiselect value, ignoring defaults from primary).

So I do not see how another PR will solve it.

heath-freenome:
@Lonli-Lokli Do you feel like you still need to solve the issue you were attempting to fix with this PR?

Lonli-Lokli:
I still think this is an important case which is useful to solve - dependant dropdowns with different set of possible values - but I myself work rounded it with non-library code.
Unfortunately I could not make PR work without breaking other parts of tests/functionality/introducing breaking things

PR #4683, opened 2025-07-02
Always make all references absolute in nested bundled schemas

GlassOfWhiskey:
This commit adjusts all internal references of a bundled schema to be absolute, allowing subschemas to be correctly resolved.

Field renderers often resolve references on subsets of the original subschema, with are always validated against the whole root schema. However, bundled schemas may contain relative references, which should be resolved against an internal `$id` field. To resolve this problem, all references in nested bundled schema are now transformed into absolute references, so that they are always resolvable against the root schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 6.0.0-beta.11
 
+## rjsf/utils
+
+- Always make all references absolute in nested bundled schemas
+
+
+# 6.0.0-beta.11
+
 ## @rjsf/antd
 
 - Set peerDependency for `@ant-design/icons` to `^6.0.0`, fixing [#4643](https://github.com/rjsf-team/react-jsonschema-form/issues/4643)
--- a/packages/utils/src/findSchemaDefinition.ts
+++ b/packages/utils/src/findSchemaDefinition.ts
@@ -1,7 +1,14 @@
 import jsonpointer from 'jsonpointer';
 import omit from 'lodash/omit';
 
-import { ID_KEY, JSON_SCHEMA_DRAFT_2020_12, REF_KEY, SCHEMA_KEY } from './constants';
+import {
+  ALL_OF_KEY,
+  ID_KEY,
+  JSON_SCHEMA_DRAFT_2019_09,
+  JSON_SCHEMA_DRAFT_2020_12,
+  REF_KEY,
+  SCHEMA_KEY,
+} from './constants';
 import { GenericObjectType, RJSFSchema, StrictRJSFSchema } from './types';
 import isObject from 'lodash/isObject';
 import isEmpty from 'lodash/isEmpty';
@@ -20,7 +27,16 @@ function findEmbeddedSchemaRecursive<S extends StrictRJSFSchema = RJSFSchema>(sc
     return schema;
   }
   for (const subSchema of Object.values(schema)) {
-    if (isObject(subSchema)) {
+    if (Array.isArray(subSchema)) {
+      for (const item of subSchema) {
+        if (isObject(item)) {
+          const result = findEmbeddedSchemaRecursive<S>(item as S, ref);
+          if (result !== undefined) {
+            return result as S;
+          }
+        }
+      }
+    } else if (isObject(subSchema)) {
       const result = findEmbeddedSchemaRecursive<S>(subSchema as S, ref);
       if (result !== undefined) {
         return result as S;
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.0.0-beta.11
```
```suggestion
# 6.0.0-beta.13
```

packages/utils/src/findSchemaDefinition.ts
```diff
@@ -30,6 +46,31 @@ function findEmbeddedSchemaRecursive<S extends StrictRJSFSchema = RJSFSchema>(sc
...
+/** Parses a JSONSchema and makes all references absolute with respect to
+ * the `baseURI` argument
+ * @param schema - The schema to be processed
+ * @param baseURI - The base URI to be used for resolving relative references
+ */
+function makeAllReferencesAbsolute<S extends StrictRJSFSchema = RJSFSchema>(schema: S, baseURI: string): S {
+  const currentURI = get(schema, ID_KEY, baseURI);
+  // Make all other references absolute
+  if (REF_KEY in schema) {
+    schema = { ...schema, [REF_KEY]: UriResolver.resolve(currentURI, schema[REF_KEY]!) };
```
heath-freenome, line 58: So we are manipulating the root schema here or the lower-level schema as we are calling `retrieveSchema()`? Honestly, with a few changes to the `createSchemaUtils()` and `Form` in core, we could probably run this code over the `rootSchema` once and cache it in the `schemaUtils` object. And then use that schema within the `Form` by updating the `getRegistry()` function and update the `schema` in the state within `Form` to pick up the cached, updated `rootSchema` from the `schemaUtils`. Does that make sense?

packages/utils/src/findSchemaDefinition.ts
```diff
@@ -103,7 +150,14 @@ export function findSchemaDefinitionRecursive<S extends StrictRJSFSchema = RJSFS
...
     if (Object.keys(remaining).length > 0) {
-      return { ...remaining, ...subSchema };
+      if (
+        rootSchema[SCHEMA_KEY] === JSON_SCHEMA_DRAFT_2019_09 ||
+        rootSchema[SCHEMA_KEY] === JSON_SCHEMA_DRAFT_2020_12
+      ) {
+        return { [ALL_OF_KEY]: [remaining, subSchema] } as S;
+      } else {
+        return { ...remaining, ...subSchema };
+      }
```
nickgros, line 154: Why do we have different logic for different standard versions? Is the `allOf` option more correct (and could we use it for draft-07)?

packages/utils/src/findSchemaDefinition.ts
heath-freenome: Actually, I just updated the `Form` in `core` to always use the `rootSchema` in `schemaUtils` in this [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4687) so all you have to do is modify the `schemaUtils` to cache the modified `rootSchema` having made references absolute.

heath-freenome:
@GlassOfWhiskey We are waiting for a rebase as well as responses to our queries

packages/utils/src/findSchemaDefinition.ts
GlassOfWhiskey: Before Draft 2019-09, `$ref` could not have other keywords alongside it.
This behavior changed in 2019-09 and later
https://json-schema.org/draft/2019-09/release-notes

packages/utils/src/findSchemaDefinition.ts
GlassOfWhiskey: I moved the logic as requested

packages/utils/src/createSchemaUtils.ts
```diff
@@ -57,7 +60,11 @@ class SchemaUtils<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
     experimental_defaultFormStateBehavior: Experimental_DefaultFormStateBehavior,
     experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>,
   ) {
-    this.rootSchema = rootSchema;
+    if (rootSchema[SCHEMA_KEY] === JSON_SCHEMA_DRAFT_2020_12) {
```
heath-freenome, line 63: So we only want to do this for 2020? What about 2019?

GlassOfWhiskey: Support for embedded schemas has been added to JSONSchema only from draft-2020-12 (see [here](https://json-schema.org/draft/2020-12/release-notes#embedded-schemas-and-bundling)). Therefore, I did not implement support for embedded schemas in prior versions of the standard.
If you want to support this too, it would be a non-standard backport of the feature to previous versions.
As you prefer.

heath-freenome: Got it... I thought that, since you added the 2019 schema checking, that it also did that. Thanks for clarifying.

heath-freenome:
Can you rebase and make sure things are all set. I also want to wait for @nickgros to come back from vacation to look at this again as well.

GlassOfWhiskey:
Just did it.
No problem in waiting a bit more to merge the PR. Take your time.

packages/utils/src/findSchemaDefinition.ts
nickgros: Ok, we may want to build a util that we can use to see if certain functionality is in a given schema version--I can imagine that RJSF will have more control flow patterns like this, and if new versions of JSON Schema are released, we would want to only update code in a single place--but I won't burden you with that request, we can build it later.

nickgros (approved):

PR #4698, opened 2025-07-23
fix(utils): Remove fixed package-type

harish-prakash:
Removing the `type` allows consumer to choose between cjs/esm

Using `type:"module"` forces the consumer to ONLY use esm. Removing the type will allow the consumer to decide between cjs/esm based on the `exports` configuration below.


heath-freenome:
@harish-prakash If you need the CJS version of the library, you can use `require` instead of `import`, due to how we've provided the `exports` directive in the `package.json`.

PR #4700, opened 2025-07-24
Expose title property for validated fields

### LINKED ISSUE #4504: Expose `title`/`ui:title` in `RJSFValidationError`
### Prerequisites

### What theme are you using?

mui

I have a field named `confirmPassword`, which has its `ui:title` set to `"Confirm Password"`.

I want to have a custom `errorTransformer` which does a bit of dynamic composition.
It should change the error message and append the title at the end, nothing special.

The problem is that I cannot easily retrieve the `ui:title` from the `RJSFValidationError` object:

Image

Notice how the `ui:title` is only present in the `message` and `stack` properties, without an easy way to dynamically extract it for composing other error messages.

The most straightforward solution would be to add a new field to `RJSFValidationError` named `title`, which takes the property's `title` value from either the JSON Schema `title` or the UI Schema `ui:title` (with `ui:title` taking precedence over `title` maybe?).

Alternatively, adding an entire `uiOptions` field containing the corresponding `ui:options` (like returned by [`getUiOptions`](https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/utility-functions/#getuioptionst--any-s-extends-strictrjsfschema--rjsfschema-f-extends-formcontexttype--any)) for the property would cover even more use cases than exposing just the `title`.

I have tinkered around with [`getUiOptions`](https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/utility-functions/#getuioptionst--any-s-extends-strictrjsfschema--rjsfschema-f-extends-formcontexttype--any), but dynamically accessing the property given how they are not always standardized in the `property` field (sometimes the property name is preceded by a dot, sometimes not) proved a great hassle.

lazaromenezes:

This exposes the `title` property on `RJSFValidationError` to allow further custom message formating.

Fixes #4504

In #4504 discussion is mentioned a `rjsf-v6` branch that I couldn't find so I'm targeting main instead. If there's another branch I could retarget this PR please let me know.

### Checklist

- [X] **I'm updating documentation**
  - [X] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
  - [X] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.
  - [X] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,16 @@ it according to semantic versioning. For example, if your PR adds a breaking cha
 should change the heading of the (upcoming) version to include a major version bump.
 
 -->
+# 6.0.0-beta.14
+
+## @rjsf/utils
+
+- Added `title` property to `RJSFValidationError` [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
+
+## @rjsf/validator-ajv8
+
+- Updated `transformRJSFValidationErrors()` to include the `title` property of a field with error fixing #4504 with [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
+
 # 6.0.0-beta.13
 
 ## @rjsf/shadcn
@@ -52,11 +62,6 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Updated `SchemaUtils` and `createSchemaUtils()` to add a new `getRootSchema()` function
-- Added `title` property to `RJSFValidationError` [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
-
-## @rjsf/validator-ajv8
-
-- Updated `transformRJSFValidationErrors()` to include the `title` property of a field with error fixing #4504 with [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
 
 # 6.0.0-beta.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,11 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Updated `SchemaUtils` and `createSchemaUtils()` to add a new `getRootSchema()` function
+- Added `title` property to `RJSFValidationError` [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
+
+## @rjsf/validator-ajv8
+
+- Updated `transformRJSFValidationErrors()` to include the `title` property of a field with error fixing #4504 with [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
 
 # 6.0.0-beta.11
 
--- a/packages/validator-ajv8/src/processRawValidationErrors.ts
+++ b/packages/validator-ajv8/src/processRawValidationErrors.ts
@@ -39,6 +39,7 @@ export function transformRJSFValidationErrors<
     let { message = '' } = rest;
     let property = instancePath.replace(/\//g, '.');
     let stack = `${property} ${message}`.trim();
+    let uiTitle = '';
     const rawPropertyNames: string[] = [
       ...(params.deps?.split(', ') || []),
       params.missingProperty,
@@ -61,10 +62,12 @@ export function transformRJSFValidationErrors<
         }
         if (uiSchemaTitle) {
           message = message.replace(`'${currentProperty}'`, `'${uiSchemaTitle}'`);
+          uiTitle = uiSchemaTitle;
         } else {
           const parentSchemaTitle = get(parentSchema, [PROPERTIES_KEY, currentProperty, 'title']);
           if (parentSchemaTitle) {
             message = message.replace(`'${currentProperty}'`, `'${parentSchemaTitle}'`);
+            uiTitle = parentSchemaTitle;
           }
         }
       });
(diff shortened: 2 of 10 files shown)


packages/validator-ajv8/src/processRawValidationErrors.ts
```diff
@@ -97,6 +102,7 @@ export function transformRJSFValidationErrors<
       params, // specific to ajv
       stack,
       schemaPath,
+      title: uiTitle,
```
heath-freenome, line 105: Thanks for the fix... can you add a unit test in `utils` to verify this fix? And update the `CHANGELOG.md` as well.

lazaromenezes: Sure! I'll try getting something done in the next few days.

lazaromenezes:
Fixed some failing tests after my change and updated the changelog. Still thinking in the best way to make a specific test for the change.

TheOneTheOnlyJJ:
As the author of #4504, I'm very thankful to see this being taken care of.

I have a single proposal/update here, which is also mentioned in #4504.

Would it be feasible to expose the entire `uiOptions` object of the incorrect field in its `RJSFValidationError`?
This way, custom error messages could also be based on information from other `uiOptions`, such as the `description`, `help`, `label`, `emptyvalue`, `placeholder`, etc. This would enable more complex use cases by allowing the creation of custom error messages based on any and every UI option the field has, not strictly limited to the title.
This is especially important when working with runtime dynamic JSON schemas (like my project does).
The title, if present, would be included in the `uiOptions` object, fixing #4504 as well, but in an even more elegant way, opening up even more possibilities.

With the `ui:title` being retrieved from `uiOptions` in this PR, could the parameters and typings of `RJSFValidationError` be updated to pass the entire `uiOptions` object instead of just the title?

nickgros:
@TheOneTheOnlyJJ would it make sense to merge this, and later you (or @lazaromenezes, or another contributor) could enhance the object with the full `uiOptions`? Or should we wait for @lazaromenezes to weigh in on your request?

lazaromenezes:
@nickgros if maintainers are okay with merging as it is I'm also okay and address uiOptions in a later pull request and put more work on the tests when exposing the full object.

heath-freenome:
> @nickgros if maintainers are okay with merging as it is I'm also okay and address uiOptions in a later pull request and put more work on the tests when exposing the full object.

The one hesitation I have is that, in adding support for the `uiOptions`, you will cause a breaking change if you transition from `title` to `uiOptions`.

lazaromenezes:
> > @nickgros if maintainers are okay with merging as it is I'm also okay and address uiOptions in a later pull request and put more work on the tests when exposing the full object.
> 
> The one hesitation I have is that, in adding support for the `uiOptions`, you will cause a breaking change if you transition from `title` to `uiOptions`.

My original thought was adding `uiOptions` as an additional property instead of replacing `title`. If you think it makes more sense having a single new property, I can change it to be `uiOptions` this week.

heath-freenome:
> > > @nickgros if maintainers are okay with merging as it is I'm also okay and address uiOptions in a later pull request and put more work on the tests when exposing the full object.
> > 
> > 
> > The one hesitation I have is that, in adding support for the `uiOptions`, you will cause a breaking change if you transition from `title` to `uiOptions`.
> 
> My original thought was adding `uiOptions` as an additional property instead of replacing `title`. If you think it makes more sense having a single new property, I can change it to be `uiOptions` this week.

If that is the case, then I'm fine with this merging... Please rebase

TheOneTheOnlyJJ:
But would it really make sense to expose both `title` and `uiOptions` separately, given that `title` is included in `uiOptions`?

It only makes sense when keeping into account that the `title` may come from the JSON schema and not exclusively from the UI Schema, but in this case, should't all the other options where this can be the case be treated the same way?

I hope I'm not unnecessarily delaying the work on this issue, but I'm just trying to keep into account all the details that may impact the usefulness of this feature, based on my experience with having no access to the `title` property.

CHANGELOG.md
```diff
@@ -51,6 +51,11 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Updated `SchemaUtils` and `createSchemaUtils()` to add a new `getRootSchema()` function
+- Added `title` property to `RJSFValidationError` [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
+
+## @rjsf/validator-ajv8
+
+- Updated `transformRJSFValidationErrors()` to include the `title` property of a field with error fixing #4504 with [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4700)
```
heath-freenome: Due to the release of beta.13, can you add a new `6.0.0-beta.14` section and put these changes into it?

lazaromenezes: Done and rebased

nickgros:
> It only makes sense when keeping into account that the title may come from the JSON schema and not exclusively from the UI Schema

For this reason, I think it's a good idea to have a top-level `title` property.

>[...] in this case, should't all the other options where this can be the case be treated the same way?

Those other options can be incrementally added to the top level as needed. I don't think it's harmful to also have the `uiOptions.title` available as well. I think the proposed API (hopefully) makes it obvious that that the `uiOptions` data exclusively comes from the uiSchema.

Let's merge this. @lazaromenezes thanks for the contribution and @TheOneTheOnlyJJ thanks for your input!

PR #4724, opened 2025-08-16
Fix 4676 - Add support for arrays in the LayoutGridField

### LINKED ISSUE #4676: Using `LayoutGridField` for an array instead of an object
### Prerequisites

### What theme are you using?

mui

### What is your question?

Hello, I am wondering how one would set up their `uiSchema` to get it to render an array of fields.

I found[ the docs](https://rjsf-team.github.io/react-jsonschema-form/docs/usage/layout-grid) and [this sandbox](ttps://rjsf-team.github.io/react-jsonschema-form/) very helpful but it only really covers objects.

My specific scenario is I have a field in my schema which is an array of 3 element tuples (arrays with max and min items set). I would like each row to be layed out such that each element in the inner list is next to one another in that row.

Such a property may look similar to:

```json
{
    "example": {
        "type": "array",
        "prefixItems": [
        {
            "maxItems": 3,
            "minItems": 3,
            "prefixItems": [
                {
                    "type": "integer"
                },
                {
                    "type": "integer"
                },
                {
                    "type": "integer"
                }
            ],
            "type": "array"
        },
        // more items
    ],
}
```

With the actual data coming in as
```json
   example: [
      [6, 16, 50],
      [7, 14, 31],
      [7, 14, 31],
      [7, 14, 31],
      [7, 21, 38],
      [7, 21, 38],
      [7, 26, 50],
    ],
```

heath-freenome:

Fixes #4676 by adding support for indexes into the `LayoutGridField`
- Updated `LayoutGridField` to add support for arrays
- Updated the `LayoutGridField` to add 100% validation for the new arrays code

Using the schema in the `LayoutGridField` tests and this `uiSchema`...

```
{
  "ui:field": "LayoutGridField",
  "ui:layoutGrid": {
    "ui:row": {
      "spacing": 2,
      "children": [
        {
          "ui:col": {
            "size": 12,
            "children": [
              "example"
            ]
          }
        },
        {
          "ui:columns": {
            "size": 4,
            "children": [
              {
                "name": "example.0.0",
                "fullWidth": true
              },
              {
                "name": "example.0.1",
                "fullWidth": true
              },
              {
                "name": "example.0.2",
                "fullWidth": true
              }
            ]
          }
        }
      ]
    }
  },
  "example": {
    "ui:field": "LayoutHeaderField"
  }
}
```

One gets the following UI:

Screenshot 2025-08-15 at 5 38 39 PM

### Checklist

### DIFF
--- a/packages/core/src/components/fields/LayoutGridField.tsx
+++ b/packages/core/src/components/fields/LayoutGridField.tsx
@@ -24,6 +24,7 @@ import {
   UI_OPTIONS_KEY,
   UI_GLOBAL_OPTIONS_KEY,
   UiSchema,
+  ITEMS_KEY,
 } from '@rjsf/utils';
 import cloneDeep from 'lodash/cloneDeep';
 import each from 'lodash/each';
@@ -39,6 +40,7 @@ import isObject from 'lodash/isObject';
 import isPlainObject from 'lodash/isPlainObject';
 import isString from 'lodash/isString';
 import isUndefined from 'lodash/isUndefined';
+import last from 'lodash/last';
 import set from 'lodash/set';
 
 /** The enumeration of the three different Layout GridTemplate type values
@@ -128,6 +130,15 @@ type UIComponentPropsType = {
  */
 function getNonNullishValue<T = unknown>(value?: T, fallback?: T): T | undefined {
   return value ?? fallback;
+}
+
+/** Detects if a `str` is made up entirely of numeric characters
+ *
+ * @param str - The string to check to see if it as number
+ * @return - True if the string consists entirely of numeric characters
+ */
+function isNumeric(str: string) {
+  return /^-?\d+(\.\d+)?$/.test(str); // Matches integers or decimals, positive or negative
 }
 
 /** The `LayoutGridField` will render a schema, uiSchema and formData combination out into a GridTemplate in the shape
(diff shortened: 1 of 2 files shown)


packages/core/src/components/fields/LayoutGridField.tsx
```diff
@@ -130,6 +132,15 @@ function getNonNullishValue<T = unknown>(value?: T, fallback?: T): T | undefined
...
 }
 
+/** Detects if a `str` is made up entirely of numeric characters
+ *
+ * @param str - The string to check to see if it as number
+ * @return - True if the string consists entirely of numeric characters
+ */
+function isNumeric(str: string) {
+  return /^-?\d+(\.\d+)?$/.test(str); // Matches integers or decimals, positive or negative
+}
```
nickgros: This doesn't seem appropriate to use for array indices. Should probably just check if it is a positive integer, we don't want to match if it's negative or if there's a decimal (neither would be valid array indices).

packages/core/test/LayoutGridField.test.tsx
```diff
@@ -1163,7 +1227,65 @@ describe('LayoutGridField', () => {
...
+        schema,
+        isRequired: false,
+        isReadonly: undefined,
+        idSchema: testGetIdSchema(path),
+        optionsInfo: undefined,
+      });
+      expect(retrieveSchemaSpy).toHaveBeenCalledTimes(2);
+      expect(toIdSchemaSpy).not.toHaveBeenCalled();
+    });
+    test('returns schema, isRequired: true, isReadonly: true, options: undefined when selecting readonly field', () => {
```
nickgros: this test title is also not accurate

packages/core/test/LayoutGridField.test.tsx
```diff
@@ -1163,7 +1227,65 @@ describe('LayoutGridField', () => {
...
+    afterEach(() => {
+      findSelectedOptionInXxxOf.mockClear();
+      retrieveSchemaSpy.mockClear();
+      toIdSchemaSpy.mockClear();
+    });
+    afterAll(() => {
+      retrieveSchemaSpy.mockRestore();
+      toIdSchemaSpy.mockRestore();
+    });
+    test('returns schema, isRequired: false, isReadonly: undefined, options when oneOf schema is requested', () => {
```
nickgros: this test title is not accurate w.r.t. what is being tested

packages/core/src/components/fields/LayoutGridField.tsx
heath-freenome: Good point! I can definitely remove the decimal part, although, with the splitting on `.` this will never be an issue. The negative index is definitely something we don't want.

packages/core/test/LayoutGridField.test.tsx
heath-freenome: oops, copy pasta

packages/core/test/LayoutGridField.test.tsx
heath-freenome: oops, copy pasta

PR #4735, opened 2025-08-26
fix: path duplication

### LINKED ISSUE #4733: Multiselect onChange not changing values
### Prerequisites

### What theme are you using?

mui

### Version

6.0.0-beta.14

### Current Behavior

Currently multiselect fields does not change form value, the `onChange` fires but value is not changed. I've tried with both default theme and mui theme, same issue.

Single select values do work.

See video:
https://github.com/user-attachments/assets/2bd0dee6-2887-4318-95d2-6be43f8f4bad

or gif
Image

### Expected Behavior

That the array values change.

### Steps To Reproduce

1. Create a multiselect field
2. Try to change the values

[Share link](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJjdXJyZW50Q29sb3IiOiIjMDBmZjAwIiwiY29sb3JNYXNrIjpbIiMwMDAwZmYiXSwidG9nZ2xlTWFzayI6dHJ1ZSwiY29sb3JQYWxldHRlIjpbIiNmZjAwMDAiXSwiYmxlbmRNb2RlIjoic2NyZWVuIn0sInNjaGVtYSI6eyJkZWZpbml0aW9ucyI6eyJDb2xvciI6eyJ0aXRsZSI6IkNvbG9yIiwidHlwZSI6InN0cmluZyIsImFueU9mIjpbeyJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyIjZmYwMDAwIl0sInRpdGxlIjoiUmVkIn0seyJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyIjMDBmZjAwIl0sInRpdGxlIjoiR3JlZW4ifSx7InR5cGUiOiJzdHJpbmciLCJlbnVtIjpbIiMwMDAwZmYiXSwidGl0bGUiOiJCbHVlIn1dfSwiVG9nZ2xlIjp7InRpdGxlIjoiVG9nZ2xlIiwidHlwZSI6ImJvb2xlYW4iLCJvbmVPZiI6W3sidGl0bGUiOiJFbmFibGUiLCJjb25zdCI6dHJ1ZX0seyJ0aXRsZSI6IkRpc2FibGUiLCJjb25zdCI6ZmFsc2V9XX19LCJ0aXRsZSI6IkltYWdlIGVkaXRvciIsInR5cGUiOiJvYmplY3QiLCJyZXF1aXJlZCI6WyJjdXJyZW50Q29sb3IiLCJjb2xvck1hc2siLCJibGVuZE1vZGUiXSwicHJvcGVydGllcyI6eyJjdXJyZW50Q29sb3IiOnsiJHJlZiI6IiMvZGVmaW5pdGlvbnMvQ29sb3IiLCJ0aXRsZSI6IkJydXNoIGNvbG9yIn0sImNvbG9yTWFzayI6eyJ0eXBlIjoiYXJyYXkiLCJ1bmlxdWVJdGVtcyI6dHJ1ZSwiaXRlbXMiOnsiJHJlZiI6IiMvZGVmaW5pdGlvbnMvQ29sb3IifSwidGl0bGUiOiJDb2xvciBtYXNrIn0sInRvZ2dsZU1hc2siOnsidGl0bGUiOiJBcHBseSBjb2xvciBtYXNrIiwiJHJlZiI6IiMvZGVmaW5pdGlvbnMvVG9nZ2xlIn0sImNvbG9yUGFsZXR0ZSI6eyJ0eXBlIjoiYXJyYXkiLCJ0aXRsZSI6IkNvbG9yIHBhbGV0dGUiLCJpdGVtcyI6eyIkcmVmIjoiIy9kZWZpbml0aW9ucy9Db2xvciJ9fSwiYmxlbmRNb2RlIjp7InRpdGxlIjoiQmxlbmQgbW9kZSIsInR5cGUiOiJzdHJpbmciLCJvbmVPZiI6W3siY29uc3QiOiJzY3JlZW4iLCJ0aXRsZSI6IlNjcmVlbiJ9LHsiY29uc3QiOiJtdWx0aXBseSIsInRpdGxlIjoiTXVsdGlwbHkifSx7ImNvbnN0Ijoib3ZlcmxheSIsInRpdGxlIjoiT3ZlcmxheSJ9XX19fSwidWlTY2hlbWEiOnsiYmxlbmRNb2RlIjp7InVpOmVudW1EaXNhYmxlZCI6WyJtdWx0aXBseSJdfSwidG9nZ2xlTWFzayI6eyJ1aTp3aWRnZXQiOiJyYWRpbyJ9fSwidGhlbWUiOiJkZWZhdWx0IiwibGl2ZVNldHRpbmdzIjp7InNob3dFcnJvckxpc3QiOiJ0b3AiLCJ2YWxpZGF0ZSI6ZmFsc2UsImRpc2FibGVkIjpmYWxzZSwibm9IdG1sNVZhbGlkYXRlIjpmYWxzZSwicmVhZG9ubHkiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlLCJleHBlcmltZW50YWxfY29tcG9uZW50VXBkYXRlU3RyYXRlZ3kiOiJjdXN0b21EZWVwIiwiZXhwZXJpbWVudGFsX2RlZmF1bHRGb3JtU3RhdGVCZWhhdmlvciI6eyJhcnJheU1pbkl0ZW1zIjoicG9wdWxhdGUiLCJhbGxPZiI6InNraXBEZWZhdWx0cyIsImNvbnN0QXNEZWZhdWx0cyI6ImFsd2F5cyIsImVtcHR5T2JqZWN0RmllbGRzIjoicG9wdWxhdGVBbGxEZWZhdWx0cyIsIm1lcmdlRGVmYXVsdHNJbnRvRm9ybURhdGEiOiJ1c2VGb3JtRGF0YUlmUHJlc2VudCJ9fSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJBbHRlcm5hdGl2ZXMifQ==)

### Environment

```markdown
- OS: Windows 11
- Node: 22.15.1
- npm: 10.9.2
```

### Anything else?

_No response_

Fronix:
Fixed #4721 that broke multiselect fields onchange by duplicating the path with `slice()` 

This fixes #4733

### Checklist

### DIFF
--- a/packages/core/src/components/fields/ObjectField.tsx
+++ b/packages/core/src/components/fields/ObjectField.tsx
@@ -1,28 +1,28 @@
-import { Component } from 'react';
 import {
-  getTemplate,
-  getUiOptions,
-  orderProperties,
+  ADDITIONAL_PROPERTY_FLAG,
+  ANY_OF_KEY,
   ErrorSchema,
   FieldProps,
   FormContextType,
   GenericObjectType,
+  getTemplate,
+  getUiOptions,
   IdSchema,
+  ONE_OF_KEY,
+  orderProperties,
+  PROPERTIES_KEY,
+  REF_KEY,
   RJSFSchema,
   StrictRJSFSchema,
   TranslatableString,
-  ADDITIONAL_PROPERTY_FLAG,
-  PROPERTIES_KEY,
-  REF_KEY,
-  ANY_OF_KEY,
-  ONE_OF_KEY,
 } from '@rjsf/utils';
-import Markdown from 'markdown-to-jsx';
 import get from 'lodash/get';
 import has from 'lodash/has';
 import isObject from 'lodash/isObject';
 import set from 'lodash/set';
 import unset from 'lodash/unset';
+import Markdown from 'markdown-to-jsx';
+import { Component } from 'react';
 
 /** Type used for the state of the `ObjectField` component */
 type ObjectFieldState = {
@@ -80,7 +80,10 @@ class ObjectField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
       }
       // Copy the current path and insert in the property name into the first location
       const changePath = Array.isArray(path) ? path.slice() : [];
-      changePath.unshift(name);
+      // Ensure we are not duplicating the path
+      if (!changePath.length || changePath[0] !== name) {
+        changePath.unshift(name);
+      }
       onChange(value, changePath, newErrorSchema, id);
     };
   };


codeliner:
I stumbled across the same issue today. @Fronix I think your fix could cause unwanted side effects. When I have an object with a property "prop" which again is an object with a property "prop", your fix would make it impossible to change the value for "prop.prop". At least that's my understanding of the method.

While debugging the issue, I identified the ArrayField.onSelectChange method to be problematic: https://github.com/rjsf-team/react-jsonschema-form/blob/357306523d1420e510a85ff76cdc84e9d584b638/packages/core/src/components/fields/ArrayField.tsx#L423

If you look at other methods of ArrayField, you can see that an empty path array is provided to onChange with a comment that the appropriate path is added by onChange: https://github.com/rjsf-team/react-jsonschema-form/blob/357306523d1420e510a85ff76cdc84e9d584b638/packages/core/src/components/fields/ArrayField.tsx#L302

I'd say onSelectChange needs to do the same (pass an empty path array to onChange). However, I don't know if this would break other use cases ...

Fronix:
@codeliner 

That is true, I didn't think of the case of `prop.prop`. One quick way I tested (and worked) is this
https://github.com/rjsf-team/react-jsonschema-form/blob/357306523d1420e510a85ff76cdc84e9d584b638/packages/core/src/components/fields/ArrayField.tsx#L419C1-L424C5

Changing to this
```typescript
  /** Callback handler used to change the value for a checkbox */
  onSelectChange = (value: any) => {
    const { onChange, idSchema } = this.props;
    // To ensure it's possible to change prop.prop.... by extracting the correct path from the idSchema.$id
    const path = idSchema && idSchema.$id ? idSchema.$id.split('_').slice(1) : [];
    onChange(value, path, undefined, idSchema && idSchema.$id);
  };
```

and then keeping this would allow for nested properties, with the same name, to work

```typescript
      if (!changePath.length || changePath[0] !== name) {
        changePath.unshift(name);
      }
```

I could also just be very unfamiliar with the inner workings, and there might be a much better way to do it but without knowing the **real** path of the properties, it feels impossible to build the path as you go and then using unshift without breaking non nested objects.

Edit: There can also be an argument to not support `prop.prop.prop ...` behaviour and throw an error if it occurs. But again, we probably need one of the maintainers to chime in here

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -80,7 +80,10 @@ class ObjectField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends Fo
       }
       // Copy the current path and insert in the property name into the first location
       const changePath = Array.isArray(path) ? path.slice() : [];
-      changePath.unshift(name);
+      // Ensure we are not duplicating the path
+      if (!changePath.length || changePath[0] !== name) {
```
heath-freenome, line 84: Since there is similar code in `ArrayField` might I consider making a similar change there? AND there is one problem with this fix...if there were nested properties with the same name, this would cause problems with incorrect nesting.

Can you provide a codesandbox example where you are seeing this problem so that we can better understand why this failed for you.

I.e.
```
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "object",
      "properties": {
        "foo": {
          "type": "string",
        }
      }
    }
  }
}
```

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -1,28 +1,28 @@
...
-  ONE_OF_KEY,
 } from '@rjsf/utils';
-import Markdown from 'markdown-to-jsx';
 import get from 'lodash/get';
 import has from 'lodash/has';
 import isObject from 'lodash/isObject';
 import set from 'lodash/set';
 import unset from 'lodash/unset';
+import Markdown from 'markdown-to-jsx';
+import { Component } from 'react';
```
heath-freenome, line 25: Can you move this back to the top? We prefer react imports to be first

heath-freenome:
@Fronix Thanks for trying... I just did some debugging myself and made the much simpler fix https://github.com/rjsf-team/react-jsonschema-form/pull/4737

PR #4773, opened 2025-09-18
feat: Add support for user `nameGenerator` function

### LINKED ISSUE #4693: Compliant form elements names for "old" submitted data (POST)
### Prerequisites

### What theme are you using?

core

### Version

6.0.0-beta-12

### Current Behavior

Hi,

First, thank for your work on that package. It's an amazing library to work with 👏

Since 6 is not out yet, is there any possibility to make (array) field names compliant with "traditional" form submission (e.g. hitting a submit button and handling a good old POST browser form request)?

Take the "Array" example from playground and look at the field names, for example "listOfStrings", HTML is:

```html
<input id="root_listOfStrings_0" name="root_listOfStrings_0" class="form-control" label="listOfStrings-0" required="" placeholder="" type="text" aria-describedby="root_listOfStrings_0__error root_listOfStrings_0__description root_listOfStrings_0__help" value="foo">
<input id="root_listOfStrings_1" name="root_listOfStrings_1" class="form-control" label="listOfStrings-1" required="" placeholder="" type="text" aria-describedby="root_listOfStrings_1__error root_listOfStrings_1__description root_listOfStrings_1__help" value="bar">
<!-- and so on -->
```

If you submit that form using a dummy submit button and check your payload, you will not get an array. Same goes for any other "array" fields (repeaters, etc.).

### Expected Behavior

I would expect the form to be 

```html
<input id="root_listOfStrings_0" name="root_listOfStrings[0]" class="form-control" label="listOfStrings-0" required="" placeholder="" type="text" aria-describedby="root_listOfStrings_0__error root_listOfStrings_0__description root_listOfStrings_0__help" value="foo">
<input id="root_listOfStrings_1" name="root_listOfStrings[1]" class="form-control" label="listOfStrings-1" required="" placeholder="" type="text" aria-describedby="root_listOfStrings_1__error root_listOfStrings_1__description root_listOfStrings_1__help" value="bar">
<!-- and so on -->
```

I already know I could submit the data using JS as a workaround. But in my case, that is not an option because I'm rendering my RJSF form inside a already there "traditional" form. And for the love of web standards, I would love RJSF to be compliant on field names. 

I tried to take a look at the code and submit a PR but I don't have a deep enough understanding of the codebase to make it work.

Note that setting a form `name` will also imply that all fields will be prefixed by that name:

```html
<form name="root">
    <input name="root[firstName]" />
</form>
```

So in case multiple `firstNames` are submitted:

```html
<form name="root">
    <input name="root[firstName][]" /> <!-- OR root[firstName][0] -->
    <input name="root[firstName][]" /> <!-- OR root[firstName][1] -->
</form>
```

Let me know if I can help (a little guidance on where to look at might be welcome).

### Steps To Reproduce

Check "Array" playground example.

### Environment

```markdown
- OS: Mac OS
- Node: 20.x
- npm: 11.x
```

### Anything else?

_No response_

nlemoine:

Fixes #4693

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,9 +86,6 @@ should change the heading of the (upcoming) version to include a major version b
 - BREAKING CHANGE: Updated `FieldTemplateProps` and `WrapIfAdditionalTemplateProps` to replace the `onKeyChange()` and `onDropPropertyClick()` callback generator props with the `onKeyRename()`, `onKeyRenameBlur()` and `onRemoveProperty()` callback props
 - BREAKING CHANGE: Updated `ObjectFieldTemplateProps` to replace the `onAddClick()` callback generator prop with the `onAddProperty()` callback prop
 - Added new hook `useDeepCompareMemo()` and its associated tests
-
-## @rjsf/utils
-
 - Added `NameGeneratorFunction` type and two built-in name generators: `bracketNameGenerator` and `dotNotationNameGenerator`
 - Updated `GlobalFormOptions` type to include optional `nameGenerator` field
 - Updated `toFieldPathId()` function to support name generation via the `nameGenerator` option in `GlobalFormOptions`
--- a/packages/core/src/components/fields/ArrayField.tsx
+++ b/packages/core/src/components/fields/ArrayField.tsx
@@ -603,7 +603,6 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
-
     // For custom widgets with multiple=true, generate a fieldPathId with isMultiValue flag
     const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
 
@@ -661,7 +660,6 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
-
     // For multi-select widgets, generate a fieldPathId with isMultiValue flag
     const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
 
@@ -715,7 +713,6 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
-
     // For file widgets with multiple=true, generate a fieldPathId with isMultiValue flag
     const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
 


packages/utils/src/schema/toPathSchema.ts
```diff
@@ -176,7 +191,17 @@ export default function toPathSchema<T = any, S extends StrictRJSFSchema = RJSFS
   name = '',
   rootSchema?: S,
   formData?: T,
+  segments: PathSegment[] = [],
   experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>,
```
heath-freenome: Putting it at the end is a better choice to avoid a breaking change here
```suggestion
  experimental_customMergeAllOf?: Experimental_CustomMergeAllOf<S>,
  segments: PathSegment[] = [],
```

packages/utils/src/schema/toPathSchema.ts
```diff
@@ -167,6 +181,7 @@ function toPathSchemaInternal<T = any, S extends StrictRJSFSchema = RJSFSchema,
  * @param [name=''] - The base name for the schema
  * @param [rootSchema] - The root schema, used to primarily to look up `$ref`s
  * @param [formData] - The current formData, if any, to assist retrieving a schema
+ * @param [segments=[]] - The path segments array to build upon
  * @param [experimental_customMergeAllOf] - Optional function that allows for custom merging of `allOf` schemas
```
heath-freenome:
```suggestion
 * @param [experimental_customMergeAllOf] - Optional function that allows for custom merging of `allOf` schemas
 * @param [segments=[]] - The path segments array to build upon
```

packages/core/src/components/Form.tsx
```diff
@@ -1158,6 +1176,14 @@ export default class Form<
...
           errorSchema={errorSchema}
           idSchema={idSchema}
+          pathSchema={(() => {
+            try {
+              return isObject(schema) ? registry.schemaUtils.toPathSchema(schema, '', formData) : undefined;
+            } catch {
+              // Silently handle malformed schemas in tests
+              return undefined;
+            }
+          })()}
```
heath-freenome: Since this is a one-shot call in the root `Form` just compute it first into a local variable above and then use that variable here

packages/core/src/components/Form.tsx
```diff
@@ -212,6 +213,16 @@ export interface FormProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
...
+  /** Optional function to generate HTML name attributes from path segments
+   * @param segments - Array of path segments representing the field hierarchy
+   * @param rootName - Optional root name prefix
+   * @returns - The generated name attribute string, or undefined to use default behavior
+   */
+  nameGenerator?: NameGeneratorFunction;
+  /** Optional root name to use with nameGenerator
+   * @default "root"
+   */
+  rootName?: string;
```
heath-freenome: If you are required to pass the nameGenerator implementation in, why not just have that generator know what it wants to use for the `root`. This will avoid an unnecessary prop.

I noticed that you've provided some sample implementations of these in `utils`. If someone wanted to just use them straight up, but with a different `rootName` they could wrap them?

packages/utils/src/types.ts
```diff
@@ -27,6 +27,17 @@ export type StrictRJSFSchema = JSONSchema7;
...
+/** Represents a segment in the path to a form field */
+export interface PathSegment {
+  /** Type of the segment - object property or array index */
+  type: 'object' | 'array';
+  /** The key or index value */
+  key: string | number;
+}
+
+/** Function to generate HTML name attributes from path segments */
+export type NameGeneratorFunction = (segments: PathSegment[], rootName?: string) => string | undefined;
```
heath-freenome: No need for the `rootName` parameter since you can implement the name generator however you want. Thus you can hard-code in the `rootName` there.

packages/core/src/components/Form.tsx
```diff
@@ -972,7 +984,13 @@ export default class Form<
...
       widgets: { ...widgets, ...this.props.widgets },
       rootSchema: schema,
-      formContext: this.props.formContext || formContext,
+      formContext: {
+        ...(this.props.formContext || formContext),
+        ...(this.props.nameGenerator && {
+          nameGenerator: this.props.nameGenerator,
+          rootName: this.props.rootName || 'root',
+        }),
+      },
```
heath-freenome: We are about to [update](https://github.com/rjsf-team/react-jsonschema-form/pull/4776) `Registry` with a `globalOptions?: GlobalFormOptions` which is designed to contain `Form` related option that must be pass along to everything. Consider adding it to there instead. If you'd rather move forward faster, feel free to initially add the prop to the `Registry`

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -534,6 +537,14 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
         const itemIdSchema = schemaUtils.toIdSchema(itemSchema, itemIdPrefix, itemCast, idPrefix, idSeparator);
 
+        // Create path schema for array item with segments
+        const itemPathSegments: PathSegment[] = [...(pathSchema?.$segments || []), { type: 'array', key: index }];
+
+        const itemPathSchema: PathSchema<T> = {
+          $name: `${pathSchema?.$name || ''}.${index}`,
+          $segments: itemPathSegments,
+        };
+
```
heath-freenome: This all seems like a lot of extra work, just for the situation where someone MIGHT be using a `nameGenerator`. I wonder if there is a way to make this whole feature be a bit more of a plug-in rather than a hard-wire? Perhaps leveraging the `registry` to wrap the whole concept, with the `pathSchema` and all into a new `getHtmlName(registry)` function... Nah that gets more challenging.

Since one of your generators basically is designed to generate a dotted path name/id which is already covered by using the `idSeparator` flag set to `'.'`, all you are wanting to accomplish is changing the id to switch the index separator from `idSeparator` to `[]`. Which takes me back to my original suggestion of adding a new option to `toIdSchema()` to accomplish that, especially since `toIdSchema()` is only called in the following places:
```
SchemaField on line 136,
LayoutGridField on line 507,
ArrayField on lines 789 and 535,
Form on 527
```

Especially since we are considering making the `idPrefix` and `idSeparator` be part of the `GlobalFormOptions` in the `registry` [here](https://github.com/rjsf-team/react-jsonschema-form/pull/4776) so you would only have to update 5 places. rather than do more wiring.

heath-freenome:
@nlemoine I guess the TLDR is I'm not a fan of the approach, as it is too complex for your use case. I am open to you arguing your case for this approach over a simple flag to update the `toIdSchema()` generation code.

nlemoine:
Hey @heath-freenome, thanks for the thorough review! I really appreciate your time and feedback.

I understand your comments and will try to address your concerns, particularly about using `toIdSchema` instead of a separate path segments system. After careful consideration, I believe we need separate ID and name generation for both technical and practical reasons. Let me explain why.

## Backend need examples

`toIdSchema` generates flat strings with consistent separators, but backends need structured notation that preserves the data hierarchy. This isn't just about changing separators or wrapping id segments within brackets - it's about preserving structural information and providing users to generate the name they might need. It may vary depending on the backend integration:

- **[PHP](https://www.php.net/manual/en/faq.html.php#faq.html.arrays)/[Rails](https://guides.rubyonrails.org/action_controller_overview.html)**: `root[tasks][0][title]` - brackets with or without array indexes
- **[Django](https://docs.djangoproject.com/en/5.2/topics/forms/formsets/)**: `root__tasks-0__title` - different separators for objects vs arrays

## The `toIdSchema` flag approach problem

Correct me if I'm wrong, but implementing a `toIdSchema`  flag would result in `id`attributes being equals to `name` attributes.

And handling array indexes would also need changing code elsewhere because array indexes are not part of `toIdSchema` arguments:

https://github.com/rjsf-team/react-jsonschema-form/blob/ba61a8a3bb930e1a764e7980f5cf64636f1c42fa/packages/core/src/components/fields/ArrayField.tsx#L534-L542

I gave a quick and dirty try to that approach and it gave me:

```html
<input id="root[listOfStrings]_0" name="root[listOfStrings]_0" class="form-control" label="listOfStrings-0" required="" placeholder="" type="text" aria-describedby="root[listOfStrings]_0__error root[listOfStrings]_0__description root[listOfStrings]_0__help" value="foo">
```

But most importantly, it would bring some more major issues, IMHO.

## HTML Requires different IDs and names

Even with a `toIdSchema` working solution (e.g. generating bracketted ids), this isn't just about backend preferences - HTML itself requires different values for `id` and `name` attributes in several fundamental cases:

### 1. Radio button groups

Radio buttons in the same group **must** share the same `name` but have different `id` values:

```html
<!-- ✅ Same name, different IDs -->
<input type="radio" id="root_color_0" name="root[color]" value="red">
<label for="root_color_0">Red</label>

<input type="radio" id="root_color_1" name="root[color]" value="blue">
<label for="root_color_1">Blue</label>

<!-- ❌ Using ID as name breaks radio grouping -->
<input type="radio" id="root_color_0" name="root_color_0" value="red">
<input type="radio" id="root_color_1" name="root_color_1" value="blue">
```

Radio group is already ok in current RJSF state but if `id` is used as `name`, this will no longer be the case.

### 2. Checkbox arrays

Multiple checkboxes that submit as an array need the same base name:

```html
<!-- ✅ PHP-style array submission -->
<input type="checkbox" id="root_hobbies_0" name="root[hobbies][]" value="reading">
<input type="checkbox" id="root_hobbies_1" name="root[hobbies][]" value="gaming">
<input type="checkbox" id="root_hobbies_2" name="root[hobbies][]" value="cooking">
<!-- Submits as: hobbies = ['reading', 'gaming'] when multiple are checked -->

<!-- ❌ Using ID as name -->
<input type="checkbox" id="root_hobbies_0" name="root_hobbies_0" value="reading">
<input type="checkbox" id="root_hobbies_1" name="root_hobbies_1" value="gaming">
<!-- Submits as separate unrelated values, not an array! -->
```

Same as above.

### 3. Select multiple

Similar issue with multi-select dropdowns:

```html
<!-- Needs array notation in name -->
<select id="root_skills" name="root[skills][]" multiple>
  <option value="js">JavaScript</option>
  <option value="python">Python</option>
</select>
```

## Why brackets in IDs are complicated

Even if we wanted to use the same bracketed format for both ID and name (like `root[tasks][0]`), it would break or weaken major parts of the web stack:

### CSS selectors

```css
/* Brackets are special characters in CSS */
#root[tasks][0] { }  /* ❌ BREAKS - interpreted as attribute selector */

/* Would require escaping everywhere */
#root\[tasks\]\[0\] { }  /* Ugly and error-prone */

/* Current RJSF approach */
#root_tasks_0 { }  /* ✅ Clean, simple, works everywhere */
```

### JavaScript DOM Queries

```javascript
// querySelector fails with special characters
document.querySelector('#root[tasks][0]')  // ❌ SyntaxError

// Would need escaping
document.querySelector('#root\\[tasks\\]\\[0\\]')  // Complex escaping

// Current clean approach
document.querySelector('#root_tasks_0')  // ✅ Works everywhere
```

---

I would have loved a much simpler solution too. Adding a flag in `toIdSchema` feels like a workaround more than a robust and flexible alternative.

I tried to think of a one that brings the minimal possible changes. It can surely be improved (JS/React is not my strong suit) and I will address your review comments if you change your mind.

However, if you are still not convinced and don't see any benefits in bringing separate logic for `id` and `name` attributes, I would totally understand. I'm ready to work on a `toIdSchema` alternative, since I'd really like that feature to make it for version 6.

What do you think? I'll be happy to refactor the implementation based on your preferences.

heath-freenome:
@nlemoine Thank you for the well documented explanation. It has convinced me and my compatriot @nickgros that there is merit in your approach. So much so, that I've just pushed a [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4787) that does a major breaking refactor of the system to simplify your work quite a bit.

In order to implement your feature, I am going to recommend the following approach after my PR merges:

1. Add the `nameGenerator` to the `GlobalFormOptions` as optional
2. Update the `FieldPathId` to add an optional `name` variable
3. Update the `toFieldPathId()` function to generate the `name` if the `nameGenerator` exists (you can check the type of the element in the `path` prop to determine object or array)
4. Update the `WidgetProps` to add the optional `htmlName` prop
5. Update everywhere the `<Widget ...>` is used in the codebase to add `htmlName={fieldPathId.name}`
6. Update all of the widget implementations in the codebase to add `htmlName || id` for the `name` props
7. Update all of the tests to deal with that change
8. Update the docs to add the optional `htmlName` to the `Widget` docs
9. Update the `v6.x migration guide` to describe this new feature as well as let custom widget developers know about the `htmlName` props

nlemoine:
HI @heath-freenome,

Thank you so much for paying so much attention to that issue and bringing so much changes to that purpose! 🙌 

I'll try to update the PR as soon as I can and will keep posted.

packages/utils/src/schema/toPathSchema.ts
nlemoine: Closing as it is outdated now.

packages/utils/src/schema/toPathSchema.ts
nlemoine: Closing as it is outdated now.

packages/core/src/components/Form.tsx
nlemoine: Closing as it is outdated now.

packages/core/src/components/Form.tsx
nlemoine: Closing as it is outdated now.

packages/core/src/components/fields/ArrayField.tsx
nlemoine: Closing as it is outdated now.

nlemoine:
Quick questions @heath-freenome 

- Do I have to update every RJSF theme?
- I'm having an issue the current implementation doesn't solve. When [displaying array](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsaXN0T2ZTdHJpbmdzIjpbImZvbyIsImJhciJdLCJtdWx0aXBsZUNob2ljZXNMaXN0IjpbImZvbyIsImJhciJdLCJmaXhlZEl0ZW1zTGlzdCI6WyJTb21lIHRleHQiLHRydWUsMTIzXSwibWluSXRlbXNMaXN0IjpbeyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn1dLCJkZWZhdWx0c0FuZE1pbkl0ZW1zIjpbImNhcnAiLCJ0cm91dCIsImJyZWFtIiwidW5pZGVudGlmaWVkIiwidW5pZGVudGlmaWVkIl0sIm5lc3RlZExpc3QiOltbImxvcmVtIiwiaXBzdW0iXSxbImRvbG9yIl1dLCJ1bm9yZGVyYWJsZSI6WyJvbmUiLCJ0d28iXSwiY29weWFibGUiOlsib25lIiwidHdvIl0sInVucmVtb3ZhYmxlIjpbIm9uZSIsInR3byJdLCJub1Rvb2xiYXIiOlsib25lIiwidHdvIl0sImZpeGVkTm9Ub29sYmFyIjpbNDIsdHJ1ZSwiYWRkaXRpb25hbCBpdGVtIG9uZSIsImFkZGl0aW9uYWwgaXRlbSB0d28iXX0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJtdWx0aXBsZUNob2ljZXNMaXN0Ijp7InR5cGUiOiJhcnJheSIsInRpdGxlIjoiQSBtdWx0aXBsZSBjaG9pY2VzIGxpc3QiLCJpdGVtcyI6eyJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyJmb28iLCJiYXIiLCJmdXp6IiwicXV4Il19LCJ1bmlxdWVJdGVtcyI6dHJ1ZX19fSwidWlTY2hlbWEiOnsibXVsdGlwbGVDaG9pY2VzTGlzdCI6eyJ1aTp3aWRnZXQiOiJjaGVja2JveGVzIn19LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In19LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IkFycmF5cyJ9) as checkboxes, there is no `index` and the `elementType` is still an object. So no bracket is generated (e.g. `root[multipleChoicesList][]`). I currently don't have the information that tells me if I should use brackets or not. Do you have any idea for that use case?
- > I noticed that you've provided some sample implementations of these in utils. If someone wanted to just use them straight up, but with a different rootName they could wrap them?

    Yes they could but say you want to output multiple forms on the same page and want different root names, you will have to create multiple wrapper functions to get this done, which is quite ugly. However, I would be fine reusing the `idPrefix` as a parameter of the name generator function to provide a custom root name. Imagine the same logic for Ids, wrapping `toIdSchema` for each customization would be cumbersome. 
 - Do you want me to remove initially provided nameGenerator functions (in utils, bracket/dot notation you mentioned in the quote above)? I just committed them to enable an easier playground preview but I would be fine and they're not part of RJSF.

heath-freenome:
> Quick questions @heath-freenome
> 
> * Do I have to update every RJSF theme?
Yes, every theme will need updating, sorry. New features need to work across all published themes

> * I'm having an issue the current implementation doesn't solve. When [displaying array](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsaXN0T2ZTdHJpbmdzIjpbImZvbyIsImJhciJdLCJtdWx0aXBsZUNob2ljZXNMaXN0IjpbImZvbyIsImJhciJdLCJmaXhlZEl0ZW1zTGlzdCI6WyJTb21lIHRleHQiLHRydWUsMTIzXSwibWluSXRlbXNMaXN0IjpbeyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn1dLCJkZWZhdWx0c0FuZE1pbkl0ZW1zIjpbImNhcnAiLCJ0cm91dCIsImJyZWFtIiwidW5pZGVudGlmaWVkIiwidW5pZGVudGlmaWVkIl0sIm5lc3RlZExpc3QiOltbImxvcmVtIiwiaXBzdW0iXSxbImRvbG9yIl1dLCJ1bm9yZGVyYWJsZSI6WyJvbmUiLCJ0d28iXSwiY29weWFibGUiOlsib25lIiwidHdvIl0sInVucmVtb3ZhYmxlIjpbIm9uZSIsInR3byJdLCJub1Rvb2xiYXIiOlsib25lIiwidHdvIl0sImZpeGVkTm9Ub29sYmFyIjpbNDIsdHJ1ZSwiYWRkaXRpb25hbCBpdGVtIG9uZSIsImFkZGl0aW9uYWwgaXRlbSB0d28iXX0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJtdWx0aXBsZUNob2ljZXNMaXN0Ijp7InR5cGUiOiJhcnJheSIsInRpdGxlIjoiQSBtdWx0aXBsZSBjaG9pY2VzIGxpc3QiLCJpdGVtcyI6eyJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyJmb28iLCJiYXIiLCJmdXp6IiwicXV4Il19LCJ1bmlxdWVJdGVtcyI6dHJ1ZX19fSwidWlTY2hlbWEiOnsibXVsdGlwbGVDaG9pY2VzTGlzdCI6eyJ1aTp3aWRnZXQiOiJjaGVja2JveGVzIn19LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In19LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IkFycmF5cyJ9) as checkboxes, there is no `index` and the `elementType` is still an object. So no bracket is generated (e.g. `root[multipleChoicesList][]`). I currently don't have the information that tells me if I should use brackets or not. Do you have any idea for that use case?

Right now the `optionId()` utility function in the `CheckboxesWidget` implementations are adding the `-{index#}` notations. Those are ids, not names. The name should be the same for all of the checkboxes, so simply passing `htmlName | id` still works fine.

> * > I noticed that you've provided some sample implementations of these in utils. If someone wanted to just use them straight up, but with a different rootName they could wrap them?
Or you could make the name generator take the `idPrefix` as an optional parameter.
>   
>   
>   Yes they could but say you want to output multiple forms on the same page and want different root names, you will have to create multiple wrapper functions to get this done, which is quite ugly. However, I would be fine reusing the `idPrefix` as a parameter of the name generator function to provide a custom root name.
Right now the only solution people have to this anyway is to render two different `Form` elements passing different `idPrefix` values, so let's just stick with that

> * Do you want me to remove initially provided nameGenerator functions (in utils, bracket/dot notation you mentioned in the quote above)? I just committed them to enable an easier playground preview but I would be fine and they're not part of RJSF.

I'm fine with NOT providing them. You will still need something to test with so perhaps adding them is helpful?

packages/utils/src/types.ts
nlemoine: As discussed in https://github.com/rjsf-team/react-jsonschema-form/pull/4773#issuecomment-3373691770, the `idPrefix` will be used as root name.

nlemoine:
> Right now the optionId() utility function in the CheckboxesWidget implementations are adding the -{index#} notations. Those are ids, not names. The name should be the same for all of the checkboxes, so simply passing htmlName | id still works fine.

Actually, no, multiple checkboxes are considered an array. So [they need brackets](https://stackoverflow.com/questions/4631224/getting-multiple-checkboxes-names-ids-with-php) (`root[multipleChoicesList][]` or `root[multipleChoicesList][0]`). 

The name is **only** generated in `nameGenerator` and [in the case I showed you](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsaXN0T2ZTdHJpbmdzIjpbImZvbyIsImJhciJdLCJtdWx0aXBsZUNob2ljZXNMaXN0IjpbImZvbyIsImJhciJdLCJmaXhlZEl0ZW1zTGlzdCI6WyJTb21lIHRleHQiLHRydWUsMTIzXSwibWluSXRlbXNMaXN0IjpbeyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn1dLCJkZWZhdWx0c0FuZE1pbkl0ZW1zIjpbImNhcnAiLCJ0cm91dCIsImJyZWFtIiwidW5pZGVudGlmaWVkIiwidW5pZGVudGlmaWVkIl0sIm5lc3RlZExpc3QiOltbImxvcmVtIiwiaXBzdW0iXSxbImRvbG9yIl1dLCJ1bm9yZGVyYWJsZSI6WyJvbmUiLCJ0d28iXSwiY29weWFibGUiOlsib25lIiwidHdvIl0sInVucmVtb3ZhYmxlIjpbIm9uZSIsInR3byJdLCJub1Rvb2xiYXIiOlsib25lIiwidHdvIl0sImZpeGVkTm9Ub29sYmFyIjpbNDIsdHJ1ZSwiYWRkaXRpb25hbCBpdGVtIG9uZSIsImFkZGl0aW9uYWwgaXRlbSB0d28iXX0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJtdWx0aXBsZUNob2ljZXNMaXN0Ijp7InR5cGUiOiJhcnJheSIsInRpdGxlIjoiQSBtdWx0aXBsZSBjaG9pY2VzIGxpc3QiLCJpdGVtcyI6eyJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyJmb28iLCJiYXIiLCJmdXp6IiwicXV4Il19LCJ1bmlxdWVJdGVtcyI6dHJ1ZX19fSwidWlTY2hlbWEiOnsibXVsdGlwbGVDaG9pY2VzTGlzdCI6eyJ1aTp3aWRnZXQiOiJjaGVja2JveGVzIn19LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In19LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IkFycmF5cyJ9), the [`path` arg](https://github.com/rjsf-team/react-jsonschema-form/pull/4773/files#diff-3a71ba6c93cdf01fc6dd4bbec719af3239f8785ce269d8c6669654e72e61a49cR27) only contains `['multipleChoicesList']`. With that single arg, I have no way to tell (inside nameGenerator) if I should add brackets or not. Once in widget, it's "too late" to change the name (no `optionId` way). That's why my first implementation added indexes as an arg of the name generator.

heath-freenome:
> > Right now the optionId() utility function in the CheckboxesWidget implementations are adding the -{index#} notations. Those are ids, not names. The name should be the same for all of the checkboxes, so simply passing htmlName | id still works fine.
> 
> Actually, no, multiple checkboxes are considered an array. So [they need brackets](https://stackoverflow.com/questions/4631224/getting-multiple-checkboxes-names-ids-with-php) (`root[multipleChoicesList][]` or `root[multipleChoicesList][0]`).
> 
> The name is **only** generated in `nameGenerator` and [in the case I showed you](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsaXN0T2ZTdHJpbmdzIjpbImZvbyIsImJhciJdLCJtdWx0aXBsZUNob2ljZXNMaXN0IjpbImZvbyIsImJhciJdLCJmaXhlZEl0ZW1zTGlzdCI6WyJTb21lIHRleHQiLHRydWUsMTIzXSwibWluSXRlbXNMaXN0IjpbeyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn0seyJuYW1lIjoiRGVmYXVsdCBuYW1lIn1dLCJkZWZhdWx0c0FuZE1pbkl0ZW1zIjpbImNhcnAiLCJ0cm91dCIsImJyZWFtIiwidW5pZGVudGlmaWVkIiwidW5pZGVudGlmaWVkIl0sIm5lc3RlZExpc3QiOltbImxvcmVtIiwiaXBzdW0iXSxbImRvbG9yIl1dLCJ1bm9yZGVyYWJsZSI6WyJvbmUiLCJ0d28iXSwiY29weWFibGUiOlsib25lIiwidHdvIl0sInVucmVtb3ZhYmxlIjpbIm9uZSIsInR3byJdLCJub1Rvb2xiYXIiOlsib25lIiwidHdvIl0sImZpeGVkTm9Ub29sYmFyIjpbNDIsdHJ1ZSwiYWRkaXRpb25hbCBpdGVtIG9uZSIsImFkZGl0aW9uYWwgaXRlbSB0d28iXX0sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJtdWx0aXBsZUNob2ljZXNMaXN0Ijp7InR5cGUiOiJhcnJheSIsInRpdGxlIjoiQSBtdWx0aXBsZSBjaG9pY2VzIGxpc3QiLCJpdGVtcyI6eyJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyJmb28iLCJiYXIiLCJmdXp6IiwicXV4Il19LCJ1bmlxdWVJdGVtcyI6dHJ1ZX19fSwidWlTY2hlbWEiOnsibXVsdGlwbGVDaG9pY2VzTGlzdCI6eyJ1aTp3aWRnZXQiOiJjaGVja2JveGVzIn19LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In19LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IkFycmF5cyJ9), the [`path` arg](https://github.com/rjsf-team/react-jsonschema-form/pull/4773/files#diff-3a71ba6c93cdf01fc6dd4bbec719af3239f8785ce269d8c6669654e72e61a49cR27) only contains `['multipleChoicesList']`. With that single arg, I have no way to tell (inside nameGenerator) if I should add brackets or not. Once in widget, it's "too late" to change the name (no `optionId` way). That's why my first implementation added indexes as an arg of the name generator.

I'm not sure I understand why the indexes are needed for the name of checkboxes in a checkbox group? Since submitting the form will automatically cause the html to return `name=X&name=Y&name=Z` to the submitted form, which, using standard HTML parsing should result in `name=[X,Y,Z]`

nlemoine:
> I'm not sure I understand why the indexes are needed for the name of checkboxes in a checkbox group?

In PHP at least, this makes a huge difference when getting posted data (same name VS same name[]):

Capture d’écran 2025-10-07 à 21 19 12
Capture d’écran 2025-10-07 à 21 19 39

Long story short: when field(s) posted values are an array, they must use brackets (set of checkboxes, select with `multiple` attribute).

heath-freenome:
> > I'm not sure I understand why the indexes are needed for the name of checkboxes in a checkbox group?
...
> Long story short: when field(s) posted values are an array, they must use brackets (set of checkboxes, select with `multiple` attribute).

My suggestion would be to create a new utility function in `@rjsf/utils` called `toOptionName<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(id: string, htmlName: string | undefined, registry: Registry<T, S, F>, index: number): string` function that uses the `nameGenerator` to append the `[index]` onto the end of the `htmlName` as needed for the index,  when `htmlName` and `nameGenerator` exist, otherwise simply return `htmlName | id`. Then update all the widgets that need it to use it for the name.

heath-freenome:
@nlemoine Will you have this feature completed sometime in the next 10 days? I'd like to get it into a prerelease of v6 before we go live with the official release

nlemoine:
Hi @heath-freenome,

  Sorry for the delay - I had a busy week with limited time to work on this.

  I've finally managed to put everything together. I couldn't go with the solution you initially suggested. The key insight is that for multiple `<input type="checkbox" name="root[hobbies][]" />` or `<select multiple name="root[hobbies][]">`, what matters is knowing whether we're handling **multiple values** rather than the specific index. A `<select multiple>` is a single field tag that handles multiple values, while checkboxes are multiple input tags that also handle multiple values - but both need the same `[]` suffix in bracket notation.

  This is why I added an optional `isMultiValue` parameter to `NameGeneratorFunction` and `toFieldPathId`. This approach:
  - Works for both checkboxes and multi-select dropdowns
  - Lets the name generator decide how to handle multi-value fields (PHP uses `[]`, but other backends might differ)
  - Keeps the solution flexible and not hardcoded to any specific notation

  Can you review this approach before I propagate these changes to all the theme libraries? I'd like your approval first before going further.

  Thanks!

packages/core/src/components/Form.tsx
```diff
@@ -195,6 +196,7 @@ export interface FormProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
    * to put the second parameter before the first in its translation.
    */
   translateString?: Registry['translateString'];
+  nameGenerator?: NameGeneratorFunction;
```
heath-freenome, line 202: Please document this new prop, here as well as in the `docs` directory. Also, update the `v6x upgrade guide.md` to mention the new feature (you can point to the docs for it to make it easy)

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -597,11 +597,16 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For custom widgets with multiple=true, generate a fieldPathId with isMultiValue flag
+    const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
+
     return (
       <Widget
         id={fieldPathId.$id}
```
heath-freenome: Maybe, for consistency, change this to:
```suggestion
        id={multiValueFieldPathId.$id}
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -647,13 +653,18 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For multi-select widgets, generate a fieldPathId with isMultiValue flag
+    const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
+
     return (
       <Widget
         id={fieldPathId.$id}
```
heath-freenome: Maybe, for consistency, change this to:
```suggestion
        id={multiValueFieldPathId.$id}
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -697,11 +709,16 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For file widgets with multiple=true, generate a fieldPathId with isMultiValue flag
+    const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
+
     return (
       <Widget
         options={options}
```
heath-freenome, line 721: Ditto on the `fieldPathId` => `multiValueFieldPathId`

CHANGELOG.md
```diff
@@ -33,49 +35,65 @@ should change the heading of the (upcoming) version to include a major version b
...
 
 ## @rjsf/utils
 
 - BREAKING CHANGE: Updated `FieldTemplateProps` and `WrapIfAdditionalTemplateProps` to replace the `onKeyChange()` and `onDropPropertyClick()` callback generator props with the `onKeyRename()`, `onKeyRenameBlur()` and `onRemoveProperty()` callback props
 - BREAKING CHANGE: Updated `ObjectFieldTemplateProps` to replace the `onAddClick()` callback generator prop with the `onAddProperty()` callback prop
 - Added new hook `useDeepCompareMemo()` and its associated tests
 
+## @rjsf/utils
+
+- Added `NameGeneratorFunction` type and two built-in name generators: `bracketNameGenerator` and `dotNotationNameGenerator`
```
heath-freenome, line 89:
```suggestion
- Added new hook `useDeepCompareMemo()` and its associated tests
- Added `NameGeneratorFunction` type and two built-in name generators: `bracketNameGenerator` and `dotNotationNameGenerator`
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -597,14 +597,19 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
     } = this.props;
-    const { widgets, globalUiOptions, schemaUtils } = registry;
+
+    const { widgets, globalUiOptions, schemaUtils, globalFormOptions } = registry;
     const { widget, title: uiTitle, ...options } = getUiOptions<T[], S, F>(uiSchema, globalUiOptions);
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For custom widgets with multiple=true, generate a fieldPathId with isMultiValue flag
```
heath-freenome, line 606:
```suggestion
    const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
    // For custom widgets with multiple=true, generate a fieldPathId with isMultiValue flag
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -647,16 +653,21 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
+
+    const { widgets, schemaUtils, globalUiOptions, globalFormOptions } = registry;
     const itemsSchema = schemaUtils.retrieveSchema(schema.items as S, items);
     const enumOptions = optionsList<T[], S, F>(itemsSchema, uiSchema);
     const { widget = 'select', title: uiTitle, ...options } = getUiOptions<T[], S, F>(uiSchema, globalUiOptions);
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For multi-select widgets, generate a fieldPathId with isMultiValue flag
```
heath-freenome, line 663:
```suggestion
    const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
    // For multi-select widgets, generate a fieldPathId with isMultiValue flag
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -597,14 +597,19 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
-    const { widgets, globalUiOptions, schemaUtils } = registry;
+
+    const { widgets, globalUiOptions, schemaUtils, globalFormOptions } = registry;
     const { widget, title: uiTitle, ...options } = getUiOptions<T[], S, F>(uiSchema, globalUiOptions);
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For custom widgets with multiple=true, generate a fieldPathId with isMultiValue flag
+    const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
```
heath-freenome, line 607: To prevent rerender due to new fieldPathId that is exactly the same (but different storage), can you import the new `useDeepCompareMemo` function from `@rjsf/utils` and do this?
```suggestion
    const multiValueFieldPathId = useDeepCompareMemo(toFieldPathId('', globalFormOptions, fieldPathId, true));
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -647,16 +653,21 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
+    const { widgets, schemaUtils, globalUiOptions, globalFormOptions } = registry;
     const itemsSchema = schemaUtils.retrieveSchema(schema.items as S, items);
     const enumOptions = optionsList<T[], S, F>(itemsSchema, uiSchema);
     const { widget = 'select', title: uiTitle, ...options } = getUiOptions<T[], S, F>(uiSchema, globalUiOptions);
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For multi-select widgets, generate a fieldPathId with isMultiValue flag
+    const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
```
heath-freenome, line 664: To prevent rerender due to new fieldPathId that is exactly the same (but different storage), can you import the new `useDeepCompareMemo` function from `@rjsf/utils` and do this?
```suggestion
    const multiValueFieldPathId = useDeepCompareMemo(toFieldPathId('', globalFormOptions, fieldPathId, true));
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -697,15 +709,20 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
     } = this.props;
-    const { widgets, globalUiOptions, schemaUtils } = registry;
+
+    const { widgets, globalUiOptions, schemaUtils, globalFormOptions } = registry;
     const { widget = 'files', title: uiTitle, ...options } = getUiOptions<T[], S, F>(uiSchema, globalUiOptions);
     const Widget = getWidget<T[], S, F>(schema, widget, widgets);
     const label = uiTitle ?? schema.title ?? name;
     const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
+
+    // For file widgets with multiple=true, generate a fieldPathId with isMultiValue flag
```
heath-freenome, line 716:
```suggestion
    const displayLabel = schemaUtils.getDisplayLabel(schema, uiSchema, globalUiOptions);
    // For file widgets with multiple=true, generate a fieldPathId with isMultiValue flag
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -697,15 +709,20 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
+    const multiValueFieldPathId = toFieldPathId('', globalFormOptions, fieldPathId, true);
```
heath-freenome, line 717: To prevent rerender due to new fieldPathId that is exactly the same (but different storage), can you import the new `useDeepCompareMemo` function from `@rjsf/utils` and do this?
```suggestion
    const multiValueFieldPathId = useDeepCompareMemo(toFieldPathId('', globalFormOptions, fieldPathId, true));
```

heath-freenome (commented):
Looks good! Just a few suggested changes. I also have a [PR](https://github.com/rjsf-team/react-jsonschema-form/pull/4808) to convert the `ArrayField` class component into a set of stateless functional components. If mine merges first, you will likely have to rework your `ArrayField` changes in that file. If you can get the changes made in the next few hours, then I will merge yours and update mine instead

packages/core/src/components/fields/ArrayField.tsx
nlemoine: Not sure what I should do about this. I can't use that React hook here in the current state of my PR. 

```
React Hook "useDeepCompareMemo" cannot be called in a class component. React Hooks must be called in a React function component or a custom React Hook
```

nlemoine:
I made some of the changes you suggested but didn't include the `useDeepCompareMemo` hooks. See my comment in your review.

This is unrelated but may I ask if there's a recommended folder structure for themes?

I noticed some of them follow this structure:
```
├── components
│   ├── fields
│   ├── templates
│   │   ├── ButtonTemplates
│   │   ├── FieldTemplate
│   └── widgets
```

While others are structured like:
```
├── AddButton
├── ArrayFieldItemTemplate
├── ArrayFieldTemplate
├── BaseInputTemplate
├── CheckboxesWidget
├── CheckboxWidget
├── DescriptionField
├── ErrorList
├── FieldErrorTemplate
├── FieldHelpTemplate
├── FieldTemplate
├── Form
├── GridTemplate
├── IconButton
├── MultiSchemaFieldTemplate
├── ObjectFieldTemplate
├── OptionalDataControlsTemplate
├── RadioWidget
├── RangeWidget
├── SelectWidget
├── SubmitButton
├── Templates
├── TextareaWidget
├── Theme
├── TitleField
├── Widgets
└── WrapIfAdditionalTemplate
```

Any advice for building a custom theme?

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: Ok, I'll merge yours then update it in mine

heath-freenome:
> I made some of the changes you suggested but didn't include the `useDeepCompareMemo` hooks. See my comment in your review.
> 
> This is unrelated but may I ask if there's a recommended folder structure for themes?
> 
> I noticed some of them follow this structure:
> 
> ```
> ├── components
> │   ├── fields
> │   ├── templates
> │   │   ├── ButtonTemplates
> │   │   ├── FieldTemplate
> │   └── widgets
> ```
> 
> While others are structured like:
> 
> ```
> ├── AddButton
> ├── ArrayFieldItemTemplate
> ├── ArrayFieldTemplate
> ├── BaseInputTemplate
> ├── CheckboxesWidget
> ├── CheckboxWidget
> ├── DescriptionField
> ├── ErrorList
> ├── FieldErrorTemplate
> ├── FieldHelpTemplate
> ├── FieldTemplate
> ├── Form
> ├── GridTemplate
> ├── IconButton
> ├── MultiSchemaFieldTemplate
> ├── ObjectFieldTemplate
> ├── OptionalDataControlsTemplate
> ├── RadioWidget
> ├── RangeWidget
> ├── SelectWidget
> ├── SubmitButton
> ├── Templates
> ├── TextareaWidget
> ├── Theme
> ├── TitleField
> ├── Widgets
> └── WrapIfAdditionalTemplate
> ```
> 
> Any advice for building a custom theme?

I'd go with the second pattern... At some point I'll get around to making all the themes have the same directory structure.

heath-freenome (approved):

heath-freenome:
@nlemoine Even though I merged this, I just realized that you don't actually have any tests that verify adding a name generator does what you expect. Can you, at minimum, add some snapshot tests that show off the name generators in action?

You would simply have to update the `arrayTests.tsx`, `objectTests.tsx` and `formTests.tsx` in the `snapshot-tests` package, build it, then run through each theme and run `npm run test:update` to update the tests in each theme. Then verify the snapshots added the names as you would expect... Thanks!

nlemoine:
Hi @heath-freenome, 

Thanks for merging this! 

Sorry, I indeed forgot to add some tests. You will extensive tests in #4809

PR #4776, opened 2025-09-20
Feature - Prep work for new feature in 4736

heath-freenome:

Breaking change refactoring in preparation for the feature in 4736
- In `@rjsf/utils`:
  - Added new `GlobalFormOptions`, refactoring the `experimental_componentUpdateStrategy` from `Registry` and `idPrefix` & `idSeparator` from `FieldProps`
  - Replaced the `experimental_componentUpdateStrategy` prop in `Registry` with `readonly globalFormOptions?: GlobalFormOptions`
  - Updated `FieldProps` to remove `idPrefix` and `idSeparator`
- In `@rjsf/core`:
  - Updated `Form` to add the `globalFormOptions` to the registry if there are any `GlobalFormOptions` values provided
  - Updated `ArrayField`, `LayoutGridField`, `ObjectField` and `SchemaField` to get `idPrefix`, `idSeparator` from the `registry.globalFormOptions`, no longer passing them on `FieldProps`
  - Updated `SchemaField` to get `experimental_componentUpdateStrategy` from the `registry.globalFormOptions` as well
  - Updated the `SchemaField` tests as needed
- In `@rjsf/daisy` updated the snapshots
- Updated the `custom-widget-fields.md` and `v6.x upgrade guide.md` to document the refactor of the `idPrefix` and `idSeparator` refactor
- Updated the `CHANGELOG.md` file accordingly

### Checklist


packages/utils/src/types.ts
```diff
@@ -395,6 +395,22 @@ export type GlobalUISchemaOptions = GenericObjectType & {
   enableMarkdownInDescription?: boolean;
 };
 
+/** The set of options from the `Form` that will be available on the `Registry` for use in everywhere the `registry` is
+ * available.
+ */
+export type GlobalFormOptions = GenericObjectType & {
```
nickgros: I don't think it makes sense to make this extend GenericObjectType, it's a known set of properties that the end user will not augment.

```suggestion
export type GlobalFormOptions = {
```

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -511,13 +511,12 @@ class ArrayField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
       onBlur,
       onFocus,
-      idPrefix,
-      idSeparator = '_',
       rawErrors,
     } = this.props;
     const { keyedFormData } = this.state;
     const fieldTitle = schema.title || title || name;
-    const { schemaUtils, formContext } = registry;
+    const { schemaUtils, formContext, globalFormOptions = {} } = registry;
```
nickgros: It doesn't look like it would be a problem in any of the components in this change, but setting the default value to `{}` here would create a new object every render. It could cause performance issues if it later got passed to a hook dependency array, which could happen in a future change or in a custom field.

Maybe `Form.getDefaultRegistry` should create one empty object instead and pass it around everywhere?

packages/core/src/components/Form.tsx
```diff
@@ -956,10 +956,15 @@ export default class Form<
...
-      experimental_componentUpdateStrategy = 'customDeep',
+      experimental_componentUpdateStrategy,
+      idSeparator,
+      idPrefix,
     } = this.props;
     const { schema, schemaUtils } = this.state;
     const { fields, templates, widgets, formContext, translateString } = getDefaultRegistry<T, S, F>();
+    const globalFormOptions = { idPrefix, idSeparator, experimental_componentUpdateStrategy };
+    // Now stringify it so that we can determine whether `globalFormOptions` actually contains data
+    const hasEmptyGlobalFormOptions = JSON.stringify(globalFormOptions) === '{}';
```
nickgros: Use lodash `isEmpty` instead?

packages/docs/docs/api-reference/form-props.md
```diff
@@ -56,6 +56,18 @@ Formerly the `validate` prop.
 The `customValidate` prop requires a function that specifies custom validation rules for the form.
 See [Validation](../usage/validation.md) for more information.
 
+## experimental_componentUpdateStrategy
```
nickgros, line 59: Thanks for adding this

packages/docs/docs/advanced-customization/custom-widgets-fields.md
```diff
@@ -397,8 +397,6 @@ A field component will always be passed the following props:
 - `readonly`: A boolean value stating if the field is read-only;
 - `autofocus`: A boolean value stating if the field should autofocus;
 - `name`: The unique name of the field, usually derived from the name of the property in the JSONSchema
-- `idPrefix`: To avoid collisions with existing ids in the DOM, it is possible to change the prefix used for ids; Default is `root`
-- `idSeparator`: To avoid using a path separator that is present in field names, it is possible to change the separator used for ids (Default is `_`)
```
nickgros, line 401: Could move this to `## The registry object` (the section just below this one)

packages/utils/src/types.ts
```diff
@@ -423,8 +439,8 @@ export interface Registry<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
   translateString: (stringKey: TranslatableString, params?: string[]) => string;
   /** The optional global UI Options that are available for all templates, fields and widgets to access */
   globalUiOptions?: GlobalUISchemaOptions;
-  /** The component update strategy used by the Form and its fields for performance optimization */
-  experimental_componentUpdateStrategy?: 'customDeep' | 'shallow' | 'always';
+  /** The optional global Form Options that are available for all templates, fields and widgets to access */
+  readonly globalFormOptions?: GlobalFormOptions;
```
nickgros: IMO if the Form always creates this (as I suggested in another comment), then this should not be optional. But ok for all of the properties inside it to be optional

```suggestion
  /** The optional global Form Options that are available for all templates, fields and widgets to access */
  readonly globalFormOptions: GlobalFormOptions;
```

packages/utils/src/types.ts
heath-freenome: You are right. That was copy-pasta from the `GlobalUIOptions`

packages/core/src/components/Form.tsx
heath-freenome: I tried and it returns false because there are keys/value pairs, even though the values are undefined. Not that it matters since I went with your suggestion.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: I'll go with that

packages/docs/docs/advanced-customization/custom-widgets-fields.md
heath-freenome: Good point

packages/core/src/components/Form.tsx
```diff
@@ -956,10 +956,16 @@ export default class Form<
...
+      experimental_componentUpdateStrategy,
+      idSeparator,
+      idPrefix,
     } = this.props;
     const { schema, schemaUtils } = this.state;
     const { fields, templates, widgets, formContext, translateString } = getDefaultRegistry<T, S, F>();
+    // Use JSON.parse(), of the `JSON.stringify`'d version of the options to filter out all undefined ones
+    const globalFormOptions = JSON.parse(
+      JSON.stringify({ idPrefix, idSeparator, experimental_componentUpdateStrategy }),
+    );
```
nickgros: This will not work if #4773 adds the `nameGenerator` function prop since it couldn't be be stringified/parsed. 

Could you use lodash `omitBy`/`isUndefined`?

```suggestion
    const globalFormOptions = omitBy({ idPrefix, idSeparator, experimental_componentUpdateStrategy }, isUndefined)
```

heath-freenome: Let me try that!

heath-freenome: I went with `isNil()` rather than `isUndefined()` because we were already including it

nickgros (approved):

Issue #512, opened 2017-03-14
liveValidate should optionally not validate "pristine" fields (at least until a submit)


glasserc:
### Prerequisites

- [x] I have read the [documentation](https://github.com/mozilla-services/react-jsonschema-form/blob/master/README.md#readme);
- [x] In the case of a bug report, I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.

### Description

In the playground, the "Validation" tab, when live validation is enabled, starts with an error ("passwords don't match"). However, the user hasn't put in a password yet, so it doesn't make sense as part of the user experience to show this error, at least until they start typing in the field.

Similarly, in the "Errors" tab, the initial state shows a bunch of errors. However, for some user experiences, this doesn't seem necessary or helpful; we're telling the user about errors that pre-existed, not errors they themselves have made.

Per https://github.com/mozilla-services/react-jsonschema-form/issues/246#issuecomment-286277405, maybe we should not validate fields until they've been "touched" somehow.

n1k0:
> In the playground, the "Validation" tab, when live validation is enabled, starts with an error ("passwords don't match")

I can't reproduce this behavior with the latest version. Maybe I'm missing a step? If not, how has this been fixed, that would been nice to know.

n1k0:
> Similarly, in the "Errors" tab, the initial state shows a bunch of errors

This one is reproducible, though in this case we provide actual `formData`, so the form immediately immediately renders errors matching the provided data. This makes sense to me, but maybe just me? If not, would this need to be documented somehow?

glasserc:
Ha ha, so apparently my browser had saved a password for localhost:8080 from when I was working on something two jobs ago, and it autofilled that password when I was hacking on rjsf. But obviously it didn't autofill the second one, so "passwords don't match". Never mind...

tuunanen:
The issue still surfaces for example if any of the fields in the "Validation" example are marked as required.

[...]

PR #4815, opened 2025-10-24
Fix 391 & 2099 with new initialFormData prop and setFieldValue() on Form

### LINKED ISSUE #391: Form is reset if setState is called in a Component
Documentation read, yes!

### Description

It looks the form is (wrongly) reset if some internal state changes within a component. Consider the following example https://jsfiddle.net/fab1en/954uLrrL/1/

1. Edit the form
2. Hit the second button which update some foo internal state
3. Notice how the original value is reset in the form.

Form should not be reset like that. Am I missing something?

### LINKED ISSUE #2099: Implementing setFieldValue [New Feature]
### Prerequisites

### Description

Hi folks, I didn't find ability to setup formData for multiple field. As I understand there is only ability to set current field. I would like to change formData with additional function which will take newFormData and path where I need to change data.
For example, I have object with fields:
```
{
  symptom: {
    title: "Condition",
    type: "object",
    required: [:id, :name],
    properties: {
      id: {
        type: "number",
        title: 'id'
      },
      name: {
        type: "string",
        title: 'name'
      }
    }
  }
}
```
And I've created my own widget, with will render select where I choose symptom, but value already includes id and name and I would like update `symptom` key with data `{ id: 1, name: 'Name' }`
```
const SymptomWidget = props => {
  const onChangeSymptom = symptom => {
    const parsedSymptom = symptom ? JSON.parse(symptom) : null
    const id = pathOr('', ['0'], parsedSymptom)
    const name = pathOr('', ['1'], parsedSymptom)
    props.onChange({ id, name })
  }
  return <ConditionSelect onChange={onChangeSymptom} />
}
```
But I get this result: `{ symptom: id: { id: 1, name: 'Name' } }`.

### Version

2.4

heath-freenome:

Fixes #391 by adding support for the `initialFormData` prop on `Form`
Fixed #2099 by adding `setFieldValue(fieldPath: string | FieldPathList, newValue?: T): void` on `Form`
- Updated `@rjsf/core` as follows:
  - Updated `FormProps` to add the new `initialFormData` prop
  - Updated `Form` so that is behaves as a "controlled" form when `formData` is passed and uncontrolled when `initialFormData` is passed
  - Added implementation of `setFieldValue(fieldPath: string | FieldPathList, newValue?: T): void` on `Form`
  - Also fixed an issue where live validation was called on the initial form render, causing errors to show immediately, partially fixing #512
  - Updated the tests for `Form`, `ArrayField`, `ObjectField` and `StringField` to deal with the change to not live validating on initial render
  - Added new tests for `Form` that verify the fixes around controlled vs uncontrolled form related to the `initialFormData` and how it affects `reset()`
  - Added new tests for `Form` that verify the new ``setFieldValue()` programmatic function
- Updated `internal.md`, `form-props.md` and `v6x upgrade guide.md` to document the new `initialFormData` prop, `setFieldValue()` function  and the other potentially breaking `Form` fix

### Checklist

### DIFF
--- a/packages/core/test/ArrayField.test.jsx
+++ b/packages/core/test/ArrayField.test.jsx
@@ -380,7 +380,7 @@ describe('ArrayField', () => {
         liveValidate: true,
       });
 
-      // trigger the errors by submitting the form since initial render no longer shows them
+      // trigger the errors by submitting the form
       submitForm(node);
 
       const matches = node.querySelectorAll('#custom');


packages/core/test/Form.test.jsx
```diff
@@ -394,10 +394,24 @@ describeRepeated('Form common', (createFormComponent) => {
     });
 
     it('should pass errors as the provided React component', () => {
+      // live validate does not run on initial render anymore
```
heath-freenome, line 397: @nickgros I initially updated the tests to verify no errors and then seeing them show up after live validate. For the other test files I just went with submitting the form to force the errors to show up. Your thoughts on whether I should also do that in this file?

packages/core/test/Form.test.jsx
```diff
@@ -5063,6 +5154,30 @@ describe('Calling reset from ref object', () => {
     expect(input.getAttribute('value')).to.eq('');
     expect(checkbox.checked).to.eq(true);
   });
+
+  it('Reset button test with initial value', () => {
```
heath-freenome:
```suggestion
  it('Reset button test with initialFormData', () => {
```

packages/core/test/Form.test.jsx
```diff
@@ -5713,3 +5828,74 @@ describe('nameGenerator', () => {
...
+        initialFormData={initialFormData}
+        schema={schema}
+        validator={validator}
+        onSubmit={handleSubmit}
+        disabled={isPending}
+        {...props}
+      />
+    );
+  };
+  it('Form reset without initial data and not being controlled', () => {
```
heath-freenome:
```suggestion
  it('show that Form resets without initial data when it is controlled', () => {
```

packages/core/test/Form.test.jsx
```diff
@@ -5713,3 +5828,74 @@ describe('nameGenerator', () => {
...
+
+    const submit = container.querySelector('button');
+    act(() => {
+      fireEvent.click(submit);
+    });
+
+    input = container.querySelector('input');
+    expect(input.value).to.eql(data.name);
+  });
+  it('Form does not reset with initial data and not being controlled', () => {
```
heath-freenome:
```suggestion
  it('show that Form does not reset with initialFormData when it is uncontrolled', () => {
```

packages/core/test/Form.test.jsx
```diff
@@ -5713,3 +5828,74 @@ describe('nameGenerator', () => {
...
+
+describe('initialFormData feature to prevent form reset', () => {
+  const schema = {
+    title: 'Reset Example',
+    properties: {
+      name: { type: 'string', title: 'Name' },
+    },
+  };
+  const data = { name: 'initial_id' };
+  const FormWrapper = ({ formData, initialFormData }) => {
```
heath-freenome, line 6012:
```suggestion
  /** This was adapted from the [example](https://playcode.io/2038613) provided in issue #391
   */
  const FormWrapper = ({ formData, initialFormData }) => {
```

packages/docs/docs/migration-guides/v6.x upgrade guide.md
```diff
@@ -961,6 +964,13 @@ Three new validator-based utility functions are available in `@rjsf/utils`:
   - This new optional flag was added to the `SchemaUtilsType` interface's version of `retrieveSchema()` as well.
 - `validationDataMerge()`: Added optional `preventDuplicates` boolean flag that causes the `mergeObjects()` call to receive `preventDuplicates` instead of `true`
 
+### initialFormData prop added to Form
+
+A new `initialFormData` prop was added to `Form` to support adding initial values for an **"uncontrolled"** React component.
```
heath-freenome:
```suggestion
A new `initialFormData` prop was added to `Form` to support providing initial values for an **"uncontrolled"** React component.
```

packages/core/src/components/Form.tsx
```diff
@@ -799,6 +818,18 @@ export default class Form<
...
+  /** Allows a user to set a value for the provided `fieldPath`, which must be either a dotted path to the field OR a
+   * `FieldPathList`. To set the root element, used either `''` or `[]` for the path. Passing undefined will clear the
+   * value in the field.
+   *
+   * @param fieldPath - Either a dotted path to the field or the `FieldPathList` to the field
+   * @param [newValue] - The new value for the field
+   */
+  setFieldValue = (fieldPath: string | FieldPathList, newValue?: T) => {
+    const path = Array.isArray(fieldPath) ? fieldPath : fieldPath.split('.');
+    this.onChange(newValue, path);
```
heath-freenome: @nickgros I could also add the id on the the `onChange()` call by using `toFieldPathId()` and then passing `$id`... Thoughts about this?
```suggestion
    const fieldPathId = toFieldPathId('', registry.globalFormOptions, path);
    this.onChange(newValue, path, undefined, fieldPathId[ID_KEY]);
```

packages/core/test/ArrayField.test.jsx
```diff
@@ -380,6 +380,9 @@ describe('ArrayField', () => {
         liveValidate: true,
       });
 
+      // trigger the errors by submitting the form since initial render no longer shows them
```
nickgros: For long-term maintenance, I don't think we care about what the behavior used to be, so I would remove that contextual information from the code comments
```suggestion
      // trigger the errors by submitting the form
```

packages/core/test/Form.test.jsx
nickgros: I think we only need one test where we verify `liveValidate` does not have initial errors on render. No need to capture that assertion in the rest (otherwise it is a redundant test IMO). But you can keep it for the tests where you already added it.

PR #4817, opened 2025-10-27
[feat] Add fallback UI for unsupported types

### LINKED ISSUE #4736: `FallbackField` feature proposal for schemas with no `type`
### Prerequisites

### What theme are you using?

core

In our application, we have (user-created) schemas that look something like this:

```json
{
  "properties": {
    "foo": {
		"title": "Foo",
		"description": "TBD"
    }
  },
  "required": ["foo"]
}
```

You may notice that `foo` does not have a `type`! Apparently, this is valid JSON Schema. When you plug this into RJSF today, we render the `UnsupportedField` template, which is basically an error message.

Image

I was thinking of implementing a new (opt-in) "FallbackField", that can be rendered if the schema has no type (or an unknown type). Similar to the `MultiSchemaField`, the `FallbackField` will create two components:
1. A `SelectWidget` that can be used to pick one of `["string", "number", "boolean"]`
2. A `SchemaField` for the value, which uses the correct widget based on the chosen type.

Each theme would implement a `FallbackFieldTemplate`, which will basically lay out the components like this:

```
Foo
  Type                    Value
  +-------------------+   +--------------------------+
  | string          v |   | your text here           |
  +--------------------   +--------------------------+
```

A user could change "string" to "boolean" to see a checkbox and interact with it. (Note: If form data is present, I think it makes the most sense to just clear the form data when the user changes the type)

```
Foo
  Type                    Value
  +-------------------+   +--+
  | boolean         v |   |✅|
  +--------------------   +--+
```

I was thinking of making this opt-in because I think most RJSF users _should_ define a type in for each of their schema properties, so the default behavior to show the `UnsupportedField` won't change. 

In `@rjsf/core`, we can expose the default `_FallbackField` in the registry. To use it, one would just have to provide it (or a custom implementation) to the `FallbackField` property:

```tsx
import Form, { getDefaultRegistry } from '@rjsf/core'
  
const schema: RJSFSchema = {  
  title: 'My form',
  description: 'My description',
  // No type!
};  
  
render(
	<Form
      schema={schema}
      validator={validator}
      fields={{ FallbackField: getDefaultRegistry().fields._FallbackField }}
	/>,
	document.getElementById('app')
);
```

_No response_

_No response_

nickgros:
Resolves #4736

- Add new opt-in feature to use a new feature to display a type selector for fields with unknown types
- Feature is implemented by new FallbackField and rendered by new FallbackFieldTemplate.
- Update SchemaField to use FallbackField
- When not opted in, FallbackField uses existing UnsupportedField component

### Checklist

### DIFF
--- a/packages/docs/docs/migration-guides/v6.x upgrade guide.md
+++ b/packages/docs/docs/migration-guides/v6.x upgrade guide.md
@@ -82,11 +82,14 @@ React 18 is officially supported on all the themes.
 
 React 19 support is expected before the end of beta (although several developers have already upgraded with no problems).
 
-### Form POTENTIAL BREAKING CHANGE
+### Form POTENTIAL BREAKING CHANGES
 
 The `onChange()` and `onSubmit()` callbacks were accidentally returning extra data along with what was listed in the `IChangeEvent` interface.
 In v6, this was fixed so that only the data described by the `IChangeEvent` interface is returned.
 If you were relying on that extra data in the callbacks, we are sorry for the inconvenience.
+
+A bug was fixed such that live validation will no longer be triggered on the data provided during the initial rendering of the `Form`.
+If you require that your initial data have errors shown, then you can [programmatically validate](../usage/validation/#validate-form-programmatically) it after the initial render.
 
 ### IdSchema replacement BREAKING CHANGE
 
@@ -969,6 +972,19 @@ Three new validator-based utility functions are available in `@rjsf/utils`:
 - `retrieveSchema()`: Added an optional `resolveAnyOfOrOneOfRefs` boolean flag that causes the internal `resolveAllSchemas()` to resolve `$ref`s inside of the options of `anyOf`/`oneOf` schemas
   - This new optional flag was added to the `SchemaUtilsType` interface's version of `retrieveSchema()` as well.
 - `validationDataMerge()`: Added optional `preventDuplicates` boolean flag that causes the `mergeObjects()` call to receive `preventDuplicates` instead of `true`
+
+### initialFormData prop added to Form
+
+A new `initialFormData` prop was added to `Form` to support providing initial values for an **"uncontrolled"** React component.
+As a result, a small change was made so that `formData` now causes the `Form` to behave as a **"controlled"** React component.
+
+The programmatic `reset()` will ensure that an **"uncontrolled"** form is reset back to the `initialFormData` values when triggered.
+
+### setFieldValue() added to Form
+
+A new programmatic function, `setFieldValue(fieldPath: string | FieldPathList, newValue?: T): void` was added to the `Form`.
+It allows a user to set a value for the provided `fieldPath`, which must be either a dotted path to the field OR a `FieldPathList`.
+To set the root element, used either `''` or `[]` for the path. Passing undefined will clear the value in the field.
 
 ### liveValidate and liveOmit performance improvement
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,20 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Updated `FormProps` to add new `onChange`/`onBlur` values for the `liveValidate` and `liveOmit` props, deprecating the `boolean` aspect of them
 - Updated `Form` to support the new feature to do `onBlur` handling of `liveValidate` and `liveOmit`
+- Updated `FormProps` to add the new `initialFormData` prop
+- Updated `Form` so that is behaves as a "controlled" form when `formData` is passed and uncontrolled when `initialFormData` is passed, fixing [#391](https://github.com/rjsf-team/react-jsonschema-form/issues/391)
+  - Also fixed an issue where live validation was called on the initial form render, causing errors to show immediately, partially fixing [#512](https://github.com/rjsf-team/react-jsonschema-form/issues/512)
+- Updated `Form` to add a new programmatic function, `setFieldValue(fieldPath: string | FieldPathList, newValue?: T): void`, fixing [#2099](https://github.com/rjsf-team/react-jsonschema-form/issues/2099)
 - Added new `FallbackField` to add opt-in functionality to control form data that is of an unsupported or unknown type ([#4736](https://github.com/rjsf-team/react-jsonschema-form/issues/4736)).
 
+## @rjsf/mantine
+
+- Updated `FieldHelpTemplate` to avoid issue when `help` `and `fieldPathId` are undefined
+
 ## Dev / docs / playground
+
 - Updated the playground to switch `liveValidate` and `liveOmit` from checkboxes to radio buttons for the new options
-- Updated `form-props.md` and `v6x upgrade guide.md` to document the new features and deprecation
+- Updated `internals.md`, `form-props.md` and `v6x upgrade guide.md` to document the new features, potential breaking changes and deprecations
 
 # 6.0.0-beta.22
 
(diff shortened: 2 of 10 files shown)


packages/snapshot-tests/package.json
```diff
@@ -55,20 +55,5 @@
   "homepage": "https://github.com/rjsf-team/react-jsonschema-form",
   "publishConfig": {
     "access": "public"
-  },
```
nickgros, line 58: This didn't quite fix the dependency cycle issue, added a better fix to @rjsf/core

packages/core/package.json
```diff
@@ -112,5 +112,21 @@
...
+            "//": "Break the dependency cycle between @rjsf/core and @rjsf/snapshot-tests by explicitly listing the dependencies needed for build",
+            "projects": [
+              "@rjsf/utils",
+              "@rjsf/validator-ajv8"
+            ],
+            "target": "build"
+          }
+        ]
+      }
+    }
```
nickgros, line 130: Last attempt did not quite fix the dependency cycle issue. Explicitly enumerating the two dependencies of `core` seems to do the trick (graph generated using `nx graph`)

Image

packages/core/src/components/fields/FallbackField.tsx
```diff
@@ -0,0 +1,164 @@
...
+    default:
+      return formData;
+  }
+}
+
+/**
+ * The `FallbackField` component is used to render a field for unsupported or unknown schema types. If
+ * `useFallbackUiForUnsupportedType` is enabled in the `globalUiOptions`, it provides a type selector
+ * @param props
+ * @constructor
```
heath-freenome: You can probably remove these

packages/utils/src/enums.ts
```diff
@@ -86,4 +86,8 @@ export enum TranslatableString {
    * provided by FileWidget
    */
   FilesInfo = '**%1** (%2, %3 bytes)',
+  /** Label for the schema type selector, used by FallbackField */
+  Type = 'Type',
+  /** Label for the 'value' field, used by FallbackField */
+  Value = 'Value',
```
heath-freenome: Consider moving these after `OptionalObjectEmptyMsg` but before the comment that says: `Strings with replaceable parameters`

packages/core/src/components/fields/FallbackField.tsx
```diff
@@ -0,0 +1,164 @@
...
+    onChange,
+    errorSchema,
+  } = props;
+  const { translateString, fields, globalFormOptions } = registry;
+  const [type, setType] = useState<JSONSchema7TypeName>(getTypeOfFormData(formData));
+
+  const uiOptions = getUiOptions<T, S, F>(uiSchema);
+
+  const typeSelectorInnerFieldPathId = useDeepCompareMemo<FieldPathId>(
+    toFieldPathId('__internal_type_selector', globalFormOptions, fieldPathId.path),
```
heath-freenome: You could also use `fieldPathId` itself
```suggestion
    toFieldPathId('__internal_type_selector', globalFormOptions, fieldPathId),
```

packages/core/src/components/fields/SchemaField.tsx
```diff
@@ -80,24 +76,7 @@ function getFieldComponent<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
...
-        return (
-          <UnsupportedFieldTemplate
-            schema={schema}
-            fieldPathId={fieldPathId}
-            reason={translateString(TranslatableString.UnknownFieldType, [String(schema.type)])}
-            registry={registry}
-          />
-        );
-      };
+  return componentName in fields ? fields[componentName] : fields['FallbackField'];
```
heath-freenome, line 79: I believe you also want to replace the `UnsupportedField` in `ArrayField` with this, right?

packages/docs/docs/migration-guides/v6.x upgrade guide.md
```diff
@@ -663,6 +663,12 @@ This template was updated to support the new [Optional Data Controls](#optional-
 
 If you have your own implementation of `TitleFieldTemplate` and want to support this new feature, please refer to the `TitleFieldTemplate` implementation for whatever theme you are using and pick up the changes.
 
+#### SchemaField - Addition of FallbackField POTENTIAL BREAKING CHANGE
+
+The SchemaField was updated to use the new FallbackField, which adds opt-in support to replace the UnsupportedField with an interactive UI instead of an error message. If you have your own implementation of `SchemaField` and want to support this new feature, please refer to the `SchemaField` implementation in `@rjsf/core` and pick up the changes.
+
+The UnsupportedField component is unaffected by these changes, and is still used by the new FallbackField.
```
heath-freenome:
```suggestion
The `SchemaField` was updated to use the new `FallbackField`, which adds opt-in support to replace the `UnsupportedField` with an interactive UI instead of an error message.
If you have your own implementation of `SchemaField` and want to support this new feature, please refer to the `SchemaField` implementation in `@rjsf/core` and pick up the changes.

The `UnsupportedField` component is unaffected by these changes, and is still used by the new `FallbackField`.
```

packages/utils/src/types.ts
```diff
@@ -569,6 +576,27 @@ export type FieldTemplateProps<
...
+/**
+ * The properties that are passed to a `FallbackFieldTemplate` implementation
+ */
+export type FallbackFieldTemplateProps<
+  T = any,
+  S extends StrictRJSFSchema = RJSFSchema,
+  F extends FormContextType = any,
+> = RJSFBaseProps<T, S, F> & {
+  typeSelector: ReactNode;
+  schemaField: ReactNode;
```
heath-freenome, line 599: Consider adding documentation for the two props. (You can copy from the docs).

packages/utils/src/types.ts
```diff
@@ -569,6 +576,27 @@ export type FieldTemplateProps<
   onRemoveProperty: () => void;
 };
 
+/**
+ * The properties that are passed to a `FallbackField` implementation
+ */
+export type FallbackFieldProps<
```
heath-freenome, line 582: Do we need a separate type for this? Or can `FallbackField` just use `FieldProps` like all of the other field implementations do?

packages/docs/docs/migration-guides/v6.x upgrade guide.md
```diff
@@ -920,6 +926,8 @@ The following new types were added to `@rjsf/utils`:
 
 - `ArrayFieldItemTemplateProps`: The properties of each element in the ArrayFieldTemplateProps.items array.
 - `ContainerFieldTemplateProps`: The common properties of the two container templates: `ArrayFieldTemplateProps` and `ObjectFieldTemplateProps`
+- `FallbackFieldProps`: The properties passed to the new `FallbackField` component
```
heath-freenome, line 933: Depending on your resolution to another comment of mine, this may go away

packages/core/src/components/fields/FallbackField.tsx
```diff
@@ -0,0 +1,164 @@
...
+ * The `FallbackField` component is used to render a field for unsupported or unknown schema types. If
+ * `useFallbackUiForUnsupportedType` is enabled in the `globalUiOptions`, it provides a type selector
+ * @param props
+ * @constructor
+ */
+export default function FallbackField<
+  T = any,
+  S extends StrictRJSFSchema = RJSFSchema,
+  F extends FormContextType = any,
+>(props: FallbackFieldProps<T, S, F>) {
```
heath-freenome, line 71: Consider simply using `FieldProps` like the rest of the field implementations?

packages/core/src/components/fields/FallbackField.tsx
```diff
@@ -0,0 +1,164 @@
...
+    <FallbackFieldTemplate
+      schema={schema}
+      registry={registry}
+      typeSelector={
+        <SchemaField
+          key={formData ? hashObject(formData) : '__empty__'}
+          id={`${id}-type-widget`}
+          fieldPathId={typeSelectorInnerFieldPathId}
+          idSchema={idSchema}
+          name={`${name}__fallback_type`}
```
heath-freenome: I don't think you need the `id` prop since, looking at `SchemaField` it does not use it.  Moreover, you should not need the `idSchema` prop as we no longer support it, right?

packages/core/src/components/fields/FallbackField.tsx
```diff
@@ -0,0 +1,164 @@
...
+          schema={typesOptionSchema as S}
+          formData={type as T}
+          onChange={(newType) => {
+            if (newType != null) {
+              onTypeChange(newType as JSONSchema7TypeName);
+            }
+          }}
```
heath-freenome: Consider adding a useCallback for this?

packages/utils/src/types.ts
nickgros: Not with this current set of changes, but we might need it if we also want to use it in `ArrayField` to pass a different message along.

heath-freenome: Will you add this to `ArrayField` in this PR or will you wait for another one?

heath-freenome (approved):

PR #4871, opened 2025-11-14
Fix issue with omitted extension for antd icons

selaux:

With `"type": "module"` in package.json most bundlers (e.g. webpack 5 and vite) expect an extension for referencing files directly inside `node_modules`. I guess the safest way is to just import them directly from the package.

For us e.g. Webpack 5 fails with:

```
BREAKING CHANGE: The request '@ant-design/icons/PlusCircleOutlined' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

And vite fails with:

```
Error: Cannot find module '/home/stefan/workspace/stracciatella-toolset/node_modules/@ant-design/icons/ExclamationCircleOutlined' imported from /home/stefan/workspace/stracciatella-toolset/node_modules/@rjsf/antd/lib/templates/ErrorList/index.js
Did you mean to import "@ant-design/icons/ExclamationCircleOutlined.js"?
```

### Checklist

### DIFF
--- a/packages/antd/README.md
+++ b/packages/antd/README.md
@@ -61,7 +61,7 @@ Ant Design theme, fields and widgets for `react-jsonschema-form`.
 ### Prerequisites
 
 - `antd >= 5`
-- `@ant-design/icons >= 5`
+- `@ant-design/icons >= 6`
 - `dayjs >= 1.8.0`
 - `@rjsf/core >= 6`
 - `@rjsf/utils >= 6`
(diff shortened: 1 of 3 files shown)


heath-freenome:
@selaux Thanks for trying to fix this on your own. While your fix actually solves the problem, it pulls in ALL of the icons into the system, which you probably didn't need. Instead I have pushed up a build-time fix (#4872) to add the `.js` suffixes into the generated distribution. That should fix it without any code changes.

selaux:
Works for me as well, thanks :slightly_smiling_face:

PR #4899, opened 2025-12-15
Extended customValidate function with errorSchema

### LINKED ISSUE #4898: Extend customValidate with errorSchema
### Prerequisites

### What theme are you using?

core

I have a specific use case where object field is represented by table where columns maps to object properties and rows are representing objects that user can select. Data for that table comes from external source so not always is complete, while all the properties are required.
When validation kicks in each property is validated and when one (or more) is missing for selected row submit is prevented, but there is no way to indicate the validation error on object level. I want to be able to add an error to object which will say something like that: "Not all required properties are filled in. please contact support to fix the data".
My idea to solve it was to write `customValidate` function that will check for that specific object field if any of its properties has an error, if so I want to add that additional error for whole object. I've read in the docs that `customValidate` is called after avj validation so I thought I will find all avj errors in `errors` param passed to that function, but I've checked code and it is not the case.

Solution that I'd like to propose is to extend `CustomValidator` type with errorSchema:
```ts
/** A `CustomValidator` function takes in a `formData`, `errors`, `uiSchema` and `errorSchema` objects and returns the given `errors`
 * object back, while potentially adding additional messages to the `errors`
 */
export type CustomValidator<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any> = (
  formData: T | undefined,
  errors: FormValidation<T>,
  uiSchema?: UiSchema<T, S, F>,
  errorSchema?: ErrorSchema<T>,
) => FormValidation<T>;
```

Then we can modify `processRawValidationErrors` to pass that errorSchema to `customValidate`.

_No response_

Stusu:
Updated the customValidate function signature and its usage to include errorSchema as the fourth argument. Documentation and type definitions were updated to reflect this change, allowing custom validation functions to access errors raised by JSON schema validation.

Fixes #4898 

### Checklist


packages/validator-ajv8/src/processRawValidationErrors.ts
```diff
@@ -185,7 +185,7 @@ export default function processRawValidationErrors<
   // Include form data with undefined values, which is required for custom validation.
   const newFormData = getDefaultFormState<T, S, F>(validator, schema, formData, schema, true) as T;
 
-  const errorHandler = customValidate(newFormData, createErrorHandler<T>(newFormData), uiSchema);
+  const errorHandler = customValidate(newFormData, createErrorHandler<T>(newFormData), uiSchema, errorSchema);
```
heath-freenome, line 188: Can you validate this change with a test so that coverage is satisfied?

Stusu:
I've updated changelog and added tests

PR #4912, opened 2026-01-04
fix: add React 19 compatibility for getWidget (defaultOptions + remov…

### LINKED ISSUE #4907: Fix getWidgets to stop using deprecated `defaultProps` for options and react-is.isForwardRef()
### Prerequisites

### What theme are you using?

utils

### Version

6.x

In order to support React 19 properly, we will remove the `mergeWidgetOptions()` function in `@rjsf/utils` `getWidget.ts` and the `MergedWidget` concept.

Since it is using `defaultProps` to merge the `options` together, this will be a breaking change and thus will have to be done in v7.

Make sure to update the migration guide to instruct people how to update their widgets to create the same behavior

Also, React 19 doesn't support `react-is.isForwardRef()` function either, so this will have to be removed as well.

Ashutoshx7:

This PR adds React 19 compatibility to [getWidget.tsx](cci:7://file:///d:/react-jsonschema-form/packages/utils/src/getWidget.tsx:0:0-0:0) by addressing two deprecated patterns:

1. **`defaultProps.options`**: React 19 deprecated `defaultProps` for function components. This PR adds support for a new static `defaultOptions` property while maintaining backwards compatibility with the legacy pattern.

2. **`react-is.isForwardRef()`**: React 19 removed this function. This PR replaces it with a more general component detection check using `$$typeof`.

fixes #4907

### Changes Made

- **[packages/utils/src/getWidget.tsx](cci:7://file:///d:/react-jsonschema-form/packages/utils/src/getWidget.tsx:0:0-0:0)**: 
  - Added support for new `defaultOptions` static property (React 19 compatible)
  - Maintained backwards compatibility with `defaultProps.options` (deprecated but still works)
  - Replaced `ReactIs.isForwardRef()` with general `$$typeof` check for exotic components

- **[packages/utils/test/getWidget.test.tsx](cci:7://file:///d:/react-jsonschema-form/packages/utils/test/getWidget.test.tsx:0:0-0:0)**: Updated tests to use the new `defaultOptions` pattern

- **[CHANGELOG.md](cci:7://file:///d:/react-jsonschema-form/CHANGELOG.md:0:0-0:0)**: Added entry under `@rjsf/utils`

- **[packages/docs/docs/advanced-customization/custom-widgets-fields.md](cci:7://file:///d:/react-jsonschema-form/packages/docs/docs/advanced-customization/custom-widgets-fields.md:0:0-0:0)**: Updated documentation to show new pattern

- **[packages/docs/docs/migration-guides/v7.x upgrade guide.md](cci:7://file:///d:/react-jsonschema-form/packages/docs/docs/migration-guides/v7.x%20upgrade%20guide.md:0:0-0:0)**: Created migration guide for the deprecation

### Checklist

  - [x] I've updated the changelog with a description of the PR


packages/utils/src/getWidget.tsx
```diff
@@ -101,10 +112,15 @@ export default function getWidget<T = any, S extends StrictRJSFSchema = RJSFSche
...
+  // - Regular function components (typeof === 'function')
+  // - Memoized components (ReactIs.isMemo)
+  // - ForwardRef and other exotic components (typeof === 'object' with $$typeof)
+  // Note: ReactIs.isForwardRef() was removed in React 19, so we use a more general check
   if (
     typeof widget === 'function' ||
-    (widget && ReactIs.isForwardRef(createElement(widget))) ||
-    ReactIs.isMemo(widget)
+    ReactIs.isMemo(widget) ||
+    (typeof widget === 'object' && widget !== null && '$$typeof' in widget)
```
heath-freenome: Isn't there another `react-is` API that you could use? Like perhaps `isElement()` or `isValidElementType()`?

Ashutoshx7: Let me know if i  my understanding is correct  or not -
My current approach was creating an element and then inspecting it
but the question is whether the value is renderable in the first place. For that, you are referring  isValidElementType (and isElement ) should be used instead of creating elements.     

i'll update the logic

packages/docs/docs/advanced-customization/custom-widgets-fields.md
```diff
@@ -266,10 +266,9 @@ function MyCustomWidget(props: WidgetProps) {
   return <input style={{ color, backgroundColor }} />;
 }
 
-MyCustomWidget.defaultProps = {
-  options: {
-    color: 'red',
-  },
+// React 19+: Use defaultOptions instead of defaultProps.options
+(MyCustomWidget as any).defaultOptions = {
```
heath-freenome: I'm not a big fan of casting to `any`... I wonder if it makes sense to update the `Widget` type to add an optional `defaultOptions?: GenericObjectType` prop onto it? I'm not quite sure whether that will work though.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -0,0 +1,73 @@
...
+import { WidgetProps } from '@rjsf/utils';
+
+function MyCustomWidget(props: WidgetProps) {
+  const { options } = props;
+  const { color, backgroundColor } = options;
+  return <input style={{ color, backgroundColor }} />;
+}
+
+// New pattern - React 19 compatible
+(MyCustomWidget as any).defaultOptions = {
```
heath-freenome: Ditto on the `any` cast

packages/utils/test/getWidget.test.tsx
```diff
@@ -42,27 +42,23 @@ const TestRefWidget: Widget = forwardRef<HTMLSpanElement, Partial<WidgetProps>>(
   );
 });
 
-TestRefWidget.defaultProps = {
-  options: { id: 'test-id' },
-};
+// React 19+: Use defaultOptions instead of defaultProps.options
+(TestRefWidget as any).defaultOptions = { id: 'test-id' };
```
heath-freenome: Ditto on the `any` cast

packages/utils/test/getWidget.test.tsx
```diff
@@ -42,27 +42,23 @@ const TestRefWidget: Widget = forwardRef<HTMLSpanElement, Partial<WidgetProps>>(
...
 function TestWidgetDefaults(props: WidgetProps) {
   const { options } = props;
   return <div {...options}>test</div>;
 }
 
-TestWidgetDefaults.defaultProps = {
-  options: { color: 'yellow' },
-};
+// React 19+: Use defaultOptions instead of defaultProps.options
+(TestWidgetDefaults as any).defaultOptions = { color: 'yellow' };
```
heath-freenome: Ditto on the `any` cast

packages/docs/docs/advanced-customization/custom-widgets-fields.md
Ashutoshx7: Updated with  WidgetWithDefaultOptions type -no more any casts

packages/utils/src/getWidget.tsx
nickgros, line 70: I don't know that we want to maintain this `defaultOptions` pattern. Custom widget authors could implement this in their widgets if they want to. With `defaultProps` gone in React 19, I think we may want to just remove `mergeWidgetOptions` in RJSF v7. With that in mind, I don't think we should add this `defaultOptions` compatibility feature. @Ashutoshx7, is there any reason that I'm not realizing that explains why we _need_ to support this in RJSF?

heath-freenome: @nickgros You raise a good point and I believe you are correct. I am curious to hear what @Ashutoshx7 has to say in reply to your question.

Ashutoshx7: sorry for the late reply !

The main motivation behind this change was to provide a smooth React 19 transition for existing custom widgets that currently rely on defaultProps.options, without breaking them immediately. The idea behind defaultOptions was to offer a React-19-safe replacement while keeping backward compatibility during the v7 lifecycle.

I’ll update the PR to drop the defaultOptions support and limit it to addressing the React 19 deprecations only (removing defaultProps usage and replacing react-is.isForwardRef). 

I’ll push the updated changes shortly.

@nickgros  
@heath-freenome

packages/utils/src/types.ts
heath-freenome, line 1: Did you actually change anything in this file or is it just whitespace. If just whitespace, please revert

packages/utils/src/getWidget.tsx
```diff
@@ -101,11 +100,9 @@ export default function getWidget<T = any, S extends StrictRJSFSchema = RJSFSche
...
-  if (
-    typeof widget === 'function' ||
-    (widget && ReactIs.isForwardRef(createElement(widget))) ||
-    ReactIs.isMemo(widget)
-  ) {
+  // Check if widget is a valid React component.
+  // We exclude strings because in RJSF, string widgets are keys in the registry, not HTML elements.
+  if (ReactIs.isValidElementType(widget) && typeof widget !== 'string') {
     return mergeWidgetOptions<T, S, F>(widget as Widget<T, S, F>);
   }
```
heath-freenome, line 107: Actually, once we drop `defaultProps` support in v7 this whole conditional and the `mergeWidgetOptions()` function will no longer be necessary. Which means that the `MergedWidget` feature will just go away.

packages/utils/test/getWidget.test.tsx
```diff
@@ -42,27 +42,20 @@ const TestRefWidget: Widget = forwardRef<HTMLSpanElement, Partial<WidgetProps>>(
   );
 });
 
-TestRefWidget.defaultProps = {
-  options: { id: 'test-id' },
-};
+TestRefWidget.defaultProps = { options: { id: 'test-id' } };
 
 function TestWidget(props: WidgetProps) {
   const { options } = props;
```
heath-freenome, line 48: Convert defaultProps into a destructure default assignment
```suggestion
  const { options, id = 'foo' } = props;
```

packages/utils/test/getWidget.test.tsx
```diff
@@ -42,27 +42,20 @@ const TestRefWidget: Widget = forwardRef<HTMLSpanElement, Partial<WidgetProps>>(
...
 }
 
-TestWidget.defaultProps = {
-  id: 'foo',
-};
-
 function TestWidgetDefaults(props: WidgetProps) {
   const { options } = props;
   return <div {...options}>test</div>;
 }
```
heath-freenome, line 55: The replacement for`defaultProps.options` merging
```suggestion
function TestWidgetDefaults(props: WidgetProps) {
  const { options } = props;
  const allOptions = { color: 'yellow', ...options };
  return <div {...allOptions}>test</div>;
}
```

packages/utils/test/getWidget.test.tsx
```diff
@@ -42,27 +42,20 @@ const TestRefWidget: Widget = forwardRef<HTMLSpanElement, Partial<WidgetProps>>(
...
 
-TestWidgetDefaults.defaultProps = {
-  options: { color: 'yellow' },
-};
+TestWidgetDefaults.defaultProps = { options: { color: 'yellow' } };
+
 
 const widgetProps: WidgetProps = {
```
heath-freenome, line 60:
```suggestion

const widgetProps: WidgetProps = {
```

packages/utils/test/getWidget.test.tsx
```diff
@@ -42,27 +42,20 @@ const TestRefWidget: Widget = forwardRef<HTMLSpanElement, Partial<WidgetProps>>(
   );
 });
 
-TestRefWidget.defaultProps = {
-  options: { id: 'test-id' },
-};
+TestRefWidget.defaultProps = { options: { id: 'test-id' } };
 
 function TestWidget(props: WidgetProps) {
```
heath-freenome, line 47: We would then also drop testing with `.defaultProps` and instead roll that into the widget implementation instead in a manner similar to the changes to line 54 below

```
const { options } = props;
const allOptions = { id: 'test-id', ...options };
```

```suggestion

function TestWidget(props: WidgetProps) {
```

heath-freenome:
Closing in favor of the simpler solution in v7 in https://github.com/rjsf-team/react-jsonschema-form/pull/5258

PR #4913, opened 2026-01-04
fix empty fieldset for pure unions

kolodny:

Fix empty fieldset rendering for pure oneOf/anyOf schemas without properties.

When a schema has `type: "object"` with `oneOf` or `anyOf` but no `properties` defined at the root level, `ObjectFieldTemplate` renders an empty `<fieldset id="root_XxxOf">` wrapper. This happens because:

1. `SchemaField` sees `type: "object"` and routes to `ObjectField`
2. `ObjectField` calls `ObjectFieldTemplate` with an empty `properties` array
3. `ObjectFieldTemplate` renders the fieldset wrapper even though it has no content
4. `MultiSchemaField` handles the actual oneOf/anyOf rendering separately

The fix checks for this "pure union" case (oneOf/anyOf present, no schema.properties, empty rendered properties array) and returns `null` to skip the empty wrapper.

**Reproduction:** Open the [RJSF Playground](https://rjsf-team.github.io/react-jsonschema-form/), select the "One Of" sample with the default theme, and inspect the DOM to see `<fieldset id="root_XxxOf"></fieldset>` with no children.

### Checklist

### DIFF
--- a/packages/core/test/StringField.test.tsx
+++ b/packages/core/test/StringField.test.tsx
@@ -778,86 +778,6 @@ describe('StringField', () => {
       expect(node.querySelector('[type=datetime-local]')).toHaveAttribute('id', 'root');
     });
 
-    // These two tests no longer work with @testing-utils/react's fireEvent since the input field does not accept
-    // invalid data, unlike when `Simulate` was being used
-    //
-    // it('should reject an invalid entered datetime and shows error', () => {
-    //   const { node, onChange } = createFormComponent({
-    //     schema: {
-    //       type: 'string',
-    //       format: 'date-time',
-    //     },
-    //     liveValidate: true,
-    //   });
-    //   let inputs = node.querySelectorAll('.form-group.rjsf-field-error input[type=datetime-local]');
-    //   expect(inputs).toHaveLength(0);
-    //   act(() => {
-    //     fireEvent.change(node.querySelector('[type=datetime-local]')!, {
-    //       target: { value: 'invalid' },
-    //     });
-    //   });
-    //   expect(onChange).toHaveBeenLastCalledWith(
-    //     expect.objectContaining({
-    //       errorSchema: { __errors: ['must be string'] },
-    //       errors: [
-    //         {
-    //           message: 'must be string',
-    //           name: 'type',
-    //           params: { type: 'string' },
-    //           property: '',
-    //           schemaPath: '#/type',
-    //           stack: 'must be string',
-    //           title: '',
-    //         },
-    //       ],
-    //     }),
-    //     'root',
-    //   );
-    //   inputs = node.querySelectorAll('.form-group.rjsf-field-error input[type=datetime-local]');
-    //   expect(inputs).toHaveLength(1);
-    // });
-    //
-    // it('should reject an invalid entered datetime and hides error', () => {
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -84,6 +84,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Fixed issue `"ui:rows"` option for `textarea` elements, fixing [#4903](https://github.com/rjsf-team/react-jsonschema-form/issues/4903).
 
+## @rjsf/core
+
+- Fixed `ObjectFieldTemplate` rendering an empty fieldset for pure oneOf/anyOf schemas (schemas with `type: "object"` and oneOf/anyOf but no properties at root level)
```
heath-freenome: Please add this to the existing `@rjsf/core` above instead of duplicating the section

CHANGELOG.md
```diff
@@ -84,6 +84,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Fixed issue `"ui:rows"` option for `textarea` elements, fixing [#4903](https://github.com/rjsf-team/react-jsonschema-form/issues/4903).
```
heath-freenome, line 87: Also, while not part of your MR, would you also move this up to the `@rjsf/shadcn` section as well?

packages/core/src/components/templates/ObjectFieldTemplate.tsx
```diff
@@ -46,6 +46,12 @@ export default function ObjectFieldTemplate<
     registry,
     options,
   );
+  const isPureUnionSchema = (schema.oneOf || schema.anyOf) && !schema.properties && properties.length === 0;
```
heath-freenome, line 52: Add a comment describing the need for this logic

packages/core/test/anyOf.test.jsx
```diff
@@ -33,6 +33,36 @@ describe('anyOf', () => {
...
+        {
+          properties: {
+            bar: { type: 'string' },
+          },
+        },
+      ],
+    };
+
+    const { node } = createFormComponent({
+      ref: React.createRef(),
```
heath-freenome: I don't see you using the `ref` anywhere so you likely can remove it.

heath-freenome:
Also, can you rebase and deal with the conflict I created by switching the tests to jest/typescript. And finally, can you show a screen capture of the playground with your change working?

CHANGELOG.md
kolodny: Moved

CHANGELOG.md
kolodny: Done

packages/core/src/components/templates/ObjectFieldTemplate.tsx
kolodny: Added comment for this

packages/core/test/anyOf.test.jsx
kolodny: Done

kolodny:
Here's the existing playground elements panel
Google Chrome 2026-01-05 20 34 24

---

And here's the fixed playground
Google Chrome 2026-01-05 20 35 16

Issue #670, opened 2017-08-16
Recursive references and uiSchema


mikebarkmin:
How do you handle recursive references with a custom uiSchema? This code does not work:

```javascript
const schema = {
  definitions: {
    block: {
      type: 'object',
      properties: {
        head: {
          type: 'string'
        },
        content: {
          type: 'string'
        },
        blocks: {
          type: 'array',
          items: {
            $ref: '#/definitions/block'
          }
        }
      }
    }
  },
  type: 'object',
  properties: {
    blocks: {
      type: 'array',
      items: {
        $ref: '#/definitions/block'
      }
    }
  }
};

const uiSchema = {
  definitions: {
    block: {
      content: {
        'ui:widget': 'textarea'
      },
      blocks: {
        items: {
          $ref: '#/definitions/block'
        }
      }
    }
  },
  blocks: {
    items: {
      $ref: '#/definitions/block'
    }
  }
};
```

n1k0:
This is being addressed in https://github.com/mozilla-services/react-jsonschema-form/pull/557

phemmer:
Was this ever addressed? I see it was closed due to #557, but #557 was also closed, and I see no other issues on the subject.

Issue #477, opened 2017-02-17
uiSchema for definitions


AnttiVikman:
Currently it seems there is no way to define ui options via uiSchema for the definitions. So where using definitions we can remove the repetition in schema, we are still forced to repeat ui options for each field in uiSchema.

Quick sample https://jsfiddle.net/9rwrn19d/1/

n1k0:
This is indeed a current limitation. Note that meanwhile you can leverage good old JavaScript to avoid too much verbosity: https://jsfiddle.net/yx09e5gp/1/

But I agree it'd be better to natively support uiSchema for definitions.

MatejMazur:
@n1k0 How would you solve uiSchema for a sample like [References in playground](https://mozilla-services.github.io/react-jsonschema-form/)?

epicfaace:
Merging this into #701.

PR #4947, opened 2026-01-26
feat: add ui:definitions for recursive and reusable uiSchema

adamrimon:

This PR adds a new `ui:definitions` property to uiSchema, enabling UI customizations for schema `$ref` references. This addresses a long-standing limitation where there was no way to define uiSchema for recursive schemas or to avoid repeating uiSchema for reused definitions.

Relates to #670 (Recursive references and uiSchema) and #477 (uiSchema for definitions)

**The Problem:**
```tsx
// Schema with recursive definition
const schema = {
  definitions: {
    node: {
      type: 'object',
      properties: {
        name: { type: 'string' },
        children: { type: 'array', items: { $ref: '#/definitions/node' } }
      }
    }
  },
  properties: { tree: { $ref: '#/definitions/node' } }
};

// Previously: No way to define uiSchema for recursive nodes
// uiSchema only applied to the first level, not deeper recursion
```

**The Solution:**
```tsx
const uiSchema = {
  'ui:definitions': {
    '#/definitions/node': {
      name: { 'ui:placeholder': 'Enter node name' }
    }
  }
};
// Now applies at ALL recursion levels
```

**Key features:**
- Define uiSchema once for schemas used in multiple places
- Support for recursive schemas at any depth
- Local overrides merge with and take precedence over definitions
- Different uiSchema for same property names in different `oneOf`/`anyOf` branches
- Keys use exact `$ref` paths (e.g., `#/definitions/address` or `#/$defs/node`)

**Changes:**
- **@rjsf/utils**: Added `expandUiSchemaDefinitions()` and `resolveUiSchema()` functions, `UiSchemaDefinitions` type, and related constants
- **@rjsf/core**: Integrated definition expansion at form initialization and runtime resolution in `SchemaField`
- **@rjsf/playground**: Updated References sample to demonstrate the feature
- **@rjsf/docs**: Added documentation for `ui:definitions`

### Checklist

### DIFF
--- a/packages/core/test/SchemaField.test.tsx
+++ b/packages/core/test/SchemaField.test.tsx
@@ -58,6 +58,7 @@ describe('SchemaField', () => {
           idSeparator: DEFAULT_ID_SEPARATOR,
           useFallbackUiForUnsupportedType: false,
         },
+        uiSchemaDefinitions: {},
       });
     });
     it('should provide expected registry with globalUiOptions as prop', () => {
@@ -98,6 +99,7 @@ describe('SchemaField', () => {
           idSeparator: DEFAULT_ID_SEPARATOR,
           useFallbackUiForUnsupportedType: false,
         },
+        uiSchemaDefinitions: {},
       });
     });
   });


packages/utils/src/types.ts
```diff
@@ -1148,6 +1161,12 @@ export type UiSchema<
     items?:
       | UiSchema<ArrayElement<T>, S, F>
       | ((itemData: ArrayElement<T>, index: number, formContext?: F) => UiSchema<ArrayElement<T>, S, F>);
```
heath-freenome: Uh oh, it looks like this was accidentally put here instead of `UIOptionsBaseType` can you move this?

packages/docs/docs/api-reference/uiSchema.md
```diff
@@ -63,6 +63,106 @@ const uiSchema: UiSchema = {
...
+  // Local overrides take precedence over ui:definitions
+  shipping_address: {
+    street_address: { 'ui:placeholder': 'Shipping street (overrides definition)' },
+  },
+};
+
+render(<Form schema={schema} uiSchema={uiSchema} validator={validator} />, document.getElementById('app'));
+```
+
+#### Local Overrides
```
heath-freenome, line 129: It seems like you are trying to mirror the `JSON Schema` concept, which does NOT support local overrides of `definitions`. So I'm curious why you chose to do it for the `UiSchema`? It seems like the potential recursive nature of the `JSON Schema` would make it hard to support overrides at a sub-field level with recursion?

packages/utils/src/resolveUiSchema.ts
```diff
@@ -0,0 +1,177 @@
...
+ * @param visited - Set of $refs already visited (to detect recursion)
+ * @returns - The expanded uiSchema with definitions merged in
+ */
+export function expandUiSchemaDefinitions<
+  T = any,
+  S extends StrictRJSFSchema = RJSFSchema,
+  F extends FormContextType = any,
+>(
+  currentSchema: S,
+  rootSchema: S,
```
heath-freenome: Because the `registry` already contains the pointers to the `rootSchema` AND `uiSchemaDefinitions` objects, I'd recommend passing that in rather than these two parameters, and then extracting them inside of this function.

packages/utils/src/resolveUiSchema.ts
```diff
@@ -0,0 +1,177 @@
...
+      resolvedSchema = currentSchema;
+    }
+  }
+
+  // Process properties (each property maps to uiSchema[propName] - flattened)
+  const properties = resolvedSchema[PROPERTIES_KEY];
+  if (properties && isObject(properties)) {
+    for (const [propName, propSchema] of Object.entries(properties as Record<string, S>)) {
+      const propUiSchema = (result[propName] || {}) as UiSchema<T, S, F>;
+      const expanded = expandUiSchemaDefinitions(propSchema, rootSchema, propUiSchema, definitions, new Set(visited));
```
heath-freenome: Why not just reuse the existing visited set instead of copying it?
```suggestion
      const expanded = expandUiSchemaDefinitions(propSchema, rootSchema, propUiSchema, definitions, visited);
```

packages/utils/src/resolveUiSchema.ts
```diff
@@ -0,0 +1,177 @@
...
+    if (subSchema && isObject(subSchema) && !Array.isArray(subSchema)) {
+      const currentUiSchema = result[keyword];
+      if (typeof currentUiSchema !== 'function') {
+        const subUiSchema = ((currentUiSchema as GenericObjectType) || {}) as UiSchema<T, S, F>;
+        const expanded = expandUiSchemaDefinitions(
+          subSchema as S,
+          rootSchema,
+          subUiSchema,
+          definitions,
+          new Set(visited),
```
heath-freenome: Why not just reuse the existing visited set instead of copying it?
```suggestion
          visited,
```

packages/utils/src/resolveUiSchema.ts
```diff
@@ -0,0 +1,177 @@
...
+      let hasExpanded = false;
+      for (let i = 0; i < schemaOptions.length; i++) {
+        const optionSchema = schemaOptions[i] as S;
+        const optionUiSchema = (uiSchemaArray[i] || {}) as UiSchema<T, S, F>;
+        const expanded = expandUiSchemaDefinitions(
+          optionSchema,
+          rootSchema,
+          optionUiSchema,
+          definitions,
+          new Set(visited),
```
heath-freenome: Why not just reuse the existing visited set instead of copying it?
```suggestion
          visited,
```

packages/core/src/components/Form.tsx
```diff
@@ -538,6 +540,12 @@ export default class Form<
 
     const rootSchema = schemaUtils.getRootSchema();
 
+    // Pre-expand ui:definitions into the uiSchema structure
```
heath-freenome: Move this code down to after these lines so that you can use the registry rather than reading the uiSchema as you have it now:
```
    // Only store a new registry when the props cause a different one to be created
    const newRegistry = this.getRegistry(props, rootSchema, schemaUtils);
    const registry = deepEquals(state.registry, newRegistry) ? state.registry : newRegistry;
```

packages/core/src/components/Form.tsx
```diff
@@ -1149,6 +1157,7 @@ export default class Form<
       translateString: customTranslateString || translateString,
       globalUiOptions: uiSchema[UI_GLOBAL_OPTIONS_KEY],
       globalFormOptions: this.getGlobalFormOptions(props),
+      uiSchemaDefinitions: uiSchema[UI_DEFINITIONS_KEY],
```
heath-freenome: Does it make sense to provide an empty object for this if not present?
```suggestion
      uiSchemaDefinitions: uiSchema[UI_DEFINITIONS_KEY] ?? {},
```

heath-freenome (commented):
Thanks for this feature... Although the tickets you referenced are closed, are there any newer tickets related to this are open?

packages/docs/docs/api-reference/uiSchema.md
adamrimon: The intent isn't to mirror `$defs` semantics, but rather to provide a way to attach uiSchema to schemas that are referenced via `$ref`. Without this, there's no way to define UI customizations for recursive schemas or avoid duplicating uiSchema for definitions used in multiple places.

Currently, uiSchema works by specifying customizations at paths that mirror the schema structure. I wanted to keep that behavior while also providing a default for `$ref`s - so `ui:definitions` provides the default, but you can still customize individual instances at their paths.

packages/utils/src/resolveUiSchema.ts
adamrimon: Copying it prevents sibling branches from interfering with each other. For example, if a schema has two properties that both reference `#/definitions/node`, reusing `visited` would cause the second property to incorrectly see `#/definitions/node` as already visited

packages/utils/src/resolveUiSchema.ts
adamrimon: Copying it prevents sibling branches from interfering with each other. For example, if a schema has two properties that both reference #/definitions/node, reusing visited would cause the second property to incorrectly see #/definitions/node as already visited

packages/core/src/components/Form.tsx
adamrimon: Done

packages/utils/src/resolveUiSchema.ts
adamrimon: Thanks, good idea

packages/utils/src/resolveUiSchema.ts
adamrimon: Same as above

packages/core/src/components/Form.tsx
adamrimon: Done

heath-freenome (approved):

PR #4968, opened 2026-02-23
fix: extraErrors not displaying on first async set after submit

### LINKED ISSUE #4965: The extraErrors is not showing the first time set
### Prerequisites

### What theme are you using?

antd

### Version

6.x

### Current Behavior

When you set the `extraErrors` value of the form in a submit callback for example, the first time you do it, the error won't be shown. But if you do it again, the errors will be displayed correctly, so are the following attempts.

### Expected Behavior

When the `extraErrors` value is set first time, the error should be shown as well.

### Steps To Reproduce

I created a repo for reproducing the error:

https://github.com/fangpenlin/react-jsonschema-form-extra-error-bug-reproduce

a link to codesandbox is available here:

https://codesandbox.io/p/github/fangpenlin/react-jsonschema-form-extra-error-bug-reproduce/master?file=%2Fsrc%2FApp.tsx%3A46%2C11

### Environment

```markdown
- OS:
- Node:
- npm:
```

### Anything else?

_No response_

adamrimon:

Fixes #4965

In v5, `UNSAFE_componentWillReceiveProps` handled prop changes, including `extraErrors`, by computing and applying new state before render. When v6 replaced it with `getSnapshotBeforeUpdate` + `componentDidUpdate`, state computation moved to after render, requiring a second render cycle via `setState` in `componentDidUpdate` to display the updated errors. That second cycle doesn't reliably fire in the browser when `extraErrors` is set asynchronously, causing errors to not appear on the first submit.

The fix adds `getDerivedStateFromProps` to detect when `extraErrors` changes and re-merge errors into state before render. This restores the single-render-cycle behavior that v5 had.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed `extraErrors` not displaying on first async set after submit, fixing [#4965](https://github.com/rjsf-team/react-jsonschema-form/issues/4965)
 - Updated multi-select ArrayFields to properly use the `items` uiSchema for enumerated options, fixing [#4955](https://github.com/rjsf-team/react-jsonschema-form/issues/4955)
 
+## @rjsf/utils
+
+- Fixed `resolveAllReferences` to preserve `$ref` on resolved schemas, enabling `ui:definitions` beyond the first recursion level, fixing [#4966](https://github.com/rjsf-team/react-jsonschema-form/issues/4966)
+
 # 6.3.1
 
 ## Dev / docs / playground
--- a/packages/core/test/uiSchema.test.tsx
+++ b/packages/core/test/uiSchema.test.tsx
@@ -2696,4 +2696,173 @@ describe('uiSchema', () => {
       });
     });
   });
+
+  describe('ui:definitions', () => {
+    it('should apply ui:definitions to a simple $ref property', () => {
+      const schema: RJSFSchema = {
+        type: 'object',
+        properties: {
+          address: { $ref: '#/$defs/address' },
+        },
+        $defs: {
+          address: {
+            type: 'object',
+            properties: {
+              street: { type: 'string' },
+              city: { type: 'string' },
+            },
+          },
+        },
+      };
+      const uiSchema: UiSchema = {
+        'ui:definitions': {
+          '#/$defs/address': {
+            street: { 'ui:placeholder': 'Enter street' },
+            city: { 'ui:widget': 'textarea' },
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/core
 
+- Fixed `extraErrors` not displaying on first async set after submit, fixing [#4965](https://github.com/rjsf-team/react-jsonschema-form/issues/4965)
```
heath-freenome, line 23: Can you add this below the existing lines instead of above

packages/core/src/components/Form.tsx
```diff
@@ -374,6 +376,38 @@ export default class Form<
...
   private _isProcessingUserChange = false;
 
+  /** When the `extraErrors` prop changes, re-merges `schemaValidationErrors` + `extraErrors` + `customErrors` into
+   * state before render, ensuring the updated errors are visible immediately in a single render cycle.
+   *
+   * @param props - The current props
+   * @param state - The current state
+   * @returns Partial state with re-merged errors if `extraErrors` changed, or `null` if no update is needed
+   */
+  static getDerivedStateFromProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(
```
heath-freenome, line 386: @adamrimon  What kind of complexities arise from this given that we already have a `getSnapshotBeforeUpdate()` method. Is it possible to merge the two? Would it be better if we did?

heath-freenome (commented):
@adamrimon Thanks for helping to make our library better. I have a question and comment

packages/core/src/components/Form.tsx
adamrimon: getSnapshotBeforeUpdate() is called by React after the call to render(), which is why we needed to add the getDerivedStateFromProps(). It is called just before render() - so it can update state in time for the first render cycle. They have different roles in React's lifecycle.
Theoretically we could move everything to getDerivedStateFromProps(), but it is a static method (with no access to instance methods like getStateFromProps()) so that would require a bigger and more complicated refactor.
I'm still studying React's lifecycle, but I think the best solution would be to simplify the whole thing by switching to a function definition of the component and use `useEffect` and `useMemo`.  For a future version, I'll try to come up with a design that handles all changes in a single place - but for now I believe this is the best solution.

heath-freenome: Fair enough. I'll approve.

PR #4971, opened 2026-02-23
fix: omitExtraData strips additionalProperties inside oneOf/anyOf

### LINKED ISSUE #4366: Unable to add new item for object with additionalProperties within anyOf / oneOf with liveOmit and omitExtraData
### Prerequisites

### What theme are you using?

core

### Version

5.x

### Current Behavior

When `liveOmit` and `omitExtraData` options are enabled, then you are unable to add a new item to an object that uses `additionalProperties` within an `anyOf` or `oneOf`.

Example json schema
```json
{
  "oneOf": [
    {
      "type": "object",
      "additionalProperties": {
        "foo": {
          "type": "string"
        }
      }
    }
  ]
}
```

RJSF options
```
liveOmit: true
omitExtraData: true
```

### Expected Behavior

Can add new items

### Steps To Reproduce

1. Visit [playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJvbmVPZiI6W3sidHlwZSI6Im9iamVjdCIsImFkZGl0aW9uYWxQcm9wZXJ0aWVzIjp7ImZvbyI6eyJ0eXBlIjoic3RyaW5nIn19fV19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwiZXhwZXJpbWVudGFsX2RlZmF1bHRGb3JtU3RhdGVCZWhhdmlvciI6eyJhcnJheU1pbkl0ZW1zIjp7InBvcHVsYXRlIjoicG9wdWxhdGUiLCJtZXJnZUV4dHJhRGVmYXVsdHMiOmZhbHNlfSwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQifSwibGl2ZU9taXQiOnRydWUsIm9taXRFeHRyYURhdGEiOnRydWV9fQ==)
2. Try and add a new item

### Environment

```markdown
Playground
```

### Anything else?

_No response_

adamrimon:

Fixes #4366

When `omitExtraData` and `liveOmit` are enabled, additional properties inside a `oneOf`/`anyOf` option are incorrectly stripped on every change. This makes it impossible to add new key-value pairs to objects with `additionalProperties` in those schemas.

The bug is specific to `oneOf`/`anyOf` because of how `retrieveSchema` works. When `additionalProperties` is directly on the root schema, `retrieveSchema` calls `stubExistingAdditionalProperties`, which expands formData keys into the schema's `properties`. This means `toPathSchema` sees them as regular properties and creates child entries for them.

But when `additionalProperties` is inside a `oneOf`/`anyOf` option, the resolved inner schema does not go through the same stubbing. So `toPathSchema` sets the `additionalProperties` flag on the `PathSchema` node but never creates child entries for the formData keys. When `getFieldNames` later walks the `PathSchema` to decide which fields to keep, it finds no children, returns an empty list, and all the data gets stripped.

The fix adds logic to `toPathSchema` that iterates over formData keys not present in `properties` and recursively builds `PathSchema` entries for them using the `additionalProperties` schema. This lets `getFieldNames` discover them naturally during traversal. It also enables deep stripping when the `additionalProperties` schema itself defines strict constraints.

### Checklist

  - [X] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.


packages/utils/test/schema/omitExtraDataTest.ts
```diff
@@ -311,6 +311,100 @@ export default function omitExtraDataTest(testValidator: TestValidatorType) {
       expect(schemaUtils.omitExtraData(schema, formData)).toEqual(formData);
     });
 
+    it('should not omit additional properties when root schema has additionalProperties', () => {
```
nickgros, line 314: These test make sense. I'm wondering what happens in the cases where we have two schemas in `oneOf` like

```json
{ 
	"oneOf": [{
		"type": "object",
		"properties": {
			"discriminator": { "const": "foo" }
		},
		"additionalProperties": true,
	},
	{
		"type": "object",
		"properties": {
			"discriminator": { "const": "bar" }
		},
		"additionalProperties": false,
	}]
}
```

Do we properly discard the additional property for each case? Does it make sense to add these as tests?

adamrimon: We do, but I added it as a test case :)

packages/utils/src/schema/toPathSchema.ts
```diff
@@ -98,6 +98,22 @@ function toPathSchemaInternal<T = any, S extends StrictRJSFSchema = RJSFSchema,
 
   if (ADDITIONAL_PROPERTIES_KEY in schema && schema[ADDITIONAL_PROPERTIES_KEY] !== false) {
     set(pathSchema, RJSF_ADDITIONAL_PROPERTIES_FLAG, true);
+    const additionalSchema =
+      typeof schema[ADDITIONAL_PROPERTIES_KEY] === 'object' ? (schema[ADDITIONAL_PROPERTIES_KEY] as S) : ({} as S);
```
heath-freenome: Using lodash's `isObject()` would result in a shorter line... and one need only cast once with a little paren switching
```suggestion
      (isObject(schema[ADDITIONAL_PROPERTIES_KEY]) ? schema[ADDITIONAL_PROPERTIES_KEY] : {}) as S;
```

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Fixed `resolveAllReferences` to preserve `$ref` on resolved schemas, enabling `ui:definitions` beyond the first recursion level, fixing [#4966](https://github.com/rjsf-team/react-jsonschema-form/issues/4966)
 
+- Fixed `omitExtraData` stripping `additionalProperties` inside oneOf/anyOf options, fixing [#4366](https://github.com/rjsf-team/react-jsonschema-form/issues/4366)
+
```
heath-freenome, line 30: No need for the extra line
```suggestion
- Fixed `resolveAllReferences` to preserve `$ref` on resolved schemas, enabling `ui:definitions` beyond the first recursion level, fixing [#4966](https://github.com/rjsf-team/react-jsonschema-form/issues/4966) 
- Fixed `omitExtraData` stripping `additionalProperties` inside oneOf/anyOf options, fixing [#4366](https://github.com/rjsf-team/react-jsonschema-form/issues/4366)

```

PR #4988, opened 2026-03-13
add removeEmptyOptionalObjects prop to prune empty optional objects

### LINKED ISSUE #4954: Optional object with required fields blocks submit after user clears input
### Prerequisites

### What theme are you using?

mui

### Version

6.2.5

### Current Behavior

I have an optional object property that contains required fields. From a JSON Schema perspective, this is valid: the object itself is optional, but if it exists, its inner fields must be present. But once the user interacts with the fields and then clears them, the form becomes impossible to submit, even though the object is optional and not listed in the parent required. Any user interaction “activates” the object permanently.

Example, [Sandbox link](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJ0ZXN0Ijp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7ImZpZWxkMSI6eyJ0eXBlIjoic3RyaW5nIn0sImZpZWxkMiI6eyJ0eXBlIjoic3RyaW5nIn19LCJyZXF1aXJlZCI6WyJmaWVsZDEiLCJmaWVsZDIiXX19fSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6Im11aSIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6dHJ1ZSwicmVhZG9ubHkiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlLCJleHBlcmltZW50YWxfY29tcG9uZW50VXBkYXRlU3RyYXRlZ3kiOiJjdXN0b21EZWVwIiwiZXhwZXJpbWVudGFsX2RlZmF1bHRGb3JtU3RhdGVCZWhhdmlvciI6eyJhcnJheU1pbkl0ZW1zIjoicG9wdWxhdGUiLCJhbGxPZiI6InNraXBEZWZhdWx0cyIsImNvbnN0QXNEZWZhdWx0cyI6ImFsd2F5cyIsImVtcHR5T2JqZWN0RmllbGRzIjoicG9wdWxhdGVBbGxEZWZhdWx0cyIsIm1lcmdlRGVmYXVsdHNJbnRvRm9ybURhdGEiOiJ1c2VGb3JtRGF0YUlmUHJlc2VudCJ9LCJ1c2VGYWxsYmFja0ZpZWxkIjpmYWxzZSwidXNlRmFsbGJhY2tVaUZvclVuc3VwcG9ydGVkVHlwZSI6ZmFsc2UsImxpdmVWYWxpZGF0ZSI6ZmFsc2UsImlkUHJlZml4Ijoicm9vdCIsImlkU2VwYXJhdG9yIjoiXyJ9LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IlNpbXBsZSJ9):
```json
{
  "type": "object",
  "properties": {
    "test": {
      "type": "object",
      "properties": {
        "field1": { "type": "string" },
        "field2": { "type": "string" }
      },
      "required": ["field1", "field2"]
    }
  }
}
```

### Expected Behavior

The form should be submittable, because: "test" object is not required. An empty and cleared optional object should be treated as “not present” and the object should be removed from formData.

### Steps To Reproduce

1. Render the form using the schema above, HTML 5 validation disabled.
2. Leave test untouched, form is submittable (expected).
3. Type something into field1 or field2.
4. Clear the input so both fields are empty.
5. Try to submit the form.

### Environment

```markdown
- OS: Windows 11
- Node: 24.6.0
- npm: 11.7.0
```

### Anything else?

Is there a way in rjsf to treat an optional object as absent when all its fields are empty? If not, would this be considered a bug or a feature request?

Suyog241005:
Fixes #4954 

### Problem 
Currently, if a user types into an optional object's fields and then clears them out, the form retains the empty object in formData (e.g., `{ test: {} }` or {` test: { field1: "" }}`). If the optional object itself has required fields, schema validation will fail on the now-empty object, indefinitely preventing form submission even though the entire object was supposed to be optional.

### Solution 
This PR introduces a new opt-in prop, `removeEmptyOptionalObjects`, to the `Form` component. When enabled, it recursively prunes optional objects from `formData` if all their properties are computationally empty (i.e. `""`, `null`, `undefined`, `[]`, or `{}`). An object is considered "optional" if it is not explicitly listed in its parent's `required` array.

This pruning occurs gracefully during standard form lifecycle events (`onChange`, `onBlur`, `programmatic` `validateForm`, and `onSubmit`) to ensure the form safely behaves as if the user never interacted with the optional tree.

### Checklist


packages/core/src/components/Form.tsx
```diff
@@ -937,6 +944,15 @@ export default class Form<
       };
     }
 
+    if (this.props.removeEmptyOptionalObjects === true) {
```
nickgros: Destructure the prop above for consistency
```ts
    const { extraErrors, omitExtraData, liveOmit, noValidate, liveValidate, onChange, removeEmptyOptionalObjects } = this.props;
```

```suggestion
    if (removeEmptyOptionalObjects === true) {
```

packages/core/src/components/Form.tsx
```diff
@@ -937,6 +944,15 @@ export default class Form<
...
+      const { schemaUtils: su, schema: s } = this.state;
+      newFormData = removeOptionalEmptyObjects(su.getValidator(), s, su.getRootSchema(), newFormData) as T;
```
nickgros: schemaUtils and schema are already in scope
```suggestion
      newFormData = removeOptionalEmptyObjects(schemaUtils.getValidator(), schema, schemaUtils.getRootSchema(), newFormData) as T;
```

packages/core/src/components/Form.tsx
```diff
@@ -1064,6 +1080,11 @@ export default class Form<
         newFormData = this.omitExtraData(formData);
         state = { formData: newFormData };
       }
+      if (this.props.removeEmptyOptionalObjects === true) {
+        const { schemaUtils: su, schema: s } = this.state;
+        newFormData = removeOptionalEmptyObjects(su.getValidator(), s, su.getRootSchema(), newFormData) as T;
+        state = { ...state, formData: newFormData };
+      }
```
nickgros: Same feedback (if a property is not yet in scope, destructure it at the beginning of the function)
```suggestion
      if (removeEmptyOptionalObjects === true) {
        newFormData = removeOptionalEmptyObjects(schemaUtils.getValidator(), schema, schemaUtils.getRootSchema(), newFormData) as T;
        state = { ...state, formData: newFormData };
      }
```

packages/core/src/components/Form.tsx
```diff
@@ -1128,6 +1149,16 @@ export default class Form<
...
+    if (this.props.removeEmptyOptionalObjects === true) {
+      const { schemaUtils, schema } = this.state;
+      newFormData = removeOptionalEmptyObjects(
+        schemaUtils.getValidator(),
+        schema,
+        schemaUtils.getRootSchema(),
+        newFormData,
+      ) as T;
+    }
+
```
nickgros: Same feedback (if a property is not yet in scope, destructure it at the beginning of the function)

```suggestion
    if (removeEmptyOptionalObjects === true) {
      newFormData = removeOptionalEmptyObjects(
        schemaUtils.getValidator(),
        schema,
        schemaUtils.getRootSchema(),
        newFormData,
      ) as T;
    }

```

packages/core/src/components/Form.tsx
```diff
@@ -1315,6 +1346,15 @@ export default class Form<
...
     }
+    if (this.props.removeEmptyOptionalObjects === true) {
+      const { schemaUtils, schema } = this.state;
+      newFormData = removeOptionalEmptyObjects(
+        schemaUtils.getValidator(),
+        schema,
+        schemaUtils.getRootSchema(),
+        newFormData,
+      ) as T;
+    }
```
nickgros: Same feedback (if a property is not yet in scope, destructure it at the beginning of the function)

```suggestion
    if (removeEmptyOptionalObjects === true) {
      newFormData = removeOptionalEmptyObjects(
        schemaUtils.getValidator(),
        schema,
        schemaUtils.getRootSchema(),
        newFormData,
      ) as T;
    }
```

packages/utils/src/removeOptionalEmptyObjects.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/** Determines whether a value is considered "empty" for the purposes of optional object pruning.
+ * A value is empty if it is `undefined`, `null`, an empty string, or an object where all own
+ * properties are themselves empty.
+ *
+ * @param value - The value to check
+ * @returns True if the value is considered empty
+ */
+function isValueEmpty(value: unknown): boolean {
+  if (value === undefined || value === null || value === '') {
```
nickgros: As a convention we have been using lodash `isNil` (please ensure you import from `'lodash/isNil'` for tree-shaking)

```suggestion
  if (isNil(value) || value === '') {
```

packages/utils/src/removeOptionalEmptyObjects.ts
```diff
@@ -0,0 +1,100 @@
...
+ * @param [formData] - The current form data to prune
+ * @returns - A new copy of `formData` with empty optional objects removed, or `undefined` if the
+ *           entire formData was pruned
+ */
+export default function removeOptionalEmptyObjects<
+  T = any,
+  S extends StrictRJSFSchema = RJSFSchema,
+  F extends FormContextType = any,
+>(validator: ValidatorType<T, S, F>, schema: S, rootSchema?: S, formData?: T): T | undefined {
+  if (!isObject(formData) || !isObject(schema)) {
```
nickgros: You might also want to support arrays? An array's `items` schema could define an inner optional object.

What do you think?

packages/core/src/components/Form.tsx
```diff
@@ -208,6 +209,12 @@ export interface FormProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
    * called. Set to `false` by default.
    */
   omitExtraData?: boolean;
+  /** If set to true, optional object properties whose fields are all empty (undefined, null, or empty string)
+   * will be automatically removed from formData. This prevents the scenario where interacting with fields inside
+   * an optional object "activates" it permanently, making the form unsubmittable when the optional object has
+   * required inner fields. This works independently of `omitExtraData`. Set to `false` by default.
+   */
+  removeEmptyOptionalObjects?: boolean;
```
nickgros, line 217: Please add corresponding documentation to form-props.md in `docs`

packages/utils/src/removeOptionalEmptyObjects.ts
Suyog241005: Great catch. I will update the utility to recursively process arrays as well, ensuring optional objects within array items are properly pruned. I will also add test cases covering both root level and nested arrays (including tuples) to ensure everything behaves as expected.

Suyog241005: Added the array support and test cases for the same

packages/core/src/components/Form.tsx
Suyog241005: Updated the docs ✅

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 6.4.1
 
+## @rjsf/core
```
heath-freenome, line 21: This is a new feature, add a new 6.5.0 section for it.

Suyog241005: done ✅

Suyog241005:
@heath-freenome I think this pr is ready to be merged

packages/core/src/components/Form.tsx
```diff
@@ -937,6 +945,19 @@ export default class Form<
       };
     }
 
+    if (removeEmptyOptionalObjects === true) {
```
heath-freenome: Since this is an optional boolean, you need only do this:

```suggestion
    if (removeEmptyOptionalObjects) {
```

packages/core/src/components/Form.tsx
```diff
@@ -1051,19 +1072,28 @@ export default class Form<
...
       const { onChange, extraErrors } = this.props;
-      const { formData } = this.state;
+      const { formData, schemaUtils, schema } = this.state;
       let newFormData: T | undefined = formData;
       let state: Partial<FormState<T, S, F>> = { formData: newFormData };
       if (omitExtraData === true && liveOmit === 'onBlur') {
         newFormData = this.omitExtraData(formData);
         state = { formData: newFormData };
       }
+      if (removeEmptyOptionalObjects === true) {
```
heath-freenome: Ditto

```suggestion
      if (removeEmptyOptionalObjects) {
```

packages/core/src/components/Form.tsx
```diff
@@ -1121,13 +1151,23 @@ export default class Form<
...
     event.persist();
-    const { omitExtraData, extraErrors, noValidate, onSubmit } = this.props;
+    const { omitExtraData, extraErrors, noValidate, onSubmit, removeEmptyOptionalObjects } = this.props;
     let { formData: newFormData } = this.state;
 
     if (omitExtraData === true) {
       newFormData = this.omitExtraData(newFormData);
     }
 
+    if (removeEmptyOptionalObjects === true) {
```
heath-freenome: ditto
```suggestion
    if (removeEmptyOptionalObjects) {
```

packages/core/src/components/Form.tsx
```diff
@@ -1310,11 +1350,20 @@ export default class Form<
    * @returns - True if the form is valid, false otherwise.
    */
   validateForm() {
-    const { omitExtraData } = this.props;
+    const { omitExtraData, removeEmptyOptionalObjects } = this.props;
     let { formData: newFormData } = this.state;
     if (omitExtraData === true) {
       newFormData = this.omitExtraData(newFormData);
     }
+    if (removeEmptyOptionalObjects === true) {
```
heath-freenome: Ditto
```suggestion
    if (removeEmptyOptionalObjects) {
```

packages/utils/src/removeOptionalEmptyObjects.ts
```diff
@@ -0,0 +1,131 @@
+import isObject from './isObject';
+import { FormContextType, GenericObjectType, RJSFSchema, StrictRJSFSchema, ValidatorType } from './types';
+import retrieveSchema from './schema/retrieveSchema';
+import isNil from 'lodash/isNil';
```
heath-freenome: Please keep the global imports first and separate from the locals
```suggestion
import isNil from 'lodash/isNil';

import isObject from './isObject';
import retrieveSchema from './schema/retrieveSchema';
import { FormContextType, GenericObjectType, RJSFSchema, StrictRJSFSchema, ValidatorType } from './types';
```

packages/utils/src/removeOptionalEmptyObjects.ts
```diff
@@ -0,0 +1,131 @@
...
+    // An empty array is considered empty; a non-empty array is not
+    return value.length === 0;
+  }
+  if (isObject(value)) {
+    const obj = value as GenericObjectType;
+    const keys = Object.keys(obj);
+    if (keys.length === 0) {
+      return true;
+    }
+    return keys.every((key) => isValueEmpty(obj[key]));
```
heath-freenome: `every()` always returns true for an empty array.
```suggestion
    return keys.every((key) => isValueEmpty(obj[key]));
```

packages/utils/src/removeOptionalEmptyObjects.ts
```diff
@@ -0,0 +1,131 @@
...
+        itemSchema = itemsSchema[index] || (resolvedSchema.additionalItems as S) || ({} as S);
+      }
+      const cleaned = removeOptionalEmptyObjects<T, S, F>(validator, itemSchema, rootSchema, item);
+      if (cleaned !== item) {
+        hasChanges = true;
+      }
+      return cleaned === undefined ? ({} as T) : cleaned;
+    });
+
+    return hasChanges ? (mapped as unknown as T) : formData;
```
heath-freenome: Is there a way to type the `formData.map()` so that you don't need the `as unknown` intermediary cast? It's ok if you can't, and I'd love for you to try first. IF you can't, at least add a comment explaining why the cast.

packages/utils/src/removeOptionalEmptyObjects.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  if (!isObject(formData)) {
+    return formData;
+  }
+
+  const { properties, required: requiredFields = [] } = resolvedSchema;
+
+  if (!properties) {
+    return formData;
+  }
```
heath-freenome: Let's combine these two conditions to bail out early
```suggestion
  const { properties, required: requiredFields = [] } = resolvedSchema;
  if (!isObject(formData) || !properties) {
    return formData;
  }
```

packages/utils/src/removeOptionalEmptyObjects.ts
```diff
@@ -0,0 +1,131 @@
...
+        // This is an optional property and the cleaned result is empty — omit it
+        continue;
+      }
+
+      result[key] = cleaned;
+      hasAnyValue = true;
+    } else if (!isRequired && isValueEmpty(value) && properties[key]) {
+      // Optional scalar property that is empty — omit it
+      continue;
+    } else {
```
heath-freenome, line 112: This is essentially doing the work that omitExtraData does... I wonder if users turn this feature on whether we actually need to run the other code?

Suyog241005: @heath-freenome Currently, `removeOptionalEmptyObjects` is a separate traversal. Based on your feedback, I can integrate the `omitExtraData` logic directly into this function to avoid redundant passes when both are enabled. Would you prefer this consolidated approach, or should I keep them separate for now?

packages/utils/src/removeOptionalEmptyObjects.ts
Suyog241005: @heath-freenome i think to remove the intermediary cast need to add this 
```
const data = formData as any[];
    const mapped = data.map((item, index) => {
      let itemSchema = itemsSchema as S;
      if (Array.isArray(itemsSchema)) {
        itemSchema = itemsSchema[index] || (resolvedSchema.additionalItems as S) || ({} as S);
      }
      const cleaned = removeOptionalEmptyObjects<T, S, F>(validator, itemSchema, rootSchema, item);
      if (cleaned !== item) {
        hasChanges = true;
      }
      // If the cleaning removed everything, return an empty object (placeholder)
      return cleaned === undefined ? ({} as T) : cleaned;
    });
    // We cast back to T since we've established at the start of this block that T is an array type
    return hasChanges ? (mapped as T) : formData;
```

packages/utils/src/removeOptionalEmptyObjects.ts
heath-freenome: I'm very curious whether this logic actually replaces what omitExtraData already does? Or do you already know that omitExtraData does more?

packages/utils/src/removeOptionalEmptyObjects.ts
heath-freenome: Hmmm, I am not a fan of the `any` cast. At some point I will likely update the linter to prevent them by default. Ok. I was just wondering

packages/utils/src/removeOptionalEmptyObjects.ts
Suyog241005: @heath-freenome Currently, `omitExtraData` strictly removes any data not defined in the schema (using PathSchema). `removeOptionalEmptyObjects` currently preserves extra data; its primary goal is to prune optional keys that are in the schema but are effectively empty, which prevents validation errors for optional objects with nested required fields.

If we want to consolidate them, I would need to update this logic to also omit extra data when requested, making it a comprehensive **data cleaning** pass.

packages/utils/src/removeOptionalEmptyObjects.ts
Suyog241005: I hear you on the `any` cast! I've updated the implementation to avoid that explicit `as any[]` cast entirely. I've also simplified the final return to a single `(mapped as T)` cast, successfully removing the unknown intermediary, and added the explanatory comment we discussed.

packages/utils/src/removeOptionalEmptyObjects.ts
heath-freenome: @nickgros I'd love your thoughts on the idea of merging the two data cleaning options? Maybe it makes sense to take a second pass at the fix after merging this PR?

@Suyog241005 I'm leaning towards merging this now, especially if you are open to  attempting the data cleaning merge in a second PR. It would be awesome to figure out if doing this merge (especially via the minimization/elimination of the use of PathSchemas which are not very performant) would improve the speed of the algorithms.

Suyog241005: > @Suyog241005 I'm leaning towards merging this now, especially if you are open to attempting the data cleaning merge in a second PR. It would be awesome to figure out if doing this merge (especially via the minimization/elimination of the use of PathSchemas which are not very performant) would improve the speed of the algorithms.

@heath-freenome Sounds like a great plan! I'm definitely open to the data cleaning pass in a second PR. I agree that minimizing the use of `PathSchema` could provide a significant performance boost, and I'd love to help explore that.

nickgros:
@heath-freenome Agree that consolidating the two into a single data cleaning step makes sense and could make for both cleaner code and improved performance. I also agree that that can be follow-up work; let's get this new feature in first!

x0k:
I don’t quite understand why:
- this feature should be a separate utility rather than a couple of checks when removing data from object/array fields
- there needs to be an option to disable this behavior

Nevertheless, as I mentioned in one of the PRs, in my project I implemented [omitExtraData](https://github.com/x0k/svelte-jsonschema-form/blob/70cbcaceb5400446130bcd336d7af01a92ab2027/packages/form/src/omit-extra-data.ts) based on schema traversal - it might be useful here as well.

PR #5001, opened 2026-03-25
Feat/UI initial value and required

### LINKED ISSUE #4983: Extend `ui:emptyValue` and complement it with `ui:initialValue` and `ui:required`
### Prerequisites

### What theme are you using?

utils

When the JSON Schema is shared across forms, comes from an API, or is auto-generated, you can't easily set per-form defaults or change which fields are required without touching the schema itself.

Example: a "person details" schema has a `country` field. A US-only form wants to hide it and pre-fill `"US"`. Right now you'd have to modify the schema or stuff the value into `formData`, which mixes up user input with form setup.

Two new uiSchema properties:

**`ui:initialValue`**: Pre-fills a field on initial render and after reset. Takes priority over `schema.default`.

This is about what the user sees, while the existing `ui:emptyValue` is about what gets stored when a field is blank. They're complementary:

- `ui:initialValue` pre-fills the field on render and reset, keeping it from being blank
- `ui:emptyValue` is stored in `formData` whenever the field is blank, whether from initial state, user clearing, or reset

This means `ui:emptyValue` would be extended from its current scope (only kicks in when a user clears a widget) to kick in whenever a field is blank, regardless of why.

**`ui:required`**: Overrides the schema's `required` for a field, but only on the UI side.

- `true`: makes a non-required field required (validation + asterisk)
- `false`: hides the required asterisk on a schema-required field

Importantly, `ui:required` only affects the form-level required check and the UI indicator. It does not change the JSON Schema itself, so schema-level validation still runs as-is. If the schema says a field is required and it's left empty, validation will still fail regardless of `ui:required: false`. The intended use for `false` is when the field is guaranteed to have a value through `initialValue` or `emptyValue`.

```js
const uiSchema = {
  country: {
    "ui:widget": "hidden",
    "ui:initialValue": "US",
    "ui:required": false,  // hidden + pre-filled, no need to show as required
  },
  nickname: {
    "ui:required": true,   // required in this form, even if the schema doesn't say so
  },
};
```

Worth noting: `ui:required: false` on a schema-required field *without* `initialValue` or `emptyValue` is misleading, since the UI says "optional" but validation still fails. A `console.warn` for this would be nice, similar to the existing warning for deprecated `ui:widget` object syntax.

`ui:emptyValue` partly covers the defaults use case, but only on the data side. It doesn't pre-fill the field for the user and doesn't help with overriding required.

adamrimon:

Fixes #4983

`ui:emptyValue` only applies when the user actively clears a field. If the user never touches it, or the form is reset, there is no equivalent. This is a gap when the JSON schema is auto-generated and a separate uiSchema layer controls the UX.

This PR:
- Extends `ui:emptyValue` to apply whenever a field is blank, regardless of the reason (initial render, reset, or user clearing), by threading `uiSchema` through the [`getDefaultFormState`](packages/utils/src/schema/getDefaultFormState.ts) pipeline
- Adds `ui:initialValue` to pre-fill a field on render/reset, taking priority over `schema.default` (useful for hidden fields that need a fixed value)
- Adds `ui:required` to override the schema's required status for a field. `true` adds the indicator and validation, `false` hides the indicator but does not suppress schema validation. Emits `console.warn` for misleading configurations

The `ui:emptyValue` extension is a breaking change (previously only fired on explicit clear), hence the major version bump to **7.0.0**.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,23 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 7.0.0
+
+## @rjsf/utils
+
+- Extended `ui:emptyValue` to apply whenever a field is blank (initial render, reset, or user clearing), not just on widget clear ([#4983](https://github.com/rjsf-team/react-jsonschema-form/issues/4983))
+- Added `ui:initialValue` to pre-fill fields on render/reset, taking priority over `schema.default` ([#4983](https://github.com/rjsf-team/react-jsonschema-form/issues/4983))
+- Added `ui:required` to override schema required status from uiSchema ([#4983](https://github.com/rjsf-team/react-jsonschema-form/issues/4983))
+
+## @rjsf/core
+
+- Updated `Form` to pass `uiSchema` through `getDefaultFormState` for `ui:emptyValue` and `ui:initialValue` support ([#4983](https://github.com/rjsf-team/react-jsonschema-form/issues/4983))
+- Updated `SchemaField` and `LayoutGridField` to respect `ui:required` override ([#4983](https://github.com/rjsf-team/react-jsonschema-form/issues/4983))
+
+## Dev / docs / playground
+
+- Added documentation and playground examples for `ui:emptyValue` (extended), `ui:initialValue`, and `ui:required` ([#4983](https://github.com/rjsf-team/react-jsonschema-form/issues/4983))
+
 # 6.4.2
 
 ## @rjsf/core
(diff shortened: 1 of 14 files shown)


heath-freenome:
@adamrimon Wow, making changes for v7 already. We have a few others queuing up for it. I hope I have some time to prepare the long-lived branch for it

heath-freenome:
@adamrimon the long-lived v7 branch is available for you to work in now

heath-freenome:
Reimplemented in v7 via https://github.com/rjsf-team/react-jsonschema-form/pull/5277

PR #5000, opened 2026-03-24
feat: add support for optgroups in SelectWidget and related documentation updates

nlemoine:
### Reasons for making this change

Fixes #1813, #580 and closes #4374

Add optgroup support to RJSF. 

### Checklist

- [x] **I'm updating documentation**
  - [x] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [x] **I'm adding or updating code**
  - [x] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.
  - [x] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [x] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [x] **I'm adding a new feature**
  - [x] I've updated the playground with an example use of the feature

PR #5004, opened 2026-03-26
fix(SelectWidget): use real enum values for option values when nameGenerator is active

nlemoine:

When htmlName is present (nameGenerator enabled), SelectWidget rendered option values as array indices (0, 1, 2) instead of real enum values (e.g. "active", "inactive"). This broke native form submission since the POSTed value was the index, not the actual value. Fixes all 10 themes.

Related to `nameGenerator` introduced in #4773 

### Checklist


heath-freenome:
@nlemoine Help me understand the problem here. How does the name generator break things? The select widget converts to/from the index value upon selection. so the caller is still given the proper value.

nlemoine:
@heath-freenome Sorry, I should have included a [playground example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJyb2xlIjoiYWRtaW4iLCJmaXJzdE5hbWUiOiJDaHVjayIsImxhc3ROYW1lIjoiTm9ycmlzIiwiYWdlIjo3NSwiYmlvIjoiUm91bmRob3VzZSBraWNraW5nIGFzc2VzIHNpbmNlIDE5NDAiLCJwYXNzd29yZCI6Im5vbmVlZCIsInRlbGVwaG9uZSI6IjEtODAwLUtJQ0tBU1MifSwic2NoZW1hIjp7InRpdGxlIjoiU2VsZWN0ICsgbmFtZUdlbmVyYXRvciBidWciLCJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJzdGF0dXMiOnsidHlwZSI6InN0cmluZyIsInRpdGxlIjoiU3RhdHVzIiwiZW51bSI6WyJhY3RpdmUiLCJpbmFjdGl2ZSIsImFyY2hpdmVkIl19LCJwcmlvcml0eSI6eyJ0eXBlIjoibnVtYmVyIiwidGl0bGUiOiJQcmlvcml0eSIsImVudW0iOlsxLDIsM10sImVudW1OYW1lcyI6WyJMb3ciLCJNZWRpdW0iLCJIaWdoIl19LCJyb2xlIjp7InR5cGUiOiJzdHJpbmciLCJ0aXRsZSI6IlJvbGUiLCJvbmVPZiI6W3siY29uc3QiOiJhZG1pbiIsInRpdGxlIjoiQWRtaW5pc3RyYXRvciJ9LHsiY29uc3QiOiJlZGl0b3IiLCJ0aXRsZSI6IkVkaXRvciJ9LHsiY29uc3QiOiJ2aWV3ZXIiLCJ0aXRsZSI6IlZpZXdlciJ9XX19fSwidWlTY2hlbWEiOnsiZmlyc3ROYW1lIjp7InVpOmF1dG9mb2N1cyI6dHJ1ZSwidWk6ZW1wdHlWYWx1ZSI6IiIsInVpOnBsYWNlaG9sZGVyIjoidWk6ZW1wdHlWYWx1ZSBjYXVzZXMgdGhpcyBmaWVsZCB0byBhbHdheXMgYmUgdmFsaWQgZGVzcGl0ZSBiZWluZyByZXF1aXJlZCIsInVpOmF1dG9jb21wbGV0ZSI6ImZhbWlseS1uYW1lIiwidWk6ZW5hYmxlTWFya2Rvd25JbkRlc2NyaXB0aW9uIjp0cnVlLCJ1aTpkZXNjcmlwdGlvbiI6Ik1ha2UgdGV4dCAqKmJvbGQqKiBvciAqaXRhbGljKi4gVGFrZSBhIGxvb2sgYXQgb3RoZXIgb3B0aW9ucyBbaGVyZV0oaHR0cHM6Ly9tYXJrZG93bi10by1qc3gucXVhbnRpem9yLmRldi8pLiJ9LCJsYXN0TmFtZSI6eyJ1aTphdXRvY29tcGxldGUiOiJnaXZlbi1uYW1lIiwidWk6ZW5hYmxlTWFya2Rvd25JbkRlc2NyaXB0aW9uIjp0cnVlLCJ1aTpkZXNjcmlwdGlvbiI6Ik1ha2UgdGhpbmdzICoqYm9sZCoqIG9yICppdGFsaWMqLiBFbWJlZCBzbmlwcGV0cyBvZiBgY29kZWAuIDxzbWFsbD5OT1RFOiBVbnNhZmUgSFRNTCwgbm90IHJlbmRlcmVkPC9zbWFsbD4gIn0sImFnZSI6eyJ1aTp3aWRnZXQiOiJ1cGRvd24iLCJ1aTp0aXRsZSI6IkFnZSBvZiBwZXJzb24iLCJ1aTpkZXNjcmlwdGlvbiI6IihlYXJ0aCB5ZWFyKSJ9LCJiaW8iOnsidWk6d2lkZ2V0IjoidGV4dGFyZWEifSwicGFzc3dvcmQiOnsidWk6d2lkZ2V0IjoicGFzc3dvcmQiLCJ1aTpoZWxwIjoiSGludDogTWFrZSBpdCBzdHJvbmchIn0sInRlbGVwaG9uZSI6eyJ1aTpvcHRpb25zIjp7ImlucHV0VHlwZSI6InRlbCJ9fX0sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJjb25zdEFzRGVmYXVsdHMiOiJhbHdheXMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOmZhbHNlLCJpZFByZWZpeCI6InJvb3QiLCJpZFNlcGFyYXRvciI6Il8ifSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJTaW1wbGUifQ==).

See the option values:

```html
❌
<select id="root_status" name="root_status" role="combobox" class="form-control" aria-describedby="root_status__error root_status__description root_status__help">
  <option value=""></option>
  <option value="0">active</option>
  <option value="1">inactive</option>
  <option value="2">archived</option>
</select>
```

When submitting the traditional way (the whole PR I worked on), the submitted values are wrong. 

```html
✅
<select id="root_status" name="root_status" role="combobox" class="form-control" aria-describedby="root_status__error root_status__description root_status__help">
  <option value=""></option>
  <option value="active">active</option>
  <option value="inactive">inactive</option>
  <option value="archived">archived</option>
</select>
```

Let me know if you need more details.

heath-freenome:
> @heath-freenome Sorry, I should have included a [playground example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJyb2xlIjoiYWRtaW4iLCJmaXJzdE5hbWUiOiJDaHVjayIsImxhc3ROYW1lIjoiTm9ycmlzIiwiYWdlIjo3NSwiYmlvIjoiUm91bmRob3VzZSBraWNraW5nIGFzc2VzIHNpbmNlIDE5NDAiLCJwYXNzd29yZCI6Im5vbmVlZCIsInRlbGVwaG9uZSI6IjEtODAwLUtJQ0tBU1MifSwic2NoZW1hIjp7InRpdGxlIjoiU2VsZWN0ICsgbmFtZUdlbmVyYXRvciBidWciLCJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJzdGF0dXMiOnsidHlwZSI6InN0cmluZyIsInRpdGxlIjoiU3RhdHVzIiwiZW51bSI6WyJhY3RpdmUiLCJpbmFjdGl2ZSIsImFyY2hpdmVkIl19LCJwcmlvcml0eSI6eyJ0eXBlIjoibnVtYmVyIiwidGl0bGUiOiJQcmlvcml0eSIsImVudW0iOlsxLDIsM10sImVudW1OYW1lcyI6WyJMb3ciLCJNZWRpdW0iLCJIaWdoIl19LCJyb2xlIjp7InR5cGUiOiJzdHJpbmciLCJ0aXRsZSI6IlJvbGUiLCJvbmVPZiI6W3siY29uc3QiOiJhZG1pbiIsInRpdGxlIjoiQWRtaW5pc3RyYXRvciJ9LHsiY29uc3QiOiJlZGl0b3IiLCJ0aXRsZSI6IkVkaXRvciJ9LHsiY29uc3QiOiJ2aWV3ZXIiLCJ0aXRsZSI6IlZpZXdlciJ9XX19fSwidWlTY2hlbWEiOnsiZmlyc3ROYW1lIjp7InVpOmF1dG9mb2N1cyI6dHJ1ZSwidWk6ZW1wdHlWYWx1ZSI6IiIsInVpOnBsYWNlaG9sZGVyIjoidWk6ZW1wdHlWYWx1ZSBjYXVzZXMgdGhpcyBmaWVsZCB0byBhbHdheXMgYmUgdmFsaWQgZGVzcGl0ZSBiZWluZyByZXF1aXJlZCIsInVpOmF1dG9jb21wbGV0ZSI6ImZhbWlseS1uYW1lIiwidWk6ZW5hYmxlTWFya2Rvd25JbkRlc2NyaXB0aW9uIjp0cnVlLCJ1aTpkZXNjcmlwdGlvbiI6Ik1ha2UgdGV4dCAqKmJvbGQqKiBvciAqaXRhbGljKi4gVGFrZSBhIGxvb2sgYXQgb3RoZXIgb3B0aW9ucyBbaGVyZV0oaHR0cHM6Ly9tYXJrZG93bi10by1qc3gucXVhbnRpem9yLmRldi8pLiJ9LCJsYXN0TmFtZSI6eyJ1aTphdXRvY29tcGxldGUiOiJnaXZlbi1uYW1lIiwidWk6ZW5hYmxlTWFya2Rvd25JbkRlc2NyaXB0aW9uIjp0cnVlLCJ1aTpkZXNjcmlwdGlvbiI6Ik1ha2UgdGhpbmdzICoqYm9sZCoqIG9yICppdGFsaWMqLiBFbWJlZCBzbmlwcGV0cyBvZiBgY29kZWAuIDxzbWFsbD5OT1RFOiBVbnNhZmUgSFRNTCwgbm90IHJlbmRlcmVkPC9zbWFsbD4gIn0sImFnZSI6eyJ1aTp3aWRnZXQiOiJ1cGRvd24iLCJ1aTp0aXRsZSI6IkFnZSBvZiBwZXJzb24iLCJ1aTpkZXNjcmlwdGlvbiI6IihlYXJ0aCB5ZWFyKSJ9LCJiaW8iOnsidWk6d2lkZ2V0IjoidGV4dGFyZWEifSwicGFzc3dvcmQiOnsidWk6d2lkZ2V0IjoicGFzc3dvcmQiLCJ1aTpoZWxwIjoiSGludDogTWFrZSBpdCBzdHJvbmchIn0sInRlbGVwaG9uZSI6eyJ1aTpvcHRpb25zIjp7ImlucHV0VHlwZSI6InRlbCJ9fX0sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJjb25zdEFzRGVmYXVsdHMiOiJhbHdheXMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOmZhbHNlLCJpZFByZWZpeCI6InJvb3QiLCJpZFNlcGFyYXRvciI6Il8ifSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJTaW1wbGUifQ==).
> 
> See the option values:
> 
> ```
> ❌
> <select id="root_status" name="root_status" role="combobox" class="form-control" aria-describedby="root_status__error root_status__description root_status__help">
>   <option value=""></option>
>   <option value="0">active</option>
>   <option value="1">inactive</option>
>   <option value="2">archived</option>
> </select>
> ```
> 
> When submitting the traditional way (the whole PR I worked on), the submitted values are wrong.
> 
> ```
> ✅
> <select id="root_status" name="root_status" role="combobox" class="form-control" aria-describedby="root_status__error root_status__description root_status__help">
>   <option value=""></option>
>   <option value="active">active</option>
>   <option value="inactive">inactive</option>
>   <option value="archived">archived</option>
> </select>
> ```
> 
> Let me know if you need more details.

But the onChange handler should do the conversion so that the onSubmit will get the proper data in the `formData`. Or are you trying to work with pure HTML and avoiding the onSubmit callback?

nlemoine:
> Or are you trying to work with pure HTML and avoiding the onSubmit callback?

All the work done in #4773 has been made to that only purpose: be able to submit a form using standard HTML. I know people forget about it now everything is JS based :) 

In a classic form submission (no JS involved): no `formData`, no `onSubmit`, etc. Everything works pretty great right now. Except that one. Note that the current PR changes only affect `nameGenerator` path (e.g. no changes when no `nameGenerator` is `undefined`).

heath-freenome:
> > Or are you trying to work with pure HTML and avoiding the onSubmit callback?
> 
> All the work done in #4773 has been made to that only purpose: be able to submit a form using standard HTML. I know people forget about it now everything is JS based :)
> 
> In a classic form submission (no JS involved): no `formData`, no `onSubmit`, etc. Everything works pretty great right now. Except that one. Note that the current PR changes only affect `nameGenerator` path (e.g. no changes when no `nameGenerator` is `undefined`).

AHHH.... Got it. Let me take a closer look then :)

packages/chakra-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -42,21 +42,39 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
     uiSchema,
   } = props;
   const { enumOptions, enumDisabled, emptyValue } = options;
+  const useRealValues = !!htmlName;
 
   const _onMultiChange = ({ value }: SelectValueChangeDetails) => {
+    if (useRealValues) {
+      return onChange(value);
+    }
     return onChange(enumOptionsValueForIndex<S>(value, enumOptions, emptyValue));
```
heath-freenome: This is a cleaner approach with a bit more self-documentation
```suggestion
    const newValue = useRealValues ? value : enumOptionsValueForIndex<S>(value, enumOptions, emptyValue);
    return onChange(value);
```

packages/chakra-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -42,21 +42,39 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
...
   };
 
   const _onBlur = ({ target }: FocusEvent<HTMLInputElement>) =>
-    onBlur(id, enumOptionsValueForIndex<S>(target && target.value, enumOptions, emptyValue));
+    onBlur(
+      id,
+      useRealValues
+        ? target && target.value
+        : enumOptionsValueForIndex<S>(target && target.value, enumOptions, emptyValue),
+    );
```
heath-freenome: This is cleaner approach with a bit more self-documentation. Can you make similar changes in the rest of the PR?
```suggestion
    const newValue = useRealValues
        ? target && target.value
        : enumOptionsValueForIndex<S>(target && target.value, enumOptions, emptyValue)
    onBlur(id, newValue);
```

packages/chakra-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -80,30 +99,37 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
...
       : [];
 
-  const formValue = (isMultiple ? getMultiValue() : getSingleValue()).map((item) => item.value);
+  const formValue = useRealValues
+    ? typeof value !== 'undefined'
+      ? Array.isArray(value)
+        ? value.map(String)
+        : [String(value)]
+      : []
+    : (isMultiple ? getMultiValue() : getSingleValue()).map((item) => item.value);
```
heath-freenome: I'm not a fan of nested ternaries like that. Can you clean this up without them?

packages/core/src/components/widgets/SelectWidget.tsx
```diff
@@ -42,42 +42,63 @@ function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F extend
...
     },
-    [onChange, multiple, enumOptions, optEmptyVal],
+    [onChange, multiple, enumOptions, optEmptyVal, useRealValues],
   );
 
   const selectedIndexes = enumOptionsIndexForValue<S>(value, enumOptions, multiple);
+  const isEmpty = typeof value === 'undefined' || (multiple && value.length < 1) || (!multiple && value === emptyValue);
   const showPlaceholderOption = !multiple && schema.default === undefined;
 
+  const selectValue = useRealValues
```
heath-freenome: Ditto on nested ternaries

nlemoine:
Thanks for your feedback and comments @heath-freenome, I addressed them all. There's probably some room for DRYness here too. I'll handle all this in #5000

x0k:
Mapping via `String(value)` introduces ambiguity between `123` and `'123'`, and only supports primitive values.

If you use `JSON.stringify`, you will also need to handle parsing from `FormData`. In that case, it’s simpler to use the index and restore the selected value with `structuredClone(options[index])`.

Also, if someone already relies on the current behavior, this would be a breaking change.

heath-freenome:
@nlemoine Given that the name generator works across all themes, I'm wondering whether it makes sense to expand this set of changes across them all.  Also @x0k makes a few interesting points that I'm curious about your response to them

nlemoine:
Thanks for the feedback @x0k, that's a good catch and I reworked the approach.

Right now, `<option value>` stores the array index, handlers read it back and resolve the real typed value through `enumOptionsValueForIndex()`. That works fine on the JS side, but is meaningless for native form submission and [a bit unusual from a classic submission POV](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/select). That would mean resolving the index server side again.

I realized it not only affects `<select>` but every choices widget like radio, checkboxes, etc. that use `enumOptionsValueForIndex()` for value resolution. This is more related to web standards than `nameGenerator` itself, although they share the same purpose. 

Rather than overloading value for both jobs, I split them:

- value="real_value": the actual enum value as a string, so native form payload contains the actual value
- data-index="0": stores the array index for internal resolution

Handlers read data-index and call `enumOptionsValueForIndex()` same as before. So:

1. No String(value) ambiguity: resolution is still index-based
2. No FormData parsing: the index lives in data-index, the real value lives where the browser expects it
3. Not breaking: onChange/onBlur/onFocus return the same typed values as before

This restores `value` purpose instead of using it as an index store and defer index storage to `data-index` (which is fine reading from in JS context). Only changed core and react-bootstrap since they use native `<select>`. Other packages are untouched.

What do you think?

If you're ok with this, I can spread this approach to `RadioWidget` and `CheckboxesWidget`.

x0k:
> resolving the index server side

This is what I meant. With the current approach on the server:
- You won’t be able to distinguish primitive values from their string representations (`123` vs `'123'`)
- Arrays and objects will be sent as `[object Object]` ([example](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJsb2NhdGlvbiI6eyJuYW1lIjoiQW1zdGVyZGFtIiwibGF0Ijo1MiwibG9uIjo1fX0sInNjaGVtYSI6eyJkZWZpbml0aW9ucyI6eyJsb2NhdGlvbnMiOnsiZW51bSI6W3sibmFtZSI6Ik5ldyBZb3JrIiwibGF0Ijo0MCwibG9uIjo3NH0seyJuYW1lIjoiQW1zdGVyZGFtIiwibGF0Ijo1MiwibG9uIjo1fSx7Im5hbWUiOiJIb25nIEtvbmciLCJsYXQiOjIyLCJsb24iOjExNH1dfX0sInR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7ImxvY2F0aW9uIjp7InRpdGxlIjoiTG9jYXRpb24iLCIkcmVmIjoiIy9kZWZpbml0aW9ucy9sb2NhdGlvbnMifSwibG9jYXRpb25SYWRpbyI6eyJ0aXRsZSI6IkxvY2F0aW9uIFJhZGlvIiwiJHJlZiI6IiMvZGVmaW5pdGlvbnMvbG9jYXRpb25zIn0sIm11bHRpU2VsZWN0Ijp7InRpdGxlIjoiTG9jYXRpb25zIiwidHlwZSI6ImFycmF5IiwidW5pcXVlSXRlbXMiOnRydWUsIml0ZW1zIjp7IiRyZWYiOiIjL2RlZmluaXRpb25zL2xvY2F0aW9ucyJ9fSwiY2hlY2tib3hlcyI6eyJ0aXRsZSI6IkxvY2F0aW9ucyBDaGVja2JveGVzIiwidHlwZSI6ImFycmF5IiwidW5pcXVlSXRlbXMiOnRydWUsIml0ZW1zIjp7IiRyZWYiOiIjL2RlZmluaXRpb25zL2xvY2F0aW9ucyJ9fSwicmF0aW5nIjp7InRpdGxlIjoiUmF0aW5nIChtYXAtYmFzZWQgZW51bU5hbWVzKSIsInR5cGUiOiJudW1iZXIiLCJlbnVtIjpbMSwyLDMsNCw1XX0sInByaW9yaXR5Ijp7InRpdGxlIjoiUHJpb3JpdHkgKGVudW1PcmRlcikiLCJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyJsb3ciLCJtZWRpdW0iLCJoaWdoIiwiY3JpdGljYWwiXX19fSwidWlTY2hlbWEiOnsibG9jYXRpb24iOnsidWk6ZW51bU5hbWVzIjpbIk5ldyBZb3JrIiwiQW1zdGVyZGFtIiwiSG9uZyBLb25nIl19LCJsb2NhdGlvblJhZGlvIjp7InVpOndpZGdldCI6IlJhZGlvV2lkZ2V0IiwidWk6ZW51bU5hbWVzIjpbIk5ldyBZb3JrIiwiQW1zdGVyZGFtIiwiSG9uZyBLb25nIl19LCJtdWx0aVNlbGVjdCI6eyJpdGVtcyI6eyJ1aTplbnVtTmFtZXMiOlsiTmV3IFlvcmsiLCJBbXN0ZXJkYW0iLCJIb25nIEtvbmciXX19LCJjaGVja2JveGVzIjp7InVpOndpZGdldCI6IkNoZWNrYm94ZXNXaWRnZXQiLCJpdGVtcyI6eyJ1aTplbnVtTmFtZXMiOlsiTmV3IFlvcmsiLCJBbXN0ZXJkYW0iLCJIb25nIEtvbmciXX19LCJyYXRpbmciOnsidWk6ZW51bU5hbWVzIjp7IjEiOiJUZXJyaWJsZSIsIjIiOiJQb29yIiwiMyI6IkF2ZXJhZ2UiLCI0IjoiR29vZCIsIjUiOiJFeGNlbGxlbnQifX0sInByaW9yaXR5Ijp7InVpOmVudW1OYW1lcyI6eyJsb3ciOiJMb3cgcHJpb3JpdHkiLCJtZWRpdW0iOiJNZWRpdW0gcHJpb3JpdHkiLCJoaWdoIjoiSGlnaCBwcmlvcml0eSIsImNyaXRpY2FsIjoiQ3JpdGljYWwgcHJpb3JpdHkifSwidWk6ZW51bU9yZGVyIjpbImNyaXRpY2FsIiwiaGlnaCIsIioiXX19LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In0sInVzZUZhbGxiYWNrRmllbGQiOmZhbHNlLCJ1c2VGYWxsYmFja1VpRm9yVW5zdXBwb3J0ZWRUeXBlIjpmYWxzZSwibGl2ZVZhbGlkYXRlIjpmYWxzZSwiaWRQcmVmaXgiOiJyb290IiwiaWRTZXBhcmF0b3IiOiJfIn0sInZhbGlkYXRvciI6IkFKVjgiLCJzYW1wbGVOYW1lIjoiRW51bWVyYXRlZCBvYmplY3RzIn0=))

If this approach is acceptable for your project, implement it as a custom widget.

In my opinion, to properly support using RJSF with native forms, it’s better to approach this differently - namely, implement `FormData` parsing based on the schema, taking into account form submission specifics (empty text fields are not sent, only checked checkboxes are submitted with value `on`, files are sent even for empty inputs, ...) and RJSF-specific behavior (indexes instead of real values in enum fields, presence of a selector value for `oneOf/anyOf` fields, ability to rename keys when using `additionalProperties` and `patternProperties`, ...).

nlemoine:
For objects/arrays, native form submission relies on proper field naming conventions. That's what `nameGenerator` and #4773 are about. The structure is expressed through field names (location[name], location[lat]), not serialized into a single value attribute.

```html
<input type="checkbox" name="pets[]" value="dog" />
<input type="checkbox" name="pets[]" value="cat" />
```

Will generate a PHP array:
```php
$_POST['pets'] = ['cat', 'dog'];
```

For the 123 vs '123' distinction, this is how HTML forms have always worked. Every value hits the server as a string. The server parses based on the expected type. That's true for <input type="number"> too, it sends "123", not the  number 123.

The data-index approach in this PR keeps the JS-side typed resolution intact (no behavior change for existing RJSF users) while making the HTML output correct for standard form submission.

There's also an accessibility/UX angle. Browser autofill depends on `<option value>` being meaningful. A `<select autocomplete="country">` with `<option value="FR">France</option>` can be autofilled. With <option value="1">France</option>, the  browser can't match anything. Same for honorifics, card types, languages, etc. Index-based values break autofill.

x0k:
The example above with your changes - try modifying the `Location` and `Locations` values:

image


> this is how HTML forms have always worked

This is not the issue; the loss of type information is caused by the encoding method (`String(value)`). As mentioned earlier, you could use `JSON.stringify`, which would allow distinguishing values (`'123'` vs `'"123"'` on the server), but this approach requires decoding and is not significantly different from handling index-based values.

> no behavior change for existing RJSF users

This is a breaking change for those already using RJSF with native forms.  

>  Index-based values break autofill

This makes sense; perhaps this behavior should be configurable (via UI Schema / Form options)?

heath-freenome:
I would hope that people who are using the nameGenerator aren't dealing with complex objects like @x0k has in his example. @nlemoine I mostly agree with the approach you are taking and hope that the code can be a DRY as possible (maybe a new utility method or 2 would be helpful?). I would also love to have this solution working across all themes and places where we did the index replacement (to solve the issues with typing). And I am intrigued by the suggestion of adding a Form/UiSchema prop that controls whether JSON.stringify/parse is used to render/select values in the components. Maybe an optional prop on the `GlobalFormOptions` that controls this?

nlemoine:
Thanks for the clarification @x0k and the example. Sorry, I didn't spot the object enum right away. 

> The example above with your changes - try modifying the Location and Locations values

`<option value>` is a string attribute meant for short identifiers: country codes, honorific prefixes, rating values. Although not technically incorrect, It was never designed to hold serialized objects. If someone has object enums and needs native form submission, `nameGenerator` decomposes the structure into properly named fields, that's its purpose.

That said, you're right that the current index-based approach is at least harmless for object enums, while String(value) would produce "[object Object]". We could fall back to the index/stringyfied for non-primitive values to avoid that regression.

> the loss of type information is caused by the encoding method (String(value))

Every HTML form input sends strings. `<input type="number" value="123">` submits "123". Server frameworks parse it based on the expected type. `JSON.stringify` would work technically (`123` vs '"123"'), but no server framework expects that encoding out of the box. `String(value)` matches how every other HTML form element behaves.

> This is a breaking change for those already using RJSF with native forms.

Fair but honestly, very unlikely. 

> perhaps this behavior should be configurable

Open to that, though I wonder if it adds complexity for a niche case. The data-index attribute is always present, so anyone who needs index-based resolution can access it. Making real values the default aligns with how HTML forms work everywhere else.

----

To summarize:
- Meaningless values (browser autocomplete, a11y, etc.) is a real problem, I think we all agree about this?
- Changing `value` attribute is a breaking change for users that handle submission the native way
- Behavior should be optin

Proposed alternative:
- Something like `useRealValuesInChoices` (or whatever name) in `GlobalFormOptions` as proposed by @heath-freenome enable the native behavior
- When enabled: primitive values are printed as srings instead of index, more complex ones are stringyfied
- When enabled: typed values are still resolved through `data-index`, `formData` stays unchanged

What do you think? 

(@heath-freenome, I'll refactor the logic, just need feedback and approval from both of you about the behavior)

x0k:
Implementing this as an opt-in feature (with the option to switch to opt-out in v7) is a good solution.

> more complex ones are stringyfied

I don’t think this is a good idea, as it’s unclear how to handle this on the server. For an opt-in feature, `String(value)` is sufficient.

> in `GlobalFormOptions`

I think using the UI schema would result in a more flexible solution, as it would allow controlling behavior both for individual fields and for the entire form (`ui:globalOptions`).

> through `data-index`

This doesn’t seem like a universal solution. It’s easy to imagine UI libraries that don’t allow passing `data` attributes or fully handle option rendering themselves (e.g., `<Select options={options} />`). In my library, I introduced the following entity:

```ts
interface EnumValueMapper {
  fromValue: (value: JSONSchema7Type | undefined) => string;
  toValue: (value: string) => JSONSchema7Type | undefined;
}
```

 Using it, you can create options with the desired encoding (indices, `String(value)`, `JSON.stringify`, etc.) and retrieve the original value in `onChange`.

heath-freenome:
> Implementing this as an opt-in feature (with the option to switch to opt-out in v7) is a good solution.
> 
> > more complex ones are stringyfied
> 
> I don’t think this is a good idea, as it’s unclear how to handle this on the server. For an opt-in feature, `String(value)` is sufficient.
> 
> > in `GlobalFormOptions`
> 
> I think using the UI schema would result in a more flexible solution, as it would allow controlling behavior both for individual fields and for the entire form (`ui:globalOptions`).
> 
> > through `data-index`
> 
> This doesn’t seem like a universal solution. It’s easy to imagine UI libraries that don’t allow passing `data` attributes or fully handle option rendering themselves (e.g., `<Select options={options} />`). In my library, I introduced the following entity:
> 
> ```ts
> interface EnumValueMapper {
>   fromValue: (value: JSONSchema7Type | undefined) => string;
>   toValue: (value: string) => JSONSchema7Type | undefined;
> }
> ```
> 
> Using it, you can create options with the desired encoding (indices, `String(value)`, `JSON.stringify`, etc.) and retrieve the original value in `onChange`.

@x0k Can you point to how your library is defining and using it by providing links for us to checkout?

x0k:
- [EnumValueMapper definition](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/form/src/options.svelte.ts#L15)
- Using [createFormOptions](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/form/src/fields/enum.ts#L82), both the options and the `mapper` are created simultaneously in [enum](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/form/src/fields/extra/enum.svelte#L42) and [multiEnum](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/form/src/fields/extra/multi-enum.svelte#L47) fields
- Using the [singleOption](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/form/src/options.svelte.ts#L113) and [multipleOption](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/form/src/options.svelte.ts#L140) helpers, adapters are created to maintain two-way binding between the actual value and the widget value  
- [Usage in a widget](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/basic-theme/src/widgets/select.svelte#L45)
- The `mapper` is passed to widgets because in one of the themes, the [combobox widget](https://github.com/x0k/svelte-jsonschema-form/blob/e8619e4bf496be2c5bd1356e793dafdd6959400d/packages/skeleton4-theme/src/lib/extra-widgets/combobox.svelte#L34) is implemented based on a multi-select component

nlemoine:
Thanks for the links @x0k, very helpful to see how you approached this in your Svelte library.

Our current implementation follows a very similar pattern:

- `enumOptionValueEncoder(value, index, useRealValues)` → maps typed values to strings for DOM attributes (equivalent to your `fromValue`)
- `enumOptionValueDecoder(value, enumOptions, useRealValues, emptyValue)` → reverse lookup from string back to typed value (equivalent to your `toValue`)

Both are standalone utility functions in `@rjsf/utils`, used across all 10 theme packages in SelectWidget, RadioWidget, and CheckboxesWidget.

Configuration is via `ui:globalOptions` as you suggested, works both per-field and form-wide:

```js
// Form-wide
uiSchema: {
  "ui:globalOptions": { useRealOptionValues: true }
}

// Per-field
uiSchema: {
  myField: { "ui:options": { useRealOptionValues: true } }
}
```

When disabled (default): index-based encoding, fully backward compatible.
When enabled: `String(value)` for primitives, falls back to index for objects/arrays.

Decoder does a reverse lookup by matching `String(option.value)` against the DOM value, with index fallback. No `data-index` dependency, the resolution is self-contained in the decoder, so it works with any UI library regardless of attribute support.

All existing tests pass across all packages. Will push the updated branch shortly.

nlemoine:
Branch updated, I'll let you check it out and give me your feedback!

CHANGELOG.md
```diff
@@ -42,6 +42,11 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed a breaking change introduced in v6.4.0 where array properties with enum items no longer used `ui:enumNames` from the array property's uiSchema ([#4985](https://github.com/rjsf-team/react-jsonschema-form/issues/4985))
 - Passed `parentUiSchema` to `ArrayFieldItem` to maintain `uiSchema`, this prop was introduced recently but was not being passed
 - Fixed `extraErrors` not displaying when `customErrors` are also present (e.g., with array fields and controlled `formData`), fixing [#4982](https://github.com/rjsf-team/react-jsonschema-form/issues/4982)
+- Fixed `SelectWidget` rendering array indices instead of real enum values in `<option value>`, breaking native form submission. Uses `data-index` for internal typed value resolution ([#4693](https://github.com/rjsf-team/react-jsonschema-form/issues/4693))
+
+## @rjsf/react-bootstrap
+
+- Fixed `SelectWidget` rendering array indices instead of real enum values in `<option value>`, breaking native form submission. Uses `data-index` for internal typed value resolution ([#4693](https://github.com/rjsf-team/react-jsonschema-form/issues/4693))
```
heath-freenome: These belong under the 6.5.0 section above now.

packages/utils/src/enumOptionValueDecoder.ts
```diff
@@ -0,0 +1,55 @@
...
+  value: string,
+  enumOptions: EnumOptionsType<S>[] | undefined,
+  emptyValue?: unknown,
+): unknown {
+  if (value === '') {
+    return emptyValue;
+  }
+  if (!Array.isArray(enumOptions)) {
+    return emptyValue;
+  }
```
heath-freenome: You can combine these into a single condition

```suggestion
  if (value === '' || !Array.isArray(enumOptions)) {
    return emptyValue;
  }
```

packages/utils/src/enumOptionValueEncoder.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param value - The typed enum value
+ * @param index - The option's position in the enumOptions array
+ * @param useRealValues - Whether to use real values or indices
+ * @returns The string to use as the DOM value attribute
+ */
+export default function enumOptionValueEncoder(value: unknown, index: number, useRealValues: boolean): string {
+  if (!useRealValues) {
+    return String(index);
+  }
+  if (value === undefined || value === null) {
```
heath-freenome: You could used lodash's `isNil()` for a shorter condition

packages/utils/test/enumOptionValueDecoder.test.ts
```diff
@@ -0,0 +1,63 @@
+import enumOptionValueDecoder from '../src/enumOptionValueDecoder';
```
heath-freenome: This tests that the export is present in the main `index.ts`
```suggestion
import { enumOptionValueDecoder } from '../src';
```

packages/utils/test/enumOptionValueEncoder.test.ts
```diff
@@ -0,0 +1,39 @@
+import enumOptionValueEncoder from '../src/enumOptionValueEncoder';
```
heath-freenome: This tests that the export is present in the main `index.ts`

```suggestion
import { enumOptionValueEncoder } from '../src';
```

packages/utils/test/enumOptionValueDecoder.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    { value: 456, label: '456' },
+  ];
+  const booleanOptions: EnumOptionsType<RJSFSchema>[] = [
+    { value: true, label: 'Yes' },
+    { value: false, label: 'No' },
+  ];
+  const objectOptions: EnumOptionsType<RJSFSchema>[] = [
+    { value: { name: 'NY' } as unknown as RJSFSchema, label: 'New York' },
+    { value: { name: 'LA' } as unknown as RJSFSchema, label: 'Los Angeles' },
+  ];
```
heath-freenome: You could move these to be outside of the `describe()` since they are const across the tests.

packages/shadcn/src/SelectWidget/SelectWidget.tsx
```diff
@@ -62,9 +64,15 @@ export default function SelectWidget<
       {!multiple ? (
         <FancySelect
           items={items}
-          selected={enumOptionsIndexForValue<S>(value ?? defaultValue, enumOptions, false) as unknown as string}
+          selected={
+            useRealValues
+              ? typeof value === 'undefined'
+                ? ''
+                : String(value)
+              : (enumOptionsIndexForValue<S>(value ?? defaultValue, enumOptions, false) as unknown as string)
```
heath-freenome: Would the encoder be better here to avoid the nested ternaries?

packages/semantic-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -108,7 +112,17 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
       name={htmlName || id}
       label={labelValue(label || undefined, hideLabel, false)}
       multiple={typeof multiple === 'undefined' ? false : multiple}
-      value={typeof value === 'undefined' ? emptyValue : selectedIndexes}
+      value={
+        useRealValues
```
heath-freenome: Would the encoder be useful here? I'd love to avoid these nested ternaries

packages/react-bootstrap/src/SelectWidget/SelectWidget.tsx
```diff
@@ -48,11 +50,24 @@ export default function SelectWidget<
   const selectedIndexes = enumOptionsIndexForValue<S>(value, enumOptions, multiple);
   const showPlaceholderOption = !multiple && schema.default === undefined;
 
+  let selectValue;
+  if (useRealValues) {
```
heath-freenome: Hmmm, this seems like it similar to the nested ternaries below... Is there another helper function that can be built to consolidate this logic in one place?

packages/primereact/src/SelectWidget/SelectWidget.tsx
```diff
@@ -103,27 +114,39 @@ function MultiSelectWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
...
+      value={
+        useRealValues
+          ? isEmpty
+            ? emptyValue
+            : multiple
+              ? value.map(String)
+              : String(value)
+          : !isEmpty && typeof selectedIndexes !== 'undefined'
+            ? selectedIndexes
+            : emptyValue
```
heath-freenome: Encoder or is there another helper that can be created?

packages/primereact/src/SelectWidget/SelectWidget.tsx
```diff
@@ -70,10 +73,18 @@ function SingleSelectWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F
...
       {...primeProps}
-      value={!isEmpty && typeof selectedIndexes !== 'undefined' ? selectedIndexes : emptyValue}
+      value={
+        useRealValues
+          ? isEmpty
+            ? emptyValue
+            : String(value)
+          : !isEmpty && typeof selectedIndexes !== 'undefined'
+            ? selectedIndexes
+            : emptyValue
```
heath-freenome: Ditto on the encoder or another helper

packages/mui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -68,7 +70,17 @@ export default function SelectWidget<
...
+      value={
+        useRealValues
+          ? isEmpty
+            ? emptyValue
+            : multiple
+              ? value.map(String)
+              : String(value)
+          : !isEmpty && typeof selectedIndexes !== 'undefined'
+            ? selectedIndexes
+            : emptyValue
```
heath-freenome: Ditto on encoder or another helper

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
```diff
@@ -76,35 +78,43 @@ export default function SelectWidget<
...
-  const selectedValues = Array.isArray(selectedIndexes) ? selectedIndexes : [selectedIndexes];
+  const selectedValues = useRealValues
+    ? Array.isArray(value)
+      ? value.map(String)
+      : value != null
+        ? [String(value)]
+        : []
+    : Array.isArray(selectedIndexes)
+      ? selectedIndexes
+      : [selectedIndexes];
```
heath-freenome: Ditto on encoder or additional helper function

packages/core/src/components/widgets/SelectWidget.tsx
```diff
@@ -42,42 +43,56 @@ function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F extend
...
+      typeof value === 'undefined' || (multiple && value.length < 1) || (!multiple && value === emptyValue);
+    if (isEmpty) {
+      selectValue = emptyValue;
+    } else {
+      selectValue = multiple ? value.map(String) : String(value);
+    }
+  } else {
+    selectValue = typeof selectedIndexes === 'undefined' ? emptyValue : selectedIndexes;
+  }
+
```
heath-freenome: Ditto on encoder or another helper function

packages/chakra-ui/test/__snapshots__/Grid.test.tsx.snap
```diff
@@ -10208,7 +10199,6 @@ exports[`Complex grid renders person and address and employment in a complex gri
                                   id="root_employment-0"
                                 >
                                   <input
-                                    aria-labelledby="radio-group::r55::radio:label:0"
```
heath-freenome, line 10211: Looks like you lost the `aria-labelledby`... Can you restore it?

packages/chakra-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -103,7 +105,19 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
...
+    const isEmpty =
+      typeof value === 'undefined' || (isMultiple && value.length < 1) || (!isMultiple && value === emptyVal);
+    if (isEmpty) {
+      formValue = isMultiple ? [] : [''];
+    } else {
+      formValue = isMultiple ? value.map(String) : [String(value)];
+    }
+  } else {
+    formValue = (isMultiple ? getMultiValue() : getSingleValue()).map((item) => item.value);
+  }
```
heath-freenome: Ditto on encoder or additional helper function?

packages/mui/test/__snapshots__/Form.test.tsx.snap
```diff
@@ -34799,7 +34802,9 @@ exports[`single fields select field multiple choice enumDisabled using checkboxe
           >
             <span
               aria-describedby="root__error root__description root__help"
-              class="MuiButtonBase-root MuiCheckbox-root MuiCheckbox-colorPrimary MuiCheckbox-sizeMedium PrivateSwitchBase-root MuiCheckbox-root MuiCheckbox-colorPrimary MuiCheckbox-sizeMedium Mui-disabled MuiCheckbox-root MuiCheckbox-colorPrimary MuiCheckbox-sizeMedium emotion-4"
+              aria-disabled="true"
```
heath-freenome: What is making these aria-disabled suddenly?

heath-freenome (commented):
@nlemoine Great work in responding to the feedback and improving your code. I've made note of a few more places where you can DRY up repetitive code. Also some of the snapshots seems like things were incorrectly changed. Plus your tests are failing in several themes.

packages/core/src/components/widgets/SelectWidget.tsx
```diff
@@ -42,42 +43,56 @@ function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F extend
...
+      return onChange(enumOptionValueDecoder<S>(newValue, enumOptions, useRealValues, optEmptyVal));
     },
-    [onChange, multiple, enumOptions, optEmptyVal],
+    [onChange, multiple, enumOptions, optEmptyVal, useRealValues],
   );
 
   const selectedIndexes = enumOptionsIndexForValue<S>(value, enumOptions, multiple);
   const showPlaceholderOption = !multiple && schema.default === undefined;
 
+  let selectValue;
```
x0k: `enumOptionValueEncoder`, or an additional helper that abstracts away the encoder implementation details, should be used to determine the currently selected value; this applies to all select widgets.

packages/mui/test/__snapshots__/Form.test.tsx.snap
nlemoine: It's a unrelated update to my changes.

nlemoine:
This should be good now, CI is green.

packages/utils/src/types.ts
```diff
@@ -420,6 +420,11 @@ export type GlobalUISchemaOptions = GenericObjectType & {
    * both. To disable the Optional Data Field UI for a specific field, provide an empty array within the UI schema.
    */
   enableOptionalDataFieldForType?: ('object' | 'array')[];
+  /** When true, enum-backed widgets (select, radio, checkboxes) render real enum values
+   *  in their value attributes instead of array indices. This enables proper native form
+   *  submission and browser autocomplete. Default: false (index-based, current behavior).
+   */
+  useRealOptionValues?: boolean;
```
heath-freenome: Can you please add documentation for this new option before we merge. Also, I wonder if you changed this to: `optionValueFormat?: 'indexed' | 'realValue';` Going this route could support future enhancements like allowing people to provide their own transformer, etc... I know this is another change through the code and it will improve maintainability and  usability. We've been doing a lot of work in other place in RJSF to switch from boolean to "options" so let's be proactive here.

packages/utils/src/enumOptionValueEncoder.ts
```diff
@@ -0,0 +1,27 @@
...
+ * String() would produce "[object Object]".
+ *
+ * When `useRealValues` is false, returns the index as a string (current default behavior).
+ *
+ * @param value - The typed enum value
+ * @param index - The option's position in the enumOptions array
+ * @param useRealValues - Whether to use real values or indices
+ * @returns The string to use as the DOM value attribute
+ */
+export default function enumOptionValueEncoder(value: unknown, index: number, useRealValues: boolean): string {
```
heath-freenome: you could potentially pass in the `UiOptionsType` instead and read the property from it? OR, you could pass in the suggested `optionValueFormat` instead. If you go the `optionValueFormat` route, then make it optional with the default being `index`. If you go the `UiOptionsType` route, then you can add an internal helper function is used across all 3 functions to set the default to `index` in one place. This would change the generics to require T, S and F.

packages/utils/src/enumOptionValueDecoder.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param value - The string value(s) from the DOM
+ * @param enumOptions - The available enum options
+ * @param useRealValues - Whether real values or indices were used for encoding
+ * @param emptyValue - The value to return for empty/missing selections
+ * @returns The original typed enum value(s)
+ */
+export default function enumOptionValueDecoder<S extends StrictRJSFSchema = RJSFSchema>(
+  value: string | string[],
+  enumOptions: EnumOptionsType<S>[] | undefined,
+  useRealValues: boolean,
```
heath-freenome: you could potentially pass in the `UiOptionsType` instead and read the property from it? OR, you could pass in the suggested `optionValueFormat` instead. If you go the `optionValueFormat` route, then make it optional with the default being `index`. If you go the `UiOptionsType` route, then you can add an internal helper function is used across all 3 functions to set the default to `index` in one place. This would change the generics to require T, S and F.

packages/utils/src/enumOptionSelectedValue.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param multiple - Whether the select allows multiple selections
+ * @param useRealValues - Whether real values or indices are used for encoding
+ * @param emptyValue - The value to return when the selection is empty
+ * @returns The value to use for the select element's `value` attribute
+ */
+export default function enumOptionSelectedValue<S extends StrictRJSFSchema = RJSFSchema>(
+  value: any,
+  enumOptions: EnumOptionsType<S>[] | undefined,
+  multiple: boolean,
+  useRealValues: boolean,
```
heath-freenome: you could potentially pass in the `UiOptionsType` instead and read the property from it? OR, you could pass in the suggested `optionValueFormat` instead. If you go the `optionValueFormat` route, then make it optional with the default being `index`. If you go the `UiOptionsType` route, then you can add an internal helper function is used across all 3 functions to set the default to `index` in one place. This would change the generics to require T, S and F.

packages/chakra-ui/src/CheckboxesWidget/CheckboxesWidget.tsx
```diff
@@ -57,8 +59,8 @@ export default function CheckboxesWidget<
     >
       {!hideLabel && label && <FieldsetLegend>{labelValue(label)}</FieldsetLegend>}
       <CheckboxGroup
-        onValueChange={(option) => onChange(enumOptionsValueForIndex<S>(option, enumOptions, emptyValue))}
-        value={selectedIndexes}
+        onValueChange={(option) => onChange(enumOptionValueDecoder<S>(option, enumOptions, useRealValues, emptyValue))}
+        value={useRealValues ? (Array.isArray(value) ? value.map(String) : []) : selectedIndexes}
```
heath-freenome: Would calling the `enumOptionSelectedValue()` to get the value for the `CheckboxGroup` work?

packages/chakra-ui/src/RadioWidget/RadioWidget.tsx
```diff
@@ -58,7 +60,7 @@ export default function RadioWidget<T = any, S extends StrictRJSFSchema = RJSFSc
         onChange={_onChange}
         onBlur={_onBlur}
         onFocus={_onFocus}
-        value={selectedIndex}
+        value={useRealValues ? (value != null ? String(value) : null) : selectedIndex}
```
heath-freenome: Ditto comment about using `enumOptionSelectedValue()` here?

packages/chakra-ui/src/SelectWidget/SelectWidget.tsx
heath-freenome: Like:  using `enumOptionSelectedValue()`

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
```diff
@@ -76,35 +78,38 @@ export default function SelectWidget<
...
+    [onFocus, id, enumOptions, optEmptyVal, useRealValues],
   );
 
-  const selectedIndexes = enumOptionsIndexForValue<S>(value, enumOptions, multiple);
-  const selectedValues = Array.isArray(selectedIndexes) ? selectedIndexes : [selectedIndexes];
+  const selectedValues: string[] = [
+    enumOptionSelectedValue<S>(value, enumOptions, !!multiple, useRealValues, multiple ? [] : ''),
+  ]
+    .flat()
+    .filter((v) => v !== '');
```
heath-freenome, line 116: Why are these needed? Also if they are, does it make sense to put them into the `enumOptionSelectedValue()` function?

packages/mantine/src/widgets/SelectWidget.tsx
```diff
@@ -92,7 +92,7 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
       name={htmlName || id}
       label={labelValue(label || undefined, hideLabel, false)}
       data={selectOptions}
-      value={multiple ? (selectedIndexes as any) : (selectedIndexes as string)}
+      value={enumOptionSelectedValue<S>(value, enumOptions, !!multiple, useRealValues, multiple ? [] : null) as any}
```
heath-freenome: I'm curious about the passing in of [] vs null here along the cast... Can that be fixed to make the `enumOptionSelectedValue()` use the proper default value?

packages/mui/src/RadioWidget/RadioWidget.tsx
```diff
@@ -56,7 +59,7 @@ export default function RadioWidget<T = any, S extends StrictRJSFSchema = RJSFSc
       <RadioGroup
         id={id}
         name={htmlName || id}
-        value={selectedIndex}
+        value={useRealValues ? (value !== undefined ? String(value) : '') : selectedIndex}
```
heath-freenome: Could you use `enumOptionSelectedValue()` here instead?

heath-freenome (commented):
@nlemoine Great work here. Thanks for being patient with all the suggested changes. I have a few more suggestions to help maintainability.

Also, it also looks like you skipped over converting the `antd` theme?

heath-freenome:
@nlemoine I hope I haven't overloaded you with too many comments

nlemoine:
@heath-freenome It's ok, don't worry about it. Just didn't have time to work on this recently. I'll be able to update the PR within the day.

heath-freenome:
> @heath-freenome It's ok, don't worry about it. Just didn't have time to work on this recently. I'll be able to update the PR within the day.

@nlemoine I totally understand how life gets in the way. I am waiting for this PR to be finished and merged so that I can release 6.5.0. You also have a few conflicts in the CHANGELOG.md to resolve. I hope you have some time this weekend to get it over the line. Thanks again for all your hard work!

packages/utils/src/types.ts
heath-freenome: @nlemoine This is one change that I'd love for you to make. If you feel like you don't have the time, I could also do it myself after you are done.

packages/utils/src/enumOptionValueEncoder.ts
nlemoine: Thanks. I went a slightly different route than your two options. 

I added a small `getOptionValueFormat(options)` helper in `@rjsf/utils` that returns `options.optionValueFormat ?? 'indexed'`. Widgets call it once:

```js
const optionValueFormat = getOptionValueFormat(options);
```

and pass the result to `enumOptionValueEncoder`, `enumOptionValueDecoder`, and `enumOptionSelectedValue`. The default lives in one place.

I didn't want to pass `UIOptionsType` into the three utilities directly because it felt like a layer violation.
They're low-level serialization helpers. They shouldn't need to know about UI schema conventions, or carry `T/S/F` generics, just to read one property.

On the format: `OptionValueFormat = 'indexed'` default param, I did that too. The three utilities already take format with 'indexed' as the default, so they work even if you call them without reading options. The separate helper exists for the widgets, otherwise each one still writes `options.optionValueFormat ?? 'indexed'` inline and the duplication comes back.

If you'd rather have the utilities take `UIOptionsType` directly, that's fine and I'll change it.

nlemoine:
> https://github.com/rjsf-team/react-jsonschema-form/pull/5004#discussion_r3034133492

Done. Both paths go through one `enumOptionSelectedValue` call now. While I was in there I noticed the old indexed path was building {label, value} objects in `getMultiValue`/`getSingleValue` and then throwing the labels away on the next line with `.map(item => item.value)`, so `valueLabelMap` in the `useMemo` wasn't actually used anywhere. I dropped that too. The `formValue` block and the helpers it fed went from about 28 lines down to 5, and the tests still pass without any snapshot updates.

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
nlemoine: Left it in place and added a comment on both call sites explaining why it's there. I did think about pulling it into a plural helper, but only 2 of ~13 callers need "always `string[]`". The other 11 pass the result straight into a component that accepts `string | string[]` based on `multiple`, so switching the singular helper's shape would just push unwraps out into all the other widgets, and a new utility for 2 callers felt worse than the 3 lines of duplication. The reason chakra and daisyui are the odd ones out: both use custom dropdowns that iterate with `selectedValues.includes(...)`, so they always want a flat string array, and the `.flat().filter(v => v !== '')` is what normalizes the helper's output into that shape.

packages/mantine/src/widgets/SelectWidget.tsx
nlemoine: The helper itself can't have one "proper default" because different consumers expect different empty values. Mantine's Select wants `null`, MultiSelect wants `[]`, the HTML `<select>` wants `''`, and chakra wants `['']`. Each caller knows what its component expects, so passing `emptyValue` explicitly is the right interface.

The `as any` cast you noticed was caused by the `Component = multiple ? MultiSelect : Select` pattern above it, not by the helper itself. TypeScript intersects the two components's prop types, which makes `value` effectively `never`, so anything assigned to it needs an escape hatch to compile.

I fixed it by splitting into two JSX branches. A `sharedProps` object holds everything common, and each branch passes its own narrow `value` (`string[]` for MultiSelect, `string | null` for Select). That removes the cast, and the `multiple ? [] : null` ternary goes with it since each branch just passes the literal its component expects.

nlemoine:
@heath-freenome I addressed all comments from your review. With slightly different approaches on some of them. I posted the details to show the reasoning behind decisions.

Supporting so much UI libs, with all their distinctive implementation details, is quite a challenge 😅 

Let me know if you spot something in the updated code.

packages/mantine/src/widgets/SelectWidget.tsx
```diff
@@ -41,70 +43,77 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
...
+    <MultiSelect
+      {...sharedProps}
+      value={enumOptionSelectedValue<S>(value, enumOptions, true, optionValueFormat, []) as string[]}
+      onChange={!readonly ? handleChange : undefined}
+    />
+  ) : (
+    <Select
+      {...sharedProps}
+      value={enumOptionSelectedValue<S>(value, enumOptions, false, optionValueFormat, null) as string | null}
       onChange={!readonly ? handleChange : undefined}
```
heath-freenome: I'm curious if there was a reason the `onChange()` handler is not in the `sharedProps`?

packages/semantic-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -28,13 +30,14 @@ function createDefaultValueOptionsForDropDown<S extends StrictRJSFSchema = RJSFS
   enumDisabled?: UIOptionsType['enumDisabled'],
   showPlaceholderOption?: boolean,
   placeholder?: string,
+  format: 'indexed' | 'realValue' = 'indexed',
```
heath-freenome: Does this work instead?
```suggestion
  format: OptionValueFormat = 'indexed',
```

packages/utils/src/enumOptionValueDecoder.ts
```diff
@@ -0,0 +1,52 @@
+import { EnumOptionsType, OptionValueFormat, StrictRJSFSchema, RJSFSchema } from './types';
+import enumOptionsValueForIndex from './enumOptionsValueForIndex';
+
+function decodeSingle<S extends StrictRJSFSchema = RJSFSchema>(
```
heath-freenome, line 16: Consider adding some JS Doc onto this function

heath-freenome (commented):
@nlemoine Nice work! I have a couple of questions/suggestions for you. Plus there are conflicts in the `CHANGELOG.md` again (sorry, merging other fixes).

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
heath-freenome: I'm ok with that for now. If a new theme comes in that causes a 3rd use case, then I will try to remember to have them DRY that up

heath-freenome:
@nlemoine Also, a new conflict in the MUI select widget due to another PR merging

heath-freenome:
@nlemoine Just a few conflict resolutions and a small change and this PR is good to go. How can I help?

packages/mantine/src/widgets/SelectWidget.tsx
nlemoine: Yeah, no real reason. I just missed it when I split the JSX. `onChange` has the same `!readonly ? handleChange : undefined` gate as `onBlur` and `onFocus`, so it belongs in `sharedProps` with them. Moved it in.

nlemoine:
@heath-freenome Should hopefully be all good now :) 

I'll be off next week and won't be able to make any more changes until April 27 at best. Feel free to take over if changes are needed and the PR blocks next release.

heath-freenome (approved):

PR #5031, opened 2026-04-11
fix(core): use ref to prevent stale formData in rapid additional property renames

restareaByWeezy:

Fixes #5021

When two additional property keys are renamed in quick succession (before React flushes the batched state update from the first rename), the second rename reads stale `formData` from the `useCallback` closure. This causes the first rename's result to be silently dropped.

The fix introduces a `useRef` to track the latest formData. After computing the renamed object, the ref is updated immediately before calling `onChange`, so any subsequent rename within the same React batch reads the up-to-date data. The ref is synced with the authoritative `formData` prop on every render.

Changes:
- `packages/core/src/components/fields/ObjectField.tsx`: read `formDataRef.current` instead of the closure-captured `formData` in `handleKeyRename`, and update the ref immediately after computing the new form data
- `packages/core/test/ObjectField.test.tsx`: add test for rapid successive key renames within a single `act()` boundary

Reproduction steps (on the playground's Additional Properties sample):
1. Click "Add Item" to create a `newKey` property
2. Rename `newKey` to `second`, press Tab
3. Immediately click the `assKickCount` key input, rename it to `renamed`, press Tab
4. Before fix: `second` disappears. After fix: both `second` and `renamed` are present.

### Checklist


heath-freenome (approved):
@restareaByWeezy Thanks for the fix!

PR #5032, opened 2026-04-11
Feature: validator-ajv8: Add support for suppressing duplicate filtering

### LINKED ISSUE #5028: v6 upgrade: Anyway to get the non-deduplicated errors?
### Prerequisites

### What theme are you using?

other

### What is your question?

I'm currently upgrading from v5 to v6 and noticed this change in behaviour.

I'm seeing an issue due to PR #4884, which added error deduplication.

I have an implementation that relies on the fact that each element receives `rawErrors`. When they receive `rawErrors`, they render in an error state. 

v6 screenshot (not expected):

Image

v5 screenshot (expected):

Image

In the v6 screenshot you can see `allOf` error ("Select at least one sorting order option.") and the first checkbox in an error state (it receives `rawErrors` `{ name: "const", property: ".listingFields.0.filterConfig.sortingOrder.asc", message: "must be equal to constant", … }`).

The second is not in an error state because it no longer receives `rawErrors` due to deduplication.

Here's my schema:

```json
{
  "title": "Sorting order",
  "description": "Choose how users can sort listings by this field.",
  "type": "object",
  "properties": {
    "asc": {
      "title": "Low to high",
      "type": "boolean",
      "default": true
    },
    "desc": {
      "title": "High to low",
      "type": "boolean",
      "default": true
    }
  },
  "anyOf": [
    {
      "required": [
        "asc"
      ],
      "properties": {
        "asc": {
          "const": true
        }
      }
    },
    {
      "required": [
        "desc"
      ],
      "properties": {
        "desc": {
          "const": true
        }
      }
    }
  ]
}
```

I'm asking for advice: How should I fix this issue? I have a couple of options in my mind.

Option 1: Use `has-error` CSS class. Possible, but my components don't actually use CSS this was and they assume that I pass a `state: "error"` property on them. In addition, while this could solve my use-case here, it wouldn't solve the issue in case I've wanted to e.g. print an error message after the checkbox, which would require information about the error (the `rawErrors` passed to each checkbox)

Option 2: Use `formContext` to pass the information from the Object field down to the checkbox component. Possible, but feels unnecessarily complicated.

Option 3: Ideal for me would be the revert the old behaviour or make it possible to override the current default behaviour with my own which wouldn't do deduplication.

I'm currently leaning towards 2, but just wanted to ask here if anyone has any simpler ideas. Thanks!

heath-freenome:

Fixes #5028 by adding configuration support to the `validator-ajv8` to suppress duplicate filtering
- Updated `CustomValidatorOptionsType` to add new `suppressDuplicateFiltering` flag prop
- Updated `AJV8Validator`, `AJV8PrecompiledValidator` and `createPrecompiledValidator()` to support passing the new flag prop
- Updated `processRawValidationErrors()` to pass the new flag prop into `transformRJSFValidationErrors()`
- Refactored the filtering of duplicates to a new `filterDuplicateErrors()` function that takes the new flag prop and suppresses the appropriate filtering
- Updated the unit tests to verify the flag works as intended
- Updated the `validator-ajv8.md` and `validation.md` docs for the new flag prop

### Checklist


packages/docs/docs/usage/validation.md
```diff
@@ -707,6 +707,51 @@ const validator = customizeValidator({ extenderFn: ajvErrors });
...
+import { customizeValidator } from '@rjsf/validator-ajv8';
+
+const schema: RJSFSchema = {
+  type: 'object',
+  properties: {
+    value: {
+      anyOf: [{ type: 'string' }, { type: 'number' }],
+    },
+  },
+};
```
nickgros, line 764: May want to use an example that exemplifies the use case in #5028 , i.e. the `allOf` is used to create validation errors that should be distributed to all affected properties.

packages/validator-ajv8/src/types.ts
```diff
@@ -18,6 +18,13 @@ export interface CustomValidatorOptionsType {
   AjvClass?: typeof Ajv;
   /** A function to call to extend AJV, such as `ajvErrors()` */
   extenderFn?: (ajv: Ajv) => Ajv;
+  /** When set, suppresses duplicate error filtering for the specified keyword(s):
+   * - `'all'`: disables all duplicate filtering
+   * - `'anyOf'`: disables filtering for `anyOf` errors only (oneOf duplicates are still filtered)
+   * - `'oneOf'`: disables filtering for `oneOf` errors only (anyOf duplicates are still filtered)
+   * When omitted, both `anyOf` and `oneOf` duplicate errors are filtered (default behavior).
+   */
+  suppressDuplicateFiltering?: 'anyOf' | 'oneOf' | 'all';
```
nickgros: WDYT about a dedicated `'none'` value for the default?

heath-freenome: I like it

packages/docs/docs/usage/validation.md
heath-freenome: Done. I just used the schema from the issue :)

nickgros (approved):

PR #5044, opened 2026-04-24
fix(@rjsf/core): prevent extraErrors duplication on array field mutation (#5041)

### LINKED ISSUE #5041: Extra errors duplicated when a sibling array field is mutated
### Prerequisites

### What theme are you using?

react-bootstrap

### Version

6.5.1

### Current Behavior

When extraErrors are set on a field outside an array, any array mutation (add / remove / reorder) causes the sibling field's __errors entries to be concatenated to themselves in errorSchema. Each interaction adds another copy, so the same error renders N+1 times after N clicks.

### Expected Behavior

extraErrors should stay stable across unrelated array mutations. A sibling field's error count should remain at 1 no matter how many times the user adds, removes, or reorders array items.

### Steps To Reproduce

https://codesandbox.io/p/sandbox/optimistic-torvalds-6f6yj6

* In example sandbox, name field starts with an error from extra errors
* Change the array field, add, remove, move an item
* The name error gets duplicated

### Environment

```markdown
- OS: MacOS 26.2 (25C56)
- Node: v20.19.5
- npm: 11.12.1
```

### Anything else?

The following analysis was generated with Opus 4.7. I took a look over it's findings and they look legitimate.

## Root cause

  The bug emerges from a three-step interaction across `SchemaField`, `ArrayField`, and `Form.processPendingChange`.

  ### 1. `SchemaField` coerces undefined `errorSchema` to an empty object

  `packages/core/src/components/fields/SchemaField.tsx` (line 181 in v6.5.1):

  ```js
  const { __errors, ...fieldErrorSchema } = errorSchema || {};
  ```

  When the parent's `errorSchema[childKey]` is `undefined` (the array has no errors of its own), this defaults to `{}` and the rest-spread passes a **truthy empty object** down to the child `ArrayField`.

  ### 2. `ArrayField` treats that empty object as "has errors" and emits a truthy `newErrorSchema`

  `packages/core/src/components/fields/ArrayField.tsx`, `handleAddItem` (and the sibling `handleCopyItem` / `handleRemoveItem` / `handleReorderClick`):

  ```js
  let newErrorSchema;
  if (errorSchema) {                  // {} is truthy → enters branch
    newErrorSchema = {};              // builds an empty re-indexed errorSchema
    for (const idx in errorSchema) { /* no-op on empty object */ }
  }
  onChange(newFormData, path, newErrorSchema);  // emits truthy {} newErrorSchema
  ```

  ### 3. `Form.processPendingChange` re-merges `extraErrors` into state that already contains them

  `packages/core/src/components/Form.tsx`, `processPendingChange` (around line 997 in v6.5.1):

  ```js
  else if (!noValidate && newErrorSchema) {       // {} is truthy → enters
    const mergedErrors = this.mergeErrors(
      { errorSchema: originalErrorSchema, errors },
      extraErrors,
      customErrors,
    );
    state = { formData: newFormData, ...mergedErrors, customErrors };
  }
  ```

  `originalErrorSchema = this.state.errorSchema` already contains the prior `extraErrors` merge applied on mount (and/or via `getDerivedStateFromProps`). `mergeErrors` → `validationDataMerge` → `mergeObjects(..., concatArrays=true)`
  then concatenates the `__errors` arrays for any field present in both the state and `extraErrors`. Identical `__errors` strings accumulate on every array onChange.

  ### Why it's silent without the array interaction

  The `extraErrors` merge is gated on `newErrorSchema` being truthy. With no array interaction, `newErrorSchema` never arrives from a field, and the duplication path is never entered — the error appears exactly once.

  ## Suggested fixes

  Any of the following would resolve the duplication; (c) is the smallest behavior change and the easiest to reason about:

  - **(a)** In `processPendingChange`, base the merge on `state.schemaValidationErrorSchema` (the un-`extraErrors`-merged base) instead of `state.errorSchema`, mirroring the approach used in `getStateFromProps` (around line 240).
  - **(b)** In `mergeErrors` / `validationDataMerge`, pass `'preventDuplicates'` so identical `__errors` strings don't accumulate.
  - **(c)** In `ArrayField.handleAddItem` and its siblings, only build `newErrorSchema` if the incoming `errorSchema` actually has keys (`Object.keys(errorSchema).length > 0`), so the empty-object default from `SchemaField` doesn't
  propagate as a truthy `newErrorSchema` and the buggy merge path is never entered.

JosephDoUrden:

`processPendingChange()` used `this.state.errorSchema` (which already contains merged `extraErrors`) as the base when re-merging `extraErrors` in the non-live-validate branch. Because `validationDataMerge` concatenates `__errors` arrays, each array mutation triggered an extra merge cycle that appended duplicate copies of any sibling field's `extraErrors`.

The fix uses `schemaValidationErrorSchema` (the un-merged AJV-only base) instead, which is consistent with how `getDerivedStateFromProps` and `getStateFromProps` handle the same merge pattern.

Fixes #5041

### Changes

- In `Form.processPendingChange()`, replaced `originalErrorSchema` with `schemaValidationErrorSchema` as the base passed to `mergeErrors()` in the `else if (!noValidate && newErrorSchema)` branch
- Added regression test: sibling field `extraErrors` must not accumulate after array item is added

### Checklist

### DIFF
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -896,7 +896,11 @@ export default class Form<
     const { extraErrors, omitExtraData, liveOmit, noValidate, liveValidate, onChange, removeEmptyOptionalObjects } =
       this.props;
     const { formData: oldFormData, schemaUtils, schema, fieldPathId, schemaValidationErrorSchema, errors } = this.state;
-    let { customErrors, errorSchema: originalErrorSchema } = this.state;
+    let { customErrors } = this.state;
+    // Use the un-merged AJV-only schema as the base for re-merging extraErrors. Mirrors the
+    // pattern in getStateFromProps/getDerivedStateFromProps and avoids the duplication that
+    // happened when state.errorSchema (already containing merged extraErrors) was passed in.
+    let mergeBaseErrorSchema: ErrorSchema<T> = schemaValidationErrorSchema as ErrorSchema<T>;
     const rootPathId = fieldPathId.path[0] || '';
 
     const isRootPath = !path || path.length === 0 || (path.length === 1 && path[0] === rootPathId);
@@ -958,11 +962,13 @@ export default class Form<
       const oldValidationError = !isRootPath ? _get(schemaValidationErrorSchema, path) : schemaValidationErrorSchema;
       // If there is an old validation error for this path, assume we are updating it directly
       if (!_isEmpty(oldValidationError)) {
-        // Update the originalErrorSchema "in place" or replace it if it is the root
+        // Apply the user-supplied newErrorSchema onto a clone of the AJV-only base, so that
+        // mergeErrors below sees the user's error at this path without mutating shared state.
         if (!isRootPath) {
-          _set(originalErrorSchema, path, newErrorSchema);
+          mergeBaseErrorSchema = _cloneDeep(schemaValidationErrorSchema) as ErrorSchema<T>;
+          _set(mergeBaseErrorSchema, path, newErrorSchema);
         } else {
-          originalErrorSchema = newErrorSchema;
+          mergeBaseErrorSchema = newErrorSchema as ErrorSchema<T>;
         }
       } else {
         if (!customErrors) {
@@ -987,7 +993,7 @@ export default class Form<
       const liveValidation = this.liveValidate(
         schema,
         schemaUtils,
-        originalErrorSchema,
+        mergeBaseErrorSchema,
         newFormData,
         extraErrors,
         customErrors,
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Form.tsx
```diff
@@ -996,7 +996,11 @@ export default class Form<
       state = { formData: newFormData, ...liveValidation, customErrors };
     } else if (!noValidate && newErrorSchema) {
       // Merging 'newErrorSchema' into 'errorSchema' to display the custom raised errors.
-      const mergedErrors = this.mergeErrors({ errorSchema: originalErrorSchema, errors }, extraErrors, customErrors);
+      const mergedErrors = this.mergeErrors(
+        { errorSchema: schemaValidationErrorSchema as ErrorSchema<T>, errors },
```
heath-freenome: I believe you may also want to update line 990 as well?

heath-freenome:
@JosephDoUrden Looks like your tests are failing

heath-freenome (approved):

Issue #4475, opened 2025-01-31
Custom Validation started to fail with upgrade to version 5.24.2


rgciocan:
### Prerequisites

- [x] I have searched the existing issues
- [x] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [ ] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

validator-ajv8

### Version

5.24.2

### Current Behavior

Using custom validation examples from the documentation, when rendering the schema I get the browser console error:

Maximum call stack size exceeded
RangeError: Maximum call stack size exceeded
    at areObjectsEqual (http://localhost:3000/bundle.js:550581:25)
    at comparator (http://localhost:3000/bundle.js:550840:17)
    at Object.equals (http://localhost:3000/bundle.js:550928:16)
    at areArraysEqual (http://localhost:3000/bundle.js:550500:20)
    at comparator (http://localhost:3000/bundle.js:550793:20)
    at Object.equals (http://localhost:3000/bundle.js:550928:16)
    at isPropertyEqual (http://localhost:3000/bundle.js:550718:15)
    at areObjectsEqual (http://localhost:3000/bundle.js:550592:14)
    at comparator (http://localhost:3000/bundle.js:550840:17)
    at Object.equals (http://localhost:3000/bundle.js:550928:16)

Also, testing the code with jest I get:
detail: RangeError: Maximum call stack size exceeded
          at comparator (/home/rgciocan/WORK/dev/sources/marketplace-web/node_modules/fast-equals/src/comparator.ts:76:29)
          at Object.equals (/home/rgciocan/WORK/dev/sources/marketplace-web/node_modules/fast-equals/src/comparator.ts:307:12)
          at isPropertyEqual (/home/rgciocan/WORK/dev/sources/marketplace-web/node_modules/fast-equals/src/equals.ts:349:11)
          at areObjectsEqual (/home/rgciocan/WORK/dev/sources/marketplace-web/node_modules/fast-equals/src/equals.ts:158:10)
[...]

PR #4446, opened 2025-01-10
feat(utils): switch lodash.isEqualWith to fast-equals comparator to improve performance

igorbrasileiro:
### Reasons for making this change
This PR replaces `lodash.isEqualWith` with `fast-equals.createCustomEqual` at `@rjsf/utils.deepEquals` as the second part of this approach https://github.com/rjsf-team/react-jsonschema-form/issues/4291#issuecomment-2569881383 .
first part https://github.com/rjsf-team/react-jsonschema-form/pull/4438 

### Checklist

- [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.
  - [ ] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #4300, opened 2024-09-13
Revert "feat: improve deepEquals performance (#4292)"

igorbrasileiro:
### Reasons for making this change
This reverts the pr #4292 due to `maximum call stack exceeded` issue reported https://github.com/rjsf-team/react-jsonschema-form/pull/4292#issuecomment-2345672065

If your PR is non-trivial and you'd like to schedule a synchronous review, please add it to the weekly meeting agenda: https://docs.google.com/document/d/12PjTvv21k6LIky6bNQVnsplMLLnmEuypTLQF8a-8Wss/edit

### Checklist

- [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.
  - [ ] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #5046, opened 2026-04-28
perf(utils, validator-ajv8): improve deepEquals performance and skip redundant Ajv schema re-registration

### LINKED ISSUE #4291: performance issue related to lodash/isEqual function

FredrikWallstrom:

Fixes [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).

On large schemas with deep `dependencies` / `oneOf` / `anyOf` branches and richly populated form data, profiling shows the per-click cost in `<Form>` is dominated by `lodash.isEqualWith` walks routed through `deepEquals`. After applying both changes in this PR `baseIsEqualDeep` drops from the top of the bottom-up panel to a small fraction of click time.

There is a history here that this PR explicitly tries not to repeat:

- [#4292](https://github.com/rjsf-team/react-jsonschema-form/pull/4292) replaced `lodash.isEqualWith` with `fast-equals.createCustomEqual`. It was reverted by [#4300](https://github.com/rjsf-team/react-jsonschema-form/pull/4300) because cyclic `formData` (e.g. an `onChange` handler storing the whole `IChangeEvent`) hit `Maximum call stack size exceeded`.
- [#4446](https://github.com/rjsf-team/react-jsonschema-form/pull/4446) re-attempted the migration after `fast-equals` shipped function comparison support. It was reverted by [#4478](https://github.com/rjsf-team/react-jsonschema-form/pull/4478) because [#4475](https://github.com/rjsf-team/react-jsonschema-form/issues/4475) reported the same `Maximum call stack size exceeded` symptom under `liveOmit` + `omitExtraData` with custom validation.

The common theme in both reverts: `fast-equals.createCustomEqual` was used **without** opting into cycle detection, and `lodash.isEqualWith`'s built-in `Stack` cache had been silently protecting consumers with cyclic inputs.

### What this PR does

Two independent perf changes, one per package, in two commits.

#### 1. `perf(utils): switch deepEquals to fast-equals with cycle detection`

- `deepEquals` now uses `createCustomEqual({ circular: true, ... })`. Cycle detection is the difference from #4292 / #4446 and is what makes this safe.
- The `lodash.isEqual` callers inside `@rjsf/utils` (`useDeepCompareMemo`, `isRootSchema`, `findSelectedOptionInXxxOf`) are migrated to `deepEquals` so the perf benefit is shared across the package — same intent as the second half of #4446.
- New regression tests in `packages/utils/test/deepEquals.test.ts` exercise self-referential, mutually-referential, and cyclic-array inputs. These tests pass on the current `lodash.isEqualWith`-based implementation (lodash also tracks cycles), they pass on this PR, and they fail (`Maximum call stack size exceeded`) on a `fast-equals` configuration that omits `circular: true` — which is the failure mode that drove both prior reverts. So the tests serve as a forward guard against re-introducing the same regression.

#### 2. `perf(validator-ajv8): cache rootSchema reference in handleSchemaUpdate`

Independent of the `deepEquals` change. `retrieveSchema` invokes `isValid` (and therefore `handleSchemaUpdate`) once per `if`/`then`/`else` and per `dependencies.<key>.oneOf` branch evaluated during a render. On a form with a large `rootSchema` containing many branches the deep-equality check inside `handleSchemaUpdate` dominates click cost, even with the faster `deepEquals` from commit 1.

This commit caches the last-seen `rootSchema` reference and short-circuits when it is referentially equal on subsequent calls. Most consumers pass a stable `rootSchema` across calls within a render, so all but the first call become O(1). The structural `deepEquals` fallback still runs when the reference changes, preserving correctness for callers that legitimately rebuild the schema. New regression tests in `packages/validator-ajv8/test/validator.test.ts` cover both the cache hit and the structural fallback.

### Trade-offs / risks

- `fast-equals` is added as a new direct dependency of `@rjsf/utils` (not a peer). Same dep that was added in #4292 / #4446.
- Behavioural difference vs `lodash.isEqualWith`: the function-equality customizer (#255) is preserved, and cycle handling is preserved via `circular: true`. The remaining surface — primitives, plain objects, arrays, RegExp, Map, Set, typed arrays — is supported by `fast-equals` and matches `lodash.isEqualWith` for the inputs RJSF compares (form data, schemas, error schemas, `FieldPathId`).
- `handleSchemaUpdate` semantics change: when called repeatedly with the same `rootSchema` reference within a render, no Ajv re-registration happens. This was already a no-op in practice (`deepEquals` returned `true` for the same content), so observable behaviour is identical; only the cost differs.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Switched `deepEquals` from `lodash.isEqualWith` to `fast-equals.createCustomEqual` with cycle detection enabled, and replaced direct `lodash.isEqual` usage in `useDeepCompareMemo`, `isRootSchema`, and `findSelectedOptionInXxxOf` with `deepEquals`, addressing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
 
+## @rjsf/validator-ajv8
+
+- Cached the most recent `rootSchema` reference in `handleSchemaUpdate` so repeated `isValid` calls with the same root schema skip the deep-equality check and Ajv re-registration, addressing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
+
 ## Dev / docs / playground
 
 - Cleaned up testing to make registry mocks simpler using `getTestRegistry()` function
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 6.5.2
 
+## @rjsf/utils
+
+- Switched `deepEquals` from `lodash.isEqualWith` to `fast-equals.createCustomEqual` with cycle detection enabled, and replaced direct `lodash.isEqual` usage in `useDeepCompareMemo`, `isRootSchema`, and `findSelectedOptionInXxxOf` with `deepEquals`, addressing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
+
 ## Dev / docs / playground
 
 - Cleaned up testing to make registry mocks simpler using `getTestRegistry()` function
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,11 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/utils
 
-- Switched `deepEquals` from `lodash.isEqualWith` to `fast-equals.createCustomEqual` with cycle detection enabled, and replaced direct `lodash.isEqual` usage in `useDeepCompareMemo`, `isRootSchema`, and `findSelectedOptionInXxxOf` with `deepEquals`, addressing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
+- Switched `deepEquals` from `lodash.isEqualWith` to `fast-equals.createCustomEqual` with cycle detection enabled, and replaced direct `lodash.isEqual` usage in `useDeepCompareMemo`, `isRootSchema`, and `findSelectedOptionInXxxOf` with `deepEquals`, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
 
 ## @rjsf/validator-ajv8
 
-- Cached the most recent `rootSchema` reference in `handleSchemaUpdate` so repeated `isValid` calls with the same root schema skip the deep-equality check and Ajv re-registration, addressing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
+- Cached the most recent `rootSchema` reference in `handleSchemaUpdate` so repeated `isValid` calls with the same root schema skip the deep-equality check and Ajv re-registration, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
 
 ## Dev / docs / playground
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -22958,6 +22958,15 @@
       "dev": true,
       "license": "Apache-2.0"
     },
+    "node_modules/fast-equals": {
+      "version": "6.0.0",
+      "resolved": "https://registry.npmjs.org/fast-equals/-/fast-equals-6.0.0.tgz",
+      "integrity": "sha512-PFhhIGgdM79r5Uztdj9Zb6Tt1zKafqVfdMGwVca1z5z6fbX7DmsySSuJd8HiP6I1j505DCS83cLxo5rmSNeVEA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=6.0.0"
+      }
+    },
     "node_modules/fast-glob": {
       "version": "3.3.3",
       "resolved": "https://registry.npmjs.org/fast-glob/-/fast-glob-3.3.3.tgz",
@@ -42187,6 +42196,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@x0k/json-schema-merge": "^1.0.3",
+        "fast-equals": "^6.0.0",
         "fast-uri": "^3.1.0",
         "jsonpointer": "^5.0.1",
         "lodash": "^4.18.1",
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -66,6 +66,7 @@
   },
   "dependencies": {
     "@x0k/json-schema-merge": "^1.0.3",
+    "fast-equals": "^6.0.0",
     "fast-uri": "^3.1.0",
     "jsonpointer": "^5.0.1",
     "lodash": "^4.18.1",
--- a/packages/utils/src/deepEquals.ts
+++ b/packages/utils/src/deepEquals.ts
@@ -1,19 +1,20 @@
-import isEqualWith from 'lodash/isEqualWith';
+import { createCustomEqual } from 'fast-equals';
 
-/** Implements a deep equals using the `lodash.isEqualWith` function, that provides a customized comparator that
- * assumes all functions are equivalent.
+/** Implements a deep equals using `fast-equals.createCustomEqual`. Functions
+ * are always considered equal, and circular references are tracked to avoid
+ * infinite recursion on self-referential inputs.
  *
  * @param a - The first element to compare
  * @param b - The second element to compare
  * @returns - True if the `a` and `b` are deeply equal, false otherwise
  */
-export default function deepEquals(a: any, b: any): boolean {
-  return isEqualWith(a, b, (obj: any, other: any) => {
-    if (typeof obj === 'function' && typeof other === 'function') {
-      // Assume all functions are equivalent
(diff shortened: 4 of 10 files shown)


FredrikWallstrom:
Quick benchmark on a ~214 KB synthetic schema (root `$ref` → wrapper → views → picker-style field configs with `dependencies.field.oneOf`). 5-run averages on Node 24, M-series Mac:

| | unpatched v6.5.1 | this PR | Δ |
|---|---:|---:|---:|
| `deepEquals(rootSchema, clone)` | 1.06 ms | 0.70 ms | ~34 % |
| `handleSchemaUpdate(rootSchema)` × 100 | 0.006 ms | 0.001 ms | ~6× |

Repro:

```js
import { performance } from 'node:perf_hooks';
import { customizeValidator } from '@rjsf/validator-ajv8';
import { deepEquals } from '@rjsf/utils';

const VIEW_COUNT = 8;
const ENTRIES_PER_VIEW = 4;
const OPTIONS_PER_PICKER = 26;
const HELP_PARAGRAPH = 'A few sentences of contextual help describing what the option means and how it should be configured.';

const HelpDefinition = { type: 'object', properties: { title: { type: 'string' }, body: { type: 'string' }, examples: { type: 'array', items: { type: 'string' } } } };
const LayoutDefinition = { type: 'object', properties: { col: { type: 'integer', minimum: 1, maximum: 12 }, row: { type: 'integer', minimum: 1 }, span: { type: 'integer', minimum: 1, maximum: 12 } } };
const ValidationDefinition = { type: 'object', properties: { min: { type: 'number' }, max: { type: 'number' }, pattern: { type: 'string' }, minLength: { type: 'integer' }, maxLength: { type: 'integer' } } };

const buildPickerOptions = () =>
  Array.from({ length: OPTIONS_PER_PICKER }, (_, i) => ({ type: 'string', const: `option_${i}`, title: `Option ${i}`, description: `Pick this when the entry should be configured for option_${i}.` }));

const buildFieldConfig = (viewIndex) => ({
  type: 'object',
  required: ['field'],
  properties: {
    field: { title: 'Field', type: 'string', oneOf: buildPickerOptions() },
    label: { type: 'string', title: 'Label' },
    visible: { type: 'boolean', title: 'Visible' },
    help: { $ref: '#/definitions/Help' },
    layout: { $ref: '#/definitions/Layout' },
    validation: { $ref: '#/definitions/Validation' },
  },
  dependencies: {
    field: {
      oneOf: Array.from({ length: OPTIONS_PER_PICKER }, (_, i) => ({
        properties: {
          field: { const: `option_${i}` },
          [`option_${i}_extra`]: { type: 'string', title: `Extra for option_${i}`, description: `View_${viewIndex} configuration when field is option_${i}. ${HELP_PARAGRAPH}` },
          [`option_${i}_format`]: { type: 'string', title: 'Format', oneOf: ['short', 'medium', 'long', 'verbose'].map((f) => ({ const: f, title: `${f} format`, description: `Render the ${f} format for option_${i}.` })) },
          [`option_${i}_help`]: { $ref: '#/definitions/Help' },
          [`option_${i}_layout`]: { $ref: '#/definitions/Layout' },
          [`option_${i}_validation`]: { $ref: '#/definitions/Validation' },
        },
      })),
    },
  },
});

const buildView = (index) => {
  const properties = {};
  for (let i = 0; i < ENTRIES_PER_VIEW; i++) {
    properties[`entries_${i}`] = { type: 'array', title: `entries_${i} (view_${index})`, items: { $ref: `#/definitions/FieldConfig_${index}` } };
  }
  return { type: 'object', properties };
};

const viewNames = Array.from({ length: VIEW_COUNT }, (_, i) => `view_${i}`);
const definitions = {
  Help: HelpDefinition,
  Layout: LayoutDefinition,
  Validation: ValidationDefinition,
  Wrapper: { type: 'object', properties: Object.fromEntries(viewNames.map((name, i) => [name, { $ref: `#/definitions/View_${i}` }])) },
};
viewNames.forEach((_, i) => {
  definitions[`View_${i}`] = buildView(i);
  definitions[`FieldConfig_${i}`] = buildFieldConfig(i);
});

const rootSchema = { $id: 'http://example.com/bench-root', $schema: 'http://json-schema.org/draft-07/schema#', $ref: '#/definitions/Wrapper', definitions };

console.log(`rootSchema serialized size: ~${Math.round(JSON.stringify(rootSchema).length / 1024)} KB\n`);

const time = (label, fn, iters) => {
  fn();
  const start = performance.now();
  for (let i = 0; i < iters; i++) fn();
  const ms = performance.now() - start;
  console.log(`${label.padEnd(60)} ${(ms / iters).toFixed(3)} ms/iter (${iters}× — total ${ms.toFixed(0)} ms)`);
};

const validator = customizeValidator({});
const cloned = JSON.parse(JSON.stringify(rootSchema));

console.log('--- deepEquals on rootSchema vs structurally-equal clone ---');
time('deepEquals(rootSchema, clone)', () => deepEquals(rootSchema, cloned), 200);

console.log('\n--- handleSchemaUpdate x100 with same rootSchema reference ---');
validator.handleSchemaUpdate(rootSchema);
time('100 × validator.handleSchemaUpdate(rootSchema)', () => {
  for (let i = 0; i < 100; i++) validator.handleSchemaUpdate(rootSchema);
}, 200);
```

x0k:
Since you are working with large schemas, you might find it useful to explore the following idea: whether it makes sense to use a second Ajv instance with `allErrors: false` for calls in the `isValid` method.

I tested this for #4990, but it had no effect there.

FredrikWallstrom:
> Since you are working with large schemas, you might find it useful to explore the following idea: whether it makes sense to use a second Ajv instance with `allErrors: false` for calls in the `isValid` method.
> 
> I tested this for #4990, but it had no effect there.

Thanks for the suggestion. I tried this on the consumer side, wrapped two customizeValidator instances behind a tiny ValidatorType adapter (one with allErrors: true for validateFormData, one with allErrors: false for isValid).

No measurable improvement, though. Most of our isValid calls return true, and since allErrors: false only short-circuits on failure, it's a no-op when validation succeeds, if I understand this correctly.

heath-freenome:
@FredrikWallstrom Thanks for the fix. Somehow you've managed to update the `antd` libraries in the `package-lock.json` which is causing the tests to fail. Might I suggest reverting that file and just reapplying the `fast-equals` addition in isolation.

FredrikWallstrom:
> @FredrikWallstrom Thanks for the fix. Somehow you've managed to update the `antd` libraries in the `package-lock.json` which is causing the tests to fail. Might I suggest reverting that file and just reapplying the `fast-equals` addition in isolation.

Oops, you are right,  thanks! Reverted package-lock.json and reapplied just the fast-equals dep, force-pushed onto the existing commit.

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 6.5.2
 
+## @rjsf/utils
+
+- Switched `deepEquals` from `lodash.isEqualWith` to `fast-equals.createCustomEqual` with cycle detection enabled, and replaced direct `lodash.isEqual` usage in `useDeepCompareMemo`, `isRootSchema`, and `findSelectedOptionInXxxOf` with `deepEquals`, addressing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
```
```suggestion
- Switched `deepEquals` from `lodash.isEqualWith` to `fast-equals.createCustomEqual` with cycle detection enabled, and replaced direct `lodash.isEqual` usage in `useDeepCompareMemo`, `isRootSchema`, and `findSelectedOptionInXxxOf` with `deepEquals`, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
```

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ should change the heading of the (upcoming) version to include a major version b
...
+
+## @rjsf/validator-ajv8
+
+- Cached the most recent `rootSchema` reference in `handleSchemaUpdate` so repeated `isValid` calls with the same root schema skip the deep-equality check and Ajv re-registration, addressing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
```
```suggestion
- Cached the most recent `rootSchema` reference in `handleSchemaUpdate` so repeated `isValid` calls with the same root schema skip the deep-equality check and Ajv re-registration, fixing [#4291](https://github.com/rjsf-team/react-jsonschema-form/issues/4291).
```

heath-freenome (approved):

PR #5056, opened 2026-05-07
fix(core): omit undefined object leaves so string validation works with patternProperties

### LINKED ISSUE #4518: Validation of a property with patternProperties is wrongly failing
### Prerequisites

### What theme are you using?

core

### Version

5.x

### Current Behavior

The validation of the property with pattern is failing, even when it is empty.

### Expected Behavior

The validation should pass in this scenario.

### Steps To Reproduce

1. Open [this shared link](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJhbm5vdGF0aW9ucyI6e319LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsiYW5ub3RhdGlvbnMiOnsidHlwZSI6Im9iamVjdCIsImFkZGl0aW9uYWxQcm9wZXJ0aWVzIjp7InR5cGUiOiJzdHJpbmcifSwicGF0dGVyblByb3BlcnRpZXMiOnsiXi4rJCI6eyJ0eXBlIjoic3RyaW5nIn19LCJkZXNjcmlwdGlvbiI6IkEgc2V0IG9mIGtleS12YWx1ZSBhbm5vdGF0aW9ucy4iLCJwcm9wZXJ0aWVzIjp7InRlc3RFbXB0eUFubm90YXRpb24xIjp7InR5cGUiOiJzdHJpbmciLCJ0aXRsZSI6ImFubm90YXRpb24xIiwiZGVzY3JpcHRpb24iOiJGaWxsIHZhbHVlLCBzdWJtaXQsIHJlbW92ZSB2YWx1ZSwgc3VibWl0In0sInRlc3RFbXB0eUFubm90YXRpb24yIjp7InR5cGUiOiJzdHJpbmciLCJ0aXRsZSI6ImFubm90YXRpb24yIiwiZGVzY3JpcHRpb24iOiJGaWxsIHZhbHVlLCBzdWJtaXQsIHJlbW92ZSB2YWx1ZSwgc3VibWl0In19fX19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2RlZmF1bHRGb3JtU3RhdGVCZWhhdmlvciI6eyJhcnJheU1pbkl0ZW1zIjoicG9wdWxhdGUiLCJhbGxPZiI6InNraXBEZWZhdWx0cyIsImNvbnN0QXNEZWZhdWx0cyI6ImFsd2F5cyIsImVtcHR5T2JqZWN0RmllbGRzIjoicG9wdWxhdGVBbGxEZWZhdWx0cyIsIm1lcmdlRGVmYXVsdHNJbnRvRm9ybURhdGEiOiJ1c2VGb3JtRGF0YUlmUHJlc2VudCJ9fSwidmFsaWRhdG9yIjoiQUpWOCJ9) of RJSF Playground example.
2. Click in the form submit button, the validation is **ok**.
3. Fill **annotation1** field with any text and then clean it.
4. Click again in the submit button and now the validation **fails**, even with the annotation object **empty**, the same state it was in **step 2**.

### Environment

```markdown
React Json Schema Playground environment - https://rjsf-team.github.io/react-jsonschema-form
```

### Anything else?

Image

abhinavgautam01:
fixes #4518

Clearing optional strings left `{ key: undefined }`, which fails `type: string` for AJV.  
Unset keys for non-oneOf/anyOf leaves; preserve explicit `undefined` for oneOf/anyOf.
- After a user typed into a nested optional string and cleared it, `Form` merged the change with lodash `set`, so the in-memory `formData` kept keys whose values were `undefined`. That is not valid 
JSON and AJV treats it like a present value that is not a string, so validation could fail even when the object looked empty again (e.g. objects mixing `patternProperties` / `additionalProperties` 
with fixed `properties`) — see [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518).
- For **oneOf** / **anyOf** fields, an explicit `undefined` must remain so default merging does not immediately re-apply a branch default when the user selects an empty option.

### Checklist

### DIFF
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -926,11 +926,14 @@ export default class Form<
         _unset(formData, path);
       } else if (!isRootPath) {
         // If the newValue is not on the root path, then set it into the form data
+        let unsetPath = false;
+        let valueForPath: T | null | undefined = newValue;
+
         if (newValue === undefined) {
           const lastSegment = path[path.length - 1];
           if (typeof lastSegment === 'number') {
-            // Array items: match ArrayField `handleChange` — AJV needs `null`, not a hole/undefined.
-            _set(formData, path, null);
+            // Array items: match ArrayField `handleChange` — AJV needs `null`, not undefined.
+            valueForPath = null as unknown as T;
           } else {
             const { field } = schemaUtils.findFieldInSchema(schema, path, oldFormData);
             const leaf = field as RJSFSchema | undefined;
@@ -938,17 +941,21 @@ export default class Form<
             if (isOneOfOrAnyOfLeaf) {
               // Keep explicit `undefined` so mergeDefaults does not immediately re-apply a branch default (e.g. cleared
               // oneOf / anyOf widget).
-              _set(formData, path, newValue);
+              valueForPath = newValue;
             } else if (leaf !== undefined) {
               // Plain leaves: omit the key instead of `{ key: undefined }`, which breaks `type: "string"` validation in
               // AJV after clearing a text input (https://github.com/rjsf-team/react-jsonschema-form/issues/4518).
-              _unset(formData, path);
+              unsetPath = true;
             } else {
-              _set(formData, path, newValue);
+              valueForPath = newValue;
             }
           }
+        }
+
+        if (unsetPath) {
+          _unset(formData, path);
         } else {
-          _set(formData, path, newValue);
+          _set(formData, path, valueForPath);
         }
       }
       // Pass true to skip live validation in `getStateFromProps()` since we will do it a bit later
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -935,7 +935,12 @@ export default class Form<
             // Array items: match ArrayField `handleChange` — AJV needs `null`, not undefined.
             valueForPath = null as unknown as T;
           } else {
-            const { field } = schemaUtils.findFieldInSchema(schema, path, oldFormData);
+            const { field } = schemaUtils.findFieldInSchema(
+              schema,
+              // `findFieldInSchema` typings use `string[]`; paths may include numeric segments (same as lodash `get`).
+              path as string[],
+              oldFormData,
+            );
             const leaf = field as RJSFSchema | undefined;
             const isOneOfOrAnyOfLeaf = leaf && (ONE_OF_KEY in leaf || ANY_OF_KEY in leaf);
             // Plain leaves: omit the key instead of `{ key: undefined }`, which breaks `type: "string"` validation in
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -45,6 +45,8 @@ import {
   NameGeneratorFunction,
   getUsedFormData,
   getFieldNames,
+  ANY_OF_KEY,
+  ONE_OF_KEY,
 } from '@rjsf/utils';
 import _cloneDeep from 'lodash/cloneDeep';
 import _get from 'lodash/get';
@@ -924,7 +926,30 @@ export default class Form<
         _unset(formData, path);
       } else if (!isRootPath) {
         // If the newValue is not on the root path, then set it into the form data
-        _set(formData, path, newValue);
+        if (newValue === undefined) {
+          const lastSegment = path[path.length - 1];
+          if (typeof lastSegment === 'number') {
+            // Array items: match ArrayField `handleChange` — AJV needs `null`, not a hole/undefined.
+            _set(formData, path, null);
+          } else {
+            const { field } = schemaUtils.findFieldInSchema(schema, path, oldFormData);
+            const leaf = field as RJSFSchema | undefined;
+            const isOneOfOrAnyOfLeaf = leaf && (ONE_OF_KEY in leaf || ANY_OF_KEY in leaf);
+            if (isOneOfOrAnyOfLeaf) {
+              // Keep explicit `undefined` so mergeDefaults does not immediately re-apply a branch default (e.g. cleared
+              // oneOf / anyOf widget).
+              _set(formData, path, newValue);
+            } else if (leaf !== undefined) {
+              // Plain leaves: omit the key instead of `{ key: undefined }`, which breaks `type: "string"` validation in
+              // AJV after clearing a text input (https://github.com/rjsf-team/react-jsonschema-form/issues/4518).
+              _unset(formData, path);
+            } else {
+              _set(formData, path, newValue);
+            }
+          }
+        } else {
+          _set(formData, path, newValue);
+        }
       }
       // Pass true to skip live validation in `getStateFromProps()` since we will do it a bit later
       const newState = this.getStateFromProps(this.props, inputForDefaults, undefined, undefined, undefined, true);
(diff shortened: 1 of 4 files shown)


abhinavgautam01:
ping @heath-freenome

packages/core/src/components/Form.tsx
```diff
@@ -924,7 +926,30 @@ export default class Form<
         _unset(formData, path);
       } else if (!isRootPath) {
         // If the newValue is not on the root path, then set it into the form data
-        _set(formData, path, newValue);
+        if (newValue === undefined) {
+          const lastSegment = path[path.length - 1];
+          if (typeof lastSegment === 'number') {
+            // Array items: match ArrayField `handleChange` — AJV needs `null`, not a hole/undefined.
```
heath-freenome:
```suggestion
            // Array items: match ArrayField `handleChange` — AJV needs `null`, not undefined.
```

packages/core/src/components/Form.tsx
```diff
@@ -924,7 +926,30 @@ export default class Form<
...
+              // oneOf / anyOf widget).
+              _set(formData, path, newValue);
+            } else if (leaf !== undefined) {
+              // Plain leaves: omit the key instead of `{ key: undefined }`, which breaks `type: "string"` validation in
+              // AJV after clearing a text input (https://github.com/rjsf-team/react-jsonschema-form/issues/4518).
+              _unset(formData, path);
+            } else {
+              _set(formData, path, newValue);
+            }
+          }
```
heath-freenome, line 953: Is there a way to DRY this up so that `_set(formData, path, newValue);` only shows up once?

packages/core/src/components/Form.tsx
abhinavgautam01: Thanks for the suggestion — agreed, “hole” wasn’t adding much. Updated the comment to match your wording in the latest push.

packages/core/src/components/Form.tsx
abhinavgautam01: Good call. I refactored that block so we only resolve what to do first: either set unsetPath (plain non–oneOf/anyOf leaf + undefined → _unset) or compute valueForPath (including null for array indices and undefined for oneOf/anyOf / fallback). Then a single branch runs either _unset(formData, path) or _set(formData, path, valueForPath). Pushed in the same follow-up commit; full @rjsf/core test suite still passes...

packages/core/src/components/Form.tsx
```diff
@@ -924,7 +926,37 @@ export default class Form<
...
+              // Keep explicit `undefined` so mergeDefaults does not immediately re-apply a branch default (e.g. cleared
+              // oneOf / anyOf widget).
+              valueForPath = newValue;
+            } else if (leaf !== undefined) {
+              // Plain leaves: omit the key instead of `{ key: undefined }`, which breaks `type: "string"` validation in
+              // AJV after clearing a text input (https://github.com/rjsf-team/react-jsonschema-form/issues/4518).
+              unsetPath = true;
+            } else {
+              valueForPath = newValue;
+            }
```
heath-freenome: Since `valueForPath` is already defaulted to `newValue` on line 930, you only need the branch that sets `unsetPath = true`, right?

abhinavgautam01: You’re right — those branches were redundant. Since valueForPath is initialized from newValue, the only extra work when newValue === undefined (non-array segment) is setting unsetPath for resolved plain leaves; oneOf/anyOf and “leaf not found” now just rely on that default. Pushed an update with a brief comment explaining why we leave undefined in those cases.

packages/core/src/components/Form.tsx
```diff
@@ -924,7 +926,38 @@ export default class Form<
...
+        if (newValue === undefined) {
+          const lastSegment = path[path.length - 1];
+          if (typeof lastSegment === 'number') {
+            // Array items: match ArrayField `handleChange` — AJV needs `null`, not undefined.
+            valueForPath = null as unknown as T;
+          } else {
+            const { field } = schemaUtils.findFieldInSchema(
+              schema,
+              // `findFieldInSchema` typings use `string[]`; paths may include numeric segments (same as lodash `get`).
+              path as string[],
```
heath-freenome: I wonder what happens if `findFieldInSchema()` gets an actual number instead of string? I'd consider adding a unit test that tests that use case.

Would it be better to do something like this instead, (especially if the use-case fails)?
```

```
```suggestion
              // `findFieldInSchema` typings use `string[]`; paths may include numeric segments (same as lodash `get`).
              path.map((s) => String(s)),
```

abhinavgautam01: Agreed — using path.map((s) => String(s)) is cleaner than a cast: we pass a real string[] and normalize numeric segments the same way property keys are used. Added a findFieldInSchema unit test with a schema whose property name is "0" and a FieldPathList starting with 0 to lock this in. Pushed in the latest commit.

packages/utils/test/schema/findFieldInSchemaTest.ts
```diff
@@ -87,6 +87,26 @@ export default function findFieldInSchemaTest(testValidator: TestValidatorType)
...
+          '0': {
+            type: 'string',
+          },
+        },
+      };
+      const fieldPathList = [0] as (string | number)[];
+      expect(
+        schemaUtils.findFieldInSchema(
+          schemaWithNumericPropertyName,
+          fieldPathList.map((s) => String(s)),
```
heath-freenome, line 103: What happens without this coercion? Does the code blow up? I'm curious more than anything. If it does not blow up, then we might want to consider updating the signature to support `(string | number)[]`. If it does blow up then maybe we update the function to support it instead? Thoughts?

abhinavgautam01: at runtime it generally doesn’t blow up — segments are just used as keys when walking properties, and JS usually treats `0` vs `"0"` the same there...

The problem we hit was mostly with `tsc`: `Form` uses `FieldPathList` (`(string | number)[]`) but `findFieldInSchema` is typed as `string | string[]`, so we either end up coercing, casting, or widening types through the stack....

Widening to `FieldPathList` everywhere is reasonable longer-term, but it likely needs updates below `findFieldInSchema` too (e.g. `getFromSchema`), not just the signature....

Here I just normalized at the call site and added a small test for the numeric segment case. Happy to do a follow-up PR to thread the wider path type through if that’s preferred....!

heath-freenome: Understood. I'd definitely love to see a follow-up PR when you get the chance. Then we can remove the mapping to string in Form since it does potentially add a tiny bit more overhead

heath-freenome (approved):

PR #5065, opened 2026-05-09
Chore: Adopt SJSF omitExtraData algorithm

heath-freenome:

Adopted the `omitExtraData()` processing algorithm from `svelte-jsonschema-form`, adding in the `removeOptionalEmptyObject()` behavior
- In `@rjsf/utils`:
  - Added `relaxOptionsForScoring()` to `retrieveSchema.ts`, adding it to the `resolveAnyOrOneOfSchemas()` for the `expandAllBranches=true` branch.
  - Rewrote `omitExtraData()` basing it on the algorithm from `SJSF`, with improvements to pick up the `removeOptionalEmptyObject()` behavior
  - Refactored the `isValueEmpty()` function from `removeOptionalEmptyObject()` for use by `omitExtraData()` exporting it from utils
  - Extracted the `shallowAllOfMerge()` function from `retrieveSchema.ts` into its own file for additional use in `omitExtraData.ts`
  - Deprecated the `removeOptionalEmptyObject()` and `toPathSchema()` functions, `toPathSchema()` on `SchemaUtilsType` and the `PathSchema` type
  - Updated the tests for `toPathSchema()` to add missing coverage after the rewrite
  - Updated the tests for `omitExtraData()` to cover `isValueEmpty()`, the ported `removeOptionalEmptyObject()` behavior and keeping coverage 100%
  - Added tests for `relaxOptionsForScoring()` to add missing coverage
- In `@rjsf/core`:
  - Updated `Form` to stop using `removeOptionalEmptyObjects()` and deprecating `removeEmptyOptionalObjects` prop.
  - Added tests for `Form` to cover the remove empty objects cases
- Updated the `form-props.md` and `utility-functions.md` to document the deprecations
- Updated the `uiSchema.md` to improve the `enumNames` documents slightly

### Checklist

### DIFF
--- a/packages/utils/src/schema/omitExtraData.ts
+++ b/packages/utils/src/schema/omitExtraData.ts
@@ -171,9 +171,12 @@ export default function omitExtraData<
 
   /** Copies schema-defined properties from `source` into `target`, applying `omit` recursively for
    * each value. Handles `properties`, `patternProperties`, `additionalProperties`, and `propertyNames`.
-   * Optional object-valued properties whose filtered result is entirely empty (per `isValueEmpty`) are
-   * silently dropped; required properties and scalar values are always written when defined. Always
-   * returns `target` — pruning of the object itself is the caller's responsibility.
+   * Optional object-valued properties are pruned when every key in the filtered result is both
+   * optional (per the inner schema's `required`) and empty (per `isValueEmpty`). This preserves
+   * optional objects whose required children have empty values, while still dropping objects whose
+   * schema-filtered content is entirely optional-and-empty. Required properties and scalar values
+   * are always written when defined. Always returns `target` — pruning of the object itself is
+   * the caller's responsibility.
    *
    * @param schema - The object schema describing which properties are allowed
    * @param source - The source form data object to read values from
@@ -185,8 +188,12 @@ export default function omitExtraData<
     const requiredSet = new Set((schema.required ?? []) as string[]);
 
     /** Recursively omits extra data from `value` via `omit`, then conditionally writes the result to
-     * `target[key]`. Optional properties whose filtered value is a plain object considered entirely
-     * empty by `isValueEmpty` are dropped; all other defined values are written when not `undefined`.
+     * `target[key]`. Optional object-valued properties are dropped when every key in the filtered
+     * result is both optional (within the inner schema's `required` list) and has an empty value per
+     * `isValueEmpty`. This per-key approach prevents required-but-empty child properties — kept by
+     * inner `setProperty` calls — from inadvertently causing the parent optional object to be dropped,
+     * while still pruning optional objects whose schema-filtered content is entirely empty. Vacuously
+     * true for `{}`, so empty results are always dropped. Scalar and array values are not pruned here.
      *
      * @param key - The property key to write on `target`
      * @param schemaDef - The schema (or boolean shorthand) that governs the value at `key`
@@ -195,8 +202,21 @@ export default function omitExtraData<
      */
     function setProperty(key: string, schemaDef: S | boolean, value: unknown, required = false) {
       const v = omit(schemaDef, value, target[key]);
-      if (!required && isObject(v) && isValueEmpty(v as GenericObjectType)) {
-        return;
+      if (!required && isObject(v)) {
+        // Resolve $ref so we can inspect the effective required list for the inner schema.
+        let sd = isSchemaObj(schemaDef as S | boolean) ? (schemaDef as S) : ({} as S);
+        if (sd.$ref !== undefined) {
+          sd = findSchemaDefinition(sd.$ref, rootSchema) as S;
+        }
+        const innerRequired = new Set((sd.required ?? []) as string[]);
+        // Drop this optional object when every key in v is both optional in the inner schema
+        // and has an empty value. Vacuously true for {} so empty objects are always dropped.
+        const shouldDrop = Object.entries(v as GenericObjectType).every(
+          ([k, val]) => !innerRequired.has(k) && isValueEmpty(val),
+        );
+        if (shouldDrop) {
+          return;
+        }
       }
       if (v !== undefined) {
         target[key] = v;
@@ -231,6 +251,8 @@ export default function omitExtraData<
       }
     }
 
+    // JSON Schema spec: absent additionalProperties defaults to true (allow all extra keys). Here
+    // we treat it as false so omitExtraData never inadvertently passes through undescribed keys.
     if (additionalProperties !== undefined && additionalProperties !== false) {
       const addlSchema = additionalProperties as S | boolean;
       if (patternPropertiesRest !== undefined) {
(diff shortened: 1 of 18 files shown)


heath-freenome:
@x0k Per your suggestion, I've adapted your code to this library. Feel free to take a look

packages/utils/src/schema/omitExtraData.ts
```diff
@@ -68,26 +83,373 @@ export function getFieldNames<T = any>(pathSchema: PathSchema<T>, formData?: T):
...
+
+    // Track keys not handled by properties/patterns so additionalProperties can pick them up.
+    let patternPropertiesRest: string[] | undefined;
+    if (patternProperties !== undefined) {
+      patternPropertiesRest = [];
+      const patterns = Object.entries(patternProperties).map(([pattern, schemaDef]): [RegExp, S | boolean] => [
+        new RegExp(pattern),
+        schemaDef as S | boolean,
+      ]);
+      const knownProperties = new Set(Object.keys(properties ?? {}));
```
x0k, line 232: Removing `getKnownProperties` will result in `bar` being preserved in the following case (if this was intentional, it would make sense to add this test to the test suite)

```ts
const schema: RJSFSchema = {
  type: 'object',
  oneOf: [
    {
      properties: {
        kind: { const: 'a' },
        foo: { type: 'string' },
      },
    },
    {
      properties: {
        kind: { const: 'b' },
        bar: { type: 'string' },
      },
    },
  ],
  additionalProperties: {
    type: 'string',
  },
};

const value = {
  kind: 'a',
  bar: 'hello',
  extra: 'keep me',
};

const result = omitExtraData(testValidator, schema, schema, value);

expect(result).toEqual({
  kind: 'a',
  bar: 'hello',
  extra: 'keep me',
});
```

packages/utils/src/schema/omitExtraData.ts
```diff
@@ -68,26 +83,373 @@ export function getFieldNames<T = any>(pathSchema: PathSchema<T>, formData?: T):
...
+export function isValueEmpty(value: unknown): boolean {
+  if (isNil(value) || value === '') {
+    return true;
+  }
+  if (Array.isArray(value)) {
+    return value.length === 0;
+  }
+  if (isObject(value)) {
+    return Object.values(value as GenericObjectType).every(isValueEmpty);
+  }
```
x0k, line 101: This function should not be recursive, or `setProperty` should call a different function. Since `omitExtraData` performs a depth-first traversal, empty objects could be collapsed when returning the result from `omit`, for example. It may even make sense to integrate all `isValueEmpty` logic directly into `omit`.

packages/utils/src/schema/omitExtraData.ts
```diff
@@ -68,26 +83,373 @@ export function getFieldNames<T = any>(pathSchema: PathSchema<T>, formData?: T):
...
+    if (!Array.isArray(oneOf) || isSelect(validator, schema, rootSchema, experimental_customMergeAllOf)) {
+      return target;
+    }
+    // Mirror svelte: relax additionalProperties:false so getClosestMatchingOption can validate data freely.
+    const options = (oneOf as Array<S | boolean>).map((d) => {
+      if (!isSchemaObj(d)) {
+        return (d ? {} : { not: {} }) as S;
+      }
+      return d.additionalProperties === false ? ({ ...d, additionalProperties: true } as S) : d;
+    });
```
x0k: Oh, it looks like this will not work with precompiled validators.

x0k (commented):
Just in case, I’m attaching my implementation of `removeOptionalEmptyObject`, which has virtually no overhead [PR](https://github.com/x0k/svelte-jsonschema-form/pull/342/changes) + [fix](https://github.com/x0k/svelte-jsonschema-form/pull/345/changes).

packages/utils/src/schema/omitExtraData.ts
```diff
@@ -68,26 +83,373 @@ export function getFieldNames<T = any>(pathSchema: PathSchema<T>, formData?: T):
...
+        const matched = patterns.find(([re]) => re.test(key));
+        if (matched === undefined) {
+          patternPropertiesRest.push(key);
+          continue;
+        }
+        setProperty(key, matched[1], value);
+      }
+    }
+
+    if (additionalProperties !== undefined && additionalProperties !== false) {
```
x0k, line 248: If the `additionalProperties` schema is not explicitly specified, its value is treated as `false`, which differs from the behavior defined by JSON Schema. It might be worth adding a note about this peculiarity.

packages/utils/src/schema/omitExtraData.ts
heath-freenome: Test was added

packages/utils/src/schema/omitExtraData.ts
heath-freenome: Updated `setProperty()` per your recommendation

packages/utils/src/schema/omitExtraData.ts
x0k: > This is the same limitation that already exists elsewhere in the codebase (e.g. MultiSchemaField) when getClosestMatchingOption is used with dynamic schemas

I wasn’t aware of this. Perhaps [precompile module](https://github.com/x0k/svelte-jsonschema-form/blob/9cecc6d9cd3e7ebae49ad858d7a5dd813e073597/packages/form/src/validators/precompile.ts) ([usage](https://github.com/x0k/svelte-jsonschema-form/blob/9cecc6d9cd3e7ebae49ad858d7a5dd813e073597/apps/docs2/src/content/docs/validators/ajv/precompile/compile-schema-script.ts)) could also be adapted at some point.

nickgros: @heath-freenome did the prior `omitExtraData` handle (or not handle) `oneOf` matching in a way that worked for precompiled validators? Does this represent a functional regression?

heath-freenome: @nickgros @x0k I took your feedback to heart and made sure the precompiled schemas work correctly now

packages/utils/src/schema/retrieveSchema.ts
```diff
@@ -694,13 +694,38 @@ export function resolveAnyOrOneOfSchemas<
     // Call this to trigger the set of isValid() calls that the schema parser will need
     const option = getFirstMatchingOption<T, S, F>(validator, formData, anyOrOneOf, rootSchema, discriminator);
     if (expandAllBranches) {
+      // Also trigger isValid() for the relaxed variants so that precompiled validators capture their hashes.
+      // omitExtraData's handleOneOf relaxes additionalProperties:false → true before scoring; those mutated
+      // schemas must be present in a precompiled validator's compiled set or isValid() will throw at runtime.
+      const relaxed = relaxOptionsForScoring<S>(anyOrOneOf);
+      for (const relaxedOption of relaxed) {
+        validator.isValid(relaxedOption, formData, rootSchema);
+      }
```
x0k: Is the result of `isValid` not being used?

Honestly, to verify that this works properly, it would be better to add integration tests (`getFirstMatchingOption`, `getClosestMatchingOption`, `omitExtraData`) for the precompiled validator.

x0k: I had been ignoring the `expandAllBranches` flag this whole time, but now I understand that it is needed for full schema traversal in order to collect schemas for the precompiled validator.

heath-freenome: Glad you came to understand the use of the flag. Sounds like the documentation of the flag needs improving?

x0k: I think the code already has enough comments (which I successfully ignored).

The only thing I would recommend at this point is running the same test suite for both runtime and precompiled validators.

heath-freenome: I believe that the additional tests added to the validator cover it

nickgros (approved):

Issue #5052, opened 2026-05-01
Add an `ata-validator` alternative to AJV


TheOneTheOnlyJJ:
### Prerequisites

- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)

### What theme are you using?

validator-ajv8

### Describe the solution you'd like

It is in RJSF's interest to have other validator options available. Especially given how the frequency of AJV updates has appeared to drop over the last year.

However, not many good options were available until now.

[ata-validator](https://github.com/ata-core/ata-validator) is a new high-performance JSON schema validator that could coexist with the existing AJV one. Pre-compiled schemas are also supported. Also related to #4614, as ata can output its errors in the Standard Schema format.

### Describe alternatives you've considered

Continue relying strictly on AJV.

mertcanaltin:
Hey @TheOneTheOnlyJJ, thanks for opening this.

happy to put together a `validator-ata` package that mirrors validator-ajv8 if there is interest.

x0k:
I just tried creating a validator based on `ata-validator`; here is some feedback:

- No support for custom formats [usage](https://github.com/rjsf-team/react-jsonschema-form/blob/d2fea869e04045d0c5813bb949abde5830ae7f9e/packages/validator-ajv8/src/createAjvInstance.ts#L51)  
- `ValidationError` does not include `parentSchema` [usage](https://github.com/rjsf-team/react-jsonschema-form/blob/d2fea869e04045d0c5813bb949abde5830ae7f9e/packages/validator-ajv8/src/processRawValidationErrors.ts#L86)  
- When using the Standard Schema interface, `issue.path` stores indices as strings - unclear why  
- Could not find a way to compile multiple schemas into an ESM bundle, also `toStandaloneModule` method is missing in `index.d.ts`  
- When compiling a CJS bundle (`bundleStandalone`), schemas with `$ref` were not compiled  [script](https://github.com/x0k/svelte-jsonschema-form/blob/35b128742198d8f41b680ba3e3407c70bd899fea/apps/docs2/src/content/docs/validators/ata/precompile/compile-schema-script.ts) [input](https://gist.github.com/x0k/9b9d33bda44ad6a548b3fe6a14d13c70) [output](https://github.com/x0k/svelte-jsonschema-form/blob/35b128742198d8f41b680ba3e3407c70bd899fea/apps/docs2/src/content/docs/validators/ata/precompile/validate-functions.js)

[...]

PR #5078, opened 2026-05-23
Add precompiled validator support to @rjsf/validator-ata

mertcanaltin:
Relates to #5052 (the precompiled part).

Adds `createPrecompiledValidator` and `compileSchemaValidators` to `@rjsf/validator-ata`, mirroring `@rjsf/validator-ajv8`, so a schema can be compiled ahead of time and loaded without runtime schema compilation.

ata's `bundleStandalone` emits an array of validators (one per parsed schema, each returning `{ valid, errors }`), so `compileSchemaValidatorsCode` wraps that into the `{ [id]: (data) => boolean }` map with an `errors` property that the precompiled consumer expects. The validator class and factory follow the ajv8 versions.

Built against `ata-validator` ^0.17.1.

A few limitations come from ata's standalone (AOT) path:

- Custom formats must be a RegExp or string pattern. A function checker can't be serialized into the bundle, so it is rejected with a clear error at compile time.
- A schema ata can't compile to standalone (for example a `$defs` entry that carries its own `$id`) surfaces a clear error rather than failing later.
- For `anyOf` and a schema-valued `additionalProperties`, the standalone output reports a single error on the parent field (the `anyOf` node) rather than per-field errors. The closest-branch detail is present but in ata's own shape, not the AJV error shape rjsf reads, so the form shows one error on the parent instead of on each sub-field. Validity itself is unaffected. This is an ata-side AOT limitation I'm tracking; I can add the per-field mapping if it matters for the precompiled path.

Tests mirror the ajv8 precompiled tests; the package suite and coverage pass.


packages/docs/docs/api-reference/validator-ata.md
```diff
@@ -10,7 +10,7 @@ See the [Validation documentation](../usage/validation.md) for examples of using
...
 - The `ata-validator` format set is always installed, so there is no opt-in flag for formats.
 - `ata-validator`'s error params are frozen, so the pre-quote pass that `validator-ajv8` runs for `ajv-i18n` is unnecessary. Localizers that mutate `message` in place still work as-is.
-- Precompiled validators (`compileSchemaValidators` / `createPrecompiledValidator`) are not yet wired through. `ata-validator`'s `bundleStandalone` output has a different shape than AJV's standalone modules and needs an adapter; this is planned as a follow-up.
+- Precompiled validators (`compileSchemaValidators` / `createPrecompiledValidator`) are supported. `ata-validator`'s `bundleStandalone` output is adapted into the validator function map the precompiled consumer expects. Two constraints carry over from the standalone (AOT) path: custom formats must be a RegExp or pre-anchored string pattern, since a function checker cannot be serialized into the bundle and is rejected at compile time; and an invalid `anyOf` or schema-valued `additionalProperties` reports a single error on the parent field rather than per-field errors. Validity is unaffected in both cases.
```
heath-freenome: > `This is an ata-side AOT limitation I'm tracking; I can add the per-field mapping if it matters for the precompiled path.`

Does this differ from how the ATA validator works in non-precompiled mode? If so, it feels worthy of making them consistent.

packages/validator-ata/src/compileSchemaValidatorsCode.ts
```diff
@@ -0,0 +1,87 @@
...
+          `@rjsf/validator-ata: custom format "${name}" is a function, which is not supported by the precompiled validator; provide a RegExp or string pattern instead`,
+        );
+      }
+      regexFormats[name] = typeof check === 'string' ? new RegExp(check) : check;
+    }
+  }
+  const formats: Record<string, (value: string) => boolean> = {};
+  for (const [name, re] of Object.entries(regexFormats)) {
+    // eslint-disable-next-line no-new-func
+    formats[name] = new Function('value', `return ${re.toString()}.test(value)`) as (value: string) => boolean;
```
heath-freenome, line 57: Would a fat-arrow function work here instead?

packages/validator-ata/test/compileSchemaValidatorsCode.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    name: { type: 'string' },
+    address: { $ref: '#/definitions/address' },
+  },
+  required: ['name'],
+  definitions: {
+    address: { type: 'object', properties: { city: { type: 'string' } }, required: ['city'] },
+  },
+};
+
+test('generates a module with a validator function per schema id', () => {
```
heath-freenome: Consider wrapping all of these `test()` calls inside of a `describe()` block

heath-freenome (commented):
@mertcanaltin Thanks for this. Just a few questions/suggestions

mertcanaltin:
Thanks @heath-freenome, updated for all three.

On the consistency question, it was worth fixing. I compared both paths against the runtime validator: `anyOf` already lined up (runtime reports a single error on the field too), but schema-valued `additionalProperties` didn't, the AOT path gave a generic error where runtime gave per-field. That's fixed in `ata-validator` 0.17.4, so the precompiled output now reports the same per-field `.field` errors as runtime. Bumped the dep to `^0.17.4`, added a test for it, and corrected the docs note.

On `new Function` vs a fat-arrow, I kept `new Function` on purpose. ata inlines these format checkers into the standalone bundle with `Function#toString`, and a fat-arrow would capture `re` and serialize as `re.test(value)`, which is undefined in the generated module (it throws `re is not defined`). Building it from a string puts the regex literal in the body so the bundle stays self-contained. Left a comment so it's clear.

Wrapped the tests in a describe.

heath-freenome (approved):

PR #5080, opened 2026-05-27
Removed jest/babel in favor of vitest

heath-freenome:
- Replaced `babel-jest` with `@swc/jest` across all packages, used "v8" coverage provider, removed dead code found during upgrade
    
- Remove all babel configs (root + 16 per-package) and babel devDependencies
    (@babel/cli, @babel/core, @babel/preset-env, @babel/preset-react,
    @babel/preset-typescript, babel-jest, etc.) from the root package.json.
    
- Add explicit @swc/jest transform config to all 15 per-package jest.config.json
    files. The root babel.config.json was silently used by Jest's auto-detection
    and was the real transformer for 12 packages that had no explicit transform.
    
- All packages with coverage thresholds use coverageProvider: "v8" instead of
    Istanbul, which accurately accounts for SWC-transpiled code.
    
- Remove dead code in getArrayDefaults, restore 100% branch coverage
    
- `computeDefaults` always handles schemas with a `const` property before
    delegating to `getArrayDefaults`, so `CONST_KEY in schema` was never true
    inside `getArrayDefaults`. The `hasConst` guard and its wrapping `if` block
    were dead code that V8 branch coverage correctly flagged as unreachable.
    
Chore: Replace jest with vitest across all packages, fix rebase artifacts
- Replace jest.fn/mock/spyOn/mocked with vi.* equivalents in tests
- Rebuild snapshot-tests lib/ output (stale compiled JS still referenced jest.*)
- Removed `@swc/*` since they are no longer needed with the conversion to `vitest`
- Also renamed `setup-test-env.(js/ts)` to `testSetup.ts` to make the naming consistent

Co-Authored-By: Claude Sonnet 4.6
### Checklist


packages/antd/test/__snapshots__/Array.test.tsx.snap
```diff
@@ -1,6 +1,6 @@
-// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`array fields array 1`] = `
+exports[`array fields > array 1`] = `
```
heath-freenome, line 3: Vitest added the `> ` at each describe/it/test block where jest didn't :(

packages/chakra-ui/test/testSetup.ts
heath-freenome, line 1: Renamed all of these to `setup-test-env.ts`... NOTE,  the `antd` one has a different name which I considered changing to match this. I still can if you think it would help maintainability

packages/core/test/anyOf.test.tsx
```diff
@@ -1402,7 +1403,7 @@ describe('anyOf', () => {
     });
 
     it('should warn when the anyOf in the uiSchema is not an array, and pass the base uiSchema down', () => {
-      const consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation();
+      const consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(noop);
```
heath-freenome, line 1406: vitest doesn't like an empty `mockImplementation()` so I used `noop` to quiet the lint error

packages/core/test/FieldTemplate.test.tsx
```diff
@@ -11,10 +11,12 @@ describe('FieldTemplate', () => {
       }
       return null;
     }
-    createFormComponent({
-      schema: { type: 'string' },
-      uiSchema: { 'ui:disabled': true },
-      templates: { FieldTemplate },
+    it('renders without throwing when exactly one child is provided', () => {
```
heath-freenome, line 14: This was just plain missing the `it` block, which claude caught and fixed

packages/core/test/LayoutMultiSchemaField.test.tsx
```diff
@@ -248,12 +250,9 @@ describe('LayoutMultiSchemaField', () => {
...
     expect(() => render(<LayoutMultiSchemaField {...props} />)).toThrow(expectedError);
-    expect(consoleErrorSpy).toHaveBeenNthCalledWith(
-      1,
-      expect.objectContaining({
-        message: expect.stringContaining(expectedError),
-        type: 'unhandled exception',
-      }),
+    expect(consoleErrorSpy).toHaveBeenCalledWith(
+      expect.stringContaining(expectedError),
+      expect.objectContaining({ message: expectedError }),
```
heath-freenome, line 255: vitest handled the object slightly differently

packages/core/test/StringField.test.tsx
```diff
@@ -2190,7 +2191,7 @@ describe('StringField', () => {
     });
 
     it('should render the file widget with download link', () => {
-      const formData = 'data:text/plain;name=file1.txt;base64,x=';
+      const formData = 'data:text/plain;name=file1.txt;base64,YQ==';
```
heath-freenome, line 2194: The old base 64 string was considered broken, so claude changed it to something that it deemed functional

packages/shadcn/resources/amethyst-haze.css
heath-freenome, line 1: Not totally sure why the css files changed

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -673,19 +673,13 @@ export function getArrayDefaults<T = any, S extends StrictRJSFSchema = RJSFSchem
...
-  if (hasConst === false) {
-    if (neverPopulate) {
-      return defaults ?? emptyDefault;
-    }
-    if (ignoreMinItemsFlagSet && !required) {
-      // If no form data exists or defaults are set leave the field empty/non-existent, otherwise
-      // return form data/defaults
-      return defaults ? defaults : undefined;
-    }
+  if (neverPopulate) {
```
heath-freenome, line 676: the babel removal caused the coverage to change, which detected dead code that was removed

CLAUDE.md
```diff
@@ -118,6 +118,6 @@ Fields select which widget to render based on schema `type` and `ui:widget`. Tem
 
 ## Testing
 
-- Jest + `@swc/jest` (fast transpilation), jsdom, Testing Library
+- Vitest 3, jsdom, Testing Library
```
heath-freenome:
```suggestion
- Vitest, jsdom, Testing Library
```

packages/validator-ata/test/validator.test.ts
```diff
@@ -309,6 +310,24 @@ describe('cloneForValidation', () => {
     }
   });
 
+  it('falls back to cloneDeep when structuredClone is unavailable', () => {
```
heath-freenome, line 313: I believe changing the coverage to v8 cause an uncovered path to be exposed and this fixes it

packages/utils/test/utcToLocal.test.ts
```diff
@@ -1,17 +1,18 @@
+import type { MockInstance } from 'vitest';
```
heath-freenome, line 1:
```suggestion
import type { MockInstance } from 'vitest';

```

packages/utils/test/resolveUiSchema.test.ts
```diff
@@ -1,10 +1,11 @@
+import noop from 'lodash/noop';
```
heath-freenome, line 1:
```suggestion
import noop from 'lodash/noop';

```

packages/utils/test/isRootSchema.test.ts
```diff
@@ -1,3 +1,4 @@
+import type { MockInstance } from 'vitest';
```
heath-freenome, line 1:
```suggestion
import type { MockInstance } from 'vitest';

```

packages/utils/test/getUiOptions.test.ts
```diff
@@ -1,3 +1,5 @@
+import type { MockInstance } from 'vitest';
+import noop from 'lodash/noop';
```
heath-freenome, line 2:
```suggestion
import noop from 'lodash/noop';

```

packages/utils/test/getTestIds.test.ts
```diff
@@ -1,13 +1,14 @@
+import uniqueIdFn from 'lodash/uniqueId';
```
heath-freenome, line 1:
```suggestion
import uniqueIdFn from 'lodash/uniqueId';

```

packages/utils/test/getDiscriminatorFieldFromSchema.test.ts
```diff
@@ -1,3 +1,5 @@
+import type { MockInstance } from 'vitest';
+import noop from 'lodash/noop';
```
heath-freenome, line 2:
```suggestion
import noop from 'lodash/noop';

```

packages/utils/test/testUtils/getTestValidator.ts
```diff
@@ -1,3 +1,4 @@
+import noop from 'lodash/noop';
```
heath-freenome, line 1:
```suggestion
import noop from 'lodash/noop';

```

packages/utils/test/schema/toPathSchemaTest.ts
```diff
@@ -1,14 +1,16 @@
+import type { MockInstance } from 'vitest';
+import noop from 'lodash/noop';
```
heath-freenome, line 2:
```suggestion
import noop from 'lodash/noop';

```

packages/utils/test/schema/getFirstMatchingOptionTest.ts
```diff
@@ -1,3 +1,4 @@
+import noop from 'lodash/noop';
```
heath-freenome, line 1:
```suggestion
import noop from 'lodash/noop';

```

packages/utils/test/parser/ParserValidator.test.ts
```diff
@@ -1,3 +1,5 @@
+import type { MockInstance } from 'vitest';
+import noop from 'lodash/noop';
```
heath-freenome, line 2:
```suggestion
import noop from 'lodash/noop';

```

packages/utils/src/getTestIds.ts
```diff
@@ -4,7 +4,7 @@ import uniqueId from 'lodash/uniqueId';
 import { TestIdShape } from './types';
 
 /** Returns an object of test IDs that can only be used in test mode. If the function is called in a test environment
- * (`NODE_ENV === 'test'`, this is set by jest) then a Proxy object will be returned. If a key within the returned
+ * (`NODE_ENV === 'test'`, this is set by Vitest) then a Proxy object will be returned. If a key within the returned
```
heath-freenome:
```suggestion
 * (`NODE_ENV === 'test'`, this is set by vitest) then a Proxy object will be returned. If a key within the returned
```

packages/core/test/testSetup.ts
```diff
@@ -1,6 +1,3 @@
 import { setImmediate } from 'timers';
 
-// atob
```
heath-freenome, line 3: The version of jsdom that works with vitest now has atob

packages/core/test/LayoutMultiSchemaField.test.tsx
```diff
@@ -31,9 +33,9 @@ import SelectWidget from '../src/components/widgets/SelectWidget';
 import { SIMPLE_ONEOF, SIMPLE_ONEOF_OPTIONS } from './testData/layoutData';
 import getTestRegistry from '../src/getTestRegistry';
 
-jest.mock('@rjsf/utils', () => ({
-  ...jest.requireActual('@rjsf/utils'),
-  getWidget: jest.fn().mockImplementation((_schema, widget, widgets) => {
+vi.mock('@rjsf/utils', async () => ({
+  ...(await vi.importActual<typeof import('@rjsf/utils')>('@rjsf/utils')),
```
nickgros: Not sure if using the argument `importActual` lets you omit the type

```suggestion
vi.mock('@rjsf/utils', async (importActual) => ({
  ...(await importActual()),
```

packages/mantine/test/Form.test.tsx
```diff
@@ -2,9 +2,9 @@ import { formTests } from '@rjsf/snapshot-tests';
 
 import WrappedForm from './WrappedForm';
 
-jest.mock('@mantine/hooks', () => ({
-  ...jest.requireActual('@mantine/hooks'),
-  useMove: jest.fn,
+vi.mock('@mantine/hooks', async () => ({
+  ...(await vi.importActual<typeof import('@mantine/hooks')>('@mantine/hooks')),
```
nickgros: Ditto if the other comment works
```suggestion
vi.mock('@mantine/hooks', async (importActual) => ({
  ...(await importActual()),
```

package.json
nickgros: Are we still using SWC? Could probably remove

packages/core/test/LayoutMultiSchemaField.test.tsx
heath-freenome: It works. I updated all of the instances of `vi.importActual()` with this pattern (although I called the variable importOriginal, per the vitest docs)

packages/utils/src/schema/getDefaultFormState.ts
nickgros: This doesn't seem like dead code to me...can you explain why `hasConst` was always `false`?

packages/utils/src/enumOptionsValueForIndex.ts
```diff
@@ -20,8 +20,8 @@ export default function enumOptionsValueForIndex<S extends StrictRJSFSchema = RJ
     return (
       valueIndex
         .map((index) => enumOptionsValueForIndex(index, allEnumOptions))
-        // Since the recursive call returns `emptyValue` when we get a bad option, only filter those out
-        .filter((val) => val !== emptyValue)
+        // The recursive call returns undefined for a bad option, filter those out
+        .filter((val) => val !== undefined)
```
nickgros, line 24: This is a logical change, is it the correct change to make?

packages/mantine/test/Form.test.tsx
heath-freenome: Got them all!

package.json
heath-freenome: Good point, let me remove it and see

heath-freenome: Looks like they can be removed

packages/utils/src/enumOptionsValueForIndex.ts
heath-freenome: I believe so. `enumOptionsValueForIndex()` has the optional `emptyValue` parameter which is returned when there is a bad option. Since we aren't actually passing in `emptyValue` to the function, then `undefined` is actually returned.

packages/utils/src/schema/getDefaultFormState.ts
heath-freenome: From Claude:

You're right that it's not logically dead code — the branch is reachable when an array schema has a `const` property and `constAsDefaults !== 'never'`. The v8 coverage provider (which replaced babel's istanbul) detected that no existing test ever exercised `getArrayDefaults` with a const array schema, so it flagged the `hasConst === true` path as uncovered.

However, restoring the code isn't the right fix. The key is that `computeDefaults` (line 249–254) **already resolves the const into `defaults` before `getArrayDefaults` is ever called**:

```typescript
if (schema[CONST_KEY] !== undefined && constAsDefaults !== 'never' && ...) {
  defaults = schema[CONST_KEY]; // const value is baked into `defaults` here
}
// ...then later:
getArrayDefaults(validator, rawSchema, props, defaults) // defaults is already ['foo', 'bar']
```

So when `getArrayDefaults` runs with a const array, `defaults` is already the const value. The early-return paths both return it correctly:
- `neverPopulate`: `return defaults ?? emptyDefault` → returns the const ✓
- `ignoreMinItemsFlagSet && !required`: `return defaults ? defaults : undefined` → returns the const ✓

The only case where the old `hasConst` check changed behavior was the pathological edge case of a schema with both `const: [...]` and `minItems > const.length` (an inherently invalid schema). There, the old code would incorrectly pad the const out to `minItems` with filler entries — the removal actually makes that case more correct.

Added 9 explicit `getArrayDefaults` tests for const array schemas (both `constAsDefaults` default and `'never'`) to close the coverage gap: the const value is returned correctly without needing to restore the `hasConst` guard.

packages/chakra-ui/test/testSetup.ts
heath-freenome: Actually, I changed my mind and renamed all 4 of them to `testSetup.ts` to match the main root one as well as the `antd` one

packages/mantine/test/setup-jest-env.ts
heath-freenome, line 1: I renamed this to testSetup.ts to match the main root one as well as the others. Only this one had so much mocking in it that it looks like an add/delete

packages/shadcn/test/setup-jest-env.ts
heath-freenome, line 1: I renamed this to testSetup.ts to match the main root one as well as the others. Only this one had so much mocking in it that it looks like an add/delete

packages/utils/test/schema/getDefaultFormStateTest.ts
```diff
@@ -2259,6 +2262,91 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
           expect(getArrayDefaults(testValidator, schema)).toBeUndefined();
         });
       });
+
+      describe('an array with const', () => {
```
heath-freenome, line 2266: These are the additional tests that Claude added regarding the `hasConst` change in `getArrayDefaults()`

nickgros (approved):

PR #5094, opened 2026-05-30
perf: memoize ArrayField/ObjectField internals and wrap IconButtons with React.memo

heath-freenome:

Related to #3183

ArrayField (NormalArray + FixedArray):
- Rename ArrayFieldItem → ArrayFieldItemInner and wrap with memo(). The as typeof cast preserves the generic signature that memo() would otherwise erase.
- Move retrieveSchema call into ArrayFieldItem via useMemo([schemaUtils, rawItemSchema, itemData]): previously called once per array item on every parent render (O(n) per keystroke); now only the item whose data changed recomputes its schema.
- Replace itemFieldPathId prop (new object every render) with parentFieldPathId (same stable reference for all items) and compute toFieldPathId() inside ArrayFieldItem via useMemo. Paired with the memoized uiOptions below, this unblocks React.memo from bailing out on unchanged items.
- Wrap uiOptions with useMemo([uiSchema, globalUiOptions]) in both NormalArray and FixedArray.
- Memoize _schemaItems, itemsSchema, formData, and canAdd in NormalArray.
- Eliminate per-item retrieveSchema calls in FixedArray: replace computed itemSchemas array with memoized _schemaItems and pass rawItemSchema per item; replace additionalSchema (used only as a boolean) with isObject(schema.additionalItems).
- Add keyedFormDataRef / errorSchemaRef (useRef pattern) so the four mutation handlers (handleAddItem, handleCopyItem, handleRemoveItem, handleReorderItems) can drop keyedFormData and errorSchema from their useCallback dependency arrays. Handlers now hold stable references across keyedFormData changes, eliminating the cascade that previously re-rendered all ArrayFieldItem instances on every keystroke.

ObjectField:
- Rename ObjectFieldProperty → ObjectFieldPropertyFn and wrap with memo(), preventing all N property components from reconciling when only one field value changes.
- Stabilize the resolved schema reference with useDeepCompareMemo: retrieveSchema returns a new object on every render; deep-comparing the result keeps the reference stable when the schema structure hasn't changed, so get(schema, [PROPERTIES_KEY, name]) returns the same reference for unchanged properties and React.memo can bail out.
- Wrap uiOptions with useMemo([uiSchema, globalUiOptions]).

IconButtons (all themes):
- Wrap every IconButton variant (base button + CopyButton, MoveDownButton, MoveUpButton, RemoveButton, ClearButton) in React.memo across core, chakra-ui, daisyui, fluentui-rc, mantine, mui, primereact, react-bootstrap, semantic-ui, and shadcn.
- Use the FnInner → memo(Fn) as typeof Fn pattern to preserve generic type parameters.
- ChakraIconButton and DaisyUIButton were already memoized; only their named wrappers needed updating.

### Performance for one key press

The performance for one keypress went from 654ms -> 401ms or about 39% improvement

#### Before

https://github.com/user-attachments/assets/ae3d2e90-2a28-4273-ad81-b55c5b62a3a3

#### After

https://github.com/user-attachments/assets/7b31152c-2107-4c51-be2f-3fcccbb48135

### Checklist


packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -222,8 +224,10 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
   const { OptionalDataControlsField } = fields;
   const formDataRef = useRef(formData);
   formDataRef.current = formData;
-  const schema: S = schemaUtils.retrieveSchema(rawSchema, formData, true);
-  const uiOptions = getUiOptions<T, S, F>(uiSchema, globalUiOptions);
+  // useDeepCompareMemo stabilizes the schema reference when formData changes but the resolved
+  // schema structure hasn't — keeping property schema refs stable so memo on ObjectFieldProperty works.
+  const schema: S = useDeepCompareMemo<S>(schemaUtils.retrieveSchema(rawSchema, formData, true));
```
heath-freenome: Maybe use a `useMemo()` to avoid the retrieveSchema recalculation

packages/shadcn/src/IconButton/IconButton.tsx
```diff
@@ -10,14 +11,7 @@ export type ShadIconButtonProps<
...
 
-/** Base button component that renders a Shadcn button with an icon for RJSF form actions.
- * This component serves as the foundation for other specialized buttons used in array operations.
- * It combines RJSF's IconButtonProps with Shadcn's ButtonProps to provide a consistent styling
- * and behavior across the form.
- *
- * @param props - The combined props from RJSF IconButtonProps and Shadcn ButtonProps, including icon and event handlers
- */
-export default function IconButton<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(
+function IconButtonFn<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(
```
heath-freenome:
```suggestion
/** Base button component that renders a Shadcn button with an icon for RJSF form actions.
 * This component serves as the foundation for other specialized buttons used in array operations.
 * It combines RJSF's IconButtonProps with Shadcn's ButtonProps to provide a consistent styling
 * and behavior across the form.
 *
 * @param props - The combined props from RJSF IconButtonProps and Shadcn ButtonProps, including icon and event handlers
 */
 function IconButtonFn<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(
```

nickgros:
@heath-freenome maybe instead we use the React compiler? https://react.dev/reference/react-compiler/compiling-libraries

heath-freenome:
> @heath-freenome maybe instead we use the React compiler? https://react.dev/reference/react-compiler/compiling-libraries

That requires going to React 19... So part of v7? Was there something you'd rather see done?

nickgros:
> That requires going to React 19... So part of v7? Was there something you'd rather see done?

We can set a target version of 17: https://react.dev/reference/react-compiler/compiling-libraries#configure-target-version.

Just wondering if you think we could do that and not worry manual memoization

heath-freenome:
> > That requires going to React 19... So part of v7? Was there something you'd rather see done?
> 
> We can set a target version of 17: https://react.dev/reference/react-compiler/compiling-libraries#configure-target-version.
> 
> Just wondering if you think we could do that and not worry manual memoization

Given that I just ripped out babel, we'd first have to build stuff with vite rather than esbuild. And then add the react compiler to that. Again something for v7 so that we can kill UMD and maybe CJS?

nickgros (approved):

PR #5106, opened 2026-06-02
feat(lint): add airbnb-extended errors.ts and best-practices.ts rules to oxlintrc and fix violations

heath-freenome:
- Enables no-console (error), no-else-return, no-param-reassign, guard-for-in,
    no-throw-literal, no-void, no-self-compare, prefer-object-has-own, radix,
    no-useless-return, no-sequences, default-case, class-methods-use-this,
    max-classes-per-file, and ~40 other best-practices rules matching
    eslint-config-airbnb-extended. 
- Fixes all violations across 73 files: extracts
    param reassignments to local variables, removes redundant else-after-return,
    converts test class components to function components, makes non-this class
    methods static, and adds oxlint-disable-next-line comments for intentional
    console calls.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
### Checklist


packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -902,11 +903,13 @@ export default function ArrayField<T = any, S extends StrictRJSFSchema = RJSFSch
       if (errorSchemaRef.current) {
         newErrorSchema = {};
         for (const idx in errorSchemaRef.current) {
-          const i = parseInt(idx);
-          if (index === undefined || i < index) {
-            set(newErrorSchema, [i], errorSchemaRef.current[idx]);
-          } else if (i >= index) {
-            set(newErrorSchema, [i + 1], errorSchemaRef.current[idx]);
+          if (Object.hasOwn(errorSchemaRef.current, idx)) {
```
nickgros, line 906: Seems odd to need this `Object.hasOwn` when it originates from a `for ... in` loop

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -128,14 +128,12 @@ function ObjectFieldPropertyFn<T = any, S extends StrictRJSFSchema = RJSFSchema,
...
-        // fields which are "mandated" by the schema, these fields can be set to undefined by clicking a "delete field"
-        // button, so set empty values to the empty string.
-        value = '' as unknown as T;
-      }
-      onChange(value, path, newErrorSchema, id);
+      // Don't set value = undefined for fields added by additionalProperties. Doing so removes them from the
+      // formData, which causes them to completely disappear (including the input field for the property name). Unlike
+      // fields which are "mandated" by the schema, these fields can be set to undefined by clicking a "delete field"
+      // button, so set empty values to the empty string.
+      const normalizedValue = value === undefined && addedByAdditionalProperties ? ('' as unknown as T) : value;
```
nickgros: Makes sense to not change `value` but IMO the ternary is harder to read here. Could we still do this:

```ts
let normalizedValue = value
if (value === undefined && addedByAdditionalProperties) {
  normalizedValue = ''
}
onChange(normalizedValue, path, newErrorSchema, id);
```

packages/core/src/components/widgets/FileWidget.tsx
```diff
@@ -94,7 +94,8 @@ function FileWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends
   const handleOnChangeEvent = (event: ChangeEvent<HTMLInputElement>) => {
     if (event.target.files) {
       // handleChange is async; DOM event handlers are void-returning, so we intentionally don't await
-      void handleChange(event.target.files);
+      // oxlint-disable-next-line no-misused-promises, no-floating-promises
+      handleChange(event.target.files);
```
nickgros: what was wrong with `void`? IMO that is better than a disable lint comment

packages/core/src/components/Form.tsx
```diff
@@ -791,7 +792,7 @@ export default class Form<
    * @return - The `extraErrors` and `customErrors` merged into the `schemaValidation`
    * @private
    */
-  private mergeErrors(
+  private static mergeErrors<T = any>(
```
nickgros, line 795: Making these static is great. Surprised we didn't catch that before

packages/mantine/src/widgets/FileWidget.tsx
```diff
@@ -36,7 +36,8 @@ export default function FileWidget<T = any, S extends StrictRJSFSchema = RJSFSch
     (files: any) => {
       if (typeof files === 'object') {
         // handleChange is async; DOM event handlers are void-returning, so we intentionally don't await
-        void handleChange(files);
+        // oxlint-disable-next-line no-misused-promises, no-floating-promises
+        handleChange(files);
```
nickgros: ditto on this `void`

packages/playground/src/components/Playground.tsx
```diff
@@ -157,6 +157,7 @@ export default function Playground({ themes, validators }: PlaygroundProps) {
         setLoaded(true);
       } catch (error) {
         alert('Unable to load form setup data.');
+        // oxlint-disable-next-line no-console
```
nickgros: may just want to disable this in the entire file, console is fine in the playground

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -742,6 +743,7 @@ export function getDefaultBasedOnSchemaType<
     case 'array': {
       return getArrayDefaults(validator, rawSchema, computeDefaultsProps, defaults as T[]);
     }
+    // no default
```
nickgros: maybe should explicitly return?

packages/utils/src/schema/omitExtraData.ts
```diff
@@ -176,6 +176,7 @@ export default function omitExtraData<
    * @returns - `target` after all schema-defined properties have been processed
    */
   function handleObject(childSchema: S, source: GenericObjectType, target: GenericObjectType): GenericObjectType {
+    const result = target;
```
nickgros: the entire purpose of these functions is to mutate `target`, so I feel like assigning it to `result` and mutating `result` is blindly complying with the linter and hiding the fact that we are mutating the argument...I think I would prefer using comments to disable the lint rule in these cases.

packages/core/src/components/fields/ArrayField.tsx
nickgros: Oh, that's literally the point of the rule: https://eslint.org/docs/latest/rules/guard-for-in

packages/utils/test/dataURItoBlob.test.ts
```diff
@@ -28,7 +28,8 @@ describe('dataURItoBlob()', () => {
   });
   it('should throw if the body is not a valid dataURI', () => {
     expect(() => dataURItoBlob('Hello%20World')).toThrow(new Error('File is invalid: URI must be a dataURI'));
-    expect(() => dataURItoBlob('javascript:alert()')).toThrow(new Error('File is invalid: URI must be a dataURI'));
+    const jsUrl = ['javascript', ':', 'alert()'].join('');
```
nickgros: Is this just avoiding a lint rule to not use `alert`? 😆  A disable comment would be fine

packages/validator-ajv8/src/validator.ts
```diff
@@ -117,15 +117,16 @@ export default class AJV8Validator<
         // `AJV8Validator#transformRJSFValidationErrors` replaces property names
         // with `title` or `ui:title`. See #4348, #4349, #4387, and #4402.
         (compiledValidator.errors ?? []).forEach((error) => {
+          const { params } = error;
```
nickgros: seems like another case where extracting the variable just makes it less obvious that we are mutating the argument

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: And, as it turns out a later rule coming in a new PR will switch that around to something else :)

nickgros (approved):

PR #5113, opened 2026-06-04
Additional eslint config airbnb extended

heath-freenome:

New rules enabled in .oxlintrc.json:

**ESLint style rules:**
- `func-names` (warn) – require named function expressions
- `new-cap` – require constructors to start with a capital letter
- `no-bitwise` – disallow bitwise operators (disable comments where intentional)
- `no-continue` – disallow `continue` statements; refactored loops to use `if` blocks
- `no-lonely-if` – disallow `if` as the only statement in an `else` block
- `no-multi-assign` – disallow chained assignment (`a = b = value`)
- `no-nested-ternary` – disallow nested ternary expressions
- `no-plusplus` – disallow `++`/`--`; replaced with `+= 1` / `-= 1`
- `no-underscore-dangle` – disallow leading/trailing underscores; renamed `_onChange/Blur/Focus/Clear` → `handleChange/Blur/Focus/Clear` across all theme packages, renamed private class fields; allow-listed `__errors` (AJV schema property) and framework-private factory function names
- `operator-assignment` – require shorthand assignment operators
- `unicode-bom` – disallow byte order marks

**React rules:**
- `react/button-has-type` – require explicit `type` on `<button>` elements
- `react/display-name` – require display names on all components
- `react/forward-ref-uses-ref` – require `ref` usage in `forwardRef` components
- `react/jsx-boolean-value` – disallow redundant `={true}` on boolean props
- `react/jsx-curly-brace-presence` – disallow unnecessary JSX curly braces around string literals
- `react/jsx-filename-extension` – restrict JSX to `.jsx`/`.tsx` files
- `react/jsx-fragments` – require fragment shorthand syntax `<>…</>`
- `react/jsx-no-constructed-context-values` – disallow object literals as context values; fixed with `useMemo`
- `react/jsx-no-script-url` – disallow `javascript:` URLs
- `react/jsx-no-useless-fragment` – disallow fragments wrapping a single child
- `react/jsx-pascal-case` – require PascalCase component names; renamed `_SchemaField` → `SchemaFieldComponent`
- `react/jsx-props-no-spread-multi` – disallow spreading props multiple times
- `react/no-array-index-key` – disallow array index as JSX key; switched enum option widgets to use `String(option.value)`, ObjectFieldTemplate to use `element.name`, and added disable comments where index is genuinely unavoidable (ErrorList, AltDateWidget, RatingWidget, etc.)
- `react/no-danger` (warn) – warn on `dangerouslySetInnerHTML`
- `react/no-did-update-set-state` – disallow `setState` in `componentDidUpdate`; added guarded disable comments in `Form` and `MultiSchemaField`
- `react/no-namespace` – disallow XML namespaces in JSX
- `react/no-redundant-should-component-update` – disallow `shouldComponentUpdate` on PureComponent
- `react/no-this-in-sfc` – disallow `this` in stateless function components
- `react/no-unstable-nested-components` – disallow components defined inside render
- `react/no-will-update-set-state` – disallow `setState` in `componentWillUpdate`
- `react/prefer-es6-class` – require ES6 class syntax for React components
- `react/require-render-return` – require explicit `return` in `render()`
- `react/self-closing-comp` – require self-closing tags for components/elements with no children
- `react/state-in-constructor` – require state initialization in the constructor
- `react/style-prop-object` – require `style` prop to be an object
- `react/void-dom-elements-no-children` – disallow children on void DOM elements

**TypeScript rules (TS files override):**
- `typescript/class-literal-property-style` – require fields over accessors for class literal properties
- `typescript/dot-notation` – prefer `obj.foo` over `obj['foo']` (replaces base `dot-notation`)
- `typescript/no-confusing-non-null-assertion` – disallow non-null assertions in positions that look like comparisons
- `typescript/no-for-in-array` – disallow `for…in` over arrays
- `typescript/no-inferrable-types` – disallow unnecessary type annotations on initialised variables
- `typescript/no-mixed-enums` – disallow enums that mix string and number members
- `typescript/no-unnecessary-parameter-property-assignment` – disallow redundant property assignments in constructors
- `typescript/no-unnecessary-qualifier` – disallow unnecessary namespace qualifiers
- `typescript/no-unsafe-enum-comparison` – disallow unsafe enum comparisons
- `typescript/no-unsafe-unary-minus` – disallow unary minus on non-numeric types
- `typescript/prefer-find` – prefer `Array.find` over `filter()[0]`

**Ignore patterns / overrides:**
- Ignore entire `**/docs/**` (avoids false-positive `tsconfig-error` from Docusaurus base tsconfig resolving `baseUrl`)
- Disable `typescript/no-var-requires` in docs (Docusaurus uses webpack `require()` for SVG imports)

### Checklist


packages/core/test/Form.test.tsx
```diff
@@ -240,7 +240,7 @@ describeRepeated('Form common', (createFormComponent) => {
       const { node } = createFormComponent({ schema, idPrefix: 'rjsf' });
       const inputs = node.querySelectorAll('input');
       const ids = [];
-      for (let i = 0, len = inputs.length; i < len; i++) {
+      for (let i = 0, len = inputs.length; i < len; i += 1) {
```
nickgros: Could tweak the rule to [`allowForLoopAfterthoughts`](https://eslint.org/docs/latest/rules/no-plusplus#options)

nickgros (approved):

packages/core/test/Form.test.tsx
heath-freenome: Good call, changes made :)

Issue #4262, opened 2024-08-01
Application freeze


jakubkanna:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [ ] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

5.x

### Current Behavior

Using json schema generated from Prisma database causing application to freeze.

### Expected Behavior

It should render form. 

### Steps To Reproduce

1. Generate schema via [prisma-json-schema-generator](https://www.npmjs.com/package/prisma-json-schema-generator)
2. Copy paste schema into [playground](https://rjsf-team.github.io/react-jsonschema-form/)

### Environment

```markdown
Playground
```

### Anything else?

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "GeneralSection": {
      "type": "object",
      "properties": {
        "id": {
          "type": "integer"
        },
        "title": {
          "type": "string"
        },
        "description": {
          "type": ["string", "null"]
        },
        "published": {
          "type": "boolean",
          "default": false
        },
        "slug": {
          "type": "string"
        },
        "tags": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Tag"
          }
        },
        "createdAt": {
          "type": "string",
          "format": "date-time"
        },
        "updatedAt": {
          "type": ["string", "null"],
          "format": "date-time"
        },
        "Project": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Project"
          }
        },
        "Work": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Work"
          }
        },
        "Post": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Post"
          }
        }
      }
    },
    "ImageRef": {
      "type": "object",
      "properties": {
        "etag": {
          "type": "string"
        },
        "public_id": {
          "type": "string"
        },
        "type": {
          "type": "string",
          "default": "IMAGE",
          "enum": ["IMAGE", "VIDEO", "THREE_D"]
        },
        "cld_url": {
[...]

PR #5118, opened 2026-06-07
Stop infinite renders caused by circular $ref object properties

### LINKED ISSUE #3907: <title>
### Prerequisites

### What theme are you using?

core

### Version

5.x

### Current Behavior

Schema:
```
{
  "type": "object",
  "properties": {
    "child": {
      "$ref": "#"
    }
  }
}
```

### Expected Behavior

Not to hang aka eventually stack overflow

### Steps To Reproduce

Paste that schema into the playground, that schema is a simplification of what happens when I attempt to use the draft-07 json-schema meta-schema.

### Environment

```markdown
N/A
```

### Anything else?

@heath-freenome , as requested in #3560.  I dont really know anything about react anymore, but just poking around for a bit, the `Form` seems to be trying to render a recursive reference. It seems like the correct behavior should either be an error something like: "Hey, we don't allow recursive references", or just build the feature which marks definitions as seen, which may require a bit of a change in strat from DFS to BFS of the schema tree.

Also, im not sure if its a goal of this project, but it seems to me some good test cases are actually the json-schema meta-schemas. But I recon that is a separate issue altogether.

heath-freenome:

Fixes #3907 and #4262
- Added GenericSymbolObjectType and RJSFMarkedSchema so call sites can cast to it when reading Symbol keys
- Adds RJSF_REF_CYCLE_KEY ('__rjsf_ref_cycle') to detect and terminate self-referential schemas that would otherwise render infinitely.
- In resolveAllReferences, when a $ref is encountered that's already in the recurseList AND we're resolving in simple (non-resolveAnyOfOrOneOfRefs) mode, the schema is tagged with __rjsf_ref_cycle: true instead of silently returning the unresolved $ref. This only fires in a direct object-property context (!resolveAnyOfOrOneOfRefs), where rendering is unconditional and cycles are genuinely infinite. Array-items and anyOf/oneOf contexts remain untagged since they are data-driven and terminate naturally.
- SchemaField checks _schema[RJSF_REF_CYCLE_KEY] after its useCallback hook (to satisfy React rules of hooks) and renders a CyclicSchemaField instead of recursing. CyclicSchemaField renders the CyclicSchemaExpandTemplate, which shows the user a message and an expand button to load the next cycle break.
- CyclicSchemaExpandTemplate is implemented for all theme packages: antd, chakra-ui, daisyui, fluentui-rc, mantine, mui, primereact, react-bootstrap, semantic-ui, and shadcn, each using their native UI library components.
- ADDITIONAL_PROPERTY_FLAG, RJSF_REF_KEY, and RJSF_REF_CYCLE_KEY are now `Symbol` constants instead of plain strings.
  -  This keeps RJSF's internal schema annotations invisible to:
      - JSON.stringify (and thus AJV/ATA schema compilation caches)
      - JSON Schema validators — AJV, ATA, and any future validator that only enumerates string-keyed properties
      - deepEquals (fast-equals) — Symbol keys are ignored, so the validator cache in ATAValidator and schemaParser cycle detection continue to work correctly without needing to strip these keys before comparison
- isRootSchema() drops the omit(resolvedSchema, RJSF_REF_KEY) call for
  the same reason — deepEquals already ignores Symbol keys
- retrieveSchema(): three lodash set() calls that used ADDITIONAL_PROPERTY_FLAG
  as a path component are replaced with direct property assignment, since
  lodash set() stringifies path elements and cannot address Symbol keys
- getDisplayLabel(), SchemaField: lodash get() calls replaced with direct
  property access (lodash get() also stringifies its path argument)
- ObjectField: lodash has() replaced with optional-chained property access
- validator-ajv8/createAjvInstance: removed the three addKeyword() calls for
  the now-Symbol keys — AJV never sees Symbol properties, so registration
  as custom keywords is unnecessary. RJSF_ADDITIONAL_PROPERTIES_FLAG (still
  a string) retains its addKeyword() call
- Updated snapshots for the new tests
- Updated documentation for the new feature
- Updated the CHANGELOG.md

### Checklist

### Preview

https://github.com/user-attachments/assets/4121c0fb-e7b8-4cfb-a23f-137e1c9301a6


x0k:
1. In the video, clicking the button adds two fields.  
2. If the initial data is `{"child":{"child":{"child":{"child":{}}}}}`, only two levels are rendered, plus an expand button.

In SJSF, I tried to solve this problem by generating a `schemaLocation` for each field in order to detect cycles at the form level (rather than in `utils`), but so far I have not found a simple way to guarantee correct JSON Pointer generation in all cases (`then`/`else`, `dependencies`, etc.).

heath-freenome:
> 1. In the video, clicking the button adds two fields.
> 
>     2. If the initial data is `{"child":{"child":{"child":{"child":{}}}}}`, only two levels are rendered, plus an expand button.
> 
> 
> In SJSF, I tried to solve this problem by generating a `schemaLocation` for each field in order to detect cycles at the form level (rather than in `utils`), but so far I have not found a simple way to guarantee correct JSON Pointer generation in all cases (`then`/`else`, `dependencies`, etc.).

Hi @x0k... Yeah it's not until we encounter the second instance of the same key that the cycle is detected, hence the information showing up 2x for each cycle since the marker is on that second item. I've been debating doing something with the recursion such that upon the cycle detection, I can somehow go back to the first instance and mark that one instead. I may spend some more time on that once this gets merged

x0k:
> hashForSchema now filters out keys starting with RJSF_REF_KEY before hashing so that internal cycle-tracking metadata does not affect the computed hash.

FYI, if you use `Symbol`s, such keys are automatically ignored by `JSON.stringify`.

x0k:
`ajv.addKeyword` does not work with `Symbol`s, so this would only work with `strict: false` (which is the default, [AJV docs](https://ajv.js.org/strict-mode.html#strict-mode)).

Actually, I think that AJV in strict mode will ignore symbols even without `addKeyword`. I'll post a demo as soon as I get to a computer.

x0k:
Here is a demo - [StackBlitz](https://stackblitz.com/edit/node-tabbmnc7?file=index.js). It works with AJV and other JSON Schema validators.

packages/utils/src/constants.ts
```diff
@@ -1,9 +1,10 @@
...
- * utility functions. In addition to those keys, there are the special `ADDITIONAL_PROPERTY_FLAG` and
- * `RJSF_ADDITIONAL_PROPERTIES_FLAG` flags that is added to a schema under certain conditions by the `retrieveSchema()`
- * utility.
+ * utility functions. In addition to those keys, there are the special `ADDITIONAL_PROPERTY_FLAG`,
+ * `RJSF_REF_KEY`, and `RJSF_REF_CYCLE_KEY` Symbols that are added to a schema under certain conditions by the
+ * `retrieveSchema()` utility. Using Symbols keeps these internal markers invisible to JSON serialisation and
+ * JSON Schema validators (AJV, ATA, etc.), which only enumerate string-keyed properties.
  */
-export const ADDITIONAL_PROPERTY_FLAG = '__additional_property';
+export const ADDITIONAL_PROPERTY_FLAG: unique symbol = Symbol('__additional_property');
```
heath-freenome:
```suggestion
export const ADDITIONAL_PROPERTY_FLAG = Symbol('__additional_property');
```

packages/utils/src/constants.ts
```diff
@@ -25,7 +26,8 @@ export const READONLY_KEY = 'readonly';
 export const REQUIRED_KEY = 'required';
 export const SUBMIT_BTN_OPTIONS_KEY = 'submitButtonOptions';
 export const REF_KEY = '$ref';
-export const RJSF_REF_KEY = '__rjsf_ref';
+export const RJSF_REF_KEY: unique symbol = Symbol('__rjsf_ref');
+export const RJSF_REF_CYCLE_KEY: unique symbol = Symbol('__rjsf_ref_cycle');
```
heath-freenome:
```suggestion
export const RJSF_REF_KEY = Symbol('__rjsf_ref');
export const RJSF_REF_CYCLE_KEY = Symbol('__rjsf_ref_cycle');
```

packages/utils/src/hashForSchema.ts
```diff
@@ -47,10 +47,12 @@ export function hashObject(object: unknown): string {
...
  * in consistent order before stringify to prevent different hash ids for the same schema.
+ * Symbol-keyed properties (RJSF_REF_KEY, RJSF_REF_CYCLE_KEY, ADDITIONAL_PROPERTY_FLAG) are
+ * automatically excluded by JSON.stringify, so no special filtering is needed.
  *
  * @param schema - The schema for which the hash is desired
  * @returns - The string obtained from the hash of the stringified schema
  */
 export default function hashForSchema<S extends StrictRJSFSchema = RJSFSchema>(schema: S) {
-  return hashObject(schema);
+  return hashString(sortedJSONStringify(schema));
```
heath-freenome, line 57:
```suggestion
  return hashObject(schema);
```

packages/utils/src/mergeSchemas.ts
```diff
@@ -15,7 +15,7 @@ import type { GenericObjectType } from './types';
  */
 export default function mergeSchemas(obj1: GenericObjectType, obj2: GenericObjectType) {
```
heath-freenome, line 16:
```suggestion
export default function mergeSchemas(obj1: GenericSymbolObjectType, obj2: GenericSymbolObjectType) {
```

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -394,7 +394,7 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
     title: uiOptions.label === false ? '' : templateTitle,
     description: uiOptions.label === false ? undefined : description,
     properties: orderedProperties.map((propertyName) => {
-      const addedByAdditionalProperties = has(schema, [PROPERTIES_KEY, propertyName, ADDITIONAL_PROPERTY_FLAG]);
+      const addedByAdditionalProperties = Boolean(get(schema.properties, [propertyName, ADDITIONAL_PROPERTY_FLAG]));
```
heath-freenome: Fix this to do:

```suggestion
      const addedByAdditionalProperties = Boolean((schema.properties?.propertyName as RJSFMarkedSchema)[ADDITIONAL_PROPERTY_FLAG]));
```

nickgros (approved):

packages/utils/src/schema/retrieveSchema.ts
```diff
@@ -384,7 +391,18 @@ export function resolveAllReferences<S extends StrictRJSFSchema = RJSFSchema>(
...
-        acc[key] = resolveAllReferences(value as S, rootSchema, childList, currentBaseURI, resolveAnyOfOrOneOfRefs);
+        // Mark cycles only when NOT in resolveAnyOfOrOneOfRefs mode. When resolveAnyOfOrOneOfRefs=true
+        // (e.g. from ObjectField), options are resolved against a shared recurseList that accumulates
+        // refs across branches, causing false positives. In simple (non-anyOf) resolution, a $ref cycle
+        // in an object property always causes an infinite render loop and must be caught.
+        acc[key] = resolveAllReferences(
+          value as S,
+          rootSchema,
+          childList,
+          baseURI,
```
x0k, line 402: @heath-freenome  Is the change from `currentBaseURI` to `baseURI` intentional? I'm asking just in case, since it looks like a bug.

heath-freenome: Hmmm, very good question. Let me update my current PR with this change and see what happens

PR #5121, opened 2026-06-12
Fix oneOf rendering when type is on the parent schema (#5119)

### LINKED ISSUE #5119: oneOf/anyOf with parent-level type renders an extra input and produces incorrect formData
### Prerequisites

### What theme are you using?

validator-ajv8

### Version

6.6.2

### Current Behavior

Hi rjsf team,

I found an issue when using oneOf or anyOf with a field that already defines its type at the parent level.

I will include a playground example, but the schema looks like this:

```
{
  "title": "Network config",
  "type": "object",
  "required": ["ipAddress"],
  "properties": {
    "ipAddress": {
      "title": "IP address",
      "type": "string",
      "oneOf": [
        {
          "title": "IPv4",
          "pattern": "^(25[0-5]|2[0-4][0-9]|1?[0-9]{1,2})(\\.(25[0-5]|2[0-4][0-9]|1?[0-9]{1,2})){3}$",
          "examples": ["192.168.1.10"]
        },
        {
          "title": "IPv6",
          "pattern": "^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|::1|::)$",
          "examples": ["2001:0db8:85a3:0000:0000:8a2e:0370:7334"]
        }
      ]
    }
  }
}
```

The same issue also happens if oneOf is replaced with anyOf.

Since ipAddress is declared as "type": "string", I would expect rjsf to render it as a string field and use the selected oneOf/anyOf branch only to apply the additional validation rules.

**Current behavior**

rjsf renders an unexpected extra input above the oneOf/anyOf selector.

It also renders the selector and then shows an unsupported field error for the selected branch:

Unsupported field schema for field root_ipAddress: Unknown field type undefined.

The unsupported schema shown is only the selected branch:

```
{
  "title": "IPv4",
  "pattern": "^(25[0-5]|2[0-4][0-9]|1?[0-9]{1,2})(\\.(25[0-5]|2[0-4][0-9]|1?[0-9]{1,2})){3}$",
  "examples": ["192.168.1.10"]
}

```
The bigger issue is that typing into the unexpected input can produce form data like this:

```
{
  "ipAddress": {
    "XxxOf": "192.168.1.10"
  }
}
```

This does not match the schema, because ipAddress is explicitly declared as a string.

### Expected Behavior

I would expect rjsf to:

- Use the parent-level type when rendering oneOf/anyOf branches that do not define their own type.
- Avoid rendering the extra input above the oneOf/anyOf selector.
- Preserve the declared data type in formData.

The resulting form data should remain a string:

```
{
  "ipAddress": "192.168.1.10"
}

```
Since the parent schema already declares "type": "string", rjsf should be able to treat each oneOf/anyOf branch as string-based for rendering purposes, unless a branch explicitly defines a different type.

For this example, a reasonable default UI could be:

```
IP address type: [IPv4 | IPv6]
IP address:      [________________]
```

or any other layout decided by rjsf, but there should be only one real input bound to ipAddress.

### Steps To Reproduce

[Playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0aXRsZSI6Ik5ldHdvcmsgY29uZmlnIiwidHlwZSI6Im9iamVjdCIsInJlcXVpcmVkIjpbImlwQWRkcmVzcyJdLCJwcm9wZXJ0aWVzIjp7ImlwQWRkcmVzcyI6eyJ0aXRsZSI6IklQIGFkZHJlc3MiLCJ0eXBlIjoic3RyaW5nIiwib25lT2YiOlt7InRpdGxlIjoiSVB2NCIsInBhdHRlcm4iOiJeKDI1WzAtNV18MlswLTRdWzAtOV18MT9bMC05XXsxLDJ9KShcXC4oMjVbMC01XXwyWzAtNF1bMC05XXwxP1swLTldezEsMn0pKXszfSQiLCJleGFtcGxlcyI6WyIxOTIuMTY4LjEuMTAiXX0seyJ0aXRsZSI6IklQdjYiLCJwYXR0ZXJuIjoiXigoWzAtOWEtZkEtRl17MSw0fTopezd9WzAtOWEtZkEtRl17MSw0fXw6OjF8OjopJCIsImV4YW1wbGVzIjpbIjIwMDE6MGRiODo4NWEzOjAwMDA6MDAwMDo4YTJlOjAzNzA6NzMzNCJdfV19fX0sInVpU2NoZW1hIjp7fSwidGhlbWUiOiJkZWZhdWx0IiwibGl2ZVNldHRpbmdzIjp7InNob3dFcnJvckxpc3QiOiJ0b3AiLCJ2YWxpZGF0ZSI6ZmFsc2UsImRpc2FibGVkIjpmYWxzZSwibm9IdG1sNVZhbGlkYXRlIjpmYWxzZSwicmVhZG9ubHkiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In0sInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOmZhbHNlLCJpZFByZWZpeCI6InJvb3QiLCJpZFNlcGFyYXRvciI6Il8ifSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJCbGFuayJ9)

### Environment

```markdown
- OS: Ubuntu 24.04
- Node: 22.22.3
- npm: 10.9.8
```

### Anything else?

_No response_

nickgros:
Fixes #5119

- Suppress the duplicate field input that appeared when a primitive or array type (e.g. type: "string") was combined with a non-select oneOf/anyOf at the same schema level
- Propagate the parent schema's type to option sub-schemas that omit their own, so the correct widget renders for the selected option instead of FallbackField


packages/core/src/components/fields/MultiSchemaField.tsx
```diff
@@ -198,10 +198,19 @@ class AnyOfField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
     let optionSchema: S | undefined | null;
 
     if (option) {
-      // merge top level required field
-      const { required: schemaRequired } = schema;
```
nickgros, line 202: We were already merging `required` down into the subschema fields, so let's do it for `type` too.

packages/core/src/components/fields/MultiSchemaField.tsx
```diff
@@ -198,10 +198,19 @@ class AnyOfField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends For
...
+      const { required: schemaRequired, type: schemaType } = schema;
+      const parentProps: Partial<S> = {};
+      if (schemaRequired) parentProps.required = schemaRequired as S['required'];
```
heath-freenome: Hmmm, we need to tell Claude to respect our lint rules
```suggestion
      if (schemaRequired) {
        parentProps.required = schemaRequired as S['required'];
      }
```

packages/core/src/components/fields/SchemaField.tsx
```diff
@@ -77,9 +77,13 @@ function getFieldComponent<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
...
-  // If the type is not defined and the schema uses 'anyOf' or 'oneOf', don't
-  // render a field and let the MultiSchemaField component handle the form display
-  if (!componentName && (schema.anyOf || schema.oneOf)) {
+  // If the schema uses 'anyOf' or 'oneOf' and is not a pure select (all-constant options),
+  // let the MultiSchemaField component handle the form display entirely.
+  // ObjectField is excluded: it renders shared properties (defined at the parent schema
+  // level) alongside the XxxOfField option selector.
+  // All other field types — including primitives and arrays — have no shared renderable
+  // properties, so the outer FieldComponent would only produce a spurious duplicate input.
+  if ((schema.anyOf || schema.oneOf) && !registry.schemaUtils.isSelect(schema) && componentName !== 'ObjectField') {
```
heath-freenome: Destructure, schemaUtils out of registry above at line 62

packages/core/test/oneOf.test.tsx
```diff
@@ -1902,6 +1902,113 @@ describe('oneOf', () => {
       }
     });
   });
+  describe('primitive type with non-select oneOf', () => {
```
heath-freenome, line 1905: I would expect the `anyOf` tests above to have a nearly identical set of tests... ?

CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.7.0
+
+## @rjsf/core
```
heath-freenome, line 29: This will conflict with my changes

heath-freenome (approved):

packages/core/src/components/fields/MultiSchemaField.tsx
nickgros: Why didn't this fail in CI?

heath-freenome: Oops, I forgot to add `curly` to the lint config... Can you do that?

heath-freenome: Nevermind: I just pushed this https://github.com/rjsf-team/react-jsonschema-form/pull/5123. Feel free to approve/merge/rebase

heath-freenome: Also, OxFmt won't automatically expand `if (foo) { asfsdfsa }` to
```
if (foo) {
  asfsdfa
}
``` 
unless the body is wider than 120 characters, sadly

PR #5133, opened 2026-06-26
Fix playground

nickgros:
Upgrading to Vite 8.1.0 seems to break the playground, related to bundling @ant-design/icons. Revert to 8.0.16 and will file a report with Vite if we cannot find a matching issue.


packages/playground/package.json
```diff
@@ -31,9 +31,6 @@
   "engines": {
     "node": ">=20"
   },
-  "peerDependencies": {
-    "react": "^18.3.1"
-  },
```
nickgros, line 36: since this is an SPA that does not have dependents, `peerDependencies` makes no sense

packages/playground/package.json
```diff
@@ -121,4 +121,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
+}
```
nickgros:
```suggestion
}

```

packages/playground/package.json
```diff
@@ -97,7 +97,7 @@
     "react-transform-catch-errors": "^1.0.2",
     "react-transform-hmr": "^1.0.4",
     "source-map-loader": "^5.0.0",
-    "vite": "^8.1.0"
+    "vite": "8.0.16"
```
nickgros, line 100: Fixes the issue. Will find/file a report with Vite

packages/playground/main
nickgros, line 1: Not sure what this empty file was doing here

heath-freenome (approved):

nickgros:
Created an issue with Vite: https://github.com/vitejs/vite/issues/22779

PR #5138, opened 2026-06-28
fix(core): prevent cleared fields from re-defaulting when another field is cleared (#5125 regression)

### LINKED ISSUE #5125: About form behavior when empty input is performed
### Prerequisites

### What theme are you using?

mui

### Version

">=6.5.3"

### Current Behavior

For fields with `default` set in JSONSchema, the input cannot be empty unless `ui:emptyValue` is specified in UISchema.

See: [Example1](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJzb21lU3RyaW5ncyI6IkNodWNrIiwic29tZU51bWJlcnMiOjF9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInJlcXVpcmVkIjpbInNvbWVTdHJpbmdzIl0sInByb3BlcnRpZXMiOnsic29tZVN0cmluZ3MiOnsidHlwZSI6InN0cmluZyIsImRlZmF1bHQiOiJDaHVjayJ9LCJzb21lTnVtYmVycyI6eyJ0eXBlIjoibnVtYmVyIiwiZGVmYXVsdCI6MX19fSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In0sInVzZUZhbGxiYWNrRmllbGQiOmZhbHNlLCJ1c2VGYWxsYmFja1VpRm9yVW5zdXBwb3J0ZWRUeXBlIjpmYWxzZSwibGl2ZVZhbGlkYXRlIjpmYWxzZSwiaWRQcmVmaXgiOiJyb290IiwiaWRTZXBhcmF0b3IiOiJfIn0sInZhbGlkYXRvciI6IkFKVjgiLCJzYW1wbGVOYW1lIjoiQmxhbmsifQ==)

Also, if you specify `ui:emptyValue` in UISchema, it will not be able to recognize the empty value, so `required` will not work in form validation, or if you specify `null`, the validation result will be displayed as a type error instead of a `required` error.

See: [Example2](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJzb21lU3RyaW5ncyI6IkNodW5rIiwic29tZU51bWJlcnMiOjF9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInJlcXVpcmVkIjpbInNvbWVTdHJpbmdzIl0sInByb3BlcnRpZXMiOnsic29tZVN0cmluZ3MiOnsidHlwZSI6InN0cmluZyIsImRlZmF1bHQiOiJDaHVjayJ9LCJzb21lTnVtYmVycyI6eyJ0eXBlIjoibnVtYmVyIiwiZGVmYXVsdCI6MX19fSwidWlTY2hlbWEiOnsic29tZVN0cmluZ3MiOnsidWk6ZW1wdHlWYWx1ZSI6IiJ9LCJzb21lTnVtYmVycyI6eyJ1aTplbXB0eVZhbHVlIjpudWxsfX0sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJjb25zdEFzRGVmYXVsdHMiOiJhbHdheXMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOiJvbkNoYW5nZSIsImlkUHJlZml4Ijoicm9vdCIsImlkU2VwYXJhdG9yIjoiXyJ9LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IkJsYW5rIn0=)

In addition, up to v6.5.2, an empty field in the form data as an argument of the `onChange` event was `undefined` with a key, but from v6.5.3 the form data is passed without the field itself.Currently, in this development environment, there are many problems related to state management due to the RJSF version update.

**JSON Schema:**

```json
{
  "type": "object",
  "properties": {
    "someStrings": { "type": "string" },
    "someNumbers": { "type": "number" }
  }
}
```

**v6.5.2:**

```jsx
const [formData, setFormData] = useState({
  someStrings: undefined,
  someNumber: undefined,
});

<Form
  schema={schema}
  formData={formData}
  onChange={(e) => {
    console.log(e.formData);
    /*
     * // If the input value is empty
     * {
     *   someString: undefined,
     *   someNumber: undefined
     * }
     */
  }}
/>;
```

**>=v6.5.3:**

```jsx
const [formData, setFormData] = useState({
  someStrings: undefined,
  someNumber: undefined,
});

<Form
  schema={schema}
  formData={formData}
  onChange={(e) => {
    console.log(e.formData);
    /*
     * // If the input value is empty
     * {
     * }
     */
  }}
/>;
```

### Expected Behavior

As with v6.5.2, I think the following behavior is natural.

- `default` takes effect only when the form is first displayed and allows empty input
- The formData of `onChange` stores `undefined` with a key so that if the field is empty, it will be known that the field is unfilled.

### Example1

1. See: [Example1](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJzb21lU3RyaW5ncyI6IkNodWNrIiwic29tZU51bWJlcnMiOjF9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInJlcXVpcmVkIjpbInNvbWVTdHJpbmdzIl0sInByb3BlcnRpZXMiOnsic29tZVN0cmluZ3MiOnsidHlwZSI6InN0cmluZyIsImRlZmF1bHQiOiJDaHVjayJ9LCJzb21lTnVtYmVycyI6eyJ0eXBlIjoibnVtYmVyIiwiZGVmYXVsdCI6MX19fSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50In0sInVzZUZhbGxiYWNrRmllbGQiOmZhbHNlLCJ1c2VGYWxsYmFja1VpRm9yVW5zdXBwb3J0ZWRUeXBlIjpmYWxzZSwibGl2ZVZhbGlkYXRlIjpmYWxzZSwiaWRQcmVmaXgiOiJyb290IiwiaWRTZXBhcmF0b3IiOiJfIn0sInZhbGlkYXRvciI6IkFKVjgiLCJzYW1wbGVOYW1lIjoiQmxhbmsifQ==)
2. Empty the value of each input item using the BackSpace key etc.
3. You can confirm that the input field cannot be empty

### Example2

1. [Example2](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJzb21lU3RyaW5ncyI6IkNodW5rIiwic29tZU51bWJlcnMiOjF9LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInJlcXVpcmVkIjpbInNvbWVTdHJpbmdzIl0sInByb3BlcnRpZXMiOnsic29tZVN0cmluZ3MiOnsidHlwZSI6InN0cmluZyIsImRlZmF1bHQiOiJDaHVjayJ9LCJzb21lTnVtYmVycyI6eyJ0eXBlIjoibnVtYmVyIiwiZGVmYXVsdCI6MX19fSwidWlTY2hlbWEiOnsic29tZVN0cmluZ3MiOnsidWk6ZW1wdHlWYWx1ZSI6IiJ9LCJzb21lTnVtYmVycyI6eyJ1aTplbXB0eVZhbHVlIjpudWxsfX0sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJjb25zdEFzRGVmYXVsdHMiOiJhbHdheXMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOiJvbkNoYW5nZSIsImlkUHJlZml4Ijoicm9vdCIsImlkU2VwYXJhdG9yIjoiXyJ9LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IkJsYW5rIn0=)
2. Empty all input fields
3. You can see that the `required` field allows empty and the optional field results in a type error.

### Environment

```markdown
- OS: Windows, macOS
- Node: v24.15.0
- npm: v11.12.1
```

### Anything else?

This symptom appears to be an unintended side effect of the fix for [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) introduced in v6.5.3.

[Releases/v6.5.3](https://github.com/rjsf-team/react-jsonschema-form/releases/tag/v6.5.3)

> Fixed processPendingChange leaving optional object keys as undefined after clearing text inputs (invalid for AJV type: "string"), by unsetting keys for resolved non-oneOf/anyOf leaves while preserving explicit undefined for oneOf/anyOf branches, fixing #4518

[Form.tsx](https://github.com/rjsf-team/react-jsonschema-form/blob/cb432190cce9285194f37ec3e3b0f0fbff3c3d88/packages/core/src/components/Form.tsx#L973-L994)

The fix for #4518 was valid — leaving `undefined` values in `formData` for `type: "string"` fields caused AJV type errors. However, this fix introduced the following tradeoffs:

1. **Fields with `default` can no longer be cleared.** When the user empties an input, the `default` value is reapplied because the key is removed from `formData` and the schema default takes effect again.
2. **`onChange` no longer reports empty fields.** Previously, `formData` included `{ someString: undefined }` for empty fields, allowing callers to distinguish "field is present but empty" from "field was never set". Now the key is omitted entirely (`{}`), which breaks state management code that relies on key presence.

Would it be possible to address both constraints — for example, by using `undefined` with keys preserved in `onChange` while only stripping them at validation time, or by providing a way to opt in to the old behavior?

Thank you for taking a look at this.

yap-house:

This fixes a regression introduced in #5136.
After clearing field A, clearing field B caused field A to be re-populated with its schema default.
The fix uses `_set(formData, path, undefined)` instead of `_unset` to keep the cleared key present in state, preventing `mergeDefaultsWithFormData` from re-applying the default. A new private helper `omitUndefinedLeaves()` strips `undefined`-valued keys before AJV validation to preserve the #4518 fix.

fixes #5125

### Checklist

  - [X] I've added and/or updated tests. I've run `pnpm exec nx run-many --target=build --exclude=@rjsf/docs && pnpm run test:update` to update snapshots, if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed `MultiSchemaField` to propagate the parent schema's `type` to option sub-schemas that don't define their own, so the correct widget (e.g. `StringField`) renders for the selected option instead of `FallbackField`, fixing [#5119](https://github.com/rjsf-team/react-jsonschema-form/issues/5119)
 - Fixed nested `constAsDefaults` values being skipped in matching conditional `allOf` schemas, fixing [#4963](https://github.com/rjsf-team/react-jsonschema-form/issues/4963)
 - Fixed `processPendingChange` so that clearing a string field that has a schema `default` value no longer re-applies the default, fixing [#5125](https://github.com/rjsf-team/react-jsonschema-form/issues/5125)
+- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; added `omitUndefinedLeaves()` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved
 
 ## @rjsf/daisyui
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed `MultiSchemaField` to propagate the parent schema's `type` to option sub-schemas that don't define their own, so the correct widget (e.g. `StringField`) renders for the selected option instead of `FallbackField`, fixing [#5119](https://github.com/rjsf-team/react-jsonschema-form/issues/5119)
 - Fixed nested `constAsDefaults` values being skipped in matching conditional `allOf` schemas, fixing [#4963](https://github.com/rjsf-team/react-jsonschema-form/issues/4963)
 - Fixed `processPendingChange` so that clearing a string field that has a schema `default` value no longer re-applies the default, fixing [#5125](https://github.com/rjsf-team/react-jsonschema-form/issues/5125)
-- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; uses `JSON.parse(JSON.stringify(...))` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved ([#5138](https://github.com/rjsf-team/react-jsonschema-form/pull/5138))
+- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; uses `JSON.parse(JSON.stringify(...))` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved
 
 ## @rjsf/daisyui
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed `MultiSchemaField` to propagate the parent schema's `type` to option sub-schemas that don't define their own, so the correct widget (e.g. `StringField`) renders for the selected option instead of `FallbackField`, fixing [#5119](https://github.com/rjsf-team/react-jsonschema-form/issues/5119)
 - Fixed nested `constAsDefaults` values being skipped in matching conditional `allOf` schemas, fixing [#4963](https://github.com/rjsf-team/react-jsonschema-form/issues/4963)
 - Fixed `processPendingChange` so that clearing a string field that has a schema `default` value no longer re-applies the default, fixing [#5125](https://github.com/rjsf-team/react-jsonschema-form/issues/5125)
-- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; added `omitUndefinedLeaves()` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved ([#5138](https://github.com/rjsf-team/react-jsonschema-form/pull/5138))
+- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; uses `JSON.parse(JSON.stringify(...))` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved ([#5138](https://github.com/rjsf-team/react-jsonschema-form/pull/5138))
 
 ## @rjsf/daisyui
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed `MultiSchemaField` to propagate the parent schema's `type` to option sub-schemas that don't define their own, so the correct widget (e.g. `StringField`) renders for the selected option instead of `FallbackField`, fixing [#5119](https://github.com/rjsf-team/react-jsonschema-form/issues/5119)
 - Fixed nested `constAsDefaults` values being skipped in matching conditional `allOf` schemas, fixing [#4963](https://github.com/rjsf-team/react-jsonschema-form/issues/4963)
 - Fixed `processPendingChange` so that clearing a string field that has a schema `default` value no longer re-applies the default, fixing [#5125](https://github.com/rjsf-team/react-jsonschema-form/issues/5125)
-- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; added `omitUndefinedLeaves()` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved
+- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; added `omitUndefinedLeaves()` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved ([#5138](https://github.com/rjsf-team/react-jsonschema-form/pull/5138))
 
 ## @rjsf/daisyui
 
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -765,9 +765,14 @@ export default class Form<
     // When a pre-resolved schema is provided (e.g., from live validation), use it directly.
     // Otherwise validate against the original schema so AJV sees the full constraint set.
     const validationSchema = retrievedSchema ?? schema;
+
+    // JSON.stringify drops keys with `undefined` values; JSON.parse on the result gives AJV a clean
+    // object that avoids spurious type errors for `type: "string"` fields that were cleared (#4518).
+    const validationFormData = formData ? JSON.parse(JSON.stringify(formData)) : undefined;
+
     return schemaUtils
       .getValidator()
-      .validateFormData(formData, validationSchema, customValidate, transformErrors, uiSchema);
+      .validateFormData(validationFormData, validationSchema, customValidate, transformErrors, uiSchema);
   }
 
   /** Renders any errors contained in the `state` in using the `ErrorList`, if not disabled by `showErrorList`. */
@@ -1065,18 +1070,13 @@ export default class Form<
       // If we have custom errors and the path has an error, then we need to clear it
       customErrors.clearErrors(path);
     }
-
-    // JSON.stringify drops keys with `undefined` values; JSON.parse on the result gives AJV a clean
-    // object that avoids spurious type errors for `type: "string"` fields that were cleared (#4518).
-    const formDataForValidation = newFormData ? JSON.parse(JSON.stringify(newFormData)) : undefined;
-
     // If there are pending changes in the queue, skip live validation since it will happen with the last change
     if (mustValidate && this.pendingChanges.length === 1) {
       const liveValidation = this.liveValidate(
         schema,
         schemaUtils,
         mergeBaseErrorSchema,
-        formDataForValidation,
+        newFormData,
         extraErrors,
         customErrors,
         retrievedSchema,
@@ -1373,12 +1373,7 @@ export default class Form<
   validateFormWithFormData = (formData?: T): boolean => {
     const { extraErrors, extraErrorsBlockSubmit, focusOnFirstError, onError } = this.props;
     const { errors: prevErrors } = this.state;
-
-    // JSON.stringify drops keys with `undefined` values; JSON.parse on the result gives AJV a clean
-    // object that avoids spurious type errors for `type: "string"` fields that were cleared (#4518).
-    const formDataForValidation = formData ? JSON.parse(JSON.stringify(formData)) : undefined;
-
-    const schemaValidation = this.validate(formDataForValidation);
+    const schemaValidation = this.validate(formData);
     // Always merge extraErrors so they remain visible in state regardless of extraErrorsBlockSubmit.
     const { errors, errorSchema } = extraErrors ? Form.mergeErrors<T>(schemaValidation, extraErrors) : schemaValidation;
     // hasError gates submission: schema errors always block; extraErrors only block when
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -817,6 +817,38 @@ export default class Form<
     return { errors, errorSchema };
   }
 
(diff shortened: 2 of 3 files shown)


packages/core/src/components/Form.tsx
```diff
@@ -817,6 +817,38 @@ export default class Form<
...
+   * `type: "string"` under `patternProperties`, which would cause a spurious type
+   * error (#4518), while still keeping the `undefined` key in state so that
+   * `mergeDefaultsWithFormData` does not re-apply schema defaults when the user has
+   * explicitly cleared a field (#5125 regression).
+   *
+   * @param formData - The form data to strip
+   * @returns - A copy of `formData` with all `undefined`-valued keys omitted
+   * @private
+   */
+  private static omitUndefinedLeaves<T>(formData: T): T {
```
heath-freenome: I'm thinking that maybe this belongs in omitExtraData?

packages/core/src/components/Form.tsx
```diff
@@ -817,6 +817,38 @@ export default class Form<
...
+    // If the formData is not an object, return it as is (base case for recursion)
+    if (!isObject(formData)) {
+      return formData;
+    }
+
+    // If the formData is an object, create a new object to hold the result
+    const result: any = {};
+
+    for (const key of Object.keys(formData as object)) {
```
heath-freenome: `isObject()` returns true for arrays. Did you want different logic for an array versus an object?

packages/core/src/components/Form.tsx
heath-freenome: Another approach might be to JSON.stringify()/JSON.parse() the object since stringifying an undefined value skips the key. Not sure which is faster, your algorithm or the serialize/deserialize

yap-house: @heath-freenome As discussed in [#5125](https://github.com/rjsf-team/react-jsonschema-form/issues/5125#issuecomment-4832078625), I tried replacing `omitUndefinedLeaves` with `schemaUtils.omitExtraData` but it caused two test failures — `omitExtraData` also strips keys not in the schema, which breaks `additionalProperties: false` and `if/then` validation. The `JSON.parse(JSON.stringify(...))` approach you suggested below ended up resolving this more cleanly.

packages/core/src/components/Form.tsx
yap-house: @heath-freenome Good catch. This is now moot since the implementation was switched to `JSON.parse(JSON.stringify(...))` per your suggestion below — it handles both objects and arrays without needing to distinguish between them.

packages/core/src/components/Form.tsx
yap-house: @heath-freenome Adopted this approach — updated the PR. It's simpler and also sidesteps the array/object distinction question above. Thank you for the suggestion!

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed `MultiSchemaField` to propagate the parent schema's `type` to option sub-schemas that don't define their own, so the correct widget (e.g. `StringField`) renders for the selected option instead of `FallbackField`, fixing [#5119](https://github.com/rjsf-team/react-jsonschema-form/issues/5119)
 - Fixed nested `constAsDefaults` values being skipped in matching conditional `allOf` schemas, fixing [#4963](https://github.com/rjsf-team/react-jsonschema-form/issues/4963)
 - Fixed `processPendingChange` so that clearing a string field that has a schema `default` value no longer re-applies the default, fixing [#5125](https://github.com/rjsf-team/react-jsonschema-form/issues/5125)
+- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; uses `JSON.parse(JSON.stringify(...))` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved ([#5138](https://github.com/rjsf-team/react-jsonschema-form/pull/5138))
```
```suggestion
- Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; uses `JSON.parse(JSON.stringify(...))` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved
```

heath-freenome (approved):

packages/core/src/components/Form.tsx
```diff
@@ -1062,13 +1065,18 @@ export default class Form<
       // If we have custom errors and the path has an error, then we need to clear it
       customErrors.clearErrors(path);
     }
+
+    // JSON.stringify drops keys with `undefined` values; JSON.parse on the result gives AJV a clean
+    // object that avoids spurious type errors for `type: "string"` fields that were cleared (#4518).
+    const formDataForValidation = newFormData ? JSON.parse(JSON.stringify(newFormData)) : undefined;
+
```
heath-freenome: Move this inside of the block before the call to liveValidate so we don't waste cycles for something we won't use all the time.

packages/core/src/components/Form.tsx
```diff
@@ -1365,7 +1373,12 @@ export default class Form<
   validateFormWithFormData = (formData?: T): boolean => {
     const { extraErrors, extraErrorsBlockSubmit, focusOnFirstError, onError } = this.props;
     const { errors: prevErrors } = this.state;
-    const schemaValidation = this.validate(formData);
+
+    // JSON.stringify drops keys with `undefined` values; JSON.parse on the result gives AJV a clean
```
heath-freenome: Actually, since both `liveValidate()` and `validateWithFormData()` both call `validate()` i believe it actually make sense to do this tenary on line 768 before we call the `validateFormData()` from `schemaUtils`

heath-freenome (changes requested):

packages/core/src/components/Form.tsx
heath-freenome: NVM, see the comment below

yap-house: @heath-freenome Thank you for your review! The answers are summarized in [the comments below](https://github.com/rjsf-team/react-jsonschema-form/pull/5138#discussion_r3549284931).

packages/core/src/components/Form.tsx
yap-house: @heath-freenome Thank you for your suggestion! I moved JSON.parse/stringify sanitization into `validate()`. Please check the commit d18749a5d.

heath-freenome (approved):

PR #5139, opened 2026-06-28
refactor: convert AnyOfField from class to functional component

heath-freenome:

Replaces the class-based AnyOfField with a functional component using useReducer for state management. State snapshots (prevOptions, prevFormData, prevFieldId) replace the three useRef prev-trackers and useEffect, using React's getDerivedStateFromProps render-phase dispatch pattern instead — React re-renders immediately from those dispatches and discards the current render before any DOM commit. onOptionChange is wrapped in useCallback.

### Checklist

### DIFF
--- a/packages/core/src/components/fields/MultiSchemaField.tsx
+++ b/packages/core/src/components/fields/MultiSchemaField.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useReducer, useRef } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import type { FieldProps, FormContextType, RJSFSchema, StrictRJSFSchema, UiSchema } from '@rjsf/utils';
 import {
   ANY_OF_KEY,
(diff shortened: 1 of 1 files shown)


packages/core/src/components/fields/MultiSchemaField.tsx
```diff
@@ -19,279 +19,280 @@ import isEmpty from 'lodash/isEmpty';
...
+  if (fieldPathId.$id !== state.prevFieldId) {
+    nextState = { ...nextState, prevFieldId: fieldPathId.$id };
+  }
 
-    const discriminator = getDiscriminatorFieldFromSchema<S>(schema);
-    const option = schemaUtils.getClosestMatchingOption(formData, options, selectedOption, discriminator);
-    return option;
+  if (nextState !== state) {
+    dispatch({ type: 'UPDATE_STATE', payload: nextState });
   }
```
nickgros: I'm not a fan of this pattern where we conditionally set state in the render, IMO it becomes too easy to accidentally create an infinite re-render loop.  But I understand that it's probably the only way to keep this logic in the component and not trigger extra render commits.

Do you share the same concerns? Should we live with this or try to find ways to avoid this in the future?

nickgros (approved):

packages/core/src/components/fields/MultiSchemaField.tsx
heath-freenome: Per Claude:
```
The reviewer's concern is valid, and there is a cleaner structure that eliminates the render-phase dispatch entirely.

The root cause is that we mixed two conceptually different things in state:
- retrievedOptions — purely derived from options; belongs in useMemo, not state
- selectedOption — genuinely stateful (user interaction + auto-matching)

Because we put retrievedOptions in state, we needed render-phase dispatch to keep it in sync. If we separate them correctly:

┌───────────────────────────────────┬───────────────────────────────────────────┐
│              Current              │                  Better                   │
├───────────────────────────────────┼───────────────────────────────────────────┤
│ retrievedOptions in reducer state │ useMemo([options, schemaUtils, formData]) │
├───────────────────────────────────┼───────────────────────────────────────────┤
│ prevOptions snapshot in state     │ gone — useMemo handles invalidation       │
├───────────────────────────────────┼───────────────────────────────────────────┤
│ prevFormData snapshot in state    │ useRef (only needed in the effect)        │
├───────────────────────────────────┼───────────────────────────────────────────┤
│ prevFieldId snapshot in state     │ useRef (only needed in the effect)        │
├───────────────────────────────────┼───────────────────────────────────────────┤
│ render-phase dispatch             │ gone                                      │
├───────────────────────────────────┼───────────────────────────────────────────┤
│ useReducer with 5-field state     │ useState<number> for selectedOption only  │
├───────────────────────────────────┼───────────────────────────────────────────┤
│ UPDATE_STATE action               │ gone                                      │
└───────────────────────────────────┴───────────────────────────────────────────┘

The auto-matching logic (selectedOption re-match on formData change) goes back into a useEffect without a deps array — same as componentDidUpdate. That's the conventional approach, it's easier to reason about, and since the effect only dispatches when matchingOption !== selectedOption, there's no infinite loop risk.

The trade-off vs the render-phase approach: useEffect fires post-commit rather than pre-commit. But the original class component had the same timing (via componentDidUpdate), so this isn't a regression — it's just returning to what the class did.
```

packages/core/src/components/fields/MultiSchemaField.tsx
```diff
@@ -18,280 +18,234 @@ import get from 'lodash/get';
...
+    uiSchema,
+  } = props;
+  const { schemaUtils } = registry;
+
+  // retrievedOptions is purely derived from options — useMemo handles re-derivation automatically
+  // when options, schemaUtils, or formData change, with no render-phase dispatch needed.
+  const retrievedOptions = useMemo(
+    () => options.map((opt: S) => schemaUtils.retrieveSchema(opt, formData)),
+    [options, schemaUtils, formData],
+  );
```
heath-freenome, line 58: @nickgros This differs from the old implementation which did it in the constructor once... Not sure whether this is better or worse

nickgros: I think the `useEffect` change is slightly more idiomatic. There is a sync happening, so I think `useEffect` is the right tool to communicate the intent and match the old behavior.

Ideally we wouldn't need `useEffect` or an update mid-render at all, but I think that would require decoupling this UI logic from the React lifecycle altogether, which could be a massive change.

PR #5156, opened 2026-07-14
Fix 4631: preserve additional property order for numeric keys

### LINKED ISSUE #4631: `patternProperties` number keys make the input get rendered at the very top of the `Form`
### Prerequisites

### What theme are you using?

core

### Version

6.0.0-beta.8

### Current Behavior

Adding a new `patternProperties` key that is a number makes the key name input get rendered at the very top of the `Form`:

Image

Notice that keys that have non-numeric characters in them are not affected by this issue.

### Expected Behavior

Keys should at least remain grouped.
It would be intuitive to always have them rendered in the order they were added.

If they should be reordered dynamically amongst themselves, I don't know.
Maybe this could be a customizable behaviour set through a new prop or `UiSchema` property? Just like lists allow reordering, maybe?

### Steps To Reproduce

1. Go to the playground `patternProperties` example.
2. Change the regex of the JSON Schema `patternproperties` property to `"(.*?)"`, so it matches anything.
3. Add a new property with a purely numeric key.
4. Observe where the field gets rendered.

### Environment

```markdown
Irrelevant
```

### Anything else?

_No response_

wanxiankai:

Fixes #4631.

JavaScript enumerates integer-like object keys before string keys. When an `additionalProperties` or `patternProperties` key was renamed to an integer, `Object.keys(schema.properties)` moved the field to the top of the form even though the user added it later.

### Changes made

- Track the render order of additional and pattern properties inside `ObjectField`.
- Keep that order in sync when dynamic properties are added, renamed, or removed.
- Keep schema-defined fields ahead of dynamic fields while preserving explicit `ui:order` as the final override.
- Add a regression test that renames the second pattern property to `1` and verifies that it stays second.
- Add the fix to the `@rjsf/core` changelog.

### Testing

- `corepack pnpm -C packages/core exec vitest run test/ObjectField.test.tsx` (249 passed)
- `corepack pnpm -C packages/core run test` (2215 passed)
- `corepack pnpm run lint`
- `corepack pnpm run build`
- `corepack pnpm run test`
- Target-file Oxfmt and `git diff --check`

### Checklist

- [ ] **I am updating documentation**
  - [ ] I have checked the rendering of Markdown text I added
- [x] **I am adding or updating code**
  - [x] I have added and updated tests
  - [x] Documentation changes are not needed for this bug fix
  - [x] I have updated `CHANGELOG.md`
- [ ] **I am adding a new feature**
  - [ ] I have updated the playground with an example of the feature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.7.1
+
+## @rjsf/core
+
+- Fixed `ObjectField` to preserve the insertion order of additional and pattern properties when keys are renamed to integers, fixing [#4631](https://github.com/rjsf-team/react-jsonschema-form/issues/4631) ([#5156](https://github.com/rjsf-team/react-jsonschema-form/pull/5156))
+
 # 6.7.0
 
 ## @rjsf/antd
@@ -43,7 +49,6 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; uses `JSON.parse(JSON.stringify(...))` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved
 - Show selected option descriptions, fixing ([#4214](https://github.com/rjsf-team/react-jsonschema-form/issues/4214))
 - Fixed `ObjectField`'s "Add" button for `additionalProperties` schemas containing a `$ref` so the new item's default value is computed via `getDefaultFormState()` against the fully-resolved referenced schema (recursively populating nested defaults) instead of only reading the resolved schema's own top-level `default`, which discarded the sibling `default` and any nested property defaults, fixing [#4266](https://github.com/rjsf-team/react-jsonschema-form/issues/4266)
-- Fixed `ObjectField` to preserve the insertion order of additional and pattern properties when keys are renamed to integers, fixing [#4631](https://github.com/rjsf-team/react-jsonschema-form/issues/4631) ([#5156](https://github.com/rjsf-team/react-jsonschema-form/pull/5156))
 
 ## @rjsf/daisyui
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Fixed a regression introduced in [#5136](https://github.com/rjsf-team/react-jsonschema-form/pull/5136) where clearing a second field caused a previously-cleared field to be re-populated with its schema default; uses `JSON.parse(JSON.stringify(...))` to strip `undefined`-valued keys from formData before AJV validation so that the [#4518](https://github.com/rjsf-team/react-jsonschema-form/issues/4518) fix is preserved
 - Show selected option descriptions, fixing ([#4214](https://github.com/rjsf-team/react-jsonschema-form/issues/4214))
 - Fixed `ObjectField`'s "Add" button for `additionalProperties` schemas containing a `$ref` so the new item's default value is computed via `getDefaultFormState()` against the fully-resolved referenced schema (recursively populating nested defaults) instead of only reading the resolved schema's own top-level `default`, which discarded the sibling `default` and any nested property defaults, fixing [#4266](https://github.com/rjsf-team/react-jsonschema-form/issues/4266)
-- Fixed `ObjectField` to preserve the insertion order of additional and pattern properties when keys are renamed to integers, fixing [#4631](https://github.com/rjsf-team/react-jsonschema-form/issues/4631)
+- Fixed `ObjectField` to preserve the insertion order of additional and pattern properties when keys are renamed to integers, fixing [#4631](https://github.com/rjsf-team/react-jsonschema-form/issues/4631) ([#5156](https://github.com/rjsf-team/react-jsonschema-form/pull/5156))
 
 ## @rjsf/daisyui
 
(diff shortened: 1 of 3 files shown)


packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -370,7 +376,30 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
...
+
+      const currentAdditionalPropertySet = new Set(currentAdditionalProperties);
+      const retainedAdditionalProperties = additionalPropertyOrder.current.filter((property) =>
+        currentAdditionalPropertySet.has(property),
+      );
+      const retainedAdditionalPropertySet = new Set(retainedAdditionalProperties);
+      const newAdditionalProperties = currentAdditionalProperties.filter(
+        (property) => !retainedAdditionalPropertySet.has(property),
+      );
+      additionalPropertyOrder.current = [...retainedAdditionalProperties, ...newAdditionalProperties];
```
heath-freenome: While I understand the logic, I feel like it is doing too much work every single render. Does it make sense to instead memoize this on the initial render and then manually keep it in sync on the additional property create/delete callbacks?

wanxiankai: Thanks, that makes sense. I pushed 3968c4a to initialize the additional property order once from the retrieved schema and then keep it in sync through the add/rename/remove callbacks. The render path now only combines the memoized schema-defined properties with the tracked additional property order.

I also reran `corepack pnpm -C packages/core exec vitest run test/ObjectField.test.tsx` (249 passed) and `corepack pnpm -C packages/core run build`.

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -238,6 +248,16 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
...
   const childFieldPathId = props.childFieldPathId ?? fieldPathId;
   const lastRenamedProperty = useRef({ previousKey: '', currentKey: undefined as string | undefined });
+  const additionalPropertyOrder = useRef<string[]>([]);
+  const additionalPropertyOrderInitialized = useRef(false);
+  if (!additionalPropertyOrderInitialized.current) {
+    additionalPropertyOrder.current = getAdditionalPropertyOrder<S>(schemaProperties);
+    additionalPropertyOrderInitialized.current = true;
+  }
+  const definedPropertyOrder = useMemo(
+    () => Object.keys(schemaProperties).filter((property) => !isAdditionalPropertySchema(schemaProperties[property])),
```
heath-freenome: Why not use the new `getAdditionalPropertyOrder()` function here instead of replicating the logic?
Also, you could move the logic from lines 253-256 inside of this function as well, reusing the same computation for both. Then you wouldn't need `additionalPropertyOrderInitialized` at all ... i.e.:

```suggestion
    () => {
      const order = getAdditionalPropertyOrder<S>(schemaProperties);
      if (!additionalPropertyOrderInitialized.current) {
        additionalPropertyOrder.current = order;
      }
      return order;
    }
```

packages/core/src/components/fields/ObjectField.tsx
heath-freenome: You probably want to memoize `schemaProperties` from schema since schema is also memoized.

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -369,8 +394,10 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
 
   if (!renderOptionalField || hasFormData) {
     try {
-      const properties = Object.keys(schemaProperties);
-      orderedProperties = orderProperties(properties, uiOptions.order);
+      const currentAdditionalProperties = additionalPropertyOrder.current.filter((property) =>
+        Object.hasOwn(schemaProperties, property),
+      );
+      orderedProperties = orderProperties([...definedPropertyOrder, ...currentAdditionalProperties], uiOptions.order);
```
heath-freenome, line 396: Won't this generate duplicates?

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ should change the heading of the (upcoming) version to include a major version b
...
 - Show selected option descriptions, fixing ([#4214](https://github.com/rjsf-team/react-jsonschema-form/issues/4214))
 - Fixed `ObjectField`'s "Add" button for `additionalProperties` schemas containing a `$ref` so the new item's default value is computed via `getDefaultFormState()` against the fully-resolved referenced schema (recursively populating nested defaults) instead of only reading the resolved schema's own top-level `default`, which discarded the sibling `default` and any nested property defaults, fixing [#4266](https://github.com/rjsf-team/react-jsonschema-form/issues/4266)
+- Fixed `ObjectField` to preserve the insertion order of additional and pattern properties when keys are renamed to integers, fixing [#4631](https://github.com/rjsf-team/react-jsonschema-form/issues/4631) ([#5156](https://github.com/rjsf-team/react-jsonschema-form/pull/5156))
```
heath-freenome: Also, please move this to the 6.7.1 section since we released 6.7.0

packages/core/src/components/fields/ObjectField.tsx
wanxiankai: Thanks, updated in cdeae96. `schemaProperties` is memoized now, the additional property order is initialized once as state and kept in sync through the add/rename/remove callbacks, and `definedPropertyOrder` now reuses `getAdditionalPropertyOrder()` instead of repeating the additional-property check.

packages/core/src/components/fields/ObjectField.tsx
wanxiankai: Done in cdeae96. I changed `schemaProperties` to a `useMemo()` derived from the memoized schema.

packages/core/src/components/fields/ObjectField.tsx
wanxiankai: Good catch. I added a `definedPropertySet` filter in cdeae96 before combining the tracked additional properties with `definedPropertyOrder`, so a key cannot be emitted from both lists.

CHANGELOG.md
wanxiankai: Done in cdeae96. I moved the entry into a new `6.7.1` / `@rjsf/core` section.

packages/core/src/components/fields/ObjectField.tsx
heath-freenome: Perfect! Can you make sure there are tests that exercise this logic?

wanxiankai: Added in 6eee33a. The new test rerenders an `additionalProperties` object after one tracked additional key becomes schema-defined and verifies that the promoted field is rendered only once while the remaining additional key stays in the dynamic key list.

I reran `corepack pnpm -C packages/core exec vitest run test/ObjectField.test.tsx` (250 passed), `corepack pnpm exec oxfmt --check packages/core/test/ObjectField.test.tsx`, and `corepack pnpm -C packages/core run build`.

heath-freenome (approved):

PR #5162, opened 2026-07-17
chore: bump ata-validator to ^1.1.0 in @rjsf/validator-ata

mertcanaltin:
#### Reasons for making this change

ata-validator 1.1.0 fixes a crash that matters for how rjsf uses the library: in environments without the native addon, which includes every browser, schemas the JS engine could not compile failed with a stack overflow (`Maximum call stack size exceeded`). Since form schemas are user-supplied, raising the minimum keeps that failure mode out of `@rjsf/validator-ata` installs; affected schemas now report a clear error instead.

1.1.0 also moves the native engine into per-platform optional packages, so the main tarball drops from about 5 MB to 120 KB. No API changes; the full `@rjsf/validator-ata` test suite passes unchanged against 1.1.0.

Changelog: https://github.com/ata-core/ata-validator/blob/master/CHANGELOG.md

#### Checklist

- [x] Added a CHANGELOG entry under 6.7.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.7.1
+
+## @rjsf/validator-ata
+
+- Updated the `ata-validator` dependency to `^1.1.0`, which fixes a stack overflow in browser environments for schemas its JS engine cannot compile and shrinks the install footprint via per-platform optional packages ([#5162](https://github.com/rjsf-team/react-jsonschema-form/pull/5162))
+
 # 6.7.0
 
 ## @rjsf/antd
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -111,6 +111,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Updated `transformRJSFValidationErrors()` to detect field name for `allOf` and `if-then-else` by falling back to the root schema to find the title, fixing [#5134](https://github.com/rjsf-team/react-jsonschema-form/issues/5134)
 - Updated the `ata-validator` dependency to `^1.0.0`, its first stable release; no API changes for `@rjsf/validator-ata` consumers ([#5157](https://github.com/rjsf-team/react-jsonschema-form/pull/5157))
+- Updated the `ata-validator` dependency to `^1.1.0`, which fixes a stack overflow in browser environments for schemas its JS engine cannot compile and shrinks the install footprint via per-platform optional packages ([#5162](https://github.com/rjsf-team/react-jsonschema-form/pull/5162))
```
heath-freenome: We released 6.7.0. Please add this to the 6.7.1 section instead

heath-freenome (approved):

CHANGELOG.md
heath-freenome: NVM, I just did it.

mertcanaltin: thanks for the edit 🙏

PR #5170, opened 2026-07-25
Fix 4136: render required asterisk for standalone boolean checkbox fields

### LINKED ISSUE #4136: Schema defines standalone checkbox as mandatory, but asterisk is not rendered.
### Prerequisites

### What theme are you using?

core

### Version

5.x

### Current Behavior

Making a standalone checkbox mandatory does NOT render an asterisk.
Refer to schema below in 'Steps to reproduce'

Form renders ok, errors shown up when clicking submit button, but no asterisk shows up.

### Expected Behavior

Making a standalone checkbox mandatory should render an asterisk, so behavior is consistent across this package.

### Steps To Reproduce

This is the schema I am using

```
const schema: RJSFSchema = {
    title: "User Information", //fieldset-Legend
    type: "object",
    required: ["firstName", "IAgreeCheckbox"], // These fields are mandatory;
    properties: {
      firstName: {
        title: "First Name",
        type: "string",
        minLength: 2,
        maxLength:20,
        pattern: "^[A-Za-z'-]*$"
      },
      age: {
        title: "Age",
        type: "number",
        minimum:18,
        maximum:100
      },
      //single-checkbox:
      IAgreeCheckbox: {
        type: "boolean",
        title: "I Agree with All the above",
        const: true //Must check
      }
    }
  };
```

  /*
  Field-level error message:
  Must be equal to constant.

  Form-level error message:
  'I Agree with All the above' must be equal to constant
  */

### Environment

```markdown
- OS:Windows 11
- Node: v14.20.0 (not relevant)
- npm: 6.14.17 (not relevant)
```

### Anything else?

It's a basic problem, the library should behave consistently.

heath-freenome:
Fixes #4136 as follows:

Add false default for required boolean properties in getDefaultFormState

Also made it so that the * only shows up when the component is required AND the value must be true

When a boolean field is listed in the parent schema's `required` array, the FieldTemplate skips its own Label (which renders the asterisk) because checkboxes render their label inline. Each CheckboxWidget was using `schemaRequiresTrueValue` (which only returns true when `const: true` or `enum: [true]`) for its display label, so fields that were merely in the `required` array never showed the `*` marker.

This separates the two concerns across all theme packages:
- HTML `required` attribute on the input continues to use `schemaRequiresTrueValue` (an unchecked checkbox submits as false, so HTML validation should only trigger when the value must be true).
- The visual required indicator now uses the `required` prop from WidgetProps (i.e. whether the field appears in the parent's `required` array).

Fixes applied to: core, react-bootstrap, antd, mui, chakra-ui, fluentui-rc, semantic-ui, shadcn, primereact, and mantine.

Adds a unit test in BooleanField.test.tsx and a shared snapshot test in formTests.tsx (covering all themes) to prevent regressions.

Eight RJSF-specific option keys (allowClearTextInputs, autofocus, enumOrder, optionsSchemaSelector, enableOptionalDataFieldForType, optionValueFormat, deprecatedHandling, style) and all 14 template component override keys were absent from uiOptionsKeys, meaning they could pass through cleanupOptions and be spread onto Mantine components, eventually reaching the DOM and triggering React unknown-prop warnings. Added the complete set of UIOptionsType keys.

### Checklist


packages/chakra-ui/src/CheckboxWidget/CheckboxWidget.tsx
```diff
@@ -35,11 +28,8 @@ export default function CheckboxWidget<
     options,
     uiSchema,
     schema,
+    required,
   } = props;
-  // Because an unchecked checkbox will cause html5 validation to fail, only add
-  // the "required" attribute if the field value must be "true", due to the
-  // "const" or "enum" keywords
-  const required = schemaRequiresTrueValue<S>(schema);
```
nickgros, line 42: @heath-freenome I'm not sure if this is the right change to make. From a UX perspective (and likely the rationale for the current logic), when a user sees a checkbox with a `required` indicator, they assume that the value must be set to `true` to proceed.

So I think we should still use `schemaRequiresTrueValue`

packages/antd/src/widgets/CheckboxWidget/index.tsx
```diff
@@ -42,7 +56,16 @@ export default function CheckboxWidget<
       {...extraProps}
       aria-describedby={ariaDescribedByIds(id)}
     >
-      {labelValue(label, hideLabel, '')}
+      {!hideLabel && label ? (
+        <>
+          {label}
+          {required && (
```
nickgros: this should use `schemaRequiresTrueValue`

packages/fluentui-rc/src/CheckboxWidget/CheckboxWidget.tsx
```diff
@@ -29,17 +29,13 @@ export default function CheckboxWidget<
...
   } = props;
   const DescriptionFieldTemplate = getTemplate<'DescriptionFieldTemplate', T, S, F>(
     'DescriptionFieldTemplate',
     registry,
     options,
   );
-  // Because an unchecked checkbox will cause html5 validation to fail, only add
-  // the "required" attribute if the field value must be "true", due to the
-  // "const" or "enum" keywords
-  const required = schemaRequiresTrueValue<S>(schema);
```
nickgros, line 41: Suggest reverting

packages/core/test/BooleanField.test.tsx
```diff
@@ -44,6 +44,23 @@ describe('BooleanField', () => {
     expect(node.querySelector('.rjsf-field label span')).toHaveTextContent('foo');
   });
 
+  it('should render a required asterisk when the boolean field is required', () => {
```
nickgros: I think this is invalid (see other comments)

nickgros (changes requested):

heath-freenome:
Thanks for the review @nickgros — I want to push back a bit on the direction here, because I think the PR is actually doing the right thing by separating two concerns that were previously conflated.

**The two concerns:**

1. **HTML `required` attribute on the `<input>`** → correctly kept as `schemaRequiresTrueValue`. An unchecked checkbox submits as `false`, so browser-native validation should only trigger when the value *must be true* (via `const: true` or `enum: [true]`). The PR preserves this.

2. **Visual asterisk (`*`) in the label** → this is what the PR changes to use the `required` prop (i.e. whether the field appears in the parent schema's `required` array).

The reviewer's concern conflates these two. The asterisk in RJSF means "this field is in the parent schema's `required` array" — that's its convention across *every other field type* (strings, numbers, objects, etc.). The original bug (#4136) is that required boolean fields show no asterisk while every other type does. That inconsistency is the core problem.

**On the UX concern:**

The ambiguity you're describing — a user seeing `*` and assuming they must check the box — is real, but I think the alternative is worse. Hiding the asterisk on a required boolean means:

- A field the schema author explicitly marked as `required` gives no visual signal to the user
- Boolean fields behave inconsistently with every other RJSF field type
- The form author's intent is silently ignored

If anything, the meaning of `required` on a boolean in JSON Schema (key must be present in the object) is somewhat unusual since booleans are never truly "absent" from submitted form data. But that's a schema-authoring concern — when someone puts a boolean in `required`, they're signaling intent, and RJSF should reflect that consistently.

The HTML `required` attribute is exactly the right place to enforce the "must be true" distinction, and the PR already handles that correctly.

Happy to add clarifying language to the PR description if that helps, but I think the design is sound and would like to keep the change as-is. Let me know if there's a specific scenario where you think this breaks down — willing to work through it.

packages/core/src/components/widgets/CheckboxWidget.tsx
```diff
@@ -40,7 +41,7 @@ function CheckboxWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F exte
   // Because an unchecked checkbox will cause html5 validation to fail, only add
   // the "required" attribute if the field value must be "true", due to the
   // "const" or "enum" keywords
-  const required = schemaRequiresTrueValue<S>(schema);
+  const inputRequired = schemaRequiresTrueValue<S>(schema);
```
nickgros:
```suggestion
  const inputRequired = required && schemaRequiresTrueValue<S>(schema);
```

packages/core/src/components/widgets/CheckboxWidget.tsx
```diff
@@ -77,15 +78,21 @@ function CheckboxWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F exte
...
           onChange={handleChange}
           onBlur={handleBlur}
           onFocus={handleFocus}
           aria-describedby={ariaDescribedByIds(id)}
         />
-        {labelValue(<span>{label}</span>, hideLabel)}
+        {labelValue(
+          <span>
+            {label}
+            {required && <span className='required'>*</span>}
```
nickgros:
```suggestion
            {inputRequired && <span className='required'>*</span>}
```

packages/core/src/components/widgets/CheckboxWidget.tsx
```diff
@@ -40,7 +41,7 @@ function CheckboxWidget<T = any, S extends StrictRJSFSchema = RJSFSchema, F exte
   // Because an unchecked checkbox will cause html5 validation to fail, only add
   // the "required" attribute if the field value must be "true", due to the
   // "const" or "enum" keywords
-  const required = schemaRequiresTrueValue<S>(schema);
+  const inputRequired = schemaRequiresTrueValue<S>(schema) && required;
```
nickgros, line 44: I think this is maybe a better variable name?
```suggestion
  const trueIsRequired = schemaRequiresTrueValue<S>(schema) && required;
```

packages/core/test/BooleanField.test.tsx
```diff
@@ -245,8 +268,12 @@ describe('BooleanField', () => {
       schema: {
         type: 'object',
         properties: {
+          // Use a string enum so the field stays absent from initial formData —
+          // required booleans now default to false, which satisfies AJV's required
+          // constraint and would suppress the validation error this test relies on.
```
nickgros: Would prefer to avoid comments like these that are basically documenting a "before and after" that we won't care about after this PR is merged. In this case I think it can be wholly removed.

In my repos I have a section in the CLAUDE.md (or a STANDARDS.md) like this, wonder if we should add this or something like it? IME it ends up reducing the amount an agent writes these, and usually removes them later in a `/code-review` pass when ensuring changes meet the code standards.

```markdown
## Code Comments

- **Prioritize Expressive Code**: Write highly readable, self-documenting code as the primary means of explanation. Use comments exclusively to provide critical context that cannot be naturally expressed through clean naming conventions and clear structure.
- **Target the Audience (Method/Class-level docs vs. Inline)**: Match documentation placement to its specific consumer:
  - **Public API**: Focus class and method comments strictly on the public contract, defining the behavior, parameters, and return values expected by the caller at that specific level of abstraction.
  - **Internal Logic (Inline Comments)**: Place all underlying execution details, algorithmic mechanics, and internal complexities entirely within inline comments inside the implementation body.
- **Document Intent, Refactor Mechanics**: Dedicate internal comments to explaining the underlying business logic, constraints, and rationale behind the code (the Why). Allow the code architecture to explain the execution (the What). Treat any impulse to write step-by-step prose about what the code is doing as an immediate signal to refactor the code into clearer, smaller functions.
- **Current State Only**: Code comments and CLAUDE.md files should exclusively describe the current state, logic, and intent of the code.
  - Keep historical context, diff explanations, and "before vs. after" commentary entirely within planning documents, commit messages, PR descriptions, or narrowly scoped as comments that are co-located with specific regression tests.
  - Limit references to past logic strictly to active, ongoing code migration paths that directly impact current execution.
- **Stable References**: Code comments and CLAUDE.md files should use reference points that survive automated refactoring and ongoing codebase evolution.
  - Point to other code exclusively through LSP-supported dynamic links or external issue keys (like PLFM-1234).
  - Define target locations using conceptual names or programmable signatures instead of brittle options like absolute file paths or hard-coded line numbers.
 ```

File: packages/core/test/BooleanField.test.tsx [hunk 9]
```diff
@@ -280,8 +307,12 @@ describe('BooleanField', () => {
...
```
nickgros: Ditto to remove this comment

packages/antd/src/widgets/CheckboxWidget/index.tsx
```diff
@@ -14,10 +14,26 @@ export default function CheckboxWidget<
...
+    onFocus,
+    readonly,
+    value,
+    required,
+    schema,
+  } = props;
   const { formContext } = registry;
   const { readonlyAsDisabled = true } = formContext as GenericObjectType;
+  const { token } = theme.useToken();
+  const inputRequired = schemaRequiresTrueValue(schema) && required;
```
nickgros: better name IMO
```suggestion
  const trueValueRequired = schemaRequiresTrueValue(schema) && required;
```

packages/antd/src/widgets/CheckboxWidget/index.tsx
```diff
@@ -38,11 +54,21 @@ export default function CheckboxWidget<
       disabled={disabled || (readonlyAsDisabled && readonly)}
       id={id}
       name={htmlName || id}
+      required={required}
```
nickgros, line 57: I think for all of these we should only pass `inputRequired` (or `trueValueRequired` if you accept my rename suggestion) or else the field has different behavior depending on if HTML validation is enabled.

JSON Schema `required` -> value must not be undefined
HTML input `required` -> value must be `true`

```suggestion
      required={trueValueRequired}
```

heath-freenome: @nickgros I think that may end up being a breaking change. Plus with the defaulting of required booleans to false, the data should always be there so this will likely be a non-starter. We can consider changing this in 7.x

packages/core/test/BooleanField.test.tsx
heath-freenome: I updated our CLAUDE.md per your feedback

packages/antd/src/widgets/CheckboxWidget/index.tsx
nickgros: The actual behavior will vary depending on the default form state behavior that the caller chooses. But that it may be a breaking change is a fair point, since `core` currently passes `required={required}`.

packages/core/test/ArrayField.test.tsx
```diff
@@ -2945,6 +2945,35 @@ describe('ArrayField', () => {
       );
     });
 
+    it('does not leave a stale, duplicated error behind when removing a valid element between two invalid ones', async () => {
```
nickgros, line 2948: I don't think you meant to commit these tests?

packages/core/test/Form.test.tsx
```diff
@@ -3757,6 +3757,36 @@ describeRepeated('Form common', (createFormComponent) => {
   });
 });
 
+describe('Error paths that collide with prototype keys', () => {
```
nickgros, line 3760: I don't think you meant to commit these tests?

nickgros (approved):

PR #5171, opened 2026-07-26
Update sortedJSONStringify

samcragg:
Re-implement `sortedJSONStringify` to preserve difference between `null` and `undefined` array elements

There's an issue where array items of `null` and `undefined` are treated the same. This means that changing an item from `undefined` to `null` (or vice versa) meant the widget was not being updated (it thought no value had changed).

### Checklist

### DIFF
--- a/packages/utils/src/hashForSchema.ts
+++ b/packages/utils/src/hashForSchema.ts
@@ -26,13 +26,18 @@ export function hashString(string: string): string {
  * @returns - The stringified object with keys sorted in a consistent order
  */
 export function sortedJSONStringify(object: unknown): string {
-  const allKeys = new Set<string>();
-  // solution source: https://stackoverflow.com/questions/16167581/sort-object-properties-and-json-stringify/53593328#53593328
-  JSON.stringify(object, (key, value) => {
-    allKeys.add(key);
-    return value;
-  });
-  return JSON.stringify(object, Array.from(allKeys).sort());
+  if (Array.isArray(object)) {
+    return `[${object.map(sortedJSONStringify).join(',')}]`;
+  }
+  if (object === null || typeof object !== 'object') {
+    return JSON.stringify(typeof object === 'function' ? null : object); // Normalise functions
+  }
+
+  const record = object as Record<string, unknown>;
+  const sortedValues = Object.keys(record)
+    .sort()
+    .map((key) => `${JSON.stringify(key)}:${sortedJSONStringify(record[key])}`);
+  return `{${sortedValues.join(',')}}`;
 }
 
 /** Stringifies an `object` and returns the hash of the resulting string. Sorts object fields
--- a/packages/utils/src/hashForSchema.ts
+++ b/packages/utils/src/hashForSchema.ts
@@ -20,12 +20,17 @@ export function hashString(string: string): string {
   return hash.toString(16);
 }
 
-/** Stringifies an `object`, sorts object fields in consistent order before stringifying it.
+/** Recursively serializes a value to JSON with object keys sorted alphabetically to produce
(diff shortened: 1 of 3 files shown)


packages/utils/src/hashForSchema.ts
```diff
@@ -26,13 +26,18 @@ export function hashString(string: string): string {
...
  */
 export function sortedJSONStringify(object: unknown): string {
-  const allKeys = new Set<string>();
-  // solution source: https://stackoverflow.com/questions/16167581/sort-object-properties-and-json-stringify/53593328#53593328
-  JSON.stringify(object, (key, value) => {
-    allKeys.add(key);
-    return value;
-  });
-  return JSON.stringify(object, Array.from(allKeys).sort());
+  if (Array.isArray(object)) {
```
heath-freenome, line 34: can you update the JS Doc for the function to include how the algorithm is now functioning.

heath-freenome (commented):
@samcragg Thanks for fixing an issue you found. Although your changes are breaking a bunch of core tests, so some of the assumptions you are making need to be reconsidered.

packages/utils/src/hashForSchema.ts
samcragg: Not a problem, I've updated that now

samcragg:
@heath-freenome thanks for taking a look so quickly. Those tests now pass locally for me.

I've added some additional tests for the actual function to prevent regression in the future (changing the utils causes pretty everything else to need to be built, so not the quickest dev cycle when you break it 😃)

heath-freenome:
Nice fix for the underlying issue, but I think the new `sortedJSONStringify` has a bug around `undefined` array elements that's worth fixing before merge.

**Repro:**
```js
sortedJSONStringify([])              // => "[]"
sortedJSONStringify([undefined])     // => "[]"   <- collides with []
sortedJSONStringify([null])          // => "[null]"
sortedJSONStringify([1, undefined])         // => "[1,]"  (invalid JSON)
sortedJSONStringify([undefined, undefined]) // => "[,]"   (invalid JSON)
```

**Root cause:** for a leaf `undefined`, the function hits:
```js
return JSON.stringify(typeof object === 'function' ? null : object);
```
`JSON.stringify(undefined)` returns the actual JS value `undefined`, not a string — even though the function's return type is declared `string`. When that lands in `object.map(sortedJSONStringify)`, `Array.prototype.join(',')` silently coerces it to `''`, so `[undefined]` renders identically to `[]`.

That's a bit worse than the original bug this PR fixes: it doesn't just conflate `null` and `undefined` in arrays, it conflates "array has an `undefined` item" with "array is empty." For `hashForSchema`/`hashObject` consumers, two arrays of different length/contents can now hash identically. It also produces invalid JSON for arrays with multiple/trailing `undefined` entries (`"[,]"`, `"[1,]"`).

**Suggested fix:** give the `undefined` leaf case a real string sentinel instead of letting it fall through to `JSON.stringify(undefined)`, e.g.:
```js
if (object === undefined) return '"__undefined__"'; // any string distinct from other outputs
```
placed before the generic `JSON.stringify(...)` fallback. Everything else (object-property filtering via `shouldIncludeValue`, matching native `JSON.stringify` semantics for dropped `undefined`/function properties) looks correct.

samcragg:
Good finds, thanks.

I've added a fix and expanded the tests and they pass now. I've also rebased the branch for the conflict in the changelog.

I've not changed the existing behaviour of `sortedJSONStringify(undefined)` returning `undefined` instead of a `string`, as I'm not sure if it's intended or not. Let me know if you do want me to change that though.

heath-freenome (approved):

PR #5174, opened 2026-07-28
fix(core): preserve seconds precision in TimeWidget and sync displayed value

sakshamspr:
## Summary

Fix `TimeWidget` so it stays in sync with the native browser `time` input and respects schema-level time precision.

Previously, the widget always normalized browser edits by appending `:00`. That caused two problems:

1. For minute-precision inputs, the browser edited `HH:mm` while the widget re-rendered `HH:mm:ss`, creating controlled input sync issues.
2. For second-precision inputs, such as when `schema.multipleOf < 60`, the widget still appended `:00`, producing invalid values like `09:22:12:00`.

## Root Cause

`TimeWidget` normalized values on change without considering whether the schema enabled seconds precision, and it rendered stored `HH:mm:ss` values directly back into native minute-precision inputs.

## Fix

- use `schema.multipleOf` as the precision signal for `time` values
- when seconds precision is enabled (`multipleOf < 60`):
  - preserve the browser value as-is
  - do not append `:00`
  - do not strip seconds on render
- when seconds precision is not enabled:
  - preserve backward-compatible stored values as `HH:mm:ss`
  - render stored `HH:mm:00` values as `HH:mm`
  - append `:00` when the browser emits `HH:mm`

## Backward Compatibility

This change does not alter the stored representation for minute-precision values, which remains `HH:mm:ss`.

The only behavioral change is:

- native minute-precision inputs now receive `HH:mm` instead of `HH:mm:ss`
- second-precision inputs now preserve `HH:mm:ss` instead of incorrectly appending `:00`

## Testing

- `pnpm nx lint @rjsf/core`
- `pnpm test -- StringField.test.tsx -t TimeWidget`
- `pnpm nx test @rjsf/core`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 6.8.0
 
+## @rjsf/core
+
+- Fixed `TimeWidget` to respect schema `multipleOf` time precision, preserving second-precision values while keeping minute-precision native inputs synchronized with their displayed value ([#5174](https://github.com/rjsf-team/react-jsonschema-form/pull/5174))
+
 ## @rjsf/validator-ata
 
 - Raised the `ata-validator` requirement to `^1.3.0`, which fixes a `$ref` that could not be resolved being treated as vacuous rather than reported as an error ([#5181](https://github.com/rjsf-team/react-jsonschema-form/pull/5181))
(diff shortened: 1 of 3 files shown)


heath-freenome (dismissed):

heath-freenome:
@sakshamspr Can you update `CHANGELOG.md` and I will merge

sakshamspr:
> @sakshamspr Can you update `CHANGELOG.md` and I will merge

updated the `CHANGELOG.md`

heath-freenome (approved):

PR #5184, opened 2026-08-08
Require ata-validator 1.6.0 in @rjsf/validator-ata

mertcanaltin:
Raises the `ata-validator` requirement in `@rjsf/validator-ata` from `^1.3.0` to `^1.6.0`.

### Why

1.4.0 is the reason to raise the floor. Where `new Function` is refused, the
validator used to survive the refusal and go on handling schemas it got wrong,
silently. It now runs on an interpreted engine instead, scoring 1286 of 1290 on
the official Draft 2020-12 suite with `eval` and `new Function` blocked, against
1188 before, with no schema failing to build.

That matters here because a page served under a strict Content-Security-Policy is
an ordinary deployment for a form library, not an edge case, and the same applies
to forms rendered on Cloudflare Workers or Deno Deploy. A caret range already
allows 1.4.0 for a fresh install; raising the floor is what moves projects with a
committed lockfile off the affected versions.

Two additions come along with it and change nothing for existing schemas. 1.5.0
adds `propertyDependencies`, which selects a subschema by the value of a property
rather than by its presence. 1.6.0 adds the JSON Schema v1 dialect, selected by
`$schema`; a schema that does not declare it takes the path it always did.

### Lockfile

The change is limited to the seven ata entries, versions and integrity hashes,
29 insertions against 29 deletions.

As in #5181, regenerating the lockfile wholesale with the pinned pnpm (10.17.1)
strips `libc:` from around sixty unrelated native packages across the workspace,
which would change optional dependency selection on musl. Those lines are put
back, so only the ata entries are touched. `pnpm install --frozen-lockfile`
accepts the result.

### Checks

Package tests pass unchanged against 1.6.0: 1274 tests across 8 files, coverage
still 100%. That includes the precompiled path, `compileSchemaValidators`,
`compileSchemaValidatorsCode`, `createPrecompiledValidator` and
`precompiledValidator`. `build:ts` and `lint` are clean.

The playground is not covered by this, since it needs a full workspace build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/validator-ata
 
 - Raised the `ata-validator` requirement to `^1.3.0`, which fixes a `$ref` that could not be resolved being treated as vacuous rather than reported as an error ([#5181](https://github.com/rjsf-team/react-jsonschema-form/pull/5181))
+- Raised the `ata-validator` requirement to `^1.6.0`, so that validation runs correctly where `new Function` is refused, such as a page under a strict Content-Security-Policy ([#5184](https://github.com/rjsf-team/react-jsonschema-form/pull/5184))
 
 ## @rjsf/validator-cfworker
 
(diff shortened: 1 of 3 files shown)


heath-freenome (approved):

mertcanaltin:
Posted this on the wrong pull request, sorry: this one had already merged. The 1.6.1 bump is #5186.

PR #5189, opened 2026-08-15
refactor: use native type guards instead of lodash, dropping antd's lodash dependency

jimmycallin:

Replaces lodash's simple type-guard functions with their native equivalents, and removes `@rjsf/antd`'s last lodash import along the way.

| | |
|---|---|
| `isNil(x)` | `x == null` (and `!isNil(x)` → `x != null`) |
| `isUndefined(x)` | `x === undefined` |
| `isString(x)` | `typeof x === 'string'` |
| `isNumber(x)` | `typeof x === 'number'` |
| `isFunction(x)` | `typeof x === 'function'` |

These are exact. The only behavioural difference from lodash is boxed primitives (`new String('x')`, `new Number(1)`), which do not occur in this codebase. Every call site is a direct call in conditional position — none are passed by reference — so there are no function-identity or memoisation effects. TypeScript narrowing is preserved, since `typeof` guards and `== null` narrow natively just as the lodash type predicates did.

`== null` is already sanctioned by the repo's lint config, which sets `eqeqeq: ["error", "always", { "null": "ignore" }]` specifically to permit it.

## The one site that needed an audit

`ObjectFieldTemplate`'s `isObject(colSpan)` is antd's last lodash import, so converting it makes **`@rjsf/antd` free of direct lodash imports**. Unlike the guards above this is not a mechanical swap — lodash's `isObject` also returns `true` for arrays and functions:

```ts
if (isObject(colSpan)) {
  if (typeof widget === 'string') { return colSpan[widget]; }
  // ...field, type
}
if (typeof colSpan === 'number') { return colSpan; }
return defaultColSpan;
```

**Arrays are unaffected** — `typeof [] === 'object'`, so they enter the branch either way, index to `undefined`, and return it.

**Functions do differ.** lodash enters the branch and returns `colSpan[widget]` (i.e. `undefined`); the `typeof` guard skips the branch and returns `defaultColSpan`. A function-valued `colSpan` is invalid configuration — it comes from `formContext.colSpan`, which is expected to be a number or a lookup object — and returning the computed default is the more sensible outcome. But it is a behaviour change, not an equivalence, so it is called out here rather than buried.

I checked this across 55 combinations of `colSpan` value (number, lookup object, empty object, array, function, `null`, `undefined`, string, `Date`, boolean) against widget/field/type context. The function case is the only divergence.

Removing lodash here also makes the `colSpanObj` intermediate dead, so it is deleted. It existed only because `isObject`'s `thing is object` predicate narrowed `colSpan` from `any` to `object`, which TypeScript will not let you index; the `typeof` guard does not narrow that way, so the lookups read `colSpan` directly.

## Measured effect: essentially none, and that is expected

Built ESM entries, bundled as a consumer's would be (peers external, nx cache bypassed):

| | min | gzip | own lodash imports |
|---|---|---|---|
| `@rjsf/antd` before | 1,216,302 | 383,334 | 3 files |
| `@rjsf/antd` after | 1,216,409 | 383,362 | **0 files** |
| `@rjsf/utils` before | 116,801 | 42,029 | 41 files |
| `@rjsf/utils` after | 116,823 | 42,010 | 37 files |

Slightly *larger* in antd's case. That is the point worth being explicit about: these are leaf functions with no internal dependency graph, and lodash's shared internals stay reachable until their last consumer goes. `@rjsf/antd` no longer imports lodash **directly**, but its bundle still contains `baseGet`, `baseIsEqualDeep` and `baseClone` transitively via `@rjsf/utils` and `@rjsf/core`.

So this PR should be judged as a safe, mechanical step toward removing the dependency — not as a size win. The size arrives when the shared chains go.

## Not included

- The remaining `isObject` sites — lodash's version, `@rjsf/utils`' own exported `isObject`, and a naive `typeof` check all disagree on arrays, Dates and functions, so each needs its own audit.
- `isEmpty` — returns `true` for `42`, `true`, `null` and `new Map()`.
- `noop` — passed as a prop reference, so it needs a stable module-level constant rather than an inlined arrow.
- `getClosestMatchingOption.ts` and `LayoutGridField.tsx` — converted in the deep-equality chain PR instead, so the two PRs share no source file and can merge in either order.

No public types or signatures change.

### The dependency is removed too

With no lodash imports left, `lodash` and `lodash-es` are also removed from `@rjsf/antd`'s `dependencies`, so consumers no longer install them on the package's behalf.

The `pnpm-lock.yaml` change is confined to that package's own importer section — a few lines — and `pnpm install --frozen-lockfile`, which is what CI runs, accepts it without rewriting the file.

One thing this does **not** do: the published bundle still contains lodash internals (`baseGet`, `baseIsEqualDeep`, `baseClone`) pulled in transitively via `@rjsf/utils` and `@rjsf/core`. Those go when the shared chains are removed there.

## Checklist

### DIFF
--- a/packages/utils/src/mergeDefaultsWithFormData.ts
+++ b/packages/utils/src/mergeDefaultsWithFormData.ts
@@ -1,5 +1,4 @@
 import get from 'lodash/get';
-import isNil from 'lodash/isNil';
 
 import type { GenericObjectType } from '../src';
 import isObject from './isObject';
@@ -104,8 +103,8 @@ export default function mergeDefaultsWithFormData<T = any>(
    */
   if (
     (defaultSupercedesUndefined &&
-      ((!(defaults === undefined) && isNil(formData)) || (typeof formData === 'number' && Number.isNaN(formData)))) ||
-    (overrideFormDataWithDefaults && !isNil(formData))
+      ((!(defaults === undefined) && formData == null) || (typeof formData === 'number' && Number.isNaN(formData)))) ||
+    (overrideFormDataWithDefaults && formData != null)
   ) {
     return defaults;
   }
--- a/packages/antd/package.json
+++ b/packages/antd/package.json
@@ -72,8 +72,6 @@
   },
   "dependencies": {
     "classnames": "^2.5.1",
-    "lodash": "^4.18.1",
-    "lodash-es": "^4.18.1",
     "rc-picker": "^4.11.3"
   },
   "devDependencies": {
--- a/packages/antd/src/templates/ObjectFieldTemplate/index.tsx
+++ b/packages/antd/src/templates/ObjectFieldTemplate/index.tsx
@@ -11,9 +11,6 @@ import type {
 import { canExpand, getTemplate, getUiOptions, titleId, buttonId } from '@rjsf/utils';
 import { Col, Row, ConfigProvider } from 'antd';
 import classNames from 'classnames';
-import isNumber from 'lodash/isNumber';
-import isObject from 'lodash/isObject';
-import isString from 'lodash/isString';
 
 /** The `ObjectFieldTemplate` is the template to use to render all the inner properties of an object along with the
  * title and description if available. If the object is expandable, then an `AddButton` is also rendered after all
@@ -74,19 +71,18 @@ export default function ObjectFieldTemplate<
         ? 24
         : 12;
 
-    if (isObject(colSpan)) {
-      const colSpanObj: GenericObjectType = colSpan;
-      if (isString(widget)) {
-        return colSpanObj[widget];
+    if (typeof colSpan === 'object' && colSpan !== null) {
+      if (typeof widget === 'string') {
+        return colSpan[widget];
       }
-      if (isString(field)) {
-        return colSpanObj[field];
+      if (typeof field === 'string') {
+        return colSpan[field];
       }
-      if (isString(type)) {
-        return colSpanObj[type];
+      if (typeof type === 'string') {
+        return colSpan[type];
       }
     }
-    if (isNumber(colSpan)) {
+    if (typeof colSpan === 'number') {
       return colSpan;
     }
     return defaultColSpan;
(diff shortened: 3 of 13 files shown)


jimmycallin:
Scope note: this PR originally also converted `utils/src/schema/getClosestMatchingOption.ts` and `core/src/components/fields/LayoutGridField.tsx`. Both files are also modified by #5190, which reindents the former substantially — so a conflict resolution there could have silently dropped these `typeof` conversions while still compiling and passing tests.

Those two files' conversions have moved to #5190. This PR now touches 11 files / 12 call sites, and the two PRs no longer share any source file, so they can merge in either order without conflict. No conversions were lost.

heath-freenome:
@jimmycallin Thanks for making all of these changes... I'm curious why remove lodash?

jimmycallin:
> @jimmycallin Thanks for making all of these changes... I'm curious why remove lodash?

Analyzing the dependency graph and bundle size of our app showed rjsf to be a large chunk of the output, in a way that felt unearned. Lodash has generally great native equivalents today, and was an easy enough starting point for me to help reduce the size.

heath-freenome:
> > @jimmycallin Thanks for making all of these changes... I'm curious why remove lodash?
> 
> Analyzing the dependency graph and bundle size of our app showed rjsf to be a large chunk of the output, in a way that felt unearned. Lodash has generally great native equivalents today, and was an easy enough starting point for me to help reduce the size.

Got it. Be sure to remove the `@types/lodash`, `@types/lodash-es` and `lodash-es`

packages/utils/src/mergeDefaultsWithFormData.ts
```diff
@@ -1,5 +1,4 @@
 import get from 'lodash/get';
```
heath-freenome, line 1: I guess you aren't fully eliminating lodash after all.

heath-freenome (approved):

packages/utils/src/mergeDefaultsWithFormData.ts
jimmycallin: get is less obvious and needs more careful attention so leaving it for a later iteration

PR #5190, opened 2026-08-15
perf(utils,core): eliminate lodash's deep-equality chain from the bundle

jimmycallin:

`@rjsf/utils` and `@rjsf/core` pull in lodash's deep-equality internals (`baseIsEqualDeep` and its supporting chain — `equalArrays`, `equalObjects`, `equalByTag`, `Stack`, `getTag`, …). That code is reachable from four separate imports, so removing any one of them individually changes the bundle by roughly nothing while the shared ~8KB stays put. Removing the last consumer is what lets it drop out.

This PR removes all of them together, plus the adjacent lodash calls in the same block.

| function | where |
|---|---|
| `transform`, `merge`, `uniq`, `flattenDeep` | `utils/src/schema/retrieveSchema.ts` (one block) |
| `uniq` | `utils/src/shouldRenderOptionalField.ts` |
| `pickBy`, `keys` | `utils/src/getChangedFields.ts` |
| `reduce` | `utils/src/schema/getClosestMatchingOption.ts` |
| `isEqual` | `core/src/components/fields/LayoutGridField.tsx` (uses the existing `deepEquals`) |

`transform` and `merge` are not merely adjacent but coupled — `transform`'s callback populates `childrenLists`, which the `merge` line then consumes — so they are rewritten as one loop.

### Also converts the type guards in these two files

`LayoutGridField.tsx` and `getClosestMatchingOption.ts` additionally have lodash's simple type guards converted to native ones (`isString`/`isNumber`/`isUndefined`/`isFunction` → `typeof`, `isNil` → `== null`). Those conversions belong here rather than in the type-guards PR (#5189) so that the two PRs share no source file and can merge in either order — this PR reindents `getClosestMatchingOption.ts` substantially, and a conflict resolution across the two could silently drop the `typeof` conversions while still compiling and passing tests.

### Measured effect

Built ESM entries bundled as a consumer's bundler would (React external, nx cache bypassed):

| | min | gzip | `baseIsEqualDeep` present |
|---|---|---|---|
| `@rjsf/utils` before | 116,801 | 42,035 | yes |
| `@rjsf/utils` after | 110,902 | **39,835** | **no** |
| `@rjsf/core` before | 387,786 | 129,309 | yes |
| `@rjsf/core` after | 381,828 | **127,042** | **no** |

≈2.2KB gzipped off `@rjsf/utils` and ≈2.3KB off `@rjsf/core`.

## Notes on equivalence

Three replacements are not literal one-for-one swaps and deserve attention:

**1. `merge(recurseList, uniq(flattenDeep(childrenLists)))`** was doing something non-obvious. `merge` on two *arrays* assigns index-wise, not by append. It behaved as an append only because every `childList` starts as a copy of `recurseList` and only ever grows, making `recurseList` an exact prefix of the deduped result. The replacement appends the missing refs directly. This is equivalent under that invariant and preserves the no-op case where `properties` yields no children — a naive clear-and-refill rewrite would wrongly empty the list there.

**2. `Object.entries(resolvedSchema[PROPERTIES_KEY] ?? {})`** — the `?? {}` preserves `lodash/transform`'s tolerance of a nullish collection. `'properties' in schema` passes for `{ properties: undefined }`, where a bare `Object.entries` throws. A test covers this and was verified to fail when the guard is removed.

**3. `getChangedFields` is now own-properties-only.** lodash's `pickBy` uses `keysIn`, which walks the prototype chain, so the function previously mixed `keys` (own) with `pickBy` (own and inherited) three lines apart. The `isPlainObject` guard means the only way to observe the difference is an enumerable property on `Object.prototype`, in which case the old behaviour reported that polluted key as a changed field — and `Form` then wrote it into the error schema as a cleared entry. The new behaviour cannot return anything that is not an own property of the form data.

`deepEquals` differs from `lodash/isEqual` in treating functions as equal and tracking circular references. The `LayoutGridField` call site compares sorted arrays of primitives, so the two agree there; they also agree on `NaN`, `-0` and sparse arrays.

No public types or signatures change.

## Review tip

`getClosestMatchingOption.ts` shows ~127 changed lines but is **+2 / −7** ignoring whitespace — the callback body de-indents one level when moving from lodash's 3-argument form to native `Array.prototype.reduce`. Reviewing with `?w=1` makes this much easier to read.

## Checklist

### DIFF
--- a/packages/core/src/components/fields/LayoutGridField.tsx
+++ b/packages/core/src/components/fields/LayoutGridField.tsx
@@ -11,6 +11,7 @@ import type {
 } from '@rjsf/utils';
 import {
   ANY_OF_KEY,
+  deepEquals,
   getDiscriminatorFieldFromSchema,
   getTemplate,
   getTestIds,
@@ -34,12 +35,8 @@ import has from 'lodash/has';
 import includes from 'lodash/includes';
 import intersection from 'lodash/intersection';
 import isEmpty from 'lodash/isEmpty';
-import isEqual from 'lodash/isEqual';
-import isFunction from 'lodash/isFunction';
 import isObject from 'lodash/isObject';
 import isPlainObject from 'lodash/isPlainObject';
-import isString from 'lodash/isString';
-import isUndefined from 'lodash/isUndefined';
 import last from 'lodash/last';
 import set from 'lodash/set';
 
@@ -177,7 +174,7 @@ export function computeFieldUiSchema<T = any, S extends StrictRJSFSchema = RJSFS
     set(fieldUiSchema, [UI_GLOBAL_OPTIONS_KEY], globalUiOptions);
   }
   let { readonly: uiReadonly } = getUiOptions<T, S, F>(fieldUiSchema);
-  if (forceReadonly === true || (isUndefined(uiReadonly) && schemaReadonly === true)) {
+  if (forceReadonly === true || (uiReadonly === undefined && schemaReadonly === true)) {
     // If we are forcing all widgets to be readonly, OR the schema indicates it is readonly AND the uiSchema does not
     // have an overriding value, then update the uiSchema to set readonly to true. Doing this will
     uiReadonly = true;
@@ -215,7 +212,7 @@ export function conditionMatches(
   const values = flatten([value]).sort();
   switch (operator) {
     case Operators.ALL:
-      return isEqual(data, values);
+      return deepEquals(data, values);
     case Operators.SOME:
       return intersection(data, values).length > 0;
     case Operators.NONE:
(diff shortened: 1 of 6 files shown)


jimmycallin:
Scope note: this PR now also carries the native type-guard conversions (`isString`/`isNumber`/`isUndefined`/`isFunction` → `typeof`) for the two files it shares with #5189 — `getClosestMatchingOption.ts` and `LayoutGridField.tsx`.

Those conversions were originally in #5189, but because this PR reindents `getClosestMatchingOption.ts` substantially, a conflict resolution could have silently dropped them while still compiling and passing tests. Carrying them here keeps the two PRs off the same files so they merge in either order.

heath-freenome (approved):

PR #4593, opened 2025-05-01
Merge the RJSF V6 long-lived branch

heath-freenome:
### Reasons for making this change

Merging in ALL of the rjsf v6 changes down to the `main` branch in preparation for the v6 beta release

### Checklist

- [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.
  - [ ] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #5194, opened 2026-08-17
chore: add knip and remove the unused code and dependencies it found

jimmycallin:

Follow-up to the lodash PRs (#5189–#5193), from the same goal of shrinking what consumers install. This adds [knip](https://knip.dev) and applies everything it found.

The bulk of the win is **unused runtime dependencies in published packages** — install weight every consumer pays for packages the source never imports:

| Package | Removed from `dependencies` |
|---|---|
| `@rjsf/core` | `prop-types` |
| `@rjsf/antd` | `rc-picker` |
| `@rjsf/chakra-ui` | `@ark-ui/react`, `@zag-js/react`, `@emotion/memoize`, `react-select` (the theme uses `chakra-react-select`) |
| `@rjsf/shadcn` | `@radix-ui/react-popover`, `@radix-ui/react-select`, `@react-icons/all-files`, `tailwindcss-animate`, `uuid` |
| `@rjsf/fluentui-rc` | `@griffel/react` |
| `@rjsf/semantic-ui` | `@fluentui/react-component-ref` |
| `@rjsf/docs` | `ajv` |

`@fluentui/react-component-ref` is safe to drop because `semantic-ui-react@2.1.5` declares its own copy, so consumers still resolve it.

Unused devDependencies also go: the root (including `tslib`, since `tsconfig.base.json` sets `importHelpers: false`), the playground's dead webpack toolchain plus twelve unused runtime deps (it builds with Vite), mantine's react-bootstrap leftovers, primereact, the three validators, and the six theme packages that never import `@rjsf/validator-ajv8` — along with their now-stale TypeScript project references.

## `@rjsf/daisyui`

**Superseded by #5224.** knip surfaced a ~20-file dead island here (daisyui arrived whole in the v6 merge (#4593) with its wiring written from scratch, orphaning the per-directory barrels and leaving most of the theme contract implemented but never exported). This PR originally added the missing `generate*` exports; #5224 then did that properly on `main` — wiring up the dead code instead of just exporting it — so after merging `main` back in, this PR no longer touches daisyui source at all. What remains: the `@rjsf/validator-ajv8` devDependency and tsconfig reference are restored (the new `DaisyUIForm.test.tsx` from #5224 imports it), and knip keeps treating daisyui's `src/` as entry points for the still-orphaned barrels, deferred to v7 with the reason inline in `knip.jsonc`. Only the test fixtures superseded by `testing/testSetup.ts` stay deleted (never published).

## Two things deliberately left for the next major

I checked the whole diff for anything that can't ship in a 6.x minor, and backed out two changes that couldn't:

1. **`Widgets` stays the `generateWidgets` function.** Every other theme exports the generated object; daisyui's `Templates` and `Theme` are already objects, so `Widgets` is the lone outlier and `<Form widgets={Widgets} />` currently hands the registry a function. Fixing it changes the runtime type of a published export, so the shape correction is deferred (#5224 added `generateWidgets` alongside; knip ignores the resulting duplicate export, reason in the config).
2. **`semantic-ui-css` stays a runtime dependency.** No source file imports it, but consumers load its stylesheet per the package README, so making it an optional peer would break them on a minor upgrade. knip ignores it instead, with the reason in the config.

## The `./lib/*.js` wildcard

Every package's `exports` map has a `./lib/*.js` wildcard, and `*` matches path separators, so every internal file is resolvable as a deep import — `@rjsf/daisyui/lib/utils.js`, `lib/daisyForm/index.js`, the barrels. Per review, the daisyui file deletions this enabled questions about are reverted and deferred to v7 (see above).

Same wildcard has a second effect worth knowing: because it makes knip treat every source file as a public entry point, **unused-export detection is effectively off repo-wide**. Narrowing those maps in a major would switch it on for free.

## knip setup

`knip.jsonc` at the root, wired into `sanity-check` and CI (gated to the 22.x matrix leg, since the analysis doesn't vary by Node version). Five suppressions, each documented inline with its reason. Every config entry was verified load-bearing by removing it and confirming knip reports again. (An earlier revision also suppressed `lodash-es` for the `lodash/*` build-time rewrite; the lodash removal PRs merging made that moot, and knip itself flagged the stale ignore.)

## Verification

After merging `main` back in (post #5224/#5219): `pnpm run build` (all 18 projects), `pnpm test` (17 projects), `pnpm run lint`, and `pnpm run knip` all pass, and every knip suppression was re-verified load-bearing against the merged tree.

`packages/chakra-ui/src/ChakraFrameProvider.tsx` fails `cs-check`, but that is pre-existing on `main` and untouched here.

## Note on merge order

~~Overlaps with the lodash PRs in three `package.json` files, all in the same `dependencies` blocks, so expect conflicts depending on merge order: `packages/antd/package.json` with #5189, `packages/semantic-ui/package.json` with #5192, `packages/shadcn/package.json` with #5193, plus `pnpm-lock.yaml` with all three. Happy to rebase this one last — it's the least urgent of the set.~~ Resolved: the lodash PRs and size-limit (#5207) merged first, and this branch is rebased on top of them. `main` was merged in again after #5224 and #5219 landed.

## Checklist

- [x] I'm updating dependencies


heath-freenome:
## Code Review

The dependency cleanup is well-reasoned and the knip config is correctly documented. Three issues survive verification:

---

### `packages/daisyui/src/theme/Theme.tsx` — `ThemeContext` becomes an unannounced breaking export removal

The PR changes `export const ThemeContext = createContext(...)` → `const ThemeContext = ...`, making it module-private. Since `packages/daisyui/src/theme/index.ts` does `export * from './Theme'`, the exported symbol disappears from the `@rjsf/daisyui/lib/theme/index.js` deep-import path (accessible today via the `"./lib/*.js"` wildcard in the exports map). The PR description lists two deferred breaking changes (`Widgets` type and `semantic-ui-css`) but doesn't mention this one. The `useTheme` hook is the functional replacement, but any consumer calling `useContext(ThemeContext)` directly will break silently at runtime with `undefined`. Worth an explicit callout in the PR description or changelog, even under the "deep imports are not blessed" policy.

---

### `packages/daisyui/src/widgets/` barrel files (multiple) — per-widget deep imports silently removed

Twelve per-widget `index.ts` barrel files (e.g. `src/widgets/CheckboxWidget/index.ts`) are deleted, and the named widget re-exports are removed from `Widgets.tsx`. Due to the `"./lib/*.js"` wildcard in the exports map, `import CheckboxWidget from '@rjsf/daisyui/lib/widgets/CheckboxWidget/index.js'` and `import { CheckboxWidget } from '@rjsf/daisyui/lib/widgets/index.js'` are currently valid import paths that now 404. The PR description discusses the wildcard policy and a question about whether these should move to the next major — this is the concrete illustration of that question. The position taken (wildcard is for lib/dist split, not frozen deep-API) is defensible, but it would be worth a short `CHANGELOG` note so any daisyui consumer who happened to reach for a deep widget import knows to switch to the `generateWidgets()` pattern.

---

### `packages/semantic-ui/test/*.test.tsx` — removal of `vi.mock('@fluentui/react-component-ref')` guards

`semantic-ui-react@2.1.5` lists `@fluentui/react-component-ref@~0.63.0` as a direct `dependency` and uses it in at least 9 of its distributed files (`Portal.js`, `Dropdown.js`, `TextArea.js`, etc.). The mocks being removed — `Ref: vi.fn().mockImplementation(({ children }) => children)` — were added because tests failed without them in jsdom. The dependency removal from `@rjsf/semantic-ui` is correct (the source never imports it directly, and `semantic-ui-react` pulls its own copy). The mock removal is plausible if vitest/jsdom now handles `@fluentui/react-component-ref`'s `Ref` component cleanly, which the PR author states is the case. If CI passes on this PR, this is fine — but if anything about the test-runner environment changes later (jsdom version bump, vitest update), these tests may silently break again. Low priority, but consider a brief comment explaining *why* the mocks were safe to remove, so the next person doesn't re-add them reflexively.

---

*Review generated with [Claude Code](https://claude.ai/code)*

jimmycallin:
Rebased onto `main` and addressed all three points.

**1 & 2 — daisyui `ThemeContext` and per-widget deep imports.** Both are now called out in the changelog under a new `## @rjsf/daisyui` heading for 6.8.0, framed as what they are: not public API, but reachable through the unblessed `@rjsf/daisyui/lib/*.js` wildcard. Each names its replacement — `useTheme()` for `ThemeContext`, `generateWidgets()` for the widget barrels — and I verified both are actually exported from `src/index.ts`, so anyone who hits the break has somewhere to go.

**3 — semantic-ui `@fluentui/react-component-ref`.** I dug into why the mocks existed rather than leaving a note in the source, so here is the whole picture in one place.

The mocks were introduced in #3150 (Sept 2022), the TypeScript conversion, as `jest.mock(...)`. At that point the suite rendered through `react-test-renderer`, which builds a synthetic tree with no host DOM nodes. `Ref` from `@fluentui/react-component-ref` resolves its child to a real DOM node, so under `react-test-renderer` it had nothing to attach to and the tests failed — hence the pass-through stub. #5080 (May 2026) migrated the suite from jest to vitest and rewrote the call as `vi.mock(...)`, but that was a mechanical translation; nobody re-checked whether the underlying failure still reproduced.

It does not. The suite now renders via `@testing-library/react` under `environment: 'jsdom'`, so `Ref` gets a genuine DOM node and behaves normally. The condition the mock was written for stopped existing when the renderer changed, not when any dependency version moved. All 146 tests pass with the mocks gone, including the `SelectWidget` and `TextareaWidget` paths that reach semantic-ui's `Dropdown` and `TextArea` — the components that actually use `Ref`.

Separately, on the dependency itself: dropping `@fluentui/react-component-ref` from `@rjsf/semantic-ui`'s `dependencies` changes nothing about what gets installed. Our source never imported it, and `semantic-ui-react@2.1.5` declares it directly — `pnpm why` in the package still resolves `0.63.1`, the same version the direct entry pinned.

**Rebase note:** the lodash-removal PRs that landed on `main` (#5189, #5192) conflicted here. Resolved so `packages/antd` keeps only `classnames` (both `rc-picker` and lodash dropped) and `packages/semantic-ui` keeps only `semantic-ui-css`. Lockfile regenerated from main's rather than hand-merged.

**Verification:** build 18/18, tests 17/17 projects, lint clean, `knip` clean. One `cs-check` failure in `packages/chakra-ui/src/ChakraFrameProvider.tsx` is pre-existing from #5175 on `main` and untouched by this branch.

jimmycallin:
Correction to my previous comment on point 3: I have dropped the explanatory comments from the semantic-ui test files.

A comment describing a mock that no longer exists in the file is documentation of history, not of the code — the next reader sees a note about machinery they cannot find. Git history and this thread already record why the mocks went away, and that is the right place for it. The tests read clean without them, and all 146 still pass.

Points 1 and 2 are unchanged — the daisyui removals stay documented in the changelog, which is where removed behavior does belong.

heath-freenome:
## Re-review

All three findings from the previous review are addressed — the changelog entry covers \`ThemeContext\` and the widget barrels, and the follow-up comment gives a complete account of why the semantic-ui mocks were safe to remove. Two things remain worth a look:

---

### \`packages/daisyui/src/DaisyUIForm.tsx\` — \`generateTheme()\` now called twice at module load

Before the PR, \`DaisyUIForm.tsx\` imported the pre-computed singleton \`Theme\` from \`./theme\` (\`const Theme = generateTheme()\` at line 37 of \`Theme.tsx\`) and passed it directly to \`withTheme\`. After the PR, \`DaisyUIForm.tsx\` calls \`generateTheme<T, S, F>()\` fresh inside \`generateForm()\`, and \`export default generateForm()\` fires it again at module load. \`Theme.tsx\`'s module-level singleton is still computed for the sake of \`export { default as Theme }\` in \`src/index.ts\`. Net result: \`getDefaultRegistry()\`, \`generateWidgets()\`, and \`generateTemplates()\` each run twice when \`@rjsf/daisyui\` is first imported.

One-time cost, not a correctness bug — but the old approach (import the singleton, don't re-generate) was cheaper. Simplest fix in \`DaisyUIForm.tsx\`: import \`Theme\` (the default) rather than \`generateTheme\`, cast it to \`ThemeProps<T, S, F>\`, and pass it to \`withTheme\`, matching the pre-refactor pattern.

---

### \`CHANGELOG.md\` — \`buttonTemplates\` removal not called out

The changelog entry correctly names \`ThemeContext\` and the per-widget barrels as the two things removed from the unblessed \`@rjsf/daisyui/lib/*.js\` deep-import path. But \`lib/templates/ButtonTemplates/index.js\` also loses its \`export default buttonTemplates\`, and \`lib/templates/index.js\` (which re-exported it as a named export) is deleted entirely. Any consumer who reached for \`import buttonTemplates from '@rjsf/daisyui/lib/templates/ButtonTemplates/index.js'\` will silently break without a migration path mentioned. The parallel to the widget-barrel callout would be something like "use \`generateTemplates()\` and destructure the button template you need." Very minor given the explicit "wildcard is not public API" stance, but worth a one-liner alongside the existing two bullets.

---

Everything else checked out: all removed \`export\` keywords have no in-repo callers, the \`@rjsf/validator-ajv8\` devDep removals are safe because \`@rjsf/snapshot-tests\` still declares it and the workspace hoists it, all runtime dep removals are confirmed unused in source, and \`tslib\` removal is correctly backed by \`"importHelpers": false\` in \`tsconfig.base.json\`.

*Review generated with [Claude Code](https://claude.ai/claude-code)*

jimmycallin:
Thanks — changelog bullet added in 554258c.

On the double `generateTheme()`: I looked, and this is the existing repo-wide pattern rather than something the PR introduced. `packages/mui/src/Theme/Theme.tsx` ends with `export default generateTheme()` for the `Theme` export, and `MuiForm.tsx`'s `generateForm()` calls `generateTheme<T, S, F>()` again — so mui pays exactly the same one-time cost at module load, as do the other themes that follow the `generateForm`/`generateTheme` contract. daisyui's `DaisyUIForm.tsx` is now character-for-character mui's `MuiForm.tsx` apart from the import path.

Importing the singleton and casting it to `ThemeProps<T, S, F>` would be cheaper, but it also loses the point of the generic: `generateTheme<T, S, F>()` types the theme at the caller's generics, whereas the cast just asserts the `any`-generic singleton is whatever the caller asked for. Since making daisyui stop being the odd theme out is the whole reason this file changed, I'd rather leave it matching. If the double-generation is worth removing it's a change for every theme at once, which feels like its own PR.

CHANGELOG.md
```diff
@@ -22,6 +22,13 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Fixed `TimeWidget` to respect schema `multipleOf` time precision, preserving second-precision values while keeping minute-precision native inputs synchronized with their displayed value ([#5174](https://github.com/rjsf-team/react-jsonschema-form/pull/5174))
 
+## @rjsf/daisyui
```
heath-freenome, line 44: Move these into 6.8.1 since we've released 6.8.0

packages/core/package.json
```diff
@@ -15,7 +15,6 @@
     "precommit": "lint-staged",
     "publish-to-npm": "pnpm run build && pnpm publish",
     "test": "vitest run",
-    "test:debug": "node --inspect-brk ../../node_modules/.bin/jest",
```
heath-freenome, line 17: Why remove this? Back in the day I'd use this to debug tests

knip.jsonc
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "https://unpkg.com/knip@6/schema-jsonc.json",
+  "workspaces": {
+    ".": {
+      // Invoked by tsc-alias through the `afterDeclarations` replacers in the
+      // per-package tsconfig.build.json files, which knip cannot follow.
+      "entry": ["tsc-alias-replacer/*.ts"]
+    },
+    "packages/daisyui": {
```
heath-freenome, line 11: Where are the rest of the packages?

heath-freenome:
@jimmycallin Can you respond to my questions please

packages/core/package.json
jimmycallin: The script was broken: it pointed at `../../node_modules/.bin/jest`, which no longer exists since the vitest migration, so knip reported `jest` as an unlisted binary and I deleted the line instead of fixing it.

Restored in f79c2a434, pointed at vitest and matching the one `@rjsf/utils` already has:

```json
"test:debug": "node --inspect-brk ../../node_modules/.bin/vitest --pool=forks"
```

knip.jsonc
jimmycallin: knip discovers all workspaces itself from `pnpm-workspace.yaml`, so every package is already analyzed. The `workspaces` block here only holds per-package overrides — the five entries are the packages that needed a suppression. The other thirteen need none, so listing them would just be empty objects.

`pnpm run knip` covers all 18 projects; you can confirm with `pnpm run knip -- --debug`, which prints the resolved workspace list.

CHANGELOG.md
jimmycallin: `main` has since moved to `6.9.0`, so these will land under that heading rather than `6.8.1`. Doing it as part of the rebase — this PR still needs one for the three `package.json` overlaps with the lodash PRs plus `pnpm-lock.yaml`, and moving the entries now would only re-conflict.

jimmycallin:
@heath-freenome was it intentional to close this pr?

jimmycallin:
@heath-freenome Reopened this PR just to verify it was intentionally closed.

heath-freenome:
> @heath-freenome Reopened this PR just to verify it was intentionally closed.

It was an accident

packages/daisyui/src/daisyForm/index.ts
```diff
@@ -1,2 +0,0 @@
-export { default } from './DaisyForm';
```
nickgros: Based on the package.json and build config, I think it's possible for consumers to write deep imports against these Daisy files, so I think this might be a breaking change.

Maybe we separate these removals of unnecessary barrels into a new PR and we'll look at starting a v7 branch very soon?

packages/daisyui/src/templates/ArrayFieldDescriptionTemplate/ArrayFieldDescriptionTemplate.tsx
```diff
@@ -1,19 +0,0 @@
-import type { FormContextType, StrictRJSFSchema, RJSFSchema, ArrayFieldDescriptionProps } from '@rjsf/utils';
-
-/** The `ArrayFieldDescriptionTemplate` component renders the description for an array field
- * with DaisyUI styling, displaying it as a small text with accent color.
- *
- * @param props - The `ArrayFieldDescriptionProps` for the component
- */
-export default function ArrayFieldDescriptionTemplate<
```
nickgros, line 10: The Daisy theme probably should be using these rather than the `core` (so I think this is really a bug). I think @heath-freenome plans to make a fix to properly use these fields, so we won't remove them

packages/daisyui/src/daisyForm/index.ts
jimmycallin: Agreed — deferred to v7. ff3e74bfa restores every deleted `src/` file (this one included), so the `lib/*.js` deep-import surface is byte-identical to `main` again. What remains of the daisyui changes is purely additive: the `generateForm`/`generateTemplates`/`generateTheme`/`generateWidgets` exports and the dependency cleanup. knip.jsonc now treats all of daisyui `src/` as entry points with the v7 deferral noted inline, so it stops flagging the orphans; the changelog removal bullets are gone too.

packages/daisyui/src/templates/ArrayFieldDescriptionTemplate/ArrayFieldDescriptionTemplate.tsx
jimmycallin: Restored in ff3e74bfa, along with `ArrayFieldTitleTemplate` — both files are back exactly as on `main`, still unregistered, so wiring them into the theme stays open for @heath-freenome's fix.

heath-freenome:
@jimmycallin Ok, I fixed the DaisyUI theme

jimmycallin:
Thanks @heath-freenome! Merged `main` back in — #5224 supersedes the daisyui changes here, so this PR no longer touches daisyui source (just restores the `@rjsf/validator-ajv8` devDependency your new test needs). Conflicts resolved, all checks green, description updated.

heath-freenome (approved):

Issue #5196, opened 2026-08-18
omitExtraData infinite-loops / OOMs when patternProperties has an empty {} schema over an array


MatinF:
### Prerequisites

- [x] I have searched the existing issues
- [x] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [x] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

6.7.1, 6.8.0
# `omitExtraData` infinite-loops / OOMs when `patternProperties` has an empty `{}` schema over an array

**Package/version:** `@rjsf/utils` 6.7.1 — the relevant code is byte-identical at tag `v6.8.0`
**Validator:** `@rjsf/validator-ajv8` (not validator-specific)
**Form props:** `omitExtraData: true` + `liveOmit: true`

## Description

`omit()` in `packages/utils/src/schema/omitExtraData.ts` returns the **source object by reference** for an empty subschema:

```ts
if (schemaDef === true || isEmpty(schemaDef as object)) {
  return source;
}
```

`setProperty` writes that reference to `target[key]`, and passes it back in on a later visit:

```ts
const v = omit(schemaDef, value, target[key]);
```

so `target` can become the *same object* as `source`. When the aliased value contains an array, `handleArray` runs:

```ts
for (let i = 0; i < source.length; i += 1) {
  target.push(omit(items as S | boolean, source[i]));
}
```

with `target === source`. The loop feeds itself and never terminates. It also mutates the caller's `formData` in place.

**Trigger:** an object with `additionalProperties: false` + `patternProperties: { "<key>": {} }` — a common way to whitelist keys injected by `dependencies` — where `<key>` is re-visited through the `dependencies` `oneOf` branch and holds an array with at least one element.

## Reproduction in the playground

<https://rjsf-team.github.io/react-jsonschema-form/>

1. Paste the schema and form data below.
2. Enable **Omit extra data** *and* **Live omit**.
3. Change the **ID format** dropdown inside *Routing → Mux → Messages*.

The tab freezes immediately, with **no console output**.

Notes on the steps:

[...]

Issue #5176, opened 2026-07-29
Inifinite loop and browser crash with omitExtraData option


slafon51:
### Prerequisites

- [x] I have searched the existing issues
- [x] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [x] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

6.7.1

### Current Behavior

Hello,
In some cases, the `omitExtraData` option can cause an infinite loop and crash the browser tab.
This appears to be due to the resolution of issue #5142 
I can reproduce this bug by using schemas that define the same object - which has a list-type property - in multiple `allOf` clauses.

### Expected Behavior

The omitExtraData option should not cause infinite loop.

### Steps To Reproduce

[...]

Issue #3920, opened 2023-10-25
Problem with contains condition in the array field and conditionally rendered subschema with the integer keys when liveOmit and omitExtraData are set to true


magaton:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

mui

### Version

5.13.2

### Current Behavior

I have hit a strange bug, not sure where to look, so I will post working and not working schemas. Hopefully, the maintainers will have an idea of what is wrong.

#1. `if-then` with `contains`. The conditionally rendered schema has `string` keys. This works.
#2. `if-then` with `contains`. The conditionally rendered schema has `integer` keys. This does not work.

Both work well when `liveOmmit` and `omitExtraData` are set to `false`

### Expected Behavior

I would expect that both scenarios work with turned on liveOmmit and omitExtraData. 
However it seems that they reset the field that is used for rendering the dependent schema, but only if that field options match the keys in the dependent schema.
This looks like a bug to me

### Steps To Reproduce

[...]

PR #5200, opened 2026-08-19
Fix 5194: remove over-reaching additionalProperties:false post-processing

heath-freenome:

Fixes #5194, #5196, #3920

The additionalProperties:false post-processing block introduced in the #5176 fix built its key allowlist solely from the parent schema's `properties`, incorrectly deleting keys written by winning oneOf/anyOf branches and if/then branches. The block is removed entirely, restoring pre-PR behaviour where branch-granted keys survive omitExtraData filtering.

### Checklist

### DIFF
--- a/packages/utils/src/schema/omitExtraData.ts
+++ b/packages/utils/src/schema/omitExtraData.ts
@@ -491,25 +491,25 @@ export default function omitExtraData<
       }
     }
 
-    const result = handleDependencies(localSchema, source, handleConditions(localSchema, source, filtered));
-
-    // Post-prune: when additionalProperties is false, delete any key that survived branch
-    // merging (oneOf/anyOf/if-then) but is not covered by the parent schema's own properties
-    // or patternProperties. Mirrors the SJSF implementation — branch-granted keys are still
-    // filtered through omit() above, but any that end up outside the schema's declared property
-    // set are removed here so the result honours additionalProperties:false.
-    if (localSchema.additionalProperties === false && isObjectValue(result)) {
+    // Post-prune after conditions (oneOf/anyOf/if-then) but BEFORE dependencies: keys added by
+    // dependency sub-schemas are legitimate extensions of the allowed property set and must not be
+    // pruned. Keys added only by winning conditional branches are still removed here so the result
+    // honours additionalProperties:false.
+    const afterConditions = handleConditions(localSchema, source, filtered);
+    if (localSchema.additionalProperties === false && isObjectValue(afterConditions)) {
       const knownKeys = new Set(Object.keys(localSchema.properties ?? {}));
       const patterns = localSchema.patternProperties
         ? Object.keys(localSchema.patternProperties).map((p) => new RegExp(p))
         : [];
-      for (const key of Object.keys(result)) {
+      for (const key of Object.keys(afterConditions)) {
         if (!knownKeys.has(key) && !patterns.some((re) => re.test(key))) {
           // oxlint-disable-next-line no-param-reassign
-          delete result[key];
+          delete afterConditions[key];
         }
       }
     }
+
+    const result = handleDependencies(localSchema, source, afterConditions);
 
     return result ?? (materializeSource ? source : undefined);
   }
--- a/packages/utils/src/schema/omitExtraData.ts
+++ b/packages/utils/src/schema/omitExtraData.ts
@@ -169,7 +169,7 @@ export default function omitExtraData<
   }
 
   /** Copies schema-defined properties from `source` into `target`, applying `omit` recursively for
-   * each value. Handles `properties`, `patternProperties`, `additionalProperties`, and `propertyNames`.
+   * each value. Handles `properties`, `patternProperties`, and `additionalProperties`.
    * Optional object-valued properties are pruned when every key in the filtered result is both
    * optional (per the inner schema's `required`) and empty (per `isValueEmpty`). This preserves
    * optional objects whose required children have empty values, while still dropping objects whose
@@ -183,7 +183,7 @@ export default function omitExtraData<
    * @returns - `target` after all schema-defined properties have been processed
    */
   function handleObject(childSchema: S, source: GenericObjectType, target: GenericObjectType): GenericObjectType {
-    const { properties, additionalProperties, patternProperties, propertyNames } = childSchema;
+    const { properties, additionalProperties, patternProperties } = childSchema;
     const requiredSet = new Set(childSchema.required ?? []);
 
     /** Recursively omits extra data from `value` via `omit`, then conditionally writes the result to
@@ -266,46 +266,51 @@ export default function omitExtraData<
       }
     }
 
-    // When propertyNames is present, the schema only constrains key names — all source keys are valid.
-    if (propertyNames !== undefined) {
-      for (const [key, value] of Object.entries(source)) {
-        // oxlint-disable-next-line no-param-reassign
-        target[key] = value;
-      }
-    }
-
     return target;
   }
 
   /** Filters array elements from `source` into `target` according to `schema.items` and
-   * `schema.additionalItems`. For tuple schemas (`items` is an array) each element is filtered by its
-   * per-index schema; elements beyond the tuple length are covered by `additionalItems` when present.
-   * For list schemas (`items` is a single schema) every element is filtered by that schema.
+   * `schema.additionalItems`. Only called when `items` is defined (guarded in `omit`). Uses index
+   * assignment (not push) so a pre-populated `target` from a prior branch is overwritten rather than
+   * appended to, preventing item duplication. For tuple schemas (`items` is an array) elements up to
+   * `min(items.length, source.length)` are filtered by their per-index schema; elements beyond the
+   * tuple are covered by `additionalItems` when present, or the target is truncated when absent.
+   * For list schemas (`items` is a single schema) every source element is filtered and `target.length`
+   * is set to `source.length`.
    *
    * @param childSchema - The array schema describing `items` and optionally `additionalItems`
    * @param source - The source array to read elements from
-   * @param target - The accumulator array to push filtered elements into
-   * @returns - `target` after all applicable source elements have been pushed
+   * @param target - The accumulator array to write filtered elements into
+   * @returns - `target` after all applicable source elements have been written
    */
   function handleArray(childSchema: S, source: unknown[], target: unknown[]): unknown[] {
     const { items, additionalItems } = childSchema;
-    if (items !== undefined) {
-      if (Array.isArray(items)) {
-        for (let i = 0; i < items.length; i += 1) {
-          target.push(omit(items[i] as S | boolean, source[i]));
-        }
-      } else {
-        for (let i = 0; i < source.length; i += 1) {
-          target.push(omit(items as S | boolean, source[i]));
-        }
+    if (Array.isArray(items)) {
+      const tupleLength = Math.min(items.length, source.length);
+      let i = 0;
+      for (; i < tupleLength; i += 1) {
+        // oxlint-disable-next-line no-param-reassign
+        target[i] = omit(items[i] as S | boolean, source[i]);
+      }
+      // Without additionalItems, truncate at the tuple boundary to drop any pre-existing items.
+      if (!additionalItems) {
+        // oxlint-disable-next-line no-param-reassign
+        target.length = tupleLength;
+        return target;
+      }
+      for (; i < source.length; i += 1) {
+        // oxlint-disable-next-line no-param-reassign
+        target[i] = omit(additionalItems as S | boolean, source[i]);
+      }
+    } else {
+      for (let i = 0; i < source.length; i += 1) {
+        // oxlint-disable-next-line no-param-reassign
+        target[i] = omit(items as S | boolean, source[i]);
       }
     }
-    // additionalItems covers tuple items beyond the items array length.
-    if (additionalItems) {
-      for (let i = target.length; i < source.length; i += 1) {
-        target.push(omit(additionalItems as S | boolean, source[i]));
-      }
-    }
+    // Truncate any pre-existing items beyond the source length.
+    // oxlint-disable-next-line no-param-reassign
+    target.length = source.length;
     return target;
   }
 
@@ -330,7 +335,7 @@ export default function omitExtraData<
       ? validator.isValid(condition as S, source as T, rootSchema)
       : condition;
     const branch = isThenBranch ? then : otherwise;
-    return branch === undefined ? target : omit(branch as S | boolean, source, target);
+    return branch === undefined ? target : omit(branch as S | boolean, source, target, false);
   }
 
   /** Applies the best-matching `oneOf` option to `source`, merging the result into `target`. When
@@ -368,7 +373,7 @@ export default function omitExtraData<
     const resolved: S = isObject(winning)
       ? resolveAllReferences<S>(winning, rootSchema, [])
       : scoringOptions[bestIndex];
-    return omit(resolved, source, target);
+    return omit(resolved, source, target, false);
   }
 
   /** Applies `anyOf` branches from `schema` to `source`, merging results into `target`. When `anyOf`
@@ -394,7 +399,7 @@ export default function omitExtraData<
     ) {
       let result = target;
       for (const branch of anyOf as (S | boolean)[]) {
-        result = omit(branch, source, result);
+        result = omit(branch, source, result, false);
       }
       return result;
     }
@@ -420,7 +425,7 @@ export default function omitExtraData<
     for (const [key, deps] of Object.entries(dependencies)) {
       // Skip property dependencies (string arrays); only process schema dependencies.
       if (key in source && !Array.isArray(deps)) {
-        result = omit(deps as S | boolean, source, result);
+        result = omit(deps as S | boolean, source, result, false);
       }
     }
     return result;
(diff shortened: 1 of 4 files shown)


packages/utils/test/schema/omitExtraDataTest.ts
```diff
@@ -1189,6 +1190,55 @@ export default function omitExtraDataTest(testValidator: TestValidatorType) {
...
+          type: 'object',
+          additionalProperties: false,
+          if: { type: 'object' } as any,
+          then: { properties: { extra: { type: 'string' } } },
+        };
+        const formData = { extra: 'keep' };
+        testValidator.setReturnValues({ isValid: [true] });
+        const result = omitExtraData(testValidator, schema, schema, formData);
+        expect(result).toEqual({ extra: 'keep' });
+      });
```
x0k: Why was this behavior chosen instead of "Prune additional properties"?

The values obtained do not pass validation: [case1](https://x0k.dev/svelte-jsonschema-form/playground3/?page=v#N4IgzgxgFgpgtgQxALhMAOgOwATYC4CeADjMtgOQD2ARgFYwR7kA0WuRATpSR3gJYwwZYPmKlsIwiTLkweDn0wBzctgC+61jmxEEePDA6YAClx79BwigD0iAfXJWp42fMUr1mttgQATX3z8lJgIADam3IYWQtgAZmFgMFq4fLFWnJG8AjGSYlYQwXIyAIKqGuXJ+LCY6WZR2VYwAB7yCE55FHIKymWealpqIMwgikQArngoaKLSFKXMOnaxAqG+MgDWMDBELNjNrRtbO+pDIJQT45OoANreGNq46CC6eFBPZNcAupWPIHCCYAQShg72wTzgYzk2AAcgB5AAq2CgCAAbjAfP5AnxgmEdHUsoInt41FhPqcUWE+L49JQOChQCF-lMELQUQAOU6+DgIWJ4ABMAAZBSh4qFEsNOAxKHAiHxQjBfCKEjA1GogA), [case2](https://x0k.dev/svelte-jsonschema-form/playground3/?page=v#N4IgzgxgFgpgtgQxALhMAOgOwATYC4CeADjMtgOQD2ARgFYwR7kA0WuCAJhwJZ7eWYEAGwAKAJ0okxfGGDIAzYWBisc2bvLLB8xUhRr1G5bAF9VuPLExbsRCVJlzs2mAA88YhDcIky5MB7cmADmxiampqomIMwgQUQArngoaNhuHl4UANYwMERhMSCUSYnJqADabM5VuOggRAiWdWTlALrmuNh1cLJgCMEwzV0gcAkB2AByAPIAKthQCABuMNicPHwCwrb2MNLcsnVVJlithYvC3ByNlGIooII9KQi0iwAchRye8ngATAAM-xQiiEylidgYlDgRG4QhgHCBShgJhMQA)

heath-freenome: Good question — there are two different versions of "restoring the adaptation" worth distinguishing:

**The original (buggy) adaptation** — what this PR removes — built the allowed-key set from the parent schema's `properties` only, so it wiped out keys written by winning `oneOf`/`anyOf` and `if`/`then` branches. That's the root cause of #5194, so restoring it exactly makes things worse.

**A corrected SJSF-style adaptation** that tracks *all* keys written by any branch during processing (parent `properties` + `patternProperties` + branch-granted keys) and only prunes what isn't covered by any of them: for your two playground examples the runtime result would be **identical to this PR** — `{ type: 'A', p_field: 'keep', extra: 'keep' }` — because `extra` is branch-granted. The data would still fail strict AJV validation for the same reason: in JSON Schema draft-07, `additionalProperties: false` only sees `properties` and `patternProperties` at the same schema level; it does not retroactively count `then`-branch properties as "known."

The choice to preserve branch-granted keys is intentional: if `type: 'A'` reveals an `extra` field and the user fills it in, stripping it on submit would be a surprising regression even if a strict validator would flag it. The spec-correct solution for cross-keyword property awareness is `unevaluatedProperties` (draft 2019-09+), not `additionalProperties: false`.

That said, if you think `omitExtraData` should enforce stricter pruning here and you have a concrete approach that handles all the branch-composition edge cases, I'm open to it — it's a real trade-off.

x0k: >  for your two playground examples the runtime result would be identical to this PR 

[omit1](https://x0k.dev/svelte-jsonschema-form/playground3/?page=o#N4IgzgxgFgpgtgQxALhMAOgOwATYC4CeADjMtgOQD2ARgFYwR7kA0WuRATpSR3gJYwwZYPmKlsIwiTLkweDn0wBzctgC+61jmxEEePDA6YAClx79BwigD0iAfXJWp42fMUr1mttgQATX3z8lJgIADam3IYWQtgAZmFgMFq4fLFWnJG8AjGSYlYQwXIyAIKqGuXJ+LCY6WZR2VYwAB7yCE55FHIKymWealpqIMwgikQArngoaKLSFKXMOnaxAqG+MgDWMDBELNjNrRtbO+pDIJQT45OoGNroIM53ZHfFd5V39sswq4-Yd5vbdywgzUQA), [omit2](https://x0k.dev/svelte-jsonschema-form/playground3/?page=o#N4IgzgxgFgpgtgQxALhMAOgOwATYC4CeADjMtgOQD2ARgFYwR7kA0WuCAJhwJZ7eWYEAGwAKAJ0okxfGGDIAzYWBisc2bvLLB8xUhRr1G5bAF9VuPLExbsRCVJlzs2mAA88YhDcIky5MB7cmADmxiampqomIMwgQUQArngoaNhuHl4UANYwMERhMSCUSYnJqMDRJkA)

This is handled by [this code](https://github.com/x0k/svelte-jsonschema-form/blob/3ee66f60d4f9ea0d08b93857966bdcc53b973648/packages/form/src/omit-extra-data.ts#L311-L322)

heath-freenome: I integrated that fix. And then fixed a bug.

x0k: The validator doesn't like this  [demo](https://x0k.dev/svelte-jsonschema-form/playground3/?page=v#N4IgzgxgFgpgtgQxALhMAOgOwATYC4CeADjMtgOQD2ARgFYwR7kA0WuRATpSR3gJYwwZYNkwI4pbCMIky5MHg59MAc3LYAvptY5sCACb6+-SmIA2ABS49+gsgDMEZsDB259MEpg+YIAoVJsuKLikhi6wdic3DC8-sJ6KmH4xJLkmACucNSx6loabsEFQcWYGiDMIMpEGXgoaCEScgCCZnwQMCyJkgDMAAyaFSCUtTV1qADaQeGR6CBECHhQc2QTALqFuHMSYGAISSvY2xkK2AByAPIAKthQCABuMHqGxnymTlHWsbZgcyVYayG9ycfH0i0oHBQoDEEnqCFo9wAHEN9BwEPY8AAmPrYlCOZyueYcBiUOBEPhmGD6PFOFwaDRAA).
I think this might be one of the reasons to switch to the new version of the JSON schema with `unevaluatedProperties`

heath-freenome: @x0k which validator? There are 3 now

x0k: All of them 
image

packages/utils/test/schema/omitExtraDataTest.ts
```diff
@@ -1189,6 +1283,56 @@ export default function omitExtraDataTest(testValidator: TestValidatorType) {
...
+        // is pruned. 'p_field' matches '^p_' and is retained.
+        const schema: RJSFSchema = {
+          type: 'object',
+          properties: { type: { type: 'string' } },
+          patternProperties: { '^p_': { type: 'string' } },
+          additionalProperties: false,
+          if: { properties: { type: { const: 'A' } } },
+          then: { properties: { extra: { type: 'string' } } },
+        };
+        const formData = { type: 'A', p_field: 'keep', extra: 'keep' };
```
nickgros: Change this value to 'drop' to match the intent of the test

```suggestion
        const formData = { type: 'A', p_field: 'keep', extra: 'drop' };
```

packages/utils/test/schema/omitExtraDataTest.ts
```diff
@@ -1236,12 +1391,13 @@ export default function omitExtraDataTest(testValidator: TestValidatorType) {
...
-        // anyOf sees [] as empty → applies all branches → returns [] as target.
-        // The outer type:'array' branch then runs handleArray with that existing [] as target,
-        // hitting the Array.isArray(target) ? target : [] true-branch.
+        // anyOf sees [] as empty → applies all branches.
+        // The branch has type:'array' so omit() returns [] as the filtered value.
+        // The outer type:'array' block then runs handleArray with that existing [] as target,
+        // hitting the Array.isArray(filtered) ? filtered : [] true-branch (filtered is an array).
         const schema: RJSFSchema = {
           type: 'array',
-          anyOf: [{ items: { type: 'string' } }, { items: { type: 'number' } }],
```
nickgros, line 1254: Thanks for fixing this test to match the description. Do we still need to have a test that matches the original test logic (or do we have one I am not seeing and this was a duplicate)?

packages/utils/src/schema/omitExtraData.ts
```diff
@@ -424,26 +429,30 @@ export default function omitExtraData<
...
    * @param schemaDef - The schema (or boolean shorthand) to filter `source` against
    * @param source - The raw form data value to filter
    * @param [target] - An optional accumulator carrying results from prior oneOf/anyOf processing
+   * @param [materializeSource=true] - When false a permissive schema (`true`/`{}`) returns `target`
+   *   rather than `source`, preventing compositional branches from aliasing the accumulator to `source`
    * @returns - The filtered value, or `undefined` when the schema rejects the value
    */
-  function omit(schemaDef: S | boolean, source: unknown, target?: unknown): unknown {
+  function omit(schemaDef: S | boolean, source: unknown, target?: unknown, materializeSource = true): unknown {
     if (source === undefined || schemaDef === false) {
```
nickgros: What do you think about `useSourceAsFallback` instead of `materializeSource`?

nickgros (approved):
Need a clean build, but generally LGTM with some minor requests. Please let me know if I should re-review once @x0k 's feedback is addressed

heath-freenome:
@x0k Any thoughts on the current state of this PR? More precisely do you have an object to it getting merged?

x0k:
I still don't understand why the `preserves keys added by a dependency sub-schema even when parent has additionalProperties:false` case was added if its result doesn't pass validation [ajv](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJuYW1lIjoiQWxpY2UiLCJhZ2UiOjMwfSwic2NoZW1hIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7Im5hbWUiOnsidHlwZSI6InN0cmluZyJ9fSwiYWRkaXRpb25hbFByb3BlcnRpZXMiOmZhbHNlLCJkZXBlbmRlbmNpZXMiOnsibmFtZSI6eyJwcm9wZXJ0aWVzIjp7ImFnZSI6eyJ0eXBlIjoibnVtYmVyIn19fX19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJjb25zdEFzRGVmYXVsdHMiOiJhbHdheXMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQiLCJuZXN0ZWREZWZhdWx0c1ByZWNlZGVuY2UiOiJkZXNjZW5kYW50V2lucyJ9LCJ1c2VGYWxsYmFja0ZpZWxkIjpmYWxzZSwidXNlRmFsbGJhY2tVaUZvclVuc3VwcG9ydGVkVHlwZSI6ZmFsc2UsImxpdmVWYWxpZGF0ZSI6ZmFsc2UsImlkUHJlZml4Ijoicm9vdCIsImlkU2VwYXJhdG9yIjoiXyJ9LCJ2YWxpZGF0b3IiOiJBSlY4Iiwic2FtcGxlTmFtZSI6IkJsYW5rIn0=), [ata](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJuYW1lIjoiQWxpY2UiLCJhZ2UiOjMwfSwic2NoZW1hIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7Im5hbWUiOnsidHlwZSI6InN0cmluZyJ9fSwiYWRkaXRpb25hbFByb3BlcnRpZXMiOmZhbHNlLCJkZXBlbmRlbmNpZXMiOnsibmFtZSI6eyJwcm9wZXJ0aWVzIjp7ImFnZSI6eyJ0eXBlIjoibnVtYmVyIn19fX19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJjb25zdEFzRGVmYXVsdHMiOiJhbHdheXMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQiLCJuZXN0ZWREZWZhdWx0c1ByZWNlZGVuY2UiOiJkZXNjZW5kYW50V2lucyJ9LCJ1c2VGYWxsYmFja0ZpZWxkIjpmYWxzZSwidXNlRmFsbGJhY2tVaUZvclVuc3VwcG9ydGVkVHlwZSI6ZmFsc2UsImxpdmVWYWxpZGF0ZSI6ZmFsc2UsImlkUHJlZml4Ijoicm9vdCIsImlkU2VwYXJhdG9yIjoiXyJ9LCJ2YWxpZGF0b3IiOiJBVEEiLCJzYW1wbGVOYW1lIjoiQmxhbmsifQ==), [cfworker](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJuYW1lIjoiQWxpY2UiLCJhZ2UiOjMwfSwic2NoZW1hIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7Im5hbWUiOnsidHlwZSI6InN0cmluZyJ9fSwiYWRkaXRpb25hbFByb3BlcnRpZXMiOmZhbHNlLCJkZXBlbmRlbmNpZXMiOnsibmFtZSI6eyJwcm9wZXJ0aWVzIjp7ImFnZSI6eyJ0eXBlIjoibnVtYmVyIn19fX19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJza2lwRGVmYXVsdHMiLCJjb25zdEFzRGVmYXVsdHMiOiJhbHdheXMiLCJlbXB0eU9iamVjdEZpZWxkcyI6InBvcHVsYXRlQWxsRGVmYXVsdHMiLCJtZXJnZURlZmF1bHRzSW50b0Zvcm1EYXRhIjoidXNlRm9ybURhdGFJZlByZXNlbnQiLCJuZXN0ZWREZWZhdWx0c1ByZWNlZGVuY2UiOiJkZXNjZW5kYW50V2lucyJ9LCJ1c2VGYWxsYmFja0ZpZWxkIjpmYWxzZSwidXNlRmFsbGJhY2tVaUZvclVuc3VwcG9ydGVkVHlwZSI6ZmFsc2UsImxpdmVWYWxpZGF0ZSI6ZmFsc2UsImlkUHJlZml4Ijoicm9vdCIsImlkU2VwYXJhdG9yIjoiXyJ9LCJ2YWxpZGF0b3IiOiJDRldvcmtlciAoRHJhZnQgNykiLCJzYW1wbGVOYW1lIjoiQmxhbmsifQ==) (click raw validation).

heath-freenome:
@x0k You're technically correct about the JSON Schema spec: `additionalProperties` is evaluated against `properties` and `patternProperties` at the **same schema level only**. `dependencies` subschemas do not merge their `properties` into the parent schema's allowed-property set — they're evaluated as a separate overlay. So `{ name: "Alice", age: 30 }` is strictly invalid against that schema regardless of what `name` holds.

But `omitExtraData` is not a validator. Its contract is: *strip data for fields RJSF did not render*. When `name` is filled in, RJSF activates the `dependencies` sub-schema and **renders an `age` input**. The user types into it. If `omitExtraData` then deletes `age`, the user has lost data they explicitly entered into a visible field — a fundamental UX contract violation.

The PR's pipeline makes the ordering explicit:

```
source → handleObject → handleConditions → [post-prune for additionalProperties:false] → handleDependencies → result
```

The post-prune step fires **after** conditional branches (`if/then`, `oneOf`, `anyOf`) but **before** `handleDependencies`. Keys that only survive because of a conditional branch *are* pruned when `additionalProperties: false` is set. Keys added by `dependencies` sub-schemas are not, because those sub-schemas directly cause RJSF to render new fields.

The schema in the test is an anti-pattern: combining `additionalProperties: false` with `dependencies` that introduce new properties creates a schema where the only valid data is data the rendered form can't produce. RJSF has to pick a side — strip rendered-field data (bad UX) or preserve it (bad strict-spec compliance). For a form library, preserving rendered-field data is the right call.

If strict schema compliance is the goal, the fix is in the schema: move `age` into the top-level `properties` (with the dependency sub-schema adding only constraints), or replace `dependencies` with `if/then`. With `if/then`, the PR's post-prune step *does* delete `age` when `additionalProperties: false` is set, so the output would actually validate.

x0k:
> Its contract is: strip data for fields RJSF did not render

This may be one source of the misunderstanding, since I had a different intent in mind for this function:

Remove data that is not declared by the JSON Schema (with no renderer-specific or other logic involved - just a clear standard to follow).

> a fundamental UX contract violation.

However, this can result in a form that cannot be submitted unless these values are removed manually, or may be impossible to submit at all if those fields are declared as required.

That doesn't seem like good UX either. Instead of choosing between two bad options, we could:

* Add a dev-time check for schemas that are incompatible with RJSF and provide suggestions for how to change the schema.
* Explicitly support the workarounds used to achieve the equivalent result.

```js
{
  type: 'object',
  properties: {
    name: { type: 'string' },
    age: {}
  },
  additionalProperties: false,
  dependencies: {
    name: {
      properties: { age: { type: 'number' } }
    }
  }
}
```

```js
{
  type: 'object',
  oneOf: [
    {
      properties: { name: { type: 'string' } },
      additionalProperties: false
    },
    {
      properties: {
        name: { type: 'string' },
        age: { type: 'number' }
      },
      required: ['name'],
      additionalProperties: false
    }
  ]
}
```

Issue #5198, opened 2026-08-18
Defaults inside `dependencies` subschemas are not populated by `getDefaultFormState`


MatinF:
### Prerequisites

- [x] I have searched the existing issues
- [x] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [x] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

antd

### Version

6.7.1 / 6.8.0

### Current Behavior

We are trying to upgrade our rjsf from a very old release to the latest, but we seem to encounter various bugs this introduces including the below:

The bug we observe consists of two related cases where a `default` that lives inside a `dependencies` subschema
never reaches the form data. Both worked in rjsf 5. They share a cause in
`computeDefaults()` in `packages/utils/src/schema/getDefaultFormState.ts`, so
they are filed together.

The visible symptom in both cases is a field the user never touched showing a
validation error, and with a `oneOf` style dependency the error is misleading:
because neither branch matches, ajv reports every branch's failure, so a missing
`p` also surfaces as `.t must be equal to one of the allowed values` pointing at
a field whose value is correct.

---

## Case 1: no form data, so the dependency is never resolved

`computeDefaults()` has a `DEPENDENCIES_KEY` branch, but it resolves the
dependency against data that is already present. Called with no form data it has
nothing to resolve against, so defaults inside the dependency are dropped.

This is what `getNewFormDataRow()` in `@rjsf/core` does when a row is added to an
array:

```ts
return schemaUtils.getDefaultFormState(itemSchema);
```

so a brand new, untouched array row is incomplete and fails validation as soon as
it renders.

### Reproduction

```js
import { createSchemaUtils } from '@rjsf/utils';
import validator from '@rjsf/validator-ajv8';

const schema = {
  type: 'object',
  required: ['t'],
  properties: { t: { type: 'integer', default: 0 } },
  dependencies: {
    t: {
      oneOf: [
        { properties: { t: { enum: [0] }, p: { type: 'integer', default: 5 } }, required: ['p'] },
        { properties: { t: { enum: [1] } } },
      ],
    },
  },
};

[...]

PR #5202, opened 2026-08-19
fix(utils): deep merge an array that sits under an object key

luantaraschi:
Fixes case 2 of #5198.

`mergeDefaultsWithFormData` has a shortcut for the case where the default under a key is an object: if none of that object's values is itself an object, it shallow spreads `{ ...default, ...formData }` and skips the recursion. `isObject` returns false for arrays, so an array of objects does not count as nested, and the shortcut fires. The form data array then replaces the merged default array whole.

Measured on the schema from the issue, with the `getDefaultFormState` arguments:

```js
merge({ rows: [{ t: 0, p: 5 }] },            { rows: [{ t: 0 }] })            // { rows: [{ t: 0, p: 5 }] }
merge({ f: { rows: [{ t: 0, p: 5 }] } },     { f: { rows: [{ t: 0 }] } })     // { f: { rows: [{ t: 0 }] } }
```

The first call never reaches the shortcut because the value under `rows` is an array, so `keyDefaultIsObject` is false. The second one goes through `f`, whose default is an object holding only an array, so it takes the spread and `p` is gone. That is why the reported schema works with the array one level below the root and fails two levels down.

The condition now counts arrays as needing the recursive path, which is where the array merge rules described in the function's own doc comment live.

### What this does not fix

Case 1 of the issue, a `dependencies` default with no form data at all, has a different cause and is untouched here. `getDefaultFormState` resolves the schema before computing defaults:

```js
retrieveSchema(validator, child, undefined, undefined)
// { type: 'object', required: ['t'], properties: { t: { type: 'integer', default: 0 } } }
```

With no form data the dependency is not triggered, so `dependencies` is gone from the schema that reaches `computeDefaults`, and its `DEPENDENCIES_KEY` branch never runs. Calling `computeDefaults` on the raw schema returns `{ t: 0, p: 5 }`, so that branch itself is fine. Fixing it means changing when the schema is resolved relative to when defaults are computed, which is a wider change than this one and I would rather not bundle it. Happy to look at it separately if you want it.

### Tests

- `mergeDefaultsWithFormData.test.ts`, unit test for an array under an object key, in the `overrideFormDataWithDefaults` block.
- `getDefaultFormStateTest.ts`, end to end test next to the existing `nested dependencies in arrays` one, with the array one object deeper.

Both fail on `main` and pass with the change.

Suites run locally: utils 1444, validator-ajv8 2681, validator-ata 1277, validator-cfworker 613, core 2235, all passing. `oxfmt --check`, `oxlint` and `tsc --noEmit` are clean for `utils`; the oxlint warnings it prints are pre-existing and in files this PR does not touch.

One thing I could not run: the husky pre-commit hook. It goes through `nx run-many` into `pnpm`, which tries to create a symlink in its own store and fails with `os error 1314` on this Windows box. I ran what it would have run, `oxlint --fix` and `oxfmt` over the three changed files, and both are no-ops.

CHANGELOG entry is in the second commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.8.1
+
+## @rjsf/utils
+
+- Fixed `mergeDefaultsWithFormData()` to take the recursive path when the default under an object key holds an array, so a default that lives inside a `dependencies` subschema is no longer dropped when the owning array sits more than one object below the root, fixing part of [#5198](https://github.com/rjsf-team/react-jsonschema-form/issues/5198) ([#5202](https://github.com/rjsf-team/react-jsonschema-form/pull/5202))
+
 # 6.8.0
 
 ## @rjsf/core
@@ -25,7 +31,6 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Fixed an issue in `sortedJSONStringify` with regards to handling arrays containing `null` and `undefined` values.
-- Fixed `mergeDefaultsWithFormData()` to take the recursive path when the default under an object key holds an array, so a default that lives inside a `dependencies` subschema is no longer dropped when the owning array sits more than one object below the root, fixing part of [#5198](https://github.com/rjsf-team/react-jsonschema-form/issues/5198) ([#5202](https://github.com/rjsf-team/react-jsonschema-form/pull/5202))
 
 ## @rjsf/validator-ata
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Fixed an issue in `sortedJSONStringify` with regards to handling arrays containing `null` and `undefined` values.
+- Fixed `mergeDefaultsWithFormData()` to take the recursive path when the default under an object key holds an array, so a default that lives inside a `dependencies` subschema is no longer dropped when the owning array sits more than one object below the root, fixing part of [#5198](https://github.com/rjsf-team/react-jsonschema-form/issues/5198) ([#5202](https://github.com/rjsf-team/react-jsonschema-form/pull/5202))
 
 ## @rjsf/validator-ata
 
--- a/packages/utils/src/mergeDefaultsWithFormData.ts
+++ b/packages/utils/src/mergeDefaultsWithFormData.ts
@@ -68,9 +68,9 @@ export default function mergeDefaultsWithFormData<T = any>(
       const keyExistsInDefaults = isObject(defaults) && key in (defaults as GenericObjectType);
       const keyExistsInFormData = key in (formData as GenericObjectType);
       const keyDefault = get(defaults, key) ?? {};
-      // Anything below this key that is an object or an array has merge rules of its
-      // own, and the shallow spread below bypasses them, so those cases take the
-      // recursive path instead.
+      // If any direct child of this key's default is itself an object or array,
+      // the shallow spread below would bypass its merge rules. Route those cases
+      // to the recursive path; the recursive call repeats this check one level deeper.
       const defaultValueNeedsDeepMerge =
         keyExistsInDefaults &&
         isObject(keyDefault) &&
--- a/packages/utils/src/mergeDefaultsWithFormData.ts
+++ b/packages/utils/src/mergeDefaultsWithFormData.ts
@@ -68,13 +68,18 @@ export default function mergeDefaultsWithFormData<T = any>(
       const keyExistsInDefaults = isObject(defaults) && key in (defaults as GenericObjectType);
       const keyExistsInFormData = key in (formData as GenericObjectType);
       const keyDefault = get(defaults, key) ?? {};
-      const defaultValueIsNestedObject =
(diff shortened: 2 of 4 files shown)


packages/utils/src/mergeDefaultsWithFormData.ts
```diff
@@ -68,13 +68,18 @@ export default function mergeDefaultsWithFormData<T = any>(
...
       const keyDefault = get(defaults, key) ?? {};
-      const defaultValueIsNestedObject =
-        keyExistsInDefaults && isObject(keyDefault) && Object.values(keyDefault).some((v) => isObject(v));
+      // Anything below this key that is an object or an array has merge rules of its
+      // own, and the shallow spread below bypasses them, so those cases take the
+      // recursive path instead.
+      const defaultValueNeedsDeepMerge =
+        keyExistsInDefaults &&
+        isObject(keyDefault) &&
+        Object.values(keyDefault).some((v) => isObject(v) || Array.isArray(v));
```
heath-freenome, line 77: **Undocumented side-effect when `mergeExtraArrayDefaults=true` and the defaults array is longer than the formData array**

The old shallow-spread path never reached the array-merge logic, so `mergeExtraArrayDefaults` was silently ignored for object keys whose defaults contained an array. With the new recursive path, that flag is now honoured — which is the *correct* behaviour, but it is a behaviour change for any caller that previously relied on formData's array winning unconditionally.

Concrete example: `defaults = { config: { tags: ['default-a', 'default-b'] } }`, `formData = { config: { tags: ['user-a'] } }`, called with `mergeExtraArrayDefaults=true`.
- **Before:** shallow-spread fires → `{ config: { tags: ['user-a'] } }` (extra default ignored)
- **After:** recursive path → `{ config: { tags: ['user-a', 'default-b'] } }` (extra default appended)

`getDefaultFormState` calls `mergeDefaultsWithFormData` with `mergeExtraArrayDefaults=true`, so any schema whose property default is a flat object containing an array *longer* than what the user submitted will now append those extra items on every render. Both new tests use equal-length arrays, so this case has no coverage. Worth adding a test (or a note in the PR description) to confirm the new behaviour is intentional.

packages/utils/src/mergeDefaultsWithFormData.ts
```diff
@@ -68,13 +68,18 @@ export default function mergeDefaultsWithFormData<T = any>(
       const keyExistsInDefaults = isObject(defaults) && key in (defaults as GenericObjectType);
       const keyExistsInFormData = key in (formData as GenericObjectType);
       const keyDefault = get(defaults, key) ?? {};
-      const defaultValueIsNestedObject =
-        keyExistsInDefaults && isObject(keyDefault) && Object.values(keyDefault).some((v) => isObject(v));
+      // Anything below this key that is an object or an array has merge rules of its
+      // own, and the shallow spread below bypasses them, so those cases take the
+      // recursive path instead.
```
heath-freenome: **Comment overstates the depth of the inspection**

`Object.values(keyDefault).some(...)` only looks at the *direct children* of `keyDefault` — one level deep. The statement "anything below this key" implies unlimited depth, which isn't what the code does. It holds end-to-end only because each recursive invocation re-runs the same one-level check at the next level down, but that dependency is invisible from the comment alone.

Suggested rewording:
```
// If any direct child of this key's default is itself an object or array,
// the shallow spread below would bypass its merge rules. Route those cases
// to the recursive path; the recursive call repeats this check one level deeper.
```

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Fixed an issue in `sortedJSONStringify` with regards to handling arrays containing `null` and `undefined` values.
+- Fixed `mergeDefaultsWithFormData()` to take the recursive path when the default under an object key holds an array, so a default that lives inside a `dependencies` subschema is no longer dropped when the owning array sits more than one object below the root, fixing part of [#5198](https://github.com/rjsf-team/react-jsonschema-form/issues/5198) ([#5202](https://github.com/rjsf-team/react-jsonschema-form/pull/5202))
```
heath-freenome: We've released 6.8.0. Move this into 6.8.1

packages/utils/src/mergeDefaultsWithFormData.ts
luantaraschi: You are right, it only looks at the direct children. I took your wording.

CHANGELOG.md
luantaraschi: Moved to 6.8.1.

packages/utils/src/mergeDefaultsWithFormData.ts
luantaraschi: Ran it both ways and it lands exactly where you said. With `mergeExtraArrayDefaults=true`, `{ config: { tags: ['default-a', 'default-b'] } }` over `{ config: { tags: ['user-a'] } }` gives `['user-a']` on main and `['user-a', 'default-b']` here.

What settled it for me is that the same two arrays at the root already appended before this PR: `mergeDefaultsWithFormData({ tags: ['default-a', 'default-b'] }, { tags: ['user-a'] }, true)` returns `['user-a', 'default-b']` on main as well. So the flag was honoured when the array was the value and dropped when it sat one object below, and the doc comment on the function says the extras get appended. I read that as the nested case catching up rather than a new rule, but say the word if you would rather have it gated.

Added a test carrying both halves, the flag on and the flag off. The first assertion fails on main.

heath-freenome (approved):

Issue #5201, opened 2026-08-19
Inline field errors stop rendering after adding item to nested array


dmitry-mamay:
### Prerequisites

- [x] I have searched the existing issues
- [x] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [x] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

6.6.2

### Current Behavior

[...]

PR #5204, opened 2026-08-20
refactor: replace path-based lodash functions with native path utilities in utils and core

jimmycallin:

Follow-up to #5189 (native type guards): this removes all **path-based** lodash usage (`get`, `set`, `setWith`, `has`, `unset`, `toPath`, `pick`) from `@rjsf/utils` and `@rjsf/core`, replacing it with a small set of native path utilities. Removing the remaining non-path lodash functions (`isEmpty`, `isObject`, etc.) is a separate effort, with dropping the `lodash` dependency from these packages entirely as the eventual goal.

## What was done

- **New `@rjsf/utils` exports**: `getByPath`, `setByPath`, `hasByPath`, `unsetByPath`, and `toPath` (~40 lines total). The contract is deliberately simpler than lodash's:
  - a bare string path is always **one literal key** — `'a.b'` means the key `'a.b'`, never a deep path;
  - an array is a list of segments;
  - `toPath()` is the explicit parser for dotted strings, supporting exactly the grammar RJSF produces (dotted segments plus bracketed non-negative integer indexes such as `'.level1.level2[2]'`). Quoted keys, negative/decimal indexes, and lodash's empty-segment handling are intentionally not supported; keys containing dots use the segment-list form.
- **Every call site audited**: sites that only ever accessed a single constant key were simplified to plain property access / optional chaining / `in` checks; the genuine dotted-string boundaries (`ErrorSchemaBuilder`, `LayoutGridField` layout paths, `getUsedFormData`, validation-error property strings) now parse explicitly with `toPath()`.
- **Typing strengthened instead of cast**: `isObject()` now narrows to `GenericObjectType`, letting guarded form-data access typecheck without assertions; several previously-needed casts and a `@ts-expect-error` were removed.
- **Behavioral fixes over lodash semantics**:
  - property names containing dots (e.g. a user-typed additional property `"a.b"`) are no longer silently expanded into nested objects;
  - `setByPath` refuses `__proto__`/`constructor`/`prototype` segments (matching lodash ≥ 4.17.19's prototype-pollution hardening), and `getByPath` only resolves those segments when they are genuine own data keys;
  - `unsetByPath` returns `false` instead of throwing on non-configurable properties in strict mode;
  - boolean property subschemas (`properties: { foo: false }`) are handled safely in `sanitizeDataForNewSchema` and `findFieldInSchema`.
- `@rjsf/utils` retains 100% test coverage, including tests for all of the behaviors above.

## Impact

**Bundle size** — minimal app doing `import Form from '@rjsf/core'` (core + utils bundled, react/react-dom external), minified + gzipped (level 9), every milestone rebuilt from source in a clean worktree and bundled identically (esbuild 0.28.1 / rolldown 1.1.5):

| milestone | esbuild | Δ | rolldown | Δ |
|---|---|---|---|---|
| v6.8.0 (before the removal effort) | 132.3 kB | — | 121.6 kB | — |
| after #5191 (`cloneDeep`/`omit`) | 130.9 kB | −1.4 kB | 120.3 kB | −1.3 kB |
| after #5190 (deep-equality chain) | 126.7 kB | −4.2 kB | 116.8 kB | −3.5 kB |
| after #5189 (type guards) | 126.6 kB | −0.1 kB | 116.7 kB | −0.1 kB |
| after #5206 (`sideEffects: false`) = current `main` | 86.7 kB | −39.9 kB | 80.5 kB | −36.2 kB |
| **after this PR (path functions → native `pathUtils`)** | **85.4 kB** | **−1.3 kB** | **79.3 kB** | **−1.2 kB** |

Cumulative across the effort including this PR: **−46.9 kB gzipped (esbuild) / −42.3 kB (rolldown)** off the whole Form bundle — the bulk from #5206 letting bundlers tree-shake the unused AJV chain, which this table now shows as its own row instead of a footnote.

This PR also converts the validator packages off `lodash/get` (measured the same way with an entry importing the validator):

| bundle | main | this PR | Δ |
|---|---|---|---|
| `@rjsf/validator-ajv8` (esbuild / rolldown) | 66.9 / 60.1 kB | 64.7 / 58.2 kB | −2.2 / −1.9 kB |
| `@rjsf/validator-cfworker` (esbuild / rolldown) | 38.2 / 33.6 kB | 32.9 / 29.5 kB | −5.3 / −4.1 kB |

`@rjsf/validator-cfworker` drops its `lodash`/`lodash-es` dependencies entirely; ajv8 and ata keep a single `lodash/isObject` import each, part of the non-path follow-up. And #5189, #5192 and #5193 additionally removed lodash from the `@rjsf/antd`, `@rjsf/semantic-ui` and `@rjsf/shadcn` theme bundles, which isn't counted above.

**Runtime** (Node micro-benchmark, 2M iterations, interleaved warmup, best of 3 rounds) on the call shapes used in the codebase:

| operation | lodash | native | ratio |
|---|---|---|---|
| `get`, single key | 17 ns | 7 ns | 0.41× |
| `get`, missing with default | 19 ns | 7 ns | 0.39× |
| `has`, single key | 17 ns | 6 ns | 0.38× |
| `set`, single key | 21 ns | 7 ns | 0.32× |
| `set`, array path into fresh object (Form onChange shape) | 91 ns | 58 ns | 0.64× |
| `get`, array path (pre-split segments) | 14 ns | 28 ns | 1.95× |
| `has`, array path | 25 ns | 28 ns | 1.12× |
| **`get` via dotted string** (`get(o,'a.b.c')` vs `getByPath(o, toPath('a.b.c'))`) | 47 ns | 86 ns | 1.84× |

Single-key access — the large majority of call sites, which look up one constant key such as `default`, `$ref` or `propertyName` — is **2.4–3× faster** than lodash, because a bare-string key takes a fast path that skips both lodash's per-call key-vs-path detection and any segment-list allocation. `setByPath` on the Form onChange shape is 36% faster.

Two shapes are slower, and both are worth stating plainly:

- **Multi-segment reads (1.95×, +14 ns).** `getByPath` resolves **own** properties only so that it agrees with `hasByPath` (see [this thread](https://github.com/rjsf-team/react-jsonschema-form/pull/5204#discussion_r3827276971)); `Object.hasOwn()` costs ~5 ns per segment. This is the semantics, not the implementation — lodash's own `has`, which performs the same own-property checks, lands at the same ~28 ns. Structural variants (indexed loops, hoisted regex, read-then-verify) were benchmarked and recover ~2 ns at most.
- **Dotted-string reads (1.84×, +39 ns).** lodash memoized its string→segments parse in a 500-entry capped cache, so a repeated path string was parsed once; `toPath()` re-parses on every call (36 ns for one segment, 58 ns for three). A memo cache would close this (measured: 5 ns on a hit), but it means module-level mutable state plus a mutation hazard — `toPath` returns an array, and handing out a shared instance lets one caller corrupt every later one, which is why lodash's own public `toPath` copies. The absolute cost is ~39 ns per call at sites that run once per field per render, i.e. single-digit microseconds for a 50-field grid, so the cache is deliberately left out. Easy to add later if a profile ever justifies it.

Neither regression is close to significant next to a React render.

## Checklist

- [x] I'm updating documentation (JSDoc on all new utilities; CHANGELOG entry under 6.9.0)
- [x] I'm adding or updating code (unit tests added/updated; `pnpm run lint`, full build, and all package test suites pass)


packages/utils/src/pathUtils.ts
heath-freenome, line 1: Can you update the docs to add these new exported functions

packages/utils/src/pathUtils.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @returns - The list of path segments, such as `['level1', 'level2', '2', 'level3']`
+ */
+export function toPath(path: string): string[] {
+  const result: string[] = [];
+  path.replace(rePathSegment, (match: string, index: string) => {
+    result.push(index ?? match);
+    return match;
+  });
+  return result;
+}
```
heath-freenome: `toPath` uses `String.prototype.replace` purely as an iterator — the mutated string it returns is discarded and `return match` exists only to suppress substitution. This misleads readers who expect `.replace` to produce something useful.

A more expressive form using `matchAll`:

```ts
export function toPath(path: string): string[] {
  const result: string[] = [];
  for (const m of path.matchAll(/[^.[\]]+|\[([\d+])\]/g)) {
    result.push(m[1] ?? m[0]);
  }
  return result;
}
```

There is also a latent hazard with the module-level global regex: `.replace` correctly resets `lastIndex` after each call, but any future refactor to an `.exec()`-in-a-loop (a natural cleanup step) would introduce a re-entrancy bug because the regex is shared across calls. Moving the regex literal inside `toPath`, or switching to a non-global pattern with `matchAll`, removes that risk entirely.

packages/utils/src/pathUtils.ts
```diff
@@ -0,0 +1,149 @@
...
+  const segments = normalizePath(path);
+  let current: any = obj;
+  for (const segment of segments) {
+    // Prototype-chain segments only resolve when they are real own data keys, never leaking internals
+    if (current == null || (isUnsafeSegment(segment) && !Object.hasOwn(current, segment))) {
+      return defaultValue as R;
+    }
+    current = current[segment];
+  }
+  return current === undefined ? (defaultValue as R) : current;
```
heath-freenome: `getByPath` and `hasByPath` have an undocumented asymmetry around prototype-chain properties that can cause confusing mismatches at call sites.

`hasByPath` (line 119) calls `Object.hasOwn` at **every** segment, so it returns `false` for any inherited property — e.g. `hasByPath({}, 'toString') === false`.

`getByPath` (this function) only guards with `Object.hasOwn` for segments in `UNSAFE_SEGMENTS`. For all other segments it falls through to `current[segment]`, which *does* traverse the prototype chain — so `getByPath({}, 'toString')` returns `Function.prototype.toString`.

The mismatch means a guard pattern like `if (hasByPath(obj, key)) { return getByPath(obj, key); }` silently misses inherited keys where `hasByPath` says `false` but `getByPath` would return a non-`undefined` value. This is safe today because RJSF form-data objects are always plain objects with no custom prototypes, but the divergence should be called out in JSDoc on both functions so future callers know not to use them interchangeably as a guard/read pair.

packages/utils/src/pathUtils.ts
jimmycallin: Done in a6eb40c — `packages/docs/docs/api-reference/utility-functions.md` now has entries for `getByPath`, `hasByPath`, `setByPath`, `unsetByPath` and `toPath`, each with parameters, return value and an example, plus an `ObjectPath` note in the Types section next to the existing `SchemaFieldPath` one calling out that a bare string is always a single literal key (the deliberate divergence from lodash).

packages/utils/src/pathUtils.ts
jimmycallin: Agreed on the diagnosis — the returned string was discarded and `return match` only existed to suppress substitution. Fixed in a6eb40c, though with a third option rather than `matchAll`, because the whole function collapses to one line:

```ts
export function toPath(path: string): string[] {
  return path.split(/[.[\]]+/).filter(Boolean);
}
```

Dots and brackets are always separators in the grammar we support, so there is nothing to capture and no alternation needed. That also removes `rePathSegment` entirely, which kills the `lastIndex` hazard you flagged at the source rather than working around it — a non-global `split` pattern has no `lastIndex` state for a future refactor to trip over.

I diffed it against the old implementation over the whole supported grammar before swapping: `a.b.c`, `a[0].b`, `.level1.level2[2].level3`, `''`, `.`, `a.0.b`, `a[10].b`, `list[0].d`, `a[x].b`, `ui:options`, `a..b`, `a[0][1]` — identical output on all of them. Added tests for the trailing-dot, `..` and `a[0][1]` cases so the empty-segment behavior is pinned, and moved the grammar notes onto the function's JSDoc now that the regex constant is gone.

Two notes on the suggested snippet, in case it gets reused elsewhere: `\[([\d+])\]` is a character class, so it matches a single digit or a literal `+` — `a[10]` falls through to the other alternative and coincidentally yields the same result, but it should be `(\d+)`. And `matchAll` requires the `g` flag, so the "non-global pattern with `matchAll`" variant would throw a `TypeError`; the global literal in your snippet is fine since being inline means each call gets a fresh regex.

packages/utils/src/pathUtils.ts
jimmycallin: Confirmed — `getByPath({}, 'toString')` resolved `Function.prototype.toString` while `hasByPath({}, 'toString')` was `false`.

Rather than document the divergence, 81abbd7 makes them symmetric, which is the behavior I'd expect a reader to assume: for the plain form data and schemas these navigate, an inherited member is never data. So `getByPath` now requires an own property at every segment, tightening it rather than loosening `hasByPath` to `in` — that direction would have re-opened the prototype read surface. It's also a net simplification, since own-only subsumes the `__proto__`/`constructor`/`prototype` special case and `isUnsafeSegment` is now only needed for writes in `setByPath`.

It fixes a latent bug too: static properties are inherited, so in `getWidget()` a widget declared `class B extends A` resolved A's `MergedWidget` and reused a wrapper bound to the parent component with the parent's `defaultOptions`.

Verified the two call sites that pass something other than a plain object — `globalThis.process.env.NODE_ENV` in `getTestIds` and `MergedWidget` in `getWidget` — are own at every level. All 1485 `@rjsf/utils` tests pass with the 100% coverage gate held, including a new case asserting `hasByPath(obj, key) === (getByPath(obj, key) !== undefined)` across own, inherited and `constructor` keys.

jimmycallin: Revisited this — you were right that symmetric is the more expected shape, and I've made them symmetric in 81abbd7 rather than just documenting the divergence.

The direction matters, and only one of the two is worth doing. Loosening `hasByPath` to use `in` would make `hasByPath({}, 'toString')` true and re-open the prototype read surface, which is the wrong way for form data. So `getByPath` is now the one that changed: every segment must be an own property, exactly as `hasByPath` already required.

```ts
for (const segment of segments) {
  // Only own data keys resolve, so inherited members and prototype internals are never read
  if (current == null || !Object.hasOwn(current, segment)) {
    return defaultValue as R;
  }
  current = current[segment];
}
```

That is a net simplification: own-only subsumes the `__proto__`/`constructor`/`prototype` special case, so `isUnsafeSegment` drops out of `getByPath` entirely and is now needed only for writes in `setByPath`. One rule instead of two.

It also fixes a latent bug in `getWidget()`. Static properties are inherited, so a widget declared as `class B extends A` resolved A's `MergedWidget` through the prototype chain and reused a wrapper bound to the *parent* component with the parent's `defaultOptions`. Own-only makes B build its own wrapper.

I checked the two call sites that pass something other than a plain object, since those are where prototype traversal could have mattered:

- `getByPath(globalThis, ['process', 'env', 'NODE_ENV'])` in `getTestIds` — own at every level (`Object.hasOwn(process.env, 'NODE_ENV')` is `true` when set, `false` when unset, so the `!== 'test'` branch is unchanged either way).
- `getByPath(AWidget, 'MergedWidget')` in `getWidget` — own, since `setByPath` writes it onto that same function object.

All 1485 `@rjsf/utils` tests pass with the 100% coverage gate held, and I added a case asserting `hasByPath(obj, key) === (getByPath(obj, key) !== undefined)` across own, inherited, `Object.create` and `constructor` keys so the pairing stays enforced.

One honest cost: `Object.hasOwn()` runs about 4 ns per segment, so multi-segment reads got slower rather than faster. The benchmark table in the PR description is updated — `get` on a 4-segment array path is now 32 ns against lodash's 21 ns, where it was 19 ns before. I tried to recover it and there is no cheaper correct form: `hasOwnProperty.call` measures 31 ns and read-then-verify 34 ns, versus 17.5 ns for prototype-traversing access. Single-key reads, which are the majority of call sites, are still well ahead (8 ns vs 13 ns). Twelve nanoseconds per deep read is a rounding error next to a React render, and I'd rather pay it than ship two functions that disagree — but flagging it explicitly in case you'd rather keep the faster asymmetric version with the JSDoc warning instead.

heath-freenome:
@jimmycallin there are a couple of current, unmerged PRs that are using `setWith()`. Is that something that you can also implement?

jimmycallin:
@heath-freenome Yes — that case is already covered. Every `setWith()` usage in those PRs is the `Object` customizer (forcing plain objects instead of arrays for numeric segments, e.g. `_setWith(acc, path, undefined, Object)` in #5205), and `setByPath` supports that directly through its fourth parameter: `setByPath(acc, path, undefined, true)`. It's what `ErrorSchemaBuilder` on this branch already uses in place of its previous `setWith` call, and it's documented in the new utility-functions docs with a test pinning the behavior.

I skipped lodash's general customizer-function signature since the `Object` customizer is the only form used anywhere in the codebase or the open PRs.

packages/utils/src/lookupFromFormContext.ts
```diff
@@ -17,10 +15,10 @@ export default function lookupFromFormContext<
   T = any,
   S extends StrictRJSFSchema = RJSFSchema,
   F extends FormContextType = any,
->(regOrFc: Registry<T, S, F> | Registry<T, S, F>['formContext'], toLookup: string, fallback?: unknown) {
+>(regOrFc: Registry<T, S, F> | Registry<T, S, F>['formContext'], toLookup: string, fallback?: any) {
```
heath-freenome: Why switch from `unknown` to `any`?

CHANGELOG.md
```diff
@@ -16,6 +16,17 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.9.0
+
+## @rjsf/utils
+
+- Added new path utilities `getByPath`, `setByPath`, `hasByPath`, `unsetByPath` and `toPath`, and switched all path-based `lodash` usage (`get`, `set`, `setWith`, `has`, `unset`, `toPath`, `pick`) over to them. Unlike their lodash counterparts, these treat a bare string as a single literal key; lodash-style dotted path strings are parsed explicitly via `toPath()`
+- `getByPath()` resolves **own** properties only, matching `hasByPath()` so the two work as a guard/read pair. Inherited members never resolve, so `getByPath({}, 'toString')` returns the default value instead of `Function.prototype.toString`. An inherited member is never form data, and the own-property rule also keeps prototype internals such as `__proto__` unreachable unless they are genuine own data keys
```
heath-freenome: You probably do not need this since it just expands the information on the previous bullet item

packages/utils/src/lookupFromFormContext.ts
jimmycallin: Good catch — reverted to `unknown` in 40d2d0d04. The switch had been to keep the inferred return type as `any` (`getByPath`'s return type infers from the fallback argument), but it turns out nothing depends on that: with `unknown` the return type becomes `unknown` and both packages still typecheck clean, so the stricter type wins.

CHANGELOG.md
jimmycallin: Done in 40d2d0d04 — folded the own-property note into the main bullet as a short clause and dropped the rest.

packages/utils/src/lookupFromFormContext.ts
heath-freenome: Hmmm, looks like this change may have affected core tests

jimmycallin:
Pushed 2742ef229, which tightens the typing of the new path utilities and fixes the CI build failure on the previous commit.

**CI fix**: 40d2d0d04 changed `lookupFromFormContext`'s fallback back to `unknown`, which flipped its return type to `unknown` and broke `@rjsf/core:build` at `LayoutGridField.tsx:418`. It now takes a trailing `R = unknown` generic (`lookupFromFormContext<T, S, F, R>`) so callers declare the expected result type, and the LayoutGridField call sites do so.

**Stricter path utilities** — no `any` anywhere, and exactly one type assertion left in `pathUtils.ts`:
- `getByPath<R = unknown>(obj, path, defaultValue?): R` — the value at a runtime path can't be known statically, so `R` is the caller's declaration (same idea as `Map.get`), defaulting to `unknown` which forces narrowing. The one documented cast lives at its return.
- Internals traverse with `unknown` cursors via `Reflect.get`/`Reflect.set`; `setByPath` uses an honest `value is object` guard; `unsetByPath` uses `Reflect.deleteProperty` (returns `false` for non-configurable properties instead of relying on `delete` throwing).
- Boundary call sites (`getDefaultFormState`, `retrieveSchema`, `toPathSchema`, `findFieldInSchema`, `getFromSchema`, `getFirstMatchingOption`) now declare their expected type via `getByPath<T>`/`<S>` instead of casting the result.
- Two sites no longer need path utilities at all: `getWidget` caches its wrapper as a declared optional `MergedWidget` property on the `Widget` type, and `ErrorSchemaBuilder` walks the `ErrorSchema` with typed reads (an `ErrorSchema` only ever nests `ErrorSchema` blocks).

Docs updated to the new signatures. Verified with a full `pnpm run build` (18 projects), `pnpm run lint`, and uncached test runs for utils (1488 tests) and core (2235 tests), all green.

jimmycallin:
Pushed 71fa938e0: the validator packages are now converted as well, so all path-based lodash usage in this PR's scope is gone from `utils`, `core`, `validator-ajv8`, `validator-ata`, and `validator-cfworker`.

- The three error processors (`processRawValidationErrors`) now use `getByPath`/`toPath` with caller-declared result types (`getByPath<string | undefined>(...)` etc.) in place of `lodash/get` with casts; dotted validation-error paths are parsed explicitly with `toPath()`, matching the rest of the PR.
- `precompiledValidator` (ajv8/ata) read the constant `$id` key with plain typed property access — no path utility needed.
- `@rjsf/validator-cfworker` no longer uses lodash anywhere, so its `lodash`/`lodash-es` dependencies are dropped entirely. ajv8/ata keep lodash only for one `lodash/isObject` import each, which belongs to the non-path follow-up.

Verified with uncached test runs for all three packages (617 + 1285 + 2697 tests), builds, and lint — all green. CHANGELOG and the PR description's validator caveat updated.

jimmycallin:
Ran a code review over the branch and pushed the results (c5fc851cb, 675dfdd22, cffa27111). It caught two real bugs I'd introduced:

**Empty paths resolved to the whole object.** `getByPath(obj, [])` returned `obj` while `hasByPath(obj, [])` was `false` — contradicting the guard/read pairing these utilities document. This was reachable: `toPath('')` yields an empty segment list, so a **root-level validation error** (empty `instancePath`) made the validators' `getByPath(uiSchema, toPath(property))` return the *entire* uiSchema, letting `getUiOptions()` pull the root `title` into an unrelated error message. lodash's `get(obj, '')` returned `undefined`, so this restores prior behavior. Same class of bug in `unsetByPath(obj, [])`, which indexed `segments[-1]` and deleted the literal `"undefined"` key. Both now return the default / report success without touching the object, with tests pinning them.

**A lost default in the layout-grid schema walk.** Converting `get(innerData, part, {})` to `getByPath<T>(innerData, part)` dropped the empty-object fallback that the line's own comment documents, changing what `retrieveSchema()` receives when drilling into absent form data. Restored.

**Performance**, from the same review: `setByPath` was taking the general route for single-key writes (segment-list allocation + `.some()` per call) — a fast path takes it from 21 ns to 6.7 ns (0.32× lodash) and removes the per-segment allocation in `ErrorSchemaBuilder`'s block walk. `LayoutGridField` was also parsing `toPath(name)` twice per render.

I also **disclosed a regression the tables were hiding**: reads via a *dotted string* are 1.84× slower than lodash, because lodash memoized its string→segments parse and `toPath()` re-parses. The PR description now shows that row with the reasoning for not adding a memo cache (module-level mutable state plus a shared-array mutation hazard, for ~39 ns at per-field-per-render sites). Happy to add the cache if you'd prefer parity.

Two review suggestions I did **not** take, for the record: memoizing `toPath` (above), and collapsing `getUsedFormData`'s `hasByPath`+`getByPath` into one walk via a sentinel — that one changes behavior for an own key whose value is `undefined`, which isn't worth a few microseconds on submit.

**Types**: there is now **no `any`** introduced anywhere in this PR — the last one, `isObject(thing: any)`, is now `unknown` narrowed with `in` checks (an `unknown` parameter still accepts every argument, so callers are unaffected), and `ObjectField`'s two `as GenericObjectType` casts turned out to be leftovers from lodash's `set` and are gone. `pathUtils.ts` contains exactly one type assertion, the documented `as R` at `getByPath`'s return.

Verified: full build (18 projects), full test suite (17 projects, utils at 100% branch coverage), and lint all green.

packages/utils/src/ErrorSchemaBuilder.ts
```diff
@@ -43,12 +40,21 @@ export default class ErrorSchemaBuilder<T = any> {
...
+      return this.errorSchema;
+    }
+    // An `ErrorSchema` only ever nests `ErrorSchema` blocks, so walk it with typed reads, creating
+    // missing blocks along the way. `setByPath` guards the write against prototype-polluting segments
+    let errorBlock: ErrorSchema = this.errorSchema;
+    for (const segment of path) {
+      let next = Object.hasOwn(errorBlock, segment) ? errorBlock[segment] : undefined;
+      if (!next) {
+        next = {};
+        setByPath(errorBlock, segment, next);
```
heath-freenome: **Correctness (confirmed):** This loop calls `setByPath(errorBlock, segment, next)` per segment but never checks whether the write was silently refused. `setByPath`'s single-key fast path refuses to write `__proto__`/`constructor`/`prototype` segments (correct prototype-pollution guard) — but those are all *legal JSON Schema property names*. For a schema with a field literally named `constructor`, `setByPath` no-ops here (never attaches `next` to `errorBlock`), yet the loop unconditionally does `errorBlock = next` on the next line and keeps walking/mutating this now-orphaned object for the rest of the path. The caller (`addErrors`/`setErrors`/`clearErrors`) ends up mutating an object that's unreachable from `this.errorSchema` — the error is silently dropped from the real tree, no exception, no warning, the field's error just never renders.

Suggest checking the return/success of the write (or switching to `getByPath`/`Object.hasOwn` + direct assignment that doesn't silently refuse) so an unsafe segment fails loudly instead of silently losing data.

packages/core/src/components/Form.tsx
```diff
@@ -1039,15 +1045,16 @@ export default class Form<
...
+        ? getByPath(schemaValidationErrorSchema, path)
+        : schemaValidationErrorSchema;
       // If there is an old validation error for this path, assume we are updating it directly
       if (!_isEmpty(oldValidationError)) {
         // Apply the user-supplied newErrorSchema onto a clone of the AJV-only base, so that
         // mergeErrors below sees the user's error at this path without mutating shared state.
         if (!isRootPath) {
           mergeBaseErrorSchema = structuredClone(schemaValidationErrorSchema);
-          _set(mergeBaseErrorSchema, path, newErrorSchema);
+          setByPath(mergeBaseErrorSchema, path, newErrorSchema);
```
heath-freenome: **Correctness (confirmed, coordination risk):** This and the `customErrors.ErrorSchema` call a few lines below both omit `createIntermediateObjects=true`, so a missing intermediate segment followed by a numeric one auto-vivifies a real array in the `ErrorSchema` tree instead of the plain numeric-keyed object RJSF's own convention requires (matching how `ErrorSchemaBuilder`'s own rewritten `getOrCreateErrorBlock` always builds plain `{}` intermediates).

I checked this specifically because it's the exact shape-mismatch bug tracked in #5201 (silently-dropped inline field errors after an `ArrayField` add/remove/reorder touches a nested array). `main` currently still has plain `_set(...)` here — the fix for #5201 (converting to `_setWith(..., Object)`) hasn't merged yet — so this PR isn't a *regression* today, but whichever of the two branches merges second will need to convert these two call sites to `setByPath(..., true)`, or the bug reappears under a different name. Flagging now so it isn't lost in a rebase/merge.

packages/utils/src/pathUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+  return path.length > 0;
+}
+
+/** Removes the own property at `path` of `obj`, mutating `obj`
+ *
+ * @param obj - The object to modify
+ * @param path - The single key or list of path segments at which to remove the property
+ * @returns - True if the property was removed or did not exist, otherwise false
+ */
+export function unsetByPath(obj: unknown, path: ObjectPath): boolean {
```
heath-freenome, line 186: **Correctness (plausible):** `unsetByPath` has no `UNSAFE_SEGMENTS`/prototype-pollution guard at all, unlike `setByPath` — it relies purely on `getByPath`'s implicit own-property gate for intermediate segments and does an unguarded `Reflect.deleteProperty` on the terminal segment. The CHANGELOG/JSDoc broadly claims parity with "lodash ≥ 4.17.19's prototype-pollution hardening," but that hardening is asymmetric across lodash's own `set`/`unset`/`get`/`has` implementations and this port only reproduces it for `setByPath`.

Not currently exploitable (today's callers don't pass attacker-controlled paths here), but a future refactor of `getByPath`'s own-property gate — which `unsetByPath` implicitly depends on for safety — could silently reopen this with no local guard or test to catch it.

packages/utils/src/schema/getFromSchema.ts
```diff
@@ -31,16 +30,17 @@ function getFromSchemaInternal<T = any, S extends StrictRJSFSchema = RJSFSchema,
...
   let fieldSchema = schema;
-  if (has(schema, REF_KEY)) {
+  // hasByPath instead of `in` because `schema` can be undefined at runtime when drilling past a non-matching xxxOf
+  if (hasByPath(schema, REF_KEY)) {
     fieldSchema = retrieveSchema<T, S, F>(validator, schema, rootSchema, undefined, experimental_customMergeAllOf);
   }
   if (isEmpty(path)) {
     return fieldSchema;
   }
   const pathList = Array.isArray(path) ? [...path] : path.split('.');
```
heath-freenome: **Consistency (plausible):** This function's `get`/`has` calls were converted to `getByPath`/`hasByPath` a few lines up, but this line's own ad hoc `path.split('.')` was left in place instead of delegating to the new `toPath()`, which supports the fuller RJSF grammar (bracketed indexes like `'items[0].name'`).

`SchemaFieldPath` string arguments now silently only work with dot-separated segments here, while sibling utilities accepting the same conceptual path type (e.g. `Form.tsx`'s `focusOnError`) accept `toPath()`'s bracket-index grammar — a bracketed path passed to this function wouldn't split correctly, creating an inconsistent grammar between utilities that both claim to accept `SchemaFieldPath`.

packages/utils/src/isObject.ts
```diff
@@ -1,21 +1,23 @@
...
  */
-export default function isObject(thing: any): thing is object {
+export default function isObject(thing: unknown): thing is GenericObjectType {
   if (typeof thing !== 'object' || thing === null) {
     return false;
   }
   // lastModified is guaranteed to be a number on a File instance
   // as per https://w3c.github.io/FileAPI/#dfn-lastModified
-  if (typeof thing.lastModified === 'number' && typeof File !== 'undefined' && thing instanceof File) {
+  if ('lastModified' in thing && typeof thing.lastModified === 'number' && typeof File !== 'undefined' && thing instanceof File) {
```
heath-freenome: **Convention (confirmed):** This line (130 chars) and the `getMonth` guard a few lines below (124 chars) exceed the repo's 120-char print width (CLAUDE.md: "Prettier: single quotes, JSX single quotes, 120-char print width"; matches `.oxfmtrc.json`'s `printWidth: 120`). These are new lines added for TS narrowing on the now-`unknown`-typed parameter. Worth a `pnpm run cs-format` pass before merge if CI runs `cs-check` in check-mode.

packages/utils/src/schema/findFieldInSchema.ts
```diff
@@ -60,33 +58,33 @@ export default function findFieldInSchema<
...
+      if (hasByPath(parentField, ONE_OF_KEY)) {
         // if this sub-path has a `oneOf` then use the formData to drill into the schema with the selected option
         parentField = findSelectedOptionInXxxOf(
           validator,
           rootSchema,
           parentField,
           fieldNameKey,
           ONE_OF_KEY,
-          get(formData, subPath),
+          getByPath<T>(formData, [subPath]),
```
heath-freenome: **Efficiency (plausible), pattern repeated elsewhere:** `subPath` is a scalar element of `FieldPathList`, but it's wrapped in a one-element array (`[subPath]`) before being passed to `getByPath`, forcing the slower general multi-segment path (with its own `Array.isArray`/`Object.hasOwn` traversal) instead of the documented single-key fast path a bare value would hit — and allocating an array for no reason. Same pattern repeats at line 80 in this file, and in `getOptionMatchingSimpleDiscriminator.ts`, `findSelectedOptionInXxxOf.ts`, `getFirstMatchingOption.ts`, and twice in `retrieveSchema.ts`.

These run per-property/per-render (this function's oneOf/anyOf path walk, `retrieveSchema`'s property-resolution loop), so the redundant allocation repeats on every keystroke in a live-validated form. It's also a bit of a correctness trap for future readers — wrapping a provably-scalar value in `[x]` looks like intentional multi-segment handling and invites someone to "generalize" a call site that was never meant to take a real path. (Contrast with `LayoutMultiSchemaField.tsx:499`'s `getByPath(formData, selectorField)`, which correctly keeps the same kind of variable un-wrapped.)

packages/utils/src/ErrorSchemaBuilder.ts
```diff
@@ -43,12 +40,21 @@ export default class ErrorSchemaBuilder<T = any> {
...
-      errorBlock = {};
-      setWith(this.errorSchema, pathOfError, errorBlock, Object);
+    // A string path is dotted-path notation, so parse it into its segments
+    const path = typeof pathOfError === 'string' ? toPath(pathOfError) : (pathOfError ?? []);
+    if (path.length === 0) {
+      return this.errorSchema;
+    }
+    // An `ErrorSchema` only ever nests `ErrorSchema` blocks, so walk it with typed reads, creating
+    // missing blocks along the way. `setByPath` guards the write against prototype-polluting segments
+    let errorBlock: ErrorSchema = this.errorSchema;
```
heath-freenome: **Altitude (plausible):** This loop hand-rolls a "walk the path, create missing plain-object containers" traversal that `setByPath(obj, path, value, /* createIntermediateObjects */ true)`'s own general multi-segment loop already implements — with a subtly different existence check (`if (!next)`, falsy, vs. `setByPath`'s `isSettable(existing)`).

This duplication is also the direct site of the unsafe-segment-drop bug flagged separately on line 55 above — a single call into `setByPath`'s own traversal (which already knows how to skip unsafe segments safely) would likely have avoided that bug rather than requiring a second, independent reimplementation of the same semantics to get right.

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -702,7 +702,7 @@ export function getArrayDefaults<T = any, S extends StrictRJSFSchema = RJSFSchem
           experimental_defaultFormStateBehavior,
           experimental_customMergeAllOf,
           rawFormData: item,
-          parentDefaults: get(defaults, [idx]),
+          parentDefaults: Array.isArray(defaults) ? defaults[idx] : undefined,
```
heath-freenome: **Correctness (plausible), minor:** Every other `get(defaults, [key])` in this file was mechanically converted to `getByPath(defaults, [key])` (which resolves any own property regardless of container type), but this one site was hand-rewritten as `Array.isArray(defaults) ? defaults[idx] : undefined` — an intentional-looking deviation from the file's own 1:1 conversion pattern.

Old lodash `get(defaults, [idx])` would still resolve `defaults[idx]` if `defaults` ever arrived as a non-Array-but-object value (e.g. a malformed/legacy `schema.default` shaped as `{0: {...}, 1: {...}}`, or a custom `computeDefaults` extension returning an object). The new code always returns `undefined` in that case, silently dropping supplied per-item defaults when populating an array field's defaults. Low likelihood in normal use, but worth an explicit sign-off that the narrowing is intentional.

heath-freenome (commented):
Automated review (8-angle finder pass + verification). 8 findings below, most-severe first. I specifically cross-checked this against my own recent work fixing #5201 (an ErrorSchema array/object shape-mismatch bug in Form.tsx/ArrayField.tsx) since this PR touches exactly those files, and traced one real new bug plus one coordination risk with that unmerged fix. The rest of this refactor is unusually careful — several angles confirmed it actually *fixes* latent lodash bugs (e.g. dotted-string auto-splitting silently corrupting object keys that legitimately contain dots) rather than introducing them.

packages/utils/src/ErrorSchemaBuilder.ts
jimmycallin: Good catch, and it went deeper than the call site. Fixed at the root in `setByPath`: the only segment that can actually mutate the prototype chain on assignment is `__proto__` (it hits the setter). `constructor` and `prototype` were only dangerous because `setByPath`’s intermediate reads resolved inherited members — `current[segment]` walked into the real `Object.prototype`. Those reads are now own-property-only, matching `getByPath`/`hasByPath`, so `["constructor", "prototype", "x"]` shadows `constructor` with a fresh own object and `UNSAFE_SEGMENTS` shrinks to just `__proto__`.

So the silent drop is gone rather than made loud: a field literally named `constructor` or `prototype` now round-trips. `__proto__` as a literal data key still cannot be written (same as lodash) and is documented as such. New tests cover all three.

packages/utils/src/ErrorSchemaBuilder.ts
jimmycallin: Agreed — collapsed onto `setByPath`’s own traversal:

```ts
let errorBlock = getByPath<ErrorSchema | undefined>(this.errorSchema, path);
if (!errorBlock) {
  errorBlock = {};
  setByPath(this.errorSchema, path, errorBlock, true);
}
return errorBlock;
```

`createIntermediateObjects` keeps every intermediate a plain `{}`, which is what the hand-rolled loop was there for. The divergent `!next` vs `isSettable(existing)` check goes with it.

packages/core/src/components/Form.tsx
jimmycallin: Done — both sites now pass `createIntermediateObjects = true`, with a comment saying why (an `ErrorSchema` nests plain objects even at numeric segments). Whichever branch merges second, this side is already correct for #5201.

packages/utils/src/pathUtils.ts
jimmycallin: Right that there was no guard, but after the `setByPath` change on the other thread I think none is needed and I have added tests to pin that down. `unsetByPath` resolves the parent through `getByPath`, which is own-property-only, and the terminal operation is a delete — `Reflect.deleteProperty(parent, "__proto__")` on a non-own `__proto__` is a no-op returning `true`, it cannot set a prototype. The dependency on `getByPath`’s own-property gate is now stated in the JSDoc and covered by a test that fails if that gate is ever loosened.

packages/utils/src/schema/getFromSchema.ts
jimmycallin: Fixed — `getFromSchema` now uses `toPath(path)`, so bracketed indexes parse the same way here as everywhere else.

packages/utils/src/isObject.ts
jimmycallin: Formatted — `pnpm run cs-format`. (CI does run `cs-check`.)

packages/utils/src/schema/findFieldInSchema.ts
jimmycallin: Agreed, and it was more widespread than the sites listed — unwrapped every provably-scalar `[x]` across `findFieldInSchema`, `getOptionMatchingSimpleDiscriminator`, `findSelectedOptionInXxxOf`, `getFirstMatchingOption`, `retrieveSchema`, `toPathSchema`, `getDefaultFormState`, `ObjectField` and `ArrayField`. All hit the single-key fast path now.

packages/utils/src/schema/getDefaultFormState.ts
jimmycallin: Not intentional, just an inconsistency — reverted to `getByPath<T>(defaults, idx)` so it matches the rest of the file and keeps the old resolve-any-own-property behaviour.

heath-freenome (approved):

PR #5205, opened 2026-08-20
fix: clear only the error of the field that changed

### LINKED ISSUE #5197: Multiple errors in array objects disappear when only one is fixed.
### Prerequisites

### What theme are you using?

mui

### Version

6.1.2

### Current Behavior

When there are errors in object validtion in the array, changin one field results in clearing error in other fields of the same object and any other object item in the same array.

Here is [playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJiYXoiOlt7fV19LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsiZm9vIjp7InR5cGUiOiJzdHJpbmcifSwiYmFyIjp7InR5cGUiOiJzdHJpbmcifSwiYmF6Ijp7InR5cGUiOiJhcnJheSIsIml0ZW1zIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7InF1eCI6eyJ0eXBlIjoic3RyaW5nIn0sImNvcmdlIjp7InR5cGUiOiJzdHJpbmcifX0sInJlcXVpcmVkIjpbInF1eCIsImNvcmdlIl19fX0sInJlcXVpcmVkIjpbImZvbyIsImJhciJdfSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6ZmFsc2UsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50IiwibmVzdGVkRGVmYXVsdHNQcmVjZWRlbmNlIjoiZGVzY2VuZGFudFdpbnMifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOmZhbHNlLCJpZFByZWZpeCI6InJvb3QiLCJpZFNlcGFyYXRvciI6Il8ifSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJCbGFuayJ9).

You may press Validate and see error alerts for missing required fields.

When you fill in "foo" field, error for it disappears - it works OK.

When you fill in "qux" field, error for it disappears, but also error for "corge" field disappears - it norks NOK.

Also when you add one item to the array, after pressing validate, there are errors near all field, if you fill in any of them, all errors disappear.

### Expected Behavior

Errors are upaded only on fields that are changed. Like in object outside array, vide "foo/bar" object.

### Steps To Reproduce

1. Use [playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJiYXoiOlt7fV19LCJzY2hlbWEiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsiZm9vIjp7InR5cGUiOiJzdHJpbmcifSwiYmFyIjp7InR5cGUiOiJzdHJpbmcifSwiYmF6Ijp7InR5cGUiOiJhcnJheSIsIml0ZW1zIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7InF1eCI6eyJ0eXBlIjoic3RyaW5nIn0sImNvcmdlIjp7InR5cGUiOiJzdHJpbmcifX0sInJlcXVpcmVkIjpbInF1eCIsImNvcmdlIl19fX0sInJlcXVpcmVkIjpbImZvbyIsImJhciJdfSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6ZmFsc2UsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoic2tpcERlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50IiwibmVzdGVkRGVmYXVsdHNQcmVjZWRlbmNlIjoiZGVzY2VuZGFudFdpbnMifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOmZhbHNlLCJpZFByZWZpeCI6InJvb3QiLCJpZFNlcGFyYXRvciI6Il8ifSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJCbGFuayJ9)
2. Press "Validate"
3. Notice errors on fields: foo, bar, baz.qux, baz.corge
4. Type "a" in field "foo"
5. Notice disappeared error on "foo"
6. Notice error on "bar" still visible
7. Type "a" in field "qux"
8. Notice disappeared error on "qux"
9. Notice wrongly disappeared error on "corge"

### Environment

```markdown
- OS: Windows 11
- Node: 22.14.0
```

### Anything else?

_No response_

luantaraschi:
Fixes #5197.

## The defect

Filling in one field of an array item clears the errors of every other field in that item, and of every other item in the same array. The reporter's steps, on a schema with `foo`/`bar` required at the root and `qux`/`corge` required inside `baz` items:

| step | errors on screen |
|---|---|
| submit | foo, bar, qux, corge |
| type into `foo` | bar, qux, corge |
| type into `baz.0.qux` | bar |

`corge` was never touched and its error should still be there.

## Why

When `formData` comes back in through props, `getSnapshotBeforeUpdate` asks `getChangedFields` which fields moved, and `getStateFromProps` clears the error of each one:

```ts
const newErrorSchema = formDataChangedFields.reduce<Record<string, undefined>>((acc, key) => {
  acc[key] = undefined;
  return acc;
}, {});
```

`getChangedFields` only walks the top level. A change to `baz.0.qux` is reported as `baz`, so the object handed to `mergeObjects` is `{ baz: undefined }` and the whole `baz` branch of the error schema goes with it. A top-level scalar like `foo` is unaffected, which is why the bug only shows up one level down and looks like an array problem.

## The change

`getChangedFields` gains an optional `deep` flag. With it, a field holding a nested object or a same-length array is descended into and the dotted path of the field that actually changed is returned (`baz.0.qux`) instead of the top-level key holding it (`baz`). Existing callers are untouched: the flag defaults to false and every current test passes unchanged.

One place where it deliberately stops short rather than guessing: **an array whose length changed**. Items shift, so nothing below them can be matched up by index and the array itself is reported. Clearing the whole branch is right there anyway, since the errors no longer line up with the items.

A name holding a `.` or a `[` is descended into like any other. The path it produces cannot be told apart from a path through nested keys, and neither can the entry an `ErrorSchema` keeps for it: `toErrorSchema` runs the property name through `toPath`, so the error of a field under a name like `has.dot` is written at `errorSchema.has.dot`, which is where the walk's own path lands.

`Form` passes `true` at the one call site that feeds the error clearing, and builds the object it hands to `mergeObjects` with `setByPath(acc, toPath(path), undefined, true)`. The `toPath` splits the path the same way `toErrorSchema` splits a validation error property, so the two address the same entry, and the forced intermediate objects keep the numeric segment of an array item as an object key, which is how an `ErrorSchema` addresses array items. `mergeObjects` already recurses into a nested right-hand side, so `{ baz: { 0: { qux: undefined } } }` clears exactly `baz.0.qux`.

Every container along that path changed together with the field, so its own `__errors` goes too, and only its own: the other fields it holds did not change and keep theirs. Without that, an array with `uniqueItems` went on reporting duplicates after one of them had been edited into something else, which the shallow clearing this replaces did handle, by dropping the whole branch.

The other `getChangedFields` call, `stateDataChangedFields`, is only read as `.length > 0` and is left shallow. The deep walk itself reaches `getStateFromProps` as a function and is called inside the `!mustValidate` branch, so with `liveValidate` on it never runs.

## Tests

- `Form.test.tsx`: the reporter's steps, asserted on the errors on screen after each one. Then a field whose name holds a dot, an object sitting under such a name, an error owned by the array rather than by the item it holds, and a same-length reorder of array items. They run under `describeRepeated`, so each one covers the four `omitExtraData`/`liveOmit` combinations.
- `getChangedFields.test.ts`: the deep flag on a nested object, on an array item, with several fields changed at once, on a name holding a dot and on one holding brackets, and the cases where it stops short (length change, non-object on one side, a name whose difference cannot be narrowed). Plus a case pinning that the shallow result did not move.

`@rjsf/utils` 1504 passing with branches at 100%, `@rjsf/core` 2255 passing.


CHANGELOG.md
```diff
@@ -21,10 +21,12 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/core
 
 - Fixed `TimeWidget` to respect schema `multipleOf` time precision, preserving second-precision values while keeping minute-precision native inputs synchronized with their displayed value ([#5174](https://github.com/rjsf-team/react-jsonschema-form/pull/5174))
+- Fixed the clearing of errors on change so that only the field that actually changed loses its error, rather than every field under the top-level key holding it, fixing [#5197](https://github.com/rjsf-team/react-jsonschema-form/issues/5197) ([#5205](https://github.com/rjsf-team/react-jsonschema-form/pull/5205))
```
heath-freenome: We've released 6.8.0. Move these to 6.8.1

luantaraschi: Moved to 6.8.1, thanks.

luantaraschi:
The red build was the utils coverage gate rather than a test failure. `getChangedFields.ts` line 28 has two sides and I only covered one: an array whose changed item is itself an object takes the mapping side, and an array of plain values takes `[String(index)]`, which nothing exercised. Added `{ items: ['one', 'two'] }` against `{ items: ['one', 'three'] }`, which comes back as `items.1`. Branches are at 1812/1812 locally now.

packages/core/src/components/Form.tsx
```diff
@@ -691,10 +693,13 @@ export default class Form<
...
+        // `formDataChangedFields` carries the path of each field that changed, so clearing has to follow that path
+        // instead of dropping the whole branch it starts in. `setWith` with `Object` keeps the numeric segment of an
+        // array item as an object key, which is how an `ErrorSchema` addresses array items.
+        const newErrorSchema = formDataChangedFields.reduce<GenericObjectType>(
+          (acc, path) => _setWith(acc, path, undefined, Object),
+          {},
+        );
         schemaValidationErrorSchema = mergeObjects(
           currentErrors.errorSchema,
           newErrorSchema,
```
heath-freenome: `_setWith` interprets a dotted string as a **nested path**, not a literal key — so any schema property whose name contains a `.` will silently fail to have its error cleared after an edit.

The `.includes('.')` guard in `getChangedFields` correctly prevents descent into a key like `"foo.bar"`, returning it as the literal string `"foo.bar"`. But that string then flows here, and lodash parses `"foo.bar"` as `foo → bar`:

```ts
setWith({}, 'foo.bar', undefined, Object)
// produces { foo: {} }  ← wrong
// wanted  { 'foo.bar': undefined }
```

The resulting `newErrorSchema` targets the wrong path, so `mergeObjects` leaves `errorSchema['foo.bar']` untouched and the stale error stays visible. The same issue exists for keys returned by the `diffFields` branch (keys new in `b` that aren't in `a`) — those also bypass the dot-guard and feed straight into `_setWith`.

The pre-PR code used `acc[key] = undefined` (plain bracket assignment), which handled literal dotted keys correctly. One option that preserves that property while still building a nested object for normal dotted-path strings:

```ts
const newErrorSchema = formDataChangedFields.reduce<GenericObjectType>((acc, path) => {
  if (path.includes('.')) {
    _setWith(acc, path, undefined, Object);
  } else {
    acc[path] = undefined;
  }
  return acc;
}, {});
```

Though this is a band-aid — the deeper fix is to carry segment arrays (not dotted strings) from `getChangedFields` so the boundary between "literal key" and "path" never has to be re-inferred by the consumer.

packages/core/src/components/Form.tsx
```diff
@@ -468,7 +470,7 @@ export default class Form<
   ): { nextState: FormState<T, S, F>; shouldUpdate: true } | { shouldUpdate: false } {
     if (!deepEquals(this.props, prevProps)) {
```
heath-freenome, line 476: The deep traversal of the entire `formData` tree runs unconditionally here, but the leaf paths it produces are only consumed inside `if (!mustValidate)` (line 693 in `getStateFromProps`). When `liveValidate={true}` — a very common configuration — `mustValidate` is `true`, that branch is skipped entirely, and the O(tree-size) traversal is discarded.

The `formDataChangedFields.length > 0` check just below (line 479, `isFormDataChanged`) only needs to know whether *something* changed, which the old shallow call already provided correctly. The leaf paths are the new capability, and they are only needed when `!mustValidate`.

A low-overhead fix: leave this call shallow, and compute the deep variant lazily inside `getStateFromProps` only when `!mustValidate` is known:

```ts
// getSnapshotBeforeUpdate — stays cheap:
const formDataChangedFields = getChangedFields(this.props.formData, prevProps.formData);

// getStateFromProps — only pay for depth when it's actually used:
if (formDataChangedFields.length > 0 && !mustValidate) {
  const deepChangedFields = getChangedFields(props.formData, prevFormData, true);
  const newErrorSchema = deepChangedFields.reduce(...);
  ...
}
```

This keeps the fix correct for the target case while avoiding a per-keypress full-tree scan on every `liveValidate` form.

packages/core/src/components/Form.tsx
luantaraschi: The name never reaches the error schema as a key, so there is no `errorSchema['foo.bar']` left behind. `toErrorSchema` runs the property through `toPath`, and the validator reports the missing property as `foo.bar`, so the error lands at the path that name spells out:

```js
validator.validateFormData({}, {
  type: 'object',
  properties: { 'foo.bar': { type: 'string' }, baz: { type: 'string' } },
  required: ['foo.bar', 'baz'],
}).errorSchema;
// {
//   foo: { bar: { __errors: ["must have required property 'foo.bar'"] } },
//   baz: { __errors: ["must have required property 'baz'"] },
// }
```

The same holds one level down (`outer['foo.bar']` lands at `outer.foo.bar`) and for a name holding brackets (`has[0]` lands at `has.0`). That is the addressing `_setWith` reproduces, so the string clears the entry the validator actually wrote. It is the bracket assignment it replaced that missed: put `acc[key] = undefined` back and the error survives while a dead key is added beside it.

```
- "bar": undefined,
+ "bar": { "__errors": ["must have required property 'foo.bar'"] },
+ "foo.bar": undefined,
```

That is the new test in 6f73c61 failing against the old line. It is green as it stands.

`diffFields` falls under the same rule: a key that is new in `b` is addressed the same way, dot or no dot.

One note on the suggested patch: as written both branches do the same thing, since `_setWith(acc, 'foo', undefined, Object)` and `acc['foo'] = undefined` produce the same object, and the dotted case is the one that goes through `_setWith`.

Where a string does lose is a schema holding both `'foo.bar'` and a nested `foo: { bar }`. Their errors already share `errorSchema.foo.bar`, so clearing one clears the other. That collision comes from `toErrorSchema` flattening the name into a path and it predates this PR. Carrying segments would have to start there rather than in `getChangedFields`, since the error schema is where the two get merged. Happy to open a separate issue for it.

packages/core/src/components/Form.tsx
luantaraschi: Done in 029bfe3. `getSnapshotBeforeUpdate` keeps the shallow call for the length check, and the deep walk goes to `getStateFromProps` as a function that is only called inside the `!mustValidate` branch, so with `liveValidate` on it never runs. It is a function rather than the previous `formData` because `prevProps` is not in scope down there.

On the size of what was being wasted: the walk is bounded by the branch that changed, not by the tree, since a top-level key that compares equal is never descended into. Measured here, average per call over 2000 iterations:

| formData | shallow | deep |
| --- | --- | --- |
| 200 objects at the top, one edited | 0.070 ms | 0.072 ms |
| array of 1000 items, one leaf edited | 0.207 ms | 0.363 ms |
| object 10 levels deep, deepest leaf edited | 0.003 ms | 0.028 ms |

Small, but there is no reason to pay it when the result gets thrown away.

packages/core/src/components/Form.tsx
```diff
@@ -55,6 +56,7 @@ import _get from 'lodash/get';
 import _isEmpty from 'lodash/isEmpty';
 import _pick from 'lodash/pick';
 import _set from 'lodash/set';
+import _setWith from 'lodash/setWith';
```
heath-freenome: There may be one challenge using this as another contributor is in the process of removing lodash from the system. See this #5204

luantaraschi: That one is a straight swap once #5204 lands. Its `setByPath` takes a `createIntermediateObjects` flag that does what the `Object` customizer does here, so the line becomes `setByPath(acc, toPath(path), undefined, true)`.

I merged the two branches locally and applied it to check: `@rjsf/utils` 1493 passing, `@rjsf/core` 2243 passing. The conflicts are the lodash import block in Form.tsx, the one line in `getChangedFields` where `get` became `getByPath`, and then the changelog and the docs page.

The `toPath` has to stay, though. A bare string is one literal key under the new contract, so `setByPath(acc, path, undefined, true)` writes `{ 'baz.0.qux': undefined }` and the error is never cleared. That variant fails 8 tests here, the two this PR adds across the four `describeRepeated` combinations, so a mechanical rewrite gets caught rather than landing quietly.

Whichever of the two lands first, I will take the rebase on this side.

heath-freenome (commented):
Automated review (8-angle finder pass + verification). 7 findings below, most-severe first. The core `setWith(..., Object)`/`mergeObjects` merge mechanics are sound (verified empirically), and the lazy-callback conversion in `getStateFromProps` is a clean, complete refactor with no regression. The important finding is that the dot-guard doesn't actually achieve what it's meant to — it was verified by running the code end-to-end, not just read statically.

packages/utils/src/getChangedFields.ts
```diff
@@ -35,7 +70,13 @@ export default function getChangedFields(a: unknown, b: unknown): string[] {
   }
   const unequalFields = Object.entries(a as object)
     .filter(([key, value]) => !deepEquals(value, get(b, key)))
-    .map(([key]) => key);
+    .flatMap(([key, value]) => {
+      if (!deep || key.includes('.')) {
```
heath-freenome: **Correctness (confirmed by running the code):** This guard doesn't actually prevent the ambiguity it's meant to. When a key contains a `.`, it's returned unparsed as a single string — but `Form.tsx`'s `_setWith(acc, path, undefined, Object)` re-splits that exact same string on the dot anyway when consuming it. So for a schema property literally named e.g. `'has.dot'` that is itself an object with multiple required sub-fields, only one of which changed, the *entire* subtree gets wiped — reproducing the exact #5197 over-clearing bug this PR exists to fix, just for this specific case.

I verified this end-to-end: `getChangedFields(new, old, true)` returns `['has.dot']` (guard fires, stops short) for a `has.dot` object with `inner`/`other` required sub-fields where only `inner` was fixed. But `_setWith({}, 'has.dot', undefined, Object)` still parses on the dot → `{has: {dot: undefined}}` → merged into the error schema, `other`'s still-valid required-property error is wiped even though it was never touched. The PR's own dotted-key test (`'foo.bar'`) only covers a simple string leaf field, so this gap isn't caught by the new tests.

Since the guard can't actually prevent the collision at the consumer side, the real fix probably needs to happen where the dotted string gets built/consumed (see the bracket-notation finding below for the same root cause) — e.g. building the ErrorSchema-clearing patch from segment arrays rather than dotted strings that get re-parsed by lodash.

packages/core/src/components/Form.tsx
```diff
@@ -690,17 +695,23 @@ export default class Form<
...
-        ) as ErrorSchema<T>;
-        errorSchema = schemaValidationErrorSchema;
+      if (!mustValidate) {
+        const formDataChangedFields = getFormDataChangedFields();
+        if (formDataChangedFields.length > 0) {
+          // `formDataChangedFields` carries the path of each field that changed, so clearing has to follow that path
+          // instead of dropping the whole branch it starts in. `setWith` with `Object` keeps the numeric segment of an
+          // array item as an object key, which is how an `ErrorSchema` addresses array items.
+          const newErrorSchema = formDataChangedFields.reduce<GenericObjectType>(
+            (acc, path) => _setWith(acc, path, undefined, Object),
```
heath-freenome: **Correctness (confirmed):** `_setWith(acc, path, undefined, Object)` parses `path` as a lodash dotted-path *string*, including bracket notation. The PR's ambiguity guard in `getChangedFields` only checks for literal `.` characters, not `[`/`]` — a schema property literally named e.g. `'a[0]'` (valid JSON) would have its clearing path corrupted: `lodash.toPath('a[0]')` produces `['a', '0']`, not `['a[0]']`, so this ends up clearing/creating `acc.a[0]` instead of the literal key `acc['a[0]']`.

The pre-fix code used plain `acc[key] = undefined` — direct property assignment with no path-string parsing at all — so this is a genuinely new hazard introduced by switching to `_setWith`. Same root cause as the dot-guard finding above: any key that flows into this reduce as a dotted string is subject to lodash's full path grammar, which is broader than the `.`-only contract the producer assumes.

packages/utils/src/getChangedFields.ts
```diff
@@ -35,7 +70,13 @@ export default function getChangedFields(a: unknown, b: unknown): string[] {
...
+        return [key];
+      }
+      const descendants = getChangedDescendants(value, get(b, key));
+      return descendants.length ? descendants.map((path) => `${key}.${path}`) : [key];
+    });
   const diffFields = difference(Object.keys(b as object), Object.keys(a as object));
```
heath-freenome: **Correctness (plausible):** `diffFields` (keys present on one side and absent on the other, i.e. added/removed between the two `formData` snapshots) are concatenated onto the result completely unguarded by the dot-check applied to `unequalFields` a few lines up — an inconsistency in this fix's own coverage.

A field whose name contains a `.` or `[` that gets added/removed between renders (e.g. a conditional/`oneOf` branch switch) skips the guard entirely and flows straight into `_setWith` with none of the (partial) protection the changed-value path has, hitting the same hazards as the two findings above via a path the guard never considers.

packages/utils/src/getChangedFields.ts
```diff
@@ -4,6 +4,33 @@ import isPlainObject from 'lodash/isPlainObject';
...
+ * @param a - The first value, representing the original data to compare
+ * @param b - The second value, representing the updated data to compare
+ * @returns - An array of dotted paths, relative to `a`
+ */
+function getChangedDescendants(a: unknown, b: unknown): string[] {
+  if (isPlainObject(a) && isPlainObject(b)) {
+    return getChangedFields(a, b, true);
+  }
+  // Arrays of a different length shift their items around, so nothing below them can be matched up by index.
+  if (Array.isArray(a) && Array.isArray(b) && a.length === b.length) {
```
heath-freenome, line 17: **Correctness (plausible):** This only falls back to reporting the whole array as changed when *lengths* differ; it doesn't detect a same-length *reorder* (items shifted position without the array growing/shrinking), and will happily descend into `oldArray[i]` vs `newArray[i]` as if they're the same logical item.

In a controlled form, an `ArrayField` move-up/move-down reorder produces a same-length array where the item at each index differs from before, even though nothing was actually *edited* — items just moved. This recurses into positionally-mismatched item pairs; if they happen to share some field values by coincidence, only the differing fields get reported as "changed" at that index — potentially clearing the wrong item's error while leaving a stale error behind for the item that actually moved. The PR's own stated rationale for the length-mismatch fallback ("items shift, so nothing below them can be matched up by index") applies equally here.

packages/core/test/Form.test.tsx
```diff
@@ -2363,6 +2363,111 @@ describeRepeated('Form common', (createFormComponent) => {
...
+          // the path that name spells out, not under the name itself
+          expect(formRef.current!.state.errorSchema).toEqual({
+            foo: { bar: { __errors: ["must have required property 'foo.bar'"] } },
+            baz: { __errors: ["must have required property 'baz'"] },
+          });
+
+          // Clearing has to reach the same place, and leave the field that was not touched alone
+          await user.type(node.querySelector('[id="root_foo.bar"]')!, 'a');
+          expect(formRef.current!.state.errorSchema).toEqual({
+            foo: { bar: undefined },
```
heath-freenome, line 2467: **Design note (worth confirming is intentional):** Clearing an error via `mergeObjects` + an `undefined` value (rather than deleting the key) leaves a stale key with an `undefined` value in the `ErrorSchema` — and this test proves that shape now reaches the *public* `errorSchema` shipped in `Form.state`/the `onChange` event, not just an internal intermediate: `formRef.current!.state.errorSchema` is asserted to equal `{ foo: { bar: undefined }, ... }`, i.e. `errorSchema.foo.bar` is present-but-`undefined`, not removed.

Third-party code using the common defensive pattern `Object.keys(errorSchema.foo ?? {}).length > 0` to decide whether to show a "section has errors" indicator would see a false positive here — the key structurally exists even though there's no real error left under it. This is `mergeObjects`'s known "never deletes stale keys" behavior showing up on the public API surface rather than staying an internal merge detail. Might be worth a deliberate cleanup pass (e.g. via `ErrorSchemaBuilder.clearErrors`, which does properly reset to `[]`) rather than leaving `undefined` sentinels in the shipped errorSchema.

CHANGELOG.md
```diff
@@ -16,6 +16,16 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.8.1
```
heath-freenome: **Heads up, not a code issue:** this inserts a new `# 6.8.1` heading at the same insertion point as an unrelated, separately in-flight fix (issue #5201, branch `fix-5201`) — whichever of the two merges second will conflict here. If resolved mechanically rather than by hand-merging the two packages' bullet lists under one shared heading, it'll either drop a changelog entry or produce two duplicate adjacent `# 6.8.1` headings. Just flagging for whoever merges second.

packages/core/src/components/Form.tsx
```diff
@@ -690,17 +695,23 @@ export default class Form<
...
-          return acc;
-        }, {});
-        schemaValidationErrorSchema = mergeObjects(
-          currentErrors.errorSchema,
-          newErrorSchema,
-          'preventDuplicates',
-        ) as ErrorSchema<T>;
-        errorSchema = schemaValidationErrorSchema;
+      if (!mustValidate) {
+        const formDataChangedFields = getFormDataChangedFields();
```
heath-freenome, line 704: **Altitude (worth discussing):** The `setWith(obj, path, value, Object)` idiom ("keep a numeric path segment as an object key, not an array index, when building an ErrorSchema tree") is now independently re-derived in at least three places: this reduce, `ErrorSchemaBuilder.getOrCreateErrorBlock`'s existing `setWith(this.errorSchema, pathOfError, errorBlock, Object)`, and a separate not-yet-merged fix for #5201 — with no shared `@rjsf/utils` primitive enforcing the invariant.

Relatedly: `ErrorSchemaBuilder` already has an array-based path type (`FieldPathList`) that bypasses lodash's string-path parser entirely. Building this error-clearing patch from `FieldPathList` segments (e.g. via `ErrorSchemaBuilder.clearErrors(path)` per changed field) rather than dotted strings re-parsed by `_setWith` would likely have sidestepped the dot/bracket hazards found above structurally, instead of needing a guard to patch around them after the fact.

packages/core/src/components/Form.tsx
heath-freenome: I believe the PR merged so please update

luantaraschi: Done. `_setWith(acc, path, undefined, Object)` is now `setByPath(acc, toPath(path), undefined, true)`, and `getChangedFields` reads with `getByPath`. The `toPath` is load bearing: a bare string is one literal key under the new contract, so without it the clearing writes `{'baz.0.qux': undefined}` and never reaches the error.

The changelog entries went into the `@rjsf/utils` and `@rjsf/core` lists under the `# 6.9.0` heading #5204 opened, so there is no second heading.

`@rjsf/utils` 1504 passing with branches at 100%, `@rjsf/core` 2255 passing.

packages/utils/src/getChangedFields.ts
luantaraschi: You are right that the guard buys nothing, and it is gone. The fix went the other way from the one suggested, though.

The ambiguity is real, and the entry being cleared carries the same one: `toErrorSchema` spells such a name out as a path too. A property literally named `has.dot` holding `inner` and `other` has its errors written under it as a path:

```js
validator.validateFormData({ 'has.dot': {} }, schema).errorSchema;
// { has: { dot: { inner: { __errors: [...] }, other: { __errors: [...] } } } }
```

So descending and producing `has.dot.inner` lands on the entry the validator wrote, and `other` keeps its own. Carrying segment arrays would address `errorSchema['has.dot'].inner`, which is never written, so the clearing would stop working for those names entirely.

Your case, before and after:

```
- { has: { dot: { inner: undefined, other: { __errors: [...] } } } }
+ { has: { dot: undefined } }
```

That is the new test failing against the guard. It is green with the guard gone.

Something else turned up while measuring this, worth its own issue. A field whose name holds a dot never displays its error at all, in any version. The error is written at `errorSchema.foo.bar` and the field reads the literal key `'foo.bar'`, so nothing renders. That predates this PR. Happy to file it.

packages/core/src/components/Form.tsx
luantaraschi: Measured, and this goes the other way too. `toErrorSchema` runs the property through `toPath`, so a property literally named `a[0]` has its error written at `errorSchema.a['0']`:

```js
validator.validateFormData({}, {
  type: 'object',
  properties: { 'a[0]': { type: 'string' }, b: { type: 'string' } },
  required: ['a[0]', 'b'],
}).errorSchema;
// { a: { 0: { __errors: ["must have required property 'a[0]'"] } }, b: { __errors: [...] } }
```

`setByPath(acc, toPath('a[0]'), undefined, true)` reproduces that addressing and clears the entry that is there. The `acc[key] = undefined` it replaced wrote the literal key `'a[0]'`, which the error schema never uses, so it cleared nothing and added a dead key beside the error it missed. Same as the dotted names above.

packages/utils/src/getChangedFields.ts
luantaraschi: `diffFields` are top-level keys and `toErrorSchema` addresses them like every other name, so the answer above applies unchanged: the dotted string is what reaches the entry, and the literal key is what misses it. With the guard gone every key takes the same route, so there is no inconsistency left to fix.

packages/utils/src/getChangedFields.ts
luantaraschi: Measured this one. The clearing goes by path rather than by item identity, and an error at a path whose value did not change is still the error of the value sitting there.

`{ baz: [{}, { corge: 'z' }] }`, submit, then move the first item down:

```
before   ["must have required property 'qux'", "must have required property 'corge'", "must have required property 'qux'"]
after    ["must have required property 'qux'", "must have required property 'qux'"]
```

Neither item holds a `qux` after the move, so both of those are the right errors for the values now at those paths. Only the `corge` that moved loses its own, and it comes back on the next validation like any edited field does. Covered by a test now.

Looking at it did turn up a real one next door. An error owned by a container rather than by a leaf was being left standing, because no container is ever the field that changed. An array with `uniqueItems` kept reporting duplicates after one of them was edited away:

```
main      errorSchema {} , on screen []
this PR   errorSchema { tags: { __errors: ["must NOT have duplicate items ..."] } } , still on screen
```

The value of every container along the path changed together with the field, so their own errors now go with it, and only their own: the other fields they hold did not change and keep theirs. Fixed with a test.

packages/core/test/Form.test.tsx
luantaraschi: That shape is what `mergeObjects` does with an `undefined` value, and the line this PR replaced produced it as well, one level higher up the tree:

```js
mergeObjects({ foo: { __errors: ['a'] }, baz: { __errors: ['b'] } }, { foo: undefined }, 'preventDuplicates');
// keys ["foo","baz"] , 'foo' in result === true , value undefined
```

So the sentinel was already reaching `Form.state.errorSchema` and the `onChange` event before this PR, at the top-level key instead of at the leaf. I agree it is worth cleaning up and `ErrorSchemaBuilder.clearErrors` looks like the right tool for it, but it changes the shape for every path that gets cleared rather than only the ones this PR adds, so I would rather not fold it in here.

CHANGELOG.md
luantaraschi: Resolved on this side. #5204 opened a `# 6.9.0` heading, so both entries went into the `@rjsf/utils` and `@rjsf/core` lists already sitting under it and my `# 6.8.1` heading is gone. One heading, nothing dropped.

packages/core/src/components/Form.tsx
luantaraschi: #5204 landed the primitive: `setByPath(obj, path, value, true)` is the `Object` customizer under another name, and `ErrorSchemaBuilder` and this reduce both go through it now.

Building the patch from `FieldPathList` instead is the half that would not have worked. `toErrorSchema` flattens a name holding a dot into a path, so a segment list would address `errorSchema['has.dot']`, which is never written, while the dotted string matches the error schema's own addressing. Measurements in the thread on the guard.

luantaraschi:
Merged main in to clear the conflict. The only file that collided was the changelog: the new per-package sections under 6.9.0 landed around my two entries, so they moved back under `@rjsf/utils` and `@rjsf/core` where they belong. `Form.test.tsx` merged on its own and the source is untouched by any of it.

Green here after rebuilding `utils` and `snapshot-tests` first: utils 1519 and core 2260.

heath-freenome:
## Review: re-check of previous findings + fresh pass

I re-fetched the current diff, pulled the branch locally, ran the actual test suites, and manually traced the merge logic rather than trusting the PR thread's summary. **All 7 findings from the previous review round are verifiably resolved**, and I found no new issues.

### Verification of previous findings

| # | Previous finding | Status | Verification |
|---|---|---|---|
| 1 | Dot-guard didn't prevent `_setWith` re-splitting a literal `.` key | **Fixed** | Guard removed entirely (`getChangedFields.ts`). Confirmed via `toErrorSchema`'s own behavior — a property named `has.dot` is written to `errorSchema.has.dot` by the validator, so descending and clearing at that same path is *correct*, not a collision. New test `should only clear the error of the field that changed under a name containing a dot` passes. |
| 2 | Bracket notation (`[`) had the same hazard, unguarded | **Fixed** | Same root fix as #1 — no guard needed once the addressing matches `toErrorSchema`'s. |
| 3 | `diffFields` bypassed the guard inconsistently | **Fixed** | Moot — guard removed, so `diffFields` and `unequalFields` now take the same path. |
| 4 | Same-length array reorder could clear the wrong item's error | **Addressed (design, not bug)** | Confirmed empirically: errors are path-addressed, not identity-addressed, so an error sitting at an unchanged path is still valid for whatever value is now there. Investigating this *did* surface a real bug — container-owned errors (e.g., array `uniqueItems`) were never cleared — which is now fixed via the `for (let i = 1; i < pathOfField.length; i++) setByPath(acc, [...pathOfField.slice(0, i), ERRORS_KEY], undefined, true)` loop in `Form.tsx`. Traced this by hand against the `tags: ['a','a']` uniqueItems test and it's correct. |
| 5 | `undefined` sentinel keys reach public `errorSchema` | **Correctly deferred** | Pre-existing `mergeObjects` behavior, not a regression from this PR — reasonable to leave out of scope. |
| 6 | Changelog heading collision with #5201/#5204 | **Fixed** | Single `# 6.9.0` heading in current `CHANGELOG.md`, both entries present, no duplicate. |
| 7 | `setWith(..., Object)` idiom re-derived in 3 places | **Fixed** | #5204's `setByPath(obj, path, value, true)` landed and is now the shared primitive used here and in `ErrorSchemaBuilder`. |

### Empirical verification (not just reading the diff)

- Checked out this branch, built `@rjsf/utils` and `@rjsf/validator-ajv8`.
- `packages/utils` tests: **1504 passed**, **100% branch coverage** — matches the numbers quoted above exactly.
- `packages/core` tests: **2257 passed** (2 more than quoted above — from an unrelated snapshot addition already on `main`, not from this PR).
- `tsc --noEmit` on `packages/core`: clean.
- Manually traced the container-clearing loop against the `baz.0.qux` and `tags` uniqueItems scenarios — the addressing is correct.

### Remaining observations (not blockers)

- The lazy `getFormDataChangedFields: () => string[]` parameter is a bit unusual as an API shape for `getStateFromProps`, but it's well-justified by the earlier perf finding (avoids the deep walk when `liveValidate` will validate anyway) and is documented with a clear why-comment.
- The `for` loop starting at `i = 1` (skipping the leaf's own container check when `pathOfField.length === 1`) is correct but non-obvious; the surrounding comment explains it adequately.

**Conclusion:** No outstanding issues. The fixes for all previously-raised findings hold up under independent verification (fresh checkout, fresh test run, manual trace) — this looks ready to merge.

heath-freenome:
@luantaraschi Looks like you'll need to rebase and deal with conflicts

luantaraschi:
Done, main is in. The collision was in the source this time: main rebuilt `getChangedFields` around `Object.keys` and a `Set` for `diffFields`, so the `deep` descent hangs off that walk now instead of the `Object.entries` one it was written against. The one test that used `cloneDeep` uses the `makeComplexObject()` factory the file already had.

utils 1527 with branches at 100%, core 2260.

heath-freenome (approved):

PR #5207, opened 2026-08-21
ci: add size-limit bundle size check with base-branch comparison on PRs

jimmycallin:

Bundle size regressions are currently invisible in CI — e.g. the missing `sideEffects` declaration fixed in #5206 shipped ~34kB gzip of unused AJV to every consumer without anything flagging it. This PR adds a [size-limit](https://github.com/ai/size-limit) check so size changes are measured, budgeted, and visible on every PR.

### What it measures

Eight checks (esbuild, minified + brotli) in `.size-limit.json`, following the tier layout [redux-toolkit](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/.size-limit.cjs) uses: every entry point is measured **both with its dependencies and again with them ignored**, plus single-export tree-shaking canaries.

| Check | Current | Budget |
|---|---:|---:|
| 1. `@rjsf/core` | 84.99 kB | 92 kB |
| 2. `@rjsf/core` (without dependencies) | 22.26 kB | 24 kB |
| 3. `@rjsf/core: Form` | 49.28 kB | 53 kB |
| 4. `@rjsf/utils` | 31.52 kB | 34 kB |
| 5. `@rjsf/utils` (without dependencies) | 17.34 kB | 19 kB |
| 6. `@rjsf/utils: getUiOptions` | 0.31 kB | 1 kB |
| 7. `@rjsf/validator-ajv8` | 36.38 kB | 39 kB |
| 8. `@rjsf/validator-ajv8` (without dependencies) | 2.62 kB | 3 kB |

The pairing is the point — the two styles catch different failures, and neither catches the other's:

- **With/without dependencies.** 74% of core's surface and 93% of `validator-ajv8` are third-party. An AJV upgrade moves checks 1 and 7 while 2 and 8 stay flat; an rjsf code regression does the reverse. A single number conflates them.
- **Tree-shaking canaries** (checks 3, 6). A full-surface check is structurally blind to tree-shaking: measured across all 15 packages, toggling `sideEffects` moved the full-surface number by 0.0% everywhere, while a single-export import of `@rjsf/utils` moved 11,974 B → 308 B.

Rebased on top of #5212, which declared `sideEffects: false` on `@rjsf/utils`. That took check 6 from 11.97 kB to 308 B — a 39x reduction — so its budget is now 1 kB rather than 13 kB. That tight budget is what makes it a canary: a tree-shaking regression in `@rjsf/utils` trips it instead of hiding under headroom. Every other check is within 0.1% of its pre-rebase value.

### How it runs

Two workflows, split so the size report comment also works on PRs from forks (a plain `pull_request` job from a fork only gets a read-only `GITHUB_TOKEN` and cannot comment):

- **`.github/workflows/size-limit.yml`** (`pull_request`, read-only token — safe to build untrusted PR code): builds and measures the PR head, then the merge base, generates a markdown diff report (`.size-limit/report.mjs`), uploads it as an artifact, and fails the check if any budget is exceeded. Budget enforcement lives entirely in this job. If the base branch predates the size-limit setup (as `main` does while this PR is open), the baseline is reported as empty instead of failing.
- **`.github/workflows/size-limit-comment.yml`** (`workflow_run`, write token — never executes PR code): downloads the artifact and posts/updates a single sticky size-report comment. Since artifact contents are produced by untrusted PR code, it validates them before use: the claimed PR number must resolve to a PR whose head SHA matches the run that produced the artifact, and the report must match the expected format.

Note: `workflow_run` workflows execute from the default branch's definition, so the comment job takes effect for PRs opened after this merges; on this PR itself only the measuring check runs.

This split goes beyond size-limit's own CI recipe, which is a plain `on: pull_request` job using `andresz1/size-limit-action` — that action has had no commits since June 2024 and has no fork story, since GitHub makes `GITHUB_TOKEN` read-only on fork PRs regardless of declared `permissions`. TanStack Query uses the upstream recipe and simply accepts that fork PRs get no size comment; rjsf takes more fork contributions, hence the split.

### Verified

- `pnpm exec size-limit` run locally against freshly built `lib/` output — all eight checks pass with the sizes above; over-budget checks exit non-zero while still emitting the JSON the report is built from.
- No shim fixture: size-limit synthesizes its own entry file from the `import` option, so the earlier `.size-limit/form-only.js` was unnecessary and is deleted. Core's `Form` is a default export, so the syntax is the unbraced `"import": "Form"` — `"{ Form }"` fails to build.

## Checklist

- [x] I'm updating documentation (changelog entry added)
- [ ] I'm adding or updating tests


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ should change the heading of the (upcoming) version to include a major version b
 ## Dev / docs / playground
 
 - Added `@rjsf/validator-cfworker` to the playground validator selector and API documentation
+- Added a `size-limit` CI check that measures the tree-shaken bundle size of importing only `Form` from `@rjsf/core` (with and without `@rjsf/utils`) plus the full `@rjsf/core` surface, enforces size budgets, and comments the size diff versus the base branch on every PR
```
heath-freenome: Can you move this to 6.8.1

.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,56 @@
...
+          # JSON; record the outcome and enforce it at the end so the report
+          # comment is produced either way.
+          if pnpm exec size-limit --json > "$RUNNER_TEMP/head.json"; then
+            echo ok > "$RUNNER_TEMP/status"
+          else
+            echo fail > "$RUNNER_TEMP/status"
+          fi
+      - name: Measure base branch
+        run: |
+          git fetch --no-tags --depth 1 origin ${{ github.event.pull_request.base.sha }}
```
heath-freenome: **Correctness (confirmed):** `git fetch` and `git checkout -f` here sit *outside* the `if [ -f .size-limit.json ] && pnpm install && ...; then ... else ...; fi` fallback guard below. GitHub Actions runs multi-line `run:` blocks with `bash -e` by default, so if `git fetch` fails (network blip, or the base SHA becoming unreachable after a force-push rewrites the base branch's history), the script aborts immediately — before the guard that's specifically designed to degrade gracefully to `base.json = '[]'` ever runs. Since `actions/upload-artifact@v7` below has no `if: always()`, no artifact gets produced at all in that case, so the downstream comment workflow silently no-ops and the PR gets no size report — contradicting this step's own comment ("report no baseline in any of those cases") and the workflow's stated goal ("the report comment is produced either way").

Suggest wrapping the fetch/checkout in the same guard, or giving `upload-artifact` `if: always()`.

.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/checkout@v7
+      - uses: pnpm/action-setup@v6
+        with:
+          run_install: false
+      - uses: actions/setup-node@v7
+        with:
+          node-version: 22
+          cache: 'pnpm'
+          cache-dependency-path: 'pnpm-lock.yaml'
+      - name: Measure PR head
```
heath-freenome, line 26: **Correctness (plausible, currently dormant):** This step is missing the `--skip-nx-cache` flag that the "Measure base branch" step below adds, with a comment explaining exactly why: `lib/` isn't a declared nx build output, so a cache hit skips the actual build without restoring `lib/` to disk. That reasoning applies identically here. I confirmed against `nx.json` that `lib/` genuinely isn't declared as an output, and that no remote/Nx Cloud caching is currently configured anywhere in the repo — so today this is safe (every run is a cold local cache miss on a fresh runner). But the moment persistent/remote nx caching gets enabled for this workflow (a natural CI-speed follow-up), a cache hit on this exact step would report success while leaving `packages/core/lib/index.js` unwritten, and `.size-limit/form-only.js`'s import would fail — silently reintroducing on the head side the exact bug class this PR just fixed on the base side.

Suggest adding `--skip-nx-cache` here too for consistency/defense-in-depth.

.size-limit/report.mjs
```diff
@@ -0,0 +1,53 @@
...
+const safe = (name) =>
+  String(name)
+    .replace(/[^\w ()@/+.,:-]/g, '')
+    .slice(0, 80);
+// size-limit budgets ("90 kB") are decimal via bytes-iec, so format with 1000.
+const fmt = (bytes) => `${(bytes / 1000).toFixed(2)} kB`;
+const rows = head.slice(0, 20).map((c) => {
+  const b = base.get(c.name);
+  const diff = b ? c.size - b.size : null;
+  const delta = diff === null ? 'new' : diff === 0 ? '=' : `${diff > 0 ? '+' : '-'}${fmt(Math.abs(diff))}`;
```
heath-freenome: **Correctness (confirmed):** Only an *exact* `diff === 0` renders as `=`; any nonzero byte diff whose absolute value rounds to 0.00 kB at 2 decimals (i.e. any 1–9 byte change) still goes through the sign branch. A 3-byte decrease computes `diff = -3` → `diff === 0` is false → `diff > 0` is false so sign is `-` → `fmt(3)` = `"0.00 kB"` → renders as `"-0.00 kB"`. A PR that shaves off a handful of bytes (pure noise) shows a confusing signed-but-zero delta instead of `=`, misleading reviewers into thinking there was a directional change.

Suggest something like `Math.abs(diff) < 5 ? '=' : ...` or rounding before the zero-check, if a tolerance is desired.

.github/workflows/size-limit-comment.yml
```diff
@@ -0,0 +1,65 @@
+# Posts the size report produced by the "Size limit" workflow as a PR comment.
```
heath-freenome, line 1: **Correctness (plausible):** Unlike `size-limit.yml`, this workflow has no `concurrency:` group. `listComments` → conditionally `createComment`/`updateComment` below (lines ~53–64) is a check-then-act sequence: if two `workflow_run` "completed" events for the same PR fire close together — e.g. a maintainer clicking "Re-run all jobs" on the Size limit check twice within the ~10–30s this job takes — both invocations can list comments before either creates one, both see no existing marker comment, and both call `createComment`, leaving a duplicate stale comment that future runs never touch again (they only ever update whichever one `.find()` picks first).

Suggest adding a `concurrency: group: size-limit-comment-${{ github.event.workflow_run.head_sha }}` (or similar) here too.

heath-freenome, line 1: **Design/altitude (worth discussing, not necessarily wrong):** Enforcement lives only in the read-only `pull_request` job (`test "$(cat status)" = ok` in size-limit.yml); this write-token job never fails even when a budget is exceeded — it only annotates rows with `:x:` in the posted markdown. The PR description discloses this, but nothing in the workflow guards against the easy misconfiguration of pointing branch protection at "Size limit comment" instead of "Size limit / size", which would silently disable enforcement while still showing a green required check.

Might be worth a one-line README/CONTRIBUTING note on which check name to require, since the comment job is the more visually prominent of the two on a PR.

.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,56 @@
+name: Size limit
```
heath-freenome, line 6: **Design/altitude (worth discussing):** This reinvents `andresz1/size-limit-action` (maintained by the size-limit authors for exactly this base-vs-head-diff + PR-comment use case) across ~120 lines of custom YAML/JS. That said, the custom fork-safety split here (read-only measuring job, separate write-token job that never executes PR-controlled code, PR derived from the trusted `workflow_run.head_sha` via the API rather than from artifact contents) is a genuinely stronger security posture than that action's simpler default pattern typically provides — so this is a real tradeoff worth a comment/note for future maintainers on why it's custom, rather than a clear mistake.

.size-limit/form-only.js
```diff
@@ -0,0 +1,5 @@
+// Entry for the size-limit checks: what a consumer bundles when importing only Form.
```
heath-freenome: **Maintainability:** This shim only re-exports `Form`, disconnected from `@rjsf/core`'s actual `exports` map (which also exposes `withTheme`, `getDefaultRegistry`, etc. from the same entry point). If a future change moves a heavy dependency behind a different export, or a new named export accidentally pulls in something large, this guardrail won't catch it — it only covers the one import path someone remembered to hand-write here. Not blocking, just worth a comment noting this needs to be kept in sync by hand as the public API evolves, or scoped in the PR description as a known limitation.

.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,56 @@
...
+          else
+            echo '[]' > "$RUNNER_TEMP/base.json"
+          fi
+      - uses: actions/upload-artifact@v7
+        with:
+          name: size-limit-report
+          path: |
+            ${{ runner.temp }}/head.json
+            ${{ runner.temp }}/base.json
+      - name: Enforce limits
```
heath-freenome, line 65: **Simplification (optional):** The status-file dance (write `ok`/`fail` here, read it back in the separate "Enforce limits" step) seems like more indirection than needed to achieve "always upload the artifact, but still fail the job on budget overrun." Letting `pnpm exec size-limit --json > ...` fail the step naturally, and adding `if: always()` to the `upload-artifact` step, would get the same behavior with one fewer step and no cross-step file handoff to reason about (and no risk of a silent typo like `= ok` vs `= Ok` breaking enforcement undetected).

heath-freenome (commented):
Automated review (8-angle finder pass + verification). 9 findings below, most-severe first. Two are confirmed correctness bugs, the rest are plausible gaps/design tradeoffs worth a look. Note: I specifically tried to break the untrusted-artifact / write-token trust boundary between the two workflows and it holds up — PR-number derivation, artifact JSON handling, and comment sanitization all look correctly defended.

.github/workflows/size-limit.yml
heath-freenome, line 26: **Efficiency (optional):** This job runs `pnpm install` + a full `nx build` twice, serially (head then base), on one runner — and the base build explicitly opts out of nx caching (`--skip-nx-cache`), which removes the one benefit that would justify keeping both measurements in the same job/runner. Splitting into two parallel jobs (each with its own fresh checkout at the right ref) would run the two install+build passes concurrently instead of back-to-back, meaningfully cutting wall-clock on a check that runs on every PR push, at the cost of one extra runner spin-up + pnpm-cache restore.

CHANGELOG.md
jimmycallin: Moved to a new 6.8.1 heading in eaa56b52.

.github/workflows/size-limit.yml
jimmycallin: Fixed in eaa56b52 — the `git fetch`/`git checkout` are now the first links of the guard's `&&` chain, so an unfetchable base sha degrades to `base.json = '[]'` like the other baseline failures, and the artifact still uploads. I left `upload-artifact` without `if: always()` deliberately: after this fix the only failures that can precede it are genuine head build/install failures, where there is no head.json and a report would be junk — the comment job already skips quietly when no artifact exists.

.github/workflows/size-limit.yml
jimmycallin: Added in eaa56b52, with the lib/-not-a-declared-output rationale moved onto the head step where it now applies to both.

.size-limit/report.mjs
jimmycallin: Fixed in eaa56b52 — reproduced first (`diff = -3` rendered `-0.00 kB`). Any `|diff| < 5` bytes now renders `=`, since 5 bytes is the smallest value that formats to a nonzero 0.01 kB at two decimals; verified with a synthetic head/base pair that a 3-byte diff shows `=` and a 10-byte diff shows `+0.01 kB`.

.github/workflows/size-limit-comment.yml
jimmycallin: Added in eaa56b52, though with a static `group: size-limit-comment` rather than keying on `head_sha`: a per-sha group still races when two runs for *different* shas of the same PR overlap (push during a re-run), which is the same duplicate-comment outcome. The job is a few seconds of API calls, so serializing all comment jobs repo-wide costs nothing and closes the race completely. No `cancel-in-progress`, so the newest run still lands its update after queueing.

jimmycallin: Noted in eaa56b52 as comments in both workflow files — a header on size-limit.yml saying branch protection must require "Size limit / size", and a matching line in size-limit-comment.yml's header saying it must not be the required one. Put it in the YAML rather than CONTRIBUTING since that's what a maintainer configuring branch protection will actually have open.

.github/workflows/size-limit.yml
jimmycallin: Added a two-line note in eaa56b52 at the top of size-limit.yml: hand-rolled instead of andresz1/size-limit-action specifically for the fork-safety split (read-only job builds untrusted code, write-token job never executes it), which the action's single-job pattern can't express.

.size-limit/form-only.js
jimmycallin: Agreed it's a known limitation — documented in eaa56b52 as a comment in the shim: it guards only the Form import path, and a new export growing a heavy dependency needs its own entry in .size-limit.json. Auto-generating a shim from the exports map would defeat the point of the check (measuring the *tree-shaken* cost of the one import consumers actually make), so hand-maintained entries per guarded path is the intended model.

.github/workflows/size-limit.yml
jimmycallin: Keeping the status file, for one reason the `if: always()` variant can't reproduce: it distinguishes budget-exceeded from build-broken. With natural step failure + `if: always()`, the base-measure and upload steps would also run after a genuine head build/install failure, producing and posting a report from a stale or missing head.json. With the status file, a real failure kills the step before `status` is written and nothing uploads, while a budget overrun still measures the base and posts the full diff. On the typo risk: a broken comparison fails closed (`test` returns nonzero), so a typo would block every PR loudly rather than silently disable enforcement.

.github/workflows/size-limit.yml
jimmycallin: Leaving this serial for now. Two jobs would double the pnpm-install + runner-setup overhead that dominates the wall clock, and the base build is usually the cheap path (it exits at the `.size-limit.json` existence check for the common case today, and shares the runner's warm pnpm store otherwise). If the check turns out to be the long pole on PRs after it's been running on real traffic, splitting is a straightforward follow-up.

CHANGELOG.md
heath-freenome: This section should be gone with the rebase

.size-limit/form-only.js
heath-freenome: Got, perhaps adding a comment to the index.ts to inform developers (and claude) that `.size-limit.json` also needs updating?

CHANGELOG.md
jimmycallin: Gone in b5b774c2

.size-limit/form-only.js
jimmycallin: Added in b5b774c2 above the named-export block in `packages/core/src/index.ts`:

```ts
// The bundle size of the `Form`-only import path is guarded by the size-limit check.
// An export added here that pulls in a heavy dependency goes unmeasured until it gets
// its own check in .size-limit.json.
```

heath-freenome: Looks like this commit was lost somewhere along the way

jimmycallin: Removed deliberately in ce79a07f4, see https://github.com/rjsf-team/react-jsonschema-form/pull/5207#issuecomment-5436102848

Short version: check 1 now measures core's entire entry point rather than just the `Form` path, so a new export pulling in a heavy dependency is caught automatically — which made the note untrue rather than merely redundant. The shim went because size-limit generates its own entry from the `import` option.

jimmycallin:
### Rethought the check structure to widen coverage

The original pair of checks both measured a `Form`-only import of `@rjsf/core`. That left two gaps: nothing covered `@rjsf/utils` or `@rjsf/validator-ajv8` at all, and a dependency growing was indistinguishable from rjsf's own code growing.

The config now follows the tier layout [redux-toolkit](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/.size-limit.cjs) uses — each entry point measured both with its dependencies and again with them ignored, plus single-export tree-shaking canaries. Eight checks, over three packages instead of one.

Two things that buys:

**Full-surface coverage.** Checks 1, 4 and 7 are bare paths with no `import` field, so they measure the entire entry point — every export in the `exports` map, dependencies included. An export that pulls in something heavy moves the number without anyone having to remember to add a check for it. Concretely: `getTestRegistry` reaches the AJV chain, and **35.33 kB** of check 1 is that chain (85.09 kB, dropping to 49.76 kB with `ajv`/`ajv-formats` ignored) — none of which a `Form`-only check at 49.28 kB can see.

**Dependency weight vs. own code.** 74% of core's surface and 93% of `validator-ajv8` are third-party. An AJV upgrade moves checks 1 and 7 while 2 and 8 stay flat; an rjsf code regression does the reverse. A single number per package conflates the two.

The `.size-limit/form-only.js` fixture is gone as part of this: size-limit synthesizes its own entry file from the `import` option, so a shim was never needed. For core's default export the syntax is the unbraced `"import": "Form"` — `"{ Form }"` fails to build.

One thing stays a manual decision: whether a given export deserves its own canary (checks 3 and 6). Only the full-surface coverage became automatic.

heath-freenome:
@jimmycallin I will approve this now. If you merge anymore PRs that shrink the size of the bundles, can you make sure to update the budgets in those PRs? Thanks!

heath-freenome (approved):

PR #5210, opened 2026-08-23
refactor: remove the last lodash usage and drop the dependency

jimmycallin:

Follow-up to #5204 (path utilities) and the type-guard/deep-equality PRs before it: this removes the **last** `lodash` usage from the repository and drops `lodash`/`lodash-es` from every remaining `package.json` — `@rjsf/utils`, `@rjsf/core`, `@rjsf/validator-ajv8`, `@rjsf/validator-ata` and `@rjsf/playground`. With this merged, no RJSF package depends on lodash.

## What was done

- **Native replacements for the remaining helpers**: `isEmpty`, `isObject`, `isPlainObject`, `each`, `flatten`, `includes`, `intersection`, `last`, `difference`, `times`, `uniqueId`, `noop`, `isEqualWith`, `isFunction`. Each call site was replaced with the narrowest native form for what it actually needed rather than a like-for-like shim — e.g. `isEmpty(enumOptions)` → `enumOptions.length === 0` on a value that is always an array, `intersection(a, b).length > 0` → `a.some((e) => b.includes(e))`, `last(items)` → `items.at(-1)`.
- **Three new `@rjsf/utils` exports**, all needed by the replacements and useful to consumers:
  - `isPlainObject()` — the stricter counterpart to `isObject()`, excluding class instances (`Error`, `Map`, …). It narrows to `Record<string, unknown>`, not `Record<string, any>`, so guarded access actually typechecks instead of silently becoming `any`.
  - `getPropertySchema(schema, property)` — reads a property sub-schema out of a schema's `properties`, defaulting to an empty schema. `properties` is typed `JSONSchema7Definition`, which cannot express "the sub-schemas are also `S`", so this function owns that single assertion instead of it being repeated at every lookup.
  - `noop()` — a do-nothing function replacing `lodash/noop`, so no module has to declare its own.
- **`lodashReplacer` deleted**: the `tsc-alias` machinery that rewrote `lodash` → `lodash-es` for the ESM builds is gone, along with the now-empty `tsconfig.replacer.json` files and the vestigial `tsc-alias` block in `packages/core/src/tsconfig.json`. `antd` and `validator-ajv8` keep their replacers for `antdIcons`/`ajv`.
- **Typing strengthened rather than cast**: 31 type assertions removed from the touched files, 1 added (the one inside `getPropertySchema`). `getChangedFields` now relies on `isPlainObject`'s type predicate instead of three `as object` casts; `unwrapErrorHandler` and `toErrorList` recurse through a private function typed on the plain-object shape, so neither the entry point nor the recursive call needs an assertion; a redundant `formData as GenericObjectType` in `createErrorHandler` is gone. No `any` is introduced anywhere in the diff.
- `@rjsf/utils` retains 100% test coverage; tests added for `isPlainObject` and `getPropertySchema`.

## Impact

**Bundle size** — minimal app entry (`import x from '<pkg>'`), react/react-dom/jsx-runtime external, esbuild 0.28.1 and rolldown 1.1.5, minified, gzip level 9. Every milestone below was checked out, reinstalled and rebuilt from source with `--skip-nx-cache`, then bundled by the identical script, so the whole table is directly comparable.

This PR, against current `main`:

| bundle | main (esbuild / rolldown) | this PR | Δ |
|---|---|---|---|
| `@rjsf/core` (core + utils) | 83.1 / 80.1 kB | **77.5 / 75.3 kB** | **−5.6 / −4.8 kB** |
| `@rjsf/validator-ajv8` | 63.8 / 60.2 kB | 61.2 / 57.9 kB | −2.6 / −2.3 kB |
| `@rjsf/validator-ata` | 76.7 / 72.9 kB | 74.1 / 70.6 kB | −2.6 / −2.3 kB |

**Cumulative across the whole effort.** The Form bundle (`import Form from '@rjsf/core'`), milestone by milestone:

| milestone | esbuild | Δ | rolldown | Δ |
|---|---|---|---|---|
| v6.8.0 (before the effort) | 129.1 kB | — | 121.7 kB | — |
| #5191 (`cloneDeep`/`omit`) | 128.0 kB | −1.1 kB | 120.8 kB | −0.9 kB |
| #5193 (shadcn), #5192 (semantic-ui) | 128.0 kB | −0.0 kB | 120.7 kB | −0.0 kB |
| #5190 (deep-equality chain) | 124.6 kB | −3.4 kB | 117.8 kB | −2.9 kB |
| #5189 (native type guards) | 124.6 kB | −0.0 kB | 117.8 kB | −0.0 kB |
| #5206 (`sideEffects: false`) | 84.0 kB | −40.6 kB | 81.0 kB | −36.8 kB |
| #5204 (path utilities) = `main` | 83.1 kB | −0.9 kB | 80.1 kB | −0.9 kB |
| **this PR (last of lodash)** | **77.5 kB** | **−5.6 kB** | **75.3 kB** | **−4.8 kB** |
| **total** | | **−51.6 kB (−39.9%)** | | **−46.4 kB (−38.1%)** |

The `#5193`/`#5192`/`#5189` rows are ~0 for this bundle because those PRs targeted the theme packages; their effect shows up in the theme table below. `#5206` is the outlier: declaring `sideEffects: false` is what finally let bundlers drop the unused AJV chain, and the lodash removals are what kept shrinking the remainder afterwards.

Same measurement across the other published bundles, v6.8.0 → this PR:

| bundle | v6.8.0 (esbuild / rolldown) | this PR | Δ |
|---|---|---|---|
| `@rjsf/validator-ajv8` | 71.5 / 66.8 kB | 61.2 / 57.9 kB | −10.3 / −8.9 kB |
| `@rjsf/validator-ata` | 92.5 / 85.0 kB | 74.1 / 70.6 kB | −18.4 / −14.4 kB |
| `@rjsf/antd` | 338.2 / 358.3 kB | 286.0 / 311.6 kB | −52.2 / −46.8 kB |
| `@rjsf/semantic-ui` | 167.6 / 197.3 kB | 117.5 / 152.1 kB | −50.1 / −45.2 kB |
| `@rjsf/shadcn` | 171.9 / 203.4 kB | 120.3 / 156.8 kB | −51.6 / −46.6 kB |

Theme bundles include their whole UI library, so the percentages are smaller, but each sheds roughly the same ~50 kB gzipped that the Form bundle does.

**Behavior** — no breaking changes. The public API change is purely additive (two new `@rjsf/utils` exports; nothing removed or renamed). Every replacement was checked against lodash's semantics; the differences are confined to inputs that are already malformed, and in each case the new behavior is the same or better:

- lodash's `isObject` counts arrays, functions and `Date`/`File` as objects; RJSF's does not. Every swapped site either sits behind an explicit `Array.isArray` branch (`findSchemaDefinition`, `withIdRefPrefix`) or operates on a value that can only be a schema. One consequence is a latent fix: `makeAllReferencesAbsolute` used to spread a nested array into an object (`{...[1,2]}` → `{0:1,1:2}`); it now leaves it alone.
- `isFormDataAvailable(fn)` returns `true` where it previously returned `false` — a function is not valid form data in the first place, and every other input (including `Date` and `File`) is unchanged.
- A non-object `ui:layoutGrid` (e.g. an array) is now ignored instead of being partially destructured.
- `toErrorList` skips a non-array `__errors` instead of throwing on it.
- The one packaging note: `lodash` is no longer installed as a transitive dependency of any RJSF package, so an app that was relying on hoisting to get lodash into its own `node_modules` will need to declare it directly.

## Checklist

- [x] I'm updating documentation (JSDoc on all new utilities, `utility-functions.md` entries, CHANGELOG under 6.9.0)
- [x] I'm adding or updating code (tests added for the new utilities; `pnpm run lint`, full build and all 17 package test suites pass, `@rjsf/utils` at 100% coverage)


packages/core/src/components/fields/LayoutMultiSchemaField.tsx
```diff
@@ -23,11 +23,12 @@ import {
...
   PROPERTIES_KEY,
   getTemplate,
+  getPropertySchema,
   getUiOptions,
   getWidget,
 } from '@rjsf/utils';
-import isEmpty from 'lodash/isEmpty';
-import noop from 'lodash/noop';
+
+const noop = () => {};
```
heath-freenome: Can we add this to the `utils` library so that this PR remains DRY?

packages/utils/test/parser/ParserValidator.test.ts
```diff
@@ -27,7 +26,7 @@ describe('ParserValidator', () => {
   let consoleErrorSpy: MockInstance;
   beforeAll(() => {
     validator = new ParserValidator(RECURSIVE_REF);
-    consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(noop);
+    consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
```
heath-freenome: If you've followed my suggestion, use the DRY'd up noop function here instead

packages/utils/test/schema/getDefaultFormStateTest.ts
```diff
@@ -21,11 +20,36 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
   describe('getDefaultFormState()', () => {
     let consoleWarnSpy: MockInstance;
     beforeAll(() => {
-      consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(noop); // mock this to avoid actually warning in the tests
+      consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {}); // mock this to avoid actually warning in the tests
```
heath-freenome: Ditto

packages/utils/test/schema/getFirstMatchingOptionTest.ts
```diff
@@ -171,7 +169,7 @@ export default function getFirstMatchingOptionTest(testValidator: TestValidatorT
 
     // simple in the sense of getOptionMatchingSimpleDiscriminator
     it('should return Bar when schema has non-simple discriminator for bar', () => {
-      const consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(noop);
+      const consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
```
heath-freenome: ditto

packages/core/src/components/fields/LayoutMultiSchemaField.tsx
jimmycallin: Done in b40a480dd — `@rjsf/utils` now exports `noop()` (with JSDoc, a docs entry and a CHANGELOG line), and all 27 local copies across `core`, `utils`, both validators and the playground import it instead. This was the last of them.

packages/utils/test/parser/ParserValidator.test.ts
jimmycallin: Done in b40a480dd — now `mockImplementation(noop)` using the shared `noop` from `@rjsf/utils`.

packages/utils/test/schema/getDefaultFormStateTest.ts
jimmycallin: Done in b40a480dd — now `mockImplementation(noop)` using the shared `noop` from `@rjsf/utils`.

packages/utils/test/schema/getFirstMatchingOptionTest.ts
jimmycallin: Done in b40a480dd — now `mockImplementation(noop)` using the shared `noop` from `@rjsf/utils`.

heath-freenome:
## Review

**+859 / −664** across 101 files. Final PR in the lodash-removal series — replaces the remaining non-path lodash helpers with native equivalents / new `@rjsf/utils` exports (`isPlainObject`, `getPropertySchema`, `noop`), deletes the `lodashReplacer` tsc-alias machinery, and drops the dependency from every `package.json` that declared it.

### Correctness — verified against original lodash semantics

Traced the non-mechanical replacements (not straight 1:1 swaps) against lodash's actual behavior:

- **`getChangedFields`** — rewritten to use a `Set` instead of `lodash.difference`; logic is equivalent, arguably clearer.
- **`conditionMatches`** (`LayoutGridField.tsx`) — `intersection(a,b).length > 0/=== 0` → `a.some(e => b.includes(e))` — equivalent for a 2-array intersection test.
- **`isFormDataAvailable(fn)`** now returns `true` for functions (previously `false`, since lodash's `isEmpty(fn)` is `true`). Called out explicitly in the PR description as intentional — a function was never valid form data to begin with.
- **`ArrayField`'s `schemaItems`**: old code guarded with `isObject(schema.items)`, which lodash also returns `true` for a *single* schema object (not just tuple arrays) — meaning a non-tuple `items` schema could previously have been silently mis-cast to `S[]`. The new `Array.isArray(schema.items)` guard is strictly more correct.
- **`toErrorList`** — now checks `Array.isArray(errors)` before mapping instead of a non-null assertion on `errorSchema[ERRORS_KEY]`, so a malformed non-array `__errors` is skipped instead of throwing. Documented in the PR body.
- **`getPropertySchema`** call sites — checked each one where the property key isn't guaranteed to come from `Object.keys(schema.properties)` (e.g. `sanitizeDataForNewSchema.ts`, `ensureFormDataMatchingSchema`, `getObjectDefaults`); in every case the `?? {}` fallback is either unreachable (key always exists) or matches the prior fallback behavior — no silent regression.
- **`setByPath(newErrorSchema, [i ± 1], …)` → `setByPath(newErrorSchema, i ± 1, …)`** in `ArrayField.tsx` — confirmed against `pathUtils.ts`: a non-array path takes the single-key fast path, functionally identical to a one-element array path.
- **`generateRowId`/`getTestIds`** — swapped `lodash.uniqueId`'s shared global counter for a local module-scoped counter. Functionally equivalent since each was only ever called with one fixed prefix.

No behavior regressions found beyond the ones already called out in the PR description (functions-as-formData, non-object `ui:layoutGrid`, non-array `__errors`, `makeAllReferencesAbsolute` array-spread fix) — all improvements on malformed/edge-case inputs, not the happy path.

### Code quality & style

- `getPropertySchema()` consolidates a repeated `(schema[PROPERTIES_KEY]?.[key] ?? {}) as S` cast (previously duplicated across ~6 call sites with a lossy assertion that could hide `undefined`) into one place.
- `isPlainObject()` correctly excludes arrays and class instances (verified via `Object.getPrototypeOf`).
- Type-safety improved incidentally: 31 assertions removed vs. 1 added; `isPlainObject`'s type predicate lets `createErrorHandler`, `unwrapErrorHandler`, and `toErrorList` drop casts in favor of real narrowing.
- Some prettier reformatting is intermixed with logic changes (e.g. `LayoutGridField.test.tsx` reflows unrelated object literals). Likely automatic fallout from import changes, not deliberate — harmless but inflates the diff. Not blocking.

### Test coverage

New dedicated test files for `getPropertySchema`, `isPlainObject`, and `noop`. `@rjsf/utils` retains 100% coverage per the PR description; existing suites updated in place where they used lodash directly.

### Risks / things to double check

- **Transitive dependents relying on hoisting**: any downstream app importing `lodash` without declaring it (relying on hoisting via RJSF's now-removed dependency) will break. Already flagged in the "Impact" section — worth a callout in release notes given it's a behavior change even if not technically a public API break.
- **Module-scoped ID counters** (`rowIdCounter`, `testIdCounter`): reset per module instance, so two separate copies of `@rjsf/core` on the same page (version mismatch, microfrontend) could theoretically produce duplicate IDs. No worse than the previous lodash-based behavior (same failure mode existed with duplicate lodash copies) — not a regression, just noting it's not newly fixed either.

### Verdict

Well-scoped, well-verified refactor with real, measured bundle-size payoff (~5.6 kB gzip off `@rjsf/core`, ~39% cumulative across the whole lodash-removal effort) and no correctness regressions found. Good to merge from a correctness standpoint.

---

heath-freenome:
> * **Transitive dependents relying on hoisting**: any downstream app importing `lodash` without declaring it (relying on hoisting via RJSF's now-removed dependency) will break. Already flagged in the "Impact" section — worth a callout in release notes given it's a behavior change even if not technically a public API break.

@jimmycallin You'll need to rebase and resolve conflicts.  Normally we don't do this, but due to this above mentioned potentially breaking change for people who didn't explicitly import lodash and instead relied on RJSF doing so for them, I would add a note in the CHANGELOG.md directly under 6.9.0 mentioning this `"potentially breaking change"`

jimmycallin:
Rebased onto `main` (through #5212) and added the callout.

**Conflicts resolved:**

- `CHANGELOG.md` — reordering/additions from #5170, #5200, #5206, #5211, #5212; kept all of main's entries and re-inserted this PR's under the right package headings.
- `packages/utils/src/schema/omitExtraData.ts` — #5200 rewrote the permissive-schema early return to `target ?? (useSourceAsFallback ? source : undefined)`. Kept that, with this PR's native `isEmpty` replacement: `schemaDef === true || !schemaDef || Object.keys(schemaDef).length === 0`. The `!schemaDef` arm only catches falsy non-`false` values, for which `isEmpty` was also `true`, so it's equivalent.
- `packages/validator-ata/package.json` — took #5211's `ata-validator@^1.7.1` and dropped `lodash`/`lodash-es`.
- `pnpm-lock.yaml` — regenerated from main's lockfile.

**Potentially breaking change note** — added directly under the `# 6.9.0` heading as requested:

> **Potentially breaking change:** `lodash` and `lodash-es` are no longer dependencies of any `@rjsf/*` package. No RJSF public API changed, but an application that imports `lodash` itself without declaring it — relying on it being hoisted into `node_modules` because RJSF depended on it — must now add `lodash` to its own `package.json`.

**Verified post-rebase:** `pnpm run build --skip-nx-cache` and `pnpm test` (17 projects) both green, `@rjsf/utils` still at 100% coverage, `pnpm run lint` clean, and no `lodash` import or `package.json` entry left anywhere in `packages/`. One pre-existing `cs-check` failure on `packages/chakra-ui/src/ChakraFrameProvider.tsx` is inherited from `main` (last touched by #5175) and untouched by this PR.

heath-freenome (approved):

PR #5216, opened 2026-08-27
fix(utils): support memo widgets via structural check, dropping react-is

jimmycallin:

Partially addresses #4907 — this removes the `react-is` usage (including the `isForwardRef()` call that issue notes is broken on React 19). The `mergeWidgetOptions()`/`defaultProps` removal from that issue is a v7 breaking change and is intentionally not part of this PR.

`getWidget()` rejected `memo()` widgets with `Unsupported widget definition`. It relied on `react-is`, whose predicates only inspect *elements* — so `ReactIs.isMemo(bareMemoComponent)` always returned `false`. The pinned `react-is@18` also doesn't recognize React 19 elements (renamed `$$typeof` tag), which silently breaks the `forwardRef` detection on React 19 too.

There were two gates rejecting wrapper widgets: `getWidget()`'s react-is checks, and `getUiOptions()`, which mistook any object-valued `ui:widget` (including a memo component) for the deprecated `{component, options}` form and dropped it — so wrapper widgets passed via `ui:widget` never even reached `getWidget()`. A new Form-level integration test caught the second gate.

Instead of enumerating wrapper kinds, both now share a structural check (`isWidgetComponent()`, newly exported): every React wrapper (`memo`, `forwardRef`, `lazy`, future ones) is an object tagged with a `$$typeof` symbol, while `isValidElement` filters out a mistakenly passed element. This:

- fixes `memo()` and `memo(forwardRef())` widgets on React 18 and 19, and `forwardRef` widgets on React 19
- adds `lazy()` widget support, and future wrapper types work without changes
- keeps the existing error messages that `hasWidget()` depends on
- removes the `react-is` and `@types/react-is` dependencies, shrinking `@rjsf/utils` (esbuild, minified, react external):

  | | minified | gzip -9 | brotli -11 |
  |---|---|---|---|
  | before | 104,489 B | 37,791 B | 33,717 B |
  | after | 102,195 B | 37,018 B | 33,024 B |
  | **saving** | **2,294 B (2.2%)** | **773 B** | **693 B** |

The existing memo test passed vacuously (it never called `getWidget`); it now exercises the real path, with new tests for memo(forwardRef), lazy, and element rejection.

## Checklist

- [x] I've verified the tests pass: `@rjsf/utils` 1512 tests (100% coverage), `@rjsf/core` 2238 tests


jimmycallin:
@heath-freenome just found your https://github.com/rjsf-team/react-jsonschema-form/issues/4907 I'm not sure if this fix makes sense once your issue has been fully implemented. Happy to close this if you think it should be fixed fully instead in a breaking change.

heath-freenome:
@jimmycallin If the only reason to add the code you did was to replace react-is and I'm planning to remove it in v7 (which I hope to start in September) then maybe we don't need this change?

jimmycallin:
@heath-freenome i believe you're right! closing

PR #5218, opened 2026-08-27
fix(utils): Use defaults when toggling between oneOf/anyOf schemas

### LINKED ISSUE #3736: Form values don't get reinitialized from defaults when toggling between anyOf/oneOf options
### Prerequisites

### What theme are you using?

core

### Version

5.x

### Current Behavior

When one of the options from anyOf/oneOf is rendered, the default values get set properly. However, after you switch to a different option, then go back, all the defaults will be gone.

### Expected Behavior

The defaults should get set every time someone changes the subschema version

### Steps To Reproduce

1. Go to https://rjsf-team.github.io/react-jsonschema-form/
2. Click the "Any Of" example
3. Notice that "First method of identification" is selected, and "First name" is set to the default value, "Chuck"
4. In the render view, change to "Second method of identification"
5. In the render view, change to "First method of identification"

Current: "First name" is empty
Expected: "First name" is set to the default value, "Chuck"

### Environment

```markdown
- OS: Windows 11
- Node & npm: whatever you use on your playground
```

### Anything else?

image

sophstad:

fixes #3736 

This change has caused us some trouble when upgrading from v4 to v6 (😅). This should restore the previous behavior where defaults are used when a field does not exist in all of the schemas within anyOf/oneOf.

Thanks a lot in advance for your help! My team is a big fan of RJSF.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.9.1
+
+## @rjsf/core
+
+- Fixed defaults not being restored when returning to an `anyOf` or `oneOf` option with disjoint properties ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
+
+## @rjsf/utils
+
+- Fixed `sanitizeDataForNewSchema()` clearing existing arrays or preserving stale `undefined` values instead of retaining data or applying defaults for properties newly defined by the incoming schema ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
+
 # 6.9.0
 
 > **Potentially breaking change:** `lodash` and `lodash-es` are no longer dependencies of any `@rjsf/*` package. No RJSF public API changed, but an application that imports `lodash` itself without declaring it — relying on it being hoisted into `node_modules` because RJSF depended on it — must now add `lodash` to its own `package.json`.
@@ -153,12 +163,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/core
 
-- Fixed defaults not being restored when returning to an `anyOf` or `oneOf` option with disjoint properties ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
 - Fixed `TimeWidget` to respect schema `multipleOf` time precision, preserving second-precision values while keeping minute-precision native inputs synchronized with their displayed value ([#5174](https://github.com/rjsf-team/react-jsonschema-form/pull/5174))
 
 ## @rjsf/utils
 
-- Fixed `sanitizeDataForNewSchema()` clearing existing arrays or preserving stale `undefined` values instead of retaining data or applying defaults for properties newly defined by the incoming schema ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
 - Fixed an issue in `sortedJSONStringify` with regards to handling arrays containing `null` and `undefined` values.
 
 ## @rjsf/validator-ata
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,10 +153,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/core
 
+- Fixed defaults not being restored when returning to an `anyOf` or `oneOf` option with disjoint properties ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
 - Fixed `TimeWidget` to respect schema `multipleOf` time precision, preserving second-precision values while keeping minute-precision native inputs synchronized with their displayed value ([#5174](https://github.com/rjsf-team/react-jsonschema-form/pull/5174))
 
 ## @rjsf/utils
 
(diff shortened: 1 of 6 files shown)


heath-freenome:
## Review

### Overview
Fixes #3736: when toggling between `anyOf`/`oneOf` branches (or generally moving to a schema with new properties), a property that's newly introduced by the new schema should get its `default` applied rather than being left/reset to `undefined`. The mechanism is a new `isNewProperty` flag in `sanitizeDataForNewSchema` that (a) lets a new property with `formValue === undefined` pick up the new schema's `default`, and (b) stops the function from recursing into a newly-introduced **array** property (which previously caused it to be wiped to `undefined`, since the old side of the recursion had no matching type).

### Verification performed
- Checked out the branch (already rebased onto current `main`, including the just-merged `getPropertySchema` refactor — mergeable cleanly).
- `packages/utils`: **1522 passed**, **100%** branch/line/statement/function coverage.
- `packages/core`: **2243 passed**, no snapshot drift.
- `tsc --noEmit` clean on both packages.
- Manually traced the modified branch against all 5 new unit tests plus the 3 new integration tests (`Form.test.tsx`, `anyOf.test.tsx`, `oneOf.test.tsx`) — logic matches expectations in each case.

### Finding: array data for a newly-introduced property skips item sanitization (regression)

**Confirmed empirically**, not just by reading:

```ts
const oldSchema = { type: 'object', properties: { idCode: { type: 'string' } } };
const newSchema = {
  type: 'object',
  properties: {
    idCode: { type: 'string' },
    tags: { type: 'array', default: [], items: { type: 'string', enum: ['a', 'b'] } },
  },
};
sanitizeDataForNewSchema(validator, newSchema, newSchema, oldSchema, { idCode: undefined, tags: ['a', 'x'] });
// before this PR: { idCode: undefined, tags: ['a'] }   -- 'x' filtered as invalid
// after this PR:  { idCode: undefined, tags: ['a', 'x'] } -- 'x' survives, unfiltered
```

The new guard `(!isNewProperty && newSchemaTypeForKey === 'array' && Array.isArray(formValue))` in `sanitizeDataForNewSchema.ts` skips the recursive call entirely whenever the array property is new — not just when `formValue` needs a fresh default. That recursive call is what does enum filtering for array items and recurses into object items. So for a newly-introduced array property that already carries stale data (e.g., leftover from a sibling `anyOf`/`oneOf` branch, or `omitExtraData: false`), invalid items are no longer stripped and nested object items are no longer sanitized — they pass through unchanged via the final `...data` spread.

This is a narrow edge case (it only bites when a "new" property already has non-default array content), but it's a real behavior change from current `main`, and it's the kind of thing the surrounding code was explicitly designed to prevent (see `enumValuesForSchema`/`replacementForInvalidEnumValue` a few lines below). Worth a second look before merge — possibly recurse as before but only fall back to preserving `formValue` when the recursive result would otherwise be `undefined`, rather than skipping recursion outright.

### Minor, out-of-scope observation
The same `isNewProperty` treatment was applied only to the `default` branch, not the `const` branch just below it. A newly-introduced property with only a `const` (no `default`) and `formValue === undefined` stays `undefined` instead of being initialized to the `const` — verified with a scratch test. This mirrors the bug being fixed but for `const` instead of `default`. Not required for this PR (issue #3736 is specifically about defaults) but worth a follow-up.

### Other notes
- The JSDoc fix (`{ ...data, ...removeOldSchemaData, ...nestedData }`) is a welcome correction — it now matches what the code actually does (it didn't before this PR either, unrelated pre-existing doc drift).
- Tests are well-targeted: `Form.test.tsx`/`anyOf.test.tsx`/`oneOf.test.tsx` each cover the reported bug at the integration level, and `sanitizeDataForNewSchemaTest.ts` isolates the four relevant unit-level cases (new+undefined, new+already-present array, existing+undefined explicit, and a regression pin for old-schema-missing-type). Good discipline pinning the pre-existing quirky "missing type wipes array" behavior so this change doesn't accidentally alter it.
- Changelog entries are accurate and correctly scoped to `@rjsf/core`/`@rjsf/utils`.

**Recommendation:** Solid fix for the reported issue with good test coverage, but the array-filtering regression above should be addressed or explicitly accepted before merging.

heath-freenome:
@sophstad The Form.test.tsx file was split up into pieces. Can you figure out which of the split up files it belongs in?

CHANGELOG.md
```diff
@@ -120,10 +120,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 ## @rjsf/core
 
+- Fixed defaults not being restored when returning to an `anyOf` or `oneOf` option with disjoint properties ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
```
heath-freenome: Move this to 6.9.1

CHANGELOG.md
```diff
@@ -120,10 +120,12 @@ should change the heading of the (upcoming) version to include a major version b
...
 - Fixed `TimeWidget` to respect schema `multipleOf` time precision, preserving second-precision values while keeping minute-precision native inputs synchronized with their displayed value ([#5174](https://github.com/rjsf-team/react-jsonschema-form/pull/5174))
 
 ## @rjsf/utils
 
+- Fixed `sanitizeDataForNewSchema()` clearing existing arrays or preserving stale `undefined` values instead of retaining data or applying defaults for properties newly defined by the incoming schema ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
```
heath-freenome: Move this to 6.9.1

heath-freenome (approved):

PR #5219, opened 2026-08-27
test: de-duplicate and shard @rjsf/core's slowest test suites

jimmycallin:

I've started working on some test modernization and cleanup, with a focus on test performance. This is the first PR in that effort, with more to follow.

This one targets the two biggest time sinks in `@rjsf/core`'s test suite:

**1. The "Form common" suite ran 4× serially inside one file.** `Form.test.tsx` contained a ~3.7k-line shared suite executed once per `omitExtraData`/`liveOmit` variant, all inside a single test file — so vitest couldn't parallelize any of it. That suite is now split by topic into six normal test files (`Form.{rendering,defaults,handlers,state,errors,props}.test.tsx`), each still run under all four variants via the existing `describeRepeated` helper — same coverage matrix, but vitest can now spread the work across workers, and each file is a readable size.

**2. The StringField suite was silently re-executed twice more.** `ArrayField.test.tsx` and `ObjectField.test.tsx` imported `StringField.test.tsx` as a module, which re-registered and re-ran its entire suite inside theirs — 302 duplicate test executions per run that looked like extra coverage but weren't. The genuinely shared cases now live in a `TextWidgetTest.tsx` helper that each suite invokes for its own scenario, and the duplication is gone.

Before/after on my machine (`pnpm test` in `packages/core`, best of 3):

|  | tests | wall time |
|---|---|---|
| before | 2260 | ~10.2s |
| after | 1958 | ~5.4s |

Same assertions, same coverage — just without running 302 of them twice and with better parallelism.

## Checklist

- [x] I'm updating documentation — CHANGELOG.md entry added

## How to review the sharding

The diff is big but almost entirely moved lines. Two mechanical checks make it reviewable:

**1. The test inventory is unchanged.** `vitest list` dumps every test's full name; after stripping file paths and the topic suffix added to the `describeRepeated` titles, the sorted unique inventories on `main` and this branch are byte-identical (1902 names each):

```bash
git checkout main && cd packages/core && pnpm vitest list | sed -E 's/^[^>]+> //' | sort -u > /tmp/before.txt
git checkout - && pnpm vitest list | sed -E 's/^[^>]+> //; s/^Form common: [a-z ]+ \{/Form common \{/' | sort -u > /tmp/after.txt
diff /tmp/before.txt /tmp/after.txt   # empty
```

(The raw counts differ — 2240 vs 1938 — because the 302 cross-import duplicates on `main` produce *identical* full names, which `sort -u` collapses. That's the duplication this PR removes.)

**2. The moved code is verbatim.** Review the diff with git's moved-line detection, which greys out lines that moved unchanged and highlights anything that didn't:

```bash
git diff main...HEAD --color-moved=dimmed-zebra -- packages/core/test
```

What remains highlighted is only the expected scaffolding: each new file's import header, the `describeRepeated('Form common: <topic>', ...)` wrappers, and the `TextWidgetTest` extraction.


packages/core/test/Form.test.tsx
heath-freenome, line 1: Might I suggest, renaming the `FormCommon.*.test.tsx` to just be `Form*.test.tsx` since they are all refactored out of `Form.test.tsx`. And then, in another, simple PR, rename `Form.test.tsx` to be another `Form.xxx.test.tsx` (where `xxx` is a representive name for the contents) so that they all follow the pattern while keeping the history of `Form.test.tsx` -> `Form.xxx.test.tsx` in git

heath-freenome:
## Review

### Overview
Pure test-infrastructure PR: splits the monolithic `Form.test.tsx` "Form common" suite (run 4× via `describeRepeated`) into six topic-scoped files so vitest can parallelize them, and replaces `ArrayField.test.tsx`/`ObjectField.test.tsx`'s cross-import of `StringField.test.tsx` (which silently re-ran its whole suite twice more) with a shared `TextWidgetTest.tsx` helper. No production code (`src/`) touched.

### Verification performed (independent, not just trusting the PR's own numbers)
- Checked out the branch — already rebased on latest `main`, mergeable.
- **Test inventory equivalence**: ran the recipe from the PR description myself rather than trusting it. `pnpm vitest list` on `main`: 2240 raw / 1882 unique names. On this branch: 1938 raw / 1882 unique names. `diff` between the two normalized, sorted-unique inventories is **empty** — byte-identical test coverage. The raw-count delta (2240 → 1938 = exactly 302) matches the claimed duplicate-execution removal precisely.
- Full `packages/core` suite: **1938 passed**, 0 failures, ~8s locally (vs. `main`'s ~14s in an earlier session on this same machine) — parallelism gain is real.
- `tsc --noEmit` on `packages/core`: clean.
- `pnpm run lint`: clean (only pre-existing, unrelated warnings in other packages).

### Finding: one hunk is not actually a mechanical move, despite the PR's framing
The PR body explicitly recommends `git diff --color-moved=dimmed-zebra` to confirm "the moved code is verbatim," and mostly that holds. But in the `Nested forms > should call provided submit handler with form state` test (moved from `Form.test.tsx` into `FormCommon.props.test.tsx`), the implementation actually changed:

```diff
-import { Portal } from 'react-portal';
...
-        return (
-          <Portal>
-            <div className='array' ref={innerRef}>...</div>
-          </Portal>
-        );
+import { createPortal } from 'react-dom';
...
+        return createPortal(
+          <div className='array' ref={innerRef}>...</div>,
+          document.body,
+        );
```

This is a good change on its own merits — `react-portal` isn't declared as a dependency in any `package.json` in the monorepo (it only resolved via lockfile hoisting as a phantom transitive dependency), whereas `createPortal` is the standard `react-dom` API and renders to the same `document.body` target, so it's behaviorally equivalent (confirmed: the test passes). But it's a deliberate rewrite sitting inside a diff otherwise presented as "almost entirely moved lines," and someone using only the suggested moved-line color trick to sanity-check the PR could plausibly skim past it as scaffolding. Worth calling out explicitly in the PR description/commit rather than leaving it for a reviewer to find via the same tool the PR suggests using to *confirm nothing changed*.

### Other minor notes
- `testUtils.tsx`: the `CreatorFn` type alias (previously needing an `// oxlint-disable-next-line no-unused-vars` pragma) was inlined into the `describeRepeated` signature, removing the suppression comment — a clean, small improvement.
- Changelog entry is accurate and appropriately scoped.

**Recommendation:** Solid, well-verified refactor — the test-inventory-equivalence check holds up independently. No blockers; flag the `Portal`→`createPortal` hunk as a called-out (not hidden) change before merging, since it's the one place the PR's own "just trust the move-detection diff" review method could mislead a reviewer.

heath-freenome:
## Follow-up: verifying no tests were lost in the move

Went a level deeper than the name-set comparison from my earlier review, since set equality alone can hide a real deletion if it happens to be masked by an unrelated duplicate. This time I compared **occurrence counts per test name** (a full multiset diff) between `main` and the current PR head (`2eddd4f9`, the rebased commit):

- **151 test names** changed frequency — every single one is a `StringField > ...`-prefixed test, and every single one goes **3 → 1** (previously executed once directly plus twice more via the `ArrayField`/`ObjectField` cross-imports; now once).
- **Zero** non-`StringField` tests changed count in either direction.
- The total reduction across all 151 sums to **exactly 302** — matching the PR's own duplicate-removal claim to the test.
- No `.only`, `.skip`, `xit`, or `xdescribe` introduced anywhere in the touched files.
- Re-ran the moved-line detection (`--color-moved=dimmed-zebra`) on the current head: same 113 added / 48 removed non-moved lines as in my previous pass — no additional test body was silently rewritten under an unchanged title beyond what was already flagged (imports, `describeRepeated` topic titles, and the `Portal`→`createPortal` swap noted in my earlier comment).
- Full `packages/core` suite: 1938/1938 passing on this exact head.

**Conclusion: no tests were lost in the move.** Every test that disappeared from the raw count was a genuine duplicate of a `StringField` test previously executed via cross-import, not a real test. Nothing new to flag beyond the `Portal`/`createPortal` hunk already called out.

packages/core/test/Form.test.tsx
jimmycallin: Done — renamed the six files to the `Form.<topic>.test.tsx` pattern in 069ce05. Left `Form.test.tsx` itself untouched here as suggested, so a follow-up PR can rename it with its git history intact.

heath-freenome (approved):

PR #5225, opened 2026-08-28
Fixes related to empty values (#5222)

### LINKED ISSUE #5222: Inconsistent empty string behaviour under additionalProperties
### Prerequisites

### Version

Live Version on the playground today

### Current Behavior

When I have the same object subschema under the root, as well as under an additionalProperties, a non-required string value in that subschema is treated differently.
In the root, an empty value omits the field in the output (as I deem correct), while inside the additionalProperty, an empty string is emitted instead.

This also happens if the subschema is a `$ref`. It also happens if the subschema is not an `object`, but directly a `string`, in which case it makes sense (as otherwise the key would directly disappear), and that might be the cause of this bug.

### Expected Behavior

I expect the forms that look the same to behave the same. In particular I would like to rely on the fact that an empty string will omit the field in the json.

### Steps To Reproduce

Please see this [playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJ0aGVfb25lX3Byb2plY3QiOnt9LCJwcm9qZWN0cyI6eyJFeGFtcGxlIjp7Im5hbWUiOiIifX19LCJzY2hlbWEiOnsidGl0bGUiOiJDb25maWciLCJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJ0aGVfb25lX3Byb2plY3QiOnsidHlwZSI6Im9iamVjdCIsInByb3BlcnRpZXMiOnsibmFtZSI6eyJ0eXBlIjoic3RyaW5nIn19fSwicHJvamVjdHMiOnsidHlwZSI6Im9iamVjdCIsImFkZGl0aW9uYWxQcm9wZXJ0aWVzIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7Im5hbWUiOnsidHlwZSI6InN0cmluZyJ9fX19fX0sInVpU2NoZW1hIjp7fSwidGhlbWUiOiJkZWZhdWx0IiwibGl2ZVNldHRpbmdzIjp7InNob3dFcnJvckxpc3QiOiJ0b3AiLCJ2YWxpZGF0ZSI6ZmFsc2UsImRpc2FibGVkIjpmYWxzZSwibm9IdG1sNVZhbGlkYXRlIjpmYWxzZSwicmVhZG9ubHkiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlLCJleHBlcmltZW50YWxfY29tcG9uZW50VXBkYXRlU3RyYXRlZ3kiOiJjdXN0b21EZWVwIiwiZXhwZXJpbWVudGFsX2RlZmF1bHRGb3JtU3RhdGVCZWhhdmlvciI6eyJhcnJheU1pbkl0ZW1zIjoicG9wdWxhdGUiLCJhbGxPZiI6InNraXBEZWZhdWx0cyIsImNvbnN0QXNEZWZhdWx0cyI6ImFsd2F5cyIsImVtcHR5T2JqZWN0RmllbGRzIjoicG9wdWxhdGVBbGxEZWZhdWx0cyIsIm1lcmdlRGVmYXVsdHNJbnRvRm9ybURhdGEiOiJ1c2VGb3JtRGF0YUlmUHJlc2VudCIsIm5lc3RlZERlZmF1bHRzUHJlY2VkZW5jZSI6ImRlc2NlbmRhbnRXaW5zIn0sInVzZUZhbGxiYWNrRmllbGQiOmZhbHNlLCJ1c2VGYWxsYmFja1VpRm9yVW5zdXBwb3J0ZWRUeXBlIjpmYWxzZSwibGl2ZVZhbGlkYXRlIjpmYWxzZSwiaWRQcmVmaXgiOiJyb290IiwiaWRTZXBhcmF0b3IiOiJfIn0sInZhbGlkYXRvciI6IkFKVjgiLCJzYW1wbGVOYW1lIjoiU2ltcGxlIn0=) as an example.

### Resolution

I don't know the codebase, nor do I know typescript particularly well, but I could let claude on the task, but manually review and create a PR. Would this be useful? If so, any points?

nickgros:
Fixes #5222

- descendants of `additionalProperties` should use default `onChange` which will respect `ui:emptyValue`
- `ui:allowClearTextInputs` button should respect `ui:emptyValue` (bonus bug)

### Checklist


packages/core/src/components/templates/BaseInputTemplate.tsx
```diff
@@ -73,7 +73,7 @@ export default function BaseInputTemplate<
     (e: MouseEvent) => {
       e.preventDefault();
       e.stopPropagation();
-      onChange(options.emptyValue ?? '');
+      onChange(options.emptyValue);
```
heath-freenome, line 76: Is this pattern anywhere else in the other themes?

heath-freenome: Yes — the identical pattern (`onChange(options.emptyValue ?? '')` in the clear-button handler) is unchanged in every other theme's `BaseInputTemplate`: antd, chakra-ui, daisyui, fluentui-rc, mantine, mui, primereact, react-bootstrap, semantic-ui, shadcn. This PR only fixes `@rjsf/core`, so those 10 packages will still store `''` instead of `ui:emptyValue` on clear-button click, diverging from core's corrected behavior. Worth a fast-follow PR applying the same one-line fix across those templates.

nickgros: Good catch, I fixed all themes and extended CLAUDE.md to guide agents to checking other themes when making or reviewing changes like this in the future.

heath-freenome (approved):

PR #5227, opened 2026-08-29
test: remove setup shims that no longer do anything

jimmycallin:
## Reasons for making this change

Part of the ongoing test modernization and cleanup: while reviewing the per-package vitest setup files, I checked whether each shim in them is still doing anything — by deleting it and running that package's full suite. Five turned out to be dead weight:

- **core**: the `setImmediate` global — the suite passes with an *empty* setup file, so the file and its config entry are removed entirely.
- **antd**: the `MessageChannel` global (once required by React 18 + antd; no longer).
- **chakra-ui**: the `structuredClone` JSON polyfill — `structuredClone` is native since Node 17, and this repo requires Node ≥ 20. (The JSON version was also subtly worse: it drops `undefined`, `Date`, `RegExp`, etc.)
- **mantine**: `cleanSnapshotSerializer` — a no-op against every current snapshot, so the serializer file goes too.
- **primereact**: the `<style>`-injection blocker and its setup file — all snapshots are identical without it.

Equally important, what was verified to still be load-bearing and is deliberately **kept**: the shared `ResizeObserver` mock (63 failures across 10 files without it), antd's and mantine's `mockViewport`/matchMedia mocks (their whole suites fail), chakra's and mui's Emotion snapshot serializers, and mantine's `Math.random` stub (29 snapshot failures).

The benefit: less mystery setup for the next person to puzzle over — every shim that remains provably earns its place.

## Checklist

- [x] I'm updating documentation — CHANGELOG.md entry added
- [x] I'm adding or updating tests — test setup only; all affected suites verified green (core 1958, antd 145, chakra-ui 144, mantine 141, primereact 141)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5228, opened 2026-08-29
test: resolve workspace packages from source via an @rjsf/source export condition

jimmycallin:

Part of the ongoing test modernization work: `git clone && pnpm install && pnpm vitest run` now works with **no build step**, and tests always exercise current TypeScript source instead of potentially stale `lib/` output.

**How it works**

- `@rjsf/utils`, `@rjsf/core`, `@rjsf/validator-ajv8` and `@rjsf/snapshot-tests` declare a custom `@rjsf/source` export condition pointing at their TS source. The namespaced name (rather than a bare `"source"` or `"development"`) means only our packages can ever match it — a third-party dependency declaring its own `source` condition can't get resolved to unshipped files, and consumers' dev servers keep resolving to the published `lib/` output.
- A new shared `testing/vitest.base.ts` opts into the condition via `resolve.conditions` and carries the common vitest defaults (`globals`, jsdom). Every package's vitest config now extends it via `mergeConfig`, declaring only its own `setupFiles`/coverage deltas. The shared setup file lives in the base (resolved against the base's own location); packages keep only their genuinely local setup files and coverage deltas, with the shared `fullCoverage()` gate exported from the base for the four 100%-threshold packages.
- A root `vitest.config.ts` with `projects` lets vitest run across every package from the repo root: `pnpm vitest run`, `pnpm vitest --project @rjsf/core`, repo-wide watch. **The nx-driven `test` scripts are unchanged** — CI keeps running exactly as today.

**Second commit: the validator-ajv8 harness shrinks to one helper**

The precompiled-validator tests used to statically import gitignored `superSchema*.cjs` fixtures, which forced generation to happen before vitest loaded the tests — the reason a `compileSchemas` npm script, a vitest `globalSetup`, and a `pnpm`→`tsx --conditions` subprocess all existed. A small `compileSuperSchema()` helper now compiles the schema entirely in memory *inside* the test context (where the source condition applies) and evaluates the generated standalone module directly — no fixture files, no temp files, no gitignore entries. The file-writing wrapper API stays covered by its own dedicated test (`compileSchemaValidators.test.ts`, fs mocked); two assertions that had become self-comparisons after the fixture removal were dropped. Minus a script, a globalSetup, a subprocess, three stale gitignore entries, and a pair of "keep these in sync" comments; the shared `expectWarn()` helper moved to the harness so the console-spy dance exists once.

**Verified** with every `lib/` and `dist/` deleted: the full root run (196 files / 9628 tests, ~24s) and standalone package runs (validator-ajv8's 2730 with no pre-generated fixtures, mui's 150) all pass from pure source.

## Checklist

- [x] I'm updating documentation — CHANGELOG.md entries added


packages/core/vitest.config.ts
```diff
@@ -1,12 +1,9 @@
...
+
+export default mergeConfig(base, {
   test: {
-    globals: true,
-    environment: 'jsdom',
-    setupFiles: ['./test/testSetup.ts', '../../testing/testSetup.ts'],
-    coverage: {
-      provider: 'v8',
-    },
+    setupFiles: ['./test/testSetup.ts'],
```
heath-freenome: `mergeConfig(base, override)` concatenates `setupFiles` as `[...base, ...override]`, always putting the shared file first. Core's original config explicitly listed `./test/testSetup.ts` **before** `../../testing/testSetup.ts` — that order is now silently reversed.

The PR description says setupFiles stay per-package deliberately because "their load order differs between packages," which suggests the ordering matters and was meant to be preserved. Currently harmless since core's `testSetup.ts` content (`global.setImmediate = setImmediate`) is order-independent, but worth a second look given the stated intent.

testing/vitest.base.ts
```diff
@@ -0,0 +1,20 @@
...
+  // "@rjsf/source" export condition, so tests run against current code with no
+  // prerequisite build. The custom name (rather than "development") keeps
+  // consumers' dev servers on the published lib/ output.
+  resolve: {
+    conditions: ['@rjsf/source'],
+  },
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    setupFiles: ['../../testing/testSetup.ts'],
```
heath-freenome: `'../../testing/testSetup.ts'` is a literal path relative to each *consuming* package's directory, not derived from this base file's own location — it silently assumes every package's `vitest.config.ts` sits exactly two directories below the repo root (`packages/<name>/vitest.config.ts`).

A future package whose vitest config lives somewhere else (nested deeper, or a package that moves) would get a silently wrong/missing setup file. Might be worth deriving this path from `import.meta.url` instead so it's independent of the consumer's location.

packages/utils/vitest.config.ts
```diff
@@ -1,13 +1,11 @@
...
-export default defineConfig({
+import base from '../../testing/vitest.base';
+
+export default mergeConfig(base, {
   test: {
-    globals: true,
-    environment: 'jsdom',
-    setupFiles: ['../../testing/testSetup.ts'],
     exclude: ['node_modules/**', 'lib-test/**'],
     coverage: {
```
heath-freenome: This ~9-line coverage block (`enabled`, `reportsDirectory`, `include`, `exclude`, 100/100/100/100 `thresholds`) is copy-pasted identically across `packages/utils`, `packages/validator-ajv8`, `packages/validator-ata`, and `packages/validator-cfworker`'s `vitest.config.ts` — all touched by this PR, which introduces `testing/vitest.base.ts` specifically to de-duplicate shared config.

Since it's the same in all four, could this move into the base config (or a second shared `coverage.base.ts` extended by these four) so changing the coverage policy doesn't require editing 4 files in lockstep?

packages/validator-ajv8/test/harness/testData.ts
```diff
@@ -1,8 +1,20 @@
+import { noop } from '@rjsf/utils';
 import type Ajv from 'ajv';
 
 import type { CustomValidatorOptionsType } from '../../src';
 
-// NOTE these are the same as the CUSTOM_OPTIONS in `compileTestSchema.js`, keep them in sync
+/** Runs `fn` with console.warn stubbed, asserts it warned with `expectedWarning`, and returns the result. */
+export function expectWarn<T>(fn: () => T, ...expectedWarning: unknown[]): T {
```
heath-freenome, line 7: This `expectWarn()` helper duplicates the same spy/assert/restore idiom already implemented in `packages/core/test/testUtils.tsx` (`setupConsoleWarnSuppression`/`setupConsoleErrorSuppression`). Given this PR adds a root `testing/` directory meant to hold exactly this kind of cross-package test infrastructure, would it make sense to consolidate into one shared helper there rather than having two independent implementations of the same pattern?

packages/validator-ajv8/test/precompiledValidator.test.ts
```diff
@@ -2,20 +2,14 @@ import type { ErrorSchema, FormValidation, RJSFSchema, RJSFValidationError, UiSc
...
-// @ts-ignore file might be not yet generated by test/harness/compileTestSchema.cjs
-import * as superSchemaFns from './harness/superSchema.cjs'; // oxlint-disable-line import/extensions
-import superSchema from './harness/superSchema.json';
-// oxlint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore file might be not yet generated by test/harness/compileTestSchema.cjs
-import * as superSchemaOptionsFns from './harness/superSchemaOptions.cjs'; // oxlint-disable-line import/extensions
+import { SUPER_SCHEMA_OPTIONS, compileSuperSchema, superSchema } from './harness/compileSuperSchema';
 
-const validateFns = superSchemaFns as unknown as ValidatorFunctions;
+const validateFns = compileSuperSchema();
```
heath-freenome, line 9: `compileSuperSchema()` with default options now does a full AJV compile + `new Function` eval at module scope here and again independently in `createPrecompiledValidator.test.ts` — the identical default-options compilation is redone per test file with no cross-file caching, unlike the old fixture-file approach where it happened once via `globalSetup`.

Not a correctness issue and cheap today given `superSchema.json`'s size, but worth flagging since it's a structural duplication that would scale poorly if the schema or option-variant count grows. A `beforeAll`-memoized shared compile (or a small module-level cache in `compileSuperSchema.ts`) could avoid the repeat work if it ever becomes noticeable.

heath-freenome (commented):
Automated review (Claude Code). Verified by checking out this PR into a worktree and running the full test suite from pure source (196 files / 9628 tests, matching the PR's own claim), plus `tsc -b`/`oxlint` on touched files. Six findings below, none blocking — this is a clean, well-scoped test-infra change.

**Not inline (file isn't part of this PR's diff, so GitHub won't let me anchor a line comment to it):**

`nx.json`'s `test` target is cached with no `inputs` covering the new shared `testing/vitest.base.ts` (or root `vitest.config.ts`). Reproduced directly: after a clean `nx test @rjsf/utils`/`@rjsf/core` run, editing `testing/vitest.base.ts` (e.g. `resolve.conditions`) and re-running `nx test <pkg>` or `pnpm test` prints "Nx read the output from the cache instead of running the command" — a stale green result gets served even though shared test behavior changed. Mainly a local warm-cache risk since CI does a fresh checkout, but `pnpm test` (`nx run-many --target=test`) is this repo's documented standard command, so it's worth adding these new shared files to the `test` target's cache inputs.

The other five findings (setupFiles ordering, a relative-path assumption in the new base config, a duplicated coverage-threshold block, a duplicated console-warn-spy helper, and a minor per-file recompilation) are left as inline comments below.

packages/core/vitest.config.ts
jimmycallin: Fixed in f8e9f32 — moot on the current head anyway: #5227 deleted core's and primereact's local setup files entirely, so both configs are now one-line re-exports of the base. Every remaining package with a local setup file (antd, chakra, mantine, mui) already listed the shared file first, so no load order changed anywhere. The PR description's ordering caveat is stale; I've removed it.

testing/vitest.base.ts
jimmycallin: Good catch — fixed in f8e9f32: the base now resolves the setup file against its own location via `fileURLToPath(new URL('./testSetup.ts', import.meta.url))`, so consuming configs can live anywhere.

packages/utils/vitest.config.ts
jimmycallin: Done in f8e9f32 — the base now exports `fullCoverage(extraExcludes)` and the four gated packages use `fullCoverage()` / `fullCoverage(['src/types.ts'])`. Verified the gate still fires (standalone coverage runs report the 100% thresholds) and all four suites pass. Related: I also added `{workspaceRoot}/testing/*` and the root `vitest.config.ts` to nx's `test`/`test:update` cache inputs per your top-level finding — reproduced your stale-cache repro before (cache hit after editing the base) and after (correctly invalidates).

packages/validator-ajv8/test/harness/testData.ts
jimmycallin: I'd keep these separate: they have different semantics — `expectWarn()` is a one-shot wrapper that *asserts* a specific warning fired and restores immediately, while core's `setupConsole*Suppression` helpers install suite-scoped beforeAll/afterAll suppression and hand back the spy. Merging them means one configurable helper doing both jobs, which I think reads worse than two small obvious ones. Happy to consolidate into `testing/` if you'd rather have one implementation, though.

packages/validator-ajv8/test/precompiledValidator.test.ts
jimmycallin: Agreed it's structural duplication, but a module-level cache can't actually help: vitest isolates each test file in its own worker process, so a cache in `compileSuperSchema.ts` never sees a second hit — within one file each variant already compiles exactly once at module scope. The only true cross-file cache is fixture files on disk, which is the machinery this PR deliberately removes (and whose cost was the pnpm+tsx subprocess, not the compile — the compile itself is ~10-20ms). If superSchema ever grows enough to matter, resurrecting a globalSetup that calls `compileSchemaValidators()` directly (no subprocess) is the escape hatch.

CHANGELOG.md
```diff
@@ -128,12 +128,11 @@ should change the heading of the (upcoming) version to include a major version b
...
 - Restructured `@rjsf/core`'s test suite for speed: the "Form common" suite was split by topic into six test files so vitest can parallelize them, and the `StringField` suite is no longer re-executed via cross-imports from the `ArrayField`/`ObjectField` suites (302 duplicate test executions removed)
-- Added `knip` and removed the unused code, exports and dependencies it reported ([#5194](https://github.com/rjsf-team/react-jsonschema-form/pull/5194))
-- Removed test-setup shims that empirically no longer do anything (each verified by deleting it and running the package's suite): core's `setImmediate` global, antd's `MessageChannel` global, chakra-ui's `structuredClone` JSON polyfill (native since Node 17), mantine's `cleanSnapshotSerializer` (a no-op on all current snapshots), and primereact's `<style>`-injection blocker
-- Upgraded the test tooling: `@testing-library/jest-dom` 6→7, `jsdom` 29→30, `@testing-library/user-event` to 14.6.6, `vitest` to 4.1.11, and declared the `@testing-library/dom` peer explicitly at the root. Self-anchored test selectors were rewritten to `:scope` to match jsdom 30's corrected element-scoped `querySelectorAll` behavior
```
heath-freenome, line 136: Did these accidentally get removed in a bad merge?

heath-freenome:
## Review

### Overview
Two related changes, bundled as one PR:
1. **Source-condition resolution**: `@rjsf/utils`, `@rjsf/core`, `@rjsf/validator-ajv8` (both its main export and `./compileSchemaValidators` subpath), and `@rjsf/snapshot-tests` gain a namespaced `@rjsf/source` export condition pointing at `./src/index.ts`. A new shared `testing/vitest.base.ts` opts into it via `resolve.conditions` and centralizes the common vitest defaults (`globals`, jsdom, `setupFiles`, coverage). Every package's `vitest.config.ts` now extends it via `mergeConfig`. A root `vitest.config.ts` adds a `projects` glob so `pnpm vitest run` works repo-wide. Nx-driven `test`/`test:update` scripts are unchanged, but `nx.json` gains explicit `inputs` for `testing/*` and the new root config so a warm nx cache doesn't serve stale results when the shared config changes.
2. **`validator-ajv8` harness simplification**: the precompiled-validator tests no longer write/read gitignored `superSchema*.cjs` fixture files via a `globalSetup` subprocess; a new `compileSuperSchema()` helper compiles and `new Function`-evaluates the AJV standalone code entirely in memory. Deletes the `compileSchemas` script, `globalSetup.ts`, three `.gitignore` files, and two now-redundant "generates the expected output" assertions.

### Verification performed (not just read-through)
- **`mergeConfig` array semantics**: confirmed in Vite's source (`mergeConfigRecursively`) that array-valued keys are concatenated (`[...arraify(existing), ...arraify(value)]`), not overwritten. So e.g. antd's `setupFiles: ['./test/testSetup.ts']` correctly merges with the base's `setupFiles: [testing/testSetup.ts]` into both files running — matching the old explicit `['../../testing/testSetup.ts', './test/testSetup.ts']` behavior exactly. No regression here.
- **Condition containment**: checked `tsconfig.base.json` (no `customConditions`) and the playground/docs build configs (no `resolve.conditions`) — the `@rjsf/source` condition can only ever be requested by the new `testing/vitest.base.ts`. TypeScript's own type resolution and production bundlers are unaffected, exactly as the PR description claims.
- **`files` field consistency**: `packages/snapshot-tests/package.json` gets `src` added to `files` (needed since it previously had none in its `exports` map); `core`/`utils`/`validator-ajv8` already had `src` in `files` prior to this PR, so the pattern is applied consistently.
- **`.gitignore` removals for `validator-ata`/`validator-cfworker`**: verified these packages' `test/harness/` directories contain no fixture-generating script (`package.json` has no `compileSchemas`-equivalent, harness dir only has `superSchema.json` + `testData.ts`) — the removed `.gitignore` entries were already dead weight, not a regression.
- **"Self-comparison" test removal**: read the deleted `compileTestSchema.ts` — the old `expectedCode` was read from a file regenerated immediately beforehand by the same `compileSchemaValidatorsCode()` call via `globalSetup`, so the deleted assertions were comparing the compiler's output against itself, not against a checked-in golden file. Calling this a "self-comparison" in the PR description is accurate; dropping it is a real but very minor loss of a determinism guard, already transparently disclosed.
- **Branch staleness**: `gh pr diff`'s CHANGELOG.md hunk appears to delete three bullets ("Added knip…", "Removed test-setup shims…", "Upgraded the test tooling…") that exist on current `main` — because this branch predates PR #5230 (jsdom 30/jest-dom 7 bump, now merged) and the knip/shims PRs. Checking `gh pr view --json mergeable,mergeStateStatus`: `mergeable: MERGEABLE` (no conflict) and `mergeStateStatus: BLOCKED` (blocked on required review, not conflicts). A clean git 3-way merge should retain all bullets from both sides — the `gh pr diff` rendering is just misleading about the *final* result, not a preview of actual content loss. CI (`build 20.x/22.x/24.x`, `size`) all pass, and CI runs against the merge commit, so the vitest restructuring here is already confirmed compatible with the intervening jsdom 30/jest-dom 7 bump and whatever `testing/testSetup.ts` changes the shims-removal PR made.

### Code quality
- The `mergeConfig(base, {...delta...})` pattern is applied uniformly and correctly across all ~14 packages; packages with zero deltas correctly collapse to a one-line `export { default } from '../../testing/vitest.base'`.
- `fullCoverage(extraExcludes)` cleanly captures the shared 100%-threshold gate while preserving each package's pre-existing exclude list (`utils`/`validator-ajv8` get no extra excludes, `validator-ata`/`validator-cfworker` keep their `src/types.ts` exclude) — checked against the pre-PR configs, no behavior change.
- The namespaced condition name (`@rjsf/source`, not `source`/`development`) is a good defensive choice — it can't collide with a third-party package's own `source` condition and can't accidentally leak into a consumer's dev server resolution.
- `nx.json`'s added `inputs` with the explanatory `"//test"` comment is a genuinely non-obvious but important correctness fix — without it, editing `testing/vitest.base.ts` wouldn't invalidate nx's test cache for any package.
- oxlint-disable comments (`oxlint-disable-next-line no-new-func, typescript/no-implied-eval`) use the project's correct disable-comment style.

### Issues / suggestions
1. **(Low, hygiene) Rebase before merge.** Branch predates #5230 and the knip/test-setup-shims PRs. GitHub reports no conflicts and CI is green against the merge commit, so this isn't blocking — but rebasing would make the CHANGELOG diff legible again and remove reliance on 3-way-merge correctness for that file.
2. **(Very low) Lost determinism guard.** The two deleted "generates the expected output" tests, while genuinely redundant self-comparisons in their old form, did incidentally guard against `compileSchemaValidatorsCode()` producing non-deterministic output across calls (e.g. object key ordering). Not worth blocking on, just noting the tradeoff is real, if small.

### Security
`new Function('exports', 'require', code)(...)` in `compileSuperSchema.ts` evaluates AJV-generated standalone code from a known, local test schema — not external/untrusted input, and confined to test code (never shipped). Consistent with AJV's own documented pattern for standalone-code usage, and consistent with the existing `require(...)` pattern already used elsewhere in this same file (`testData.ts`). No concern.

### Recommendation
Solid, well-scoped refactor with real verification behind its claims (196 files / 9628 tests from clean `lib`/`dist`, confirmed independently via the merge-commit CI run). Safe to merge once the required review/rebase housekeeping is cleared — no functional blockers found.

CHANGELOG.md
jimmycallin: Yes, good catch.

heath-freenome (approved):

PR #5230, opened 2026-08-29
test: upgrade the test tooling (jest-dom 7, jsdom 30, user-event, vitest)

jimmycallin:

Part of the ongoing test modernization work: brings the test tooling to current versions.

- `@testing-library/jest-dom` 6.9 → **7.0** (root, core, daisyui devDeps, and the `@rjsf/snapshot-tests` peer range)
- `jsdom` 29 → **30**
- `@testing-library/user-event` → 14.6.6 (core and shadcn), `vitest`/`@vitest/coverage-v8` → 4.1.11
- `@testing-library/react` → 16.3.3 (fixes an `act()` re-entrancy issue when dispatching events)
- `@testing-library/dom` declared explicitly at the root — jest-dom 7 makes it a required peer
- `pnpm dedupe` run on top, collapsing a few duplicated transitive versions (`picomatch`, `lru-cache`, `@napi-rs/wasm-runtime`)

The one substantive code change: jsdom 30 corrects element-scoped `querySelectorAll` behavior, so self-anchored selectors like `node.querySelectorAll('form .form-group …')` (where `node` *is* the form) no longer match themselves — they're rewritten to `:scope .form-group …` in `SchemaField.test.tsx` plus one selector in `Form.errors.test.tsx`. Same assertions, spec-correct anchoring.

Verified: the full repo suite passes on the new stack (all 17 projects via nx, including every theme's snapshot suite — no snapshot changes needed).

**Heads-up for maintainers:** both majors formally raise their Node floor to 22 (`jest-dom` 7 declares `engines.node >=22`; jsdom 30 dropped its Node 20 compatibility code), while this repo declares `engines.node >=20` and CI tests 20.x. pnpm doesn't enforce `engines`, and the CI matrix is the empirical check — but if Node 20 support matters long-term, holding these two at 6.x/29.x until the repo drops Node 20 is the alternative.

Sequencing note: this owns the lockfile churn, so it shares `pnpm-lock.yaml` and root devDeps lines with #5194 — whichever lands second needs a lockfile regen on rebase. Orthogonal to everything else that's open.

## Checklist

- [x] I'm updating documentation — CHANGELOG.md entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,7 +134,6 @@ should change the heading of the (upcoming) version to include a major version b
 - Added `knip` and removed the unused code, exports and dependencies it reported ([#5194](https://github.com/rjsf-team/react-jsonschema-form/pull/5194))
 - Upgraded the test tooling: `@testing-library/jest-dom` 6→7, `jsdom` 29→30, `@testing-library/user-event` to 14.6.6, `vitest` to 4.1.11, and declared the `@testing-library/dom` peer explicitly at the root. Self-anchored test selectors were rewritten to `:scope` to match jsdom 30's corrected element-scoped `querySelectorAll` behavior
 
-
 # 6.8.0
 
 ## @rjsf/core
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,8 @@ should change the heading of the (upcoming) version to include a major version b
 - Added `size-limit` CI checks that enforce bundle size budgets for `@rjsf/core`, `@rjsf/utils` and `@rjsf/validator-ajv8` — each measured both with and without its runtime dependencies, plus single-export tree-shaking canaries — and comment the size diff versus the base branch on every PR
 - Restructured `@rjsf/core`'s test suite for speed: the "Form common" suite was split by topic into six test files so vitest can parallelize them, and the `StringField` suite is no longer re-executed via cross-imports from the `ArrayField`/`ObjectField` suites (302 duplicate test executions removed)
 - Added `knip` and removed the unused code, exports and dependencies it reported ([#5194](https://github.com/rjsf-team/react-jsonschema-form/pull/5194))
+- Upgraded the test tooling: `@testing-library/jest-dom` 6→7, `jsdom` 29→30, `@testing-library/user-event` to 14.6.6, `vitest` to 4.1.11, and declared the `@testing-library/dom` peer explicitly at the root. Self-anchored test selectors were rewritten to `:scope` to match jsdom 30's corrected element-scoped `querySelectorAll` behavior
+
 
 # 6.8.0
 
--- a/package.json
+++ b/package.json
@@ -30,17 +30,18 @@
   "homepage": "https://github.com/rjsf-team/react-jsonschema-form",
   "devDependencies": {
     "@size-limit/preset-small-lib": "^13.0.3",
-    "@testing-library/jest-dom": "^6.9.1",
-    "@testing-library/react": "^16.3.2",
+    "@testing-library/dom": "^10.4.1",
+    "@testing-library/jest-dom": "^7.0.1",
+    "@testing-library/react": "^16.3.3",
     "@types/node": "^24.13.2",
     "@types/react": "^18.3.31",
     "@types/react-dom": "^18.3.7",
-    "@vitest/coverage-v8": "^4.1.9",
+    "@vitest/coverage-v8": "^4.1.11",
     "cross-env": "^10.1.0",
     "esbuild": "^0.28.1",
     "husky": "^9.1.7",
     "is-ci": "^4.1.0",
-    "jsdom": "^29.1.1",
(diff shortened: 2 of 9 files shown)


heath-freenome (approved):

heath-freenome:
## Review

### Overview
Pure test-infrastructure bump, no runtime/production code touched:
- `@testing-library/jest-dom` 6.9 → 7.0.1, `jsdom` 29 → 30.0.1, `@testing-library/react` 16.3.2 → 16.3.3, `@testing-library/user-event` → 14.6.6, `vitest`/`@vitest/coverage-v8` → 4.1.11, `@testing-library/dom` newly pinned explicitly at root (now a required jest-dom 7 peer).
- One substantive code change: 15 test selectors across `SchemaField.test.tsx` and `Form.errors.test.tsx` rewritten from self-anchored (`form .form-group …`) to `:scope`-anchored (`:scope .form-group …`).
- Rest is `pnpm-lock.yaml` churn from the bump + a `pnpm dedupe` pass.

CI is green on all three Node majors (20.x, 22.x, 24.x), plus lint and size checks.

### Correctness — verified, not just trusted
I pulled the pre-change source for both touched test files to check the `:scope` rewrite by hand, since a mechanical selector rewrite is exactly the kind of change that can silently loosen an assertion:
- In `Form.errors.test.tsx:745`, the old selector `.rjsf-field-string .error-detail li` only "worked" because jsdom ≤29 buggily let the context node satisfy the leading compound selector against itself (`fieldNodes[1]` is itself `.rjsf-field-string`). Jsdom 30 fixes that per spec, so the new `:scope .error-detail li` is the correct replacement — same intent, no longer relying on the bug.
- In `SchemaField.test.tsx`, `node` is the rendered `<form>` from `createFormComponent`, so `form .form-group …` had the identical self-match dependency; `:scope > .form-group …` / `:scope .form-group …` are exact, spec-correct equivalents.

No behavior loosening in either file — this is a clean, correct mechanical fix, not a rubber-stamp rewrite.

### Issues found

**1. CHANGELOG.md will conflict on rebase (minor, but concrete)**
The new bullet is inserted immediately after the "Restructured `@rjsf/core`'s test suite…" line. Current `main` already has a different bullet ("Added `knip`…", PR #5194) inserted at that exact same point, merged after this PR's branch point. The three-way PR diff doesn't show it, but it's on `main` now — so this insertion will conflict on rebase, in addition to the `pnpm-lock.yaml`/root-devDeps conflict with #5194 already flagged in the PR description. Not a blocker, just confirms the rebase will touch two files, not one.

**2. Double blank line introduced before `# 6.8.0`**
The diff adds a trailing blank line after the new bullet, on top of the existing blank line before the version header:
```
+- Upgraded the test tooling: ...
+

# 6.8.0
```
Every other version boundary in the file uses a single blank line. Minor nit, one-line fix.

**3. `libc:` metadata silently dropped for many optional native packages in `pnpm-lock.yaml`**
Packages like `@ata-validator/native-linux-*`, `@nx/nx-linux-*`, `@oxlint/binding-linux-*`, `@parcel/watcher-linux-*`, `lightningcss-linux-*`, `@rollup/rollup-linux-*`, `@tailwindcss/oxide-linux-*` keep the *same version* but lose their `libc: [glibc]` / `libc: [musl]` disambiguation line. `lockfileVersion` and the pinned `packageManager: pnpm@10.17.1` are unchanged, so this isn't an intentional lockfile-format migration — it reads like a side effect of the `pnpm dedupe` pass. Since this field is what lets pnpm pick the right native binary on musl-based systems (e.g. Alpine Docker images), it's worth a sanity check that installs still resolve correctly there before merging — CI here only runs glibc-based `ubuntu-latest` runners, so this wouldn't be caught by the existing matrix either way.

**4. Node-floor tightening (already disclosed, empirically fine)**
The PR description calls this out, and it goes slightly further than stated: `jsdom@30.0.1`'s lockfile entry declares `engines: {node: ^22.22.2 || ^24.15.0 || >=26.0.0}` (not just "22+"), and `jest-dom@7.0.1` declares `>=22`. `pnpm install` doesn't enforce `engines` (no `engine-strict` in `.npmrc`), and the `build (20.x)` job passed, so this isn't blocking today — just flagging that it's tighter than the PR text implies, worth remembering if a future patch of jsdom starts using Node 22.22+-only APIs at runtime rather than just declaring the engine.

### Test coverage
No new tests needed or added — appropriate for a tooling bump. The 15 changed assertions preserve identical intent/coverage; verified above rather than assumed.

### Security
All changed packages are devDependencies (test tooling only); none ship in the published bundles (confirmed by the passing `size` check, which showed no diff). No production security surface.

### Recommendation
Safe to merge. Fix the CHANGELOG double-blank-line nit, and expect a rebase to resolve the CHANGELOG + lockfile conflicts with #5194 (already anticipated by the author). Worth a quick check that the `libc` field disappearance isn't masking a `pnpm dedupe` regression, but it's not release-blocking for this PR.

PR #5232, opened 2026-08-29
build: consolidate the TypeScript configuration

jimmycallin:

The root `tsconfig.build.json`'s compiler options only ever reached `@rjsf/core`. Every other package's `src` project extends `tsconfig.base.json` directly, and a referenced project does not inherit from the config that references it — so for 14 of 15 packages that file was dead. The two bases had drifted apart, and the published output drifted with them:

| | `@rjsf/core` | every other package |
|---|---|---|
| target | `esnext` | `ES2018` |
| declarationMap | yes | no |
| sourceMap | no | yes |

Nobody chose that split. Along the way this also fixes a stale root reference list, a `@rjsf/daisyui` config that emitted build artifacts into `src`, and a published package that node cannot import.

## What this does

**One base.** `tsconfig.build.json` is folded into `tsconfig.base.json`, pinned at `ES2022` (`esnext` makes the emitted output move with every TypeScript upgrade) with `declaration`, `declarationMap` and `sourceMap` on for everyone. Options already implied by `strict` and `esModuleInterop` are dropped. Turning the unused-symbol checks on for `core` surfaced two dead bindings (`PROPERTIES_KEY`, `autofocus`), removed here.

**One test base.** The 15 near-identical test configs are hoisted into a new `tsconfig.test.json`; each package's is now four lines. They had drifted too — only `@rjsf/daisyui`'s re-included `../src` and emitted into `../dist`, and none pulled in the jest-dom matcher types, which had left 27 real type errors invisible.

**Explicit `.js` extensions** on relative imports in `src` and `test`. `tsc` now emits node-loadable ESM on its own, making `tsc-alias --resolveFullPaths` redundant: it drops out of 13 of 15 build scripts. `@rjsf/antd` and `@rjsf/validator-ajv8` keep it for their bare-specifier replacers. The oxlint `import/extensions` rule flips to `"always"` to match.

**`verbatimModuleSyntax`** enabled, with zero fallout — type imports were already marked consistently.

**Root project references** completed: `snapshot-tests` was listed twice, and `mantine`, `primereact`, `validator-ata` and `validator-cfworker` were missing entirely.

**`packages/daisyui/tsconfig.json`** becomes a solution file like the other 14. It had a real `include` and no `outDir`, so `tsc -b` at the root emitted 312 `.js`/`.d.ts`/`.map` files directly into `daisyui/src`. Nothing ran the solution build, so nobody had hit it.

**`"type": "module"` on `@rjsf/snapshot-tests`**, which is published and emits ESM into `lib/` but would be parsed as CommonJS by node.

## Why not full `nodenext`

`moduleResolution: nodenext` is the boring correct setting for packages node imports directly, but it is not reachable yet — it produces 223 errors, none of them ours:

- **118** — `react-bootstrap@2.10.10`'s `esm/*.d.ts` are not valid JSX element types under node ESM resolution
- **13** — `ajv/dist/standalone` is a deep import outside ajv's `exports` map (exactly what the `ajvReplacer` exists for), and ajv 8's `export =` class is not constructable from ESM
- **4** — `@cfworker/json-schema`'s dual package yields two nominally distinct `Validator` classes

The explicit extensions above are the half of that migration that works today under `bundler` resolution, and they are what actually removes `tsc-alias`. The resolution flip can follow when the dependency typings allow it.

## Verification

`pnpm run sanity-check` (lint + knip + build + test) passes, as does `cs-check`. The built ESM was additionally loaded in real node to confirm the extensions are correct without any post-processing:

```
node -e "import('./packages/utils/lib/index.js')"  # 156 exports
node -e "import('./packages/core/lib/index.js')"   # 7 exports
```

`tsc -b tsconfig.json` (the solution build, newly coherent) goes from 61 errors to 4. The remaining 4 are pre-existing source bugs left for separate fixes, listed in the review notes below.

## Checklist

- [x] I'm adding or updating tests (existing suites cover this; no behavior change intended)

---

**For the reviewer — four pre-existing errors this newly surfaces, not fixed here:**

- `playground/src/samples/customObject.tsx:22` — `titleId<T>(...)` passes a type argument to a function that takes none
- `playground/src/samples/customFieldAnyOf.tsx:30` — missing `fieldPathId` on `FieldTemplateProps`
- `chakra-ui/test/WrappedForm.tsx:8` — `EnvironmentProvider` no longer accepts `environment`
- `validator-ata/test/utilsTests/schema.test.ts:37` — `ajvOptionsOverrides` should be `ataOptionsOverrides`, meaning that `discriminator: true` override is silently doing nothing


heath-freenome:
@jimmycallin Anymore PRs coming from you that you'd like to get into the 6.9.0 release?

jimmycallin:
> @jimmycallin Anymore PRs coming from you that you'd like to get into the 6.9.0 release?

no go ahead. currently doing some typescript modernization that will pop up as a set of PRs over the coming days (this pr was supposed to open on my fork for me to review which is why i closed it). i'm ok with it landing in a later release.

PR #5233, opened 2026-08-30
build: import real .ts/.tsx paths in source and drop tsc-alias

jimmycallin:
First step in a TypeScript modernization effort.

## Problem

Node's ESM resolver does not guess extensions: `./foo` fails, only `./foo.js` resolves. Our source is extensionless, so `tsc` emits extensionless imports, and `tsc-alias` then string-replaces them in the emitted files. The imports we publish come from regex post-processing rather than the compiler, including two hand-written replacer plugins compiled to `.cjs` on every build.

## Change

Source imports name the real file. TypeScript's `rewriteRelativeImportExtensions` rewrites them on emit:

```ts
// packages/core/src/index.ts
import Form from './components/Form.tsx';

// packages/core/lib/index.js — straight from tsc, no post-processing
import Form from './components/Form.js';
```

This removes `tsc-alias`, `tsc-alias-replacer/`, both `tsconfig.replacer.json` files, the `compileReplacer` scripts and `move-file-cli`. `build:ts` is now just `tsc -b tsconfig.build.json`.

## Why `.ts` in source rather than `.js`

- **The compiler checks the import.** A bad relative path is a build error, not something a regex may or may not patch.
- **What you read is what runs.** Jump-to-definition, grep and the emitted output agree.
- **It prepares for `node foo.ts`.** Node's type stripping requires exact `.ts` extensions and does no extension or index lookup. Writing `.js` today would mean a second repo-wide rewrite later.

## Two things the codemod could not do

- `@rjsf/validator-ajv8` imported the directory `ajv/dist/standalone`, and a replacer patched it to `.../index.js` after emit. Source now imports `ajv/dist/standalone/index.js` directly. The `@ant-design/icons` replacer matched nothing and is simply deleted.
- `packages/daisyui/test/tsconfig.json` emitted into `../dist`, the bundle output directory. It now type-checks with `noEmit` like every other test project.

## Review notes

- The rewrite is an AST codemod over static imports/exports, type-only imports/exports, literal `import()` and literal import types. The repo has no computed dynamic imports.
- Test projects set `rewriteRelativeImportExtensions: false` with `allowImportingTsExtensions`. They never emit, and TypeScript refuses to rewrite across a project boundary (TS2878). `packages/utils/test` becomes `emitDeclarationOnly` for the same reason.
- `import/extensions` in `.oxlintrc.json` now requires `.ts`/`.tsx` on relative imports, so the convention is enforced going forward.
- **Caveat:** TypeScript leaves `.ts`/`.tsx` specifiers in emitted `.d.ts` files. It resolves those to `.d.ts` correctly, verified from consumer projects under both `node` and `bundler` resolution.

## Verification

Lint, knip, a clean full build and the full test suite pass. All 806 emitted `.js` and 806 `.d.ts` relative specifiers resolve to existing files.


heath-freenome:
## Review

### Overview
885 files, but almost entirely one mechanical codemod: every relative import/export/type-import in the monorepo gets an explicit `.ts`/`.tsx` extension (`./Form` → `./Form.tsx`), enabled by TypeScript 6's `rewriteRelativeImportExtensions` compiler option, which rewrites those specifiers to `.js` during `tsc`'s emit. This replaces the old `tsc-alias` post-processing pass (plus two hand-written "replacer" plugins) that did the same job via regex over already-emitted files. `build:ts` (the `lib/` pipeline) shrinks to just `tsc -b tsconfig.build.json`.

### Does this affect the UMD/CJS builds?

**No — verified, not assumed.** Each package's actual build scripts confirm `build:cjs`, `build:esm` (the `dist/` bundles), and `build:umd` are byte-for-byte unchanged by this PR:

```
build:ts:  rimraf ./lib && tsc -b tsconfig.build.json && tsc-alias ...   →  rimraf ./lib && tsc -b tsconfig.build.json
build:cjs: esbuild ./src/index.ts --bundle --outfile=dist/index.cjs ...  →  unchanged
build:esm: esbuild ./src/index.ts --bundle --outfile=dist/index.esm.js  →  unchanged
build:umd: rollup dist/index.esm.js --format=umd --file=dist/*.umd.js   →  unchanged
```

The key fact: `build:cjs`/`build:esm`/`build:umd` bundle **straight from `src/`** via esbuild/rollup — they never read `lib/`, never ran `tsc-alias`, and don't care about `rewriteRelativeImportExtensions` (that's a `tsc`-only compiler option; esbuild has its own resolver). This PR only touches the `tsc`→`lib/` pipeline (the `"import"` export-condition target) and the emitted `.d.ts` files.

Confirmed esbuild's resolver treats extensionless and explicit-`.tsx` relative imports identically, by bundling the same two-file example both ways in both `--format=esm` and `--format=cjs`:
```
$ esbuild old-extensionless.tsx --bundle --format=esm --outfile=old.out.js
$ esbuild new-explicit.tsx --bundle --format=esm --outfile=new.out.js
$ diff old.out.js new.out.js
# only difference: esbuild's auto-generated var name (derived from the entry filename) — no functional diff
```
Same result under `--format=cjs`. So the CJS/ESM(dist)/UMD bundles for every package's main entry point should be functionally identical before and after this PR.

### The one place it *does* touch `dist/`
`compileSchemaValidatorsCode.ts` changes a **bare package specifier** (not a relative import, so outside the codemod's stated scope): `import standaloneCode from 'ajv/dist/standalone'` → `'ajv/dist/standalone/index.js'`. Since `esbuild --packages=external` leaves external specifiers untouched in the output, this string change *does* flow verbatim into `dist/compileSchemaValidators.cjs`/`.esm.js`. Checked whether this is actually consequential:

- **CJS (`require`)**: tested both forms against the real installed `ajv` (which has no `exports` map) — both resolve fine, since CJS `require` does directory-index fallback. No behavior change.
- **Real Node ESM `import`**: tested both forms directly — the bare form **hard-fails**: `ERR_UNSUPPORTED_DIR_IMPORT: Directory import '.../ajv/dist/standalone' is not supported... Did you mean to import "ajv/dist/standalone/index.js"?` The explicit form resolves fine.
- **But**: checking `@rjsf/validator-ajv8`'s `exports` map shows there is no `./dist/*.esm.js` (or any) wildcard exposing `.esm.js` files publicly — only `.`, `./compileSchemaValidators`, `./lib`, `./lib/*.js`, `./dist`, and `./dist/*.cjs` are reachable, and all of those route ESM consumers to `lib/`, not `dist/*.esm.js`. So `dist/compileSchemaValidators.esm.js` is unreachable via the package's public contract — it's rollup-UMD-input plumbing only (and there's no UMD build for `compileSchemaValidators` anyway, since `src/index.ts` doesn't re-export it).

Net effect: this fix is safe and correct, but its consumer-facing relevance is essentially nil — it corrects an already-unreachable intermediate file. The `lib/compileSchemaValidators.js` path (the one real ESM consumers hit) was already being fixed correctly by the old `ajvReplacer.ts` tsc-alias plugin; this PR just moves that correctness into source, which is strictly better (no post-processing step to keep in sync) but not a behavior change for consumers either way.

### Other things checked
- **`daisyui/test/tsconfig.json`**: real fix, not just codemod noise — it previously had `outDir: "../dist"` with no `noEmit`, meaning the test project's type-check was emitting compiled JS *into the actual bundle output directory* alongside the real esbuild-built `dist/` artifacts. Now `noEmit: true`, matching every other test project. Worth calling out as an independently valuable bug fix buried in a big mechanical PR.
- **`.oxlintrc.json`**: `import/extensions` flips from forbidding to *requiring* `.ts`/`.tsx` on relative imports, so the new convention is enforced going forward rather than just applied once.
- **Test-project tsconfigs**: correctly get `rewriteRelativeImportExtensions: false` + `allowImportingTsExtensions: true` (test projects don't emit, so rewriting would be meaningless and TS rejects it across the source-project boundary anyway per the PR's TS2878 note).
- **`knip.jsonc`/root `package.json`**: clean removal of the now-dead `tsc-alias-replacer/*.ts` entry and the `move-file-cli`/`tsc-alias` devDependencies.

### Verification gap worth naming
The PR's own "Verification" section covers `lint`, `knip`, a clean build, the test suite, and a relative-specifier resolution check for `lib/`'s 806 files — all real checks, but all either source-level (vitest via the `@rjsf/source` condition) or `lib/`-scoped. None of it actually `require()`/`import`s the produced `dist/*.cjs` / `dist/*.esm.js` / `*.umd.js` files in a bare Node process to confirm the bundles work end-to-end post-build. Given the analysis above concludes those bundles are unaffected, this isn't blocking here — but it's the one category of regression (a genuine `dist/`-bundle break) that none of the stated verification would actually catch, for this or any future PR touching build scripts.

### Recommendation
Sound refactor. Verified directly rather than by inference: **UMD and CJS builds are unaffected** — this PR's blast radius is `lib/` and `.d.ts` output only, with one incidental (and harmless) fix to an already-unreachable `dist/` intermediate. No functional risk found for the published bundles.

CHANGELOG.md
```diff
@@ -123,6 +123,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Replaced all `lodash/get` usage with the new `@rjsf/utils` path utilities
 - Replaced `lodash/isObject` with `isObject` from `@rjsf/utils` and dropped the `lodash`/`lodash-es` dependencies entirely
 - Declared `"sideEffects": false` in `package.json`, allowing bundlers to tree-shake unused exports
+- Changed the `standaloneCode` import to name the file, `ajv/dist/standalone/index.js`, instead of the directory subpath `ajv/dist/standalone`. A `tsc-alias` replacer used to patch this into the emitted output; the source now says what it means. No public API changed
```
heath-freenome: Move these to 6.9.1

CHANGELOG.md
```diff
@@ -148,6 +149,8 @@ should change the heading of the (upcoming) version to include a major version b
...
 - Simplified `@rjsf/validator-ajv8`'s precompiled-validator test harness: the tests now compile `superSchema` in memory via a small `compileSuperSchema()` helper, replacing the generated gitignored `superSchema*.cjs` fixtures and with them the `compileSchemas` npm script and the vitest `globalSetup` that regenerated them
 - Updated the `emptyValue` and `allowClearTextInputs` uiSchema docs to indicate relationship each other.
+- Relative TypeScript imports now name their real `.ts`/`.tsx` source file (and an explicit `index.ts` for directory imports), and TypeScript's `rewriteRelativeImportExtensions` emits the `.js` specifiers directly. This removes `tsc-alias` and its post-emit string rewriting entirely, along with the `tsc-alias-replacer/` directory, both `tsconfig.replacer.json` files, the `compileReplacer` scripts and `move-file-cli`. Naming the source file rather than the output is deliberate: Node's native type stripping requires exact `.ts` extensions and does no extension or directory-index searching, so this avoids a second repository-wide import migration later
```
heath-freenome: Move these to 6.9.1

jimmycallin: done

heath-freenome (approved):

PR #5185, opened 2026-08-09
Fix defaults for single-element allOf-wrapped $ref

nkgotcode:
## Summary

- Populate defaults through a single-element `allOf` that wraps a `$ref`.
- Preserve recursive-reference protection and add focused regression coverage.

## Testing

- `pnpm --filter @rjsf/utils exec vitest run test/schema.test.ts --coverage.enabled=false` — 667 tests passed.
- `pnpm --filter @rjsf/utils run lint`
- `pnpm --filter @rjsf/utils run cs-check`
- `pnpm --filter @rjsf/utils run build`

Fixes #5177

Issue #5146, opened 2026-07-01
`json-schema-merge` update


x0k:
I have released `@x0k/json-schema-merge@1.0.4`:

Symbols are now preserved when merging schemas ([test](https://github.com/x0k/json-schema-merge/blob/52c23e6a7133254cb799a37a4ad34b420fdbe712/src/lib/json-schema/merge/symbols.test.ts)).

As I understand it, extracting schemas containing the `contains` keyword was introduced as a workaround for incorrect merging of `pattern` values in https://github.com/rjsf-team/react-jsonschema-form/pull/4312.

[...]

PR #5235, opened 2026-08-31
fix: populate defaults for a $ref wrapped in a single-element allOf

### LINKED ISSUE #5177: default from $ref is not populated when the $ref is wrapped in a single-element allOf
### Prerequisites

### What theme are you using?

core

### Version

6.x

### Current Behavior

When a property is defined as `allOf: [{ "$ref": "..." }]` instead of a direct `$ref`,
the `default` declared in the referenced subschema is **not** populated into `formData`,
even with `experimental_defaultFormStateBehavior.allOf: "populateDefaults"` enabled.

On top of that, the select widget visually displays `Cat` - the first enum option -
while `formData` is `{}`. So the UI shows a selection the form data does not contain.

### Expected Behavior

1. `allOf: [{ $ref }]` and a bare `$ref` are semantically equivalent for a single-element
   `allOf`, so both should produce `formData = { "animal": "Dog" }`.
2. Independently of (1): the widget should never display an enum option that is not
   present in `formData`. When the value is `undefined` it should render an empty /
   placeholder selection rather than silently falling back to the first enum entry.

### Steps To Reproduce

Schema that works - `formData` becomes `{ "animal": "Dog" }`:

```json
{
  "type": "object",
  "properties": {
    "animal": { "title": "Animal", "$ref": "#/definitions/Animal" }
  },
  "required": ["animal"],
  "definitions": {
    "Animal": { "type": "string", "enum": ["Cat", "Dog", "Bird"], "default": "Dog" }
  }
}
```

Same schema with the `$ref` wrapped in `allOf` - `formData` stays `{}`:

```json
{
  "type": "object",
  "properties": {
    "animal": { "title": "Animal", "allOf": [{ "$ref": "#/definitions/Animal" }] }
  },
  "required": ["animal"],
  "definitions": {
    "Animal": { "type": "string", "enum": ["Cat", "Dog", "Bird"], "default": "Dog" }
  }
}
```

1. Open the playground link below (start with empty `formData`).
2. Set `experimental_defaultFormStateBehavior.allOf` to `populateDefaults`.
3. Observe that the select shows `Cat` while `formData` remains `{}`, whereas the
   `$ref` variant shows `Dog` and yields `{ "animal": "Dog" }`.

Playground links:
- [working ($ref)](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJhbmltYWwiOiJEb2cifSwic2NoZW1hIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7ImFuaW1hbCI6eyJ0aXRsZSI6IkFuaW1hbCIsIiRyZWYiOiIjL2RlZmluaXRpb25zL0FuaW1hbCJ9fSwicmVxdWlyZWQiOlsiYW5pbWFsIl0sImRlZmluaXRpb25zIjp7IkFuaW1hbCI6eyJ0eXBlIjoic3RyaW5nIiwiZW51bSI6WyJDYXQiLCJEb2ciLCJCaXJkIl0sImRlZmF1bHQiOiJEb2cifX19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6ZmFsc2UsInJlYWRvbmx5IjpmYWxzZSwib21pdEV4dHJhRGF0YSI6ZmFsc2UsImxpdmVPbWl0IjpmYWxzZSwiZXhwZXJpbWVudGFsX2NvbXBvbmVudFVwZGF0ZVN0cmF0ZWd5IjoiY3VzdG9tRGVlcCIsImV4cGVyaW1lbnRhbF9kZWZhdWx0Rm9ybVN0YXRlQmVoYXZpb3IiOnsiYXJyYXlNaW5JdGVtcyI6InBvcHVsYXRlIiwiYWxsT2YiOiJwb3B1bGF0ZURlZmF1bHRzIiwiY29uc3RBc0RlZmF1bHRzIjoiYWx3YXlzIiwiZW1wdHlPYmplY3RGaWVsZHMiOiJwb3B1bGF0ZUFsbERlZmF1bHRzIiwibWVyZ2VEZWZhdWx0c0ludG9Gb3JtRGF0YSI6InVzZUZvcm1EYXRhSWZQcmVzZW50IiwibmVzdGVkRGVmYXVsdHNQcmVjZWRlbmNlIjoiZGVzY2VuZGFudFdpbnMifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOmZhbHNlLCJpZFByZWZpeCI6InJvb3QiLCJpZFNlcGFyYXRvciI6Il8ifSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJTaW1wbGUifQ==)
- [broken (allOf)](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6e30sInNjaGVtYSI6eyJ0eXBlIjoib2JqZWN0IiwicHJvcGVydGllcyI6eyJhbmltYWwiOnsidGl0bGUiOiJBbmltYWwiLCJhbGxPZiI6W3siJHJlZiI6IiMvZGVmaW5pdGlvbnMvQW5pbWFsIn1dfX0sInJlcXVpcmVkIjpbImFuaW1hbCJdLCJkZWZpbml0aW9ucyI6eyJBbmltYWwiOnsidHlwZSI6InN0cmluZyIsImVudW0iOlsiQ2F0IiwiRG9nIiwiQmlyZCJdLCJkZWZhdWx0IjoiRG9nIn19fSwidWlTY2hlbWEiOnt9LCJ0aGVtZSI6ImRlZmF1bHQiLCJsaXZlU2V0dGluZ3MiOnsic2hvd0Vycm9yTGlzdCI6InRvcCIsInZhbGlkYXRlIjpmYWxzZSwiZGlzYWJsZWQiOmZhbHNlLCJub0h0bWw1VmFsaWRhdGUiOmZhbHNlLCJyZWFkb25seSI6ZmFsc2UsIm9taXRFeHRyYURhdGEiOmZhbHNlLCJsaXZlT21pdCI6ZmFsc2UsImV4cGVyaW1lbnRhbF9jb21wb25lbnRVcGRhdGVTdHJhdGVneSI6ImN1c3RvbURlZXAiLCJleHBlcmltZW50YWxfZGVmYXVsdEZvcm1TdGF0ZUJlaGF2aW9yIjp7ImFycmF5TWluSXRlbXMiOiJwb3B1bGF0ZSIsImFsbE9mIjoicG9wdWxhdGVEZWZhdWx0cyIsImNvbnN0QXNEZWZhdWx0cyI6ImFsd2F5cyIsImVtcHR5T2JqZWN0RmllbGRzIjoicG9wdWxhdGVBbGxEZWZhdWx0cyIsIm1lcmdlRGVmYXVsdHNJbnRvRm9ybURhdGEiOiJ1c2VGb3JtRGF0YUlmUHJlc2VudCIsIm5lc3RlZERlZmF1bHRzUHJlY2VkZW5jZSI6ImRlc2NlbmRhbnRXaW5zIn0sInVzZUZhbGxiYWNrRmllbGQiOmZhbHNlLCJ1c2VGYWxsYmFja1VpRm9yVW5zdXBwb3J0ZWRUeXBlIjpmYWxzZSwibGl2ZVZhbGlkYXRlIjpmYWxzZSwiaWRQcmVmaXgiOiJyb290IiwiaWRTZXBhcmF0b3IiOiJfIn0sInZhbGlkYXRvciI6IkFKVjgiLCJzYW1wbGVOYW1lIjoiU2ltcGxlIn0=)

### Environment

```markdown
- OS: macOS 26.5.2
- Node: 25.2.1
- npm: 11.6.2
```

### Anything else?

_No response_

fahadhewad:

Fixes #5177.

When `experimental_defaultFormStateBehavior.allOf` is set to `populateDefaults`, `computeDefaults()` only merged a schema's `allOf` for schemas that resolve to an object, because `getObjectDefaults()` is the only caller of `retrieveSchema()` on that path. A property declared as `allOf: [{ "$ref": "#/definitions/Animal" }]` therefore produced no default at all, while the equivalent bare `$ref` correctly produced `{ "animal": "Dog" }`.

This change merges the `allOf` of a non object schema in `computeDefaults()` as well, so the wrapped form and the bare form behave the same. Object schemas keep going through `getObjectDefaults()` as before, and the default `skipDefaults` behavior is unchanged.

`retrieveSchema()` leaves `allOf` in place when it cannot merge every subschema, for example when one of them uses `contains`. Recursing on such a schema would never terminate, so the merged schema is only used once the `allOf` has actually been resolved. A test covers that case as well.

No documentation change was needed, since the behavior now matches what the `populateDefaults` value is already documented to do.

The second point raised in the issue, about the select widget showing an enum option that is not present in `formData`, is a rendering concern in the widget rather than a default form state one, so it is not addressed here.

### Checklist

### DIFF
--- a/packages/utils/src/schema/getDefaultFormState.ts
+++ b/packages/utils/src/schema/getDefaultFormState.ts
@@ -404,6 +404,28 @@ export function computeDefaults<T = any, S extends StrictRJSFSchema = RJSFSchema
       )
     ] as S;
     schemaToCompute = mergeSchemas(remaining, schemaToCompute) as S;
+  } else if (
+    ALL_OF_KEY in schema &&
+    experimental_defaultFormStateBehavior?.allOf === 'populateDefaults' &&
+    getSchemaType<S>(schema) !== 'object'
+  ) {
+    // `allOf` on an object schema is already resolved by `getObjectDefaults()`. On any other schema
+    // nothing resolves it, so the defaults of the subschemas are lost. This happens, for instance,
+    // for a single-element `allOf` wrapping a `$ref` to a string, which is equivalent to using the
+    // `$ref` directly. Merge the `allOf` here so those defaults are picked up as well.
+    const mergedSchema = retrieveSchema<T, S, F>(
+      validator,
+      schema,
+      rootSchema,
+      rawFormData,
+      experimental_customMergeAllOf,
+    );
+    // `retrieveSchema()` leaves `allOf` in place when it cannot merge every subschema, such as when
+    // one of them uses `contains`. Recursing on such a schema would never terminate, so only use it
+    // once the merge has actually resolved the `allOf`.
+    if (!(ALL_OF_KEY in mergedSchema)) {
+      schemaToCompute = mergedSchema;
+    }
   }
 
   if (schemaToCompute) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 -->
 
+# 6.9.1
+
+## @rjsf/utils
+
+- Fixed `computeDefaults()` to merge a non-object schema's `allOf` when `experimental_defaultFormStateBehavior.allOf` is set to `populateDefaults`, so a `$ref` wrapped in a single-element `allOf` now populates the same defaults as the bare `$ref` does, fixing [#5177](https://github.com/rjsf-team/react-jsonschema-form/issues/5177)
+
 # 6.9.0
 
 > **Potentially breaking change:** `lodash` and `lodash-es` are no longer dependencies of any `@rjsf/*` package. No RJSF public API changed, but an application that imports `lodash` itself without declaring it — relying on it being hoisted into `node_modules` because RJSF depended on it — must now add `lodash` to its own `package.json`.
--- a/packages/utils/test/schema/getDefaultFormStateTest.ts
+++ b/packages/utils/test/schema/getDefaultFormStateTest.ts
@@ -3078,6 +3078,70 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
           }),
         ).toEqual({ animalInfo: { animal: 'Cat', food: 'meat' } });
       });
+      it('should populate the default of a `$ref` wrapped in a single-element `allOf`', () => {
+        const schema: RJSFSchema = {
+          type: 'object',
+          properties: {
+            animal: { title: 'Animal', allOf: [{ $ref: '#/definitions/Animal' }] },
+          },
+          required: ['animal'],
+          definitions: {
+            Animal: { type: 'string', enum: ['Cat', 'Dog', 'Bird'], default: 'Dog' },
+          },
+        };
+
+        expect(
+          computeDefaults(testValidator, schema, {
+            rootSchema: schema,
+            experimental_defaultFormStateBehavior: { allOf: 'populateDefaults' },
+          }),
+        ).toEqual({ animal: 'Dog' });
+      });
+      it('should populate the defaults of a `$ref` to an object wrapped in a single-element `allOf`', () => {
+        const schema: RJSFSchema = {
+          type: 'object',
+          properties: {
+            pet: { allOf: [{ $ref: '#/definitions/Pet' }] },
+          },
+          definitions: {
+            Pet: {
+              type: 'object',
+              properties: {
+                name: { type: 'string', default: 'Rex' },
+                legs: { type: 'integer', default: 4 },
+              },
+            },
+          },
+        };
+
+        expect(
+          computeDefaults(testValidator, schema, {
+            rootSchema: schema,
+            experimental_defaultFormStateBehavior: { allOf: 'populateDefaults' },
+          }),
+        ).toEqual({ pet: { name: 'Rex', legs: 4 } });
+      });
+      it('should not recurse endlessly on an `allOf` that cannot be merged', () => {
+        // An `allOf` subschema using `contains` is left in place by `retrieveSchema()` instead of
+        // being merged away, so the merged schema must not be resolved a second time.
+        const schema: RJSFSchema = {
+          type: 'object',
+          properties: {
+            list: {
+              type: 'array',
+              items: { type: 'string' },
+              allOf: [{ contains: { const: 'a' } }],
+            },
+          },
+        };
+
+        expect(
+          computeDefaults(testValidator, schema, {
+            rootSchema: schema,
+            experimental_defaultFormStateBehavior: { allOf: 'populateDefaults' },
+          }),
+        ).toEqual({});
+      });
     });
 
     describe('default form state behaviour: allOf = "skipDefaults"', () => {
@@ -3125,6 +3189,25 @@ export default function getDefaultFormStateTest(testValidator: TestValidatorType
             experimental_defaultFormStateBehavior: { allOf: 'skipDefaults' },
           }),
         ).toEqual({ animalInfo: { animal: 'Cat' } });
+      });
+      it('should not populate the default of a `$ref` wrapped in a single-element `allOf`', () => {
+        const schema: RJSFSchema = {
+          type: 'object',
+          properties: {
+            animal: { title: 'Animal', allOf: [{ $ref: '#/definitions/Animal' }] },
+          },
+          required: ['animal'],
+          definitions: {
+            Animal: { type: 'string', enum: ['Cat', 'Dog', 'Bird'], default: 'Dog' },
+          },
+        };
+
+        expect(
+          computeDefaults(testValidator, schema, {
+            rootSchema: schema,
+            experimental_defaultFormStateBehavior: { allOf: 'skipDefaults' },
+          }),
+        ).toEqual({});
       });
     });
     describe('default form state behavior: arrayMinItems.populate = "never"', () => {


heath-freenome:
## Review

### Overview

`computeDefaults()` only resolved `allOf` (under `experimental_defaultFormStateBehavior.allOf: 'populateDefaults'`) via `getObjectDefaults()`, which is reached only when a schema is type `object`. A property like `{ allOf: [{ $ref: '#/definitions/Animal' }] }` where `Animal` is a `string` enum never got its default populated, unlike the equivalent bare `$ref`. This PR adds a new branch in `computeDefaults()` (`packages/utils/src/schema/getDefaultFormState.ts`) that, for non-object schemas with `allOf` under `populateDefaults`, calls `retrieveSchema()` to merge the `allOf` and recomputes defaults on the merged schema — but only if the merge fully resolved (no leftover `allOf`), avoiding infinite recursion on unmergeable `allOf` (e.g. one using `contains`). Includes a CHANGELOG entry and three new tests plus one `skipDefaults` regression test.

### Code quality / correctness

- **Branch placement is correct.** The new `else if` sits after `ONE_OF_KEY`/`ANY_OF_KEY` and is gated on `getSchemaType(schema) !== 'object'`, so it doesn't double-merge the object case already handled in `getObjectDefaults()` (confirmed by tracing both code paths).
- **Non-object check is on the wrapper schema, not the resolved target** — correctly triggers even for `{ allOf: [{ $ref: ... }] }` wrapping an *object* target (test 2), since the wrapper itself has no `type`/`properties` until merged.
- **Recursion safety net is sound and tested.** `retrieveSchema()` leaves `allOf` in place when a subschema can't merge (`contains`); the `!(ALL_OF_KEY in mergedSchema)` guard prevents recursing on that, and the added test exercises exactly this case.
- Comments explain the non-obvious "why" (recursion termination, object-vs-non-object split) rather than restating the code.
- Reuses existing imports (`ALL_OF_KEY`, `getSchemaType`, `retrieveSchema`) and mirrors the `shouldRetrieveAllOf` pattern already used elsewhere in the same file (`ensureFormDataMatchingSchema`, `getObjectDefaults`) — consistent with existing conventions.
- Scoped entirely to `@rjsf/utils`, the single shared source for default computation, so the fix propagates to `core` and all theme packages without additional changes.

### Test coverage

- Positive case: `$ref`-wrapped primitive (`allOf`) now populates default — matches the bare `$ref` behavior.
- Positive case: `$ref`-wrapped object via `allOf` also populates nested defaults.
- Negative/safety case: unmergeable `allOf` (`contains`) doesn't infinite-loop and returns `{}`.
- Regression case: `skipDefaults` behavior explicitly unaffected.
- Good coverage for the fix's surface area; naming and structure match existing tests in the file.

### Risks / things worth double-checking

- **Behavior generalizes beyond `$ref`.** The fix isn't `$ref`-specific — any non-object schema with a *mergeable* `allOf` (e.g. an array schema with `allOf: [{minItems: 2}]`) will now also get its `allOf` merged for default purposes under `populateDefaults`. This looks like a correct, desirable generalization of the documented `populateDefaults` semantics, not scope creep, but it's a slightly larger behavior change than the PR description ("`$ref` wrapped in `allOf`") suggests — worth a conscious sign-off rather than a surprise.
- No usages of `populateDefaults` exist in `snapshot-tests` or `core` test suites, so this shouldn't perturb existing snapshots; low regression risk since the behavior is opt-in.
- CHANGELOG entry is well-formed and follows the "append after existing bullets" convention; adding a new `# 6.9.1` heading matches the repo's pattern of contributors adding the next version heading post-release.

### Verdict

Small, well-targeted, well-tested fix. No correctness issues found. The main suggestion is not a change request but a documentation nit: the PR description could note that the merge now applies to any non-object `allOf` schema under `populateDefaults`, not just `$ref`-wrapped ones, so reviewers evaluate the full blast radius.

---
*Review generated with [Claude Code](https://claude.com/claude-code)*

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -404,6 +404,28 @@ export function computeDefaults<T = any, S extends StrictRJSFSchema = RJSFSchema
...
+    const mergedSchema = retrieveSchema<T, S, F>(
+      validator,
+      schema,
+      rootSchema,
+      rawFormData,
+      experimental_customMergeAllOf,
+    );
+    // `retrieveSchema()` leaves `allOf` in place when it cannot merge every subschema, such as when
+    // one of them uses `contains`. Recursing on such a schema would never terminate, so only use it
+    // once the merge has actually resolved the `allOf`.
```
x0k, line 425: The `contains` case can be fixed (as described in #5146). The only remaining case is when the schema contains multiple `if`/`then`/`else` conditions [explanation](https://github.com/rjsf-team/react-jsonschema-form/blob/e0f36c9434cc3e187e6f3d4ec9990fccac44a568/packages/utils/src/schema/omitExtraData.ts#L467-L470).

Maybe in this case we should add handling for it directly in `retrieveSchema`?

heath-freenome: @x0k can you explain your idea more?

x0k: I wonder if something like this would work instead of the current handling of `allOf` in `retrieveSchema`:
```js
try {
  const { allOf, ...rest } = resolvedSchema;
  const toMerge = [rest]
  let hasCondition = rest.if !== undefined;
  for (const schemaDef of allOf) {
    if (typeof schemaDef !== 'boolean' && schemaDef.if !== undefined) {
      if (hasCondition) {
        const r = retrieveSchema(schemaDef, rawFormData)
        if (!isDeepEqual(r, { not: {} })) {
          toMerge.push(r)
        }
        continue;
      }
      hasCondition = true
    }
    toMerge.push(schemaDef)
  }
  resolvedSchema = mergeAllOf({ allOf: toMerge })
}
```

heath-freenome, line 425: This comment overstates what the guard protects against. `retrieveSchema()` only *leaves `allOf` in place* for the deliberate `contains` case. A genuine merge failure (e.g. `mergeAllOf`/`experimental_customMergeAllOf` throwing) is caught a few lines down in `retrieveSchemaInternal` and instead **drops `allOf` entirely**, returning a schema without it. Since the guard here only checks `!(ALL_OF_KEY in mergedSchema)`, that failed-and-discarded case reads as "successfully resolved" and this branch will silently compute defaults from the allOf-stripped schema, losing whatever defaults lived in the subschemas.

That's not a new problem you're introducing — `getObjectDefaults()` has the exact same characteristic already for object schemas going through the identical `retrieveSchema()` call, so it's an accepted existing tradeoff, not a regression. But the comment should describe it accurately rather than implying full protection:

```suggestion
    // `retrieveSchema()` isn't guaranteed to fully resolve `allOf`: it deliberately leaves it in place
    // when a subschema uses `contains`, since recursing on that would never terminate — that's the case
    // this guard protects against. A genuine merge failure instead drops `allOf` entirely rather than
    // leaving it in place (the same tradeoff `getObjectDefaults()` already makes for object schemas), so
    // this only guarantees no infinite recursion, not that every subschema was actually merged.
```

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -404,6 +404,28 @@ export function computeDefaults<T = any, S extends StrictRJSFSchema = RJSFSchema
       )
     ] as S;
     schemaToCompute = mergeSchemas(remaining, schemaToCompute) as S;
+  } else if (
+    ALL_OF_KEY in schema &&
+    experimental_defaultFormStateBehavior?.allOf === 'populateDefaults' &&
+    getSchemaType<S>(schema) !== 'object'
+  ) {
```
heath-freenome, line 411: Minor maintainability nit (not a correctness issue): the `ALL_OF_KEY in schema && experimental_defaultFormStateBehavior?.allOf === 'populateDefaults'` gate you're adding here is now duplicated a third time — it already exists (reformatted slightly) as `ensureFormDataMatchingSchema`'s `shouldRetrieveAllOf` and `getObjectDefaults`'s `shouldRetrieveSchema` further down in this same file. Three independently-maintained copies of the same condition make it easy for a future tweak (e.g. to the recursion guard discussed above) to land on only one or two of the three call sites.

GitHub won't let me drop a one-click suggestion at those other two spots since they're outside this PR's diff, but here's the shape I'd suggest, added near the top of the file (below `const PRIMITIVE_TYPES = ...`):

```ts
function shouldPopulateAllOfDefaults<S extends StrictRJSFSchema = RJSFSchema>(
  schema: S,
  experimental_defaultFormStateBehavior?: Experimental_DefaultFormStateBehavior,
): boolean {
  return Boolean(experimental_defaultFormStateBehavior?.allOf === 'populateDefaults' && ALL_OF_KEY in schema);
}
```

Then here:
```suggestion
  } else if (
    shouldPopulateAllOfDefaults(schema, experimental_defaultFormStateBehavior) &&
    getSchemaType<S>(schema) !== 'object'
  ) {
```

...and reuse it at the other two sites:
```ts
// ensureFormDataMatchingSchema
const shouldRetrieveAllOf = shouldPopulateAllOfDefaults(schema, experimental_defaultFormStateBehavior);
```
```ts
// getObjectDefaults
const shouldRetrieveSchema =
  shouldPopulateAllOfDefaults(schema, experimental_defaultFormStateBehavior) ||
  (experimental_defaultFormStateBehavior?.emptyObjectFields !== 'skipEmptyDefaults' && IF_KEY in schema);
```

Optional cleanup — happy to see this PR ship without it if you'd rather keep the diff minimal.

heath-freenome (commented):
A couple of follow-up suggestions from digging into this fix while comparing it against #5185 for the same issue (#5177). Nothing blocking — the core fix (the new branch, the `getSchemaType(...) !== 'object'` gate, and the recursion guard) is correct and well-tested; these are just a comment-accuracy nit and an optional dedup suggestion.

packages/utils/src/schema/getDefaultFormState.ts
heath-freenome: @x0k Could you write up a new issue with your proposal?

heath-freenome:
## CI fix

The `build` checks were failing because `@rjsf/utils` requires 100% branch coverage and line 448 (`if (!(ALL_OF_KEY in mergedSchema))`) only ever took its true branch.

Root cause: `retrieveSchema()`'s `allOf`-merge step always strips the `allOf` key from its result — on a successful merge, and also in its `catch` block when the merge throws (e.g. incompatible `type`/`enum`, or previously `contains`). So the guard added here to protect against `retrieveSchema()` "leaving `allOf` in place" was checking for a case that can't happen given `retrieveSchema()`'s current contract. That was also true when this PR was written, but became directly observable once #5288 upgraded `@x0k/json-schema-merge` on `main`: the `contains` example this guard/test targeted now merges successfully too (previously it likely still hit the `catch` path, which drops `allOf` anyway — so the guard was unreachable dead code either way, just harder to notice before).

Pushed a fix (using maintainer-edit access) that:
- Drops the guard and its now-inaccurate comment in `computeDefaults()` — `schemaToCompute` is now assigned directly from `retrieveSchema()`'s result.
- Updates the "should not recurse endlessly…" test's name/comment, since it was asserting protection against a scenario that doesn't occur; the test itself (contains-in-allOf → `{}`) still passes and is still a useful regression check.

Verified locally: `pnpm test` in `packages/utils` passes all 1585 tests at 100%/100%/100%/100% coverage, and `tsc --noEmit` is clean.

---
*Investigated and fixed with [Claude Code](https://claude.com/claude-code)*

heath-freenome (approved):

packages/utils/src/schema/getDefaultFormState.ts
x0k: #5292

PR #5234, opened 2026-08-30
Releasing 6.9.0

heath-freenome:
### Reasons for making this change

Releasing 6.9.0

Bumping shadcn css files 

### Checklist

- [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests. I've run `pnpm exec nx run-many --target=build --exclude=@rjsf/docs && pnpm run test:update` to update snapshots, if needed.
  - [ ] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #5236, opened 2026-08-31
ci: post the size report on fork PRs and measure every released package

jimmycallin:
Answers [@heath-freenome on #5234](https://github.com/rjsf-team/react-jsonschema-form/pull/5234#issuecomment-5471228752), *"did we want to measure the sizes of all packages that we release?"*, and fixes why that report never appeared on #5233.

## 1. Fork PRs never got a comment

The size report is two workflows: one measures (runs fork code, read-only token), one comments (write token, never runs fork code). The comment job looked up the PR from the commit sha:

```
$ gh api repos/rjsf-team/react-jsonschema-form/commits/c51123c40.../pulls
422 No commit found for SHA
```

A fork's commit does not exist in the base repo, so this lookup always fails for fork PRs. The job warned and returned: green check, no comment. Only base-repo branches got comments, which is why nobody noticed.

**Fix:** look the PR up by head repository and branch. Both come from the `workflow_run` payload, which fork code cannot write to, so the trust split between the two workflows is unchanged.

```js
const { data: prs } = await github.rest.pulls.list({
  owner, repo, state: 'all',
  head: `${run.head_repository.owner.login}:${run.head_branch}`,
});
const pr = prs.find((p) => p.head.sha === headSha);
```

Since a silent skip is how this went unnoticed, two more changes:
- The job now **fails** when it cannot identify the PR. Budget overruns still never fail this job; that remains `Size limit / size`.
- The report **stamps the measured sha**, so a comment that lands after the branch moved says which commit it describes.

## 2. Measure every released package

`.size-limit.json` (8 hand-written checks, 3 packages) becomes `.size-limit.js`, derived from each package's `package.json`: **29 checks across all 15 released packages**. Adding a package or changing a dependency needs no config edit.

Each package gets two rows:
- **installed**: ignores only its peer dependencies, so a theme is measured without the UI library behind it;
- **without dependencies**: its own code only. Omitted for packages that have no dependencies.

**Budgets are unchanged.** Only the three packages that already had one enforce one. Per-theme budgets would mostly generate bump commits; the delta column in the PR comment is what catches a regression. All 8 pre-existing numbers reproduce exactly.

<details>
<summary>All 29 checks, measured locally at 84c424e0c</summary>

| Check | Size | | Check | Size |
| --- | --- | --- | --- | --- |
| `@rjsf/antd` | 6.85 kB | | `@rjsf/react-bootstrap` | 7.54 kB |
| `@rjsf/antd (without dependencies)` | 6.45 kB | | `@rjsf/react-bootstrap (without dependencies)` | 5.30 kB |
| `@rjsf/chakra-ui` | 18.15 kB | | `@rjsf/semantic-ui` | 5.35 kB |
| `@rjsf/chakra-ui (without dependencies)` | 10.81 kB | | `@rjsf/semantic-ui (without dependencies)` | 5.35 kB |
| `@rjsf/core` | 80.20 kB | | `@rjsf/shadcn` | 37.84 kB |
| `@rjsf/core (without dependencies)` | 22.18 kB | | `@rjsf/shadcn (without dependencies)` | 8.68 kB |
| `@rjsf/core: Form` | 44.57 kB | | `@rjsf/utils` | 29.78 kB |
| `@rjsf/daisyui` | 49.67 kB | | `@rjsf/utils (without dependencies)` | 17.24 kB |
| `@rjsf/daisyui (without dependencies)` | 9.66 kB | | `@rjsf/utils: getUiOptions` | 0.31 kB |
| `@rjsf/fluentui-rc` | 5.58 kB | | `@rjsf/validator-ajv8` | 36.34 kB |
| `@rjsf/mantine` | 11.43 kB | | `@rjsf/validator-ajv8 (without dependencies)` | 2.61 kB |
| `@rjsf/mantine (without dependencies)` | 7.07 kB | | `@rjsf/validator-ata` | 55.25 kB |
| `@rjsf/mui` | 6.30 kB | | `@rjsf/validator-ata (without dependencies)` | 2.34 kB |
| `@rjsf/primereact` | 6.27 kB | | `@rjsf/validator-cfworker` | 7.24 kB |
| | | | `@rjsf/validator-cfworker (without dependencies)` | 2.32 kB |

`@rjsf/snapshot-tests` is published but ships no bundle, so it has no `build:ts` and is not measured.

</details>

## 3. Incidentals

- **`build:ts` is now an nx target** with `lib/` as its declared output. size-limit measures `lib/`, so CI no longer builds the cjs/esm/umd bundles, and `--all` resolves to exactly the released packages since only they declare the target.
- **`.size-limit/report.mjs` moved to `.github/size-limit/report.mjs`.** size-limit's config search trips over a `.size-limit` directory at the root (`EISDIR`). Latent until the config stopped being `.size-limit.json`.

## Verification

- `size-limit` exits 0 across all 29 checks; the 8 pre-existing checks reproduce their previous numbers exactly.
- `report.mjs` rendered against real head/base json. `knip` and `oxfmt --check` clean. Both workflows parse as YAML.
- The new lookup verified live: `pulls.get(5233)` returns head `f57ff07e`, the sha the old lookup 422'd on.

The workflow paths only run in CI, so **this PR is itself the first live test of the fork path.**


heath-freenome:
## Code review findings

Found 4 issues in the new size-limit CI plumbing:

**1. `.github/workflows/size-limit-comment.yml:53`** — Spurious `setFailed` on a benign rapid-push race

The PR lookup by head branch can fail even when nothing is broken: if a contributor pushes a new commit right after the "Size limit" run for the previous commit finishes, the job can't find a PR matching the old SHA (since the PR's head has already moved on) and calls `core.setFailed` even though the new commit's own run will complete normally.

**2. `.github/workflows/size-limit.yml:41`** — Missing fast-fail guard for base commits without `.size-limit.js`

The old `[ -f .size-limit.json ]` short-circuit was removed but not replaced with an equivalent check for the new `.size-limit.js`. When the base commit lacks the config (e.g. an old base, or mid-rollout of this change), the job now pays for a full `pnpm install` + multi-package `build:ts` before `size-limit --json` fails and falls back to `echo '[]'`, instead of skipping straight to the fallback.

**3. `.github/workflows/size-limit-comment.yml:53`** — Unpaginated `pulls.list` call

This call uses the default `per_page=30` instead of `github.paginate` (used elsewhere in this file). A branch name reused across 30+ historical PRs (e.g. a long-lived dependency-bump branch) could have its current PR fall outside the first page, causing `.find(p => p.head.sha === headSha)` to miss it and incorrectly trigger `setFailed`.

**4. `.size-limit.js:40`** — `extraIgnore` silently dropped for dependency-free packages

`extraIgnore` is only threaded through on the "(without dependencies)" row, which itself is only emitted when `deps.length > 0`. A package with zero runtime dependencies but a configured `extraIgnore` would silently have that exclusion have no effect, since the row that applies it is never generated.

---

jimmycallin:
Thanks, all four addressed in a1c65ddc7:

1. **Stale run race** — when no PR matches the run's sha but PRs exist for that head, the job now logs and skips (the newer commit's own run posts the comment). It only fails when there is no PR for the head at all, which is the broken-handoff case the failure was meant to catch.
2. **Base guard** — `[ -f .size-limit.js ]` is back in the base chain before `pnpm install`, so a base without the config falls straight through to the empty baseline.
3. **Pagination** — the `pulls.list` call goes through `github.paginate` with `per_page: 100`.
4. **`extraIgnore` without deps** — the "(without dependencies)" row is emitted when either `deps` or `extraIgnore` is non-empty.

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ should change the heading of the (upcoming) version to include a major version b
...
 - Fixed `packages/daisyui/test/tsconfig.json`, which was configured to emit into `../dist` — the esbuild/rollup bundle output directory — instead of type-checking without emit like every other test project
+- Fixed the size-limit report never being posted on a pull request from a fork. The comment workflow resolved the PR from its head sha, which the base repository cannot associate with a fork's commit, so it warned and skipped — leaving a green check and no comment. The PR is now found by head repository and branch, both of which come from the trusted `workflow_run` payload, and the job fails loudly rather than skipping when no PR matches
+- Extended the size-limit checks to cover every released package rather than just `@rjsf/core`, `@rjsf/utils` and `@rjsf/validator-ajv8`. The checks are derived from each package's own `package.json`, so a new package or a dependency change needs no config edit; only the packages that already had budgets enforce one, the rest are measured and reported
```
heath-freenome: These now belong in 6.10.1

heath-freenome: @jimmycallin also, you have conflicts to resolve.

jimmycallin: Moved to a new 6.10.1 heading and rebased onto main; conflicts resolved.

heath-freenome:
## Follow-up review

The PR head has changed since my earlier review (the size-limit-comment.yml lookup logic was reworked from a `pulls.list`-by-branch-name approach to using `run.head_repository`). 3 new findings:

**1. `.github/workflows/size-limit-comment.yml:55`** — Unchecked null dereference on `head_repository`

`run.head_repository.owner.login` is dereferenced with no null check, but GitHub's `workflow_run` payload documents `head_repository` as nullable when the source repo has been deleted. If a contributor opens a PR from a fork and deletes the fork before the triggering "Size limit" run's `workflow_run` event fires, this throws an opaque TypeError instead of reaching the `core.setFailed('No PR for ...')` path this PR added specifically to make broken handoffs legible.

**2. `.github/workflows/size-limit-comment.yml:57`** — No tie-breaking in PR lookup

`prs.find((p) => p.head.sha === headSha)` takes the first list match with no tie-breaking. If a PR is closed and reopened as a new PR from the same branch without new commits (same head sha), `pulls.list({state: 'all', head})` returns both records — depending on ordering, the comment could get posted/updated on the old closed PR instead of the currently-running open one.

**3. `.github/workflows/size-limit.yml:11`** — PR number known but not propagated (root cause of #1/#2)

The PR number (`github.event.pull_request.number`) is already available in `size-limit.yml` but isn't written into the uploaded artifact, forcing `size-limit-comment.yml` to re-derive the PR via a heuristic (head owner+branch+sha) instead of an exact lookup. Writing a `pr_number.txt` alongside the existing artifact contents and using `pulls.get(pr_number)` in the comment workflow would make the lookup exact instead of heuristic, eliminating both issues above.

---

jimmycallin:
Thanks, all three addressed in c4edbaab5 by taking the root-cause route from finding 3:

1. **PR number propagated** — `size-limit.yml` writes `pr_number` into the artifact next to `head.json`/`base.json`, and `size-limit-comment.yml` does an exact `pulls.get`. Because the measuring job runs PR code, the number is treated as untrusted: the comment is posted only if the PR's `head.sha` equals `workflow_run.head_sha`, which fork code cannot influence. A mismatch is skipped (the branch moved on, or the artifact named a PR it does not belong to); a missing number or a 404 fails the job so a broken handoff stays visible.
2. **`head_repository` null dereference** — the lookup and the concurrency group no longer read `head_repository` at all; the group is keyed on `head_sha`.
3. **Tie-breaking** — moot with an exact lookup; a closed-and-reopened branch resolves to whichever PR the artifact names, and the sha check still has to pass.

Same commit also moves `size-limit` and `@size-limit/preset-small-lib` out of the root `devDependencies` into their own pnpm project under `.github/size-limit` (with its own lockfile), installed with `--ignore-workspace` only by the size-limit workflow. The config moved with them and resolves packages from the repo root, so `pnpm install` in the repo no longer pulls in the size tooling.

One caveat for verifying: `workflow_run` workflows always execute from the default branch, so the comment step on this PR still runs main's version of `size-limit-comment.yml`. The new lookup gets its first real run on the next PR after this merges; the measuring side (`size-limit.yml`, the nested install, the `pr_number` in the artifact) is exercised here.

heath-freenome:
## Follow-up review (round 3)

The PR number is now propagated via the artifact, which resolves the two findings from my previous comment. 2 new issues introduced by this change:

**1. `.github/workflows/size-limit-comment.yml:53`** — Uncaught crash on old-format artifacts during rollout

The comment job now unconditionally reads a `pr_number` file from the downloaded artifact with no fallback, but this artifact schema is new in this PR. Since `workflow_run` always executes using the workflow file currently on the default branch, an in-flight "Size limit" run that started on any open PR *before* this PR merges will have uploaded the old-format artifact (only `head.json`/`base.json`, no `pr_number`). If that run's completion event fires after this PR merges, `readFileSync('pr_number', 'utf8')` throws an uncaught `ENOENT`, failing the job with an unhandled exception instead of the intended graceful "no PR found" path — reintroducing, as a crash, the exact class of broken-handoff failure this PR was written to fix. Worth guarding with a try/catch (or checking existence) that falls back to `core.setFailed` during the rollout window.

**2. `.github/workflows/size-limit-comment.yml:21`** — Narrowed concurrency group can race on rapid pushes

The concurrency group was narrowed from per-branch to per-commit-sha, which removes serialization between comment jobs for two different commits of the same PR. Two quick pushes to the same fork PR (commit A then commit B) can have their "Size limit" workflow_runs complete close together; their comment jobs now fall into different concurrency groups and can run the check-then-act sticky-comment update concurrently instead of serialized as before. The `pr.head.sha !== headSha` guard usually self-corrects since it reads the PR fresh, but there's a narrow window where commit A's job reads `pr.head.sha` before B's push lands, passes the check, and writes/updates the sticky comment with stale (commit A) data — a case the old per-branch concurrency group would have serialized away.

---

jimmycallin:
Thanks, both addressed in f6cb30662:

1. **Old-format artifacts** — `pr_number` is read only if it exists; a run started before this merges now ends in `core.setFailed("Artifact has no PR number for <sha> (uploaded by an older size-limit.yml?)")` instead of an uncaught `ENOENT`.
2. **Concurrency** — the group is per PR branch again, keyed on `head_repository.full_name` + `head_branch`, so two quick pushes to one PR run their comment jobs in order and the later job's head-sha check sees the moved branch. `head_repository` being null for a deleted fork is fine here: the expression yields an empty segment, and the lookup itself no longer reads it.

heath-freenome (approved):

PR #5242, opened 2026-09-03
build: one tsconfig per package, typecheck tests in CI, stop publishing tsbuildinfo

jimmycallin:
## Problem

Each published package carried four tsconfigs: a `tsconfig.json` solution referencing `src` and `test`, a `tsconfig.build.json` solution referencing only `src`, the `src/tsconfig.json` that actually emitted, and `test/tsconfig.json`. The root had two, `tsconfig.base.json` and `tsconfig.build.json`, disagreeing on `target` and `noUnusedParameters`. Which root config governed a package's emit depended on what its `src/tsconfig.json` happened to extend, which produced bugs invisible from any single file:

1. **Inconsistent published output.** Only core's emitting config extended the root `tsconfig.build.json`, so core shipped `esnext` with declaration maps; the other 14 shipped ES2018 with JS source maps.
2. **`tsconfig.tsbuildinfo` is published.** TypeScript writes it under `outDir` at the config's position relative to `rootDir` — with the emitting config inside `src/`, that is `lib/tsconfig.tsbuildinfo`, and `files` publishes `lib/`. In 6.8.0 it is 446 kB of `@rjsf/antd`'s 959 kB and 265 kB of `@rjsf/mui`'s 856 kB.
3. **Tests were never typechecked.** CI ran `tsc` on `src` only, and Vitest strips types without checking them, so the 15 `test/tsconfig.json` files were editor-only.
4. **The root solution was wrong.** It referenced `snapshot-tests` twice and omitted `mantine`, `primereact`, `validator-ata` and `validator-cfworker`.
5. **Chakra's `type-check` checked nothing** — `tsc --noEmit` against a solution config with `files: []`.
6. **Libraries saw every hoisted `@types/*`**, so library code could compile against Node globals it must not use.
7. **Bloated reference graph.** Source projects referenced their dependencies' solution configs, which include the test project, so building core pulled `utils/test` and `validator-ajv8/test` in.

## Fix

One source project per package at `packages/<name>/tsconfig.json`, one test project at `packages/<name>/test/tsconfig.json` extending a shared root `tsconfig.test.json`, and one root `tsconfig.base.json` with a single deliberate value per option. The root solution lists every source and test project once, and a root `typecheck` script (`tsc --build`) runs in CI.

Decisions worth calling out:

- **Target is `esnext`.** Every theme depends on core and core already shipped `esnext`, so consumers already ran it. Core's `lib/*.js` is byte-identical to main apart from a `sourceMappingURL` trailer; of the rest measured, utils, mui and antd shrink 1–4% and validator-ajv8 grows 0.6%.
- **All packages ship `.js.map` and `.d.ts.map`** — source is published, so both resolve.
- **`"types": []` in the base**, with Node-using packages opting in; **`"lib": ["ESNext"]`** in the three validators, so a validator cannot depend on browser globals.
- **`verbatimModuleSyntax` on.** Zero cost: the source already used `import type` wherever required. The one import it would have stopped eliding, `import React` in utils' `shouldRender.ts`, is type-only and is now `import type`, so emitted output is unchanged.
- **`build:ts` is plain `tsc -b`.** The old `rimraf ./lib` only guarded against stale outputs from renamed or deleted sources, which cannot reach a release because CI builds from a fresh checkout. It also forced deleting `tsconfig.tsbuildinfo`, since TypeScript trusts the build-info over the presence of outputs ([microsoft/TypeScript#50646](https://github.com/microsoft/TypeScript/issues/50646), Working as Intended), so every build was a full rebuild. `--force` is no substitute: it rebuilds every referenced project too, which races under Nx's parallel builds. The build-info is now an Nx `build` output next to `lib/`, so a cache restore keeps the incremental state coherent.

Typechecking tests for the first time surfaced stale ones, fixed here: chakra's test wrapper passed a prop `EnvironmentProvider` does not have, `validator-ata`'s test passed an AJV-only option its validator ignores, antd and shadcn used jest-dom matchers without their types, and two playground samples had drifted from the core API.

## Effect

Same machine and run, clean outputs before each cold build; tarballs are `npm pack --dry-run` after a full `nx build`. Main is `bd1775eac`.

| | main | PR |
|---|---:|---:|
| tsconfig files / lines | 66 / 1173 | 36 / 451 |
| projects in `tsc -b packages/core` graph | 10 | 3 |
| `tsc -b packages/core`, cold | 4.4 s | 2.9 s |
| all 16 source projects, cold | 17.0 s | 16.0 s |
| root `tsc --build` (now includes tests) | red | green |
| `@rjsf/antd` tarball | 184 kB | 99 kB |
| `@rjsf/mui` tarball | 162 kB | 114 kB |
| `@rjsf/utils` tarball | 591 kB | 588 kB |
| `@rjsf/core` tarball | 485 kB | 497 kB |

Core grows because it gains the `.js.map` files the other packages already shipped, minus its 60 kB `tsbuildinfo`. `dist/` is untouched.

## Next steps

Each builds on this topology and is kept separate so it can be reviewed on its own merits. Items 2, 3 and 5 are where this setup still deviates from current guidance (the [TypeScript handbook's library recommendations](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html), the `tsc --init` defaults since 5.9, and [Nx's project-references setup](https://nx.dev/docs/technologies/typescript/guides/switch-to-workspaces-project-references)).

1. **Let tsdown emit JavaScript; `tsc` emits only declarations.** Today `tsc -b` produces `lib/*.js` file by file while esbuild and rollup produce the `dist/` bundles from the same source, so two toolchains must agree on target and module settings. Cost after this PR: `emitDeclarationOnly: true`, one line. It cannot land here because until tsdown is wired up nothing else produces `lib/*.js`.
2. **Check published ESM with Node's module rules (`module: nodenext`).** The handbook calls `moduleResolution: bundler` "infectious" for libraries: it accepts specifiers Node rejects, so `tsc` cannot tell us when `lib/` is unloadable by Node. Trying it already found three shipped defects — `@rjsf/react-bootstrap` imports `react-bootstrap/Col` and an extensionless icon path, so Node cannot load its `lib/index.js` at all; `@rjsf/antd` type-imports the directory `antd/es/select` (erased at runtime, but unresolvable under Node rules); `@rjsf/validator-ajv8` default-imports AJV whose declarations disagree with its CommonJS runtime shape. The source fixes are the real content and deserve their own review.
3. **Turn on `erasableSyntaxOnly`.** The flag that actually delivers "runnable under Node's native type stripping" by rejecting enums, namespaces and parameter properties. Enabling it today produces four errors, all exported enums: `TranslatableString` and `AdditionalItemsHandling` in `@rjsf/utils`, `GridType` and `Operators` in `@rjsf/core`. Replacing them with `as const` objects changes public types, so it needs its own changelog entry.
4. **Move the shared test helpers in `packages/utils/test` into a package export.** The validator tests import them, which is why `utils/test` is the one test project that still emits and why validator test projects reference another package's tests. Nx's recipe is a `./testing` subpath in `exports` pointing at source; that or a private workspace package makes every test project `noEmit`.
5. **Let Nx infer `typecheck` and `build` from the tsconfigs.** `nx.json` sets `useInferencePlugins: false`, so references are maintained by hand — and this PR found them drifted. `@nx/js/typescript` infers the targets and `nx sync` regenerates references from actual imports. Its `typecheck` target used to hardcode `tsconfig.json`, which here would check only `src`; [nrwl/nx#34274](https://github.com/nrwl/nx/issues/34274) is closed and the plugin now takes a `configName`, so verify it accepts `test/tsconfig.json` first. Best done after step 1.


CHANGELOG.md
```diff
@@ -51,6 +60,7 @@ should change the heading of the (upcoming) version to include a major version b
 - Updated `CheckboxWidget` to append a `*` onto the end of the label when the field is required AND the schema value is hardcoded to `true`, fixing [#4136](https://github.com/rjsf-team/react-jsonschema-form/issues/4136)
 - Declared `"sideEffects": false` in `package.json`, allowing bundlers to tree-shake unused exports
 - Fixed `BaseInputTemplate`'s `ui:allowClearTextInputs` clear button to store `ui:emptyValue` (`undefined` by default) instead of always storing `""`, so clicking it now behaves exactly like clearing the input by typing. **Potentially breaking change:** an app that never sets `ui:emptyValue` and clicks the clear button on a field (rather than backspacing it) previously got `""` back; it now gets `undefined`, matching every other way of emptying the field.
+- Removed the `environment` prop passed to `EnvironmentProvider` in the test wrapper. `EnvironmentProvider` has no such prop, so it was silently ignored; dropping it leaves the default environment in place and lets the package's `type-check` script cover the test files
```
nickgros: This got appended to the wrong release version

tsconfig.base.json
```diff
@@ -1,30 +1,25 @@
...
     "composite": true,
-    "target": "ES2018",
+    "target": "esnext",
     "module": "esnext",
-    "lib": ["dom", "esnext"],
-    "importHelpers": false,
-    "declaration": true,
-    "sourceMap": true,
+    "moduleResolution": "bundler",
+    "lib": ["DOM", "ESNext"],
```
nickgros, line 7: In the description, you claim 

> Validators compile without the DOM lib. @rjsf/validator-ajv8, -ata and -cfworker override lib to ["ESNext"], so a validator cannot accidentally depend on browser globals.

But that doesn't appear to be true? The validator tsconfigs don't include a `lib` property

nickgros (changes requested):

tsconfig.base.json
jimmycallin: You are right, I accidentally never pushed that. Fixed now, thanks.

CHANGELOG.md
jimmycallin: Moved to 6.9.1 under a new `## @rjsf/chakra-ui` heading in fe3de75d5.

nickgros (approved):

nickgros:
Thanks @jimmycallin ! Merge when you are ready

Issue #5241, opened 2026-09-03
Widget for a number field get a string as value


sthenault:
### Prerequisites

- [x] I have searched the existing issues
- [x] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [ ] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

mui

### Version

6.8

### Current Behavior

We've a custom widget for a number field which breaks since 6.8 because its `value` prop is given as a string instead of a number as it was up to 6.7.

It appears this occurs on browser with eg. french locale setup (actually language with decimal separator != '.')

### Expected Behavior

get value as a number as before.

### Steps To Reproduce

- define a custom widget for some number field
- setup your browser's locale to "french"
- check the type of the `value` prop given to your widget

### Environment

```markdown
- OS: debian trixie
- Node: 24.20
- npm: 11.19
```

nickgros:
Proposed fix in #5243

PR #5243, opened 2026-09-04
fix(core): preserve numeric formData and dot-typed input in comma-decimal locales

### LINKED ISSUE #5199: Number field gets empty after writing a decimal in comma seperated browser
### Prerequisites

### What theme are you using?

shadcn

### Version

6.8.0

### Current Behavior

Browser Language: Turkish which means decimal seperated is comma ','
Just a regular number field with no validations.
You can use NUMBERS sample from https://rjsf-team.github.io/react-jsonschema-form/
When you write 2.1 or 2,1 etc the number field gets empty.
Reverting to 6.7.0 fixes the problem.

### Expected Behavior

Number field shows the value as 2,1 or 2.1 depending on the browser language.

### Steps To Reproduce

1. Set browser language to turkish.
2. https://rjsf-team.github.io/react-jsonschema-form/ NUMBERS sample
3. Try to write 2.1 or 2,1 to the first number field.

### Environment

```markdown
- OS: Windows 11
https://rjsf-team.github.io/react-jsonschema-form/
```

### Anything else?

_No response_

nickgros:

NumberField stringified formData into a locale-formatted value for every widget except radio/select/hidden, so custom and format-registered widgets received a string instead of a number (#5241), and typing "." instead of the locale separator got silently discarded, corrupting the committed value (#5199).

- getInputProps() now defaults type:number schemas to a text input when the locale decimal separator isn't ".", instead of a native number input that silently rejects/eats non-"." input. Shared by every theme's BaseInputTemplate, so no per-field uiSchema mutation is needed.
- NumberField now only reformats displayValue for the resolved "text" widget (allow-list) instead of excluding radio/select/hidden (deny-list), so every other widget keeps receiving a real number.
- The cached-input redisplay check now canonicalizes to "." before matching, so a typed "." is preserved as a pending decimal point instead of being reformatted away mid-keystroke.
- Escapes the numeric value before embedding it in that regex, fixing an unescaped "." acting as a wildcard against stale cached input.

Fixes #5199, #5241

### Checklist


heath-freenome:
## Code Review

### Overview

This fixes two related bugs in comma-decimal locales (e.g. `pl`, `de`):

1. **#5241** — `NumberField` stringified `formData` into a locale-formatted string for every widget except `radio`/`select`/`hidden` (a deny-list), so custom widgets, `updown`/`range`, and format-registered widgets received a string instead of a `number`.
2. **#5199** — Typing `.` instead of the locale separator was silently discarded, and the cached-input regex embedded the raw numeric value unescaped, so a literal `.` acted as a regex wildcard against stale cached input.

The fix spans two packages:
- `@rjsf/utils/getInputProps.ts`: `type: number` schemas now default to a `text` input (instead of native `type=number`) when the locale decimal separator isn't `.`, unless `inputType` is explicitly set. This is centralized so every theme's `BaseInputTemplate` picks it up for free.
- `@rjsf/core/NumberField.tsx`: switches the widget check from a deny-list to an allow-list (only reformat display value when the *resolved* widget is `text`), canonicalizes the cached value to `.` before matching, and escapes the numeric value before building the cache-match regex.

### Correctness

- **Allow-list vs. deny-list fix is correct and actually more precise than before.** The `defaultWidget` computation added to `NumberField.tsx` (schema format → registered widget → fallback `text`/`select`) is an exact mirror of the real resolution logic in `StringField.tsx` (`getWidget`/`hasWidget`), so `widget === 'text'` correctly identifies exactly the case where `getInputProps()` will render a real text input needing locale-formatted display. This also fixes a latent bug the old deny-list didn't cover: `updown`/`range` widgets were previously getting reformatted display values too.
- **Regex escaping fix is a real, valid bug fix.** The old code built `new RegExp(...String(value)...)` without escaping the decimal point, so `.` acted as "any character" against `lastValue`. The new `escapedValue` uses the standard escape-regex character class, and it's covered by a dedicated test.
- **`getInputProps.ts` change is well-scoped**, only firing when `defaultType` is falsy — but see the scope gap below.

### Scope gap worth flagging

`UpDownWidget` and `RangeWidget` both call `BaseInputTemplate` with an explicit `type='number'`/`type='range'` prop, which is passed to `getInputProps()` as `defaultType`. Since the new locale-detection branch only runs in the `!defaultType` case, **`UpDownWidget` still forces a native `type="number"` input in comma-decimal locales**, and remains susceptible to the exact browser-input bug this PR fixes for the default text-rendering path (silently discarded `.`, mismatched locale format). This may be an intentional scope decision (a user who explicitly opts into `ui:widget: 'updown'` is choosing native spinner behavior), but worth confirming whether that's deliberate or a follow-up gap.

### Code quality / maintainability

- **Duplicated widget-resolution logic.** The `defaultWidget` computation in `NumberField.tsx` (~lines 96–100) is now a verbatim copy of the same block in `StringField.tsx`. That's necessary today, but the two must be kept in sync manually — if `StringField`'s resolution logic changes in the future, `NumberField` could silently regress into the bug being fixed here. Worth a one-line comment noting the coupling, or (better, as a follow-up) extracting a shared `resolveDefaultWidget(schema, uiSchema, registry)` helper into `@rjsf/utils`.
- Comments added throughout explain the *why* well (e.g., "its own `.` is a literal character here, not the regex wildcard").

### Test coverage

Strong coverage overall — new tests exercise the default `text` input type in comma locales, typing `,` and `.` as decimal points, the stale-cache wildcard regression, and passing real `number`s to custom/format-registered widgets. Two minor nits:

1. **`packages/utils/test/getInputProps.test.ts` cleanup pattern is inconsistent and slightly unsafe.** The two new tests call `vi.stubGlobal(...)` then manually `vi.unstubAllGlobals()` at the end of the test body, rather than in an `afterEach`. If the assertion between those two lines throws, the stub leaks into later tests in the same file (e.g. the very next test, `'returns type=number when schema has number type and we are not auto-defaulting'`, assumes an unstubbed/default navigator and would fail for an unrelated reason). The established convention elsewhere (`getDecimalSeparator.test.ts`, and the new `Comma decimal separator locales` describe block in `NumberField.test.tsx`) uses `afterEach(() => vi.unstubAllGlobals())`/`vi.restoreAllMocks()`. Suggest aligning for consistency and safety.
2. **Minor test-title/assertion mismatch:** `'returns type=number, step=any when schema has number type and an explicit inputType overrides the locale'` asserts `toEqual({ type: 'number' })` — no `step` key is present or asserted, so the "step=any" in the title is misleading.

### Other

- **Changelog placement** correctly appends as the last bullet under `@rjsf/core` and `@rjsf/utils`, per project convention.
- **UX tradeoff worth surfacing:** switching `type=number` → `type=text` for comma-locale users loses native spinner buttons, HTML5 numeric validation, and (on mobile) the numeric keypad. Adding `inputMode="decimal"` alongside `type="text"` in `getInputProps.ts` would restore the mobile numeric keypad without reintroducing the native-input parsing bug — worth considering as a follow-up.

### Summary

Solid, well-tested fix for two real, previously-reported bugs. The regex-escaping fix and the allow-list widget check are both correct and improvements over the status quo. Main things worth a look before merge: confirm whether `UpDownWidget` was intentionally left out of scope, and tighten the two test-hygiene nits in `getInputProps.test.ts`. Neither is blocking.

---
*Posted via automated code review (Claude Code).*

packages/core/src/components/fields/NumberField.tsx
```diff
@@ -88,14 +93,15 @@ function NumberField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends
...
     const { widget = defaultWidget } = getUiOptions(uiSchema);
 
-    // Do not convert the value to a locale-specific string for radio, select,
-    // or hidden widgets because option matching relies on the original numeric value.
-    if (widget !== 'radio' && widget !== 'select' && widget !== 'hidden') {
+    // Only the built-in text widget renders as a plain text input in this locale (see
+    // getInputProps()), so it's the only one whose displayed value needs to match the
+    // locale separator. Every other widget (custom, radio, select, hidden,
+    // format-registered, ...) keeps receiving the numeric formData its contract promises.
+    if (widget === 'text') {
```
heath-freenome: NumberField decides whether to locale-format the displayed number by checking only the resolved widget name (`widget === 'text'`), not the actual DOM input type that `getInputProps()` ends up choosing, so an explicit `ui:options.inputType` override can desync from it.

Repro: schema `{ type: 'number' }` with `uiSchema: { 'ui:options': { inputType: 'number' } }` (no `ui:widget`), in a comma-decimal locale (e.g. Polish/German). `resolveDefaultWidget` returns `'text'`, so this `if (widget === 'text')` branch fires and formats `displayValue` as a comma string like `"2,3"`. That string then reaches `getInputProps()` (packages/utils/src/getInputProps.ts:29), where `options.inputType` being explicitly set takes priority over the locale check and forces `inputProps.type = 'number'`. The result is a native `<input type="number" value="2,3">`, which browsers reject/blank out — the exact bug class (#5199) this PR fixes, still reachable through the documented `inputType` override.

heath-freenome (commented):
Rereview: found one remaining correctness gap in this fix (inline comment below).

packages/core/src/components/fields/NumberField.tsx
```diff
@@ -88,14 +93,18 @@ function NumberField<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends
...
+    const { widget = defaultWidget, inputType } = getUiOptions(uiSchema);
+
+    // Only the built-in text widget renders as a plain text input in this locale (see
+    // getInputProps()), so it's the only one whose displayed value needs to match the
+    // locale separator. Every other widget (custom, radio, select, hidden,
+    // format-registered, ...) keeps receiving the numeric formData its contract promises.
+    // An explicit `inputType` option takes priority over the locale detection inside
+    // getInputProps(), so a `text` widget with `inputType: 'number'` (or any type other than
+    // `text`) still ends up in a native, locale-unaware input; skip the comma formatting there too.
+    if (widget === 'text' && (inputType === undefined || inputType === 'text')) {
```
heath-freenome: Nice fix for the earlier `inputType: 'number'` case, but this condition now over-restricts: it only comma-formats when `inputType` is `undefined` or `'text'`, so any other explicit `ui:options.inputType` that renders as a plain text input (e.g. the docs' own `inputType: 'tel'` example, or `'search'`/`'email'`/`'url'`) skips the conversion too.

Repro: `{ type: 'number' }` with `uiSchema: { 'ui:options': { inputType: 'tel' } }` and `formData: 2.3` in a comma-decimal locale. `getInputProps()` sets `inputProps.type = options.inputType = 'tel'` directly — a `<input type="tel">` renders arbitrary text fine, so it would happily display a comma-formatted value. But this condition (`inputType === undefined || inputType === 'text'`) is false for `'tel'`, so `displayValue` stays dot-formatted ('2.3') instead of the locale-correct '2,3'. Before this PR, the exclusion here only checked widget name (not radio/select/hidden), so this exact case comma-formatted correctly — this is a regression for any non-`'number'`, non-`'text'` `inputType`.

Might be worth checking against an allowlist/denylist of input types that actually reject non-numeric characters (e.g. just `'number'`) rather than gating on `inputType === 'text'`.

heath-freenome (commented):
Rereview after the previous fix landed: found one more edge case in the same inputType-gating logic (inline comment below).

packages/core/src/components/fields/NumberField.tsx
heath-freenome: Suggested fix: gate on the one `inputType` that actually renders a native, locale-unaware `<input type="number">` (`'number'` itself), rather than allowlisting only `undefined`/`'text'`. That way `'tel'`, `'email'`, `'search'`, etc. still get comma-formatted since they render as plain text.

```suggestion
    // An explicit `inputType: 'number'` takes priority over the locale detection inside
    // getInputProps(), which still ends up rendering a native, locale-unaware
    // `<input type="number">`; skip the comma formatting there. Every other inputType
    // (including the default 'text', and any other value like 'tel' or 'email') renders as a
    // plain text input that displays a locale-formatted string just fine.
    if (widget === 'text' && inputType !== 'number') {
```

heath-freenome (commented):
Added a suggested code fix for the inputType-gating edge case.

heath-freenome (approved):

PR #5244, opened 2026-09-04
build: TypeScript 7, module nodenext, erasableSyntaxOnly, noEmit test projects

jimmycallin:
Targets `v7`; changelog entries are in `CHANGELOG-v7.md`.

## Problem

Four of the next steps listed in #5242, each small on its own now that there is one tsconfig per package:

1. **`tsc` could not see that `@rjsf/react-bootstrap` does not load under Node.** `moduleResolution: bundler` accepts specifiers Node rejects. Importing main's `lib/index.js` with `node --input-type=module` fails on the extensionless `@react-icons/all-files/bs/BsPlus` (and, past that, on `react-bootstrap/Col`-style directory subpaths).
2. **Four exported `enum`s** (`TranslatableString`, `AdditionalItemsHandling` in utils, `GridType`, `Operators` in core) keep the source from running under Node's type stripping.
3. **`packages/utils/test` emitted declarations** into `lib-test/` for one reason: the three validator test projects referenced it as a project to reuse the shared schema suite.
4. **TypeScript 6.0.3.** 7.0.2 is out.

## Fix

**`module: nodenext`** in the base config, replacing `module: esnext` + `moduleResolution: bundler`. Of the projects in the root solution, only chakra-ui's source and test projects keep bundler resolution, with a comment saying why (`@chakra-ui/react` ships ESM declarations whose relative re-exports have no extension, so every named import from it fails under Node's rules).

The root and playground `package.json` gain `"type": "module"` so the playground is checked the same way as the published packages (its `"type": "commonjs"` made every file CommonJS under `nodenext`); `scripts/get-version-tag.js` becomes `.cjs` since it uses `require`. Verified with `pnpm start`: the page renders, the `AJV8_es` validator returns Spanish messages, and the 2019-09 and 2020-12 validators validate.

Test projects deliberately stay on `nodenext`. I first tried `bundler` for them, and `@cfworker/json-schema` split into two incompatible `Validator` declarations: TypeScript resolves a referenced project's files with that project's options, but in CommonJS mode when the referencing program's `module` is not `nodenext`, so the same import matched `require` in one file and `import` in another. Keeping every program on the same setting is what avoids that class of bug.

Source defects the flag found, all in shipped `lib/`:

- `@rjsf/react-bootstrap` imported `react-bootstrap/Col` (a directory with its own `package.json`, which Node's ESM loader refuses) and extensionless icon files. Components are now named imports from `react-bootstrap`, icons name their `.js` file. Verified: Node now imports its `lib/index.js`.
- `@rjsf/antd` type-imported the directory `antd/es/select`; now `antd/es/select/index.js`. Erased at runtime, so declarations only.
- `@rjsf/validator-ajv8` default-imported `ajv`, `ajv-formats` and `ajv/dist/standalone`, all CommonJS whose declarations say `export default`. Under Node's rules an ESM default import of a CommonJS module is `module.exports`, so TypeScript now types those as the whole module. `Ajv` is imported by the name ajv also exports at runtime; the other two are called through `.default`, which their runtime sets to the function itself, so the spelling is correct under Node, bundlers and TypeScript alike.

Test-only changes the flag required: JSON imports carry `with { type: 'json' }`, `userEvent` is imported by name (user-event's single `.d.ts` is CommonJS-flavoured, so its default import is the module object), ajv's `dist/2019` and `dist/2020` subpaths name their `.js` file and import the classes by name, and the `ajv-i18n` imports get an ambient `export =` declaration in each consumer, since its own declarations describe a default export its CommonJS runtime does not have.

**`erasableSyntaxOnly`**, with the four enums replaced by `as const` objects and same-named union types. `GridType.ROW`-style access and every value are unchanged (`AdditionalItemsHandling` keeps `0`, `1`, `2`), so no call site in the repo moved. It is still a public type change: a member used as a type must become `typeof GridType.ROW`, and the numeric enum's reverse mapping (`AdditionalItemsHandling[0]`) is gone. The changelog marks both packages as breaking, which is why this targets `v7`.

**`utils/test` is `noEmit`.** Each validator test project lists `../../utils/test/schema` and `../../utils/test/testUtils` in `include` and drops the project reference. The `./testing` subpath export #5242 floated does not get there: a composite project must list every file it compiles, so the files land in `include` either way, and the export would only add indirection. A private package would be a fourth project for 18 files.

**TypeScript 7.0.2.** Nothing in CI needed a change: the native compiler takes the same options, no code in the repo imports the TypeScript API that 7.0 dropped, and lint, knip, typecheck and the full test suite pass. The docs package's editor-only tsconfig extended `@tsconfig/docusaurus`, which sets `baseUrl`, removed in 7; Docusaurus's own `@docusaurus/tsconfig` sets it too, so its options are inlined without it (it is editor-only and not part of the root solution). The full test suite passed on the bare version bump; every test-side change in this PR came from `nodenext`, not from the compiler.

Emitted output differs from main only cosmetically: import specifiers keep the source's single quotes, `jsx-runtime` imports are sorted, `stableTypeOrdering` sorts union members in `.d.ts`, and inferred return types print through their alias (`ValidationData<T>` instead of its expansion).

## Effect

Same machine, clean outputs before each cold build, best of three. Main is `1a2bb662b`. Tarballs are `npm pack --dry-run` after a full `nx build`.

| | main | PR |
|---|---:|---:|
| TypeScript | 6.0.3 | 7.0.2 |
| tsconfig lines | 451 | 452 |
| `tsc -b packages/core`, cold | 2.6 s | 0.9 s |
| root `tsc --build` (sources + tests), cold | 21.7 s | 5.2 s |
| packages whose `lib/index.js` Node can `import()` | 11 / 16 | 12 / 16 |
| `@rjsf/react-bootstrap` tarball | 82.7 kB | 81.6 kB |
| `@rjsf/utils` tarball | 588.2 kB | 589.7 kB |
| `@rjsf/core` tarball | 497.4 kB | 497.7 kB |

`@rjsf/antd`, `@rjsf/mui` and `@rjsf/validator-ajv8` are within 0.1 kB. `dist/` is untouched.

The four packages Node still cannot import are outside what `tsc` can see, and unchanged here: `@rjsf/daisyui` imports a `.css` file (bundler-only by design); `@rjsf/primereact` imports `primereact/button`-style subpaths that are directories with their own `package.json`, which TypeScript resolves through `types` but Node refuses, and the `.esm.js` file inside has no matching `.d.ts`, so no spelling satisfies both; `@rjsf/fluentui-rc` fails inside `@fluentui/react-icons`' own extensionless chunk import; `@rjsf/snapshot-tests` calls `vi.mock` at module top level through Vitest's injected globals (`vi`, `describe`, `test`, `expect`), so outside a Vitest run `vi` is not defined; it is a test suite and only ever loads inside Vitest, but importing those names from `vitest` explicitly would drop its dependence on consumers enabling `globals: true`.

## Next steps

Items 1 and 5 of #5242 remain: let tsdown emit JavaScript so `tsc` emits declarations only, then let Nx infer `typecheck` and `build` from the tsconfigs. Both build on this topology and neither is small.


packages/react-bootstrap/src/AddButton/AddButton.tsx
```diff
@@ -1,7 +1,7 @@
-import { BsPlus } from '@react-icons/all-files/bs/BsPlus';
+import { BsPlus } from '@react-icons/all-files/bs/BsPlus.js';
 import type { FormContextType, IconButtonProps, RJSFSchema, StrictRJSFSchema } from '@rjsf/utils';
 import { TranslatableString } from '@rjsf/utils';
-import Button from 'react-bootstrap/Button';
+import { Button } from 'react-bootstrap';
```
heath-freenome, line 4: The idea behind this was to limit the size of the imports... Would importing `import Button from 'react-bootstrap/Button/index.js'` work as well?

CHANGELOG.md
```diff
@@ -25,14 +25,21 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/core
 
 - Fixed defaults not being restored when returning to an `anyOf` or `oneOf` option with disjoint properties ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
+- **Potentially breaking change:** `GridType` and `Operators` are now `as const` objects with same-named union types instead of `enum`s, so the source is valid under Node's type stripping. Values and `GridType.ROW`-style member access are unchanged; code that used a member as a type must write `typeof GridType.ROW` instead of `GridType.ROW`
```
heath-freenome: @jimmycallin I will be starting up the v7 long-lived branch soon. Maybe we can make this change on it instead? Maybe this whole PR belongs as part of v7?

CHANGELOG.md
```diff
@@ -25,14 +25,21 @@ should change the heading of the (upcoming) version to include a major version b
...
+
+## @rjsf/react-bootstrap
+
+- Fixed `lib/index.js` being unloadable by Node: it imported `react-bootstrap/Col`-style directory subpaths and extensionless `@react-icons/all-files` paths, which only bundlers resolve. Components are now imported from `react-bootstrap` itself and icon files by their full `.js` name
 
 ## @rjsf/utils
 
 - Fixed `sanitizeDataForNewSchema()` clearing existing arrays or preserving stale `undefined` values instead of retaining data or applying defaults for properties newly defined by the incoming schema ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
+- **Potentially breaking change:** `TranslatableString` and `AdditionalItemsHandling` are now `as const` objects with same-named union types instead of `enum`s, so the source is valid under Node's type stripping. Values (`AdditionalItemsHandling` keeps `0`, `1`, `2`) and `TranslatableString.ArrayItemTitle`-style member access are unchanged; code that used a member as a type must write `typeof TranslatableString.ArrayItemTitle`, and the numeric enum's reverse mapping (`AdditionalItemsHandling[0]`) no longer exists
```
heath-freenome: Ditto on v7 specific changes

CHANGELOG.md
jimmycallin: sure, let me know when v7 is ready and i'll retarget to it

heath-freenome:
@jimmycallin the v7 long-lived branch has been created. Please rebase this PR onto it. NOTE that a new `CHANGELOG-v7.md` was created for changes in that branch. It will get merged into `CHANGELOG.md` upon the merge of the branch back into `main`

packages/react-bootstrap/src/AddButton/AddButton.tsx
heath-freenome: @jimmycallin Thoughts?

jimmycallin: Unfortunately no. `react-bootstrap/Button/` is a directory containing only a `package.json` (`main: ../cjs/Button.js`, `module: ../esm/Button.js`), so there is no `index.js` there to import:

```
import('react-bootstrap/Button/index.js')  -> ERR_MODULE_NOT_FOUND
import('react-bootstrap/Button')           -> ERR_UNSUPPORTED_DIR_IMPORT
import('react-bootstrap/esm/Button.js')    -> ERR_MODULE_NOT_FOUND (./ThemeProvider)
import('react-bootstrap')                  -> ok
```

Even pointing straight at `esm/Button.js` fails, because react-bootstrap's own ESM output uses extensionless relative imports internally. Their ESM build is bundler-only; the root import is the only path Node can load (it falls through `main` to CJS since the package has no `exports` map).

Bundle size should be unaffected: react-bootstrap declares `"sideEffects": false`, so `import { Button } from 'react-bootstrap'` tree-shakes to the same thing as the subpath import in any bundler. The subpath form only ever mattered for bundlers that don't tree-shake barrels, which none of the current ones are.

packages/utils/vitest.config.ts
```diff
@@ -7,7 +7,6 @@ export default mergeConfig(base, {
     // Most utils are pure functions; the DOM-dependent test files opt back into
     // jsdom with a /** @vitest-environment jsdom */ pragma.
     environment: 'node',
-    exclude: ['node_modules/**', 'lib-test/**'],
     coverage: fullCoverage(),
```
heath-freenome, line 12: Removing the `'lib-test/**'` exclude means vitest will pick up stale compiled test artifacts if `lib-test/` exists on disk (produced by the old `emitDeclarationOnly`/`outDir: '../lib-test'` build in `packages/utils/test/tsconfig.json`).

Confirmed empirically: `packages/utils/lib-test/` is gitignored but currently exists in a fresh checkout with 250 compiled `*.test.js` files (e.g. `ErrorSchemaBuilder.test.js` importing `'../src'`). Any contributor or CI cache that built this directory before pulling this PR will have vitest discover and re-run every stale compiled duplicate alongside the real `.ts` tests — inflating counts, doubling runtime, and risking spurious pass/fail results that mask real regressions.

packages/validator-ajv8/test/tsconfig.json
```diff
@@ -1,15 +1,12 @@
 {
   "extends": "../../../tsconfig.test.json",
-  "include": ["./", "./**/*.json"],
+  "include": ["./", "./**/*.json", "../../utils/test/schema", "../../utils/test/testUtils"],
```
heath-freenome, line 3: `validator-ajv8`, `validator-ata`, and `validator-cfworker` test tsconfigs each now directly `include` the same `packages/utils/test/schema` and `testUtils` source (duplicated verbatim across all three) instead of referencing one shared compiled project.

A full `tsc --build` re-typechecks the same shared source three times instead of once, which will scale worse as the shared suite grows (a tradeoff acknowledged in the PR description). Separately, adding a new shared test directory under `packages/utils/test`, or a new validator package, requires updating all three include lists in lockstep — missing one silently drops that validator's coverage of the shared suite rather than failing loudly.

packages/playground/src/app.tsx
```diff
@@ -10,15 +10,16 @@ import { Theme as shadcnTheme } from '@rjsf/shadcn';
...
+import { Ajv2019 } from 'ajv/dist/2019.js';
+import { Ajv2020 } from 'ajv/dist/2020.js';
 
 import type { PlaygroundProps } from './components/index.ts';
 import Playground from './components/index.ts';
 import Layout from './layout/index.ts';
 
-const esV8Validator = customizeValidator({}, localize_es);
+// ajv-i18n's declarations describe an ES module with a default export, but its CommonJS runtime is the localizer function itself
+const esV8Validator = customizeValidator({}, localize_es as unknown as typeof localize_es.default);
```
heath-freenome: The ajv-i18n CJS/ESM interop cast (`X as unknown as typeof X.default`) is duplicated verbatim here and in `packages/validator-ajv8/test/validator.test.ts`, with a near-identical explanatory comment, instead of being factored into one shared helper.

If ajv-i18n's typings or the interop convention change later, both call sites need to be found and fixed independently — a maintainer updating one copy can easily miss the other.

packages/docs/tsconfig.json
```diff
@@ -1,7 +1,19 @@
 {
   // This file is not used in compilation. It is here just for a nice editor experience.
-  "extends": "@tsconfig/docusaurus/tsconfig.json",
+  // Docusaurus's own preset (@docusaurus/tsconfig) still sets `baseUrl`, which TypeScript 7 removed, so its
+  // options are inlined here without it.
   "compilerOptions": {
```
heath-freenome, line 5: Inlining `@tsconfig/docusaurus`'s options here (to drop the TS7-incompatible `baseUrl`) also dropped `esModuleInterop` and the `types` array (`node`, `@docusaurus/module-type-aliases`, `@docusaurus/theme-classic`) that the preset set, with no replacement.

`packages/docs/src/components/HomepageFeatures/index.tsx` imports `'./styles.module.css'`; that ambient `*.module.css` module declaration comes from `@docusaurus/module-type-aliases`, which is no longer in `types`. Since this tsconfig is editor-only, the practical effect is a spurious "Cannot find module" error in editors for that file and any other docs file relying on Docusaurus's ambient asset/module types.

heath-freenome (commented):
Automated review findings — see inline comments.

packages/utils/vitest.config.ts
jimmycallin: Restored the exclude with a comment saying why (stale checkouts). Verified with a throwing `lib-test/stale.test.js`: not discovered.

packages/docs/tsconfig.json
jimmycallin: Added `"types": ["node", "@docusaurus/module-type-aliases"]`. Verified against `HomepageFeatures/index.tsx`: without them it reports TS2307 on `styles.module.css` and TS2591 on `require`; with them, zero errors. Left out `@docusaurus/theme-classic` since it is not a dependency of the docs package (the preset listed it but it never resolved here). `esModuleInterop` stays out: it is a no-op under `moduleResolution: bundler` with `module: esnext` for type-checking a `noEmit` project, and adding it back did not change the error count.

packages/validator-ajv8/test/tsconfig.json
jimmycallin: Keeping this as is. The three lists are 18 files re-checked three times, and the whole root `tsc --build` (sources and tests) is 5.2 s, so there is nothing to win yet. The silent-drop claim does not hold: with `composite`, a validator test that imports a shared file not in its `include` fails with TS6307 ("File ... is not listed within the file list of project"). Verified by removing `../../utils/test/schema` from validator-ata's include and running `tsc -b`. And vitest runs the tests regardless of tsconfig, so coverage of the shared suite does not depend on these lists at all. A shared `extends` base cannot carry the paths either: a child's `include` replaces the parent's, so each project would still list them.

packages/playground/src/app.tsx
jimmycallin: Leaving these as two casts. They are not the same value: the playground casts one localizer (`ajv-i18n/localize/es`), the test casts the whole localizer map (`ajv-i18n`), so a helper would be a generic identity function `<T>(m: T) => m as unknown as T['default']`. The only package both sites depend on is `@rjsf/validator-ajv8`, so the helper would become a public runtime export whose sole purpose is to paper over ajv-i18n's typings, in exchange for removing one line in a test.

jimmycallin: Went further than deduplicating: both casts are gone in 70687d8. ajv-i18n's problem is that its `.d.ts` says `export default` while its runtime assigns straight to `module.exports`. Each consumer now carries a small ambient `declare module` with `export =` describing the real shape, so the default import is typed as the function without any cast. That also made the pre-existing `as Localizer` in the test unnecessary.

packages/playground/src/ajv-i18n.d.ts
```diff
@@ -0,0 +1,8 @@
+// ajv-i18n's own declarations describe an ES module with a default export, but its CommonJS runtime assigns the
+// localizer straight to module.exports, so under Node's rules the default import would be typed as the module object.
+declare module 'ajv-i18n/localize/es/index.js' {
```
heath-freenome, line 3: We probably want to file a bug against AJV-8 for this

heath-freenome (approved):

PR #5245, opened 2026-09-05
fix(utils): populate dependency defaults without form data

### LINKED ISSUE #5198: Defaults inside `dependencies` subschemas are not populated by `getDefaultFormState`

sjh9714:

Fixes #5198.

When `getDefaultFormState()` receives no form data, its initial schema retrieval removes unresolved dependencies before property defaults can select a dependent branch. Compute defaults from the unresolved schema only for that case, while preserving the existing retrieved-schema path for supplied form data and schemas without dependencies.

The regression test covers a property default that selects a `oneOf` dependency containing another default.

### Testing

- `pnpm --dir packages/utils exec vitest run test/schema.test.ts -t 'should populate dependency defaults when formData is undefined' --coverage.enabled=false`
- `pnpm --dir packages/utils test`
- `pnpm --dir packages/utils cs-check`
- `pnpm --dir packages/utils lint`
- `pnpm --filter @rjsf/utils build`
- `pnpm exec nx run-many --target=build --exclude=@rjsf/docs`

### Checklist

  - [x] I've added and run the focused regression test and the full utils test suite.
  - [x] Documentation updates are not needed for this bug fix.
  - [x] I've updated the changelog with a description of the fix.


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ should change the heading of the (upcoming) version to include a major version b
 ## @rjsf/utils
 
 - Fixed `sanitizeDataForNewSchema()` clearing existing arrays or preserving stale `undefined` values instead of retaining data or applying defaults for properties newly defined by the incoming schema ([#3736](https://github.com/rjsf-team/react-jsonschema-form/issues/3736))
+- Fixed defaults from a dependency subschema being omitted when `getDefaultFormState()` is called without form data ([#5198](https://github.com/rjsf-team/react-jsonschema-form/issues/5198))
```
heath-freenome: Move this to a new 6.10.1 section please since we just released 6.10.0

heath-freenome:
Reviewed this diff (reproduced directly against the PR's code in an isolated worktree, compared against a `main` baseline). Found two correctness issues in the fix at `packages/utils/src/schema/getDefaultFormState.ts:865`:

**1. Fix drops `allOf`-derived defaults when the schema also has `dependencies`**

When `formData` is undefined and the schema has top-level `dependencies`, the fix substitutes the raw, unretrieved `theSchema` for the resolved `schema` before calling `computeDefaults`. But `computeDefaults` never resolves `allOf` itself (only `retrieveSchema` does), so any `allOf`-derived properties/defaults are silently dropped when a schema combines top-level `allOf` with top-level `dependencies`.

Repro:
```js
const schema = {
  type: 'object',
  allOf: [{ properties: { fromAllOf: { type: 'string', default: 'A' } } }],
  properties: { type: { type: 'integer', default: 0 } },
  dependencies: { type: { oneOf: [/* ... */] } },
};
getDefaultFormState(validator, schema); // no formData
```
- On `main` (pre-PR): `{ type: 0, fromAllOf: 'A' }` — dependency default missing (the bug being fixed).
- On this PR: `{ type: 0, value: 5 }` — dependency default now populated, but `fromAllOf` is now silently missing. New regression for any schema mixing top-level `allOf` and `dependencies`.

**2. Fix's `dependencies` check misses schemas reached via `$ref`**

The guard `DEPENDENCIES_KEY in theSchema` checks only the schema as literally passed in, so it does nothing when the schema with `dependencies` is only reachable through a `$ref` (e.g. `{ $ref: '#/definitions/dep' }`). The original bug (#5198) this PR fixes remains unfixed for this common ref/definitions pattern.

Repro:
```js
const schema = {
  $ref: '#/definitions/dep',
  definitions: {
    dep: {
      type: 'object',
      properties: { type: { type: 'integer', default: 0 } },
      dependencies: {
        type: {
          oneOf: [
            { properties: { type: { enum: [0] }, value: { type: 'integer', default: 5 } }, required: ['value'] },
            { properties: { type: { enum: [1] } } },
          ],
        },
      },
    },
  },
};
getDefaultFormState(validator, schema, undefined, schema);
```
Result: `{ type: 0 }` — `value: 5` is still missing, i.e. the same defect this PR claims to fix still reproduces when the dependency-bearing schema sits behind a `$ref` instead of being inlined at the top level.

sjh9714:
Fixed both cases and added regression tests. allOf defaults are preserved, and dependencies now work through $ref.

sjh9714:
Could you approve the workflow runs?

heath-freenome:
Follow-up review after the latest changes. Reproduced directly against the current PR branch in a worktree (existing tests + 100% coverage still pass; added a targeted scratch test to confirm the gap below).

**1. `dependencies` nested inside an `allOf` branch is still dropped**

The current fix (`packages/utils/src/schema/getDefaultFormState.ts:865`) only reattaches `dependencies` found via `resolveAllReferences`, which does not merge `allOf`. So a `dependencies` keyword declared inside an `allOf` branch (rather than directly on the schema/`$ref` target) is still silently dropped when `formData` is undefined — bug #5198 is only partially fixed.

Repro:
```js
const schema = {
  type: 'object',
  allOf: [
    {
      properties: { type: { default: 0 } },
      dependencies: {
        type: {
          oneOf: [{ properties: { type: { enum: [0] }, value: { default: 5 } }, required: ['value'] }],
        },
      },
    },
  ],
};
getDefaultFormState(validator, schema, undefined, schema);
```
Result: `{ type: 0 }` instead of `{ type: 0, value: 5 }`.

**2. Redundant full schema walk on every call (efficiency)**

`resolveAllReferences(theSchema, ...)` now runs unconditionally whenever `formData` is undefined, duplicating the reference resolution `retrieveSchema()` (line 860) already performs internally via the same function — even when the schema has no `$ref` or `dependencies` at all.

This path is hit on a common interactive flow: `ArrayField.tsx:171` calls `schemaUtils.getDefaultFormState(itemSchema)` (no `formData`) every time a user adds an array item, so large/deeply-nested item schemas now get walked twice per click.

sjh9714:
I added tests for your example, nested allOf and $ref inside allOf. They now pass. I also removed the extra reference walk by keeping dependencies until the defaults are computed.

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -857,7 +857,18 @@ export default function getDefaultFormState<
...
+  const [schema] = retrieveSchemaInternal<T, S, F>(
+    validator,
+    theSchema,
+    rootSchema ?? ({} as S),
+    formData,
+    undefined,
+    undefined,
+    experimental_customMergeAllOf,
+    undefined,
+    formData === undefined,
```
heath-freenome, line 870: **Correctness (confirmed):** the fix only preserves dependency-branch defaults when `formData` is strictly `undefined`. `formData` of `{}` (a very common "no data yet" convention) still hits the original bug (#5198).

Repro: `getDefaultFormState(validator, schema, {}, schema)` with a root-level `dependencies` keyed on a property whose default selects a `oneOf` branch — the exact schema shape from this PR's own added test, just called with `{}` instead of `undefined` — returns `{"type":0}` instead of `{"type":0,"value":5}`. Verified by direct execution against this PR's worktree.

Real-world impact: `packages/core/src/components/Form.tsx` computes `defaultsFormData` from `props.formData`, which is commonly initialized to `{}` by consumers, so those forms keep hitting the bug this PR claims to fix.

packages/utils/src/schema/retrieveSchema.ts
```diff
@@ -93,6 +93,7 @@ function normalizeBooleanSchema<S extends StrictRJSFSchema = RJSFSchema>(schema:
  * @param recurseList - The list of recursive references already processed
  * @param [formData] - The current formData to assist retrieving a schema
  * @param [experimental_customMergeAllOf] - Optional function that allows for custom merging of `allOf` schemas
+ * @param [preserveDependencies=false] - Leave dependencies unresolved for default computation
  * @returns - A list of schemas with the appropriate conditions resolved, possibly with all branches expanded
  */
 export function resolveCondition<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(
```
heath-freenome: **Simplification (plausible):** this threads a new `preserveDependencies` boolean through the entire shared schema-resolution pipeline (`resolveCondition`/`resolveSchema`/`resolveReference`/`retrieveSchemaInternal`, used by every consumer including `SchemaField`/`ObjectField` rendering), when only `getDefaultFormState`'s single call site needs the behavior change.

Every future recursive call site added to this file must remember to forward `preserveDependencies` positionally (already ~10 call sites do); one that's missed silently reverts to eager dependency resolution mid-recursion. A narrower fix confined to `getDefaultFormState.ts` (which already has its own `DEPENDENCIES_KEY` handling in `computeDefaults`) would avoid adding a permanent knob to general-purpose infrastructure that every other caller of `retrieveSchema` must now reason about.

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -857,7 +857,18 @@ export default function getDefaultFormState<
   if (!isObject(theSchema)) {
     throw new Error(`Invalid schema: ${theSchema}`);
   }
-  const schema = retrieveSchema<T, S, F>(validator, theSchema, rootSchema, formData, experimental_customMergeAllOf);
+  // Without formData, dependency conditions need the defaults that computeDefaults will generate.
+  const [schema] = retrieveSchemaInternal<T, S, F>(
```
heath-freenome: **Code quality (plausible):** the previous one-line `retrieveSchema(...)` wrapper call was replaced with a 9-argument positional `retrieveSchemaInternal(...)` call containing multiple bare `undefined` placeholders just to reach the new trailing `preserveDependencies` slot.

Harder to read/maintain than the previous wrapper call; a future parameter added to `retrieveSchemaInternal` risks silent positional misalignment here since three of the nine arguments are unlabeled `undefined`.

heath-freenome (commented):
Automated review found two correctness issues

One correctness finding couldn't be attached inline because it's in code this PR doesn't touch: in `getDefaultFormState.ts`, `computeDefaults`'s `DEPENDENCIES_KEY` branch (around line 877, downstream of the new `retrieveSchemaInternal` call added at line 861) passes the raw, possibly-`undefined` `rootSchema` into `resolveDependencies` → `withExactlyOneSubschema` → `retrieveSchemaInternal` → `resolveAllReferences` → `findSchemaDefinition`. If `rootSchema` is omitted and a dependency branch's default resolution hits a `$ref`, this now **throws** (`Could not find a definition for #/...`) instead of degrading gracefully as it did before this PR (verified by direct execution against both this branch and `main`). The new call at line 861 already guards its own `rootSchema` with `?? ({} as S)`, but that fallback schema isn't what flows into `computeDefaults` — the original, unguarded `rootSchema` still does.

Repro:
```js
getDefaultFormState(validator, {
  type: 'object',
  definitions: { extra: { type: 'string', default: 'RefDefault' } },
  properties: { type: { type: 'integer', default: 0 } },
  dependencies: {
    type: { oneOf: [{ properties: { type: { enum: [0] }, extraField: { $ref: '#/definitions/extra' } } }] },
  },
}, undefined, undefined); // rootSchema omitted -> throws on this branch, returns {type:0} on main
```

See the inline comment for the other finding (formData `{}` not covered by the fi

heath-freenome:
@sjh9714 Just a few more correctness fixes and a rebase/conflict resolution and we can merge this

heath-freenome:
@sjh9714 Hopefully we can get this merged before the weekend

sjh9714:
An empty object now gets the dependency defaults, and leaving out the root schema no longer throws. I added tests for both and resolved the conflict with main.

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -857,13 +857,25 @@ export default function getDefaultFormState<
...
+    validator,
+    theSchema,
+    rootSchema ?? ({} as S),
+    formData,
+    undefined,
+    undefined,
+    experimental_customMergeAllOf,
+    undefined,
+    formData === undefined || (isObject(formData) && Object.keys(formData).length === 0),
```
heath-freenome: **Simplification:** `formData === undefined || (isObject(formData) && Object.keys(formData).length === 0)` duplicates the identical `hasNoExistingData` expression already computed elsewhere in this file (in `computeDefaults`) for the same "is form data effectively empty" purpose. Worth extracting into a shared helper (e.g. `isEmptyFormData(formData)`) so the two copies can't silently drift apart on a future edit to one but not the other.

packages/utils/src/schema/getDefaultFormState.ts
```diff
@@ -857,13 +857,25 @@ export default function getDefaultFormState<
...
+  );
 
   // Get the computed defaults with 'shouldMergeDefaultsIntoFormData' set to true to merge defaults into formData.
   // This is done when for example the value from formData does not exist in the schema 'enum' property, in such
   // cases we take the value from the defaults because the value from the formData is not valid.
   const defaults = computeDefaults<T, S, F>(validator, schema, {
-    rootSchema,
+    // Preserved dependencies may still contain references to the original schema.
+    rootSchema: rootSchema ?? (DEPENDENCIES_KEY in schema ? theSchema : undefined),
```
heath-freenome: **Correctness (plausible, pre-existing gap not a regression):** this fallback is keyed on `DEPENDENCIES_KEY in schema` at the resolved schema's immediate top level, rather than being a general "rootSchema omitted → default to theSchema" fix. A `oneOf`/`anyOf` wrapper whose selected branch has `dependencies`+`$ref` still throws with `rootSchema` omitted, since `DEPENDENCIES_KEY in schema` is false here (oneOf/anyOf resolution happens later, inside `computeDefaults`, using whatever `rootSchema` was already decided at this point). Confirmed by execution this same scenario throws identically on `main`, so it's not new — but an unconditional `rootSchema ?? theSchema` fallback would cover it at the same depth as the rest of this fix, rather than special-casing on one keyword.

heath-freenome (commented):
Follow-up review on the latest commit (`8cd3b7279`). Reproduced directly against this PR branch in an isolated worktree: full `packages/utils` suite (1577 tests) passes, clean `tsc -b` typecheck.

**All 6 previously-raised issues verified fixed:**
1. `formData={}` now gets dependency defaults, same as `undefined` ✅
2. Omitting `rootSchema` with a `$ref` inside a dependency branch no longer throws ✅
3. `allOf`-derived defaults preserved alongside `dependencies` ✅
4. Dependencies reachable via `$ref` ✅
5. Dependencies nested inside `allOf` ✅
6. Redundant reference walk removed ✅

Two smaller notes from this round below (inline) — one duplicated-logic simplification, and one residual (pre-existing, not a regression) gap in how general the `rootSchema` fallback is. Nice work chasing all these edge cases down.

heath-freenome:
@sjh9714 just two small fixes and we'll merged

sjh9714:
Shared the empty data check and added tests for oneOf and anyOf. Dependency references now resolve for empty data without rootSchema. I kept the fallback conditional because making it unconditional breaks an existing AJV 2020 array test.

heath-freenome (approved):

Issue #4856, opened 2025-11-06
Dark mode on the playground


JBSopra:
### Prerequisites

- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)

### What theme are you using?

mui

### Is your feature request related to a problem? Please describe.

N/A

### Describe the solution you'd like

Hello,  Please, if not already considered, can you add dark mode to the playground?

### Describe alternatives you've considered

N/A

heath-freenome:
@JBSopra Since the playground is primarily using the Bootstrap 3 theme, dark mode could be challenging. We've also mixed in `Mui` for some components, so maybe it won't be too hard for someone to use the Mui theme logic to enable dark mode? Unfortunately there are many more issues that the 2 main contributors to the library need to focus on. Are you able to do the work?

JBSopra:
> [@JBSopra](https://github.com/JBSopra) Since the playground is primarily using the Bootstrap 3 theme, dark mode could be challenging. We've also mixed in `Mui` for some components, so maybe it won't be too hard for someone to use the Mui theme logic to enable dark mode? Unfortunately there are many more issues that the 2 main contributors to the library need to focus on. Are you able to do the work?

I could certainly have a look when i find the time myself. I understand regarding the other issues.

GraceyDugar:
Hi! I'm new to open source and would like to work on this. @JBSopra, are you still planning to, or is it OK if I take it? My plan is to start with dark mode for the playground UI (toggle + system preference) using MUI's theme support, and leave the form preview theming for a follow-up. @heath-freenome, does that scope sound right?

heath-freenome:
> Hi! I'm new to open source and would like to work on this. [@JBSopra](https://github.com/JBSopra), are you still planning to, or is it OK if I take it? My plan is to start with dark mode for the playground UI (toggle + system preference) using MUI's theme support, and leave the form preview theming for a follow-up. [@heath-freenome](https://github.com/heath-freenome), does that scope sound right?

@GraceyDugar Given that this has been sitting idle for over 10 months I would say that you working on it is fine. I'm interested to see what you come up with.

JBSopra:
@GraceyDugar  Absolutely go ahead, had no time during the work day for this unfortunately

PR #5246, opened 2026-09-05
feat: add propertyNames enum support for additionalProperties (#4682)

### LINKED ISSUE #4682: Use `enum` to populate dropdown list for object key
### Prerequisites

### What theme are you using?

N/A

We have the following JSON schema structure:
```json
{
  "additionalProperties": {
    "type": "string"
  },
  "propertyNames": {
    "enum": [
      "a",
      "b",
      "c",
      "d"
    ]
  },
  "type": "object"
}
```
The object keys have an enumerated list of allowed values. 

`react-jsonschema-form` picks this `enum` correctly up in the validation:
Image

However, as an end user it is difficult to understand from the generated user interface what is wrong, and what the allowed values are.

When key is an `enum` with a defined list of allowed values, show it as a dropdown instead of a text input field, in order to support the user in filling in valid values.

Add the allowed values in the `enum` to the field description, such that the user have it documented. However, this has the risk of the description and the JSON schema `enum` field getting out of sync, and in addition it is quicker for the user to select a valid value from dropdown instead of parsing the description, and writing in a valid string.

CJstate:
## Summary
Implements support for the `propertyNames` schema keyword with `enum` constraint in `additionalProperties` objects. When `propertyNames.enum` is defined, the key input for additional properties now renders as a dropdown (SelectWidget) instead of a text input, allowing users to select from predefined key names.

## Changes
- **@rjsf/utils**: Added `propertyNamesEnum` to `WrapIfAdditionalTemplateProps` and `FieldTemplateProps` with conditional type inference
- **@rjsf/core**: Updated `ObjectField`, `SchemaField`, and core `WrapIfAdditionalTemplate` to propagate `propertyNamesEnum` through the component chain
- **All 10 theme packages**: Updated `WrapIfAdditionalTemplate` in mui, react-bootstrap, chakra-ui, daisyui, fluentui-rc, mantine, primereact, semantic-ui, shadcn to conditionally render SelectWidget when `propertyNamesEnum` is available
- **Playground**: Added dark mode toggle for MUI theme

## Testing
- All 3800+ tests pass across all packages
- Lint passes with no new errors

## Related
Closes #4682
Addresses #4856 (dark mode toggle)

### DIFF
--- a/packages/core/src/components/fields/ObjectField.tsx
+++ b/packages/core/src/components/fields/ObjectField.tsx
@@ -97,6 +97,8 @@ interface ObjectFieldPropertyProps<
   handleKeyRename: (oldKey: string, newKey: string) => void;
   /** Callback that handles the removal of an additionalProperties-based property with key */
   handleRemoveProperty: (keyName: string) => void;
+  /** The enum values from the parent schema's propertyNames, if available */
+  propertyNamesEnum?: S extends { propertyNames: { enum: infer E } } ? E : never;
 }
 
 /** The `ObjectFieldProperty` component is used to render the `SchemaField` for a child property of an object
@@ -205,6 +207,7 @@ function ObjectFieldPropertyFn<T = any, S extends StrictRJSFSchema = RJSFSchema,
       disabled={disabled}
       readonly={readonly}
       hideError={hideError}
+      propertyNamesEnum={props.propertyNamesEnum}
     />
   );
 }
@@ -420,6 +423,11 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
       const addedByAdditionalProperties = isAdditionalPropertySchema(schema.properties?.[propertyName]);
       const fieldUiSchema = addedByAdditionalProperties ? uiSchema.additionalProperties : uiSchema[propertyName];
       const hidden = getUiOptions<T, S, F>(fieldUiSchema).widget === 'hidden';
+      const propertyNamesSchema = schema.propertyNames;
+      const propertyNamesEnum =
+        propertyNamesSchema && typeof propertyNamesSchema === 'object' && 'enum' in propertyNamesSchema
+          ? propertyNamesSchema.enum
+          : undefined;
       const content = (
         <ObjectFieldProperty<T, S, F>
           key={getStableKey(propertyName)}
@@ -440,6 +448,7 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
           disabled={disabled}
           readonly={readonly}
           hideError={hideError}
+          propertyNamesEnum={propertyNamesEnum as any}
         />
       );
       return {
--- a/packages/utils/src/types.ts
+++ b/packages/utils/src/types.ts
@@ -657,6 +657,8 @@ export type FieldTemplateProps<
   onKeyRenameBlur: (event: FocusEvent<HTMLInputElement>) => void;
   /** Callback used to handle the removal of the additionalProperty */
   onRemoveProperty: () => void;
+  /** The enum values from the parent schema's propertyNames, if available */
+  propertyNamesEnum?: S extends { propertyNames: { enum: infer E } } ? E : never;
 };
 
 /**
@@ -934,6 +936,8 @@ export type WrapIfAdditionalTemplateProps<
 > = RJSFBaseProps<T, S, F> & {
   /** The field or widget component instance for this field row */
   children: ReactNode;
+  /** The enum values from the parent schema's propertyNames, if available */
+  propertyNamesEnum?: S extends { propertyNames: { enum: infer E } } ? E : never;
 } & Pick<
     FieldTemplateProps<T, S, F>,
     | 'id'
--- a/packages/chakra-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
+++ b/packages/chakra-ui/src/WrapIfAdditionalTemplate/WrapIfAdditionalTemplate.tsx
@@ -1,6 +1,7 @@
+import { useCallback } from 'react';
 import { Grid, GridItem, Input } from '@chakra-ui/react';
 import type { FormContextType, RJSFSchema, StrictRJSFSchema, WrapIfAdditionalTemplateProps } from '@rjsf/utils';
-import { ADDITIONAL_PROPERTY_FLAG, buttonId, TranslatableString } from '@rjsf/utils';
+import { ADDITIONAL_PROPERTY_FLAG, buttonId, TranslatableString, getWidget } from '@rjsf/utils';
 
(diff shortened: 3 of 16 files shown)


packages/utils/src/types.ts
```diff
@@ -934,6 +936,8 @@ export type WrapIfAdditionalTemplateProps<
 > = RJSFBaseProps<T, S, F> & {
   /** The field or widget component instance for this field row */
   children: ReactNode;
+  /** The enum values from the parent schema's propertyNames, if available */
+  propertyNamesEnum?: S extends { propertyNames: { enum: infer E } } ? E : never;
```
heath-freenome, line 940: Since you already added this to the FieldTemplateProps, just add it to the pick list below

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -440,6 +448,7 @@ export default function ObjectField<T = any, S extends StrictRJSFSchema = RJSFSc
           disabled={disabled}
           readonly={readonly}
           hideError={hideError}
+          propertyNamesEnum={propertyNamesEnum as any}
```
heath-freenome, line 451: Do not cast this to any, use the proper type for it's definition.

packages/utils/src/types.ts
```diff
@@ -657,6 +657,8 @@ export type FieldTemplateProps<
   onKeyRenameBlur: (event: FocusEvent<HTMLInputElement>) => void;
   /** Callback used to handle the removal of the additionalProperty */
   onRemoveProperty: () => void;
+  /** The enum values from the parent schema's propertyNames, if available */
+  propertyNamesEnum?: S extends { propertyNames: { enum: infer E } } ? E : never;
```
heath-freenome, line 661: Consider creating an explicit type for this since you are specifying it in many places to DRY up the type

packages/core/src/components/fields/ObjectField.tsx
```diff
@@ -97,6 +97,8 @@ interface ObjectFieldPropertyProps<
   handleKeyRename: (oldKey: string, newKey: string) => void;
   /** Callback that handles the removal of an additionalProperties-based property with key */
   handleRemoveProperty: (keyName: string) => void;
+  /** The enum values from the parent schema's propertyNames, if available */
+  propertyNamesEnum?: S extends { propertyNames: { enum: infer E } } ? E : never;
```
heath-freenome, line 101: Use a type explicitly exported from `@rjsf/utils` per my recommendation below

heath-freenome:
Thanks for tackling `propertyNames.enum` support! One structural concern plus two correctness bugs I found while looking at the `SelectWidget : <native input>` pattern that's now duplicated across all 10 `WrapIfAdditionalTemplate` copies (core + 9 themes).

## Bugs in the duplicated block

1. **`getWidget(widgets, 'SelectWidget')` uses the wrong signature.** `getWidget`'s actual signature is `getWidget(schema, widget, registeredWidgets)` (`packages/utils/src/getWidget.tsx`). This call passes the widgets map as `schema` and omits `registeredWidgets` (defaults to `{}`), so it throws (`No widget 'SelectWidget' for type 'undefined'...`) as soon as `propertyNamesEnum` is set — the dropdown path is never actually reachable at runtime. It should be a direct registry lookup instead: `registry.widgets.SelectWidget`. No existing test exercises this path, which is how it got through green CI.
2. **`as any` on its own lines** (e.g. in the mui/core/etc. diffs, right before `registry={registry as any}`) isn't a type cast — in JSX that becomes two stray boolean props `as={true} any={true}` on the widget. Harmless at runtime but dead code from what looks like a broken find/replace.

Minor: `onFocus={handleSelectBlur}` wires the *commit* handler to focus, which would fire a key-rename as soon as the field gains focus, before anything is selected.

## DRY proposal

The following logic is byte-for-byte identical in all 10 files: resolving the SelectWidget, mapping `propertyNamesEnum` to `enumOptions`, and adapting `onKeyRenameBlur`'s `(event) => void` signature to the widget's `(id, value) => void` signature. Suggest pulling it into a single helper in `@rjsf/utils` (plain function, not a hook — nothing here needs memoization, so it avoids rules-of-hooks and can early-return):

```ts
// packages/utils/src/getAdditionalPropertyKeySelectProps.ts
export interface AdditionalPropertyKeySelectResult<T, S, F> {
  Widget: Widget<T, S, F>;
  props: WidgetProps<T, S, F>;
}

export default function getAdditionalPropertyKeySelectProps<T, S, F extends FormContextType>(options: {
  id: string;
  label: string;
  keyLabel?: string;
  required?: boolean;
  disabled?: boolean;
  readonly?: boolean;
  propertyNamesEnum?: unknown[];
  onKeyRenameBlur: (event: FocusEvent<HTMLInputElement>) => void;
  registry: Registry<T, S, F>;
  uiSchema?: UiSchema<T, S, F>;
}): AdditionalPropertyKeySelectResult<T, S, F> | undefined {
  const { id, label, keyLabel, propertyNamesEnum, onKeyRenameBlur, registry, uiSchema, ...rest } = options;
  if (!propertyNamesEnum?.length) return undefined;

  const Widget = registry.widgets.SelectWidget; // direct lookup — fixes the getWidget misuse
  const commit = (value: unknown) => onKeyRenameBlur({ target: { value } } as any);

  return {
    Widget,
    props: {
      id,
      name: id,
      value: label,
      ...rest,
      options: { enumOptions: propertyNamesEnum.map((value) => ({ value, label: String(value) })), enumDisabled: [] },
      onChange: commit,
      onBlur: (_id: string, value: unknown) => commit(value),
      schema: { type: 'string', enum: propertyNamesEnum },
      registry,
      uiSchema,
      label: keyLabel,
    },
  };
}
```

Each theme's `WrapIfAdditionalTemplate` then shrinks to:

```tsx
const additionalKeySelect = getAdditionalPropertyKeySelectProps({
  id: `${id}-key`,
  label,
  keyLabel,
  required,
  disabled: disabled || readonly,
  readonly,
  propertyNamesEnum,
  onKeyRenameBlur,
  registry,
  uiSchema,
});

...

{additionalKeySelect ? (
  <additionalKeySelect.Widget key={label} {...additionalKeySelect.props} />
) : (
  <Input key={label} defaultValue={label} disabled={disabled || readonly} id={`${id}-key`} name={`${id}-key`}
    onBlur={!readonly ? onKeyRenameBlur : undefined} type='text' mb={1} />
)}
```

This cuts the added lines per theme from ~55 to ~15, keeps each theme's native fallback input untouched and readable, and — importantly — fixes the `getWidget` bug in exactly one place instead of 10.

Happy to help push this through if useful.

heath-freenome:
@CJstate In addition to all the above feedback, you also need to update CHANGELOG.md to add the appropriate comments to 6.10.1

heath-freenome:
Closed by https://github.com/rjsf-team/react-jsonschema-form/pull/5321

PR #5247, opened 2026-09-05
build: bundle lib/ and dist/ with tsdown; tsc only typechecks

jimmycallin:
## Problem

Item 1 of the next steps in #5242: each package build is four steps, `tsc -b` (emit `lib/` + typecheck), two `esbuild` invocations and a `rollup` pass whose only job is UMD. `tsc` is both the typechecker and the JavaScript emitter, so `tsc --build` from the root and the package builds both write into `lib/`.

## Fix

Each package build is one step:

```
tsdown -c ../../tsdown.base.mts
```

One shared `tsdown.base.mts` drives every package, no per-package configs. It only transpiles; typechecking is the existing root `pnpm run typecheck` (`tsc --build`):

1. emits `lib/`: per-file ESM **and declarations**, mirroring `src/` so the `./lib/*.js` deep-import exports keep resolving. Declarations come from tsdown's tsc-based generator, so the published types are what tsc inferred before, JSDoc included.
2. emits `dist/`: the CJS, ESM and UMD bundles at their existing paths.

**tsc emits no JavaScript any more.** The package tsconfigs stay composite with `outDir: lib` and references (the topology from #5242), so editors and `tsc --build` keep working, but they are `emitDeclarationOnly`: the root typecheck can only rewrite the declarations with equivalent content, never the JavaScript. snapshot-tests is the one package still built by tsc; it drops its project references because `tsc -b` followed them and re-emitted core/utils/validator-ajv8's `lib/` over tsdown's output (Nx already orders it after them). The build no longer produces a tsbuildinfo, so it leaves the Nx build outputs.

## Details

- Bundle and UMD global names derive from `package.json`, with a 3-entry table for the names that predate the convention (core's `JSONSchemaForm`, fluentui-rc's copy-pasted `core.umd.js`, shadcn's `rjsf-shadcn`).
- `deps.neverBundle: true` reproduces esbuild's `--packages=external`; `comments.jsdoc: false` reproduces its comment stripping in the bundles.
- **One intentional difference:** the guessed browser-global names for externals inside the UMD `<script>`-tag branch are rolldown's (`fast_equals`) instead of rollup's (`fastEquals`). Neither matches the globals those libraries actually expose, so that branch was never usable.
- Re-export-only `index.ts` modules no longer get a source map. There is nothing in them to map.
- `nx.json`: the shared config joins the build target's inputs so a warm cache cannot serve stale bundles when it changes. `knip.jsonc`: ignores `src/index.ts` in chakra-ui, react-bootstrap and shadcn (pre-existing `Form`/`default` duplicate export that the old `esbuild ./src/index.ts` script line happened to mask).
- `rollup` + its platform binaries leave the lockfile. `esbuild` stays as a transitive dependency of `tsx` and `@size-limit/esbuild`.
- shadcn `build:css` (separate commit): `build-css.ts` spawned `npx @tailwindcss/cli` per theme, serially, and npx's package resolution was ~2.1 s of each ~2.25 s. It now calls the `tailwindcss` binary directly and compiles the ten themes in parallel: ~25 s → ~1 s, identical output.

## Verification

Against a clean build of main (`1a2bb662b`), same machine:

- every package emits the same set of modules; the runtime export surface (`Object.keys`) of every CJS and ESM bundle is identical; the typed export surface of every `lib/index.d.ts`, compared through the TypeScript checker, is identical; UMD global names are unchanged; declarations keep every JSDoc block.
- `pnpm typecheck` passes with 0 errors and leaves every `lib/*.js` byte-identical.
- `nx run-many --target=build`, `pnpm run lint`, `pnpm run knip`, `pnpm test` (17 projects) and `pnpm exec size-limit` pass.
- Cold `nx run-many --target=build --skip-nx-cache` for the 16 library packages: 44 s on main, 33 s here.

**With #5244 merged in** (TypeScript 7, `module: nodenext`): the `typescript` package no longer ships the JS compiler API, so rolldown-plugin-dts switches to its `tsgo` generator, which shells out to the native compiler with `--emitDeclarationOnly --noCheck`. Verified on a scratch merge: build passes, same `.d.ts` count and JSDoc count per package, `declarationMap`s emitted, root typecheck 0 errors, cold build 7.8 s. After #5244 lands, `dts.generator: 'tsgo'` should be pinned in `tsdown.base.mts` so the choice is explicit rather than a side effect of the installed TypeScript version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ should change the heading of the (upcoming) version to include a major version b
 - `build:ts` is now plain `tsc -b`. The old `rimraf ./lib` also deleted the build-info, forcing a full rebuild every time; the build-info is now an Nx `build` output alongside `lib/` so cache restores stay coherent
 - Added `"type": "module"` to `@rjsf/snapshot-tests`, which publishes ESM `.js` files
 - Enabled `verbatimModuleSyntax`, so type-only imports must be written as `import type`. The one import it affected, `React` in `@rjsf/utils`'s `shouldRender.ts`, is now type-only, so emitted output is unchanged
+- Replaced the per-package `tsc -b` emit plus `esbuild` + `rollup` bundling with a single `tsdown` run driven by one shared `tsdown.base.mts`, which derives bundle and global names from each `package.json`. tsdown now emits everything that is published: per-file ESM and declarations in `lib/` and the CJS, ESM and UMD bundles in `dist/`. The build only transpiles; typechecking is the separate root `tsc --build`, and tsc no longer emits JavaScript (the package tsconfigs are `emitDeclarationOnly`). Published files keep the same paths, export surfaces and types. The one difference is inside the UMD wrapper's `<script>`-tag branch, where the guessed browser-global names for externals (e.g. `fastEquals`) are now rolldown's guesses (e.g. `fast_equals`); neither set matched the globals those libraries actually expose, so that branch was never usable
 
 # 6.9.0
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,7 +40,7 @@ cd packages/playground && pnpm start
 pnpm run sanity-check
 ```
 
-Individual package builds output three module formats: `build:cjs`, `build:esm`, `build:umd`.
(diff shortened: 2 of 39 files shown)


jimmycallin:
tsdown has node 22 as earliest version, so i'll wait for next major before migrating to it

jimmycallin:
Continued against `v7` in https://github.com/rjsf-team/react-jsonschema-form/pull/5255, rebased on top of the Node engines change (#5253).

Issue #3812, opened 2023-08-04
Plan for UMD bundle


nickgros:
In RJSF v6 we should either
* Deprecate / remove the UMD bundles, or
* Spend time ensuring the UMD bundles work and follows common practices, and build it with esbuild if possible, and create documentation for core + each theme that supports a UMD bundle

nickgros:
We will revisit in v7

PR #5253, opened 2026-09-09
chore!: support only active Node.js LTS lines in v7

jimmycallin:
## Reasons for making this change

Sets `engines.node` to `^22.18.0 || ^24.11.0 || >=26.0.0` across all packages for v7, matching the support policy [tsdown](https://github.com/rolldown/tsdown/blob/main/package.json) uses:

- **22.18.0** is the first Node release with native TypeScript type stripping enabled by default, which the upcoming tsdown-based build relies on.
- **24.11.0** is where Node 24 became LTS. There is no code difference from 24.10, this is purely "active LTS lines only".
- **Node 20, 23, and 25** are dropped as end-of-life or odd-numbered lines.

Also:
- CI matrix moves from `20.x, 22.x, 24.x` to `22.x, 24.x, 26.x`
- Adds a `7.x upgrade guide` with a Node support section, mirroring the v6 guide
- Adds `CHANGELOG-v7.md` (the v7 changelog named in the branch setup commit, not yet created) with a `# 7.0.0` section and the breaking-change entry. `CHANGELOG.md` is untouched

Pre-existing prettier warnings on `CLAUDE.md` and `ci.yml` are unchanged by this PR (they fail on `v7` too).

## Checklist

- [x] I'm updating documentation
- [ ] I'm adding or updating code
- [x] I'm adding or updating a **theme**? Only `engines` in each theme's `package.json`
- [x] I've updated the `CHANGELOG.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5255, opened 2026-09-09
build!: ESM-only packages built by one tsdown run; tsc only typechecks

jimmycallin:
Supersedes #5247. Targets `v7`, where #5253 has already raised `engines.node` to `^22.18.0 || ^24.11.0 || >=26.0.0`; tsdown needs Node 22.18+ to load its TypeScript config natively, so it could not target `main` while Node 20 was supported. Every number below is measured against `v7` at 1712a49 (the main-to-v7 merge, #5259). That merge set the root `typescript` back to `^6.0.3` although #5244 and the v7 changelog have it at 7.0.2; this PR pins it back to `^7.0.2`, so the lockfile carries one TypeScript and tsdown's declaration generator runs on the native `tsgo` binary. Changelog entries are in `CHANGELOG-v7.md`.

## The problem: one build, four tools, two writers of `lib/`

Every library package today builds like this:

```
tsc -b                      # typecheck + emit lib/*.js and lib/*.d.ts
esbuild src/index.ts …cjs   # bundle dist/index.cjs
esbuild src/index.ts …esm   # bundle dist/<name>.esm.js
rollup dist/<name>.esm.js   # re-bundle the ESM bundle into dist/<name>.umd.js
```

Three problems fall out of that shape.

**1. `tsc` is both the typechecker and the JavaScript compiler.** The root `pnpm run typecheck` is `tsc --build`, so it *also* writes `lib/*.js`. Two different commands own the same output directory. Whichever ran last wins, and `tsc -b` in one package follows project references and re-emits its dependencies' `lib/` too. It works today only because both writers happen to produce the same bytes.

**2. The build is slow because it is four serial processes per package**, and the typechecker is the slowest of them. Parsing and checking the program is most of the wall clock, and it runs on every build even though nothing about the emitted JavaScript depends on the types.

**3. Three of the four outputs are redundant.** `dist/` holds a CJS bundle, an ESM bundle of the same code that already ships per-file in `lib/`, and a UMD bundle that exists for `<script>`-tag loading nobody can actually use (its external globals never matched what the libraries expose). Maintaining them costs a second bundler with 25 platform binaries in the lockfile.

## The fix: ESM only, one transpile step, typechecking stays where it is

Each package build becomes a single command against one shared config:

```
tsdown -c ../../tsdown.base.mts
```

tsdown (rolldown underneath) emits `lib/` and nothing else: per-file ESM mirroring `src/`, so the `./lib/*.js` deep-import exports keep resolving, plus `.d.ts` from tsdown's tsc-based declaration generator. The published types are what tsc inferred before, JSDoc included.

**`dist/` is gone.** The CJS, bundled-ESM and UMD outputs are no longer built or published, and the `./dist` and `./dist/*.cjs` export paths are removed. Each `package.json` points `main` at `lib/index.js`. Export subpaths are plain strings (`"./lib/*.js": "./lib/*.js"`), with a `default` condition only where the `@rjsf/source` test condition sits beside it; TypeScript finds the sibling `.d.ts` of whatever `.js` it resolves, so no `types` condition is needed, and Node's native `require(esm)` (unflagged on every supported Node line) still resolves the packages for CommonJS callers.

**`tsc` no longer emits JavaScript anywhere.** Every package tsconfig stays composite with references, so editors and the root `tsc --build` keep working exactly as now, but they are all `emitDeclarationOnly`. Typechecking is the existing `pnpm run typecheck`; it cannot overwrite the build's JavaScript any more, only rewrite declarations with equivalent content. One writer per output, with no exceptions: `snapshot-tests` builds through the same config.

The config is eight options with no per-package knowledge: where to find sources, where to put them, `unbundle`, `platform: 'neutral'` (the packages run in browsers and SSR; it also keeps the `.js`/`.d.ts` extensions), sourcemaps, `deps.neverBundle` so every dependency (workspace packages included) stays an import, and `dts.generator: 'tsgo'`. Declarations come on by themselves from the `types` field; the generator is named because TypeScript 7 no longer ships the JS compiler API, so rolldown-plugin-dts drives the native `tsgo` binary instead. It infers that on its own when TypeScript 7 is installed, but naming it keeps the choice from moving with the inference rules. tsdown prints `TypeScript 7.0 does not yet have a stable API and is experimental` once per build; that is its warning, not ours. `nx.json` lists the config as a build input so a warm cache cannot serve stale output after a config change.

### Also in this PR

- **shadcn `build:css`** spawned `npx @tailwindcss/cli` once per theme, serially. `npx` resolution was 2.1 s of each 2.25 s. It now calls the `tailwindcss` binary directly, compiles the ten themes in parallel and writes straight to `resources/` instead of hopping through `dist/`: ~25 s → ~1 s, identical CSS.
- The `compileSchemaValidators` docs snippets use `import` instead of `require`, and the v7 upgrade guide gets an "ESM-only packages" section.

### If you contribute

Cold builds get four to six times faster. Same machine, Nx cache skipped, outputs and `tsbuildinfo` deleted before each run, best of two; 15 library packages plus docs and playground.

| | before (`v7`) | after |
|---|---|---|
| Cold build, parallel | 40 s | 6 s |
| Cold build, serial (what CI does) | 63 s | 12 s |
| Root `typecheck` right after a build | 9 s | 5 s |

The "before" column is mostly the three bundler processes per package. The typecheck row compares TypeScript 6.0.3 on `v7` (what #5259 left there) against TypeScript 7 here; the build no longer leaves a `tsbuildinfo` behind, so that 5 s is a full declaration pass rather than an incremental one. Build plus typecheck together drop from 49 s to 11 s.

`rollup` and its 25 platform binaries leave; `tsdown`, `rolldown` and the `yuku` toolchain arrive, each downloading only the current platform's binary. `esbuild` remains as a transitive dependency of `size-limit`.

### If you consume the packages

**Breaking: the packages are ESM-only.**

- `import` from a bundler or from Node: unchanged.
- `require()` from Node: still works on every supported Node version via native `require(esm)`.
- Deep imports through `@rjsf/<package>/lib/<file>.js`: unchanged.
- `require('@rjsf/<package>/dist/…')` or a UMD `<script>` tag from a CDN: no longer available; switch to an ESM CDN or a bundler.

Runtime `dependencies` and `peerDependencies` are untouched, and `size-limit` (minified + gzipped, measured on `lib/index.js`) moves by at most 0.3 kB in either direction across its 27 entries. Tarballs roughly halve because `dist/` is gone, even though `lib/*.js.map` now embed `sourcesContent` so a debugger can step into the original TypeScript without the source present:

| `pnpm pack` | `v7` | this PR |
|---|---|---|
| `@rjsf/core` | 504 kB | 250 kB |
| `@rjsf/utils` | 610 kB | 323 kB |
| `@rjsf/mui` | 123 kB | 67 kB |
| `@rjsf/validator-ajv8` | 69 kB | 37 kB |

## Verification

Against a clean build of the base branch, same machine:

- every package emits the same set of `lib/` modules, and the same number of `.d.ts` files.
- the typed export surface of every `lib/index.d.ts` is identical (every export name, kind and type string, read through the TypeScript API), up to two display differences in how `tsgo` prints a type: union members in a different order in `toFieldPathId`'s signature, and `Field.RootProps` instead of `ChakraField.RootProps` for `getChakra`, the same type under its import alias.
- every `.d.ts` keeps every JSDoc block, with one exception: the file-header comment above the imports in `@rjsf/daisyui`'s `DaisyUIFrameProvider.tsx`, which is attached to no declaration. `tsc` carried it into the `.d.ts`, `tsgo` does not. The exported function keeps its own JSDoc.
- `.d.ts.map` files are emitted for every declaration except the 56 `index.d.ts` that only re-export. TypeScript 7 itself emits maps for those; rolldown-plugin-dts drops them. A declaration map on a pure re-export file only points at the re-export line, so go-to-definition still lands on the re-exported module's map.
- `pnpm pack` on core: 0 `dist/` files, `lib/` and `src/` only.
- a scratch consumer typechecks against the published export maps under both `moduleResolution: bundler` and `nodenext`.
- `require()` and `import` of `@rjsf/core`, `@rjsf/utils`, `@rjsf/mui`, `@rjsf/validator-ajv8`, `@rjsf/validator-ajv8/compileSchemaValidators` and a `lib/*.js` deep import all resolve from a Node 24 script.
- `pnpm typecheck` passes with 0 errors and leaves every `lib/*.js` byte-identical (checked by hashing all of them before and after).
- `lint`, `knip`, `build`, `typecheck`, `cs-check` and `test` (16 projects) pass locally.

The lockfile is `v7`'s plus this branch's dependency changes, with `typescript@7.0.2` as the only TypeScript in it, so tsdown and rolldown-plugin-dts resolve against the same compiler the root `typecheck` uses.

Fixes #3812


knip.jsonc
```diff
@@ -10,25 +10,40 @@
...
       "entry": ["src/**/*.{ts,tsx}"],
-      // Widgets.tsx: its `default` is a compat alias for `generateWidgets` (same
-      // deep-import reasoning as above), which knip reports as a duplicate export.
+      // Widgets.tsx: its `default` is a compat alias for `generateWidgets`
+      // (same deep-import reasoning as above), which knip reports as a
+      // duplicate export.
       "ignore": ["src/widgets/Widgets.tsx"],
     },
+    "packages/react-bootstrap": {
+      "ignore": ["src/index.ts"],
```
heath-freenome, line 34: Nit: this comment narrates pre-PR vs. post-PR build behavior ("...now that the build no longer names the entry in an esbuild script line, which previously masked it"), which is before/after commentary. Per the repo's CLAUDE.md guidance, comments should explain current rationale, not diff-style history — that belongs in the commit message/PR description. A future reader with no knowledge of the old esbuild-based build will find this refers to a masking mechanism that no longer exists in the codebase, making it harder to judge whether the ignore is still needed.

heath-freenome (commented):
Automated review found one style nit (see inline comment). Build was verified locally (install, build-serial, typecheck, lint, knip) plus an external scratch-consumer check of type/require/import resolution — no correctness issues found.

heath-freenome:
This also resolves https://github.com/rjsf-team/react-jsonschema-form/issues/3812

knip.jsonc
jimmycallin: Rewrote the comment to state only the current rationale (the named-plus-default `Form` export is what knip reports). The esbuild history is gone from the comment.

jimmycallin:
Rebased on v7 after the main→v7 merge (#5259). Added `Fixes #3812` and the PR/issue links to the changelog entries.

.github/workflows/size-limit.yml
```diff
@@ -27,15 +27,14 @@ jobs:
...
           pnpm install --frozen-lockfile --filter '!@rjsf/docs' --filter '!@rjsf/playground'
           pnpm install --frozen-lockfile --dir .github/size-limit --lockfile-dir "$GITHUB_WORKSPACE/.github/size-limit"
           echo ${{ github.event.pull_request.number }} > "$RUNNER_TEMP/pr_number"
-          # size-limit measures lib/, not the cjs/esm/umd bundles, and only the
-          # released packages declare build:ts. Safe to let nx cache: lib/ is a
-          # declared build:ts output in nx.json, so a hit restores it to disk.
-          pnpm exec nx run-many --target=build:ts --all
+          # size-limit measures each package's lib/. Safe to let nx cache: lib/
+          # is a declared build output in nx.json, so a hit restores it to disk.
+          pnpm exec nx run-many --target=build --all --exclude=@rjsf/docs,@rjsf/playground
```
heath-freenome, line 37: Is this something that belongs in the main branch too? Or do we just wait for the big merge for 7.x beta?

.github/workflows/size-limit.yml
```diff
@@ -58,7 +57,7 @@ jobs:
             [ -f .github/size-limit/.size-limit.js ] &&
             pnpm install --frozen-lockfile --filter '!@rjsf/docs' --filter '!@rjsf/playground' &&
             pnpm install --frozen-lockfile --dir .github/size-limit --lockfile-dir "$GITHUB_WORKSPACE/.github/size-limit" &&
-            pnpm exec nx run-many --target=build:ts --all; then
+            pnpm exec nx run-many --target=build --all --exclude=@rjsf/docs,@rjsf/playground; then
```
heath-freenome, line 60: Ditto?

heath-freenome (approved):
Something to consider for the main branch. This branch is good to go

.github/workflows/size-limit.yml
jimmycallin: Yes, ported to `main` in #5261 so both branches carry the same workflow and size-limit config. On `main` the `build` target also emits the `dist/` bundles size-limit does not read, so the measurement build is slower there (29 s vs 14 s locally) until v7 merges.

.github/workflows/size-limit.yml
jimmycallin: Same, covered by #5261.

PR #5258, opened 2026-09-09
Fix 4907: remove defaultProps merging from getWidget

heath-freenome:

Fixes #4907

React 19 dropped defaultProps support for function components, so getWidget()'s merging of a widget's defaultProps.options into its rendered options was already broken there and only ever worked for forwardRef/memo-wrapped widgets. Remove the merging entirely along with the now-unused react-is dependency and the MergedWidget property on the Widget type; widgets should apply their own option defaults via destructuring instead.

### Checklist

### DIFF
--- a/CHANGELOG-v7.md
+++ b/CHANGELOG-v7.md
@@ -18,6 +18,10 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 7.0.0
 
+## @rjsf/utils
+
+- **BREAKING CHANGE** Removed `defaultProps`-based options merging from `getWidget()`; a widget passed as a component is now returned as-is instead of being wrapped in a cached `MergedWidget`. Apply option defaults inside the widget itself instead (e.g. via destructuring). The now-unused `MergedWidget` property was also removed from the `Widget` type, and the `react-is` dependency was dropped
+
 ## Dev / docs / playground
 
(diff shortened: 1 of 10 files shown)


jimmycallin (approved):
Checked out the branch and verified locally: `@rjsf/utils` tests pass at 100% coverage, the core `uiSchema` suite passes, and knip is clean after dropping `react-is`. The change is correct and simpler than what it replaces. Returning the widget as-is is just as stable for reconciliation as the old cached `MergedWidget` wrapper, so no remount regression.

Two small follow-ups, neither blocking:

- **Changelog**: `CHANGELOG.md` isn't in the diff. Since this is a breaking change to `@rjsf/utils`, it needs a bullet under that heading (and a major bump on the upcoming version heading if not already there).
- **Stale API reference**: `packages/docs/docs/api-reference/utility-functions.md` (the `getWidget()` section, ~line 658) still says the widget "is wrapped with a `MergedWidget`". The migration guide and custom-widgets docs were updated but this one was missed.

One observation for the record: the guard is now `widget && typeof widget !== 'string'`, which is looser than the old function/forwardRef/memo check. That's a net improvement (`React.lazy` and other exotic component types now work instead of throwing), at the cost that a garbage `ui:widget: {}` surfaces as a React "element type is invalid" error rather than the `getWidget` error. Seems fine given the `{component, options}` misuse is already caught in core with its own console error.

PR #5260, opened 2026-09-10
chore: upgrade @rjsf/mantine to Mantine v9, require React 19 across all themes

heath-freenome:

Mantine v9 requires React ^19.2.0, so this also upgrades react/react-dom (and the pnpm-workspace.yaml override pinning them) to 19.2+ across the whole repo and raises every theme package's `react` peer dependency floor from `>=18` to `>=19`, matching the "React 19 is officially supported on all the themes" policy already declared in the v7 upgrade guide.

- @rjsf/mantine: bump @mantine/core, @mantine/dates, @mantine/hooks to ^9.6.1; peer deps now require mantine >=9
- @rjsf/mantine: make ObjectFieldTemplate's and GridTemplate's root Container fluid so forms fill their available width instead of centering at Mantine's default 960px max-width
- @rjsf/core, @rjsf/chakra-ui, @rjsf/shadcn: convert forwardRef-wrapped components to plain function components using React 19's ref-as-prop
- @rjsf/core: FormProps['onSubmit']'s event is now SubmitEvent<any>
  instead of the deprecated FormEvent<any>
- @rjsf/utils: widen ObjectFieldTemplatePropertyType['content'] to ReactElement<any>, since React 19 defaults a bare ReactElement's props to unknown instead of any
- Fix playground crash ("Cannot destructure property 'fields' of 'undefined'") on load and theme switch: it stored withTheme()'s result directly via useState()/a state setter, and React treats a bare function argument there as a lazy initializer/updater to call immediately rather than a value to store — safe only because withTheme() used to return a forwardRef object, not a plain function
- Fix the handful of other type errors React 19's stricter types surfaced (useRef without a non-null initial value now returns RefObject<T | null>, MutableRefObject is deprecated in favor of RefObject)
- Update snapshots across all theme packages for the resulting Mantine 9 / React 19 rendering differences (new useId format, new ARIA attributes on Mantine's Select/Combobox internals, etc.)

### Checklist


packages/mantine/src/templates/GridTemplate.tsx
```diff
@@ -16,7 +16,7 @@ export default function GridTemplate(props: GridTemplateProps) {
   // Grid with fluid container
   if (fluid) {
     return (
-      <Container p='4' mx={0} w='100%'>
+      <Container fluid p='4' mx={0} w='100%'>
```
jimmycallin, line 19: Mantine 9 renamed `Grid`'s `gutter` prop to `gap` (9.6.1's `Grid.d.ts` has `gap`/`rowGap`/`columnGap`, no `gutter`), but `ui:row` options are still spread straight onto `<Grid {...rest}>`. Every consumer's `gutter` is now silently ignored and leaks to the DOM as an unknown attribute.

The accepted snapshot records the regression instead of catching it: `Grid.test.tsx.snap` now has four `gutter="xs"` attributes on `mantine-Grid-root` (zero before) and `--grid-gap: var(--mantine-spacing-md)` where the base had `--grid-gutter: var(--mantine-spacing-xs)`. `packages/mantine/test/Grid.test.tsx:190` and `playground/src/samples/layoutGrid.tsx:1103` both use `gutter` verbatim.

Either map `gutter` → `gap` here for compatibility, or document the rename in `CHANGELOG-v7.md` and the upgrade guide and update the test/sample. Neither mentions it today.

jimmycallin (commented):
Reviewed the source hunks, snapshots, tests and lockfile. Two regressions, two packaging/doc gaps, and a few cleanups inline. Items that fall outside the diff:

- React 19.3 was just released, consider bumping to it. 
- **Published READMEs contradict the new floors.** `packages/mantine/README.md` still lists `@mantine/core|hooks|dates >= 8` (peer is now `>=9`, React 19.2+), and `packages/mui/README.md` still says "Material UI 7 requires React 18, so you will need to upgrade" (peer is now `react >=19`). Someone following the mui README on React 18 gets a null `formRef.current` from `withTheme()`'s form with nothing pointing at the React version.
- **Dead React 18 assertion.** `packages/core/test/Form.props.test.tsx` ~L388-394 still has `if (renderErrorSuppression.consoleSpy.mock.calls.length > 0)` waiting for React 18's "Function components cannot be given refs" warning. react-dom 19.2.8 has no such string, so the branch never runs. The two sibling React-18 console assertions were removed; this one can go too.
- **`promise-function-async` workarounds.** Three `oxlint-disable-next-line typescript/promise-function-async` comments (core `LayoutGridField.tsx`, chakra `select.tsx`, playground `ErrorBoundary.tsx`) plus the `Boolean(mark.label)` coercion in chakra `slider.tsx` all exist because @types/react 19's `ReactNode` now includes `Promise`. One override in `.oxlintrc.json` for `**/*.tsx` (the file already has per-glob overrides for JS and tests) replaces all four, and the next component that does `return children;` won't need a fifth.

packages/core/src/components/Form.tsx
```diff
@@ -1,4 +1,4 @@
-import type { ElementType, FormEvent, ReactNode, Ref, RefObject } from 'react';
+import type { ElementType, ReactNode, Ref, RefObject, SubmitEvent } from 'react';
```
jimmycallin, line 1: `SubmitEvent` only exists in `@types/react` 19.2.10 and later (checked the published `index.d.ts` for 19.1.0, 19.2.0, 19.2.7, 19.2.8, 19.2.9: absent in all). It is now part of the public `FormProps` type, but the root devDependency is `^19.2.7`, core declares no `@types/react` peer, and neither the changelog nor the upgrade guide states a minimum. A consumer on React 19 with `@types/react` 19.2.9 gets `TS2305: Module 'react' has no exported member 'SubmitEvent'` from `lib/components/Form.d.ts`. CI passes only because the lockfile pins 19.2.18.

Either bump the root floor to `^19.2.10` and document the `@types/react` minimum next to the `SubmitEvent` changelog bullet, or keep `FormEvent<any>` (which `SubmitEvent` extends and which exists in every 19.x).

packages/mantine/package.json
```diff
@@ -63,17 +63,17 @@
...
-    "@mantine/core": ">=8",
-    "@mantine/dates": ">=8",
-    "@mantine/hooks": ">=8",
+    "@mantine/core": ">=9",
+    "@mantine/dates": ">=9",
+    "@mantine/hooks": ">=9",
     "@rjsf/core": "workspace:^",
     "@rjsf/utils": "workspace:^",
-    "react": ">=18"
+    "react": ">=19"
```
jimmycallin: `@mantine/core@9.6.1` and `@mantine/hooks@9.6.1` both declare `peerDependencies.react: ^19.2.0` in the lockfile (Mantine 9 hooks use `useEffectEvent`, which 19.0/19.1 don't have), and `CHANGELOG-v7.md` says this package "requires React 19.2+". The peer here should be `>=19.2` to match, otherwise a project on 19.1.x gets no complaint from rjsf and Mantine's Select/DateInput throw on mount.

packages/utils/src/types.ts
```diff
@@ -873,7 +873,7 @@ export type ArrayFieldTemplateProps<
 /** The properties of each element in the ObjectFieldTemplateProps.properties array */
 export interface ObjectFieldTemplatePropertyType {
   /** The html for the property's content */
-  content: ReactElement;
+  content: ReactElement<any>;
```
jimmycallin: This adds an `any` to a public type solely so antd can read `element.content.props.schema` / `.uiSchema`. A typed version compiles against antd's exact usage: make the interface generic (`ObjectFieldTemplatePropertyType<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>`) with `content: ReactElement<Pick<FieldProps<T, S, F>, 'schema' | 'uiSchema'>>`, and thread `<T, S, F>` through `properties` on `ObjectFieldTemplateProps` (~L896). A non-generic `Pick<FieldProps, ...>` does not typecheck because of the `UiSchema<T, S, F>` mismatch.

Simpler still: `ObjectField` has `schema` and `uiSchema` in hand where it builds `{ content, name, ... }`, so it could expose them as fields and no theme would need to read element props at all.

packages/playground/src/components/Playground.tsx
```diff
@@ -60,12 +60,12 @@ export default function Playground({ themes, validators }: PlaygroundProps) {
   const playGroundFormRef = useRef<any>(null);
 
   // oxlint-disable-next-line react/hook-use-state
-  const [FormComponent, setFormComponent] = useState<ComponentType<FormProps>>(withTheme({}));
+  const [FormComponent, setFormComponent] = useState<ComponentType<FormProps>>(() => withTheme({}));
```
jimmycallin: `FormComponent` is a pure function of `theme` and `themes`: every `setFormComponent` (L68) is preceded by `setTheme(newTheme)`, every caller passes `themes[name]`, and `themes` is a module-scope constant in `app.tsx`. Deriving it removes the duplicated state, the `react/hook-use-state` disable, both lazy-initializer wrappers, the throwaway `withTheme({})` mount that the effect immediately replaces, and makes the crash this PR fixes structurally impossible:

```ts
const FormComponent = useMemo(() => withTheme(themes[theme].theme), [themes, theme]);
```

(`stylesheet` has to stay state since `Editors.tsx:127` sets it from subthemes.)

packages/chakra-ui/src/components/ui/select.tsx
```diff
@@ -113,21 +104,27 @@ export const SelectValueText = forwardRef<HTMLSpanElement, SelectValueTextProps>
...
+ * to apply to the whole arrow function, which is why the extra wrapping parens below are required
+ * — removing them would make the cast apply to the returned JSX instead and break the type.
+ *
+ * @param props - The properties for the select root component.
+ * @param [props.positioning] - The positioning properties for the select component.
+ * @param [props.children] - The content to display inside the select root.
+ * @returns The rendered select root component.
  */
-export const SelectRoot = forwardRef<HTMLDivElement, ChakraSelect.RootProps>((props, ref) => (
+export const SelectRoot = (({ ref, ...props }: ChakraSelect.RootProps & { ref?: Ref<HTMLDivElement> }) => (
```
jimmycallin: The cast and the five-line comment defending the parentheses (L116-120) go away if this is a generic function declaration. Chakra 3.35's `SelectRootComponent` is `<T extends CollectionItem>(props: SelectRootProps<T> & RefAttributes<HTMLDivElement>) => JSX.Element`, and this typechecks with the repo's strict flags, satisfies `const check: ChakraSelect.RootComponent = SelectRoot;`, and infers `d.value` as `string[]` at the sole call site in `SelectWidget.tsx:112`:

```tsx
export function SelectRoot<T extends CollectionItem>({ ref, ...props }: ChakraSelect.RootProps<T> & RefAttributes<HTMLDivElement>) {
  // same body
}
```

`{ ref?: Ref<HTMLDivElement> }` is also just `RefAttributes<HTMLDivElement>`. The cast hides future prop-type drift from the compiler.

packages/core/src/withTheme.tsx
```diff
@@ -13,37 +12,43 @@ export type ThemeProps<T = any, S extends StrictRJSFSchema = RJSFSchema, F exten
...
-      const widgets = { ...themeProps?.widgets, ...propWidgets };
-      const templates = {
-        ...themeProps?.templates,
-        ...propTemplates,
-        ButtonTemplates: {
-          ...themeProps?.templates?.ButtonTemplates,
-          ...propTemplates?.ButtonTemplates,
-        },
-      };
+  return function ThemedForm({
```
jimmycallin, line 25: Not confirmed against a React 19 install, so take as a suggestion: returning `memo(ThemedForm)` would restore the non-callable object contract `forwardRef` had (`{ $$typeof: REACT_MEMO_TYPE, ... }`), so `useState(withTheme(x))` keeps working and the docblock warning, upgrade-guide section, changelog bullet and both playground patches become unnecessary. The repo already does `memo(X) as typeof X` in `IconButton`, `ArrayField`, `ObjectField`, `SchemaField`, and React 19's `memo` forwards `ref` as a prop.

The one wrinkle is `experimental_componentUpdateStrategy: 'always'`, which a plain `memo` would defeat when a parent re-renders with identical references, so it would need a comparator along the lines of `(a, b) => a.experimental_componentUpdateStrategy !== 'always' && shallowEqual(a, b)`. Still smaller than a breaking change for every downstream `useState(withTheme(...))` caller.

packages/mantine/src/templates/GridTemplate.tsx
heath-freenome: Fixed in 3aff7d4e7 — `GridTemplate` now destructures `gutter` and maps it to Mantine 9's `gap` prop (`gap = gutter` as the default, so an explicit `gap` still wins if both are provided). Verified against the snapshot: `--grid-gap` now resolves to the intended value instead of falling back to the `md` default, and the stray `gutter="xs"` DOM attribute is gone.

Documented in `CHANGELOG-v7.md` and the upgrade guide — existing `ui:row`/`ui:col` options using `gutter` keep working, but should be renamed to `gap` going forward.

packages/mantine/package.json
heath-freenome: Fixed in 3aff7d4e7 — tightened to `>=19.2`, matching what `@mantine/core`/`@mantine/hooks` 9.6.1 themselves declare as their own peer requirement.

packages/core/src/components/Form.tsx
heath-freenome: Went with bumping the floor rather than reverting to `FormEvent`. One correction on the specifics: I pulled and grepped the actual published `@types/react` tarballs for every 19.2.x patch — `SubmitEvent` isn't in 19.2.10 either, it's not in *any* of 19.2.0 through 19.2.17. It only shows up starting at 19.2.18, so the problem was a bit worse than described.

Fixed in 3aff7d4e7: bumped the root `@types/react` floor to `^19.2.18` and documented the minimum next to the `SubmitEvent` entry in the upgrade guide, so a consumer on an older 19.2.x patch gets a clear pointer instead of a confusing `TS2305`.

packages/utils/src/types.ts
heath-freenome: Took the typed-generic route in 3aff7d4e7: `ObjectFieldTemplatePropertyType` is now `<T, S, F>` (defaulted like every other RJSF generic type), with `content: ReactElement<Pick<FieldProps<T, S, F>, 'schema' | 'uiSchema'>>` instead of `any`. Confirmed it lines up structurally with what `ObjectField.tsx` actually builds (`ObjectFieldPropertyProps extends Omit<FieldProps<T,S,F>, 'name'>`), and threaded `<T, S, F>` through all five consumers — `@rjsf/antd`, `@rjsf/core`'s own `ObjectFieldTemplate`, `@rjsf/mantine`'s, the playground's `customObject` sample, and `ObjectFieldTemplateProps['properties']` itself.

Didn't take the 'simpler still' route of exposing `schema`/`uiSchema` as direct fields instead of reading through `content.props` — that's a real API shape change beyond a type fix, wanted to scope it separately rather than bundle it into this PR.

packages/playground/src/components/Playground.tsx
heath-freenome: Applied basically as suggested in 3aff7d4e7: `FormComponent` is now `useMemo(() => withTheme(themes[theme].theme), [themes, theme])`. Removed the duplicated `useState`, the `react/hook-use-state` lint suppression, the throwaway `withTheme({})` initial mount, and both lazy-initializer call sites — and yes, `stylesheet` stayed as real state since `Editors.tsx` still sets it independently for subthemes.

packages/chakra-ui/src/components/ui/select.tsx
heath-freenome: Applied in 3aff7d4e7 — `SelectRoot` is now `function SelectRoot<T extends CollectionItem>({ ref, ...props }: ChakraSelect.RootProps<T> & { ref?: Ref<HTMLDivElement> })`, no more cast to `ChakraSelect.RootComponent` or the comment explaining the wrapping parens. Confirmed it still typechecks cleanly under the repo's strict flags and the full chakra-ui test suite still passes.

packages/core/src/withTheme.tsx
heath-freenome: Didn't take this one, though the core mechanism checks out: `React.memo()` does return a plain object (`{ $$typeof: REACT_MEMO_TYPE, ... }`), not a function, so it would sidestep the `useState(withTheme(x))` footgun the same way `forwardRef` used to. But the `experimental_componentUpdateStrategy: 'always'` concern you raised is real — `memo`'s shallow-equal bailout would run *above* `Form`'s own `shouldComponentUpdate`, so it'd need a custom comparator to not silently defeat `'always'`, and I didn't want to ship that untested just to restore old implicit behavior.

Holding off more deliberately, though: we're about to split `Form` into two stateless functional components along the uncontrolled/controlled boundary, which is going to change how `withTheme()` and ref-forwarding work here more fundamentally. Didn't want to patch this twice, so the current explicit approach (docblock warning + upgrade guide entry + fixed call sites) stays for now, and the ref/memo question gets revisited as part of that split.

heath-freenome:
Replied inline to each of the 7 line comments. For the items called out in the review body but not tied to a specific line:

- **React 19.3** — bumped in 3aff7d4e7. Worth flagging: doing so without a version-pinning override forked the dependency graph into two react copies (19.2.8 and 19.3.0, with `react-dom@19.3.0` paired against `react@19.2.8` in part of the tree — confirmed via `pnpm why react` and a real test failure inside `@rjsf/shadcn`). Re-added the `pnpm-workspace.yaml` override, now pinned to 19.3.0, with a comment explaining why.
- **Stale READMEs** — fixed both (`@rjsf/mantine`'s prerequisites, `@rjsf/mui`'s React-18 note).
- **Dead React 18 assertion in `Form.props.test.tsx`** — removed. Confirmed via a standalone repro that React 19 never emits the "Function components cannot be given refs" warning for this case, so the branch could never run.
- **`.oxlintrc.json` override to replace the three `promise-function-async` disables** — tried it, and it worked (verified the rule stays live elsewhere by temporarily reintroducing the same pattern in an unrelated file and confirming it still got flagged), but reverted it in favor of keeping the three individual `oxlint-disable-next-line` comments. Prefer the false-positive suppression staying visible at each call site rather than in a config file three files might quietly rely on later.

Everything above plus the 7 inline fixes is in 3aff7d4e7, pushed.

jimmycallin:
Re the react/react-dom override added in 3aff7d4: the fork isn't caused by anything in the dependency graph, so I think we can fix it at the source and skip the override.

**Where 19.2.8 comes from**

Only the root `package.json` and the playground declare `react` directly. Every other package has it as a peer with a loose range (`>=18`). Those peer resolutions are recorded in `pnpm-lock.yaml`. When the direct ranges were bumped to `^19.3.0` and `pnpm install` ran, pnpm only re-resolved the entries whose specifier changed — the ~578 peer-resolved entries pointing at `react@19.2.8` were left as-is because 19.2.8 still satisfies `>=18`. `react-dom` moved to 19.3.0 (requested directly) and got paired against those stale peers. It's lockfile inertia, not a real version conflict.

**Repro** (on this branch before the override): bump the two ranges, `pnpm install` → lockfile has `react@19.2.8` + `react@19.3.0`, with 578× `react-dom@19.3.0(react@19.2.8)`.

**Fix**

```
pnpm dedupe
```

That re-resolves the whole graph. After running it the lockfile has a single `react@19.3.0` and all 744 `react-dom` entries are paired with it. No override needed.

**Suggestion**

- Drop `react`/`react-dom` from `overrides` — it masks a stale-lockfile problem rather than solving a conflict, and it also silently forces any third-party package that has a legitimately narrower react peer range.
- Run `pnpm dedupe` after the bump and commit the lockfile.
- Optionally add `pnpm dedupe --check` to CI so a future bump can't reintroduce a fork. Note it currently exits non-zero even after deduping react (other packages in the lockfile are duplicated too), so it'd need a one-time full dedupe first.

heath-freenome:
Good catch, and thanks for the actual repro — you're right, it was lockfile inertia, not a real conflict. Fixed in 070311f14:

- Dropped `react`/`react-dom` from `pnpm-workspace.yaml`'s `overrides`.
- Ran `pnpm dedupe` and committed the result. Lockfile now has a single `react@19.3.0` with every `react-dom` entry paired against it; confirmed `node_modules/react`'s symlink resolves to 19.3.0 only.
- `pnpm dedupe` also caught one other stale entry (`postcss@8.5.26`) along the way — after that, `pnpm dedupe --check` passes cleanly.
- Full sanity check (typecheck, lint, cs-check, knip, build, test) is green.

Didn't add `pnpm dedupe --check` to CI — leaving that one for a separate decision rather than bundling a CI workflow change into this PR.

heath-freenome:
Following up on the previous comment — added both `pnpm dedupe --check` guards discussed there:

**Pre-commit** (`5c385fe88`): a root-level `lint-staged` entry runs `pnpm dedupe --check` whenever `pnpm-lock.yaml` is staged. Needed its own wiring since the existing `pre-commit:husky` script only runs `nx run-many --target=precommit`, which covers the `packages/*` projects — the repo root isn't an nx project, so it gets a sibling `pre-commit:lockfile` script instead. Also had to wrap it as `sh -c 'pnpm dedupe --check'`, since `dedupe --check` errors on any extra positional argument and lint-staged always appends the staged filename to the command. Verified end-to-end with a real commit: confirmed it no-ops when the lockfile isn't staged, and (by deliberately breaking the lockfile with a trailing newline first) confirmed it catches and blocks a bad one, with lint-staged's stash-based rollback restoring the working tree cleanly.

**CI** (`d9090bec6`): added right after the frozen-lockfile install, gated to run once (`node-version == '24.x'`) the same way `knip` already is — dedupe-ability is a property of the lockfile, not the Node version. `--frozen-lockfile` alone wouldn't have caught the original react/react-dom duplication, since the stale peer entries were valid, just redundant.

Both pass cleanly against the current lockfile.

jimmycallin (approved):

PR #5262, opened 2026-09-10
Remove deprecated v7 APIs from @rjsf/core and @rjsf/utils

heath-freenome:

Removes APIs that were marked @deprecated for removal in a future major release, per the v7 migration guide:

- Form.getUsedFormData()/getFieldNames() instance methods, and the getUsedFormData/getFieldNames exports from @rjsf/utils
- Form.omitExtraData() instance method (use SchemaUtils.omitExtraData instead); internal callers now go through schemaUtils directly via a new private Form#omitFormExtraData helper
- FormProps.removeEmptyOptionalObjects, which had already become a no-op
- removeOptionalEmptyObjects() from @rjsf/utils, superseded by the pruning built into omitExtraData()
- toPathSchema() (both the @rjsf/utils export and the SchemaUtils/SchemaUtilsType method) and the PathSchema/FieldPath types
- UiSchema['ui:rootFieldId'], superseded by the Form.idPrefix prop
- boolean (true/false) values for the liveValidate/liveOmit props; only 'onChange' | 'onBlur' | undefined remain

Updates all call sites across core, utils, and the shared validator test suites, the API reference docs, and the playground's live-settings UI (which now models "off" with its own sentinel instead of leaking a boolean into Form), and documents the removals in the v7 upgrade guide and CHANGELOG-v7.md.

Verified: full lint/typecheck/knip/build/test pass across every package (100% coverage maintained in utils and the validator packages), and the playground's Live validation/Live omit controls were exercised in a real browser to confirm the new prop values still drive live validation correctly.

### Checklist


jimmycallin:
Found one regression in the playground:

**P2 — Shared URLs without `liveSettings` now crash.** The new reads at [Playground.tsx:254–255](https://github.com/rjsf-team/react-jsonschema-form/blob/346f84390db09dc8b84f358525384c612bb3c22d/packages/playground/src/components/Playground.tsx#L254-L255) assume `liveSettings` exists, but `load()` sets it to `undefined` when the decoded URL payload omits that property. Previously, spreading `undefined` into the form props was harmless; the new property access throws `Cannot read properties of undefined (reading 'liveValidate')`.

Reproduction: load a playground URL whose base64-encoded hash contains:

```json
{"schema":{"type":"string"},"formData":"hello"}
```

Use optional chaining for both reads (`liveSettings?.liveValidate` and `liveSettings?.liveOmit`), or default the loaded settings to `{}`.

Confirmed with a focused rendering test that passes against the PR base (`2d6524403`) and fails against the PR head (`346f84390`). The 1,187 focused core and schema utility tests also passed.

heath-freenome:
Thanks for catching this, @jimmycallin! Fixed in 24ee8622c.

The root cause: `load()` set `liveSettings` state directly to the decoded payload's `liveSettings` value, which is `undefined` when a shared URL predates that field. That was harmless before — spreading `undefined` into JSX props is a no-op — but the new `liveSettings.liveValidate`/`liveOmit` reads (added to map the removed boolean prop values onto the string ones) don't tolerate that.

Extracted a `normalizeLiveSettings()` helper that defaults a missing `liveSettings` to `{}` and folds in the existing legacy-boolean conversion (`true` → `'onChange'`, `false` → `'off'`), so both the crash and the "old shared URL with `true`/`false`" case are covered from one place. Added `packages/playground/test/components/liveSettings.test.ts` covering both directly.

Verified against a live dev server with your exact repro payload (`{"schema":{"type":"string"},"formData":"hello"}`) — loads cleanly now, and the Live omit/Live validation radios correctly default to "off".

jimmycallin (approved):

Issue #4328, opened 2024-10-04
The UISchema "ui:description" no longer supports the <small> tag


felipecarrillo100:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

React-Bootstrap

### Version

    "@rjsf/core": "^5.21.1",
    "@rjsf/react-bootstrap": "^6.0.0-alpha.0",

### Current Behavior
The description markdown no longer supports the tag \<small\> , it used to work in previous versions but currently it doesn't work.  You can witness this in your own playground.
{
  "lastName": {
    "ui:autocomplete": "given-name",
    "ui:enableMarkdownInDescription": true,
    "ui:description": "Make things **bold** or *italic*. Embed snippets of `code`. \<small\>And this is a small texts.\</small\> "
  },
}  
Renders to:
Make things **bold** or *italic*. Embed snippets of `code`.  \<small\>And this is a small texts.\</small\>

### Expected Behavior

The correct rendering should be:
Make things **bold** or *italic*. Embed snippets of `code`. <small>And this is a small texts.</small>

Where "And this is a small texts." in smaller size and tag <small> is not printed to screen

### Steps To Reproduce

You can reproduce this in your own ONLINE playground.  Just go to urL
https://rjsf-team.github.io/react-jsonschema-form/

Take the sample "Simple"  and look at the "A registration form", Input "Last Name". You will see the "small" tag is printed as text: \<small\>And this is a small texts.\</small\>.  The \<small\> tag is printed as text instead of being used as format.

### Environment

```markdown
- OS:
- Node:
- npm:
```

heath-freenome:
@felipecarrillo100 According to this [article](https://meta.stackexchange.com/questions/53800/markdown-extension-for-really-small-tiny-text), it seems that the `<small>` tag was rejected recently by IIRC. If you follow the advice in the article you could use `<sup>` or a combination of `<sup><sub>` tags

heath-freenome:
[...]

PR #5263, opened 2026-09-10
feat!: make markdown opt-in via @rjsf/core/markdown to shrink the Form bundle

jimmycallin:

A form that never renders markdown still includes the markdown parser today. `RichDescription`, `RichHelp`, `ObjectField`, `UnsupportedField` and `FileWidget` import `markdown-to-jsx` directly, so setting the markdown flags to `false` does not remove that code from the bundle. The parser is roughly a third of `@rjsf/core`.

This PR was split: the explicit theme composition (`createForm`, `@rjsf/core/testing`) is now #5266 and the deep-import removal is #5265. Each targets `v7` independently; whichever lands last will need a small rebase.

### Solution: a `MarkdownTemplate` slot, registered only when needed

A new `templates.MarkdownTemplate` renders markdown-capable text. The core default returns the text as-is and imports no parser. Applications that need markdown register the renderer exported by `@rjsf/core/markdown`.

The renderer is deliberately not re-exported from the package root. Bundlers and Node's ESM loader resolve every import in the graph before tree-shaking, so a root re-export would force every consumer to have `markdown-to-jsx` installed even when nothing uses it. Keeping it on its own subpath is what lets `markdown-to-jsx` become an optional peer dependency instead of a dependency.

The existing `ui:enableMarkdownInDescription` and `ui:enableMarkdownInHelp` flags keep their separate meanings and still default to `false`. They decide **where** the registered renderer is used; registration decides **which** renderer. Global options, per-field overrides and `ui:MarkdownTemplate` all work through the usual `getTemplate()` lookup. The included renderer disables raw HTML parsing.

### Breaking change for markdown users: one install, one import, one registration

**Before:** enabling a flag was enough because core included the parser.

```tsx
import Form from '@rjsf/core';

<Form
  schema={schema}
  validator={validator}
  uiSchema={{ 'ui:enableMarkdownInDescription': true, 'ui:enableMarkdownInHelp': false }}
/>;
```

**After:** install `markdown-to-jsx`, keep the same schema, strings and flags, and register the renderer.

```tsx
import Form from '@rjsf/core';
import MarkdownTemplate from '@rjsf/core/markdown';

<Form
  schema={schema}
  validator={validator}
  templates={{ MarkdownTemplate }}
  uiSchema={{ 'ui:enableMarkdownInDescription': true, 'ui:enableMarkdownInHelp': false }}
/>;
```

**Without that registration, existing markdown is displayed as plain text even if its flag is `true`. There is no automatic parser loading or migration warning.** Registering the renderer alone does not enable markdown; the flags must also be enabled. Themed forms keep importing `Form` from the theme and take the renderer from `@rjsf/core/markdown`.

### Other breaking changes

| Change | Who is affected / migration |
| --- | --- |
| `markdown-to-jsx` is an optional peer dependency instead of a dependency | Applications that register the renderer add `markdown-to-jsx` to their own dependencies. Applications that don't no longer install it. |
| `TemplatesType` gains a required `MarkdownTemplate` | Authors constructing complete template maps must add the slot. Spreading core `Templates` supplies the plain-text default. |
| Built-in translated strings are plain text | The English `InvalidObjectField`, `UnsupportedField*` and `FilesInfo` strings no longer contain markdown formatting. A custom `translateString` that returns markdown is rendered through `RichDescription`, so it needs both the registered renderer and `ui:enableMarkdownInDescription` (per field or via `ui:globalOptions`). |
| Raw HTML parsing is disabled for file information | Code relying on HTML inside `FileWidget`'s file list must provide its own renderer. |

### Effect on bundle size

Size-limit, minified + brotli, peers excluded:

| Row | v7 | This PR |
| --- | --- | --- |
| `@rjsf/core: Form` canary | ~44 kB | ~21 kB |
| `@rjsf/core (without dependencies)` | ~22 kB | ~21.4 kB |
| `@rjsf/core` root | ~80 kB | ~56 kB |

The root row is still large because on `v7` the root index exports `getTestRegistry`, which pulls the AJV chain in; #5266 moves it to `@rjsf/core/testing` and brings the root down to the `Form` size. The `Form` canary is the real signal: importing only `Form` includes zero bytes from `markdown-to-jsx`.

### Validation

Lint, knip, build, root typecheck, all test targets, source formatting and size-limit pass on this branch, rebased on `v7` after #5260 and #5264. Tests cover independent description/help flags, local and global overrides, `ui:MarkdownTemplate` precedence, missing renderer fallback to plain text, and file-info markdown honouring the per-field flag.

### Checklist

  - [x] I've added/updated tests and verified the existing snapshots
  - [x] The playground registers the renderer and retains sample markdown flags


packages/core/src/markdown.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const TEST_IDS = getTestIds();
+
+/** Renders markdown with raw HTML parsing disabled. Only exported from `@rjsf/core/markdown`, never the root, so that
+ * `markdown-to-jsx` can be an optional peer dependency: a consumer that never imports this subpath never has to
+ * install or resolve it.
+ */
+export default function MarkdownTemplate({ children }: MarkdownTemplateProps) {
+  return (
+    <Markdown options={{ disableParsingRawHTML: true }} data-testid={TEST_IDS.markdown}>
```
heath-freenome, line 13: There is an issue https://github.com/rjsf-team/react-jsonschema-form/issues/4328#issuecomment-2408003248 which might cause the `options={{ disableParsingRawHTML: true }}` to be removed. Maybe we should add additional props on the template to support this?

heath-freenome: OR, better yet, have a second implementation, markdownUnsafeHTML that the user can choose to use instead

jimmycallin: Agreed that #4328 is a real regression and worth closing out — but I think the template slot in this PR already covers it, without a new export.

A consumer who wants raw HTML registers three lines:

```tsx
import { Markdown } from 'markdown-to-jsx/react';

templates.MarkdownTemplate = ({ children }) => <Markdown>{children}</Markdown>;
```

They already have `markdown-to-jsx` installed — it's a peer dependency the moment they use markdown at all. So #4328 is closed by a docs example rather than by code.

On adding `options` props to the template: that would put `markdown-to-jsx`'s option type back into `@rjsf/utils`' public surface, which is the coupling the subpath split exists to remove. It's also meaningless for anyone who registers a non-`markdown-to-jsx` renderer (remark, markdown-it), which the slot deliberately allows. The slot *is* the option-passing mechanism.

On a second `markdownUnsafeHTML` implementation: more defensible, since a named export makes the security tradeoff explicit and greppable instead of every app quietly re-deriving it. But it's a second subpath, second export, second docs section and second test suite for a component whose whole body is `<Markdown>{children}</Markdown>` — and the name doesn't buy the safety it implies, since nothing stops someone registering an unsafe template anyway. I'd lean toward documenting it instead.

heath-freenome: I'm fine with documenting it so users can customize that behavior appropriately... One small change might be to support passing an optional UiSchema and registry, to give custom template authors a means to support additional behaviors

jimmycallin:
Rebased onto current `v7` (`72da33978`). Seven conflicts, all from PRs that landed since: the size-limit config rewrite (#5272), the `experimental_` prefix drop (#5267), and the PrimeReact removal (#5273).

Notable resolutions:

- **`packages/core/package.json`** — #5272 made `@rjsf/validator-ajv8` an optional peer for `@rjsf/core/testing`. This PR does the same for `markdown-to-jsx`. Both are now optional peers, and `dependencies` is empty as a result, so `@rjsf/core` ships with no runtime dependencies at all.
- **`.github/size-limit/.size-limit.js`** — took v7's rewritten mechanism wholesale and dropped this branch's `extraIgnore`, which the new optional-peer handling supersedes. Because that config now measures every export subpath automatically, `@rjsf/core/markdown` gets its own row for free.
- **Budgets re-measured rather than guessed:** `@rjsf/core` **21.66 kB** (was a 47 kB budget) and `Form` **21.39 kB** (was 44.30 kB). Budgets set to 24 kB each. The `own` budget is gone along with the dependencies it excluded.
- Two tests importing `getTestRegistry` from the package root now import from `@rjsf/core/testing`.

Verified: root `typecheck`, lint, knip, `cs-check`, build, full test suite, and size-limit all clean.

### Interaction with #5266 and #5278

I test-merged both rather than guessing.

**#5278 (`ErrorSchema` types)** — only `CHANGELOG-v7.md` conflicts. `packages/utils/src/types.ts` auto-merges cleanly: #5278 rewrites the `ErrorSchema`/`FormValidation` block, this PR touches `GlobalUISchemaOptions` and `TemplatesType`. No semantic overlap.

**#5266 (theme composition)** — `CHANGELOG-v7.md` and `packages/core/test/SchemaField.test.tsx` conflict; the latter is only that both edit the same import line (#5266 swaps `getDefaultRegistry` for `generateTheme`, this PR adds the markdown import on the next line).

**One thing that merges cleanly but is wrong**, worth catching whoever rebases last: #5266 adds an explicit `export { ... }` list of every core template, and this PR adds `MarkdownTemplate` to `generateTemplates()`. Git merges both without complaint, and the result has `MarkdownTemplate` registered in `generateTemplates()` but missing from that export list — the only core template not individually importable from `@rjsf/core`. It needs adding to the list by hand.

Worth noting that after #5266 the default `MarkdownTemplate` becomes a root export of `@rjsf/core`. That is harmless — it is the plain-text passthrough and imports no parser — but it is a new public export neither PR's changelog currently mentions. The `markdown-to-jsx` renderer stays on `@rjsf/core/markdown` only, which is what keeps it an optional peer.

packages/utils/src/types.ts
```diff
@@ -354,6 +354,12 @@ export type RegistryWidgetsType<
   F extends FormContextType = any,
 > = Record<string, Widget<T, S, F>>;
 
+/** The properties that are passed to a `MarkdownTemplate` implementation */
+export interface MarkdownTemplateProps {
+  /** The markdown text to render */
+  children: string;
```
heath-freenome, line 364: @jimmycallin Thoughts about adding registry and UiSchema for people who want to customize and get access to the globalOptions and/or the field specific uiSchema?

jimmycallin: Good call, done in 039d84667. `MarkdownTemplateProps<T, S, F>` now picks `uiSchema` and `registry` from `RJSFBaseProps`, and `RichDescription`/`RichHelp` pass them through. I left `schema` out on purpose: the renderer has no use for it and neither call site has it in scope, so adding it would ripple a new prop through `RichDescription`, `RichHelp`, `FileWidget` and `ObjectField` for nothing. Also rebased on `v7` now that #5266 landed.

heath-freenome:
## Code review findings

(Existing review threads on `markdown.tsx`'s `disableParsingRawHTML` and `MarkdownTemplateProps` shape already cover that ground, so I focused on other areas.)

**1. `packages/core/src/components/templates/UnsupportedField.tsx:31`** — Per-field `ui:enableMarkdownInDescription` silently ignored

`UnsupportedField` renders its message through `RichDescription` without forwarding the field's `uiSchema`, so per-field `ui:enableMarkdownInDescription` can never reach it — only the global option works.

A form that sets `ui:enableMarkdownInDescription: true` on a specific field (not globally) plus registers a MarkdownTemplate — exactly the workflow the PR's own migration guide describes ("needs both the registered renderer and ui:enableMarkdownInDescription (per field or via ui:globalOptions)") — won't get markdown rendering for that field's unsupported-field message. Because `UnsupportedField.tsx` calls `<RichDescription description={...} registry={registry} />` with no `uiSchema` prop, and none of its call sites (`FallbackField.tsx:115`, `ArrayField.tsx:1090-1095`) pass `uiSchema` either, `RichDescription` falls back to its default `uiSchema = {}` and only `registry.globalUiOptions.enableMarkdownInDescription` is consulted. This contradicts the documented behavior — the sibling paths added in this same PR (`ObjectField.tsx`'s error path and `FileWidget.tsx`'s `FilesInfo`) correctly forward `uiSchema` and don't have this gap.

**2. `packages/core/src/components/RichDescription.tsx:25`** — `RichDescription`/`RichHelp` duplication (simplification)

`RichDescription` and `RichHelp` are near-identical duplicates (same guard/lookup logic, differing only in prop name and ui-option key) that this PR touches without consolidating. This is likely *why* #1 exists — any change to the markdown-gating logic (e.g. adding registry/uiSchema to `MarkdownTemplateProps`, already under discussion elsewhere in this PR) has to be made twice, in `RichDescription.tsx` and the structurally identical `packages/core/src/components/RichHelp.tsx`, risking the two drifting out of sync the way the `UnsupportedField` gap already shows one path lagging the others.

---

packages/core/src/markdown.tsx
jimmycallin: Documented in 6068e2f56 — `custom-templates.md` now shows an `UnsafeHtmlMarkdownTemplate` that registers `<Markdown>` without `disableParsingRawHTML`, with a note that it's only safe when you control the markdown source. That covers #4328 without a second export.

`uiSchema` and `registry` on `MarkdownTemplateProps` landed earlier in 039d84667.

jimmycallin:
Both findings fixed in 6068e2f56.

**1. `UnsupportedField` ignoring per-field flags** — confirmed, and it was two bugs in a line: the template didn't forward `uiSchema` to `RichDescription`, and neither call site (`FallbackField`, `ArrayField`) passed one in. Both now do, so `ui:enableMarkdownInDescription` on the field reaches the message the same way it does for `ObjectField`'s error path and `FileWidget`'s file info.

Added a test covering both call sites (invalid type, and an array without `items`) with a `translateString` that returns markdown. It fails on the previous commit and passes now.

**2. `RichDescription`/`RichHelp` duplication** — agreed that the duplication is why #1 drifted, so this is fixed at the root: the guard, option lookup and template resolution now live once in an internal `RichText` component, parameterised by which `ui:options` flag enables it. `RichDescription` and `RichHelp` stay as public exports and are five-line wrappers, so nothing changes for themes that import them.

Also from the `disableParsingRawHTML` thread: the raw-HTML renderer is now documented in `custom-templates.md`.

Lint, knip, root typecheck, `cs-check` and the full test suite pass.

heath-freenome:
## Follow-up review

Both issues from my earlier comment are confirmed fixed in the current HEAD (`6068e2f56`):
- `UnsupportedField.tsx` now forwards `uiSchema` to `RichDescription` (along with `FallbackField.tsx` and `ArrayField.tsx`), with a regression test in `packages/core/test/markdownOptions.test.tsx`.
- `RichDescription`/`RichHelp` duplication resolved — shared logic extracted into a new internal `RichText.tsx`, with both now thin wrappers over it.

The other reviewers' threads on `disableParsingRawHTML` and `MarkdownTemplateProps` also look closed out (docs example added, `registry`/`uiSchema` added to the type).

Two new findings from the current HEAD:

**1. `packages/core/src/components/templates/index.ts:27`** — `MarkdownTemplate` not exported

`MarkdownTemplate` is imported and used inside `generateTemplates()` but omitted from the named `export { ... }` list, so it's the only core template not individually importable from `@rjsf/core`. `import { MarkdownTemplate } from '@rjsf/core'` — the same pattern that works for `ArrayFieldTemplate`, `ObjectFieldTemplate`, etc. — silently returns `undefined`. This looks like it was already noticed: there's a PR comment from the author flagging this exact gap ("this needs adding to the list by hand") while merge-testing against #5266, but it's still missing from HEAD.

**2. `packages/core/src/components/RichDescription.tsx:43`** — `TEST_IDS` silently dropped from public API with no changelog note

`RichDescription.TEST_IDS` and `RichHelp.TEST_IDS`, previously present on these publicly-exported components, were removed when the markdown-gating logic moved to `RichText`/`MarkdownTemplate`, with no mention in `CHANGELOG-v7.md`'s breaking-changes entries for this PR. Since #5266, `RichDescription`/`RichHelp` are root exports of `@rjsf/core`, so any downstream code relying on `RichDescription.TEST_IDS.markdown` — the exact pattern this repo's own `RichDescription.test.tsx` used before this PR — now gets `undefined.markdown` with no migration pointer to the replacement (`MarkdownTemplate.TEST_IDS` from `@rjsf/core/markdown`).

---

jimmycallin:
Both fixed in `9a84b4f0c`.

**1. `MarkdownTemplate` not exported** — confirmed, and you're right that I'd spotted it while merge-testing and then lost it. It's now in the named export list in `packages/core/src/components/templates/index.ts`, so `import { MarkdownTemplate } from '@rjsf/core'` resolves like every other template. No bundle cost: the root export is the plain-text passthrough, and `markdown-to-jsx` still only enters through `@rjsf/core/markdown`.

**2. `TEST_IDS` dropped with no note** — also confirmed. They genuinely can't come back on `RichDescription`/`RichHelp`, since neither renders markdown any more, so this is a migration-pointer fix rather than a restoration. `CHANGELOG-v7.md` has a breaking-change bullet for the removal naming `MarkdownTemplate.TEST_IDS.markdown` from `@rjsf/core/markdown` as the replacement, and the markdown section of the v7 upgrade guide says the same.

Lint, knip, root typecheck, `cs-check` and the full `@rjsf/core` suite pass.

---

Unrelated, spotted while editing the changelog: `CHANGELOG-v7.md:52` on `v7` itself carries a stray `||||||| parent of c459bce95 (...)` merge marker, landed by #5266. Not from this branch, so I've left it alone — happy to send a one-line PR against `v7`.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -196,6 +220,8 @@ This closes [#3930](https://github.com/rjsf-team/react-jsonschema-form/issues/39
 
 The following new types were added to `@rjsf/utils`:
 
+- `MarkdownTemplateProps`: the props of a `MarkdownTemplate` (`children`, the markdown string)
```
heath-freenome: This needs to be synced with the later changes made to the interface... Including the generic types.

heath-freenome (dismissed):
One small fix and this is good to go

heath-freenome:
## Follow-up review (round 3)

Both round-2 issues are confirmed fixed as of current HEAD (`9a84b4f0c`):
- `MarkdownTemplate` is now properly exported from `templates/index.ts` and re-exported from `@rjsf/core`.
- `RichDescription.TEST_IDS`/`RichHelp.TEST_IDS` removal now has an explicit breaking-change bullet in `CHANGELOG-v7.md` pointing to `MarkdownTemplate.TEST_IDS`, also documented in the v7 upgrade guide.

Also re-verified the round-1 `uiSchema`-forwarding fix holds across all call sites (`ObjectField.tsx`, `UnsupportedField.tsx`, `DescriptionFieldTemplate.tsx`, `FileWidget.tsx`, `FieldHelpTemplate.tsx`).

Three new findings from the current HEAD:

**1. `packages/core/src/components/widgets/FileWidget.tsx:79`** — File-info markdown silently became flag-gated, with a style regression the docs don't call out

Pre-PR, `FilesInfo` called `<Markdown>{translateString(FilesInfo, ...)}</Markdown>` unconditionally (no ui:option gate existed for it at all), rendering the filename in bold via `**%1**`. Post-PR it routes through `RichDescription`, which only renders through `MarkdownTemplate` when `ui:enableMarkdownInDescription` is true — and separately, `enums.ts`'s `FilesInfo` string dropped the `**...**` bold markers entirely, so the old bold styling can't come back even with the flag enabled. An app that upgrades and follows the PR's own migration example (installs `@rjsf/core/markdown` and registers it as `templates.MarkdownTemplate`) but doesn't separately set `ui:enableMarkdownInDescription` on file-upload fields will see previously-bold file names silently downgrade to flat text. Neither `CHANGELOG-v7.md` nor the v7 upgrade guide singles out FileWidget's rendering becoming flag-gated (only "raw HTML parsing is disabled" is mentioned).

**2. `packages/snapshot-tests/package.json:37`** — `markdown-to-jsx` peerDependency isn't marked optional, inconsistent with the PR's own pattern

`@rjsf/snapshot-tests` declares `markdown-to-jsx` as a required peerDependency, unlike `@rjsf/core`'s own `markdown-to-jsx` peer, which is explicitly `optional: true`. No consuming theme package (mui, antd, mantine, react-bootstrap, shadcn, fluentui-rc, chakra-ui, daisyui) lists `markdown-to-jsx` as a dependency, even though all of them pull in `@rjsf/snapshot-tests` as a devDependency — so `pnpm install` reports an unmet-peer-dependency warning for each. Doesn't fail the build today only because pnpm isn't configured with strict peer dependencies, but it's a latent inconsistency with the PR's stated goal of making `markdown-to-jsx` fully optional.

**3. `packages/utils/src/enums.ts:79`** (minor/simplification) — Duplicate section-divider comment

Trimming the old markdown-specific comment left a duplicate `// Strings with replaceable parameters` header directly above `InvalidObjectField`, identical to the one already above `CycleDetected` a few lines earlier — reads as a copy-paste leftover to future editors.

---

jimmycallin:
Round 2 was already answered in `9a84b4f0c`; round 3 is handled in `f36fcbbd0`.

**1. `FileWidget` file info** — accurate, and the bold loss is deliberate. `FilesInfo` was the one string rendered through `markdown-to-jsx` unconditionally, which is exactly what this PR removes: with markdown opt-in, a `**%1**` default would render as literal asterisks for everyone who doesn't register a `MarkdownTemplate` — a worse regression than losing the bold. Restoring it without markdown would mean splitting the string into name/rest and marking it up in JSX, which takes the ordering freedom away from translators for a font weight.

So the behaviour stands and the docs now say so: `CHANGELOG-v7.md` and the upgrade guide both record that file info used to render markdown unconditionally, that the bold file name came from that, and that it now respects `ui:enableMarkdownInDescription` like every other translatable string.

**2. `markdown-to-jsx` peer on `@rjsf/snapshot-tests`** — leaving it required. `src/formTests.tsx` imports `@rjsf/core/markdown`, so anything running this suite genuinely needs `markdown-to-jsx`; marking the peer optional would say the opposite. Core's peer is optional because core works without it — snapshot-tests doesn't. I also can't reproduce the warnings: `pnpm install --force` on this branch reports no unmet peers for any theme (pnpm resolves missing peers automatically), and every theme's tests pass.

**3. Duplicate divider in `enums.ts`** — real, removed.

Also dropped while in there: a `|||||||` merge-conflict marker that survived a rebase into `CHANGELOG-v7.md` (line 53, above the daisyui `generateTheme()` bullet). Good catch trigger, thanks — it wasn't in either review.

Lint, root typecheck, `cs-check` and the test suite pass.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
heath-freenome: @jimmycallin update this to match the true description, with generic types... then merge

heath-freenome (approved):
https://github.com/rjsf-team/react-jsonschema-form/pull/5263/changes/f36fcbbd047e3d1b95449e9860076a514c3875cd#r3998200582

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Updated in `0eaa32ce9`:

> **`MarkdownTemplateProps<T, S, F>`**: the props a `MarkdownTemplate` receives, generic over the same `<T, S, F>` as every other template's props. `children` is the markdown string to render; `uiSchema` and `registry` are picked from the form's base props, so an implementation can read `ui:options`, reach `registry.translateString`, or fall back to another template.

That matches the interface as it now stands (`Pick<RJSFBaseProps<T, S, F>, 'uiSchema' | 'registry'> & { children: string }`), and the bullet is bolded like the entries in the sections below it.

PR #5265, opened 2026-09-10
chore!: remove ./lib deep-import exports and unused daisyui tailwind files

jimmycallin:

Every package publishes `./lib` and `./lib/*.js` export entries so that consumers can deep-import any built file. That makes every file in `lib/` public API: nothing can be moved, renamed or deleted without a breaking change, and knip has to treat all of `src/` as entry points (the daisyui override) instead of finding dead code. It also kept unreferenced tailwind files alive in `@rjsf/daisyui` because the wildcard made them "reachable".

Split out of #5263, which now only contains the markdown opt-in; the theme-composition half is #5266.

### Changes

- Every `package.json` exports only its root (`.`); the validators keep `./compileSchemaValidators`. The `./lib` and `./lib/*.js` entries are gone.
- `@rjsf/daisyui`: `Widgets` is now the generated widgets object like every other theme, not an alias of `generateWidgets`. The unreferenced `src/styles.css`, `tailwind.config.js` and `tailwind.config.json` are removed, with the `tailwindcss` dependency and `daisyui` devDependency that only they used.
- knip: the daisyui all-of-`src`-is-an-entry override is removed. The shadcn block now ignores `src/css/*.css` and the `tailwindcss` dependency explicitly, since the wildcard export was what previously made them count as used.

### Migration

Replace deep imports with public root exports: `import Form from '@rjsf/mui'`, and take templates and widgets from the themes' `Templates` and `Widgets` exports. For daisyui, replace `Widgets()` with `generateWidgets()` or use `Widgets` directly.

### Checklist

  - [x] I've added/updated tests and verified the existing snapshots


packages/utils/package.json
```diff
@@ -10,9 +10,7 @@
     ".": {
       "@rjsf/source": "./src/index.ts",
       "default": "./lib/index.js"
-    },
-    "./lib": "./lib/index.js",
-    "./lib/*.js": "./lib/*.js"
+    }
```
heath-freenome, line 13: What if people are deep-linking to a utility function?

jimmycallin:
might revisit this later but closing for now

PR #5043, opened 2026-04-23
Fix 5034 by converting Form to functional component

heath-freenome:
### Reasons for making this change
Fixes #5034

- Refactored `Form` into a functional component, moving a bunch of computations into helper function in `formUtils.ts`
- Updated `core/index.ts` to update exports of new `Form` related types
- Updated `withTheme.ts` to use new `FormRef` type
- Updated tests for `Form`, `NumberField` and `withTheme` to use the new `FormRef` type
- Added unit tests for all functions in `formUtils`

### Checklist

- [ ] **I'm updating documentation**
  - [ ] I've [checked the rendering](https://rjsf-team.github.io/react-jsonschema-form/docs/contributing) of the Markdown text I've added
- [ ] **I'm adding or updating code**
  - [ ] I've added and/or updated tests. I've run `npx nx run-many --target=build --exclude=@rjsf/docs && npm run test:update` to update snapshots, if needed.
  - [ ] I've updated [docs](https://rjsf-team.github.io/react-jsonschema-form/docs) if needed
  - [ ] I've updated the [changelog](https://github.com/rjsf-team/react-jsonschema-form/blob/main/CHANGELOG.md) with a description of the PR
- [ ] **I'm adding a new feature**
  - [ ] I've updated the playground with an example use of the feature

PR #5266, opened 2026-09-10
feat(core)!: export core components and generators in place of getDefaultRegistry

jimmycallin:
### The problem

Core had exactly one door into its own components: `getDefaultRegistry()`. To write a widget that wraps `TextWidget`, you asked for the entire registry and indexed into it.

```tsx
import { getDefaultRegistry } from '@rjsf/core';

const { TextWidget } = getDefaultRegistry().widgets;
```

Two things are wrong with that, and they're worth separating because only one of them is about size.

**It returns something that isn't real.** A `Registry` is what `Form` threads through the render tree at runtime: component maps *plus* `rootSchema`, `formContext`, `schemaUtils` and `translateString`. `getDefaultRegistry()` had no form to describe, so it invented placeholders for all four — and `Form` overwrote every one of them. Half the return value was genuine, half was fiction, and callers had to know which half. Asking for the default `TextWidget` should not require materializing a fake `rootSchema`.

**It couples reuse to the whole set.** `getDefaultRegistry()` builds every field, widget and template map. All of them are reachable from its return value, so no bundler can prove any of them unused. Reaching for one component means keeping all of them — not a tooling gap, but a direct consequence of returning an object that references everything.

### What changed

- Core exports every field, widget and template individually, so `import { TextWidget } from '@rjsf/core'` reaches exactly one component.
- Added `generateFields()`, `generateWidgets()`, `generateTemplates()` and `generateTheme()` for callers that want a whole set. `buildRegistry()` and `getTestRegistry()` are built on `generateTheme()`.
- **BREAKING:** `getDefaultRegistry()` is removed. Use a direct import, one of the generators, or `generateTheme()`.
- **BREAKING:** the `Form.getRegistry()` static is now a standalone `buildRegistry()` export, taking the same arguments. It never used `this`, and statics do not survive the eventual conversion of `Form` to a function component.
- Migration guide section, and docs examples switched from registry access to direct imports.
- Replaced the avoidable `any`s in the files this branch already touches.

### What this buys today

The code you write, before and after:

```tsx
// Before
import { getDefaultRegistry } from '@rjsf/core';

const { TextWidget } = getDefaultRegistry().widgets;

export function UppercaseTextWidget(props: WidgetProps) {
  return <TextWidget {...props} onChange={(v) => props.onChange(v?.toUpperCase())} />;
}
```

```tsx
// After
import { TextWidget } from '@rjsf/core';

export function UppercaseTextWidget(props: WidgetProps) {
  return <TextWidget {...props} onChange={(v) => props.onChange(v?.toUpperCase())} />;
}
```

Same component, honest import, no fabricated registry in the middle.

**This does not shrink your bundle, and it's important to be clear about why.** `buildRegistry()` calls `generateTheme()`, which names every component, and `Form` is the universal entry point — reached directly or through `withTheme()`. So any app that renders a form still ships the full set, exactly as before. Bundle size for real consumers is unchanged.

The reason it can't simply be made lazy is worth understanding, because it constrains every future attempt: component selection is **string-keyed at runtime**. `getWidget()` looks up `registeredWidgets[name]`, where `name` comes from `ui:widget` or from the schema's `type`/`format`. Schemas are data, frequently fetched from a server. No static analysis can know which names will appear, so no bundler can decide what is safe to drop. That's inherent to rjsf's model.

What the individual exports change is that reuse is no longer *coupled* to the whole set. A bundle that imports one component now contains one component (`TextWidget` is a quarter of a kilobyte, against ~62 KB for the full set). That doesn't help an app that also imports `Form` — but it's the precondition for the change that would.

### Open question, not proposed here

The change that would actually cut `Form`'s weight is to stop defaulting: require the component set to be passed in, as `withTheme()` already does, instead of `buildRegistry()` falling back to `generateTheme()`. The ceiling is real — the full set is roughly 62 KB of the ~78 KB a bare `import Form` costs.

The cost is rjsf's zero-config `<Form schema={...} validator={...} />`, which is a lot to give up. Flagging it for discussion; deliberately out of scope here.

### Typing changes

- `FormProps['onBlur']` and `FormProps['onFocus']` take `data: unknown` instead of `data: any`. `Form` only forwards this value to your handler and never inspects it, so it has no basis for claiming a type. **BREAKING** for handlers that used the value without narrowing — `(id, value: string) => ...` now needs a guard or a cast. The in-repo playground needed exactly that fix, which is a fair preview of the migration cost.
- `FormProps['onSubmit']`'s event is `SubmitEvent<HTMLFormElement>` rather than `SubmitEvent<any>`, which also gives you a typed `currentTarget`.
- Test-local `useState<any>` / `formData: any` / `errorSchema as any` replaced with the shapes those tests actually use.

Left alone: the `T = any` / `F extends FormContextType = any` generic defaults. They're the library-wide convention across ~900 sites, and narrowing them is a separate API decision that shouldn't ride along here.

### Migration

```diff
-import { getDefaultRegistry } from '@rjsf/core';
-
-const { widgets } = getDefaultRegistry();
-const { TextWidget } = widgets;
+import { TextWidget } from '@rjsf/core';
```

If you used the whole registry rather than picking components out of it:

```diff
-import { getDefaultRegistry } from '@rjsf/core';
-
-const { fields, widgets, templates } = getDefaultRegistry();
+import { generateTheme } from '@rjsf/core';
+
+const { fields, widgets, templates } = generateTheme();
```

If you called the `Form.getRegistry()` static:

```diff
-import Form from '@rjsf/core';
+import { buildRegistry } from '@rjsf/core';

-const registry = Form.getRegistry(props, schema, schemaUtils);
+const registry = buildRegistry(props, schema, schemaUtils);
```

Most code never needs it. Inside a field, widget or template, use the `registry` prop you already receive — it is the registry the `Form` actually rendered with, including its real `schemaUtils`. To build one for a test, use `getTestRegistry()` from `@rjsf/core/testing`.

`generateTheme()` returns only `{ fields, widgets, templates }`. If you were reading `rootSchema`, `formContext`, `schemaUtils` or `translateString` off the old return value, those were the placeholders `Form` discarded — take them from the `registry` prop your component already receives.

### Checklist

  - [x] I've verified the existing tests pass and root `typecheck`, `knip` and lint are clean


packages/core/src/CoreForm.ts
```diff
@@ -0,0 +1,15 @@
+import type { FormContextType, RJSFSchema, StrictRJSFSchema } from '@rjsf/utils';
+
+import BaseForm from './components/Form.tsx';
+import { generateTheme } from './Theme.ts';
+
+/** The core form renders with the default fields, widgets and templates underneath any passed as props. */
+export default class Form<
```
heath-freenome, line 7: @jimmycallin We are considering on replacing the Form class with 2 stateless functional component versions. The first version will accept a `formData` prop and will represent a "controlled" Form. The second will NOT take a `formData` prop at all and represent the "uncontrolled" Form. It is highly likely that one will wrap the other somehow. The design has not yet manifested into existence yet.

The thinking behind the two components is to deal with a ton of code related to eliminating/simpifying the `getSnapshotBeforeUpdate()`, `componentDidUpdate()`, `getDerivedStateFromProps()` processing of `formData` changed outside of the Form onChange processing. We'd welcome any ideas you may have around this too...

Perhaps these design changes will facilitate that work... OR make it even more challenging. At minimum we are DEFINITELY eliminating all remaining class-based components in v7, of which Form is the only one remaining.

jimmycallin: I would be happy to take on the work to make it into separate controlled/uncontrolled states. I don't think this PR will cause additional complexity for this, but to be sure we can postpone this until that's done.

heath-freenome: Cool! I'd love to brainstorm with you on the details of a plan before diving too deeply into implementation. Also, there are 2 PRs ([one](https://github.com/rjsf-team/react-jsonschema-form/pull/5043) from me and [one](https://github.com/rjsf-team/react-jsonschema-form/pull/5045) from Nick) that both attempted it in different ways before we realized it was a breaking change and that we might want to separate the controlled/uncontrolled aspects

jimmycallin: Cleaned this up, reduced its scope, and made it ready for review anyways!

heath-freenome:
**Shared mutable `formContext` singleton** — `packages/core/src/components/Form.tsx`

```js
const EMPTY_FORM_CONTEXT = {};
...
formContext: props.formContext ?? (EMPTY_FORM_CONTEXT as F),
```

Previously, `getDefaultRegistry()` returned `formContext: {} as F` — a fresh object literal on every call, so every `Form` instance without an explicit `formContext` got its own empty object. Now they all share one module-level object. Since `formContext` is documented as a place to stash lookup maps/caches, any field or widget that writes a property onto it will leak that mutation into every other `Form` on the page that also omits `formContext` — previously impossible.

As a bonus note: the comment justifying this ("keeps the same identity across renders") doesn't appear to be needed — `getDerivedStateFromProps` already reuses the previous `state.registry` via `deepEquals` when nothing changed, so per-instance identity was already stable. The singleton only adds the cross-instance leak risk, with no offsetting benefit.

**Suggestion:** create a fresh `{}` per `Form` instance (e.g. via an instance field or `useMemo`) rather than a module-level constant.

jimmycallin:
Good catch — fixed by dropping the module-level constant and going back to `{} as F` per call.

You're right that the identity argument doesn't hold: `getDerivedStateFromProps` already reuses `state.registry` when `deepEquals` says nothing changed, so the shared object bought nothing and only opened the cross-instance mutation path. Folded into the `getDefaultRegistry` commit, and the branch is rebased onto current `v7`.

heath-freenome (approved):

PR #5267, opened 2026-09-11
Drop the experimental_/Experimental_ prefix from stabilized FormProps and types

heath-freenome:

These features (defaultFormStateBehavior, customMergeAllOf, componentUpdateStrategy, and their corresponding @rjsf/utils types) have been experimental for several versions while their API stabilized. In v7 they're no longer experimental, so the prefix is removed from FormProps, the @rjsf/utils types/parameters, docs, and tests.

Also removes the repo-wide format/format-check npm scripts, which duplicated the per-package cs-format/cs-check scripts but over the whole repo (including markdown and configs) and weren't enforced in CI.

### Checklist


jimmycallin:
Regarding componentUpdateStrategy: My hope is that with https://github.com/jimmycallin/react-jsonschema-form/pull/50 and https://github.com/jimmycallin/react-jsonschema-form/pull/52 in place, componentUpdateStrategy will disappear and we won't need any custom update strategy. To me exposing this to users feels like a low-level leaky abstraction that exists to mitigate a less than optimal architecture. How do you feel about keeping it experimental until we can completely remove it instead?

heath-freenome:
> Regarding componentUpdateStrategy: My hope is that with [jimmycallin#50](https://github.com/jimmycallin/react-jsonschema-form/pull/50) and [jimmycallin#52](https://github.com/jimmycallin/react-jsonschema-form/pull/52) in place, componentUpdateStrategy will disappear and we won't need any custom update strategy. To me exposing this to users feels like a low-level leaky abstraction that exists to mitigate a less than optimal architecture. How do you feel about keeping it experimental until we can completely remove it instead?

Sounds good, I'll revert that one

jimmycallin (approved):

Issue #4964, opened 2026-02-20
Remove `extraErrorsBlockSubmit` flag on Form and make `extraErrors` always block validation


heath-freenome:
### Prerequisites

- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)

### What theme are you using?

core

### Is your feature request related to a problem? Please describe.

When the `extraErrorsBlockSubmit` flag was added back in the day, it was done to avoid a breaking change to how `extraErrors` affected the form submission. 

### Describe the solution you'd like

Given that users are likely to want them to block form submission, I recommend that we either remove the flag entirely OR change it to `extraErrorsAreWarnings` instead to switch the default behavior.

PR #5268, opened 2026-09-11
Fix 4964: replace extraErrorsBlockSubmit with extraErrorsAreWarnings

heath-freenome:

extraErrors now block form submission by default, the same as JSON Schema validation errors. Set extraErrorsAreWarnings to true to opt back into the old non-blocking behavior.

Fixes #4964

### Checklist


jimmycallin:
Reviewed the rename + default inversion. Core logic looks sound to me — `hasExtraErrors = errors.length > schemaValidation.errors.length` is actually an improvement over the old truthiness check, since an `extraErrors` object carrying no actual messages (e.g. `{ foo: { __errors: [] } }`) no longer blocks submission. I traced both branches of `validationDataMerge` and the comparison holds in each. Rename coverage is complete; the only remaining `extraErrorsBlockSubmit` hits are in `CHANGELOG.md` history and `versioned_docs/version-5.24.10`, which correctly stay as-is.

Three things:

1. **`packages/docs/docs/usage/validation.md:566`** — the async-validation section still ends with "these errors are 'display only' and will not block the user from submitting the form again." That's no longer true by default, and `form-props.md#extraErrors` links readers to this page ("See Validation for more information"), so the two docs pages now contradict each other. Probably wants to say extraErrors block by default and name `extraErrorsAreWarnings` as the opt-out.

2. **`packages/docs/docs/migration-guides/v7.x upgrade guide.md:98`** — the migration note doesn't mention the consequence that bites the common async pattern: an app that sets `extraErrors` from a failed server response and leaves the prop populated can now never resubmit, because every subsequent submit re-merges the stale `extraErrors` and calls `onError`. Previously those stale errors were merely displayed. One sentence telling upgraders to clear `extraErrors` on resubmit would save some debugging.

3. **`packages/core/src/components/Form.tsx:1352`** (pre-existing, not introduced here, but adjacent) — `validateFormWithFormData` merges only `extraErrors`, never `this.state.customErrors`, unlike `getDerivedStateFromProps` (L707) and the `getStateFromProps` side (L863). So a component-raised custom error is wiped from `state.errors` on a submit attempt and never blocks submission. If the goal is "extra errors behave like schema errors", customErrors look like the remaining inconsistency.

Minor test note: the case at `Form.behaviors.test.tsx:1380` moved from "prop omitted" to `extraErrorsAreWarnings: true`, so nothing now exercises schema-errors-plus-extraErrors under the new default. Low value, since that path is gated by the first clause of `hasError`.

heath-freenome:
Thanks for the thorough review! Pushed a follow-up commit (8c0fc1d2f) addressing all three points:

1. **`validation.md`** — updated the async-validation section to say `extraErrors` block by default, and name `extraErrorsAreWarnings` as the opt-out.
2. **v7 upgrade guide** — added a note that an app which sets `extraErrors` from a failed server response and leaves it populated will now re-block every resubmit (and re-call `onError`), and that it should clear `extraErrors` before/on resubmit, the same as a schema error.
3. **`customErrors` inconsistency** — agreed this was the right call. `validateFormWithFormData()` now merges `state.customErrors` the same way `getDerivedStateFromProps`/`liveValidate` already do, so a field/widget-raised custom error is no longer silently dropped and non-blocking on submit. Added a test using the documented "raise an error from a custom widget" pattern, which also turned up a stale bug in that same docs example — it was calling `onChange(value, [], errorSchema, id)`, but `WidgetProps.onChange` only takes `(value, errorSchema, id)`. Fixed that too.

While wiring in `customErrors`, I initially derived `hasError` by diffing `errors.length` across two chained `Form.mergeErrors()` calls, which turned out to rely on an undocumented invariant in `validationDataMerge` (its flat error list is never deduped, only its `errorSchema` tree is). Reworked it to check `toErrorList()` on each source directly instead, so it doesn't depend on that.

On the minor test note: agreed it's low value given that path is already gated by the first clause of `hasError`, so I left it as-is.

Full suite (1969 tests), lint, and format checks all green.

jimmycallin (approved):

Issue #3930, opened 2023-10-30
Format time in compliance with Draft 7


aiibe:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [ ] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

mui

### Version

5.13.0

### Current Behavior

The WidgetTime currently outputs time in the format `20:20:39` without including a timezone offset

### Expected Behavior

According to [JSON Schema Draft 7](https://json-schema.org/understanding-json-schema/reference/string#dates-and-times) compliance, the WidgetTime should output time with a timezone to be considered valid. The time format should include a timezone offset, something like 20:20:39+00:00

### Steps To Reproduce
Use the playground 
```json
{
  "title": "Date and time widgets",
  "type": "object",
  "properties": {
    "time": {
      "type": "string",
      "format": "time"
    }
  }
}
```
### Environment

```markdown
- OS: MacOS 13.2.1
- Node: v16.20.2
- npm: 8.19.4
```

### Anything else?

A quick solution is to override WidgetTime with a custom one.

nickgros:
I think there are two issues here:
1. Whether the widget shows (and maybe lets you change) timezone
2. Whether the formData time has a properly formatted timezone

I think we might have to solve the formData issue before the widget issue. It would be great to be in compliance with the JSONSchema spec but I think this could be a breaking change for anyone who uses time fields. I think we could put it in v5 with a feature flag and make it the default behavior in v6.

Once we properly support time zones, a TZ picker could be added to the widget, or we could add another widget to @rjsf/core to support that use case.

PRs are welcome to improve this issue!

PR #5270, opened 2026-09-11
Fix 3930: make time/date-time formats RFC 3339-compliant

heath-freenome:

Fixes #3930

format: "time" values produced by TimeWidget lacked a timezone offset, violating RFC 3339. Upgrades ajv-formats to 3.x, which now enforces the offset requirement, and updates TimeWidget/DateTimeWidget across @rjsf/core, @rjsf/antd, @rjsf/daisyui, and @rjsf/mantine to pad seconds and append/strip the local UTC offset accordingly.

Also adds format: "iso-time"/"iso-date-time" support, mapped to the same widgets, as an opt-out for fields that need to keep accepting offset-less values without a schema-breaking change.

New @rjsf/utils exports: localTimeToOffsetTime(), offsetTimeToLocalTime(), padTimeSeconds(), useTimeWidgetProps(), and getDateTimeLocalValue(), shared across the affected theme widgets to avoid duplicating the offset/seconds handling in each one.

### Checklist


packages/mantine/src/widgets/DateTime/DateTimeInput.tsx
```diff
@@ -17,13 +17,33 @@ const dateParser = (input: string, format: string) => {
...
+  }
+  const d = dayjs(input);
+  return d.isValid() ? d.toDate() : null;
+};
+
+const offsetValueFormatter = (date?: Date) => {
+  if (!date || !dayjs(date).isValid()) {
+    return '';
+  }
+  return date.toISOString();
```
jimmycallin: **High.** `offsetValueFormatter` calls `date.toISOString()`, but Mantine 9's `DateInput` declares `onChange?: (value: DateStringValue | null) => void` and calls `setValue(...)` with a **string** (`YYYY-MM-DD`) from `getDayProps` and `handlePresetSelect`. So for `format: "date-time"`/`"datetime"`, picking a date from the calendar dropdown — the primary interaction — throws `TypeError: date.toISOString is not a function`.

Reproduced by mocking `@mantine/dates`, capturing the props passed to `DateInput`, and calling `captured.onChange('2024-01-15')`. The existing tests pass only because they drive the typed-input path, where the custom `dateParser` returns a real `Date`. The pre-existing `dateFormat` path was already string-safe via `dayjs(date).format(...)`.

```suggestion
    return dayjs(date).toISOString();
```

Worth adding a test that drives `onChange` with a date string so this path is covered.

packages/utils/src/offsetTimeToLocalTime.ts
```diff
@@ -0,0 +1,11 @@
+const TIME_OFFSET_REGEX = /(?:Z|[+-]\d{2}:?\d{2})$/i;
```
jimmycallin: **Low.** `TIME_OFFSET_REGEX` requires offset minutes, but `ajv-formats` accepts `[+-]\d\d(?::?\d\d)?` — so a valid stored `20:20:39+05` doesn't match, passes through unchanged into `<input type="time">`, renders blank, and is then silently overwritten on the next edit. Making the minutes group optional covers it.

packages/utils/src/localTimeToOffsetTime.ts
```diff
@@ -0,0 +1,22 @@
+import pad from './pad.ts';
+
+/** Appends the browser's current local UTC offset to a bare `time` string (`HH:MM` or `HH:MM:SS`), producing
+ * a `time` compliant with the JSON Schema `time` format (RFC 3339 `full-time`), which requires a timezone.
+ * The wall-clock value itself is left untouched; only the offset annotation is added.
+ *
+ * @param time - A time string without a timezone offset
+ * @returns - The `time` string suffixed with `Z` (UTC) or a `+HH:MM`/`-HH:MM` offset
+ */
+export default function localTimeToOffsetTime(time: string) {
```
jimmycallin, line 10: **Low.** A `time` that already carries a foreign offset round-trips to a different instant: stored `13:10:30+02:00` displays as bare `13:10:30`, and a UTC-5 user re-picking the same displayed time commits `13:10:30-05:00`.

This is intentional and documented in the widget JSDoc, but the upgrade guide's "let the widget resave the value" advice doesn't warn that it also reinterprets values that already had an offset — worth a sentence there.

CHANGELOG-v7.md
```diff
@@ -62,10 +69,13 @@ should change the heading of the (upcoming) version to include a major version b
...
 
 ## @rjsf/validator-ajv8
 
 - Imports the `Ajv` class by name and reaches `ajv-formats` and `ajv/dist/standalone` through `.default`, which is where Node exposes those CommonJS modules' default export to ESM importers. No runtime change; the previous default-import spelling only typechecked under bundler resolution ([#5244](https://github.com/rjsf-team/react-jsonschema-form/pull/5244))
+- **BREAKING CHANGE** Upgraded `ajv-formats` from 2.x to 3.x. The `time` and `date-time` formats now require a timezone offset (`Z` or `±HH:MM`) per RFC 3339 instead of treating it as optional, so previously-valid `formData` without an offset will fail validation. The new `iso-time`/`iso-date-time` formats keep accepting offset-less values and are mapped to `TimeWidget`/`DateTimeWidget` like `time`/`date-time` are, so a field using them keeps its native time/date-time picker ([#3930](https://github.com/rjsf-team/react-jsonschema-form/issues/3930))
```
jimmycallin: **Low (docs accuracy).** `date-time` already required a timezone offset in `ajv-formats` 2.1.1: `date_time()` calls `time(dateTime[1], /* withTimeZone */ true)`, and the fast-mode regex ends `(?:z|[+-]\d\d(?::?\d\d)?)$` with no `?`. Only `time` changed in v3.

As written, users will think stored `date-time` data that previously validated may now fail and may do a needless backfill — suggest scoping the "previously optional" claim to `time`. Same wording in `packages/docs/docs/migration-guides/v7.x upgrade guide.md:111`.

jimmycallin (commented):
Reviewed the RFC 3339 time/date-time work. One blocking issue in the Mantine widget, two low-severity edge cases in the new `@rjsf/utils` helpers, and one docs accuracy nit. Everything else checked out: `iso-time`/`iso-date-time` are genuinely new in `ajv-formats` 3, the widget-map additions resolve correctly, `padTimeSeconds`/`getDateTimeLocalValue` handle empty and non-string values, the antd and daisyui commit paths are string-safe, and the new snapshot tests are appended at the end of `formTests` so they don't shift `useId` ids.

jimmycallin:
### Question: should `alt-datetime` interpret the picked wall-clock as local too?

Not introduced by this PR, but it sits right next to what you're changing, so it seems worth deciding while the time handling is being looked at.

The two date-time entry paths disagree about which timezone the user's picked wall-clock is in:

- `DateTimeWidget` → `localToUTC(dateString)` → `new Date(dateString).toJSON()`. `new Date("2024-01-15T13:10")` parses as **local**, so the picked time is converted local → UTC.
- `AltDateTimeWidget` → `useAltDateWidgetProps` → `toDateString(dateObject)` → `Date.UTC(year, month - 1, day, hour, minute, second)`. The picked components are treated as already being **UTC**, with no conversion.

For a user in UTC+2 picking 13:10 on 2024-01-15:

| path | committed value |
|---|---|
| `format: date-time` (native `datetime-local`) | `2024-01-15T11:10:00.000Z` |
| `uiSchema` `"ui:widget": "alt-datetime"` | `2024-01-15T13:10:00.000Z` |

Same gesture, same schema, two hours apart — and the `alt-datetime` one is off by the user's offset, which also makes it inconsistent with the local-offset convention `localTimeToOffsetTime` establishes in this PR.

Is the `Date.UTC` behavior intentional (alt-date fields meaning "enter the UTC components directly"), or is it a latent bug? If the latter, v7 seems like the moment to align it, and it'd be a small change in `toDateString`. Happy to open a separate issue if you'd rather keep it off this PR.

heath-freenome:
Good catch on the `alt-datetime`/`toDateString()` inconsistency — verified it: `Date.UTC(year, month - 1, day, hour, minute, second)` treats the picked wall-clock components as already being UTC, while `DateTimeWidget`'s `new Date(dateString).toJSON()` treats them as local and converts. So yes, a UTC+2 user picking 13:10 does get two different committed instants depending on which widget rendered the field, and the `alt-datetime` behavior is inconsistent with the local-offset convention this PR establishes for `time`.

I'd rather keep it out of this PR's scope, though — it's a separate code path (`toDateString`/`AltDateTimeWidget`, untouched here) and deserves its own investigation and test coverage rather than getting folded into an already-large diff. Happy to have you file the follow-up issue, or I can if that's easier — just let me know which you'd prefer.

packages/mantine/src/widgets/DateTime/DateTimeInput.tsx
heath-freenome: Confirmed — verified against the installed Mantine 9 types (`DateInputProps.onChange?: (value: DateStringValue | null) => void`, and `DateStringValue = string`), so `onChange` really does deliver a plain string for the calendar-click/preset path, not a `Date`.

Fixed by routing through `dayjs(value)` instead of assuming a `Date`:

```ts
const offsetValueFormatter = (value?: DateStringValue | Date | null) => {
  const d = dayjs(value);
  return value && d.isValid() ? d.toISOString() : '';
};
```

Also added the test you suggested — mocked `DateInput` to invoke `onChange` with a raw date string the way the calendar-click path does. I reproduced the exact `TypeError: date.toISOString is not a function` against the pre-fix code with it, then confirmed it passes with the fix. It lives in its own file (`DateTimeWidget.calendarOnChange.test.tsx`) since `vi.mock` hoists module-wide and would otherwise disable the real `DateInput` for the rest of the suite's tests.

Committed locally, pushing shortly.

packages/utils/src/offsetTimeToLocalTime.ts
heath-freenome: Confirmed — verified `ajv-formats` accepts an hour-only offset (`20:20:39+05` validates `true` against `format: time`). Made the minutes group optional to match its grammar:

```ts
const TIME_OFFSET_REGEX = /(?:Z|[+-]\d{2}(?::?\d{2})?)$/i;
```

Added a test case confirming `14:30:00+05` strips to `14:30:00`. Committed locally, pushing shortly.

packages/utils/src/localTimeToOffsetTime.ts
heath-freenome: Fair point — added a sentence to the upgrade guide's "let the widget resave" advice:

> Note this reinterprets the wall-clock value in the browser's *current* local offset, so if the stored value already carried a different offset (e.g. entered in another timezone), resaving it will represent a different instant rather than preserving the original one.

Committed locally, pushing shortly.

CHANGELOG-v7.md
heath-freenome: Good catch — verified against ajv-formats 2.1.1's own source: its `date-time` regex already had no trailing `?` on the offset group (only bare `time`'s did), so `date-time` always required an offset in 2.x. Corrected both this CHANGELOG-v7.md bullet and the same wording in the upgrade guide to scope "previously optional" to `time` alone, with a note that `date-time` already required one.

Committed locally, pushing shortly.

jimmycallin (approved):

PR #5272, opened 2026-09-11
feat(core)!: move getTestRegistry to @rjsf/core/testing

jimmycallin:

`getTestRegistry()` sat on the production root entry of `@rjsf/core` while depending on `@rjsf/validator-ajv8`, a devDependency. Consumers never got the validator, but core's `lib/` still pulled the AJV chain into every size measurement, and a test-only helper was part of the runtime API surface.

Split out of #5266 so that PR is only about the component exports.

### Changes

- `getTestRegistry()` is a named export of the new `@rjsf/core/testing` entry point (`src/testing.ts`). Its positional arguments are unchanged; `rootSchema` now defaults to `{}`.
- `@rjsf/validator-ajv8` becomes an optional peer dependency of `@rjsf/core` that only the testing entry needs.
- Three theme test helpers (`daisyui`, `shadcn`, `react-bootstrap`) passed an options object where positional arguments were expected, so the object landed in `rootSchema`. They now pass positional arguments.
- size-limit measures every export subpath as its own row and ignores only required peers, so a subpath's row includes what it actually reaches. The `@rjsf/core` row drops from 92 kB to 45 kB because the AJV chain is no longer counted against the main entry; the new `@rjsf/core/testing` row carries it at 76 kB. Node-only subpaths are declared per package in the config.
- Migration guide section and `CHANGELOG-v7.md` entry.

### Checklist

  - [x] I've verified the existing tests pass (`core`, `daisyui`, `shadcn`, `react-bootstrap`) and root `typecheck`, `knip` and lint are clean

### DIFF
--- a/.github/size-limit/.size-limit.js
+++ b/.github/size-limit/.size-limit.js
@@ -48,7 +48,7 @@ module.exports = released.flatMap(({ dir, pkg }) => {
   const optionalPeers = Object.keys(pkg.peerDependenciesMeta ?? {});
   const peers = [...Object.keys(pkg.peerDependencies ?? {}), 'react-dom'].filter((dep) => !optionalPeers.includes(dep));
   const { installed, own, canaries = [], nodeOnly = [] } = PACKAGES[pkg.name] ?? {};
-  const subpaths = Object.entries(pkg.exports).filter(
+  const subpaths = Object.entries(pkg.exports ?? {}).filter(
     ([key]) => key !== '.' && !key.startsWith('./lib') && !nodeOnly.includes(key),
   );
 
--- a/.github/size-limit/.size-limit.js
+++ b/.github/size-limit/.size-limit.js
@@ -6,8 +6,8 @@ const ROOT = join(__dirname, '..', '..');
 // Budgets only where one already existed; elsewhere the PR comment's delta column
 // catches a regression without the bump commits a per-theme budget invites.
 // Canaries are single-export imports that fail if tree-shaking regresses.
-// Every export subpath is measured as its own entry, except the Node-only ones.
-const NODE_ONLY_SUBPATHS = ['./compileSchemaValidators']; // imports fs
+// Every export subpath is measured as its own entry, except those listed in a package's `nodeOnly`, which import
+// Node built-ins such as `fs` and cannot be bundled for a browser.
 const PACKAGES = {
   '@rjsf/core': {
     installed: '47 kB',
@@ -19,8 +19,18 @@ const PACKAGES = {
     own: '19 kB',
     canaries: [{ label: 'getUiOptions', import: '{ getUiOptions }', limit: '1 kB' }],
   },
-  '@rjsf/validator-ajv8': { installed: '39 kB', own: '3 kB' },
+  '@rjsf/validator-ajv8': { installed: '39 kB', own: '3 kB', nodeOnly: ['./compileSchemaValidators'] },
+  '@rjsf/validator-ata': { nodeOnly: ['./compileSchemaValidators'] },
 };
+
+// The `default` condition is what a consumer's bundler resolves; a subpath without one has no browser entry to measure
+function subpathEntry(pkg, key, target) {
+  const entry = typeof target === 'string' ? target : target.default;
+  if (!entry) {
+    throw new Error(`${pkg.name} exports "${key}" without a "default" condition; add one or list it in nodeOnly`);
+  }
+  return entry;
+}
 
 const released = readdirSync(join(ROOT, 'packages'))
   .filter((dir) => existsSync(join(ROOT, 'packages', dir, 'package.json')))
@@ -33,25 +43,20 @@ const released = readdirSync(join(ROOT, 'packages'))
 module.exports = released.flatMap(({ dir, pkg }) => {
   const path = join(ROOT, 'packages', dir, 'lib', 'index.js');
   const deps = Object.keys(pkg.dependencies ?? {});
-  // react-dom is never declared but is always the host's to provide; devDependencies (which @rjsf/core/testing
-  // reaches) are never installed by a consumer. An optional peer is measured: the consumer only installs it for the
-  // subpath that needs it, so its cost belongs to that subpath's row.
+  // react-dom is never declared but is always the host's to provide. An optional peer is measured: the consumer only
+  // installs it for the subpath that needs it, so its cost belongs to that subpath's row.
   const optionalPeers = Object.keys(pkg.peerDependenciesMeta ?? {});
-  const peers = [
-    ...Object.keys(pkg.peerDependencies ?? {}),
-    ...Object.keys(pkg.devDependencies ?? {}),
-    'react-dom',
-  ].filter((dep) => !optionalPeers.includes(dep));
-  const { installed, own, canaries = [] } = PACKAGES[pkg.name] ?? {};
+  const peers = [...Object.keys(pkg.peerDependencies ?? {}), 'react-dom'].filter((dep) => !optionalPeers.includes(dep));
+  const { installed, own, canaries = [], nodeOnly = [] } = PACKAGES[pkg.name] ?? {};
   const subpaths = Object.entries(pkg.exports).filter(
-    ([key]) => key !== '.' && !key.startsWith('./lib') && !NODE_ONLY_SUBPATHS.includes(key),
+    ([key]) => key !== '.' && !key.startsWith('./lib') && !nodeOnly.includes(key),
   );
 
   return [
     { name: pkg.name, path, ignore: peers, ...(installed && { limit: installed }) },
     ...subpaths.map(([key, target]) => ({
       name: `${pkg.name}${key.slice(1)}`,
-      path: join(ROOT, 'packages', dir, typeof target === 'string' ? target : target.default),
+      path: join(ROOT, 'packages', dir, subpathEntry(pkg, key, target)),
       ignore: peers,
     })),
     ...(deps.length
--- a/.github/size-limit/.size-limit.js
+++ b/.github/size-limit/.size-limit.js
@@ -6,14 +6,13 @@ const ROOT = join(__dirname, '..', '..');
 // Budgets only where one already existed; elsewhere the PR comment's delta column
 // catches a regression without the bump commits a per-theme budget invites.
 // Canaries are single-export imports that fail if tree-shaking regresses.
+// Every export subpath is measured as its own entry, except the Node-only ones.
+const NODE_ONLY_SUBPATHS = ['./compileSchemaValidators']; // imports fs
 const PACKAGES = {
   '@rjsf/core': {
-    installed: '92 kB',
+    installed: '47 kB',
     own: '24 kB',
-    // getTestRegistry pulls the AJV chain into core's lib/, but it is a
-    // devDependency there — a consumer installing @rjsf/core never gets it.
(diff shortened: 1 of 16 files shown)


.github/size-limit/.size-limit.js
```diff
@@ -34,18 +33,33 @@ const released = readdirSync(join(ROOT, 'packages'))
...
-  // react-dom is never declared but is always the host's to provide.
-  const peers = [...Object.keys(pkg.peerDependencies ?? {}), 'react-dom'];
-  const { installed, own, extraIgnore = [], canaries = [] } = PACKAGES[pkg.name] ?? {};
+  // react-dom is never declared but is always the host's to provide; devDependencies (which @rjsf/core/testing
+  // reaches) are never installed by a consumer. An optional peer is measured: the consumer only installs it for the
+  // subpath that needs it, so its cost belongs to that subpath's row.
+  const optionalPeers = Object.keys(pkg.peerDependenciesMeta ?? {});
+  const peers = [
+    ...Object.keys(pkg.peerDependencies ?? {}),
+    ...Object.keys(pkg.devDependencies ?? {}),
```
heath-freenome: `peers` (used as `ignore` for every row, including the new per-subpath ones) now includes every devDependency of the package — not just its optional peers. `ajv` is a plain devDependency of `@rjsf/core` (used for core's own tests), so it ends up in this list and gets excluded from the `@rjsf/core/testing` row too, even though `testing.ts` imports `@rjsf/validator-ajv8`, which depends on `ajv` for the actual validation engine (~36 kB per this file's own PACKAGES table).

That defeats the PR's stated goal for this exact row ("a subpath's row includes what it actually needs") — a future AJV regression reachable only through the testing subpath won't move the CI number at all.

Might be worth deriving `ignore` per-subpath instead of reusing one `peers` list for every row, so only the *unrelated* devDependencies (not deps the subpath itself pulls in) are excluded.

.github/size-limit/.size-limit.js
```diff
@@ -6,14 +6,13 @@ const ROOT = join(__dirname, '..', '..');
 // Budgets only where one already existed; elsewhere the PR comment's delta column
 // catches a regression without the bump commits a per-theme budget invites.
 // Canaries are single-export imports that fail if tree-shaking regresses.
+// Every export subpath is measured as its own entry, except the Node-only ones.
+const NODE_ONLY_SUBPATHS = ['./compileSchemaValidators']; // imports fs
```
heath-freenome: `NODE_ONLY_SUBPATHS` is a hand-maintained, package-agnostic string allowlist with nothing tying it to the actual reason a subpath is Node-only (importing `fs`). It's correct today only because `@rjsf/validator-ajv8` and `@rjsf/validator-ata` happen to name their Node-only export the same thing.

If a later PR adds a Node-only subpath under a different name, or reuses `./compileSchemaValidators` for something that isn't Node-only, it'll either get silently measured as a browser bundle (bogus size number) or crash the bundler on an `fs` import — with no obvious link back to this array. Not blocking, but worth a comment noting the coincidence, or deriving this from something more direct (e.g. an explicit per-package flag in `PACKAGES`).

.github/size-limit/.size-limit.js
```diff
@@ -34,18 +33,33 @@ const released = readdirSync(join(ROOT, 'packages'))
...
+  const subpaths = Object.entries(pkg.exports).filter(
+    ([key]) => key !== '.' && !key.startsWith('./lib') && !NODE_ONLY_SUBPATHS.includes(key),
+  );
 
   return [
     { name: pkg.name, path, ignore: peers, ...(installed && { limit: installed }) },
-    ...(deps.length || extraIgnore.length
+    ...subpaths.map(([key, target]) => ({
+      name: `${pkg.name}${key.slice(1)}`,
+      path: join(ROOT, 'packages', dir, typeof target === 'string' ? target : target.default),
```
heath-freenome: `typeof target === 'string' ? target : target.default` assumes every object-shaped `exports` entry defines a `default` condition. True for all 16 packages today, but a future subpath export with only e.g. `@rjsf/source`/`types`/`import` conditions and no `default` makes `target.default` `undefined`, and `path.join(ROOT, 'packages', dir, undefined)` throws — breaking size-limit config generation for the *entire* monorepo (this is a single `module.exports` shared across all packages), with an error that doesn't point back to the missing key. Low risk, but a `target.default ?? target['@rjsf/source']` fallback or an explicit check would make the failure mode less mysterious.

packages/core/package.json
```diff
@@ -43,8 +47,14 @@
   },
   "peerDependencies": {
     "@rjsf/utils": "workspace:^",
+    "@rjsf/validator-ajv8": "workspace:^",
```
heath-freenome, line 50: `@rjsf/shadcn` and `@rjsf/react-bootstrap`'s test helpers now import `getTestRegistry` from `@rjsf/core/testing` (which imports `@rjsf/validator-ajv8` internally), but neither package declares `@rjsf/validator-ajv8` as its own devDependency. Resolution of that bare specifier walks up from `packages/core/src/` (where `testing.ts` lives), so it only works today because `@rjsf/core` *also* still lists `@rjsf/validator-ajv8` as a devDependency in addition to the new optional peer here.

If a future contributor sees the optional-peer declaration and removes the seemingly-redundant devDependency from `@rjsf/core`, shadcn's and react-bootstrap's test suites break with an unresolvable-module error that has no obvious connection to that change. `@rjsf/daisyui` doesn't have this problem since it already declares `@rjsf/validator-ajv8` itself. Worth adding it as a devDependency to shadcn and react-bootstrap directly rather than relying on core's transitively.

CHANGELOG-v7.md
```diff
@@ -35,6 +35,7 @@ should change the heading of the (upcoming) version to include a major version b
 - **BREAKING CHANGE** Replaced the `extraErrorsBlockSubmit` prop with `extraErrorsAreWarnings`, inverting the default: `extraErrors` now block form submission unless `extraErrorsAreWarnings` is set to `true`, fixing [#4964](https://github.com/rjsf-team/react-jsonschema-form/issues/4964)
 - Fixed `validateFormWithFormData()` (used by form submission and the `validateForm()` instance method) silently dropping a `customError` raised by a field/widget's `onChange` on submit; it's now merged in and blocks submission the same way schema and (non-warning) `extraErrors` do
 - **BREAKING CHANGE:** Removed the private `_internalFormWrapper` prop from `FormProps` and `ThemeProps`; its only consumer was the removed `@rjsf/semantic-ui` theme. Use `tagName` to render a different element in place of `<form>`
+- **BREAKING CHANGE:** `getTestRegistry()` moved to `@rjsf/core/testing` as a named export; `@rjsf/validator-ajv8`, which it needs, is an optional peer dependency that only this entry point requires. Its positional arguments remain supported. Fixed three theme test helpers that previously passed an options object instead of positional arguments
```
heath-freenome, line 39: This PR also fixes real bugs in `daisyui`, `shadcn`, and `react-bootstrap`'s test helpers (all three previously passed an options object into `getTestRegistry` where positional args were expected, silently dropping `templates`/`widgets`), but only this one bullet is added, under `@rjsf/core`.

This file's own header instructs listing a PR under each affected package's heading when it affects multiple packages. Might be worth a short bullet under `@rjsf/daisyui`, `@rjsf/shadcn`, and `@rjsf/react-bootstrap` too, so anyone scanning those themes' sections sees their test helpers changed.

heath-freenome (commented):
Automated review found a few issues worth a look — one likely defeats a core goal of this PR (the `@rjsf/core/testing` size row still excludes `ajv`), the rest are lower-severity robustness/documentation notes.

.github/size-limit/.size-limit.js
jimmycallin: Good catch, this did hide ajv. Only required peers and react-dom are ignored now; the testing row went from 45 kB to 76 kB with the AJV chain counted. Fixed in ef5f076c8.

.github/size-limit/.size-limit.js
jimmycallin: Moved to a per-package `nodeOnly` list in `PACKAGES`, next to the budgets, so the exclusion sits with the package that owns the subpath. Fixed in ef5f076c8.

.github/size-limit/.size-limit.js
jimmycallin: Added a `subpathEntry()` helper that throws naming the package and subpath when there is no `default` condition, rather than falling back to another condition, since only `default` is what a consumer's bundler resolves. Fixed in ef5f076c8.

packages/core/package.json
jimmycallin: Added `@rjsf/validator-ajv8` as a devDependency of `@rjsf/shadcn` and `@rjsf/react-bootstrap`, matching daisyui. Fixed in ef5f076c8.

CHANGELOG-v7.md
jimmycallin: Added a bullet under `@rjsf/daisyui` (new heading), `@rjsf/react-bootstrap` and `@rjsf/shadcn`. Fixed in ef5f076c8.

.github/size-limit/.size-limit.js
```diff
@@ -34,18 +43,28 @@ const released = readdirSync(join(ROOT, 'packages'))
...
   const deps = Object.keys(pkg.dependencies ?? {});
-  // react-dom is never declared but is always the host's to provide.
-  const peers = [...Object.keys(pkg.peerDependencies ?? {}), 'react-dom'];
-  const { installed, own, extraIgnore = [], canaries = [] } = PACKAGES[pkg.name] ?? {};
+  // react-dom is never declared but is always the host's to provide. An optional peer is measured: the consumer only
+  // installs it for the subpath that needs it, so its cost belongs to that subpath's row.
+  const optionalPeers = Object.keys(pkg.peerDependenciesMeta ?? {});
+  const peers = [...Object.keys(pkg.peerDependencies ?? {}), 'react-dom'].filter((dep) => !optionalPeers.includes(dep));
+  const { installed, own, canaries = [], nodeOnly = [] } = PACKAGES[pkg.name] ?? {};
+  const subpaths = Object.entries(pkg.exports).filter(
```
heath-freenome: `Object.entries(pkg.exports)` assumes every released, non-snapshot-tests package has an `exports` field, with no guard analogous to the one `subpathEntry()` gets for a missing `default` condition.

A future theme package (or a restored/renamed one) that ships only `main`/`module` without an `exports` map — still valid Node/npm resolution — would make `pkg.exports` undefined, so `Object.entries(undefined)` throws inside `released.flatMap(...)`. That's the single `module.exports` shared by every package, so it would crash size-limit config generation for the entire monorepo's CI run, not just for the offending package.

heath-freenome (commented):
Re-reviewed after the latest fixes — all five prior review comments look genuinely addressed. One new latent issue found, noted inline.

.github/size-limit/.size-limit.js
heath-freenome: I'll approve this assuming you fix the small issue

heath-freenome (approved):

.github/size-limit/.size-limit.js
jimmycallin: Fixed in d68060c68 — `Object.entries(pkg.exports ?? {})`. A package that ships only `main`/`module` now contributes no subpath rows instead of throwing; its main row still measures `lib/index.js` as before. Verified the generated config is byte-identical for the current 16 packages (28 entries).

Also rebased onto current `v7`.

PR #5274, opened 2026-09-12
fix 5250: sanitize stale data for a dependencies/if branch nested inside a property

### LINKED ISSUE #5250: A nested `dependencies` branch switch never sanitizes, so a narrowed select keeps an invalid value
### Prerequisites

### What theme are you using?

antd

### Version

@rjsf/core` 6.7.1 (also present in 6.8.0 and on `main`)

## Description

When a `dependencies` branch narrows the allowed values of a sibling field, the
old value is left in the form data. The select then shows one of the remaining
options while the data still holds the old one, and the field reports an error
the user cannot clear, because every option the widget offers is already the one
being displayed.

`sanitizeDataForNewSchema` handles this correctly when called directly. The
problem is when `Form` calls it. In `getStateFromProps` the sanitize step is
gated on the ROOT level retrieved schema having changed:

```ts
computedRetrievedSchema = this.updateRetrievedSchema(retrievedSchema ?? schemaUtils.retrieveSchema(rootSchema, formData));
if (shouldSanitize && !preventInfiniteSanitize.includes(formHash) &&
    !deepEquals(computedRetrievedSchema, state.retrievedSchema)) {
  // sanitizeDataForNewSchema(...)
}
```

A dependency deeper in the tree never changes the root schema, so the whole step
is skipped. Instrumenting `sanitizeDataForNewSchema` on the same schema at
different depths:

| Dependency location | root retrievedSchema changed | sanitize calls | result |
| --- | --- | --- | --- |
| at the root | true | 1 | `{f: 0, b: 0}`, valid |
| one level down | false | 0 | `{m: {f: 0, b: 1}}`, 3 errors |
| two levels down | false | 0 | same |

### Expected Behavior

A value that the newly active `dependencies` branch no longer allows should be
sanitized, regardless of how deeply the owning object is nested. Today that only
happens when the dependency is on the root schema.

### Steps To Reproduce

Start with this [playground](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJtIjp7InQiOjEsImIiOjF9fSwic2NoZW1hIjp7InR5cGUiOiJvYmplY3QiLCJwcm9wZXJ0aWVzIjp7Im0iOnsidHlwZSI6Im9iamVjdCIsInRpdGxlIjoiTWVzc2FnZSIsInJlcXVpcmVkIjpbInQiLCJiIl0sInByb3BlcnRpZXMiOnsidCI6eyJ0aXRsZSI6IkZyYW1lIGZvcm1hdCIsInR5cGUiOiJpbnRlZ2VyIiwiZGVmYXVsdCI6MCwib25lT2YiOlt7InRpdGxlIjoiU3RhbmRhcmQiLCJlbnVtIjpbMF19LHsidGl0bGUiOiJGRCIsImVudW0iOlsxXX1dfSwiYiI6eyJ0aXRsZSI6IkJpdC1SYXRlIFN3aXRjaCIsInR5cGUiOiJpbnRlZ2VyIiwiZGVmYXVsdCI6MH19LCJkZXBlbmRlbmNpZXMiOnsidCI6eyJvbmVPZiI6W3sicHJvcGVydGllcyI6eyJ0Ijp7ImVudW0iOlswXX0sImIiOnsib25lT2YiOlt7InR5cGUiOiJpbnRlZ2VyIiwidGl0bGUiOiJEaXNhYmxlIiwiZW51bSI6WzBdfV19fX0seyJwcm9wZXJ0aWVzIjp7InQiOnsiZW51bSI6WzFdfSwiYiI6eyJvbmVPZiI6W3sidHlwZSI6ImludGVnZXIiLCJ0aXRsZSI6IkRpc2FibGUiLCJlbnVtIjpbMF19LHsidHlwZSI6ImludGVnZXIiLCJ0aXRsZSI6IkVuYWJsZSIsImVudW0iOlsxXX1dfX19XX19fX19LCJ1aVNjaGVtYSI6e30sInRoZW1lIjoiZGVmYXVsdCIsImxpdmVTZXR0aW5ncyI6eyJzaG93RXJyb3JMaXN0IjoidG9wIiwidmFsaWRhdGUiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsIm5vSHRtbDVWYWxpZGF0ZSI6dHJ1ZSwicmVhZG9ubHkiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlLCJleHBlcmltZW50YWxfY29tcG9uZW50VXBkYXRlU3RyYXRlZ3kiOiJjdXN0b21EZWVwIiwiZXhwZXJpbWVudGFsX2RlZmF1bHRGb3JtU3RhdGVCZWhhdmlvciI6eyJhcnJheU1pbkl0ZW1zIjoicG9wdWxhdGUiLCJhbGxPZiI6InNraXBEZWZhdWx0cyIsImNvbnN0QXNEZWZhdWx0cyI6ImFsd2F5cyIsImVtcHR5T2JqZWN0RmllbGRzIjoicG9wdWxhdGVBbGxEZWZhdWx0cyIsIm1lcmdlRGVmYXVsdHNJbnRvRm9ybURhdGEiOiJ1c2VGb3JtRGF0YUlmUHJlc2VudCIsIm5lc3RlZERlZmF1bHRzUHJlY2VkZW5jZSI6ImlubmVybW9zdFdpbnMifSwidXNlRmFsbGJhY2tGaWVsZCI6ZmFsc2UsInVzZUZhbGxiYWNrVWlGb3JVbnN1cHBvcnRlZFR5cGUiOmZhbHNlLCJsaXZlVmFsaWRhdGUiOmZhbHNlLCJpZFByZWZpeCI6InJvb3QiLCJpZFNlcGFyYXRvciI6Il8ifSwidmFsaWRhdG9yIjoiQUpWOCIsInNhbXBsZU5hbWUiOiJCbGFuayJ9)

`T` is FD and `B` is Enable, which is valid. Change `T` to Standard, where the schema only allows Disable.

Expected: `B` becomes Disable and the form is valid.
Actual: `B` displays Disable but the data is still 1, and the form reports
"must be equal to one of the allowed values" with no way to clear it. Setting `B`
to Disable first and then changing `T` works, which shows the value is simply
never sanitized.

Removing the `m` wrapper so the dependency sits at the root makes it work.

```json
{
  "type": "object",
  "properties": {
    "m": {
      "type": "object",
      "title": "Message",
      "required": ["t", "b"],
      "properties": {
        "t": {
          "title": "Frame format",
          "type": "integer",
          "default": 0,
          "oneOf": [
            { "title": "Standard", "enum": [0] },
            { "title": "FD", "enum": [1] }
          ]
        },
        "b": { "title": "Bit-Rate Switch", "type": "integer", "default": 0 }
      },
      "dependencies": {
        "t": {
          "oneOf": [
            {
              "properties": {
                "t": { "enum": [0] },
                "b": { "oneOf": [{ "type": "integer", "title": "Disable", "enum": [0] }] }
              }
            },
            {
              "properties": {
                "t": { "enum": [1] },
                "b": {
                  "oneOf": [
                    { "type": "integer", "title": "Disable", "enum": [0] },
                    { "type": "integer", "title": "Enable", "enum": [1] }
                  ]
                }
              }
            }
          ]
        }
      }
    }
  }
}
```

```json
{ "m": { "t": 1, "b": 1 } }
```

### Environment

```markdown
- OS:
- Node:
- npm:
```

### Anything else?

_No response_

heath-freenome:

Fixes #5250

A dependencies/if branch switch nested inside an object property never sanitized a sibling field's now-invalid value. retrieveSchema() only resolves dependencies/if declared directly on the schema it's given, so the root-level retrieved schema Form used to decide whether to sanitize never reflected a conditional resolved deeper in the tree, and sanitizeDataForNewSchema() only resolved $ref (not dependencies/if/allOf) on each property's keyed schema before comparing it.

- Form now also treats sanitization as needed when the schema has a dependencies/if nested below its root (via the new schemaHasNestedConditional() utility, which resolves $ref, patternProperties, tuple items and allOf/anyOf/oneOf to find it, and merges local sibling keywords with a ref target rather than discarding them), since the plain root-schema-equality check can't see it.
- sanitizeDataForNewSchema() now resolves each property's old/new keyed schema fully (dependencies, if/then/else, allOf, not just $ref) before comparing them, including per-element for array items, and skips a redundant second resolution when a key's old and new raw schema are identical.

### Checklist

### DIFF
--- a/packages/core/src/components/Form.tsx
+++ b/packages/core/src/components/Form.tsx
@@ -53,6 +53,7 @@ import {
   ID_KEY,
   getUsedFormData,
   getFieldNames,
+  schemaHasNestedConditional,
   ANY_OF_KEY,
   ONE_OF_KEY,
 } from '@rjsf/utils';
@@ -616,6 +617,11 @@ export default class Form<
     let computedRetrievedSchema: S;
     let wasSanitized = false;
     const preventInfiniteSanitize: string[] = [];
+    // A `dependencies`/`if` branch switch nested inside an object property never changes the ROOT retrieved
+    // schema (only the schema's own top-level `dependencies`/`if` get resolved into it), so comparing
+    // `computedRetrievedSchema` to `state.retrievedSchema` below can't detect it. When the schema has such a
+    // nested conditional, skip that (root-only) short-circuit so sanitization is still attempted (#5250).
+    const hasNestedConditionalSchema = shouldSanitize && schemaHasNestedConditional(rootSchema, rootSchema);
     do {
       formData = schemaUtils.getDefaultFormState(
         rootSchema,
@@ -631,10 +637,10 @@ export default class Form<
       if (
         shouldSanitize &&
         !preventInfiniteSanitize.includes(formHash) &&
-        !deepEquals(computedRetrievedSchema, state.retrievedSchema)
+        (hasNestedConditionalSchema || !deepEquals(computedRetrievedSchema, state.retrievedSchema))
       ) {
         // Sanitize the form data if shouldSanitize is true, we haven't already processed this same formData AND
-        // we have a different retrieved schema from when we last ran the state
+        // either the retrieved schema changed or the schema has a nested conditional that the check above can't see
         const sanitizedFormData = schemaUtils.sanitizeDataForNewSchema(
           computedRetrievedSchema,
           state.retrievedSchema,
--- a/packages/utils/src/schemaHasNestedConditional.ts
+++ b/packages/utils/src/schemaHasNestedConditional.ts
@@ -1,0 +1,78 @@
+import {
+  ADDITIONAL_PROPERTIES_KEY,
+  ALL_OF_KEY,
+  ANY_OF_KEY,
+  DEPENDENCIES_KEY,
+  IF_KEY,
+  ITEMS_KEY,
+  ONE_OF_KEY,
+  PATTERN_PROPERTIES_KEY,
+  PROPERTIES_KEY,
+  REF_KEY,
+} from './constants.ts';
+import findSchemaDefinition from './findSchemaDefinition.ts';
+import isObject from './isObject.ts';
+import type { RJSFSchema, StrictRJSFSchema } from './types.ts';
+
+/** Recursively checks whether the given raw `schema` contains a `dependencies` or `if` keyword anywhere below its
+ * top level, e.g. inside a nested object's `properties`, a `$ref`, an array's tuple `items`, or a
+ * `patternProperties` entry. `retrieveSchema()` only resolves the `dependencies`/`if` declared directly on the
+ * schema it is given, so a root-level retrieved schema never reflects a conditional branch switch that happens
+ * deeper in the tree. `Form` uses this to detect when a comparison of root-level retrieved schemas can't be trusted
+ * to decide whether sanitization is needed.
+ *
+ * @param schema - The raw schema node to search
+ * @param rootSchema - The root schema, used to resolve any `$ref`s encountered while searching
+ * @param [atRoot=true] - Whether `schema` is the root of the search, whose own `dependencies`/`if` don't count
+ * @param [seenRefs=[]] - The `$ref`s already resolved along this branch of the search, to guard against cycles
+ * @returns - True if a `dependencies` or `if` keyword exists below the root of the schema
+ */
+export default function schemaHasNestedConditional<S extends StrictRJSFSchema = RJSFSchema>(
+  schema: S | boolean | undefined,
+  rootSchema: S,
+  atRoot = true,
+  seenRefs: string[] = [],
+): boolean {
+  if (!isObject(schema)) {
+    return false;
+  }
+  let resolved: S = schema;
+  let refs = seenRefs;
+  if (typeof resolved[REF_KEY] === 'string') {
+    const ref: string = resolved[REF_KEY];
+    if (refs.includes(ref)) {
+      return false;
+    }
+    refs = [...refs, ref];
+    try {
+      const target = findSchemaDefinition<S>(ref, rootSchema);
+      if (isObject(target)) {
+        // Merge, don't replace: sibling keywords declared alongside `$ref` are preserved and applied by
+        // `retrieveSchema()` (via `{ ...refSchema, ...localSchema }`), so a local `dependencies`/`if` next to a
+        // `$ref` must still be seen here, not just ones found on the ref target.
+        const { [REF_KEY]: _ref, ...localSchema } = resolved;
+        resolved = { ...target, ...localSchema } as S;
+      }
+    } catch {
+      // An unresolvable $ref will already have surfaced elsewhere (e.g. when rendering the field); treat it as
+      // having no nested conditional here rather than letting this best-effort check throw.
+      return false;
+    }
+  }
+  if (!atRoot && (DEPENDENCIES_KEY in resolved || IF_KEY in resolved)) {
+    return true;
+  }
+  const properties = resolved[PROPERTIES_KEY];
+  const patternProperties = resolved[PATTERN_PROPERTIES_KEY];
+  const items = resolved[ITEMS_KEY] as S | boolean | (S | boolean)[] | undefined;
+  const nestedSchemas: (S | boolean | undefined)[] = [
+    ...(isObject(properties) ? (Object.values(properties) as (S | boolean)[]) : []),
+    ...(isObject(patternProperties) ? (Object.values(patternProperties) as (S | boolean)[]) : []),
+    ...(Array.isArray(items) ? items : [items]),
+    resolved[ADDITIONAL_PROPERTIES_KEY] as S | boolean | undefined,
+    ...((resolved[ALL_OF_KEY] as (S | boolean)[] | undefined) ?? []),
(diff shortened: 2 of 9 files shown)


packages/core/src/components/Form.tsx
```diff
@@ -616,6 +617,11 @@ export default class Form<
     let computedRetrievedSchema: S;
     let wasSanitized = false;
     const preventInfiniteSanitize: string[] = [];
+    // A `dependencies`/`if` branch switch nested inside an object property never changes the ROOT retrieved
+    // schema (only the schema's own top-level `dependencies`/`if` get resolved into it), so comparing
+    // `computedRetrievedSchema` to `state.retrievedSchema` below can't detect it. When the schema has such a
+    // nested conditional, skip that (root-only) short-circuit so sanitization is still attempted (#5250).
+    const hasNestedConditionalSchema = shouldSanitize && schemaHasNestedConditional(rootSchema, rootSchema);
```
nickgros: Note from LLM review:

----

schemaHasNestedConditional(rootSchema, rootSchema) runs a full recursive schema walk on every getStateFromProps call (once per prop/state change, before the sanitize loop) rather than being cached/memoized keyed on the schema reference. For large/deep schemas this is a repeated O(schema size) walk on every render-triggering update — not incorrect, but a perf tradeoff the PR doesn't mention.

**Suggested fix**: Cache it, keyed off schemaUtils, instead of recomputing on every getStateFromProps call.

Why it's safe: rootSchema only changes when schemaUtils is recreated (line 599, gated by the already-cheap doesSchemaUtilsDiffer check).  schemaHasNestedConditional(rootSchema, rootSchema) is a pure function of rootSchema alone — it doesn't depend on formData — so its result is invariant across every call that reuses the same schemaUtils/rootSchema (i.e. every render/prop-change where only formData moved).

Simple fix: move the computation inside the if (!schemaUtils || schemaUtils.doesSchemaUtilsDiffer(...)) branch and stash the result alongside schemaUtils (either as a field on state, e.g. state.hasNestedConditionalSchema, recomputed only when schemaUtils is recreated, or a WeakMap<rootSchema, boolean> memo if you don't want to touch state shape). Then at line 624, read the cached value instead of calling the walker unconditionally:

 ```ts
if (!schemaUtils || schemaUtils.doesSchemaUtilsDiffer(...)) {
  schemaUtils = createSchemaUtils<T, S, F>(...);
  hasNestedConditionalSchema = schemaHasNestedConditional(schemaUtils.getRootSchema(), schemaUtils.getRootSchema());
} else {
  hasNestedConditionalSchema = state.hasNestedConditionalSchema;
}
 ```

This collapses the walk from "once per getStateFromProps call" to "once per actual schema change" — the same amortization schemaUtils itself already gets — with no behavior change and a one-field addition to state.

packages/utils/src/schemaHasNestedConditional.ts
```diff
@@ -0,0 +1,78 @@
...
+  rootSchema: S,
+  atRoot = true,
+  seenRefs: string[] = [],
+): boolean {
+  if (!isObject(schema)) {
+    return false;
+  }
+  let resolved: S = schema;
+  let refs = seenRefs;
+  if (typeof resolved[REF_KEY] === 'string') {
```
nickgros, line 43: The $ref-resolve-and-merge step in schemaHasNestedConditional.ts duplicates logic that already exists (and is already tested) in retrieveSchema.ts.

 Where:
 - schemaHasNestedConditional.ts (new, lines ~343–357): finds $ref, guards cycles with a local seenRefs list, calls findSchemaDefinition, then does resolved = {
   ...target, ...localSchema } to keep local sibling keywords.
 - retrieveSchema.ts:374–383 (existing): same shape — finds $ref, guards cycles with recurseList, calls findSchemaDefinition, then does resolvedSchema = {
   ...refSchema, ...localSchema, [RJSF_REF_KEY]: $ref }.

 These are two independent implementations of "resolve one $ref hop, merging local siblings over the ref target, without infinite-looping on cycles." The new file re-derives it by hand instead of reusing the existing one.

----

IMO, no need to refactor until we get a 3rd case, but just leaving this here for you to consider.

nickgros (approved):

jimmycallin (approved):

packages/core/src/components/Form.tsx
heath-freenome: Good catch, fixed in 6bbf4009f — `hasNestedConditionalSchema` is now destructured from `state` and only recomputed inside the `doesSchemaUtilsDiffer` branch (right alongside `schemaUtils` itself), then threaded through `nextState` for reuse. Exactly the approach you suggested.

packages/utils/src/schemaHasNestedConditional.ts
heath-freenome: Agreed, leaving this as-is for now per your suggestion — will revisit if a third case for this ref-merge pattern shows up.

Separately: while re-reviewing this PR I found that `schemaHasNestedConditional()` didn't descend into an `if`'s `then`/`else` branches or a root-level schema-style `dependencies` value, and `sanitizeDataForNewSchema()` only resolved an array's `items` schema when it had a *direct* `$ref` (missing `allOf`-wrapped refs) — both gaps meant the fix didn't fully close #5250 for those shapes. Fixed both in 6bbf4009f with regression tests.

Issue #4916, opened 2026-01-06
Type-safe extensible UiSchema


kolodny:
### Prerequisites

- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)

### What theme are you using?

core

### Is your feature request related to a problem? Please describe.

Currently, `UiSchema` is loosely typed. Any widget can be assigned to any field, and there's no autocomplete or validation for type-specific options like `minValue` (numbers) or `placeholder` (strings).

```tsx
// Current behavior - no type errors, no autocomplete
const uiSchema: UiSchema = {
  name: { 'ui:widget': 'RangeWidget' },  // Wrong! RangeWidget is for numbers
  age: { 'ui:placeholder': 'Enter age' }, // Wrong! placeholder is for strings
};
```

### Core Pattern: `{ when, then }`

[...]

Issue #4985, opened 2026-03-11
Enumerated Object labels are `[object Object]`


lemueldizon:
### Prerequisites

- [x] I have searched the existing issues
- [x] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [x] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [x] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

mantine

### Version

6.4.1

### Current Behavior

For the select input and checkboxes input it where the enums are objects, it doesn't apply the `ui:enumNames` properly and renders out `[object Object]`. This happens regardless of any theme.

### Expected Behavior

It should render the `ui:enumNames` as expected.

### Steps To Reproduce

It happens on the playground site you have as well so I'll use it as the example:
- Go to: https://rjsf-team.github.io/react-jsonschema-form/
- Then click on `Enumerated Objects`
- See that the location select and checkboxes input are rendered as `[object Object]`

### Environment

```markdown
- OS: 
- Node:
- npm:
```

x0k:
This is a breaking change in https://github.com/rjsf-team/react-jsonschema-form/pull/4961. Previously `enumNames` were taken from the array’s UI schema itself rather than from `items` @nickgros

nickgros:
Since the `enumNames` really does apply to the item schema, I think putting it in `items` is the right way to do it. Since we explicitly had this pattern in an example, then I think it's fair to consider this a breaking change and fix it (at least in RJSF v6)

I think we should do the following:

- Fix the example to properly put `enumNames` in `items`
- To avoid a breaking change, restore this case (probably just use the enumNames defined by the array property's uiSchema as a fallback)
- Decide whether we should deprecate this pattern and remove in a later major release

nickgros:
Fix proposed in https://github.com/rjsf-team/react-jsonschema-form/pull/4989

PR #5275, opened 2026-09-12
Fix 4916: add opt-in type-safe Checks parameter to UiSchema

heath-freenome:

Fixes #4916

`UiSchema<T>` already checks that a nested key names a real field of `T`, but it doesn't check the _values_ given to a field's `ui:widget`/`ui:field`/`ui:options` against that field's type - a widget that doesn't apply to the field it's on (`{ 'ui:widget': 'RangeWidget' }` on a `string` field), or a typo in a `ui:`-prefixed option name (`ui:wigdet`), only surfaces at runtime, if at all.

This PR closes that gap by giving `UiSchema` a fourth, opt-in type parameter: `UiSchema<T, S, F, Checks>`. Passing a `Checks` union narrows `ui:widget`, `ui:field`, and `ui:options` (and their `ui:`-prefixed equivalents, e.g. `ui:placeholder`) to only the values valid for each field's form-data type, recursing into nested objects/arrays the same way. This is purely additive and fully backward compatible - `UiSchema<T>` with no `Checks` behaves exactly as it always has.

`@rjsf/utils` has no widgets of its own, so it contributes only the `UiOptionsCheck<When, Then>` mechanism for building a `Checks` union, not a vocabulary. `@rjsf/core` exports `CoreUiOptionsChecks`, describing its own built-in widget/field/option vocabulary; themes and consumers extend it with their own `UiOptionsCheck` entries unioned in via `Checks`.

Because `FormProps['uiSchema']` stays typed as the open `UiSchema<T, S, F>` (no `Checks`) for backward compatibility, an inline `uiSchema` object literal passed directly as a JSX prop gets no narrowing from `Checks` on its own. Rather than add a wrapper component to force this, the PR documents using TypeScript's `satisfies` operator to check an inline literal against the closed form while leaving its own type in place for the prop.

Documents all of this at the end of uiSchema.md (cross-referenced from the v7 upgrade guide), including the known gaps (no fallback to unrestricted strings when `T` is `any`, and reduced narrowing on additionalProperties-shaped fields).

### Checklist


jimmycallin:
Thanks for this — the problem it targets is real, and the core observation in `withStrictUiSchema`'s doc comment is exactly right and not obvious: a `UiSchema | UiOptions` union catches nothing, because the permissive member absorbs every mistake. That diagnosis is the starting point for everything below.

I prototyped an alternative rather than argue from theory, and it changes what I think this PR should ship. This is long because the reasoning is worth having in the record; the concrete asks are in the last section. Every number below is measured on the current `v7` head.

---

## 1. Why `UiSchema` is permissive today, precisely

```ts
export type UiSchema<T = any, S, F> = GenericObjectType & …   // GenericObjectType = Record<string, any>
```

That one index signature does two unrelated jobs:

- **It lets a uiSchema nest by field name.** `{ firstName: { 'ui:widget': … } }` — `firstName` is an arbitrary string, and the only way a type admits arbitrary string keys is an index signature, *unless* it can enumerate the names, which means deriving them from `T`.
- **It keeps the `ui:` namespace open** for theme- and application-specific directives, which is a designed extension point.

It also swallows every typo — `ui:wigdet` is just another string key. I checked whether the `ui:` prefix can be carved back out of an index signature so typos could be caught while nesting stays open. Three attempts, none catch it:

```ts
type A = { 'ui:widget'?: string } & Record<string, any>;
type B = { [K in string as K extends `ui:${string}` ? never : K]?: unknown } & { 'ui:widget'?: string };
type C = { [k: `ui:${string}`]: unknown; [k: string]: unknown; 'ui:widget'?: string };

const a: A = { 'ui:wigdet': 'oops' };  // no error
const b: B = { 'ui:wigdet': 'oops' };  // no error
const c: C = { 'ui:wigdet': 'oops' };  // no error
```

TypeScript cannot express "any string key *except* those matching a template." So strictness of any kind requires removing the index signature — and then field-name nesting only works when `T` is known. **That is the real reason this feature has to be opt-in, and this PR is right about that.**

## 2. But field-name strictness *can* come from inference, on `UiSchema` itself

`FormProps<T>` already has `formData?: T`, so `T` is inferred today whenever typed form data is passed. The question is whether `UiSchema<T>` can be strict on field names when `T` is known and byte-for-byte permissive when it is `any` — with no new API.

It can, using the same technique that fixed `ErrorSchema` in #5278: **normalise the data type, then map.** Mapping over a normalised object type is never homomorphic over a primitive or array, and keeping the top level a plain mapped type (rather than a conditional) means it doesn't defer for an unresolved `T`, so `{}` stays assignable inside the library.

```ts
type UiSchemaChildData<T> =
  IsAnyType<T> extends true ? GenericObjectType                        // unknown data: every field name open
  : [NonNullable<T>] extends [readonly unknown[]] ? Record<never, never> // arrays nest via `items`, not by index
  : [NonNullable<T>] extends [object] ? UnionMembersMerged<NonNullable<T>>
  : Record<never, never>;                                              // leaves have no nested fields

/** Every key of every member of a union: a oneOf/anyOf field's uiSchema legitimately names keys from any branch */
type UnionMembersMerged<T> = {
  [K in T extends unknown ? keyof T : never]: T extends unknown ? (K extends keyof T ? T[K] : never) : never;
};

type UiSchemaChild<V, S, F> = IsAnyType<V> extends true ? any : UiSchema<V, S, F>;

export type UiSchema<T = any, S, F> = {
  [K in keyof UiSchemaChildData<T>]?: UiSchemaChild<UiSchemaChildData<T>[K], S, F>;
} & MakeUIType<UIOptionsBaseType<T, S, F>> & { /* the existing explicit ui: keys, items, ui:definitions… */ };
```

The `GenericObjectType &` at the front becomes a mapped type over the normalised data. Nothing else in `UiSchema` changes.

**Behaviour, measured against a probe:**

| Call | Result |
| --- | --- |
| `formData={typed}`: known field, nested field, array `items`, field behind an optional property | accepted |
| `formData={typed}`: `oneOf` field — common key, branch-only key, keys from both branches | accepted |
| `formData={typed}`: `ui:order` and a custom `ui:myTheme:thing` | accepted — `ui:` stays open |
| untyped form: arbitrary nesting, `whatever: 1`, even `ui:wigdet` | accepted — identical to today |
| bare `UiSchema` literal with no type argument | accepted — unchanged |
| `formData={typed}`: **unknown top-level field** | **rejected** |
| `formData={typed}`: **unknown nested field** | **rejected** |
| `formData={typed}`: **key in no branch of a `oneOf` union** | **rejected** |

Two properties of the result, checked directly: `string extends keyof UiSchema<Known>` is **false** (field names are closed), while `` `ui:${string}` extends keyof UiSchema<Known> `` is **true** (the directive namespace is open). The second comes for free: `UIOptionsBaseType` is string-keyed (`string extends keyof UIOptionsBaseType` holds today), so `MakeUIType`'s `ui:${…}` remap already yields a `ui:${string}` index signature. No new machinery is needed to keep themes' directives working.

**Cost, measured:** 19 errors across the whole monorepo, in four mechanical buckets:

1. **Five undeclared structural keys the index signature was hiding** — `uiSchema.additionalProperties`, `.additionalItems`, `.oneOf`, `.anyOf` (read by `ObjectField`, `ArrayField`, `optionsList`, `MultiSchemaField`), and the legacy unprefixed `classNames` that `SchemaField` strips per #439. These are real, documented features that only type-checked by accident. Declaring them next to `items` is a documentation improvement the strict type forces.
2. **Three `uiSchema = {}` destructure defaults** (`Form.tsx`, `ObjectField.tsx`) that now infer `{}`; annotate as `UiSchema<T, S, F>`.
3. **Eight runtime-key traversal sites** (`getUiOptions`, `optionsList`, `LayoutGridField`, and `SchemaField`'s `Omit`-of-a-generic around `ui:options`) — the boundary pattern `toErrorList` and `unwrapErrorHandler` already use: view as `Record<string, unknown>` where you walk by runtime key.
4. **Three theme one-liners** — daisyui's `interface … extends UiSchema` must become `type … = UiSchema & …` (TS forbids `extends` on a type with an index signature); antd and mantine `WrapIfAdditionalTemplate` build a uiSchema via spread and need the intermediate typed.

Breaking only for typed users whose uiSchema names fields absent from `T` — `dependencies`-added properties, or `additionalProperties` keys when `T` has no index signature — so it belongs in the v7 migration guide. `oneOf`/`anyOf` is handled. Untyped users are untouched.

Three things I checked so they don't surprise anyone later:

- A uiSchema returned from a helper with an *explicitly declared interface* type is rejected by this design. **It is rejected by today's `UiSchema` too** — the `ui:${string}` signature above already requires an index signature interfaces don't have. Pre-existing, not a regression; worth a docs note.
- The single `any` in `UiSchemaChild` is deliberate and measured: `unknown` there costs 39 errors vs 19 and forces every untyped user to narrow before touching a nested entry; a recursive open-object type costs 751 and rejects `whatever: 1`. `any` in → `any` out is the faithful encoding of "the caller passed `any`", the same conclusion as #5278.
- `UnionMembersMerged` is non-homomorphic, so it drops `?` from the child *data* type. That's harmless: only its keys and values are consumed, and optionality is re-applied on the uiSchema side — the optional-property probe passes.

## 3. What this means for this PR

Once field names are checked by inference, the *only* thing left to opt into is what this PR's `UiOptions` uniquely adds: restricting `ui:widget`/`ui:field` to a declared vocabulary and closing the `ui:` namespace. That tier genuinely must be explicit — you cannot check a vocabulary nobody has declared — and this PR's `UiOptionsCheck` / `Checks` mechanism is the right shape for declaring it.

But three consequences follow:

**The HOC becomes unnecessary.** `withStrictUiSchema` renders `FormComponent` with the props it received — its own comment says it "does no runtime validation or transformation". It exists to bind `T` once. With `T` inferred from `formData`, the binding already happens, and the remaining opt-in is a per-literal declaration that `satisfies` expresses directly on TypeScript 7 (which this repo already uses in nine places):

```ts
const ui = { … } satisfies UiSchema<MyData, RJSFSchema, any, MyThemeChecks>;
```

No wrapper element, no `ref` plumbing, no `StrictUiSchemaFormProps`, no new module.

**The vocabulary tier should be a type parameter on `UiSchema`, not a second type.** `UiOptions` collides with the existing `UIOptionsType` (the resolved `ui:options` bag) — they differ by casing and a suffix. Rather than rename it, fold it in:

```ts
UiSchema<MyData>                              // field names checked; ui: open  (from §2)
UiSchema<MyData, RJSFSchema, any, MyChecks>   // + vocabulary checked; ui: closed (this PR's machinery)
```

`Checks = never` preserves today's behaviour, and a trailing defaulted type parameter is non-breaking for every existing `UiSchema<T, S, F>`. One type, one name, opt-in by type argument. I have **not** prototyped this half; the thing to verify first is that gating the namespace on `Checks` — a conditional — only defers for code that is *generic over* `Checks`. Internals aren't (they use `UiSchema<T, S, F>` with `Checks` defaulted), so it should resolve everywhere that matters.

**`CoreUiOptionsChecks` belongs in `@rjsf/core`.** It hand-codes `'TextWidget' | 'TextareaWidget' | …`, a duplicate of what `generateWidgets()` registers, and lives in `@rjsf/utils`, which knows nothing about core's widgets. Nothing makes anyone update it when a widget is added. The mechanism belongs in `utils`; core's vocabulary belongs next to core's widgets — which is also the pattern this PR already prescribes for themes.

## 4. Recommendation

1. **Land field-name strictness on `UiSchema<T>` itself, as its own PR** — the §2 change plus the 19 fixes. I have it prototyped and can open it.
2. **In this PR, keep the `UiOptionsCheck`/`Checks` machinery and the type tests; drop `withStrictUiSchema` and `StrictUiSchemaFormProps`.** Document `satisfies` as the way to apply a vocabulary.
3. **Expose the vocabulary as `UiSchema`'s fourth type parameter** instead of a separate `UiOptions` type, once (1) is in.
4. **Move `CoreUiOptionsChecks` to `@rjsf/core`.**

jimmycallin:
Opened #5279 for recommendation 1 — field-name strictness on `UiSchema<T>` itself, the §2 type change plus the 19 fixes it forces. It doesn't touch anything in this PR; the two are independent until the vocabulary tier moves onto `UiSchema`'s fourth type parameter, which needs #5279 first.

packages/docs/docs/api-reference/uiSchema.md
```diff
@@ -956,3 +956,74 @@ const uiSchema: UiSchema = {
 - [AntD Customization](themes/antd/uiSchema.md)
 - [Chakra-UI Customization](themes/chakra-ui/uiSchema.md)
 - [MUI Customization](themes/mui/uiSchema.md)
+
+## Type-safe UiSchema
+
+`UiSchema` is deliberately loose: it blends in `Record<string, any>`, so any key and any value type-check, including a widget name that doesn't apply to the field it's on (`{ 'ui:widget': 'RangeWidget' }` on a `string` field) or a typo in a `ui:`-prefixed option name (`ui:wigdet`). Those mistakes only surface at runtime, if at all.
+
+For consumers who type their form data with an explicit TypeScript type, `@rjsf/utils` also exports `UiOptions<T, S, F, Checks>`, an opt-in, stricter alternative. `UiOptions` narrows `ui:widget`, `ui:field` and `ui:options` (and their `ui:`-prefixed equivalents, e.g. `ui:placeholder`) to only the values valid for each field's form-data type, and recurses into nested objects/arrays typing each field the same way:
```
nickgros, line 964: Attempt to improve clarity

```suggestion
`UiSchema` is loosely typed. It uses `Record<string, any>`, so any key and any value are valid types, 
including a widget name that may not support the field it's on (`{ 'ui:widget': 'RangeWidget' }` on a `string` field), or a typo in a `ui:`-prefixed option name (`ui:wigdet`). Those mistakes may only surface at runtime.

For consumers who type their form data with an explicit TypeScript type, `@rjsf/utils` also exports `UiOptions<T, S, F, Checks>`, an opt-in, stricter alternative. `UiOptions` narrows `ui:widget`, `ui:field` and `ui:options` (and their `ui:`-prefixed equivalents, e.g. `ui:placeholder`) to only the values valid for each field's form-data type, and recurses into nested objects/arrays typing each field the same way:
```

packages/docs/docs/api-reference/uiSchema.md
```diff
@@ -956,3 +956,74 @@ const uiSchema: UiSchema = {
...
+const uiSchema: UiOptions<FormData> = {
+  age: { 'ui:widget': 'RangeWidget' }, // ok - RangeWidget is valid for `number`
+  bio: {
+    // @ts-expect-error RangeWidget is not valid for a `string` field
+    'ui:widget': 'RangeWidget',
+  },
+};
+```
+
+This is purely additive - `UiSchema` is unchanged, and adopting `UiOptions` is entirely opt-in, incrementally, field by field or form by form.
```
nickgros, line 983: Attempt to improve clarity

```suggestion
Using the `UiOptions` type is not mandatory when providing `uiSchema` in TypeScript; adopting it is entirely opt-in, and can be adopted incrementally, field by field or form by form.
```

packages/docs/docs/api-reference/uiSchema.md
```diff
@@ -956,3 +956,74 @@ const uiSchema: UiSchema = {
...
+
+### Extending the widget/option vocabulary
+
+The widget/field names and options that `UiOptions` recognizes out of the box come from `CoreUiOptionsChecks`, a union of `UiOptionsCheck<When, Then>` rules: "when a field's type is assignable to `When`, the names/options in `Then` become valid for it." Themes and consumers extend that vocabulary with their own `Checks` union, passed as `UiOptions`'s fourth type parameter, rather than by modifying `CoreUiOptionsChecks` itself:
```
nickgros, line 987: Just adding keyword `augment`, since type augmentation is the typical way for libraries to support overriding these kinds types if they don't or can't add generic support.


```suggestion
The widget/field names and options that `UiOptions` recognizes out of the box come from `CoreUiOptionsChecks`, a union of `UiOptionsCheck<When, Then>` rules: "when a field's type is assignable to `When`, the names/options in `Then` become valid for it." Themes and consumers extend that vocabulary with their own `Checks` union, passed as `UiOptions`'s fourth type parameter, rather than by modifying or augmenting `CoreUiOptionsChecks` itself:
```

packages/docs/docs/api-reference/uiSchema.md
```diff
@@ -956,3 +956,74 @@ const uiSchema: UiSchema = {
...
+const uiSchema: MyUiOptions<{ active: boolean }> = {
+  active: { 'ui:widget': 'ToggleWidget' },
+};
+```
+
+### Known gaps
+
+- With no type parameter (`T` defaulting to `any`), `UiOptions` does **not** fall back to `UiSchema`-style unrestricted strings for `ui:widget`/`ui:field` - they're still limited to the names declared in `CoreUiOptionsChecks`/`Checks`. Pass an actual widget/field component instance instead of a string, or extend `Checks`, for anything not already covered.
+- `UiOptions` does not yet support `ui:globalOptions`, `ui:fieldReplacesAnyOrOneOf` or `ui:definitions`; use `UiSchema` for uiSchema fragments that need those.
+- A field whose form-data shape includes an index signature (e.g. from `additionalProperties`/`patternProperties`) only gets type-checking for its explicitly-declared keys; dynamic keys still type-check but without narrowing beyond the index signature's value type.
```
nickgros, line 1004: Use of "type-check" as an intransitive verb is unusual and is an LLM smell

```suggestion
- A field whose form-data shape includes an index signature (e.g. from `additionalProperties`/`patternProperties`) only gets type-checking for its explicitly-declared keys; dynamic keys are type-checked but without narrowing beyond the index signature's value type.
```

packages/docs/docs/api-reference/uiSchema.md
```diff
@@ -956,3 +956,74 @@ const uiSchema: UiSchema = {
...
+
+const StrictForm = withStrictUiSchema<FormData>(Form);
+
+// Compile error: RangeWidget is not valid for `bio`, a string field.
+<StrictForm schema={schema} uiSchema={{ bio: { 'ui:widget': 'RangeWidget' } }} validator={validator} />;
+```
+
+A plain `UiSchema<T,S,F> | UiOptions<T,S,F,Checks>` union on `uiSchema` would not achieve this: as long as `UiSchema`'s permissive `Record<string, any>` remains an accepted alternative on the same prop, TypeScript only needs the literal to match _one_ member of the union, and it always matches the permissive one. `withStrictUiSchema` only exposes `UiOptions`, which is what actually closes the gap.
+
+`UiOptions` has no runtime representation - `withStrictUiSchema` does no runtime validation or transformation of its own. It renders the wrapped component with the exact same props it was given; the extra safety exists only at compile time, for code that renders through the wrapper instead of `Form` directly.
```
nickgros, line 1029: I think this negative example is unnecessary. And deleted the two paragraphs that I believe are just explaining basic TypeScript behavior.

```suggestion
Declaring an intermediate `const uiSchema: UiOptions<FormData> = {...}` and passing that to `Form`'s `uiSchema` prop gets you the narrowing above, because `FormProps['uiSchema']` is typed as `UiSchema<T, S, F>`, which is permissive enough to accept a well-formed `UiOptions` value with no cast. But that also means an inline `uiSchema` object literal passed directly as a JSX prop gets **no narrowing at all**.

`@rjsf/core` exports `withStrictUiSchema`, a Higher-Order Component that wraps `Form` (or a themed `Form` returned by `withTheme()`) and returns a version of it whose `uiSchema` prop only accepts `UiOptions<T, S, F, Checks>`, with no permissive `UiSchema` fallback for a mistake to slip through:

```tsx
import { withStrictUiSchema } from '@rjsf/core';

const StrictForm = withStrictUiSchema<FormData>(Form);

// Compile error: RangeWidget is not valid for `bio`, a string field.
<StrictForm schema={schema} uiSchema={{ bio: { 'ui:widget': 'RangeWidget' } }} validator={validator} />;
```
```

File: packages/core/src/withStrictUiSchema.tsx [hunk 6]
```diff
@@ -0,0 +1,50 @@
...
+ * fails `UiOptions`'s stricter check simply matches the permissive member instead, with no error. Only exposing
+ * `UiOptions` - with no permissive fallback for a mistake to slip through - actually catches it, which is what this
+ * wrapper is for.
+ *
+ * `UiOptions` is a compile-time-only construct with no runtime representation, so this wrapper does no runtime
+ * validation or transformation of its own - it renders `FormComponent` with the exact same props it was given, and
+ * the extra type safety exists only at compile time, for consumers that render through this wrapper instead of
+ * `FormComponent` directly.
+ */
+export default function withStrictUiSchema<
```
nickgros, line 35: Do we really need this if someone can just use `satisfies`? (@jimmycallin also called this out)

heath-freenome:
Thanks both — this is now rebased on `v7` with #5279 in it, and I've acted on the parts of the feedback that were already clear-cut. Status against Jimmy's four recommendations:

1. **Field-name strictness on `UiSchema<T>` itself** — merged as #5279. Thanks for prototyping and shipping this; it's a real improvement independent of everything else here.
2. **Drop `withStrictUiSchema`/`StrictUiSchemaFormProps`** — done (7a867c8aa, 1d1b1e1c1). @nickgros, this directly answers your question on the HOC file too — agreed, it wasn't earning its keep once an inline literal can be checked with `satisfies` directly. I verified the `satisfies` pattern actually catches the same mistake the HOC did before documenting it as the replacement in uiSchema.md.
3. **Fold `UiOptions`/`Checks` into `UiSchema`'s 4th type parameter** — not started yet. Jimmy, you flagged this is unprototyped and named a specific thing to verify first: that gating the `ui:` namespace on a `Checks` conditional only defers for code generic over `Checks`, not for internals using `UiSchema<T,S,F>` with `Checks` defaulted. I want to spike that before touching `UiSchema` itself, same way #5279 was de-risked before landing.
4. **Move `CoreUiOptionsChecks` to `@rjsf/core`** — not started; makes most sense to do once (3)'s shape is settled, since the type may change form when it becomes a `UiSchema` type param instead of a `UiOptions` one.

@nickgros — I also folded in your wording suggestions on the surviving prose (the "not mandatory" phrasing, "modifying or augmenting", and the "type-checked" fix). One thing worth flagging: your suggested edit to the opening paragraph was accurate when you wrote it, but #5279 landed in between, so I updated it further rather than take the suggestion as-is — it now says field *names* are checked by `UiSchema<T>` itself, and `UiOptions` is what's still needed for widget/option *values*.

Will follow up here once I've spiked (3).

heath-freenome:
All four of Jimmy's recommendations are now done on this branch (not pushed yet — @heath-freenome is pushing separately):

3. **Fold `UiOptions`/`Checks` into `UiSchema`'s fourth type parameter** (ea2a0ee98). `UiSchema<T, S, F, Checks = never>` — `Checks` defaulting to `never` preserves today's fully permissive behavior exactly; a concrete `Checks` union narrows `ui:widget`/`ui:field`/`ui:options` and closes the `ui:` namespace. The separate `UiOptions` type is gone.

   Before implementing, I verified the specific risk you flagged (that you hadn't prototyped this half): built the mechanism as a temporary edit, ran the full monorepo typecheck (zero new errors — every internal usage uses the 3-arg form with `Checks` defaulted, exactly as predicted), then wrote targeted tests for the "generic over `Checks`" deferral scenario specifically. It resolves correctly at each call site.

   One thing the spike didn't catch that the real implementation needed to handle: leaving `'ui:options'` as an unconditional sibling (rather than folding it into the `Checks`-gated part) would have let its own index signature leak permissive extra keys into the "closed" case. Caught and fixed that before shipping it.

   Nice side effect: three of `UiOptions`'s documented "known gaps" (`ui:globalOptions`, `ui:fieldReplacesAnyOrOneOf`, `ui:definitions`) turned out to already be resolved — those were always unconditional siblings on `UiSchema`, so once `UiOptions` became `UiSchema` itself, they were never actually missing.

4. **Move `CoreUiOptionsChecks` to `@rjsf/core`** (e0e4417e6). `@rjsf/utils`'s `UiSchema` no longer auto-includes any vocabulary — `UiVocabularyPart` uses `Checks` alone now, not `CoreUiOptionsChecks | Checks`, so passing `@rjsf/core`'s vocabulary (or a theme's own) is entirely on the caller, unioned in explicitly. `UiOptionsCheck` (the generic `{when,then}` mechanism) stayed in `utils`, since that's what any theme builds its own vocabulary from.

   This also meant splitting the type tests along the dependency line: `packages/utils/test/uiSchemaTypes.test.ts` now exercises the generic mechanism with a made-up example vocabulary instead of importing `@rjsf/core`, and the scenarios that specifically validated real widget names moved to a new `packages/core/test/CoreUiOptionsChecks.test.ts`.

2. **Drop `withStrictUiSchema`** (7a867c8aa, 1d1b1e1c1) — done earlier, documented `satisfies` as the replacement, verified it actually catches the same mistake before writing it up. @nickgros, this is the direct answer to your question on that file.

1. **Field-name strictness on `UiSchema<T>`** — your #5279, merged.

Also folded in your wording suggestions on the docs, @nickgros, and corrected a paragraph that #5279 made stale in the meantime (`UiSchema<T>` now checks field names itself; `Checks` is what's still needed for widget/option values).

Ready for another look whenever convenient.

packages/docs/docs/api-reference/uiSchema.md
heath-freenome: This suggestion's target text (the `Record<string, any>`/`UiOptions` opening) is stale now - #5279 landed after you wrote this and closed field-name checking on `UiSchema<T>` itself, and `UiOptions` later folded into `UiSchema`'s fourth `Checks` parameter. The opening paragraph has been rewritten to describe that current shape (see the current uiSchema.md, "Type-safe UiSchema" section).

packages/docs/docs/api-reference/uiSchema.md
heath-freenome: Folded in - the "not mandatory... opt-in... incrementally" phrasing is in the current "Passing `Checks` is entirely opt-in..." sentence, updated for `UiSchema`'s `Checks` param instead of the old `UiOptions` type.

packages/docs/docs/api-reference/uiSchema.md
heath-freenome: Good catch, and this is now even more relevant since `CoreUiOptionsChecks` moved to `@rjsf/core`: extending the vocabulary should union in your own `UiOptionsCheck` entries rather than touch that export directly. Just pushed this to the "Extending the widget/option vocabulary" section, adapted to the current shape (341bdfbd6).

packages/docs/docs/api-reference/uiSchema.md
heath-freenome: Fixed - that exact sentence is now in the "Known gaps" section verbatim.

packages/docs/docs/api-reference/uiSchema.md
heath-freenome: Agreed on cutting the negative example - just did in 341bdfbd6. The rest of this suggestion (the `withStrictUiSchema` code block) no longer applies since the HOC was dropped in favor of `satisfies` (see below and the PR description) - that section now shows the `satisfies` pattern directly instead.

packages/core/src/withStrictUiSchema.tsx
heath-freenome: Agreed - dropped it. Verified `satisfies` catches the same mistake before documenting it as the replacement; see uiSchema.md's "Applying it to an inline uiSchema with satisfies" section.

packages/utils/src/types.ts
```diff
@@ -1258,6 +1258,94 @@ export type UIOptionsType<
...
+    widget?: WidgetsFor<T, Checks> | Widget<T, S, F>;
+    field?: FieldsFor<T, Checks> | Field<T, S, F>;
+  } & RawOptsFor<T, Checks>;
+}
+
+/** @internal Common `ui:*` options valid on any field regardless of its type, kept separate from a `Checks` union
+ * since they aren't type-specific vocabulary - available whenever `Checks` narrows the vocabulary, same as they
+ * always are through `UIOptionsBaseType` when it doesn't.
+ */
+interface ClosedCommonUiOptions {
```
jimmycallin: **`ClosedCommonUiOptions` drops the type-agnostic vocabulary that `UIOptionsBaseType` supplies in the open branch.**

The `[Checks] extends [never]` branch is `MakeUIType<UIOptionsBaseType<T, S, F>>`, which includes (a) the 14 per-field template overrides picked from `TemplatesType` (`ui:FieldTemplate`, `ui:ObjectFieldTemplate`, `ui:ArrayFieldTemplate`, `ui:BaseInputTemplate`, `ui:TitleFieldTemplate`, …) and (b) all of `GlobalUISchemaOptions` (`ui:label`, `ui:enableMarkdownInDescription`, `ui:enableMarkdownInHelp`, `ui:allowClearTextInputs`, `ui:duplicateKeySuffixSeparator`, `ui:enableOptionalDataFieldForType`, …). The closed branch replaces that with `ClosedCommonUiOptions`, which lists none of them, and none of them can be recovered through `Checks` because they aren't form-data-type-specific.

Verified against this branch (`tsc --build packages/core/test/tsconfig.json`), with `CoreUiOptionsChecks` passed:

```
error TS2353: ... ''ui:label'' does not exist in type 'UiSchema<string, any, any, CoreUiOptionsChecks>'
error TS2353: ... ''ui:ObjectFieldTemplate'' does not exist in type 'UiSchema<{ a: string; }, any, any, CoreUiOptionsChecks>'
error TS2353: ... ''ui:enableMarkdownInDescription'' does not exist in type 'UiSchema<string, any, any, CoreUiOptionsChecks>'
```

`ui:label: false` and per-field template overrides are among the most-used uiSchema keys in the docs, so opting into `Checks` breaks most real uiSchemas. `ClosedCommonUiOptions` should carry the template `Pick<TemplatesType…>` and `GlobalUISchemaOptions` (minus its `GenericObjectType` index signature, which is what makes the open form permissive) rather than a hand-picked subset.

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * This is intentionally a starting set covering the options that already exist on `UIOptionsBaseType` - it does not
+ * attempt to model every widget/option combination (for example, per-widget options like `RangeWidget`'s min/max
+ * come from the JSON Schema itself, not from `ui:options`, so they are not repeated here).
+ */
+export type CoreUiOptionsChecks =
+  | UiOptionsCheck<
+      string,
+      {
+        widget?:
```
jimmycallin, line 41: **The widget vocabulary lists only registered component names and omits every documented alias, which is the spelling the docs and most user code actually use.**

`getWidget`'s `widgetMap` (packages/utils/src/getWidget.tsx) is what resolves a `ui:widget` string, and it accepts the lowercase aliases per schema type: `text`, `textarea`, `password`, `email`, `uri`, `data-url`, `radio`, `select`, `hidden`, `date`, `datetime`, `date-time`, `iso-date-time`, `alt-date`, `alt-datetime`, `time`, `iso-time`, `color`, `file`, `hostname`, `ipv4`, `ipv6` (string); `text`/`select`/`updown`/`range`/`radio`/`hidden` (number, integer); `checkbox`/`radio`/`select`/`hidden` (boolean); `select`/`checkboxes`/`files`/`hidden` (array). None of them appear here.

Verified on this branch — every one of these idiomatic, documented forms is now a compile error under `CoreUiOptionsChecks`:

```
error TS2322: Type '"textarea"' is not assignable to type '"AltDateTimeWidget" | ... | undefined'.
error TS2322: Type '"hidden"'   is not assignable to type '"AltDateTimeWidget" | ... | undefined'.
error TS2322: Type '"radio"'    is not assignable to type '"CheckboxWidget" | "HiddenWidget" | "RadioWidget" | "SelectWidget" | ...'.
error TS2322: Type '"range"'    is not assignable to type '"HiddenWidget" | "RadioWidget" | "RangeWidget" | ...'.
```

The aliases need to be in each `widget?:` union alongside the PascalCase names; ideally derived from `widgetMap` so the two can't drift.

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,88 @@
...
+      null,
+      {
+        widget?: 'HiddenWidget';
+        field?: 'NullField';
+      }
+    >
+  | UiOptionsCheck<
+      unknown[],
+      {
+        widget?: 'CheckboxesWidget' | 'SelectWidget' | 'FileWidget';
```
jimmycallin: **`HiddenWidget` is missing from the array vocabulary, and object fields get no `widget` names at all — so hiding an array or object field becomes untypable.**

`widgetMap.array` includes `hidden: 'HiddenWidget'`, and `SchemaField` hides *any* field (object and array included) via `uiOptions.widget === 'hidden'` (packages/core/src/components/fields/SchemaField.tsx:260), before schema type is consulted. Verified on this branch:

```
error TS2322: Type '"HiddenWidget"' is not assignable to type
  '"CheckboxesWidget" | "FileWidget" | "SelectWidget" | Widget<string[], any, any> | undefined'.   // array field
error TS2322: Type 'string' is not assignable to type 'Widget<{ a: string; }, any, any> | undefined'. // object field: widget union is `never`
```

So `{ myArray: { 'ui:widget': 'hidden' } }` and `{ myObject: { 'ui:widget': 'hidden' } }` — both working today — stop compiling once `Checks` is passed. Add `HiddenWidget`/`hidden` to the array entry and give the object entry a `widget?: 'HiddenWidget' | 'hidden'`.

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,88 @@
...
+          | 'RadioWidget'
+          | 'FileWidget'
+          | 'HiddenWidget';
+        field?: 'StringField';
+        placeholder?: string;
+        rows?: number;
+        inputType?: string;
+        autocomplete?: HTMLInputElement['autocomplete'];
+        autocapitalize?: HTMLInputElement['autocapitalize'];
+        emptyValue?: string;
```
jimmycallin, line 64: **Several options are scoped to one type here but are genuinely valid on others, so closing the namespace turns working uiSchemas into compile errors.**

Verified on this branch under `CoreUiOptionsChecks`:

```
error TS2353: ''ui:emptyValue''  does not exist in type 'UiSchema<number,  any, any, CoreUiOptionsChecks>'
error TS2353: ''ui:placeholder'' does not exist in type 'UiSchema<number,  any, any, CoreUiOptionsChecks>'
error TS2353: ''ui:inline''      does not exist in type 'UiSchema<boolean, any, any, CoreUiOptionsChecks>'
```

- `emptyValue` is declared `string`-only, but `NumberField`/`BaseInputTemplate` apply it to numeric inputs too (and `UIOptionsBaseType` types it `any`).
- `placeholder` is `string`-only, but `SelectWidget` reads it for number/boolean enums.
- `inline` is array-only, but it is `RadioWidget`/`CheckboxesWidget`'s option — it applies to boolean, number and string enum fields.
- `filePreview` is `string`-only, but it is read for `ui:widget: files` array fields as well.

Because the closed branch removes `UIOptionsBaseType` entirely, anything missed here is unrecoverable by the consumer short of unioning in extra `UiOptionsCheck`s to patch the library's own vocabulary.

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,88 @@
...
+        widget?: 'CheckboxesWidget' | 'SelectWidget' | 'FileWidget';
+        field?: 'ArrayField';
+        addable?: boolean;
+        orderable?: boolean;
+        removable?: boolean;
+        copyable?: boolean;
+        inline?: boolean;
+      }
+    >
+  | UiOptionsCheck<
```
jimmycallin, line 138: **`object & { length?: never }` is a structural matcher that silently misses real shapes, and the miss is invisible because `RawOptsFor` degrades to `unknown` rather than erroring.**

`readonly T[]` is not assignable to `unknown[]`, and its `length: number` is not assignable to `length?: never`, so a `readonly string[]` field matches *no* check in the union. `UnionToIntersection<never>` resolves to `unknown`, so `ui:options` quietly degrades to just `{ widget?: Widget; field?: Field }` — no error pointing at the cause, just every option gone:

```
error TS2353: Object literal may only specify known properties, and 'addable' does not exist in type
  '{ widget?: Widget<readonly string[], any, any>; field?: Field<readonly string[], any, any>; }'
```

The same hole catches any form-data object that happens to declare a `length` property (`interface Track { length: number; title: string }`), which then loses `ui:order` and `ui:field`. Matching arrays with `readonly unknown[]` (which covers mutable arrays too) and excluding arrays from the object matcher with something other than `length` would close both.

packages/utils/src/types.ts
```diff
@@ -1258,6 +1258,94 @@ export type UIOptionsType<
...
+  then: Then;
+}
+
+/** @internal Distributes over a `Checks` union, collapsing each member into an intersection of its `then` type
+ * (minus `widget`/`field`) for every member whose `when` the field type `T` is assignable to.
+ */
+type UnionToIntersection<U> = (U extends unknown ? (x: U) => void : never) extends (x: infer I) => void ? I : never;
+
+/** @internal The union of `then.widget` names valid for form-data type `T`, drawn from `Checks`. */
+type WidgetsFor<T, Checks> =
```
jimmycallin, line 1323: **`Then extends { widget?: infer W }` matches even when `Then` has no `widget` key, so a check written without one silently widens the vocabulary instead of contributing nothing.**

`UiOptionsCheck`'s `Then` defaults to `GenericObjectType` (`{ [k: string]: any }`), so a consumer writing `UiOptionsCheck<string>` — a perfectly natural way to add a couple of `ui:options` without touching widgets — infers `W = any`, and `'ui:widget'?: any | Widget<…>` re-opens `ui:widget` for every `string` field in the whole uiSchema, with no diagnostic. `FieldsFor` has the same shape.

Guarding on the key's presence (`Then extends { widget?: infer W } ? (unknown extends W ? never : W) : never`, or requiring `'widget' extends keyof Then`) keeps a `widget`-less check inert.

packages/utils/src/types.ts
```diff
@@ -1258,6 +1258,94 @@ export type UIOptionsType<
...
+ * exports it for its users to pass as `Checks` (e.g. `@rjsf/core`'s `CoreUiOptionsChecks`), and a consumer of a
+ * theme adds their own domain-specific options the same way, unioned into what they pass.
+ */
+export interface UiOptionsCheck<When = any, Then = GenericObjectType> {
+  when: When;
+  then: Then;
+}
+
+/** @internal Distributes over a `Checks` union, collapsing each member into an intersection of its `then` type
+ * (minus `widget`/`field`) for every member whose `when` the field type `T` is assignable to.
```
jimmycallin: Nit: this doc comment describes `RawOptsFor` (distributing over `Checks`, collapsing `then` minus `widget`/`field` into an intersection), not `UnionToIntersection`, which is a generic union-to-intersection helper with no knowledge of `Checks`. The comment landed on the wrong declaration.

packages/utils/src/types.ts
heath-freenome: Confirmed and fixed in cd5dd58b0. Extracted the template-override picks into a shared `UIOptionsTemplateOverrides<T,S,F>` (used by both `UIOptionsBaseType` and the closed vocabulary) and split `GlobalUISchemaOptions` into `GlobalUISchemaOptionsKeys` (its declared keys, no index signature) + the existing `GenericObjectType &`-wrapped export. `ClosedCommonUiOptions` is now the generic `CommonUiOptions<T,S,F>`, carrying both, so `ui:label`, `ui:ObjectFieldTemplate`, `ui:enableMarkdownInDescription`, etc. are all available under `Checks` again. Added a regression test asserting these compile under `CoreUiOptionsChecks`.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in cd5dd58b0. Made `getWidget.tsx`'s `widgetMap` an exported `as const`, and added `WidgetAliasFor<Type>` (`keyof (typeof widgetMap)[Type]`) derived directly from it. Each `CoreUiOptionsChecks` branch now unions in `WidgetAliasFor<'string'>`/`'number'`/`'boolean'`/`'array'` alongside the PascalCase names, so it can't drift from what `getWidget` actually resolves. Added a regression test covering the `textarea`/`hidden` aliases.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in cd5dd58b0. Added `HiddenWidget` (plus the `hidden` alias, via `WidgetAliasFor<'array'>`) to the array branch, and gave the object branch a `widget?: 'HiddenWidget' | 'hidden'` it previously lacked entirely. Added a regression test for hiding both an array and an object field.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in cd5dd58b0. Added `placeholder`/`emptyValue` to the number branch, `placeholder`/`inline` to the boolean branch, `inline` to the string branch, and `filePreview` to the array branch, matching what `BaseInputTemplate`/`SelectWidget`/`RadioWidget`/`CheckboxesWidget`/`FileWidget` actually read. Added a regression test asserting these compile for the right types.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in cd5dd58b0, using the exact fix you suggested. Changed the array `when` to `readonly unknown[]` and the object exclusion from `{ length?: never }` to `{ [Symbol.iterator]?: never }`. Before writing the fix I verified both failure modes and the corrected behavior with standalone `tsc` type-level assertions: a `readonly string[]` now correctly matches only the array branch, and a `{ length: number; title: string }`-shaped interface is no longer wrongly excluded from the object branch. Added a regression test covering both cases.

packages/utils/src/types.ts
heath-freenome: Confirmed and fixed in cd5dd58b0. Added a `DeclaresKey<Then, K>` helper (`string extends keyof Then ? false : K extends keyof Then ? true : false`) and gated both `WidgetsFor` and `FieldsFor` on it before doing the `infer`. Verified empirically with a standalone `tsc` probe first: `UiOptionsCheck<string>` (bare, `Then` defaulting to `GenericObjectType`) really did widen `WidgetsFor` to `any` before the fix, resolves to `never` after; a `Then` with unrelated explicit keys and no index signature was already inert either way (so this isn't a behavior change there); and a real, well-formed `widget` union is untouched. Added a regression test unioning in a widget-less check and asserting it doesn't reopen `ui:widget`.

packages/utils/src/types.ts
heath-freenome: Good catch, fixed in cd5dd58b0 - moved the doc comment down to describe `RawOptsFor`, and gave `UnionToIntersection` its own one-liner noting it's a generic helper with no knowledge of `Checks`.

packages/utils/src/types.ts
```diff
@@ -1258,6 +1270,114 @@ export type UIOptionsType<
...
+    : never
+>;
+
+/** @internal The `ui:widget`/`ui:field`/`ui:options` shape for a single field of form-data type `T`, narrowed by
+ * `Checks` and still allowing a one-off `Widget`/`Field` component instance instead of a registered name.
+ */
+interface UiOptionsComponentPart<T, S extends StrictRJSFSchema, F extends FormContextType, Checks> {
+  'ui:widget'?: WidgetsFor<T, Checks> | Widget<T, S, F>;
+  'ui:field'?: FieldsFor<T, Checks> | Field<T, S, F>;
+  'ui:options'?: {
```
jimmycallin, line 1363: **Closed mode drops every common option from `ui:options`.**

`ui:options` here is only `{ widget, field } & RawOptsFor<T, Checks>` — `CommonUiOptions` is applied solely through the `ui:`-prefixed form (`MakeUIType<CommonUiOptions>` in `UiVocabularyPart`). So the grouped spelling of the most common options stops compiling as soon as `Checks` is passed, while the prefixed spelling of the very same option is fine:

```ts
type C<T> = UiSchema<T, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
const a: C<{ bio: string }> = { bio: { 'ui:options': { label: false } } };      // error
const b: C<{ bio: string }> = { bio: { 'ui:options': { title: 'Bio' } } };      // error
const c: C<{ bio: string }> = { bio: { 'ui:options': { classNames: 'x', disabled: true } } }; // error
const d: C<{ bio: string }> = { bio: { 'ui:options': { help: 'h' } } };         // error
const e: C<{ bio: string }> = { bio: { 'ui:options': { FieldTemplate: () => null } } }; // error
```

(verified with `tsc` against this branch). `ui:options: { label: false }` is one of the most-used forms in the docs, so any adopter of `Checks` has to rewrite it. Intersecting `CommonUiOptions<T, S, F>` into the `ui:options` shape fixes all of these at once.

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,114 @@
...
+      object & { [Symbol.iterator]?: never },
+      {
+        // Only the literal `'hidden'` has any runtime effect here: `ObjectField`/`SchemaField` decide to hide an
+        // object field via `uiOptions.widget === 'hidden'`, a strict comparison to that exact alias - unlike
+        // string/number/boolean/array fields, an object field never resolves `ui:widget` through `getWidget`, so
+        // `'HiddenWidget'` (the PascalCase component name) does nothing for it and is deliberately not offered here.
+        widget?: 'hidden';
+        field?: 'ObjectField';
```
jimmycallin, line 146: **`field?: 'ObjectField'` makes the layout grid (and the multi-schema fields) uncompilable under `Checks`.**

`LayoutGridField` is used exactly this way in `docs/usage/layout-grid.md`, and both halves of it are rejected:

```ts
type C<T> = UiSchema<T, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
const g1: C<{ person: { first: string } }> = { person: { 'ui:field': 'LayoutGridField' } };
// error: Type '"LayoutGridField"' is not assignable to type '"ObjectField" | Field<...> | undefined'
const g2: C<{ person: { first: string } }> = { person: { 'ui:layoutGrid': { 'ui:row': ['first'] } } };
// error: 'ui:layoutGrid' does not exist in type 'UiSchema<...>'
```

Same for a union-typed (anyOf/oneOf) field, where `'ui:field': 'AnyOfField'`/`'OneOfField'` (with `ui:fieldReplacesAnyOrOneOf`) is the documented override and is rejected because the field name union is per-primitive-type.

`generateFields()` registers `AnyOfField`, `OneOfField`, `LayoutGridField`, `LayoutHeaderField`, `SchemaField`, `OptionalDataControlsField`, `CyclicSchemaField`, `FallbackField` too — a `Checks` vocabulary that only lists the five per-type fields turns supported core features into type errors. `ui:layoutGrid` also needs a home (common options, or the object check).

packages/utils/src/types.ts
```diff
@@ -1258,6 +1270,114 @@ export type UIOptionsType<
...
+      ? Then extends { field?: infer Fl }
+        ? T extends When
+          ? Fl
+          : never
+        : never
+      : never
+    : never;
+
+/** @internal The intersection of every `then` option (minus `widget`/`field`) valid for form-data type `T`. */
+type RawOptsFor<T, Checks> = UnionToIntersection<
```
jimmycallin: **`UnionToIntersection` collapses an option declared with different types in two checks to `never`.**

When `T` matches more than one check, every `then` is intersected, so an option whose type differs per check becomes unusable. `emptyValue` is exactly that (`string` / `number` / `boolean` / `unknown[]` across the four primitive checks). Two very ordinary cases hit it:

```ts
// T = any (i.e. `UiSchema<any, S, F, CoreUiOptionsChecks>`, the default when there's no form-data type)
const a: UiSchema<any, RJSFSchema, FormContextType, CoreUiOptionsChecks> = { 'ui:emptyValue': '' };
// error TS2322: Type 'string' is not assignable to type 'undefined'

// a oneOf/anyOf field whose data is a union
const b: UiSchema<{ v: string | number }, RJSFSchema, FormContextType, CoreUiOptionsChecks> = {
  v: { 'ui:emptyValue': '' },
};
// error TS2322: Type 'string' is not assignable to type 'undefined'
```

`any` distributes into both branches of `T extends When`, so it picks up *every* check and lands on `string & number & boolean & unknown[]` = `never`. The error message (`not assignable to type 'undefined'`) gives a user no clue what's wrong. A union of the matching `then`s for a given key (rather than an intersection), or at minimum a `Known gaps` entry, is needed here.

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,114 @@
...
+      {
+        widget?: 'HiddenWidget';
+        field?: 'NullField';
+      }
+    >
+  | UiOptionsCheck<
+      readonly unknown[],
+      {
+        widget?: 'CheckboxesWidget' | 'SelectWidget' | 'FileWidget' | 'HiddenWidget' | WidgetAliasFor<'array'>;
+        field?: 'ArrayField';
```
jimmycallin, line 124: **The array check is missing `placeholder` and the enum options that multi-select arrays actually read.**

```ts
type C<T> = UiSchema<T, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
const a: C<{ tags: string[] }> = { tags: { 'ui:placeholder': 'pick' } };   // error
const b: C<{ tags: string[] }> = { tags: { 'ui:enumNames': ['a'] } };      // error
const c: C<{ tags: string[] }> = { tags: { 'ui:enumDisabled': ['a'] } };   // error
const d: C<{ tags: string[] }> = { tags: { 'ui:enumOrder': ['a'] } };      // error
```

All four work at runtime on an array field:

- `ArrayField.tsx:207` — `const itemsUiSchema = (uiSchema?.items ?? uiSchema)` before `optionsList()`, i.e. enum options are deliberately read off the *array's own* uiSchema when there is no `items` entry (the #4985 regression fix), and `CheckboxesWidget`/`SelectWidget` read `options.enumDisabled`.
- `ArrayField.tsx:268`/`281` — `options` from `getUiOptions(uiSchema)` is passed straight to the multi-select widget, which is what the `ui:placeholder` fix earlier in this same v7 changelog relies on ("Fixed `ui:placeholder` being dropped for multi-select arrays, file arrays, custom array widgets and boolean fields").

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,114 @@
...
+        enumDisabled?: EnumValue[];
+        enumNames?: string[] | Record<string | number, string>;
+        enumOrder?: EnumValue[];
+      }
+    >
+  | UiOptionsCheck<
+      boolean,
+      {
+        widget?: 'CheckboxWidget' | 'RadioWidget' | 'SelectWidget' | 'HiddenWidget' | WidgetAliasFor<'boolean'>;
+        field?: 'BooleanField';
```
jimmycallin: @/private/tmp/claude-501/-Users-jimmy-dev-react-jsonschema-form/54d17f25-15fa-48ad-a862-737a602b2b29/scratchpad/c5.md

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,114 @@
...
+        // runtime), so only one needs including here.
+        widget?:
+          | 'TextWidget'
+          | 'RangeWidget'
+          | 'UpDownWidget'
+          | 'SelectWidget'
+          | 'RadioWidget'
+          | 'HiddenWidget'
+          | WidgetAliasFor<'number'>;
+        field?: 'NumberField';
```
jimmycallin: @/private/tmp/claude-501/-Users-jimmy-dev-react-jsonschema-form/54d17f25-15fa-48ad-a862-737a602b2b29/scratchpad/c6.md

packages/core/src/CoreUiOptionsChecks.ts
jimmycallin, line 101: **The boolean check is missing `enumNames` (and `enumDisabled`/`enumOrder`), which `BooleanField` reads.**

```ts
const b: UiSchema<{ flag: boolean }, RJSFSchema, FormContextType, CoreUiOptionsChecks> = {
  flag: { 'ui:enumNames': ['Yes', 'No'] },
};
// error TS2353: 'ui:enumNames' does not exist in type 'UiSchema<boolean, ...>'
```

`BooleanField.tsx:45` destructures `enumNames` out of `getUiOptions()` and uses it to label the two options, and the remaining `...options` (including `enumDisabled`) go to `RadioWidget`/`SelectWidget`, both of which read `options.enumDisabled`. Relabelling a boolean's Yes/No via `ui:enumNames` is a long-standing documented pattern; it becomes a compile error under `Checks`.

packages/core/src/CoreUiOptionsChecks.ts
jimmycallin, line 85: **The number check is missing `inputType`, `autocomplete` and `autocapitalize`, which apply to numeric inputs too.**

```ts
type C<T> = UiSchema<T, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
const a: C<{ age: number }> = { age: { 'ui:inputType': 'tel' } };     // error
const b: C<{ age: number }> = { age: { 'ui:autocomplete': 'off' } };  // error
```

`getInputProps()` (`packages/utils/src/getInputProps.ts:28` and `:60`) applies `options.inputType` and `options.autocomplete` for every input the `BaseInputTemplate` renders, with no branch on schema type — `ui:inputType: 'tel'` on a number field is in fact the documented way to get a numeric keypad. Same for `autocapitalize`.

packages/utils/src/types.ts
heath-freenome: Confirmed and fixed in 8bc64e219. `UiOptionsComponentPart`'s `'ui:options'` shape now intersects `CommonUiOptions<T,S,F>` alongside `{ widget, field } & RawOptsFor`, so `label`, `title`, `classNames`, `disabled`, `help`, `FieldTemplate`, etc. all compile through the grouped `ui:options` form again, not just their `ui:`-prefixed equivalents. Verified against your exact repro (`a` through `e`) with a standalone `tsc` probe before committing.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in 8bc64e219. Added a new `UiOptionsCheck<unknown, {...}>` member to `CoreUiOptionsChecks` - `When = unknown` matches every field regardless of type, so it layers `field?: 'AnyOfField' | 'OneOfField' | 'LayoutGridField' | 'LayoutHeaderField' | 'LayoutMultiSchemaField' | 'SchemaField' | 'OptionalDataControlsField' | 'CyclicSchemaField' | 'FallbackField'` and `layoutGrid?: LayoutGridSchemaType` (imported as a type from `LayoutGridField.tsx`) on top of each type's own field names, rather than replacing them. `ui:field: 'LayoutGridField'`, `ui:layoutGrid`, and `ui:field: 'AnyOfField'/'OneOfField'` on a union-typed field all compile now - verified with a standalone `tsc` probe against your `g1`/`g2` repro plus the anyOf/oneOf case before committing.

packages/utils/src/types.ts
heath-freenome: Confirmed and fixed in 8bc64e219, with your suggested union-by-key approach. Replaced `UnionToIntersection` with `UnionToMergedKeys`, which unions each key's value type across every matching `Checks` member instead of intersecting them - `RawOptsFor` now uses it. `UiSchema<any, ..., CoreUiOptionsChecks>['ui:emptyValue']` and a `{ v: string | number }` field's `v['ui:emptyValue']` both resolve to a real union (`string | number | ...`) instead of `never` now. Verified against both of your repro cases (`a` and `b`) with a standalone `tsc` probe before committing.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in 8bc64e219. Added `placeholder`, `enumDisabled`, `enumNames`, and `enumOrder` to the array check, matching what `ArrayField`'s multi-select path (the `items ?? uiSchema` fallback you cited) actually reads. Verified against all four of your repro cases (`a` through `d`) with a standalone `tsc` probe before committing.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in 8bc64e219. Added `enumDisabled`, `enumNames`, and `enumOrder` to the boolean check, matching what `BooleanField`/`RadioWidget`/`SelectWidget` read. Relabelling a boolean's Yes/No via `ui:enumNames` compiles again - verified with a standalone `tsc` probe before committing.

packages/core/src/CoreUiOptionsChecks.ts
heath-freenome: Confirmed and fixed in 8bc64e219. Added `inputType`, `autocomplete`, and `autocapitalize` to the number check, matching `getInputProps()`, which applies them to every input `BaseInputTemplate` renders with no branch on schema type. `ui:inputType: 'tel'`/`ui:autocomplete` on a number field compile again - verified against both of your repro cases (`a` and `b`) with a standalone `tsc` probe before committing.

packages/docs/docs/api-reference/uiSchema.md
```diff
@@ -975,3 +975,67 @@ const uiSchema: UiSchema = {
...
+- A field whose form-data shape includes an index signature (e.g. from `additionalProperties`/`patternProperties`) only gets type-checking for its explicitly-declared keys; dynamic keys are type-checked but without narrowing beyond the index signature's value type.
+
+### Applying it to an inline uiSchema with `satisfies`
+
+Declaring an intermediate `const uiSchema: UiSchema<FormData, RJSFSchema, FormContextType, CoreUiOptionsChecks> = {...}` and passing that to `Form`'s `uiSchema` prop gets you the narrowing above, because `FormProps['uiSchema']` is typed as the open `UiSchema<T, S, F>` (no `Checks`), which is permissive enough to accept a well-formed closed value with no cast. An inline `uiSchema` object literal passed directly as a JSX prop, though, is checked against that same open type and gets no narrowing at all - check it against the closed form with TypeScript's `satisfies` operator instead. `satisfies` validates the expression while leaving the expression's own type in place for the surrounding `uiSchema` prop, so no intermediate variable or cast is needed:
```
jimmycallin, line 1096: **A closed `UiSchema` is not actually assignable to `FormProps['uiSchema']` — this documented workflow does not compile.**

Verified against this PR's head with `tsc --strict`:

```ts
type Closed = UiSchema<FormData, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
declare const closed: Closed;
const asProp: FormProps<FormData, RJSFSchema, FormContextType>['uiSchema'] = closed;
```

```
error TS2322: Type 'Closed' is not assignable to type 'UiSchema<FormData, RJSFSchema, GenericObjectType> | undefined'.
  Types of property ''ui:options'' are incompatible.
    ... is not assignable to type 'Record<string, string | number | boolean | object | any[] | null | undefined>'.
      Index signature for type 'string' is missing in type '{ widget?: ... } & RawOptsFor<...> & Partial<...> & GlobalUISchemaOptionsKeys & { ...; }'.
```

The open `UiSchema`'s `'ui:options'` is `UIOptionsType`, which intersects `Record<string, ...>`. The closed `'ui:options'` (`UiOptionsComponentPart`) is an intersection containing an **interface**, so TypeScript does not synthesize an implicit index signature for it and the assignment is rejected. The `satisfies` recipe below works (fresh object literals do get an implicit index signature), but the "declare an intermediate `const` and pass it to `Form`" path — the primary, first-described usage — requires a cast.

Worth adding a test that actually assigns a closed `UiSchema` to `FormProps['uiSchema']`; none of the new tests exercise that, which is why this slipped through.

packages/utils/src/types.ts
```diff
@@ -1272,7 +1402,8 @@ export type UiSchemaDefinitions<
   T = any,
   S extends StrictRJSFSchema = RJSFSchema,
   F extends FormContextType = any,
-> = Record<string, UiSchema<T, S, F>>;
+  Checks = never,
+> = Record<string, UiSchema<T, S, F, Checks>>;
```
jimmycallin, line 1415: `UiSchemaDefinitions` reuses the **root** `T` for every `$ref` entry, which was harmless while the type was open but makes `ui:definitions` largely unusable once `Checks` is passed.

```ts
type Closed = UiSchema<FormData, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
const defs: Closed = { 'ui:definitions': { '#/$defs/x': { 'ui:widget': 'TextWidget' } } };
// error TS2322: Type '"TextWidget"' is not assignable to type '"hidden" | Widget<FormData, ...> | undefined'
```

`FormData` is an object, so every definition entry inherits the object vocabulary (`'hidden'` only) no matter what the `$ref` actually resolves to. A definition for a string or number subschema can never name its widget. Since the data type behind a `$ref` isn't knowable from `T`, the entries probably need to fall back to `UiSchema<any, S, F, Checks>` under `Checks`.

packages/utils/src/types.ts
```diff
@@ -1263,6 +1275,124 @@ export type UIOptionsType<
...
+ *
+ * TypeScript has no way to tell "a key admitted only by an index signature" apart from "a key admitted by an index
+ * signature that also happens to declare it explicitly", so a `Then` combining the two (e.g. `GenericObjectType &
+ * { widget: 'Foo' }`) is treated as declaring neither - a narrow, silent under-inclusion rather than a type error.
+ * Not a concern for `Then`s written as an explicit object literal, which is the documented, expected shape.
+ */
+type DeclaresKey<Then, K extends PropertyKey> = string extends keyof Then ? false : K extends keyof Then ? true : false;
+
+/** @internal The union of `then.widget` names valid for form-data type `T`, drawn from `Checks`. */
+type WidgetsFor<T, Checks> =
```
jimmycallin, line 1323: A field whose form-data type is `unknown` gets **no** valid `ui:widget`/`ui:field` name at all, not even a fallback.

```ts
interface FormData { blob: unknown; things: unknown[] }
type Closed = UiSchema<FormData, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
const a: Closed = { blob:   { 'ui:widget': 'TextWidget' } };
const b: Closed = { things: { items: { 'ui:widget': 'TextWidget' } } };
// both: error TS2322: Type 'string' is not assignable to type 'Widget<unknown, RJSFSchema, GenericObjectType> | undefined'
```

`unknown extends When` is false for every concrete `When` (`string`, `number`, `object & …`, `readonly unknown[]`), and the only member it does match (`UiOptionsCheck<unknown, …>`) declares no `widget`, so `WidgetsFor<unknown, Checks>` collapses to `never`. `any` is handled (it matches every branch), but `unknown` — which is what `unknown[]`'s `items` and any deliberately-untyped field resolve to — is not. Treating `unknown` like `any` here (`unknown extends T ? AllWidgets : …`) would close the hole.

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,148 @@
...
+        inline?: boolean;
+        enumDisabled?: EnumValue[];
+        enumNames?: string[] | Record<string | number, string>;
+        enumOrder?: EnumValue[];
+      }
+    >
+  | UiOptionsCheck<
+      null,
+      {
+        widget?: 'HiddenWidget';
```
jimmycallin: The `null` check has the same type/runtime mismatch the object check's comment (below) was written to avoid — inverted.

`SchemaField.tsx:254` hides a field with `uiOptions.widget === 'hidden'` for **every** type, not just objects, and `NullField` never resolves a widget (it renders `null` and calls `getWidget` nowhere; `widgetMap` has no `'null'` entry either). So for a `null`-typed field:

- `'HiddenWidget'` — offered here, but has no runtime effect whatsoever.
- `'hidden'` — the only value that actually does anything, and it's rejected.

```suggestion
      null,
      {
        widget?: 'hidden';
        field?: 'NullField';
      }
```

packages/core/src/CoreUiOptionsChecks.ts
```diff
@@ -0,0 +1,148 @@
...
+        inline?: boolean;
+        filePreview?: boolean;
+        placeholder?: string;
+        emptyValue?: unknown[];
+        enumDisabled?: EnumValue[];
+        enumNames?: string[] | Record<string | number, string>;
+        enumOrder?: EnumValue[];
+      }
+    >
+  | UiOptionsCheck<
```
jimmycallin, line 138: `object & { [Symbol.iterator]?: never }` also matches class instances such as `Date`, so a `Date`-typed field is given the object vocabulary and can only be hidden:

```ts
interface FormData { when: Date }
type Closed = UiSchema<FormData, RJSFSchema, FormContextType, CoreUiOptionsChecks>;
const ui: Closed = { when: { 'ui:widget': 'DateWidget' } };
// error TS2322: Type '"DateWidget"' is not assignable to type '"hidden" | Widget<Date, ...> | undefined'
```

`'DateWidget'`, `'AltDateWidget'`, `'date'`, `'alt-date'` etc. are all rejected, and `'ui:order'` is offered instead. `@rjsf/utils` already has an `AtomicValue` type (used by `UiSchemaFieldMembers` to keep `Date`/`RegExp`/`Map`/… out of nested field names) for exactly this distinction; excluding it from this `When` would keep such fields on the string vocabulary.

packages/utils/src/getWidget.tsx
```diff
@@ -1,9 +1,11 @@
...
 
-/** The map of schema types to widget type to widget name
+/** The map of schema types to widget type to widget name. `as const` so its keys and values stay literal types,
+ * letting `WidgetAliasFor` derive a per-type alias union from it directly instead of a hand-copied one that can
+ * drift out of sync.
  */
-const widgetMap: Record<string, Record<string, string>> = {
+export const widgetMap = {
   boolean: {
     checkbox: 'CheckboxWidget',
```
jimmycallin, line 10: `widgetMap` is now exported but nothing outside this file imports the value (only `WidgetAliasFor` uses it, in-file, via `typeof`), and it isn't re-exported from `packages/utils/src/index.ts` — knip's unused-exports check should fail CI here. Either re-export it or drop `export` and keep the `as const`.

jimmycallin: knip doesn't report it due to exports being counted as the public api for utils. might be worth to look over this.

jimmycallin:
the reviews start to return less value comments now with no critical findings, so i'm approving this one

jimmycallin (approved):

packages/utils/src/getWidget.tsx
heath-freenome: Confirmed and fixed in 02cae7f8b. Nothing outside `getWidget.tsx` imported `widgetMap` (`WidgetAliasFor` only reads its type, and it isn't re-exported from the package index), so I dropped the `export` and kept the `as const`. The emitted `getWidget.d.ts` still declares it for `WidgetAliasFor`'s `typeof`, just without exporting it. `tsc --build`, the utils build, `knip` and the utils tests (100% coverage) all pass.

heath-freenome: Agreed, that explains why CI didn't flag it. Knip treats every `@rjsf/utils` export as public API, so it can't catch an unused one there. I've fixed this instance (02cae7f8b) but left the knip config alone, since tightening it would likely surface other exports across the package and belongs in its own PR.

PR #5278, opened 2026-09-12
fix(utils)!: describe error trees correctly — ErrorSchema and FormValidation from one shared type

jimmycallin:
`ErrorSchema<T>` could not describe an error on any field of a typed form. Neither could `FormValidation<T>`. Both are fixed here by one shared tree type.

This description is long on purpose: the bug is a TypeScript subtlety that is easy to reintroduce, so it is worth understanding rather than just reviewing.

---

## 1. What an error schema is at runtime

`toErrorSchema()` builds a tree that mirrors the form data. Each node holds its own messages under the `__errors` key, and a child per field:

```js
// formData: { name: 'x', address: { city: '' }, tags: ['a'] }
{
  __errors: ['form-level problem'],
  name:    { __errors: ['too short'] },
  address: { city: { __errors: ['unknown city'] } },
  tags:    { 0: { __errors: ['not allowed'] } },
}
```

Two details matter later:

- A **leaf** field's node is just `{ __errors: [...] }`. There are no children below it.
- An **array**'s node is keyed by index but is a *plain object*, not an array. `ErrorSchemaBuilder` writes numeric path segments as object keys and never auto-vivifies an array.

`createErrorHandler()` builds the same tree for `customValidate`, with an `addError()` on every node. That is `FormValidation<T>`.

## 2. The type that was supposed to describe it

```ts
export type ErrorSchema<T = any> = FieldErrors & { [key in keyof T]?: ErrorSchema<T[key]> };
export type FormValidation<T = any> = FieldValidation & { [key in keyof T]?: FormValidation<T[key]> };
```

Read plainly this looks right: a node is its own errors, plus an optional child node per key of `T`. It is wrong, and the reason is a rule about mapped types that is easy to miss.

## 3. The rule: homomorphic mapped types preserve the thing they map

A mapped type written in exactly the form `{ [K in keyof T]: … }` is called **homomorphic**. TypeScript gives it a special behaviour: when `T` is a primitive or an array, the mapped type **is not an object type at all — it is `T` itself**.

You can watch it happen in three lines:

```ts
type Mapped<T> = { [K in keyof T]?: string };

type A = Mapped<{ a: number }>;  // { a?: string }     — as expected
type B = Mapped<string>;         // string             — not an object!
type C = Mapped<string[]>;       // (string | undefined)[] — an array, not a record
```

This is deliberate: it is what lets `Partial<T>`, `Readonly<T>` and friends pass arrays and primitives through unchanged. It is exactly wrong when the mapped type is meant to describe a *different* structure, which is our case — an error tree is not shaped like the data, it is shaped like a tree *about* the data.

## 4. What that rule did to `ErrorSchema`

**A leaf field's errors became unrepresentable.** For a string field, the mapped half collapses to `string`, and intersecting an object with a primitive leaves the primitive:

```ts
ErrorSchema<string>
// = FieldErrors & { [K in keyof string]?: … }
// = FieldErrors & string
// ≈ string
```

So this fails, with the compiler saying something that looks nonsensical until you know the rule above:

```ts
const e: ErrorSchema<{ name: string }> = { name: { __errors: ['too short'] } };
//                                         ~~~~
// Type '{ __errors: string[]; }' is not assignable to type 'string'.
```

**An array's errors became an array.** `ErrorSchema<string[]>` resolved to an array type, while the runtime builds `{ 0: {...} }`. The type and the value disagreed about the basic shape.

**`ErrorSchema<any>` rejected every concrete error schema.** `keyof any` is `string | number | symbol`, so the mapped half becomes an index signature:

```ts
ErrorSchema<any>  // ≈ FieldErrors & { [x: string]: ErrorSchema<any> | undefined }
```

An object *type* is only assignable to an index-signature type if it has an index signature of its own — TypeScript infers one implicitly for fresh object literals, but not for a named type. And `FormProps['extraErrors']` is `ErrorSchema<T>` with `T = any` by default. So a user who correctly typed their error schema could not pass it to `extraErrors`.

Put together: **only a root-level `__errors` ever type-checked.** Every other use needed an escape hatch, which is why 18 `as unknown as ErrorSchema` assertions had piled up in this repo alone. Those casts were the symptom; this is the cause.

## 5. Why nobody noticed `FormValidation`

It has the identical defect, but it is *read*, not *constructed*. `FormValidation<string>` collapses to `FieldValidation & string`, and an intersection still surfaces the members of both halves — so `errors.name?.addError('…')` keeps working and nothing complains. The type was wrong the whole time; the usage pattern just never asked it the question that exposes it.

## 6. The fix: normalise the data type, *then* map

The collapse is triggered by the syntactic form `[K in keyof T]`. So the fix is not to special-case the result — it is to stop mapping over `keyof T` directly. Map over a **normalised** object type instead, and the mapped type is never homomorphic:

```ts
type ErrorTree<V, Node> = Node & {
  [key in keyof ErrorTreeChildData<V>]?: ErrorTreeChild<ErrorTreeChildData<V>[key], Node>;
};

export type ErrorSchema<T = any>    = ErrorTree<T, FieldErrors>;
export type FormValidation<T = any> = ErrorTree<T, FieldValidation>;
```

`ErrorTreeChildData<V>` answers one question — *what are this node's children?* — and normalises each kind of value to a plain object before anything is mapped:

| `V` | children | why |
| --- | --- | --- |
| `any` | `Record<string, any>` | shape unknown, so the node is unconstrained |
| `Item[]` | `Record<number, Item>` | a node per index, as a plain keyed object |
| `Date`, `File`, `RegExp`, a function | none | an object, but holds no form fields |
| an object | itself | the normal case |
| a primitive | none | a leaf has no children |

`NonNullable<V>` is applied before each check, so an optional property (`address?: { city: string }`) is tested as `{ city: string }` rather than `{ city: string } | undefined` — otherwise the conditional distributes across the union and produces a union of two node types.

The two types now differ only in the payload each node carries: `FieldErrors` for `ErrorSchema`, `FieldValidation` for `FormValidation`. One definition, one place to fix, both correct.

### The non-obvious constraint

It is tempting to write the top level as a conditional instead:

```ts
type ErrorTree<V, Node> = Node & (V extends … ? … : …);   // don't
```

That breaks the library. For an *unresolved* generic `T` — which is what internal code sees, e.g. inside `Form<T, S, F>` — TypeScript cannot evaluate the conditional, so it defers it, and `{}` stops being assignable to `ErrorSchema<T>`. That costs about a dozen internal sites. Keeping the conditional *inside* `keyof …` leaves a mapped type at every level, so `{}` stays assignable. Worth knowing before anyone "simplifies" this later.

### What now type-checks

```ts
type FormData = { name: string; address?: { city: string }; tags: string[]; when: Date };

const errors: ErrorSchema<FormData> = {
  __errors: ['form-level problem'],
  name:    { __errors: ['too short'] },
  address: { city: { __errors: ['unknown city'] } },
  tags:    { 0: { __errors: ['not allowed'] } },
  when:    { __errors: ['bad date'] },
};

// a form whose root is not an object works too
const scalarRoot: ErrorSchema<string>   = { __errors: ['bad'] };
const arrayRoot:  ErrorSchema<string[]> = { 0: { __errors: ['bad item'] } };
```

## 7. Alternatives that were measured and rejected

Each of these was implemented and typechecked across all 17 projects, rather than reasoned about:

| Approach | Result |
| --- | --- |
| **This PR** | **0 errors** |
| Key remapping (`[K in keyof T as K]`) to defeat homomorphism | Does not work — still collapses for primitives, and for arrays produces a record of `push`/`toString`/`length` |
| Make the top level conditional | 25 errors: `{}` no longer assignable to `ErrorSchema<T>` for unresolved `T` |
| Drop the `<T>` parameter for a plain recursive interface | 174 errors; removes a public generic used throughout user code |
| `any` branch → `ErrorSchema` | 38 errors |
| `any` branch → `FieldErrors & { [key: string]: unknown }` | 56 errors |

## 8. Two things deliberately left alone

**The `__errors` magic key stays.** Storing a node's own errors *in the same namespace as its children*, discriminated by a reserved key, is the deeper root cause: it is why the index signature collides with `__errors`, and why `any` needs a short-circuit. A node shaped `{ errors, children }` makes the whole class of problem disappear. It is also the format users write in `extraErrors`, that every theme reads, and that is baked into the docs — so that is a separate and much larger conversation, not a rider on a type fix.

**Internals still assert at the boundary.** A recursive walk keyed by a runtime string cannot be expressed against a type derived from the data, so `toErrorList()` and `unwrapErrorHandler()` continue to work on `Record<string, unknown>` — their existing comments already explain this. The derived type earns its keep where a field is named by the developer, not inside the machinery. This is the same split Zod uses: a flat canonical `issues` list for traversal, a derived `format()` view for reading by field.

## 9. Two latent bugs this exposed

Both were invisible while every `ErrorSchema<X>` was accidentally assignable to every other:

- **`Form.liveValidate()`** declared `originalErrorSchema: ErrorSchema<S>` — an error schema keyed by the *schema* type rather than the form data type. Every caller passes `state.errorSchema`, which is `ErrorSchema<T>`. Fixing the declaration also removed the `as ErrorSchema<T>` its body needed.
- **`ArrayField`'s `handleChange`** declared `ErrorSchema<T>` while filling a slot typed `ErrorSchema<T[]>`, casting on the way out. It now declares the type its contract specifies, and the cast is gone.

## 10. Fallout and migration

- 18 `as unknown as ErrorSchema` assertions removed (16 in `Form.behaviors.test.tsx`, 2 in `arrayTests.tsx`).
- A `createErrorHandler(...) as FormValidation` cast and a `@ts-expect-error` in the utils tests became unnecessary.
- Nothing else in the monorepo needed changing.

`packages/utils/test/errorSchemaTypes.test.ts` pins every shape for both trees — leaves, nested objects, optional properties, arrays of leaves and of objects, atomic values, non-object roots, untyped schemas, and `addError()` at each level. It asserts on real values rather than types alone, so it fails if the types and the runtime builders ever drift apart.

**Breaking change**, but only for code that depended on the old shapes — for example treating `ErrorSchema<SomePrimitive>` as that primitive, a value that could never hold errors in the first place. If you worked around the old shapes with a cast, drop it:

```diff
-const extraErrors = { foo: { __errors: ['x'] } } as unknown as ErrorSchema;
+const extraErrors: ErrorSchema<FormData> = { foo: { __errors: ['x'] } };
```

Nodes are optional, because the tree only contains fields that are actually present, so reach into them with `?.`:

```diff
-errors.address.city.addError('unknown city');
+errors.address?.city?.addError('unknown city');
```

## Checklist

  - [x] I've verified the existing tests pass and root `typecheck`, `knip`, lint and `cs-check` are clean
  - [x] `@rjsf/utils` still reports 100% coverage


heath-freenome:
## Code review findings

**1. `packages/utils/src/types.ts:265`** — `ErrorTreeChildData<V>` falls through to `Record<never, never>` for a mixed/union-typed field, so `ErrorSchema<T>` rejects error nodes that `toErrorSchema()`/`createErrorHandler()` actually build at runtime.

For example, with `type Data = { thing: string | { city: string } }`:
```ts
const x: ErrorSchema<Data> = { thing: { city: { __errors: ['bad'] } } };
```
fails to compile (`TS2353: Object literal may only specify known properties, and 'city' does not exist in type 'ErrorTree<...>'`), even though this is exactly the node a real form emits when the field's current value is the object variant of a `oneOf`/`anyOf`-shaped property or a mixed-type tuple. The PR's migration guide claims tuples/leaves/arrays/objects are all handled, but this common RJSF shape (`oneOf`/`anyOf`) is not.

**2. `packages/core/src/components/fields/ArrayField.tsx:898`** — `handleAddItem`, `handleCopyItem`, `handleRemoveItem`, and `handleReorderItems` (around lines 898/921, 938/961, 978/991, 1009/1031) still declare `newErrorSchema: ErrorSchema<T>` and cast it `as ErrorSchema<T[]>` at their `onChange()` call — the exact same mismatch the PR fixed only in the neighboring `handleChange` (line 1042).

This contradicts the PR's own stated rationale (`ArrayField`'s `handleChange` declared `ErrorSchema<T>` while filling a slot typed `ErrorSchema<T[]>`, needing a cast on the way out) — that identical defect, and its now-unnecessary cast, survives unfixed in four sibling functions in the same file, so the type-safety gap this PR set out to close is still open for add/copy/remove/reorder.

---

Lint, typecheck, and tests all pass cleanly on this branch — these are gaps not currently exercised by the test suite.

jimmycallin:
Both findings fixed in b938b0fdf, and the branch is rebased onto current `v7` (`9ee6e2a4a`) — only `CHANGELOG-v7.md` conflicted.

**1. Union-typed fields rejecting their children** — confirmed, and the cause was one step below where the report points: the conditional that picks a value's children tested `NonNullable<V>` rather than a naked type parameter, so it never distributed over a union. `string | { city: string }` matched neither the array branch nor the object branch and fell through to `Record<never, never>`.

The branch selection now happens in a small `ChildDataOf<V>` helper with a naked `V`, so it distributes, and the mapped type unions the keys each member contributes (`ChildKeys`) and the child data behind each key (`ChildAt`). A union-typed field accepts the children of every member it can hold, which is what a `oneOf`/`anyOf` property needs:

```ts
type Data = { thing: string | { city: string } };
const x: ErrorSchema<Data> = { thing: { city: { __errors: ['bad'] } } }; // compiles
const y: ErrorSchema<Data> = { thing: { __errors: ['bad'] } }; // still compiles
```

Type tests cover the object member, the scalar member, a union of two object shapes (both key sets accepted), and `string | string[]` (indexed node). They fail to compile on the previous commit. The migration guide's shape section now mentions unions.

**2. `ArrayField`'s four sibling handlers** — also confirmed. `handleAddItem`, `handleCopyItem`, `handleRemoveItem` and `handleReorderItems` each declared `ErrorSchema<T>` for a value the array's `onChange` takes as `ErrorSchema<T[]>`. All four now declare `ErrorSchema<T[]>` and the four `as ErrorSchema<T[]>` casts are gone.

Lint, knip, root typecheck, `cs-check`, build and the full test suite pass.

heath-freenome:
## Follow-up review

The two issues from the earlier review comment appear to be addressed. Two new findings from a re-review:

**1. `packages/core/src/components/fields/ArrayField.tsx:621`** (duplicated at line 760 in `FixedArray`) — `itemErrorSchema` is force-cast to `ErrorSchema<T[]>` when its natural type is `ErrorSchema<T>` — the same wrong-generic-argument bug this PR fixed in the sibling mutation handlers (`handleAddItem`/`handleCopyItem`/`handleRemoveItem`/`handleReorderItems`), left unfixed here.

`errorSchema` on `NormalArray`/`FixedArray` is `ErrorSchema<T[]>` (inherited via `FieldProps<T[]>`), so `errorSchema[index]` naturally resolves to `ErrorSchema<T> | undefined` — confirmed by direct `tsc` reproduction: assigning that expression to an `ErrorSchema<T[]>`-typed variable fails to compile, while assigning it to `ErrorSchema<T>` succeeds. The code force-casts it back to `ErrorSchema<T[]>` before handing it down as `itemErrorSchema` (declared with that same wrong type at line 377) to the per-item `ItemSchemaField`, mistyping a single item's error node as the whole array's. This is precisely the type-safety hole this PR set out to close, reopened one call site over from where it was already fixed.

**2. `packages/utils/src/types.ts:256`** — `AtomicValue = Date | RegExp | File | Blob | ((...args: never[]) => unknown)` is a hardcoded allowlist deciding which object types are error-tree leaves, rather than a general rule. Any other opaque object value (`Map`, `Set`, a custom class instance, `FileList`, `ArrayBuffer`, etc.) falls through to the `V extends object ? V : ...` branch and is typed as if it had per-property child error nodes.

A form field backed by a custom widget that treats its value as an atomic leaf (e.g. a `Money`/`Decimal`/`LatLng` value object that never gets nested `toErrorSchema()` children) still type-checks `ErrorSchema<FormData>` as though `amount`/`currency`-style properties were valid nested error keys for that field — an over-permissive type that accepts an error shape the runtime tree-builder never actually produces for that field, for any atomic type outside this fixed four-member list.

---

packages/utils/src/types.ts
```diff
@@ -249,23 +249,61 @@ export interface FieldErrors {
   __errors?: FieldError[];
 }
 
-/** Type describing a recursive structure of `FieldErrors`s for an object with a non-empty set of keys */
-export type ErrorSchema<T = any> = FieldErrors & {
-  /** The set of errors for fields in the recursive object structure */
-  [key in keyof T]?: ErrorSchema<T[key]>;
+/** True when `V` is `any`, which a conditional type otherwise matches on every branch at once */
+type IsAny<V> = 0 extends 1 & V ? true : false;
```
heath-freenome, line 253: Consider adding these new types into the v7.x upgrade guide as well.

jimmycallin:
Fixed in `a44eebda4`.

**1. `itemErrorSchema` force-cast** — real, though the cast can't fully disappear: the natural type does flow from `errorSchema[index]`, but `ItemSchemaField` is `registry.fields.SchemaField` off a `Registry<T[], S, F>`, so its `errorSchema` prop is `ErrorSchema<T[]>` no matter what we hand it. Declaring `itemErrorSchema` as `ErrorSchema<T>` and dropping the two casts at source just relocates the error to the `<ItemSchemaField>` call — confirmed with `tsc`.

So the type is now truthful everywhere it can be, and the one lie is localized: both `NormalArray` and `FixedArray` read `errorSchema?.[index]` with no assertion, `ArrayFieldItemInner` declares `itemErrorSchema?: ErrorSchema<T>`, and the single remaining cast sits at the `ItemSchemaField` call, right beside the `formData={itemData}` cast that exists for the same reason (and is already commented as such). The underlying cause is that the registry is parameterized by the array's data type, so every per-item prop in this file is expressed as `T[]` — that's a wider refactor than a type fix to `ErrorSchema`.

**2. `AtomicValue` as an allowlist** — accurate description, but I'm leaving it. TypeScript has no general "opaque object" predicate: a `Money`/`Decimal`/`LatLng` value object is structurally an object with properties, indistinguishable at the type level from a nested form object. Every heuristic I can express (e.g. "has method-valued properties") misclassifies ordinary data objects that happen to carry a `toJSON`, which would be the worse failure — a *rejected* error shape the runtime does build, rather than an accepted one it doesn't.

The four members are there because they're the DOM/JS values RJSF itself puts in form data (`FileWidget` produces `File`/`Blob`; date widgets round-trip `Date`). For anything else the fallback is over-permissive, which accepts an error node that never occurs — noise, not a hole. Given the choice between over- and under-permissive for an unknowable case, over- is the safe direction.

packages/utils/src/types.ts
jimmycallin: Done in `740acbc4c`.

The helper types themselves (`ErrorTree`, `ErrorTreeChildData`, `ChildDataOf`, `ChildKeys`, `ChildAt`, `AtomicValue`, `IsAny`) are deliberately unexported — the public surface is still just `ErrorSchema<T>` and `FormValidation<T>`, so listing the internals under "New types" would document names nobody can import, and would pin an implementation I'd rather stay free to change.

What a reader actually needs is the rule those types encode, so the guide's `ErrorSchema<T>`/`FormValidation<T>` section now states it as a table — which field types get a node per property, a node per index, or no children at all — plus the two cases that can't be inferred from the examples: an optional property is described by the type it holds, and a class instance outside the atomic list is treated as an object.

Let me know if you'd rather the helpers were exported; that's a separate call from documenting them.

heath-freenome:
## Follow-up review (round 3)

The issues from the prior two rounds are resolved: the union-type fallthrough and sibling-handler casts were fixed in `b938b0fdf`, and the `itemErrorSchema` cast was relocated in `a44eebda4`. `AtomicValue`'s hardcoded allowlist was reconsidered and judged a deliberate, reasonable tradeoff rather than a bug.

One new finding from this round:

**`packages/utils/src/types.ts:271`** — `ChildDataOf` treats a heterogeneous tuple exactly like a homogeneous array, giving every index the union of all element types instead of that position's own type, so `ErrorSchema<T>` accepts error nodes shaped like the wrong tuple member at a given index.

For `type Data = { pair: [string, { city: string }] }`, `ErrorSchema<Data>` wrongly type-checks `{ pair: { 0: { city: { __errors: ['bad'] } } } }` (position 0 is a `string` but accepts an object-shaped child) and equally wrongly type-checks `{ pair: { 1: { __errors: ['bad'] } } }` (position 1 is an object but accepts a bare leaf) — verified by direct `tsc` compilation of the PR's type against both cases (0 errors), while an unrelated bogus key at the same position is correctly rejected, showing the constraint is real but not position-aware. The migration guide's table states "An array or tuple → A node per index" with no caveat, overclaiming precision the implementation doesn't provide for tuples with differently-shaped members.

---

jimmycallin:
Fixed in `ed701f393`.

Confirmed: `ChildDataOf` normalized every `readonly unknown[]` to `Record<number, V[number]>`, which is right for an array and wrong for a tuple — `[string, { city: string }]` gave both positions `string | { city: string }`.

A tuple is now detected by `number extends V['length']` being false, and its child data is built from its own index keys (`Extract<keyof V, `${number}`>`), so each position keeps its own type. Plain arrays are unchanged — there is no per-position type there to keep.

```ts
type Data = { pair: [string, { city: string }] };

const ok1: ErrorSchema<Data> = { pair: { 0: { __errors: ['bad label'] } } };
const ok2: ErrorSchema<Data> = { pair: { 1: { city: { __errors: ['bad city'] } } } };
const bad1: ErrorSchema<Data> = { pair: { 0: { city: { __errors: ['x'] } } } }; // error
const bad2: ErrorSchema<Data> = { pair: { 2: { __errors: ['x'] } } };           // error
```

`errorSchemaTypes.test.ts` pins all four, the two rejections as `@ts-expect-error` — so the test fails to compile if the tuple case ever collapses back to an array. The guide's table now lists arrays and tuples as separate rows.

Lint, knip, root typecheck, `cs-check` and the full test suite pass.

heath-freenome:
## Follow-up review (round 4)

All issues from the prior three rounds are confirmed fixed at current HEAD (`ed701f393`): `Form.tsx:806` uses `ErrorSchema<T>`, `ArrayField.tsx`'s four sibling handlers now use `ErrorSchema<T[]>`, the `itemErrorSchema` cast is relocated to the `ItemSchemaField` call site, and fixed-length tuple positions now use `Extract<keyof V, \`${number}\`>` correctly.

One new, narrower edge case surfaced this round:

**`packages/utils/src/types.ts:275`** — A variadic (rest-element) tuple is normalized like a plain homogeneous array, so a fixed leading position loses its own type and accepts a child shape from a later, differently-typed position.

For `type Data = { pair: [string, ...({ city: string })[]] }` (position 0 is a leaf `string`, later positions are objects), `ErrorSchema<Data>` wrongly type-checks `{ pair: { 0: { city: { __errors: ['bad'] } } } }` — position 0 accepts a nested `city` error node it should reject as a leaf. This is the same class of bug already fixed for fixed-length tuples (`number extends V['length']` correctly detects and special-cases those), but a tuple with a rest element still has `number extends V['length']` evaluate to `true`, so it falls into the plain-array branch (`Record<number, V[number]>`) and every index gets the union of all element types instead of its own. The migration guide's table row "A tuple → A node per position, each of that position's own type" makes this guarantee with no caveat for rest-element tuples — verified by direct `tsc` compilation (0 errors) against the PR's exact type definitions.

---

jimmycallin:
Fixed in `ded5a1b2b`.

Confirmed: `number extends V['length']` is true for `[string, ...X[]]`, so it took the plain-array branch and every position got `string | X`.

The array and tuple branches are now one mapped type. Its keys are the tuple's declared positions (`Extract<keyof V, `${number}`>`, empty for a plain array) plus `number` when the length is open. That gives a fixed tuple only its positions, an array only the index signature, and a variadic tuple both — and because they live in the same object type rather than an intersection, a declared position takes precedence over the index signature:

```ts
type Data = { labelled: [string, ...{ city: string }[]] };

const ok1: ErrorSchema<Data> = { labelled: { 0: { __errors: ['bad label'] } } };
const ok2: ErrorSchema<Data> = { labelled: { 3: { city: { __errors: ['bad city'] } } } };
const bad: ErrorSchema<Data> = { labelled: { 0: { city: { __errors: ['x'] } } } }; // error
```

The rest positions hold the union of every element type, as an array does. `errorSchemaTypes.test.ts` pins the three cases, the rejection as `@ts-expect-error`, and the guide's table gains a row for a tuple with a rest element.

Root typecheck, uncached lint, knip, `cs-check` and the `@rjsf/utils` suite pass.

heath-freenome:
## Follow-up review (round 5)

The variadic-tuple issue from round 4 is fixed in `ded5a1b2b` — confirmed via direct `tsc` compilation of the three pinned cases in `errorSchemaTypes.test.ts`. All issues from rounds 1–4 (union fallthrough, `ArrayField` sibling casts, the `itemErrorSchema` cast, fixed-length tuple handling) remain fixed at current HEAD. Root typecheck and the `@rjsf/utils` suite (100% coverage) both pass cleanly.

Two new findings from this round:

**1. `packages/utils/src/types.ts:272`** — a field typed `unknown` (as opposed to `any`) is forced to a leaf-only error node, rejecting the nested error shape `toErrorSchema()`/`createErrorHandler()` actually build when the runtime value is an object.

For `type Data = { thing: unknown }` (or the common `Record<string, unknown>` pattern for loosely-typed/`additionalProperties: true` schema data):
```ts
const e: ErrorSchema<Data> = { thing: { nested: { __errors: ['bad'] } } }; // TS2353
```
`any` gets an explicit `IsAny` short-circuit to an unconstrained node, but `unknown` falls through `ChildDataOf<V>`'s final `Record<never, never>` branch, since `unknown extends object` is false — it's treated identically to a primitive leaf. The migration guide's shape table lists object/array/tuple/primitive/Date/File/union/any but never mentions `unknown`, so this gap is undocumented and inconsistent with how `any` is deliberately handled.

**2. `packages/utils/src/types.ts:255`** — the `AtomicValue` allowlist's justifying comment cites the wrong runtime function. It says the four members mirror what's treated as opaque at runtime, but the function that actually walks form data to build `FormValidation` (`createErrorHandler()`, via `isPlainObject()`) treats *every* class instance — not just `Date`/`File`/`RegExp`/functions — as a leaf.

`isPlainObject()` returns `false` (i.e. runtime leaf, no recursion) for any value whose prototype isn't `null` or `Object.prototype`, so a `Map`, `Set`, or any custom class instance (e.g. a `Money`/`LatLng` value object) is already a leaf at runtime under `createErrorHandler()`. But `ChildDataOf<V>` only special-cases the four `AtomicValue` members before falling to `V extends object ? V : ...`, so `FormValidation<{ amount: Money }>` type-checks as though `Money`'s properties were valid nested `addError()` targets — a shape `createErrorHandler()` never actually produces for that field. This is the same class of gap round 2 raised and round 2's response accepted as a deliberate "over-permissive, hence safe" tradeoff — but the real governing function (`isPlainObject()`, also used by `toErrorList()`/`unwrapErrorHandler()`) is far stricter than the four-member list the tradeoff was framed around, so the type's blind spot is every class instance, not just the ones outside an arbitrary allowlist. Worth at minimum fixing the comment so it doesn't point at the wrong function; whether the underlying allowlist itself is worth widening is the same judgment call round 2 already made.

---

jimmycallin:
Fixed in `8ca92f345`.

**1. `unknown` forced to a leaf** — confirmed. `unknown extends object` is false, so it fell past the object branch to `Record<never, never>` while `any` was short-circuited explicitly. `ErrorTreeChildData` now tests `unknown extends V`, which is true for both, so a field typed `unknown` (and a `Record<string, unknown>` form data type) contributes unconstrained children:

```ts
type Data = { loose: unknown };
const leaf: ErrorSchema<Data> = { loose: { __errors: ['bad'] } };
const nested: ErrorSchema<Data> = { loose: { nested: { __errors: ['bad'] } } }; // compiles now
```

Both are pinned in `errorSchemaTypes.test.ts`, and the guide's shape table row is now `any` or `unknown`.

**2. `AtomicValue`'s comment** — the allowlist itself was already narrowed to `Date | File` in `86a4a5b40` (pushed before this round, but after the commit the review compiled against, so `RegExp`/`Blob`/the function type are gone). The comment was still wrong in the way you describe, though: it named `isObject()`, which special-cases nothing of the sort. It now names `isPlainObject()` as the predicate `createErrorHandler()` and `toErrorList()` actually recurse through, says that every class instance is therefore a leaf at runtime, and says plainly that the two members are the class instances RJSF itself puts in form data — leaving any other one over-permissive, which is the round-2 tradeoff stated as such rather than dressed up as a runtime match.

Root typecheck and the `@rjsf/utils` suite pass.

heath-freenome:
## Follow-up review (round 6)

Round 5's `unknown`-handling fix (`8ca92f345`) holds — verified by compiling `types.ts` against ~15 adversarial edge cases (unresolved generics, `Record<string, unknown>`, readonly tuples, nested `unknown` in arrays/optional/nullable positions, unions, arrays-of-unions, `never`, optional tuple elements, mixed index-signature objects, nested arrays, literal unions) with `tsc --strict`. All behaved as claimed.

One new, minor finding:

**`packages/utils/src/types.ts:258`** — the `AtomicValue` comment claims `Date`/`File` are "the two class instances RJSF itself puts in form data," but RJSF's own widgets never actually store either as an instance there.

`useFileWidgetProps.ts`'s `onChange` is typed `(value?: string | null | (string | null)[]) => void`, and every `DateWidget`/`DateTimeWidget` across `packages/core`, `packages/daisyui`, `packages/mantine`, and `packages/antd` calls `.toISOString()`/produces a string before calling `onChange` — so RJSF's built-in widgets put ISO strings and data-URL strings into formData, never `Date` or `File` instances. The allowlist may still be a reasonable choice for user-authored custom widgets or app-level TS types, but the comment's stated justification for why these two specific classes were chosen is factually wrong, and a reader would be misled about what actually motivates the allowlist.

---

heath-freenome (approved):
@jimmycallin approved with a request for a small documentation change per the comment above

PR #5279, opened 2026-09-12
fix(utils)!: check uiSchema field names against the form data type

jimmycallin:

`UiSchema<T>` accepts every string key today, because it intersects `GenericObjectType` (`Record<string, any>`). That index signature does real work — it is what lets a uiSchema nest by field name — but it also means a misspelled or renamed field is silently ignored, with no error and no rendering effect. This is the field-name half of [#4916](https://github.com/rjsf-team/react-jsonschema-form/issues/4916), and it follows from the review on #5275: the vocabulary half (restricting `ui:widget` to a declared set, closing the `ui:` namespace) genuinely has to be opted into, but **field names do not** — `FormProps` already infers `T` from `formData`.

### Solution: derive the per-field entries from the data type

Same technique as #5278: normalise the data type, then map over it.

```ts
export type UiSchema<T = any, S = RJSFSchema, F = any> = {
  [K in keyof UiSchemaChildData<T>]?: UiSchemaChild<UiSchemaChildData<T>[K], S, F>;
} & MakeUIType<UIOptionsBaseType<T, S, F>> & { /* the existing ui: keys, items, ui:definitions … */ };
```

`UiSchemaChildData<T>` decides what may nest: `any` keeps every field name open, an array nests through `items` rather than by index, a leaf has no nested fields, and an object contributes its own keys. Nothing else in `UiSchema` changes.

```ts
type FormData = { firstName: string; address: { city: string } };

const uiSchema: UiSchema<FormData> = {
  firstName: { 'ui:widget': 'textarea' },
  address: { city: { 'ui:placeholder': 'City' } },
  lastName: { 'ui:widget': 'text' }, // error: `lastName` is not a field of FormData
};
```

Two properties hold, and are asserted as type-level tests: `string extends keyof UiSchema<Known>` is **false** (field names are closed) while `` `ui:${string}` extends keyof UiSchema<Known> `` is **true** (the directive namespace stays open). The second comes for free — `UIOptionsBaseType` is string-keyed, so `MakeUIType`'s remap already produces a `ui:${string}` signature. Theme and application directives keep working untouched.

`oneOf`/`anyOf` fields are typed as unions, and a uiSchema for one legitimately names keys from any branch, since which branch is present depends on the current value. `UnionMembersMerged` collects the keys of every member, so all branches' keys are accepted.

### What does not change

- **Untyped uiSchemas.** `UiSchema` with no type argument, or with `any` form data, accepts exactly what it accepted before, arbitrary nesting included. The single `any` in `UiSchemaChild` is deliberate: `any` in → `any` out is the faithful encoding of "the caller passed `any`", the same conclusion as #5278.
- **The `ui:` namespace**, at every level.
- **Runtime behaviour.** This PR is types plus the call sites they forced; no rendering changes.

### Five structural keys the index signature was hiding

`additionalProperties`, `additionalItems`, `anyOf`, `oneOf` (the last two positional lists, one uiSchema per subschema) and the legacy unprefixed `classNames` are read at runtime — by `ObjectField`, `ArrayField`, `optionsList`, `MultiSchemaField` and `SchemaField` — but were never declared; they only type-checked by accident. They are declared now, which is a documentation improvement the strict type forced.

Two existing tests passed `oneOf`/`anyOf` an object where a list belongs, deliberately, to exercise the "not an array" runtime warning; they now carry a `@ts-expect-error`.

### The rest of the cost

19 errors across the monorepo, in four mechanical buckets, all in this PR:

1. The five undeclared keys above.
2. Three `uiSchema = {}` destructure defaults (`Theme.ts`, `ObjectField`, daisyui's `getDaisy`) that inferred `{}`; annotated.
3. Runtime-key traversal (`getUiOptions`, `ObjectField`'s property loop, `optionsList`) — the boundary pattern `toErrorList` and `unwrapErrorHandler` already use: view as a keyed record where you walk by runtime key.
4. Spreading a generic uiSchema (`LayoutGridField`, antd and mantine `WrapIfAdditionalTemplate`) flattens the mapped half, which TS cannot re-check against an unresolved `T`; each names the type it still is. daisyui's `DaisyUiSchema` became a `type` alias, since TS forbids `interface … extends` a type with an index signature.

### Breaking change

Only for typed users whose uiSchema names a field absent from `T` — a property added by a `dependencies` subschema, or an `additionalProperties` key when `T` has no index signature. The migration guide covers those, plus the positional-list keys and the pre-existing (not new) rejection of a uiSchema typed by a declared `interface`. Untyped users are unaffected.

### Relationship to other PRs

- **#5275** — this is recommendation 1 from [my review there](https://github.com/rjsf-team/react-jsonschema-form/pull/5275#issuecomment-5648042303). With field names checked by inference, what remains to opt into is only the vocabulary tier, which that PR's `UiOptionsCheck`/`Checks` machinery is the right shape for.
- **#5278** — `IsAnyType` here and `IsAny` there are the same helper arrived at independently; whichever lands second should collapse them into one.

### Validation

Lint, knip, root typecheck, `cs-check`, build and all 15 test targets pass. New type tests in `packages/utils/test/uiSchemaTypes.test.ts` cover the accepted shapes, the `oneOf` branches, an untyped uiSchema, the rejected field names (via `@ts-expect-error`, which fails the build if they ever start compiling) and the two key-openness properties.

### Checklist

  - [x] I've added/updated tests


jimmycallin:
Removed every type assertion the first commit introduced — f2f721ebe. No `as` remains in this PR's diff.

| Was | Now |
| --- | --- |
| `uiSchema = {} as UiSchema<T, S, F>` (×3) | annotate the destructuring pattern — `{}` is already assignable to `UiSchema<T, S, F>`, only the *default* needed a contextual type |
| `uiSchema as Record<string, unknown>` in `getUiOptions()` | `Object.entries()`, which hands over key and value together, so there is nothing to index |
| `(uiSchema as Record<string, UiSchema>)[propertyName]` in `ObjectField` | `getByPath()`, which is how the rest of the codebase reads a uiSchema by a runtime key, and types its own result (it also drops the accidental prototype-key lookup the bare index had) |
| `{ ...uiSchema } as UiSchema<T, S, F>` (×4) | `Object.assign({}, uiSchema)` into an annotated target |
| `remainingUiSchema as UiSchema<T, S, F>` in `SchemaField` | a copy plus three `delete`s, replacing the rest-destructure whose `Omit<…>` was what could not be assigned back |

The last two are the substantive ones. A spread of a generic uiSchema drops the per-field half of the type — TypeScript discards a deferred mapped type when it flattens an object spread — and `Omit` of one does the same. `Object.assign`'s return type is *derived* from its arguments and checked against the annotation, so the type is verified rather than asserted; that is the difference from the `as` it replaces. It trips `prefer-object-spread`, suppressed at those four lines with that reason in the comment.

`getDaisy()` lost its assertion by dropping the `{}` default entirely and optional-chaining instead — a line shorter than before.

Lint, knip, root typecheck, `cs-check`, build and all 15 test targets pass.

heath-freenome:
Code review found two correctness issues in the new `UiSchema<T>` mapped type in `packages/utils/src/types.ts` (verified by compiling minimal repros against the repo's own `tsc`):

1. **Reserved-key collision (~line 1222):** The fixed/reserved `UiSchema` keys (`items`, `additionalProperties`, `additionalItems`, `anyOf`, `oneOf`, `classNames`) collide with a same-named field in the caller's own form-data type, since the mapped per-field entry and the fixed literal property are intersected on the same key.

   Example: given `interface FormData { classNames: { base: string } }` (or more commonly a field literally named `items`, e.g. `{ items: OrderItem[] }`), `const ui: UiSchema<FormData> = { classNames: { base: {...} } }` fails to compile with TS2322 — the mapped part wants `UiSchema<{base:string}>` for that key while the fixed literal `classNames?: string` forces `& string`, so no object value can ever satisfy it. The migration guide only lists three fix-ups (index signature, declared positional lists, `interface`-typed uiSchema) and doesn't mention this collision, and `packages/utils/test/uiSchemaTypes.test.ts` never exercises a data field named `items`/`classNames`/etc., so this shipped untested.

2. **`additionalProperties` typed against the wrong type (~line 1247):** `additionalProperties?: UiSchema<T, S, F>` types the nested uiSchema for a schema's `additionalProperties` value against the parent object's own data type `T`, instead of the additional property's own (unrelated) value type — unlike `items`/`additionalItems`, which correctly use `ArrayElement<T>` for the element type.

   Example: a schema `{ properties: { name: string }, additionalProperties: { type: 'object', properties: { city: string } } }` with `T = { name: string }` (no index signature) rejects the valid `uiSchema.additionalProperties = { city: {...} }` since `city` isn't a key of `T`, while the nonsensical `uiSchema.additionalProperties = { name: {...} }` type-checks fine. The field-name checking this PR adds is effectively backwards for `additionalProperties`.

jimmycallin:
Both checked against `tsc`; one is fixed, the other is a runtime limitation the type now makes visible, and the guide says so. Pushed as `f2d9fb70e`.

**2. `additionalProperties` typed against the parent** — confirmed and fixed. It is now `UiSchema<AdditionalPropertyData<T>>`, where that helper is `T`'s index-signature value type when it has one and `any` when it does not. So `UiSchema<{ name: string }>` accepts `additionalProperties: { city: {…} }` and rejects nothing under it, while `UiSchema<Record<string, { city: string }>>` accepts `city` and rejects `zip`. Both are pinned in `uiSchemaTypes.test.ts`. `ObjectField` now reads the entry through `getByPath` with the same type argument as every other per-property uiSchema, so no new cast.

**1. Reserved-key collision** — partly confirmed, and I am leaving the type as it is.

- `items`, `additionalItems`, `additionalProperties` do not collide in practice: the declared key holds a uiSchema, so the intersection of the per-field entry with it accepts the entry. The existing test already exercises a data field named `items` (`items: { label: string }[]` with `items: { items: { label: {…} } }`) and it compiles.
- `classNames`, `anyOf`, `oneOf` do collide: a string or a positional list has nothing in common with a uiSchema, so a data field with one of those names cannot get a per-field entry. That mirrors runtime behavior rather than adding a restriction: `SchemaField` deletes `classNames` from every uiSchema it receives before handing it to the field, so an object's child entry under that name never reaches `ObjectField`; and `MultiSchemaField` and `optionsList` read `anyOf`/`oneOf` as positional lists wherever the schema has the keyword. A per-field entry under those names was never reliably delivered, so a type that accepted it would be lying.

Making the declared keys yield to a same-named data field would require conditional types on the fixed part, and core reads `uiSchema.items`, `uiSchema.additionalItems`, `uiSchema[ONE_OF_KEY]` on a generic `T`, where a deferred conditional would need casts at every one of those sites. Not worth it for names that do not work at runtime anyway. The guide's "things to fix" list gains a bullet naming the three keys, why they are limited, and that `items`/`additionalItems`/`additionalProperties` as field names are fine.

Root typecheck, uncached lint, knip, `cs-check`, the `@rjsf/utils` and `@rjsf/core` suites pass.

heath-freenome:
Re-review turned up two confirmed high-severity issues plus a few lower-severity cleanups.

## High severity

1. **Runtime crash regression** (`packages/core/src/components/fields/SchemaField.tsx` ~line 207): the new `consumesStyling` check applies the `in` operator directly to `uiSchema` and `uiSchema['ui:options']`, which throws a `TypeError` instead of returning `false` when either is a non-null, non-object truthy value (e.g. a string) — a regression from the old destructuring-based strip, which never threw for that input.

   Reproduced: rendering `<Form schema={{type:'object',properties:{foo:{type:'string'}}}} uiSchema={{'ui:options':'oops'}} validator={validator} />` throws an uncaught `TypeError: Cannot use 'in' operator to search for 'classNames' in oops` on the PR's HEAD commit, and renders cleanly on the pre-PR base commit. The same crash is reachable via a nested field's own uiSchema being a non-object truthy value (e.g. `uiSchema: { foo: 'oops' }`), since `resolveUiSchema()`'s `localUiSchema || {}` fallback only replaces falsy values, so `'ui:classNames' in uiSchema` throws directly on the string.

2. **Type-level regression for union fields** (`packages/utils/src/types.ts` ~line 394): `UiSchemaChildData`/`UnionMembersMerged` don't exclude array members when merging a union field's keys, so a form-data field typed as a union of an array and a plain object pulls in `Array.prototype` method names as merged keys, causing legitimate `UiSchema<T>` literals for that field to fail to typecheck with a confusing, unrelated error.

   Confirmed via `tsc --build` on the PR's HEAD: for `interface Data { thing: string[] | { foo: string } }`, `const ui: UiSchema<Data> = { thing: { foo: { 'ui:title': 'Foo' } } }` fails with `TS2322` about `toLocaleString` being incompatible with the `ui:${string}` index signature. A plain object-only union compiles fine, isolating the array member's inherited method names as the trigger. Pre-PR, such a field's uiSchema was accepted unconditionally as `any`.

## Lower severity

3. **Shared object reference** (`SchemaField.tsx` ~line 212): when `consumesStyling` is false, `fieldUiSchema` is now the exact same object reference as `uiSchema` (no defensive copy), whereas the old code always produced a fresh object. A custom/third-party Field or Widget that mutates `props.uiSchema['ui:options']` in place (instead of spreading a new object) previously only corrupted its own subtree's copy; now the mutation leaks to sibling consumers of the same object within `SchemaFieldRender`.

4. **Duplicated fix logic** (`packages/antd/src/templates/WrapIfAdditionalTemplate/index.tsx` ~line 37): the `baseUiSchema = uiSchema ?? {}` plus spread-with-`ui:options`-override fix is duplicated verbatim in `packages/mantine/src/templates/WrapIfAdditionalTemplate.tsx` instead of being extracted into a shared `@rjsf/utils` helper.

5. **Unnecessary non-null assertions** (`packages/utils/test/optionsList.test.ts` ~line 220, and `resolveUiSchema.test.ts`): eight `.oneOf![0]`/`.anyOf![index]`-style assertions were added to work around the new array-typed `oneOf`/`anyOf`, instead of typing the local fixture once so TS narrows without `!`.

6. **Redundant type instantiation** (`packages/daisyui/src/utils.ts` ~line 5): `DaisyUiSchema` intersects the full `UiSchema<T,S,F>` with a redeclared `ui:options`, while `DaisyUiOptions` separately re-derives `UiSchema<T,S,F>['ui:options']` for the same override — `UiSchema<T,S,F>` is instantiated twice where the prior `Omit<UiSchema<...>, 'ui:options'>` computed it once.

jimmycallin:
Fixed in `f58be2cd1` — both high-severity findings and one of the three cleanups; the other two I'm leaving, with reasons.

## High severity

**1. `in` on a non-object** — confirmed, and reproduced exactly as you describe: `uiSchema={{'ui:options': 'oops'}}` throws `TypeError: Cannot use 'in' operator to search for 'classNames' in oops`. The `!!consumedUiOptions` guard I wrote only excluded falsy values, which is the same mistake `resolveUiSchema()`'s `localUiSchema || {}` makes for a nested field's own uiSchema.

Both checks now go through `isObject()`, which is already imported in this file for the same reason. Fixing `resolveUiSchema()` instead would only have covered the nested case — the form-level `ui:options` one reaches the `in` through a perfectly ordinary object — so the guard belongs where the `in` is.

A test in `packages/core/test/uiSchema.test.tsx` renders both shapes; it fails on the previous commit with that exact TypeError.

**2. Array members in a union field's merged keys** — confirmed. `string[] | { foo: string }` isn't wholly an array, so it skipped the array branch, and `UnionMembersMerged` then took `keyof string[]` — `length`, `toLocaleString`, the lot. `UnionMembersMerged` is now applied to `Exclude<NonNullable<T>, readonly unknown[]>`, which makes a union's array member behave the way a lone array member already did: it contributes no field names, because an array nests through `items` rather than by key. Three cases pinned in `uiSchemaTypes.test.ts` — the object branch's `foo` accepted, `items` accepted, `toLocaleString` rejected as `@ts-expect-error`.

## Lower severity

**5. Non-null assertions in tests** — fixed, by removing the need for them rather than by adding an annotation. In `resolveUiSchema.test.ts` the four sites now assert on `result.oneOf` as a whole (`expect(result.oneOf).toEqual([...])`), which also subsumes the separate `toHaveLength` and `toBeUndefined` assertions — eight lines become four, and nothing is asserted about a value TypeScript can't see. In `optionsList.test.ts` the two fixtures use `satisfies UiSchema` instead of an annotation, so they keep being checked as uiSchemas while `.anyOf[index]` stays a known array. The remaining `!` in that file are on `RJSFSchema.oneOf`/`anyOf` and predate this PR.

**3. Shared `uiSchema` reference** — accurate, and deliberate. The copy exists only to strip consumed styling keys, and most uiSchemas carry none, so copying every field's uiSchema on every render to defend against a component that mutates its own props buys nothing real: React props are not a mutable buffer, no field or widget in this repo or any theme writes to `props.uiSchema`, and one that did was already corrupting the object its parent holds. The comment above the branch says this is why the copy is conditional.

**4. Duplicated `baseUiSchema` in antd and mantine** — leaving it. It's four lines in two files, and the only two places in the repo that override a `ui:options` entry this way. A shared `@rjsf/utils` export has to be named, documented, tested and supported forever; that's a larger and more permanent thing than the duplication it removes. Worth revisiting if a third theme needs it.

**6. `DaisyUiSchema` instantiating `UiSchema` twice** — leaving it. TypeScript memoizes an instantiation of the same generic with the same arguments, so the second reference resolves to the cached one rather than re-computing the mapped type; there's no measurable instantiation cost to remove here. The intersection replaced `Omit<UiSchema<…>, 'ui:options'>` because `Omit` flattens the per-field mapped type away, which is the bug this PR is about — writing it once through a local alias wouldn't change either the cost or the shape.

Root typecheck, uncached lint, knip, `cs-check`, and the `@rjsf/utils` (1508) and `@rjsf/core` (1994) suites pass.

jimmycallin:
## Self-review pass on the new `UiSchema<T>` mapped type — four more defects, fixed in `a4b951643`

The `in`-operator crash was a symptom of a general shape (an operation total for objects, partial for everything else), so I went looking for the rest of it, and separately probed the mapped type against data shapes the tests don't cover. Every case below was confirmed by compiling it against this branch's `tsc`, and every one now has a test that fails on the previous commit.

### 1. A union of a primitive and an object was typed as a leaf — the worst of these

```ts
type Data = { thing: string | { city: string } };
const ui: UiSchema<Data> = { thing: { city: { 'ui:title': 'City' } } };
// TS2353: 'city' does not exist in type 'UiSchema<string | { city: string; }>'
```

`[NonNullable<T>] extends [object]` is a whole-union test, so a union containing any primitive fails it and falls through to "no children". `oneOf: [{ type: 'string' }, { type: 'object', … }]` is about as common as `oneOf` gets, and its uiSchema was rejected outright.

This is the same defect round 1 of #5278 found in `ErrorSchema` (`ErrorTreeChildData` testing `NonNullable<V>` instead of a naked parameter). It was fixed there and shipped unfixed here — the existing test only covers a union of two *objects*, which passes either way.

### 2. `UiSchema<{ when: Date }>` could not be written at all

```ts
const ui: UiSchema<{ when: Date }> = { when: { 'ui:widget': 'alt-date' } };
// TS2322: Types of property 'toString' are incompatible.
//   Type '() => string' is not assignable to type 'UiSchema<() => string>'
```

Not over-permissiveness — a hard compile break on the ordinary case. `keyof Date` is method names, so the mapped type declared `toString?: UiSchema<() => string>`, and an object literal's *apparent* type carries `Object.prototype.toString`, which can never satisfy that. Any form typed with a `Date` or `File` field — i.e. anything using `AltDateWidget` or `FileWidget` — fails to compile. The test `Data` interface has neither, so nothing caught it.

### 3. A field typed `unknown` was a leaf

Same gap round 5 of #5278 found for `ErrorSchema`: `unknown extends object` is false, so it fell past the object branch while `any` was short-circuited explicitly.

### 4. A tuple's `items` entry was rejected

`ArrayElement<[string, { city: string }]>` is `string | { city: string }` — defect 1 again, reached by a different route.

### The fix

All four are one decision made in one place. `UiSchemaChildData` no longer asks whether the whole of `T` is an object; it *selects* the members that can hold fields and merges them:

```ts
type UiSchemaFieldMembers<T> = Exclude<Extract<NonNullable<T>, object>, readonly unknown[] | AtomicUiValue>;
type UiSchemaChildData<T> = unknown extends T ? GenericObjectType : UnionMembersMerged<UiSchemaFieldMembers<T>>;
```

`Extract<…, object>` drops primitive members rather than letting them through (`keyof string` is `String.prototype`'s methods — the same leak you found for arrays), `Exclude` drops arrays and atomic values, and `unknown extends T` covers both `any` and `unknown`. `IsUnion` is gone: merging a one-member union is the identity, so the special case earned nothing.

Method-valued keys are now dropped from the merge too, which is what actually fixes defect 2 generally rather than for `Date` alone:

```ts
type FieldKeys<X> = { [K in keyof X]-?: NonNullable<X[K]> extends (...args: never[]) => unknown ? never : K }[keyof X];
```

A method is not a form field, so there is nothing to put a uiSchema on — and mapping one poisons assignability for every object literal. `Map`, a custom value object, anything whose API is methods: all safe now.

Five cases pinned in `uiSchemaTypes.test.ts` (mixed union, tuple `items`, `Date`/`File` as leaves with `getTime` rejected as `@ts-expect-error`, `unknown`); all five fail to compile on the previous commit.

### Runtime robustness, the rest of the `in` finding

Two more places, from grepping the diff for partial operations on a value the types never validated:

- **`resolveUiSchema()`** — `localUiSchema || {}` passed a non-object straight through, which is how a string reached `SchemaField` in the first place. It now normalizes to `{}`, so a value a caller put under a field's name cannot reach any field, widget or template that reasonably expects an object. Four cases tested; they fail on the previous commit.
- **`SchemaField`** — the `ui:options` destructure was still guarded by `if (consumedUiOptions)` rather than `isObject()`, so a truthy non-object was spread into an indexed object (`{0:'o',1:'o',…}`) instead of being ignored. Guarded to match the check above it.

The crash test is now a table over `'oops'`, `42`, `true`, `['oops']` and `null` at four positions (root `ui:options`, a field's own uiSchema, a field's own `ui:options`, `ui:globalOptions`) — 20 cases. Worth noting what that turned up: **two of them throw on `v7` itself**, before this PR. `uiSchema: { 'ui:options': null }` gives `TypeError: Cannot destructure property 'classNames' of 'fieldUiSchema[UI_OPTIONS_KEY]' as it is null` on the base branch, because the old code destructured through a `!`. So this PR fixed one pre-existing crash while introducing a different one; the table covers both classes now.

Root typecheck, uncached lint, knip, `cs-check` and all 15 test projects pass.

heath-freenome:
Re-reviewed with attention to the previous rounds of feedback on this PR (both the automated review comments and the human review from `heath-freenome`) plus the latest fixes.

**Good news:** every previously-flagged issue was independently re-verified as genuinely fixed as of the current head commit — the reserved-key collision, the `additionalProperties` mistyping, the `in`-operator `TypeError` on non-object uiSchema/`ui:options`, the array-method-name leak in union merging, and the primitive+object union/`Date`/`File`/tuple handling. These were re-confirmed by compiling `tsc` repros and rendering against the actual PR HEAD, not just reading the diff.

However, the fix pattern introduced a few new issues in its current state:

## Confirmed

1. **`WrapIfAdditionalTemplate` still has the non-object hazard it was supposed to fix** (`packages/antd/src/templates/WrapIfAdditionalTemplate/index.tsx` ~line 70, duplicated in `packages/mantine/src/templates/WrapIfAdditionalTemplate.tsx` ~line 55): spreading `baseUiSchema[UI_OPTIONS_KEY]` assumes it's an object, but nothing upstream validates a field's `ui:options` value — `resolveUiSchema` only normalizes the top-level uiSchema, not the `ui:options` sub-key. If an additional-property field's uiSchema is `{ 'ui:options': 'oops' }`, `{ ...'oops', block: true }` silently spreads the string's characters as numeric keys instead of producing `{ block: true }`.

2. **`LayoutGridField.tsx` has the same gap** (~line 164): `computeFieldUiSchema` reads a nested field's uiSchema via `getByPath` and spreads it without the `isObject` normalization this same PR added to `resolveUiSchema.ts` and `SchemaField.tsx` for the equivalent read. A `ui:layout` grid grouping a field whose own uiSchema entry is a non-object (e.g. `uiSchema: { someField: 'oops' }`) spreads the string the same way.

3. **`FieldKeys` still leaks non-function intrinsic properties** (`packages/utils/src/types.ts` ~line 1229): it only excludes function-valued keys, so a property like `Map.prototype.size` or `RegExp.prototype.source` still type-checks as a valid uiSchema field name. `UiSchema<{ counts: Map<string, number> }>` accepts `{ counts: { size: { 'ui:title': 'x' } } }` with no error — `AtomicUiValue` is hardcoded to `Date | File` only, so this contradicts the PR's own self-review claim that `Map` is now handled safely.

## Plausible / lower confidence

4. **`SchemaField.tsx` ~line 213**: `fieldUiSchema` can now alias the exact same object `resolveUiSchema()` returned instead of always being cloned, removing the previous invariant that every field got a private copy. Not exploitable today (no current mutator of a received uiSchema), but a latent regression if any future field/widget mutates its uiSchema prop in place.

5. **`ObjectField.tsx` ~line 426**: switching to `getByPath(uiSchema, propertyName)` means a schema property literally named `toString`/`constructor`/etc. now resolves to `undefined` instead of the inherited `Object.prototype` member the old bracket-index code returned. Likely a desirable side effect, but it's an undocumented, untested behavior change on the same line the PR touched for type reasons.

## Minor

6. **`SchemaField.tsx` ~line 210**: the `isObject(uiSchema)` half of the `consumesStyling` check is unreachable dead code, since `uiSchema` here is always `resolveUiSchema()`'s return value (guaranteed to be an object). The adjacent comment claiming neither is guaranteed to be an object could mislead a future editor into trusting this as a safety net.

7. **Duplicated fix logic**: the `baseUiSchema`/`buttonUiOptions` fix in `WrapIfAdditionalTemplate` is duplicated byte-for-byte across antd and mantine instead of being extracted to a shared `@rjsf/utils` helper, per this repo's own guidance to generalize duplicated theme-package logic.

jimmycallin:
Addressed in `59c33011f`. Items 1, 2 and 6 are fixed; 3, 4, 5 and 7 I'm leaving, with reasons.

## Fixed

**1 and 2. Spreading a non-object `ui:options`** — accurate, but worth noting both sites are unchanged from `v7`: the antd/mantine spread and the `computeFieldUiSchema` spread are byte-for-byte the base branch's, and spread never throws, so neither is a regression of this PR. Fixed anyway, at the root rather than per site: `resolveUiSchema()` now drops a non-object `ui:options` the same way it already replaces a non-object field uiSchema. Every field, widget and template receives its uiSchema through that function, so antd, mantine and `SchemaField` all inherit the guarantee without their own guards. Four cases pinned in `resolveUiSchema.test.ts`, including that the caller's object is not mutated.

`LayoutGridField` is the one reader that runs *before* resolution — it indexes the caller's raw uiSchema by field name — so it gets its own `isObject()` checks on both the field entry and its `ui:options`. Six cases in `LayoutGridField.test.tsx`.

**6. Dead `isObject(uiSchema)` in `SchemaField`** — confirmed and removed, along with the `isObject(consumedUiOptions)` halves that the `resolveUiSchema()` change makes equally unreachable. The comment now says where the guarantee comes from instead of claiming there is none. The 20-case crash table still passes, so the guarantee is tested where it's relied on.

## Leaving

**3. `FieldKeys` accepts `Map.prototype.size`** — accurate, and not a defect I'll chase. It is over-permissive only: no valid uiSchema is rejected, which is what my earlier claim about `Map` was about (its *methods* no longer poison assignability). Form data is JSON, so a field typed as a `Map` or `RegExp` is not a shape anyone writes, and enumerating every built-in with a non-method member is a list that can only ever be incomplete. `AtomicUiValue` names the two non-JSON types the widgets actually produce.

**5. `getByPath` in `ObjectField` for prototype-named properties** — accurate and desirable, as you say. Not adding a test: a schema property named `toString` never had a working uiSchema entry (the bare index returned a function), so there is no prior behavior to preserve or migration to document.

**4 and 7** — unchanged from the previous round; the reasoning there still holds.

Root typecheck, uncached lint, knip, `cs-check`, and the `@rjsf/utils` (1528) and `@rjsf/core` (2019) suites pass.

packages/utils/src/types.ts
```diff
@@ -1256,15 +1256,48 @@ export type UiSchemaDefinitions<
   F extends FormContextType = any,
 > = Record<string, UiSchema<T, S, F>>;
 
+/** True when `V` is `any`, which a conditional type otherwise matches on every branch at once */
+type IsAnyType<V> = 0 extends 1 & V ? true : false;
```
heath-freenome: Isn't this the same as `IsAny<>` from the other PR? Did you want to remove it?

jimmycallin: It is; `IsAny` came in with #5278, which this branch now sits on. Removed the duplicate in `414a4977b`, and `AtomicUiValue` with it, since `AtomicValue` is the same `Date | File`. Both trees share the two helpers now.

heath-freenome:
Re-reviewed again against the latest fixes. All previously flagged issues remain fixed. One new issue found:

**`packages/utils/src/resolveUiSchema.ts` ~line 48**: `resolveUiSchema()` normalizes a non-object local `uiSchema` and a non-object `ui:options`, but not a non-object `registry.uiSchemaDefinitions[ref]` value — so `result = { ...definitionUiSchema }` still spreads a malformed `ui:definitions` entry's characters as bogus numeric keys instead of treating it as absent.

Repro: a caller sets `uiSchema: { 'ui:definitions': { '#/$defs/Foo': 'oops' } }` and a schema resolves `$ref: '#/$defs/Foo'` with no local override. `definitionUiSchema` is the truthy string `'oops'`, `localUiSchema` is `undefined`, so the `else if` branch runs `{ ...'oops' }`, producing `{0:'o',1:'o',2:'p',3:'s'}` instead of `{}`. That corrupted uiSchema is then handed to SchemaField/ObjectField/getUiOptions for the whole ref'd subtree, silently dropping every intended `ui:definitions` customization for that path.

Confirmed via `node -e "console.log(JSON.stringify({...'oops'}))"` → `{"0":"o","1":"o","2":"p","3":"s"}`. This line is unchanged from the pre-PR base, so the self-review claim that "every field, widget and template receives its uiSchema through that function, so ... all inherit the guarantee without their own guards" isn't fully true — the `mergeObjects()`-based branch (when a local override exists) is protected by #5281's non-object-left-hand fix, but this direct-spread branch is not.

heath-freenome (approved):
@jimmycallin approving pending that one new fix with test passes and your own claude code review is in the clear

jimmycallin:
Two of the three confirmed findings were already fixed on the head you reviewed against's successor — `59c33011f` landed about 45 minutes after that comment — and the third is real and fixed in `60cb6de68`. Details per item.

## Confirmed

**1. `WrapIfAdditionalTemplate` spreading a non-object `ui:options`** — fixed in `59c33011f`, upstream of the template. `resolveUiSchema()` now deletes a non-object `ui:options` from the uiSchema it returns, and every field's uiSchema reaches a template through `SchemaField` → `resolveUiSchema()`, so the spread in antd and mantine sees an object or nothing. The templates themselves still assume an object, deliberately: the invariant is established once where uiSchemas are normalized rather than re-checked in each of the dozen places that read `ui:options`. Covered by the non-object cases in `resolveUiSchema.test.ts` and the 20-case table in `core/test/uiSchema.test.tsx`, which includes a non-object `ui:options` on a field's own uiSchema.

**2. `LayoutGridField.computeFieldUiSchema`** — fixed in the same commit. It reads the caller's uiSchema *before* `resolveUiSchema()` has seen it, so both the field's entry and its `ui:options` are now `isObject()`-guarded before the spread, with a comment saying why this site guards where others don't. Table cases in `LayoutGridField.test.tsx`, failing on the previous commit.

**3. `FieldKeys` accepting `Map.size`** — confirmed, and my self-review claim that `Map` was "safe now" was wrong: dropping method keys handles `get`/`set`, but `size` is a property. Fixed in `60cb6de68` by extending `AtomicValue`, which `UiSchema` and `ErrorSchema` have shared since `414a4977b`, to the built-in classes with non-method properties: `Blob`, `RegExp`, `URL`, `Promise`, `Map`, `Set`, `WeakMap`, `WeakSet`. Both trees treat them as leaves now. A user-defined class instance still gets its properties offered as field names; TypeScript cannot express "plain object", so there is no general rule to write, and the comment on `AtomicValue` states the limit. Pinned in both `uiSchemaTypes.test.ts` and `errorSchemaTypes.test.ts` (`counts: Map<string, number>` accepted as a leaf, `counts.size` rejected with `@ts-expect-error`); both directives are flagged unused by `tsc` on the previous commit.

## Plausible

**4. `fieldUiSchema` aliasing the resolved uiSchema** — unchanged from my reply in the last round: the copy only ever existed to strip consumed styling keys, no field or widget in this repo or any theme writes to `props.uiSchema`, and one that did was already corrupting its parent's object. If the maintainers would rather keep the every-field-gets-a-copy invariant than save the allocation, it is a one-line change and I'll make it; I just don't think it buys anything real.

Root typecheck, knip, `cs-check`, and the `@rjsf/utils` suite (1529, 100% coverage) pass.

Issue #5217, opened 2026-08-27
Performance: `FieldPathId` objects break memoization — stabilize identity instead of deep-comparing


jimmycallin:
Every field receives a `fieldPathId` prop: `{ path: (string|number)[], $id: string, name? }`. It is rebuilt by `toFieldPathId()` on every render, at every level of the field tree. React's memoization (`React.memo`, `shouldComponentUpdate`) compares props by *reference*, so a freshly-built-but-equal object looks like a change, and the memo boundary silently stops working.

Today we compensate downstream of the problem:

- `useDeepCompareMemo(toFieldPathId(...))` at six call sites in core (ArrayField ×3, ObjectField, LayoutGridField, FallbackField) to launder the unstable reference;
- `deepEquals` in `SchemaField`'s memo comparator and `shouldRender`.

This works, but it's the pattern the React docs explicitly warn against: custom comparators doing deep equality "can become incredibly slow and can freeze your app" ([react.dev/reference/react/memo](https://react.dev/reference/react/memo)), and the cost is paid on *every* render at *every* boundary. It also can't be fixed with `useMemo`: the instability cascades (each field's id depends on the parent's unstable id object), `Form` is a class component, and `LayoutGridField` creates ids inside loops where hooks can't go. `useMemo` is also documented as a discardable cache, "not a semantic guarantee."

The root cause is a data-modeling choice: **field identity is represented as an object recreated during render, while consumers rely on reference identity.**

## How other libraries solved this

**Every major React form library identifies a field by a primitive string path** — `'friends[0].firstName'` — not an object. react-hook-form (`_fields` keyed by name string, `get(_fields, name)`), Formik (`getIn`/`setIn` over dot paths), TanStack Form (`name` string into an external store), React Final Form (`registerField(name: string, ...)`). A string's identity *is* its value, so memo boundaries, map keys, and subscription keys are stable for free. This is also react.dev's first-listed advice for memo: "prefer accepting individual values as props."

The performant ones go one step further and **invert data flow**: state lives outside React and fields *subscribe* by their name string (react-hook-form's pub-sub subjects + uncontrolled inputs, Final Form's subscription model, TanStack's store selectors), so one keystroke re-renders one field instead of the tree. The lone library architected like RJSF — Formik, prop/context-driven with deep-compare compensation (`FastField`) — is the ecosystem's documented performance cautionary tale (formik#1974, formik#2335).

[...]

PR #5280, opened 2026-09-12
feat!: identify fields by a FieldPath string and an id instead of the fieldPathId object (#5217, part 1)

jimmycallin:

This is the first half of #5217, targeting `v7`. It changes **how a field knows who it is**. The second half, which removes the deep prop comparison this change makes unnecessary, is a separate stacked PR so each can be reviewed on its own.

## The problem, from the top

Every field, widget and template in RJSF receives a `fieldPathId` prop today:

```ts
fieldPathId: { $id: 'root_friends_0_firstName', path: ['friends', 0, 'firstName'], name?: string }
```

It is built by `toFieldPathId()` on every render, at every level of the field tree. That has two consequences.

**1. It defeats React's memoization.** `React.memo` decides whether to skip a re-render by comparing each prop with `===`. Two objects are only `===` if they are the *same* object. A `fieldPathId` rebuilt on every render is a new object every time, so to `React.memo` it always looks changed, even when nothing about the field moved. Every memo boundary below it silently stops working.

Here is the shape of the problem in isolation:

```ts
const a = { $id: 'root_name', path: ['name'] };
const b = { $id: 'root_name', path: ['name'] };
a === b; // false, so memo re-renders
'name' === 'name'; // true, so memo skips
```

RJSF compensates for this downstream of the cause: `useDeepCompareMemo(toFieldPathId(...))` at six call sites launders the unstable reference back into a stable one, and `SchemaField`'s memo comparator runs `deepEquals` over **all** props. The React docs [warn against exactly that](https://react.dev/reference/react/memo#specifying-a-custom-comparison-function): "deep equality checks can become incredibly slow and can freeze your app".

**2. It doesn't always tell the truth.** For `oneOf`/`anyOf`, RJSF renders two components for the *same* data address: the shared-properties field and the selected-option field. They can't share a DOM id, so a fake `'XxxOf'` segment was appended to `fieldPathId.path` purely to make the id unique. That means `path` sometimes pointed somewhere the data doesn't live, and `ObjectField`/`ArrayField` carried a compensating `childFieldPathId` prop to remember the real address for writes.

## The change

A field is identified by two strings instead of one object:

```diff
-fieldPathId: { $id: 'root_friends_0_firstName', path: ['friends', 0, 'firstName'] }
+fieldPath: 'friends[0].firstName'   // WHERE the data lives. Drives onChange writes.
+id:        'root_friends_0_firstName' // WHAT the DOM element is called. Drives htmlFor, focus, a11y.
```

A string's identity is its value, so both props are reference-stable across renders with no memoization at all. This is [react.dev's first piece of advice](https://react.dev/reference/react/memo#minimizing-props-changes) for making memo effective: "accept individual values instead of a whole object".

### Why two props and not one?

Usually `id` is derivable from `fieldPath`. The `oneOf`/`anyOf` case above is where they diverge: the two components rendering the same data get the same `fieldPath` but different `id`s. The disambiguation now lives in the only place it means anything, the DOM id, and `fieldPath` is always a truthful data address. The `childFieldPathId` compensation prop disappears because nothing needs it anymore.

Ids are byte-identical to v6, including the `XxxOf` suffixes. Every theme's snapshot suite passes with zero updates.

### The `FieldPath` grammar

```
''                          the root form (exported as ROOT_FIELD_PATH)
friends                     property
friends[0]                  array element
friends[0].firstName        nested property
a\.b                        a property literally named "a.b"
```

Property names are separated by `.`, array indexes are bracketed, and `\ . [ ]` inside a property name are backslash-escaped. That last rule is what makes the grammar unambiguous. The old `idSeparator` join collides the moment a property name contains the separator; this doesn't.

The index/key distinction survives a round trip. `toFieldPath(0, 'a')` gives `a[0]` and parses back to `['a', 0]`; `toFieldPath('0', 'a')` gives `a.0` and parses back to `['a', '0']`. That matters because `setByPath` treats a number as an array index and a string as an object key.

### `FieldPath` is a branded string

```ts
type FieldPath = string & { readonly [FIELD_PATH_BRAND]: true };
```

A DOM id is also a string. With a bare `string` alias, `fieldPathToId(id, ...)` or `onChange(value, id)` would typecheck and silently produce garbage. The brand makes the compiler reject a plain string where a `FieldPath` is expected, at zero runtime cost. Paths are constructed with `toFieldPath()` or start from `ROOT_FIELD_PATH`, and a `FieldPath` is still assignable anywhere a `string` is accepted, so `<label htmlFor={fieldPath}>` would compile if you wanted it to.

The only `as FieldPath` casts in the codebase are inside `fieldPath.ts`. Everywhere else gets a `FieldPath` by calling into it.

### `onChange` carries the string

```diff
-onChange: (newValue: T | undefined, path: FieldPathList, es?: ErrorSchema<T>, id?: string) => void;
+onChange: (newValue: T | undefined, fieldPath: FieldPath, es?: ErrorSchema<T>, id?: string) => void;
```

Fields forward the `fieldPath` prop they already hold. The string is parsed back into segments exactly once, in `Form.processPendingChange()`, where `setByPath` needs a list. Nothing in the field tree parses paths.

### The new API in `@rjsf/utils`

| Function | Purpose | Replaces |
| --- | --- | --- |
| `toFieldPath(segment, parentPath?)` | Append a property name or array index. One string concat. | `toFieldPathId(name, options, parent)` |
| `fieldPathToList(path)` | Parse back to `FieldPathList`, indexes as numbers. | `fieldPathId.path` |
| `fieldPathToId(path, globalFormOptions)` | Derive the DOM id in a single pass, no intermediate list. | `fieldPathId.$id` |
| `fieldPathToName(path, globalFormOptions, isMultiValue?)` | Derive the HTML `name`. | `fieldPathId.name` |
| `fieldPathEndsWithIndex(path)` | Does the path address an array element? | ad-hoc `path.at(-1)` checks |

The grammar lives in one file. Consumers that need a grammar fact, such as `ArrayField` asking "is this an array item?", go through a helper instead of re-encoding parser knowledge locally.

### On the name-generation concern from the issue

@heath-freenome pointed out that `fieldPathId` carries information that affects name generation. It does, and all of it is preserved. The one piece not derivable from the path, the multi-value flag `ArrayField` baked in at construction time, becomes an argument at the point of use: `fieldPathToName(fieldPath, globalFormOptions, isMultiValue)`. It's a local fact where it's needed instead of state carried down the tree. All `nameGenerator` snapshot tests pass unchanged.

### On keeping a deprecated `FieldPathId` alias

The issue thread floated keeping `FieldPathId` as a deprecated alias for pure pass-through components. I chose not to. An alias from an object type to a string wouldn't help the consumers that actually break, which are the ones reading `.$id` or `.path`, and pure pass-throughs already compile because `fieldPath` and `id` are ordinary props. Happy to add it if you'd still like the softer landing.

## How to migrate

For the vast majority of custom templates it's a rename:

```diff
-const { fieldPathId } = props;
-<label htmlFor={fieldPathId.$id}>
-<h2 id={titleId(fieldPathId)}>
+const { id } = props;
+<label htmlFor={id}>
+<h2 id={titleId(id)}>
```

For custom fields that call `onChange`:

```diff
-onChange(newValue, fieldPathId.path, errorSchema, fieldPathId.$id);
+onChange(newValue, fieldPath, errorSchema, id);
```

The v7 migration guide has the full table. `FieldProps` and `FieldTemplateProps` receive both `fieldPath` and `id`. Every other template that received `fieldPathId` (`ArrayFieldTemplate`, `ArrayFieldTitleTemplate`, `ArrayFieldDescriptionTemplate`, `ArrayFieldItemButtonsTemplate`, `CyclicSchemaExpandTemplate`, `FieldErrorTemplate`, `FieldHelpTemplate`, `ObjectFieldTemplate`, `UnsupportedFieldTemplate`) receives `id`. The id helpers (`titleId()`, `descriptionId()`, `helpId()`, `errorId()`, `examplesId()`, `buttonId()`, `optionalControlsId()`, `ariaDescribedByIds()`) accept only a `string`.

## Breaking changes

| Change | Who is affected |
| --- | --- |
| `fieldPathId` prop replaced by `fieldPath` + `id` on fields and field templates, by `id` on other templates | Every custom field, widget and template that read `fieldPathId` |
| `FieldProps['onChange']` second argument is a `FieldPath` string, not a `FieldPathList` | Custom fields that call `onChange` with a path |
| `toFieldPathId()` and the `FieldPathId` type removed | Code that built ids manually |
| `childFieldPathId` prop removed | Nobody outside core should have depended on it |
| Id helpers accept only `string` | Callers that passed the object; pass `id` instead |
| `Form.setFieldValue('')` yields id `root` instead of `root_` | Disclosed bug fix; there is a test |

## How other libraries model this

Every major React form library identifies a field by a dot-and-bracket **name string**, not an object: react-hook-form, React Final Form, TanStack Form and Formik all do. Three of the four pair it with an external store and per-field subscriptions. Formik doesn't, and its prop-driven rendering with `FastField`'s deep-ish `shouldComponentUpdate` is the ecosystem's documented performance cautionary tale ([formik#1974](https://github.com/jaredpalmer/formik/issues/1974), [formik#2335](https://github.com/jaredpalmer/formik/issues/2335)). That's why this PR is honest about being *necessary but not sufficient*: stable identity is the prerequisite; deleting the deep compare is the follow-up PR.

Schema-driven libraries chose differently and worse for our needs. JSONForms mints ids from a global counter registry, nondeterministic across mount order and a [measured performance problem](https://jsonforms.discourse.group/t/expensive-id-generation-mechanism-with-createid-and-isuniqueid/1737). uniforms uses random `uniforms-X` prefixes, hostile to snapshot tests and deep links. RJSF's deterministic path-derived ids are the nicest scheme in this group. This PR keeps them and only moves where the disambiguation lives.

[SJSF](https://github.com/x0k/svelte-jsonschema-form) is the closest relative of the schema-driven group. Identity there is a branded path, `FieldPath = Brand<"sjsf-path", Readonly<Array<string | number>>>`, and the DOM id is derived from it by a pluggable `FormIdBuilder.fromPath()` — `idPrefix` plus `_`-joined segments, the same deterministic path-derived scheme RJSF uses. Two differences are instructive. It keeps identity as an array rather than a string, which costs it nothing because Svelte's reactivity doesn't compare props by reference: the array-versus-string choice is forced on us by React memo, not by the problem domain. And what we spell `XxxOf` it makes collision-proof — `oneof`/`anyof`, like `title`, `errors` and `submit`, are *pseudo-elements*, ordinary path segments encoded behind a reserved `__sjsf_pseudo_element` prefix that no real property name produces. That is the shape of the fix for the `XxxOf` id collision raised in review, for a release that is already changing ids on purpose.

## Where to look when reviewing

1. `packages/utils/src/fieldPath.ts`: the whole grammar, five small functions. If this is right, everything else is plumbing.
2. `packages/core/src/components/fields/SchemaField.tsx` around `fieldPathProps`: the one place `id` and `fieldPath` diverge.
3. `packages/core/src/components/Form.tsx` `processPendingChange()`: the single parse point.
4. `packages/utils/test/fieldPath.test.ts`: round trips, escaping, the unclosed-bracket case from the issue review, and a property test that `fieldPathToId` matches the list-based derivation.

## Validation

Rebased on `v7` after #5266. Lint, knip, build, root typecheck, every test target, source formatting and the Docusaurus build pass. `@rjsf/utils` is at 100% coverage. Zero snapshot changes in any theme, which is the empirical check that ids are unchanged.

The `fieldPathToList` unclosed-bracket issue found in the review on #5217 is fixed and tested: a malformed `a[unclosed` is read as a property name, matching how `fieldPathToId` already treats it, instead of looping.

## Checklist

- [x] **I'm updating documentation**: `custom-templates.md`, `custom-widgets-fields.md`, `layout-grid.md`, `utility-functions.md`, the v7 migration guide
- [x] **I'm adding or updating code**: tests updated, snapshots unchanged, `CHANGELOG-v7.md` updated


heath-freenome:
## Code review findings

**CONFIRMED**

1. **`CHANGELOG-v7.md:52`** — Unresolved git merge-conflict marker committed into the changelog under the `@rjsf/daisyui` heading. Line 52 reads `||||||| parent of c459bce95 (...)` followed by an orphaned bullet — a diff3-style conflict marker never cleaned up during rebase/merge. Ships as garbage in the release notes read by every consumer upgrading to v7.

2. **`packages/utils/src/index.ts`** — Removal of the publicly-exported `useDeepCompareMemo` hook is undisclosed in breaking-changes docs. It was a documented, exported `@rjsf/utils` hook; this PR deletes it entirely but the breaking-changes table, `CHANGELOG-v7.md`, and the v7 upgrade guide say nothing about it, so any consumer importing it for custom fields/widgets hits a hard build break on upgrade with no pointer to the cause.

3. **`CLAUDE.md:123`** — Repo-root `CLAUDE.md` still lists the deleted `useDeepCompareMemo` hook as a key `@rjsf/utils` export. Should be updated now that the hook is gone.

**PLAUSIBLE**

4. **`packages/core/src/components/fields/ObjectField.tsx:409`** — `OptionalDataControlsField` may receive the `XxxOf`-suffixed id instead of the field's true id when an optional object field also has its own `oneOf`/`anyOf`. For a non-required object schema with `oneOf`/`anyOf` whose branches share type `'object'` and `enableOptionalDataFieldForType` includes `'object'`, `SchemaField` renders `ObjectField` with id `${fieldId}${idSeparator}XxxOf`. `ObjectField`'s `optionalDataControl` appears to pass this suffixed id straight through instead of resolving the true field id (as the old code did via `childFieldPathId`), which would give the Add/Remove control's DOM id an unexpected `_XxxOf` segment — breaking the "ids are byte-identical to v6" claim for this schema shape. Worth a quick test with this combination to confirm.

5. **`packages/core/src/components/fields/SchemaField.tsx:195`** — The `XxxOf`-suffixed id is built via ad-hoc string concatenation (`` `${fieldId}${globalFormOptions.idSeparator}XxxOf` ``) instead of going through the `fieldPath.ts` helper, contradicting the PR's own stated principle that the id grammar lives in one place. If `idSeparator` join semantics change later, this line could silently drift out of sync since it never calls `fieldPathToId`.

6. **`packages/utils/src/fieldPath.ts:101`** — `fieldPathToList`, `fieldPathToId`, and `fieldPathEndsWithIndex` each re-implement the same escaping/traversal grammar with no shared tokenizer. `fieldPathToId`'s hand-rolled scanner duplicates `fieldPathToList`'s backslash/`.`/`[`/`]` handling (justified only by an unbenchmarked "hot path" comment), and `fieldPathEndsWithIndex`'s regex independently assumes the same escaping invariant. Correctness across all three currently relies on a fixed-fixture property test; a future grammar extension applied to one function could silently desync the others for uncovered path shapes.

7. **`packages/core/test/Form.behaviors.test.tsx:1522`** — New comment describes removed old behavior ("root_") rather than explaining current behavior. The repo's own `CLAUDE.md` bans before/after or diff-style commentary in code comments; the added comment names the deleted old `toFieldPathId` output ("trailing-separator `root_`"), which becomes orphaned trivia once that history fades rather than explaining current behavior on its own terms.

---

jimmycallin:
All seven addressed in `37e2a2345`; one of them turned out to be a v6 bug the PR had fixed without saying so.

**1. Conflict marker** — removed. It was introduced by this branch's own conflict resolution, not inherited from `v7`, which is clean.

**2/3. `useDeepCompareMemo` removal** — disclosed in the `@rjsf/utils` changelog entry and in the guide's `FieldPath` section, with `useMemo()` as the replacement for a custom field that used it on its own props. The stale `CLAUDE.md` bullet is updated.

**4. Optional-data control id under `XxxOf`** — confirmed, but it only occurs when the `oneOf`/`anyOf` schema is *itself* `type: 'object'` with its own `properties`, so that `SchemaField`'s main `FieldComponent` is an `ObjectField`. Rendering that shape on `v7` gives the outer field's control and the selected option's control the **same** DOM id, `root_obj__optionalAdd`, twice in one document. This branch gives the outer one `root_obj_XxxOf__optionalAdd`, consistent with its own title id `root_obj_XxxOf__title` right next to it. I kept the new behavior: reproducing the duplicate would take extra code to preserve a bug. The changelog and the guide's "byte-identical" sentence now name this one exception, and a test renders the shape and asserts every id is unique and both controls are present. `ArrayField` uses the same `id` and gets the same treatment.

**5. `XxxOf` id concatenation** — now `fieldPathToId(toFieldPath('XxxOf', fieldPath), globalFormOptions)`: the id a child named `XxxOf` would have, produced by the same two helpers as every other id. No string arithmetic on the separator remains in core.

**6. Three parsers** — the "hot path" claim was unbenchmarked, as you noted, so the hand-rolled scanner is gone. `fieldPathToId()` is `[idPrefix, ...fieldPathToList(fieldPath)].join(idSeparator)` and `fieldPathEndsWithIndex()` checks whether the last parsed segment is a number. `fieldPathToList()` is the only code that knows the grammar. The property test that compared the two derivations is deleted, since it would compare a function with itself.

**7. Test comment** — reworded to describe the current behavior only.

Root typecheck, uncached lint, knip, `cs-check`, and the `@rjsf/utils` (100% coverage), `@rjsf/core` and `@rjsf/mui` suites pass.

heath-freenome:
## Code review findings (re-review)

**Correctness**

1. **`packages/core/src/components/Form.tsx:832`** — `setFieldValue` silently drops embedded empty-string path segments instead of preserving them as literal keys, retargeting the write to a different `formData` address. `form.setFieldValue(['a', '', 'b'], value)` previously wrote to `formData.a[''].b` (old `toFieldPathId` only special-cased a top-level `''` argument). The new `path.reduce((acc, segment) => toFieldPath(segment, acc), ROOT_FIELD_PATH)` calls `toFieldPath` per segment, and `toFieldPath` treats any empty segment as "names no field, return parent unchanged," so the empty segment is silently swallowed and the write lands on `formData.a.b` instead — a different address. No test covers an embedded empty segment.

2. **`packages/core/src/components/fields/SchemaField.tsx:197`** — The `'XxxOf'` disambiguation suffix is an unescaped magic segment name, so a real schema property literally named `XxxOf` collides with the synthetic oneOf/anyOf wrapper id. A oneOf/anyOf schema whose selected branch has a property named `XxxOf` causes `toFieldPath('XxxOf', fieldPath)` for that real property to collide with `fieldPathToId(toFieldPath('XxxOf', fieldPath), globalFormOptions)` computed here for the wrapper's id — two different DOM elements get the same id, breaking `htmlFor`/aria association. This predates the PR, but the PR rewrote this exact code path while introducing a dedicated escaping grammar and had the opportunity to make the suffix collision-proof.

**Efficiency**

3. **`packages/utils/src/fieldPath.ts:99`** — `fieldPathToId`/`fieldPathToName` re-parse the entire ancestor path from scratch on every call instead of building ids incrementally as the tree is walked. The deleted `toFieldPathId.ts` carried the parent's already-computed path forward and concatenated one segment (O(1) per level). The new version fully re-parses the whole accumulated path string via `fieldPathToList`, invoked once per rendered field at every nesting level (`ObjectField.tsx:128`, `ArrayField.tsx:434`, `FallbackField.tsx:125`, `LayoutGridField.tsx:690`, `Form.tsx:833`). For a schema nested to depth D, this is O(D) parsing per field and O(D²) summed over a deep/linear schema, versus O(D) total before.

4. **`packages/core/src/components/fields/ArrayField.tsx:1035`** — `handleChange` calls `fieldPathEndsWithIndex`, which fully re-parses the entire changed field path just to check whether the trailing segment is numeric, on every bubbled change event. Typing into a text field nested inside array items fires this handler on every keystroke; each keystroke re-parses the full path string (with escape handling) merely to answer a boolean "does this end in `[N]`" question — work that scales with path depth on a hot interactive path where only the last segment's type is needed.

**Reuse / Simplification**

5. **`packages/utils/src/fieldPath.ts:47`** — `fieldPathToList` hand-rolls a character-by-character path tokenizer that duplicates the parsing responsibility already carried by `pathUtils.ts`'s `toPath`, giving the package two independently-maintained path grammars (differing in escape-handling and index-typing). A future syntax change must now be implemented and reasoned about twice.

6. **`packages/antd/src/templates/CyclicSchemaExpandTemplate/index.tsx:14`** (and 9 other theme packages) — Every theme's `CyclicSchemaExpandTemplate` hand-builds the expand button's DOM id with a literal template string instead of a shared `idGenerators` helper, duplicated identically across 10 packages. This PR already touches all 10 files (renaming `fieldPathId` to `id`) but leaves the ad hoc concatenation duplicated verbatim in every theme.

7. **`packages/core/src/components/Form.tsx:832`** — The idiom for folding a `FieldPathList` into a `FieldPath` via `reduce`+`toFieldPath` is duplicated verbatim in multiple files (`Form.tsx:832`, `LayoutGridField.test.tsx:763`, `fieldPath.test.ts:59`) instead of being exported once from `fieldPath.ts`. If the grammar's list-to-path folding rule ever changes, each copy must be found and updated independently.

8. **`packages/core/src/components/fields/SchemaField.tsx:176`** — `fieldPathProps` is redeclared as a full two-field object in the oneOf/anyOf branch even though `fieldPath` is identical in both branches and only `id` ever differs — needless indirection left over from the old two-varying-field (`fieldPathId`/`childFieldPathId`) shape this PR was meant to simplify.

---

jimmycallin:
Fixed in `a24a82fc1` — findings 1, 6 and 7. Findings 5 and 8 were already addressed in `59a33992d`, which this review appears to predate. 2, 3 and 4 I'm leaving, with reasons; 3 and 4 I benchmarked rather than argued.

## Correctness

**1. Embedded empty path segments** — confirmed, and the cause is one level below `setFieldValue`: the grammar could not *represent* an empty property name, so no call site could have got this right. `toFieldPath('', parent)` returned the parent, and `fieldPathToList` dropped the segment on the way back, which means a `FieldPathList` did not round-trip through a `FieldPath`.

The fix is that a `.` now separates two names rather than terminating one: the parser treats whatever follows a `.` as a segment even when it is empty, and `toFieldPath` no longer special-cases `''`. `toFieldPath('', 'a')` is `a.`, which parses back to `['a', '']`. The root stays the one place an empty name cannot live, since the root path is itself empty — that is `toFieldPath('')`, unchanged. This also fixes `ObjectField`, which would previously have given a schema property named `''` its parent's path and id.

A `Form.behaviors` test sets `['foo', '', 'bar']` on a schema with a property named `''` and asserts both the `formData` address and the `root_foo__bar` id; it fails on the previous commit. `fieldPath.test.ts` pins the parse of `a.`, `a..b` and `a[0].`.

**2. `XxxOf` as a magic segment** — real, pre-existing, and I'm leaving it. `XxxOf` is v6's literal suffix, and the whole contract of this PR is that ids are byte-identical to v6 apart from the one exception already documented. Making the suffix collision-proof means a segment no property name can produce, which means a different id for every `oneOf`/`anyOf` field in every form — a much larger break than the collision it avoids, which needs a `oneOf`/`anyOf` branch with a property literally named `XxxOf`. Worth doing in a release that is already changing ids on purpose, not here.

## Efficiency

**3 & 4** — I benchmarked rather than reasoned about it, since round 1's finding 6 asked for the opposite change (it called the hand-rolled scanner's "hot path" justification unbenchmarked, which it was) and I'd rather not flip the code back and forth on intuition. One million calls each, paths built by `toFieldPath` with an array index every third level:

| depth | path length | `fieldPathToId` | `fieldPathEndsWithIndex` |
| --- | --- | --- | --- |
| 3 | 23 | 0.18 µs | 0.11 µs |
| 5 | 37 | 0.27 µs | 0.17 µs |
| 10 | 78 | 0.51 µs | 0.35 µs |
| 20 | 170 | 1.13 µs | 0.82 µs |

The O(D²) is real and visible in the table, but the constant is a character loop over a string that is tens of bytes long. A 1000-field form nested ten deep spends ~0.5 ms total in `fieldPathToId` across a full render, against a React commit measured in tens of milliseconds; a keystroke inside an array item costs 0.35 µs in `fieldPathEndsWithIndex`. Carrying a parallel `FieldPathList` down the tree to avoid that means every component prop, every template signature and every theme carrying two representations of field identity instead of one — which is the `fieldPathId` object this PR exists to remove. Happy to revisit if a profile of a real form ever shows these functions.

## Reuse / simplification

**7. The `reduce` + `toFieldPath` idiom** — fixed, and it's the natural home for finding 1: `fieldPathFromList()` is now exported as the inverse of `fieldPathToList()`, `Form.setFieldValue` calls it, and a round-trip test covers every list shape the parser produces, the empty segment included.

**6. `CyclicSchemaExpandTemplate`'s button id** — fixed. `expandButtonId()` joins the other id generators in `idGenerators.ts`, and all ten templates (core plus nine themes) use it instead of their own `` `${id}-button` ``. Note it can't go through the existing `buttonId()`: that one produces `id__suffix`, while this id is `id-button`, so reusing it would change the id.

**5. Two path grammars** — already addressed in `59a33992d`: `toPath()` now carries a doc block saying it is *not* the `FieldPath` parser, that the two grammars only look alike, and that feeding an escaped path to it mangles it. They stay separate on purpose — `toPath` reads the unescaped dotted paths in AJV error properties, where `a.b` is two segments; `fieldPathToList` reads the escaped grammar, where `a\.b` is one. Merging them would mean one of the two callers loses the distinction it needs.

**8. `fieldPathProps` in the `oneOf`/`anyOf` branch** — already fixed in `59a33992d`: the object is gone, replaced by a single `fieldComponentId` variable, since the id was the only thing that varied.

Root typecheck, uncached lint, knip, `cs-check` and all 15 test projects pass (`@rjsf/utils` at 100% coverage).

jimmycallin:
## Self-review pass on the `FieldPath` encoding — `8f2e909f7`

The empty-segment bug was a round-trip failure, and round-tripping is the invariant this whole PR rests on: field identity stopped being an object and became a *string encoding* of one. That was covered by a handful of hand-picked escaping cases, which is sampling, not proof — and round 1's response deleted the one property test the PR had (correctly at the time: after unifying the parsers it compared a function with itself).

So it now proves the invariant instead of sampling it.

### Exhaustive round-trip

```ts
const SEGMENTS = ['a', '', '.', '[', ']', '\\', 'a.b', 'a[0]', 'a\\b', '0', 0];
```

Every reserved character alone and embedded, an empty name, a numeric-*looking* name that must stay a string, and a real numeric index. Every list of one to three of them — 1463 cases — must satisfy `fieldPathToList(fieldPathFromList(list))` deep-equals `list`. It's exhaustive over the interesting space rather than random, so there's no seed to chase and no flake.

**220 of those 1463 fail on the grammar as it stood before the empty-segment fix**, and the failure message names the first counterexample rather than just saying "false". Round-tripping also implies the encoding is injective, which is the property that actually matters at runtime: no two distinct fields can ever share a `FieldPath`.

The one case that legitimately does not round-trip is now pinned as its own test rather than living in a filter: a *leading* empty segment is the root, because the root path is itself the empty string, and `setFieldValue('')` is documented to address the root.

### A second id deviation from v6, which I should have disclosed

Reviewing the empty-segment fix against what it replaced: v7's `toFieldPathId` had

```ts
const childPath = fieldPath === '' ? [] : [fieldPath];
```

— it dropped an empty segment at *every* level, not just at the root. So a schema property named `''` took its parent's id, and the two collided in the document. This branch gives it `root_obj_`, its own id.

That's the same kind of exception as the `XxxOf` one already documented, and for the same reason (v6 emitted a duplicate id), but the changelog and the guide's "byte-identical to v6" sentence only named the first. Both now name this one too, and a render test asserts the input is `#root_obj_` and that every id in the form is unique.

While there: this also fixed a latent inconsistency in `LayoutGridField`. `getSchemaDetailsForField` walks the data with `getByPath(innerData, part)`, which treats `''` as a real key, while the path walk through `toFieldPath` skipped it — so a grid path with an empty part addressed one thing in the data and another in the id. The two agree now.

### What I looked for and did not find

- **Ad-hoc `FieldPath` string surgery outside `fieldPath.ts`.** Grepped for `split`/`slice`/`indexOf`/`startsWith`/template concatenation on a `FieldPath`, and for `as FieldPath`. Every construction and every parse is inside `fieldPath.ts`; the only external string operation is `setFieldValue`'s documented dotted-string split, which runs on its `string | FieldPathList` argument rather than on a `FieldPath`. No second parser crept back in.
- **Non-integer and negative indexes.** `toFieldPath(-1)` and `toFieldPath(1.5)` round-trip; covered by the enumeration through the `0` segment and checked by hand for the others.
- **Numeric property names vs array indexes.** `'0'` stays a string and `0` stays a number through the round-trip — that distinction is what `fieldPathEndsWithIndex` depends on, and it is now covered by 1463 cases rather than two.

Root typecheck, uncached lint, knip, `cs-check` and all 15 test projects pass.

heath-freenome:
## Code review findings (re-review, accounting for prior feedback and latest fixes)

Most of the prior review feedback has been addressed. Two items remain:

**New finding (correctness)**

1. **`packages/utils/src/fieldPath.ts:23`** — `toFieldPath('', ROOT_FIELD_PATH)` collapses to the root path itself, so a schema property literally named `''` directly under the root object is indistinguishable from the root form, and editing it silently replaces the entire `formData`.

   Schema `{ type: 'object', properties: { '': { type: 'string' } } }` (property `''` as a *direct* child of root). `ObjectField.tsx` computes `innerFieldPath = toFieldPath('', fieldPath)` where `fieldPath` is `ROOT_FIELD_PATH`, yielding `innerFieldPath === ROOT_FIELD_PATH`; its DOM id also collapses to `idPrefix` ('root'), colliding with the root form container. Typing into this field fires `onChange` with `fieldPath=''`, which `Form.tsx`'s `processPendingChange` parses via `fieldPathToList('')` → `[]` and treats as `isRootPath` (`formData = isRootPath ? newValue : structuredClone(oldFormData)`), replacing the **entire** form's `formData` with just the typed string, discarding every sibling property.

   This PR explicitly fixed the analogous *nested* case (property `''` below another object now gets its own id) and disclosed it in `CHANGELOG-v7.md`, but the top-level case remains broken; this data-loss consequence is untested (only the nested case is covered in `Form.behaviors.test.tsx`) and undisclosed.

   Note: the function's own docstring already acknowledges this exact limitation ("At the root it is the only name the grammar cannot hold"), so this may be a known, accepted tradeoff rather than an oversight — but the silent full-form data-loss consequence seems worth either guarding against or calling out explicitly as a risk.

**Carried over from the prior round (no action expected here)**

2. **`packages/core/src/components/fields/SchemaField.tsx:195`** — The synthetic `'XxxOf'` id-disambiguation segment is still unescaped, so a real schema property literally named `XxxOf` inside a `oneOf`/`anyOf` branch collides with the synthetic wrapper's id, giving two different DOM elements the same id and breaking `htmlFor`/aria association. This is pre-existing behavior from v6, and per the author's response in the previous round it's being intentionally left for a later release that's already changing ids on purpose — flagging only because it was raised before and remains unaddressed, not as a new defect.

---

jimmycallin:
Item 1 fixed in `271b64547`.

The grammar now holds an empty name under the root: it is written as a lone `.`, so `toFieldPath('')` is `.`, `['', 'a']` is `..a` and `['', 0]` is `.[0]`, all distinct from the root `''`. The parser needed no change — a leading `.` already read as an empty first name — so the fix is the one branch in `toFieldPath`, and the docstring's "only name the grammar cannot hold" is gone. With that, the exhaustive round-trip test drops its leading-`''` exclusion: all 1463 lists round-trip, and the "root exception" test is replaced by one pinning the three spellings above.

`Form.behaviors.test.tsx` renders `{ '': string, other: string }` at the root, asserts the input's id is `root_`, types into it, and checks `onChange` reports `{ '': 'x', other: 'kept' }` against `root_`. It fails on the previous commit with the whole `formData` replaced, which is the data loss you described.

One piece of fallout: `setFieldValue('')` relied on the collapse to mean the root, so it now maps `''` to `[]` explicitly, which is what its docstring already said it does. The `CHANGELOG-v7.md` entry now says the `''` fix applies at every level, root included.

Item 2 is unchanged, as agreed.

---

Since the root case was missed one step from the nested one, I also re-checked every other special-path consumer with rendered `Form` tests: root detection in `processPendingChange`, `setFieldValue`, `fieldPathToName` and both name generators; root properties named `.`, `[`, `]`, `\`, `0`, `a.b`, `a[0]`; `'0'` as a key inside an array item; arrays of arrays; a root `additionalProperties` key `''`; `omitExtraData` with a `''` property; `idSeparator: '.'` with `idPrefix: ''`. All correct, and no raw string operations on a `FieldPath` exist outside `fieldPath.ts`.

One pre-existing limitation of the same class, outside this PR and unchanged from `v7` and v6, for the record: a validation error on a property named `''` lands on the root of the `errorSchema`, because `processRawValidationErrors` turns AJV's `instancePath` `/` into the dotted `property` string `.` and then `.split('.').filter(Boolean)` drops the empty segment. That is the `RJSFValidationError.property` grammar, which cannot represent an empty name at all, so it is a separate change to the validator's error-path format rather than something this PR should absorb.

x0k:
Could you please add [SJSF](https://github.com/x0k/svelte-jsonschema-form) to the `How it's implemented in other libraries` section, just in case?

jimmycallin:
@x0k Added, thanks — and it was worth reading rather than just listing. SJSF turns out to be the closest relative in that section: a branded path (`Readonly<Array<string | number>>`) with the DOM id derived by a pluggable `FormIdBuilder.fromPath()`, i.e. the same deterministic path-derived ids, with the path/id split this PR is introducing already factored out behind an interface.

Two things I called out in the body. The array-vs-string difference costs you nothing because Svelte doesn't compare props by reference — that choice is forced on us by React memo, not by the domain. And your pseudo-elements (`oneof`/`anyof`/`title`/`errors` encoded behind a reserved `__sjsf_pseudo_element` prefix) are exactly the collision-proof answer to the `XxxOf` suffix a reviewer flagged here; we're keeping the v6-compatible spelling in this PR and that's the shape I'd want when we do change ids on purpose.

x0k:
@jimmycallin Good writeup - and the SJSF section is fair. The pseudo-element call-out is spot on.

One thing missing from the array-vs-string analysis is that `FieldPath` is an interned backed by a trie, so the same logical path always has the same object identity.

This makes the array useful beyond prop comparison:

1. **No parsing for ID builders:** `fromPath()` gets ready-to-iterate segments; a string would need to be parsed first.
2. **Cheap structural operations:** parent/child paths are derived with `concat()`/`slice()`, without parse/serialize cycles.
3. **O(1) array-child detection**: `typeof path[path.length - 1] === "number"` is enough to tell whether a field is an array child
4. **The trie is reused elsewhere:** cycle detection and subtree traversal operate directly on paths.

heath-freenome:
## Re-review: new issue found

Re-checked against the latest fixes and prior feedback. One genuine new issue, plus one lead I checked and ruled out.

### Confirmed: stale docs for `toFieldPath('')`

**`packages/docs/docs/api-reference/utility-functions.md:1452`**

The doc still says: "An empty `segment` names no field, so the parent path is returned unchanged" — i.e. implying `toFieldPath('', parentPath) === parentPath`.

That was true before the empty-property-name fix in `271b64547` landed in this PR. Now (`packages/utils/src/fieldPath.ts`) an empty segment gets its own distinct `FieldPath`:
- `toFieldPath('', ROOT_FIELD_PATH)` → `'.'`
- `toFieldPath('', 'a')` → `'a.'`

Neither equals the parent path. The doc wasn't updated alongside that fix. A plugin author following the current doc text to hand-build paths would reproduce the exact collision/data-loss bug this PR spent multiple review rounds fixing.

### Ruled out: `setFieldValue('', ...)` mapping to root

Considered flagging `Form.tsx`'s `setFieldValue(fieldPath: string, ...)` treating `''` as the root path as a data-loss hazard (no way to address a root property literally named `''` via the string overload). Checked the live branch — this is intentional and documented: the docstring states `''` or `[]` both mean root, and there's a comment explaining the tradeoff. Not a bug.

heath-freenome (approved):
Approved pending the doc fix

PR #5281, opened 2026-09-12
fix(utils): replace a non-object left-hand value when merging an object over it

jimmycallin:
`mergeObjects()` and `mergeSchemas()` recurse whenever a key is present on both sides and the right-hand value is an object. They never check that the *left-hand* value is one too.

```ts
mergeObjects({ a: 5 }, { a: { b: 1 } });
// TypeError: Cannot use 'in' operator to search for 'b' in 5

mergeSchemas({ items: [{ type: 'string' }] }, { items: { type: 'string' } });
// { items: { 0: { type: 'string' }, type: 'string' } }
```

The first is the recursive call landing on a primitive, where `key in obj1` throws. The second is the same call landing on an array: `{ ...[{…}] }` seeds the accumulator with the array's indices, and the right-hand keys are merged on top of them.

A *falsy* left-hand value (`false`, `null`, `undefined`, `0`) already did the right thing, because `obj1 ? obj1[key] : {}` and the `obj1 &&` guard treated it as absent — guards that read as null checks but were in fact the only thing keeping a boolean schema (`additionalProperties: false` merged with `additionalProperties: { … }`) from crashing. This makes every non-object take that path: the right-hand object replaces the left-hand value.

## Where this bites

`mergeSchemas()` is how `allOf`, `dependencies` and `if`/`then`/`else` subschemas are folded together in `retrieveSchema()`, and how `MultiSchemaField` merges parent properties into a `oneOf`/`anyOf` option. JSON Schema lets a keyword hold either a boolean, an array or an object — `items`, `additionalProperties`, `additionalItems`, `contains` — so two branches of an `allOf` disagreeing about which is a schema the merge can reach, not a hypothetical.

`mergeObjects()` merges `uiSchema` fragments (`resolveUiSchema()`), `schema.default` values (`getDefaultFormState()`) and error schemas (`validationDataMerge()`), where a scalar meeting an object under the same key is just as reachable.

## The change

```diff
-const left = obj1 ? obj1[key] : {},
+const left = obj1[key],
   right = obj2[key];
-if (obj1 && key in obj1 && isObject(right)) {
-  acc[key] = mergeObjects(left, right, concatArrays);
+if (key in obj1 && isObject(right)) {
+  acc[key] = mergeObjects(isObject(left) ? left : {}, right, concatArrays);
```

The `obj1 ?` / `obj1 &&` guards go with it. Both parameters are required and non-nullable, and the only caller that ever passed a falsy object was this recursion — `@rjsf/utils` holds 100% branch coverage, and after the fix those branches are unreachable, which is how they were spotted.

Extracted from a larger `any`-elimination branch, where the type checker surfaced it; it stands on its own.

## Checklist

  - [x] I've verified the full build (18 projects) and test suite (15 projects) pass, and that root `typecheck`, `knip`, `lint` and `cs-check` are clean
  - [x] `@rjsf/utils` still reports 100% coverage
  - [ ] I've updated the migration guide — not applicable, this is a bug fix with no API change


CHANGELOG-v7.md
```diff
@@ -48,7 +48,7 @@ should change the heading of the (upcoming) version to include a major version b
 
 - Fixed the `getTestRegistry()` test helper passing an options object where positional arguments were expected, which silently dropped its templates and widgets; it now imports from `@rjsf/core/testing` ([#5272](https://github.com/rjsf-team/react-jsonschema-form/pull/5272))
 - **BREAKING CHANGE** `TimeWidget` pads missing seconds and appends the browser's local UTC offset to its value on change, matching `@rjsf/core`'s `TimeWidget` (`format: "iso-time"` still pads seconds but skips the offset, since that format's timezone is optional). `DateTimeWidget` similarly formats a `format: "iso-date-time"` value as a naive local date-time string instead of a UTC ISO string, stripping a stored offset before parsing so it displays as the naive wall-clock time it represents; an unparsable stored value resolves to no selected date rather than crashing the calendar on an `Invalid Date` ([#3930](https://github.com/rjsf-team/react-jsonschema-form/issues/3930))
-||||||| parent of c459bce95 (refactor(core,daisyui): put registry assembly in Theme.ts, drop daisyui's no-op fields)
+  ||||||| parent of c459bce95 (refactor(core,daisyui): put registry assembly in Theme.ts, drop daisyui's no-op fields)
```
heath-freenome: Aren't you removing this in a different PR? Why change it here?

packages/utils/src/mergeObjects.ts
```diff
@@ -17,10 +17,10 @@ export default function mergeObjects(
 ) {
   return Object.keys(obj2).reduce(
     (acc, key) => {
-      const left = obj1 ? obj1[key] : {},
+      const left = obj1[key],
```
heath-freenome, line 20: Dropping the `obj1 ? obj1[key] : {}` guard means a falsy/undefined top-level `obj1` now throws instead of being tolerated as `{}`. Confirmed by direct execution: `mergeObjects(undefined, { a: 1 })` used to return `{ a: 1 }`; after this change it throws `TypeError: Cannot read properties of undefined (reading 'a')`. Same pattern in `mergeSchemas.ts` (line 17).

Every internal call site always passes real objects, so the test suite (rightly, per the PR description) can't reach this branch — but this is a public export (`@rjsf/utils`), and any external/loosely-typed caller relying on the old top-level leniency will now crash instead of degrading gracefully. Worth calling out explicitly in the changelog entry, or confirming this narrowing is intentional and acceptable as a breaking change for direct callers of `mergeObjects`/`mergeSchemas`.

CHANGELOG-v7.md
heath-freenome: This line is a literal unresolved git merge-conflict marker (`||||||| parent of c459bce95 (...)`), and this diff re-indents it by two spaces instead of removing it — so the broken conflict marker ships in `CHANGELOG-v7.md` after this merges, followed by an orphaned bullet fragment. The marker predates this PR (already on `v7`), but since this diff touches this exact line, it'd be worth cleaning it up here rather than perpetuating it.

heath-freenome (commented):
Automated review found two items worth a look — see inline comments.

CHANGELOG-v7.md
jimmycallin: The changelog diff was against a stale `v7`. Rebased; this PR now only adds its own entry under `@rjsf/utils`.

jimmycallin: Gone after the rebase — current `v7` no longer carries the marker, and this branch no longer touches that line.

packages/utils/src/mergeObjects.ts
jimmycallin: Confirmed the behavior change. Intentional: both signatures have always declared `obj1: GenericObjectType`, and the `obj1 ? … : {}` guard was unreachable for any caller that type-checks. Restoring it would need a test that passes `undefined` through a cast just to keep `@rjsf/utils` at 100% coverage, which I would rather not encode. The changelog entry now states it explicitly as a breaking change for direct callers (`86d9213ed`).

heath-freenome (approved):

PR #5284, opened 2026-09-13
fix: latent bugs surfaced while strengthening the types

jimmycallin:
Six bugs found by tightening the generic defaults in `@rjsf/utils` (`T = any` / `F extends FormContextType = any` → `unknown`, `GenericObjectType = Record<string, any>` → `Record<string, unknown>`) and following what the compiler flagged. A prop read off a bag typed `Record<string, any>` is `any` whether or not anything ever supplies it, so "read a prop nobody passes" type-checks and silently does nothing.

**The type change itself is not in this PR** — only the fixes, each of which stands on its own against the current types.

## Fixes

**`@rjsf/utils`**

- `mergeObjects()` / `mergeSchemas()` recursed whenever a key was present on both sides and the right-hand value was an object, without checking the left-hand one. A truthy primitive (`{a: 5}` merged with `{a: {b: 1}}`) threw `TypeError: Cannot use 'in' operator`; an array merged its indices into the result. This is [#5281](https://github.com/rjsf-team/react-jsonschema-form/pull/5281), applied **unchanged** to the v6 line so the two branches don't diverge — it's the first commit here, byte-identical to what landed on `v7`.
- `useFileWidgetProps.handleChange` is typed `(files: FileList)` but themes legitimately hand it arrays; it now accepts `FileList | File[]`. `processFiles()` already used `Array.from()`, so this is additive.

**`@rjsf/core`**

- `ui:placeholder` was dropped for multi-select arrays and custom array widgets — both read `placeholder` from props, which nothing passes. Now read from `getUiOptions()`, as `StringField` does.
- A `ui:FieldTemplate` / `ui:FieldErrorTemplate` override on a `LayoutMultiSchemaField` was ignored: it resolved both templates by passing `options` (the anyOf/oneOf *schemas*) where `getTemplate()` wants the *UI* options.

**`@rjsf/mantine`**

- ⚠️ `onChangeOverride` received the value instead of the change event, unlike its declared type and `@rjsf/core`. **Potentially breaking for anyone using `onChangeOverride` with this theme.**
- A non-multiple `FileInput` dropped the file: Mantine hands back a lone `File`, which isn't iterable, so `Array.from()` gave `[]`. Its clear button sends `null` for a single input and `[]` for a multiple one; `null` threw, `[]` was concatenated onto the existing value. Both now report an empty value.

**`@rjsf/chakra-ui`**

- `getChakra()` `delete`d keys straight out of `uiSchema['ui:options'].chakra`, mutating the caller's own uiSchema as a side effect of rendering. Now builds a filtered copy.
- `ChakraUiSchema` used `Omit<UiSchema, 'ui:options'>`. `UiSchema` has a string index signature, so `keyof UiSchema` includes `string` and `Omit` erased every named member, leaving a type that accepted anything. It intersects instead.

## Verification

12 tests, each confirmed failing against its unfixed source: 3 merge regressions (from #5281), 2 for `ui:placeholder`, 1 for the `LayoutMultiSchemaField` template override, 3 for the mantine template (including that `onChangeOverride` gets the event), 1 for the mantine `FileWidget`, and 4 for `getChakra()` (including that it leaves its argument alone). `@rjsf/utils` keeps 100% coverage.

`lint`, `knip`, `build`, `typecheck`, `test` and `cs-check` all clean.

## Next

Follow-ups in the wider `any` removal (lint ratchet, the `StrictRJSFSchema`/`= any` sweep, `GenericObjectType`, `EnumOptionsType['value']`, the `ErrorSchema`/`FormValidation` reshape) are v7 work and are not proposed for this branch.


jimmycallin:
Superseded by #5285, which targets `v7` — the `@rjsf/mantine` `onChangeOverride` fix is a breaking change, so it belongs on the major, and the merge fix here was already #5281 on `v7`.

PR #5285, opened 2026-09-13
fix: latent bugs surfaced while strengthening the types

jimmycallin:
Five bugs found by tightening the generic defaults in `@rjsf/utils` (`T = any` / `F extends FormContextType = any` → `unknown`, `GenericObjectType = Record<string, any>` → `Record<string, unknown>`) and following what the compiler flagged. A prop read off a bag typed `Record<string, any>` is `any` whether or not anything ever supplies it, so "read a prop nobody passes" type-checks and silently does nothing.

**The wider type change is not in this PR** — only the fixes, plus the typing of the paths they touch so those files need no second pass when it lands.

Targets `v7` because of the `@rjsf/mantine` breaking change below.

## Fixes

**`@rjsf/core`**

- `ui:placeholder` was dropped for multi-select arrays and custom array widgets — both read `placeholder` from props, which nothing passes. Now read from `getUiOptions()`, as `StringField` does.
- A `ui:FieldTemplate` / `ui:FieldErrorTemplate` override on a `LayoutMultiSchemaField` was ignored: it resolved both templates by passing `options` (the anyOf/oneOf *schemas*) where `getTemplate()` wants the *UI* options.

**`@rjsf/mantine`**

- ⚠️ **BREAKING CHANGE:** `onChangeOverride` received the input's value instead of the change event, unlike its declared type, `@rjsf/core`, and every other theme. Documented in the [v7 upgrade guide](https://github.com/rjsf-team/react-jsonschema-form/blob/v7/packages/docs/docs/migration-guides/v7.x%20upgrade%20guide.md), with a before/after.
- A non-multiple `FileInput` dropped the file: Mantine hands back a lone `File`, which isn't iterable, so `Array.from()` gave `[]`. Its clear button sends `null` for a single input and `[]` for a multiple one; `null` threw, `[]` was concatenated onto the existing value. Both now report an empty value.

**`@rjsf/chakra-ui`**

- `getChakra()` `delete`d keys straight out of `uiSchema['ui:options'].chakra`, mutating the caller's own uiSchema as a side effect of rendering. Now builds a filtered copy.
- `ChakraUiSchema` used `Omit<UiSchema, 'ui:options'>`. `UiSchema` has a string index signature, so `keyof UiSchema` includes `string` and `Omit` erased every named member, leaving a type that accepted anything. It intersects instead.

**`@rjsf/utils`**

- `useFileWidgetProps.handleChange` is typed `(files: FileList)` but themes legitimately hand it arrays; it now accepts `FileList | File[]`. `processFiles()` already used `Array.from()`, so this is additive.

## Types on the touched paths

- `mergeObjects()` / `mergeSchemas()` read their operands as `unknown` instead of `any`, so every use is guarded by `isObject()` or `Array.isArray()` rather than assumed — which is exactly what hid the crash fixed in #5281.
- mantine's `BaseInputTemplate` and `FileWidget` take `T = unknown` / `F extends FormContextType = FormContextType` and constrain `S` to `RJSFSchema`. Both are leaves, so it's contained.

## Verification

9 tests; the 7 behavioural ones were each confirmed failing against unfixed source (2 for `ui:placeholder`, 1 for the `LayoutMultiSchemaField` template override, 1 for `onChangeOverride` getting the event, 2 for clearing a mantine `FileInput`, 1 for `getChakra()` leaving its argument alone).

`build`, `typecheck`, `test`, `lint`, `knip` and `cs-check` all clean.

## Next

The finished end state is at https://github.com/jimmycallin/react-jsonschema-form/pull/48 (`no-explicit-any` 529 → 0 across `utils` + `core` + `validator-ajv8`), reference only. The plan is to land it as a stack: this PR, then the lint ratchet (warnings only), local `any` removal, the mechanical `StrictRJSFSchema`/`= any` sweep, then `GenericObjectType`, `EnumOptionsType['value']`, and the `ErrorSchema`/`FormValidation` reshape. Per-*package* staging isn't possible, since `tsc -b` project references mean `utils` and every consumer must move together.


jimmycallin:
Pushed a cleanup commit (no change to what the PR fixes):

- The "an empty selection is a cleared input" rule moved from mantine's `FileWidget` into `useFileWidgetProps.handleChange`, where `handleRemove` already encodes the same convention. `core`'s `FileWidget` gets the same behaviour, and the widget keeps only the Mantine-shaped part (normalizing `File | File[] | null`). Covered by two new `@rjsf/utils` tests.
- mantine's `BaseInputTemplate` now picks the override at the JSX (`onChange={onChangeOverride ?? handleChange}`), matching `@rjsf/core` and the other seven themes, rather than branching inside the callback.

Note the changelog now also lists the `useFileWidgetProps` behaviour change under `@rjsf/utils`.

heath-freenome:
## Code review findings

**`packages/mantine/src/templates/BaseInputTemplate.tsx:96`** — the `onChangeOverride` fix only covers the `TextInput` path; `NumberInput`'s `handleNumberChange` never checks `onChangeOverride` at all, so it's silently never invoked for numeric fields.

For a form with a `type: 'number'`/`'integer'` schema field in `@rjsf/mantine` using a `ui:options.onChangeOverride`: the app's override is never called (not even with the old, wrong string argument) because `<NumberInput onChange={!readonly ? handleNumberChange : undefined}>` is hardcoded and `handleNumberChange` ignores `onChangeOverride` entirely. This contradicts the PR's own `CHANGELOG-v7.md` claim that "an `onChangeOverride` used with this theme now receives the `ChangeEvent`, as it does in every other theme" — untrue for numeric fields, unlike `@rjsf/core`'s `BaseInputTemplate`, which wires `onChangeOverride` uniformly for all input types via a single `<input>` element. No test in `packages/mantine/test/BaseInputTemplate.test.tsx` exercises `onChangeOverride` with a numeric schema, so the gap is untested.

---

Otherwise I verified the PR's core fixes (chakra-ui `getChakra`/`ChakraUiSchema`, core `ui:placeholder`, `LayoutMultiSchemaField` template resolution, mantine `FileInput` normalization, `useFileWidgetProps` empty-selection handling, `mergeObjects`/`mergeSchemas` typing) against the actual Mantine API types and callers — all correct.

jimmycallin:
Fixed in `0383b9f9c`.

Confirmed: `<NumberInput onChange={!readonly ? handleNumberChange : undefined}>` was hardcoded and `handleNumberChange` never looked at `onChangeOverride`, so a numeric field's override was not merely passed the wrong argument — it was never called at all, and the changelog's claim was untrue for that path.

It can't be fixed by wiring the override into `handleNumberChange`: Mantine omits the native `onChange` from `NumberInputProps` and redefines it as `(value: string | number) => void`, and internally feeds it through `useUncontrolled`, so no `ChangeEvent` ever exists on that path. There is nothing of the declared type to hand the override.

So a numeric field with an `onChangeOverride` now renders the `TextInput` branch — the plain `<input>` that `@rjsf/core` and the other seven themes render for a numeric field, receiving `{...inputProps}` including `type='number'`. The override gets the real `ChangeEvent`, and since it is what handles the input from there, the stepper controls and number formatting it replaces have nothing left to drive. Without an override the `NumberInput` path is untouched.

Two tests in `packages/mantine/test/BaseInputTemplate.test.tsx` now render a `type: 'number'` schema: one asserts the override is called once with the event and `onChange` is not, the other that without an override `onChange` still receives the parsed `42` rather than `'42'`. The first fails on the previous commit.

The changelog entry and the guide section both name the numeric case now.

Root typecheck, uncached lint and the full `@rjsf/mantine` suite (162 tests) pass.

packages/mantine/src/widgets/FileWidget.tsx
```diff
@@ -33,12 +35,12 @@ export default function FileWidget<T = any, S extends StrictRJSFSchema = RJSFSch
...
   const handleOnChange = useCallback(
-    (files: any) => {
-      if (typeof files === 'object') {
-        // handleChange is async; DOM event handlers are void-returning, so we intentionally don't await
-        // oxlint-disable-next-line no-floating-promises, no-void
-        void handleChange(files);
-      }
+    (files: File[] | File | null) => {
+      // Mantine's `FileInput` hands back a `File[]` when `multiple` and a lone `File` otherwise, and `null` when cleared
+      const selected = files ? [files].flat() : [];
```
heath-freenome: Minor: `[files].flat()` normalizes `File | File[] | null` via a throwaway wrapper-array allocation plus flatten. `Array.isArray(files) ? files : [files]` does the same with less work and reuses the existing array reference in the multi-file case.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome, line 1: **Line 326** (`ArrayAsFiles`, not touched by this diff): this function still drops `ui:placeholder` — the same bug this PR just fixed two functions above/below it in this file (`ArrayAsMultiSelect`, `ArrayAsCustomWidget`).

`getUiOptions()` here destructures `{ widget = 'files', title: uiTitle, ...options }` without pulling `placeholder` out first, and `options={options}` only puts it inside the options bag. But `FileWidget`/`BaseInputTemplate` read `placeholder` off the top-level `WidgetProps.placeholder`, not `options.placeholder`, and no `placeholder` prop is set on `<Widget>` here. So a schema like `{ type: 'array', items: { type: 'string', format: 'data-url' } }` with `uiSchema: { 'ui:placeholder': 'Drop files here' }` still renders with no placeholder.

packages/mantine/src/widgets/FileWidget.tsx
heath-freenome, line 1: **Line 68** (`value={value || ''}`, not touched by this diff): passing `''` instead of `null` for an empty file value breaks Mantine `FileInput`'s internal `hasValue` check.

Mantine computes `hasValue = Array.isArray(_value) ? _value.length !== 0 : _value !== null` and only renders `placeholder` when `!hasValue`. Since this widget always passes a string (never `null`) for the controlled `value`, `hasValue` is permanently `true`: a fresh `FileWidget` with `ui:placeholder: 'Choose a file'` never shows the placeholder, and one with `ui:options: { clearable: true }` shows an active Clear button before any file is chosen.

This is the exact feature area the PR adds two new tests for (`FileWidget.test.tsx`), but both start from non-empty `formData` so neither catches it.

packages/mantine/src/templates/BaseInputTemplate.tsx
```diff
@@ -96,32 +94,35 @@ export default function BaseInputTemplate<
...
+      value={value}
+      min={typeof min === 'number' ? min : undefined}
+      max={typeof max === 'number' ? max : undefined}
+      aria-describedby={ariaDescribedByIds(id, !!schema.examples)}
+    />
+  ) : (
+    <TextInput
+      onChange={!readonly ? (onChangeOverride ?? handleChange) : undefined}
+      {...componentProps}
+      {...inputProps}
```
heath-freenome, line 119: When `onChangeOverride` is supplied for a numeric schema, this renders Mantine's `TextInput` via `{...inputProps}` (not `restInputProps`), which still carries `type: 'number'`, `step`, `min`, `max` from `getInputProps()` — so the DOM node stays a native `<input type="number">`, not the plain text field the fix and its changelog/migration-guide entry claim ("renders Mantine's `TextInput` rather than its `NumberInput`").

Confirmed against `@mantine/core` 9.6.1 source: `Input.mjs` spreads `...rest` (including `type`) onto the underlying `<input>`, and `TextInput`'s own types declare no `type` override, so the attribute passes straight through.

A numeric field with an `onChangeOverride` meant to apply custom formatting (e.g. thousands separators like `1,234` or a currency prefix) gets a native `type="number"` input: browsers block non-numeric keystrokes (commas, letters, currency symbols) before any `change` event fires, so `onChangeOverride` never sees them — silently breaking the exact custom-formatting use case the override exists for. The new tests only exercise `fireEvent.change` with an already-valid numeric string (`'42'`), which jsdom doesn't reject, so this regression is untested and would only surface in a real browser.

heath-freenome (commented):
Re-review pass, checking the latest fixes against previous review threads. Found one more instance of the same `ui:placeholder` bug pattern this PR already fixes elsewhere, a related regression in the mantine `FileWidget`'s empty-value handling, a minor efficiency nit, and one additional issue in the `BaseInputTemplate` numeric/`onChangeOverride` fix where the DOM node stays `type="number"` despite the changelog claiming it renders as plain text.

packages/mantine/src/widgets/FileWidget.tsx
jimmycallin: Changed in `109abcfd5`. `null` still needs a branch, so it is `Array.isArray(files) ? files : [files].filter((file) => file !== null)` rather than a bare `[files]` — the lint config forbids the nested ternary that would express it directly.

packages/core/src/components/fields/ArrayField.tsx
jimmycallin: Confirmed and fixed in `05d4f4ea3`: `ArrayAsFiles` now pulls `placeholder` out of `getUiOptions()` and passes it as the top-level prop, like the other two. A test in `ArrayField.test.tsx` renders a `data-url` array with `ui:placeholder` through a stub `FileWidget` and fails on the previous commit. Changelog entry covers it (`6b2dcf939`).

packages/mantine/src/widgets/FileWidget.tsx
jimmycallin: Confirmed against the same `hasValue` line in 9.6.1. Fixed in `109abcfd5`: `value={value || null}`. A multiple input already passed `[]`, which fails Mantine's length check, so only the single case changed. The new test renders an empty single input with `ui:placeholder` and `clearable` and asserts the placeholder is shown and no Clear button exists; it fails on the previous commit. The 48 snapshot updates in that commit are the placeholder span now rendering for an empty input, which shifts every later `useId`. Changelog entry extended.

packages/mantine/src/templates/BaseInputTemplate.tsx
jimmycallin: The DOM node keeping `type="number"` is deliberate, not a gap. `@rjsf/core`'s `BaseInputTemplate` spreads the same `inputProps` onto its `<input>` with `onChange={onChangeOverride || handleChange}`, so a numeric field with an override is a `type="number"` input in core and in the other seven themes too. Stripping `type`, `step`, `min` and `max` here would make mantine the one theme where supplying an override changes the input type. The keystroke-blocking point is real, but it applies to every theme equally and is a separate change if wanted.

What was wrong was the wording: "plain text field" overstated it. `6b2dcf939` makes the changelog and the guide say the `TextInput` keeps `type="number"`, as core's `<input>` does.

jimmycallin:
Re-review pass over the sibling paths of the bug patterns this PR fixes, since the last round found two of them missed on neighbouring code. Two more, both fixed:

- **`@rjsf/antd` `BaseInputTemplate`** had the exact `onChangeOverride` gap fixed in mantine: its `InputNumber` branch was wired straight to `onChange`, so an override on a numeric field was never called. Same fix — with an override the `Input` branch renders, still `type="number"`, and the override receives the `ChangeEvent` (`b1d666b47`). New `packages/antd/test/BaseInputTemplate.test.tsx` fails on the previous commit. Changelog bullet under `@rjsf/antd`.
- **`BooleanField`** stranded `placeholder` in the options bag like `ArrayField` did, so `ui:placeholder` on a boolean rendered as a select never reached `SelectWidget` (`a4af207f6`). Test in `BooleanField.test.tsx`; the core changelog bullet now lists boolean fields.

Checked and clean: every other theme's `BaseInputTemplate` has a single input path with `onChange={onChangeOverride || handleChange}`; every other `getUiOptions()` → widget site passes `placeholder` top-level; the `value || ''` siblings in mantine, antd, daisyui, chakra and shadcn are all string inputs where `''` is correct, and mantine `SelectWidget`/`DateTimeInput` already pass `null`.

packages/mantine/src/widgets/FileWidget.tsx
heath-freenome: Fair — the lint constraint explains it. LGTM.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: Confirmed fixed, with a test that catches the regression. Thanks.

packages/mantine/src/widgets/FileWidget.tsx
heath-freenome: Confirmed fixed, with a test that catches the regression. Thanks.

packages/mantine/src/templates/BaseInputTemplate.tsx
heath-freenome: Good point — matching core and the other themes is the right call, and the wording fix resolves my actual concern.

packages/antd/src/templates/BaseInputTemplate/index.tsx
```diff
@@ -72,8 +72,11 @@ export default function BaseInputTemplate<
 
   const { min, max, ...restInputProps } = inputProps;
 
-  const input =
-    inputProps.type === 'number' || inputProps.type === 'integer' ? (
+  // `InputNumber` reports only the parsed value, never the `ChangeEvent` an `onChangeOverride` is declared to
+  // receive, so a widget that supplies one gets the plain input every other theme renders for a numeric field.
+  const isNumeric = !onChangeOverride && (inputProps.type === 'number' || inputProps.type === 'integer');
+
+  const input = isNumeric ? (
```
heath-freenome, line 79: `const input = isNumeric ? (` is shorter than the ternary head it replaced, but the `<InputNumber .../>` and `<Input .../>` JSX blocks below kept their old (deeper) indentation, so the file is no longer formatted correctly.

Confirmed by running `oxfmt --check` on this file in a clean checkout of the PR head (`b1d666b47`) — it fails. `pnpm run cs-check` (part of `sanity-check`, which CI runs) will fail on this PR as committed. `oxfmt` (no `--check`) should auto-fix it.

heath-freenome (commented):
All four points from the previous review are addressed — three with real fixes and test coverage, one (the mantine `type="number"` behavior) with a fair pushback that it matches core and every other theme; only the changelog wording needed fixing, and it was. Replied and resolved each thread individually.

One likely CI-breaking issue (inline comment on `packages/antd/src/templates/BaseInputTemplate/index.tsx:79`): the antd port of the numeric/`onChangeOverride` fix left the JSX under the new (shorter) ternary head at its old indentation, so `oxfmt --check` fails on this file as committed — confirmed locally against `b1d666b47`.

**One correctness risk I can't anchor inline**, since the file isn't touched by this PR: `useFileWidgetProps.handleChange` (`packages/utils/src/useFileWidgetProps.ts`) now treats *any* empty `FileList`/`File[]` as "clear the value," to fix the mantine multi-file append bug. But `packages/core/src/components/widgets/FileWidget.tsx:104-108` (`handleOnChangeEvent`) only guards `event.target.files` being null/undefined, not empty — it forwards an empty-but-truthy `FileList` straight into the new clear path. On platforms where reopening the native file picker and cancelling re-fires `change` with an empty `FileList` (a documented quirk on some mobile/embedded WebViews), a user with an existing selection in any core-based theme (antd, mui, bootstrap, semantic-ui, chakra-ui, daisyui, shadcn, fluentui-rc) would have it silently cleared — previously a harmless no-op. Worth a guard in `FileWidget.tsx`'s `handleOnChangeEvent` (`if (event.target.files?.length)`) alongside the `useFileWidgetProps` fix, or confirming this widget is intentionally out of scope for that behavior change.

Non-blocking: the `isNumeric = !onChangeOverride && (inputProps.type === 'number' || inputProps.type === 'integer')` guard (and its comment) is now duplicated near-verbatim between `mantine/BaseInputTemplate.tsx` and `antd/BaseInputTemplate/index.tsx`. Per this repo's CLAUDE.md guidance on generalizing logic duplicated across themes, a shared helper in `@rjsf/utils` might be worth it if a third theme ends up needing the same decision — not asking for a change in this PR.

heath-freenome (approved):
@jimmycallin two small fixes for you to handle. I will trust that your local claude code review will settle any tweaks on those fixes before you merge

packages/antd/src/templates/BaseInputTemplate/index.tsx
jimmycallin: Confirmed with `oxfmt --check` and fixed in `84065375a`. `cs-check` now passes on the file.

jimmycallin:
Added the guard in `84065375a`: core's `FileWidget` now only forwards `event.target.files` when it has a length, so an empty `FileList` from a cancelled picker is a no-op again rather than reaching the new clear path in `useFileWidgetProps`.

One nuance worth noting: before this PR an empty list already cleared a *single* file input on that path (`processFiles([])` gave `[]`, and `onChange(newValue[0])` sent `undefined`); only the multiple case was a no-op. The guard makes both a no-op, which is the behaviour a cancel should have. The mantine widget is unaffected, since its clear button intentionally goes through the empty-selection path.

Test in `StringField.test.tsx` fires a `change` with `files: []` against a form that has a value and asserts `onChange` is not called; it fails on the previous commit. Changelog bullet under `@rjsf/core`.

On the duplicated `isNumeric` guard: agreed on leaving it until a third theme needs it.

PR #5287, opened 2026-09-13
feat(shadcn)!: style through the consumer's Tailwind build instead of shipping compiled stylesheets

jimmycallin:

`@rjsf/shadcn` compiled ten full Tailwind builds from `src/css/<theme>.css`, committed the output to `resources/`, and documented it as a consumer import (`import '@rjsf/shadcn/dist/<theme>.css'`). Those files served one real purpose: the playground swaps a single `<link>` at runtime and fetched them from `cdn.jsdelivr.net/gh/...@main`. That is why 370 KB of build output was committed to git, and why it drifted from the source whenever someone forgot to rebuild.

For a library consumer they were the wrong artifact in every dimension:

- each one is a **complete Tailwind build including preflight**, so importing it resets the rest of the page;
- each one carries **one frozen token set**, so it cannot follow the application's own shadcn theme;
- the utility set is **fixed at rjsf's build time**, so a consumer's `ui:classNames` or custom widget classes are missing.

On this branch they had also fallen out of `files` when the output moved from `dist/` to `resources/`, so nothing was being published — the documented import failed on `main` for one reason (no `exports` entry, fixed in #5286) and would have failed on `v7` for another.

This is also the shape both Tailwind and shadcn recommend. Tailwind's guidance for libraries is to ship source and have the consumer's build scan it via `@source`, not to bundle utilities ([docs](https://tailwindcss.com/docs/detecting-classes-in-source-files), [discussion](https://github.com/tailwindlabs/tailwindcss/discussions/18545)). shadcn's model is that the application owns the design tokens.

## What this PR does

**`@rjsf/shadcn`**

- Removes `src/css/`, `resources/`, `build-css.ts`, the `build:css` script and the `@tailwindcss/cli` / `tailwindcss` dev dependencies. `build` is now just the `tsdown` run every other package has.
- Rewrites the README theming section: one `@source "../node_modules/@rjsf/shadcn/lib"` line, the pnpm symlink caveat, and the CSS-variable contract the components rely on. The "Dark mode" and "RTL" sections drop their "(Tailwind only)" qualifier since Tailwind is now the only path.
- Adds a `BREAKING CHANGE` section to the v7 upgrade guide and a note to `themes.md` that `@rjsf/shadcn` ships no stylesheet.

**Playground**

- The token sets move to `packages/playground/src/themes/shadcn/<theme>.css`. The `@theme inline` mapping — byte-identical across all ten files — moves into one shared `base.css` along with `@import "tailwindcss"` and the `@source` pointing at the aliased shadcn source. Each theme is ~50 lines shorter.
- `@tailwindcss/vite` compiles them during the playground build; `app.tsx` references each through a `?url` import instead of a jsDelivr URL. No more committed build output, no more branch-pinned CDN fetch.
- `clean-slate` existed as a theme but was never listed in the switcher; it is now.

**Housekeeping**

- Drops the two `knip.jsonc` entries that existed only for `resources/*.css` and the `@tailwindcss/cli` binary.
- The removed build script passed `--content` globs to the Tailwind v4 CLI, which has no such flag (`tailwindcss --help` lists `-i -o -w -m --optimize --cwd --map`); rebuilding without it produced byte-identical output, so nothing is lost.

## Verification

- `pnpm run build` in the playground emits one compiled asset per theme (`default-*.css` 39 KB); checked that it contains the `:root` and `.dark` token blocks, utilities from `button.tsx` (`.size-9`), preflight, and soft-pop's `body { letter-spacing }` rule, and that `index-*.js` references the hashed asset URL.
- `@rjsf/shadcn`: 153 tests pass. Snapshots are unaffected — they never included CSS.
- Root `pnpm run typecheck`, `pnpm run knip` (no findings, no hints), `lint` and `cs-check` on both touched packages: clean.

## Checklist

- [x] I've updated the upgrade guide (this branch documents breaking changes there rather than in `CHANGELOG.md`)

## Follow-up

The remaining structural issue is that `@rjsf/shadcn` vendors its own `components/ui/*` and Radix primitives, so a consumer's customized shadcn `button.tsx` is ignored — only CSS variables carry over. The idiomatic fix is publishing a shadcn registry (`registry.json` at the repo root; items declare `registryDependencies: ["button", "input", ...]` and the CLI resolves them against the user's own `components/ui`). That is a separate PR.


heath-freenome:
## Code review findings

**1. `packages/playground/src/themes/shadcn/base.css:1`** — consolidating all ten theme CSS files' `@theme inline` blocks into one shared `base.css` silently drops the soft-pop-only tracking-scale overrides (`--tracking-tighter/tight/normal/wide/wider/widest`, all derived from `calc(var(--tracking-normal) ...)`) that existed in the old `packages/shadcn/src/css/soft-pop.css`.

soft-pop's `:root { --tracking-normal: normal; }` (unique among the themes — every other theme uses `0em`) previously drove Tailwind's `tracking-tight`/`tracking-widest` utility classes via the theme's own `@theme inline` mapping. `alert.tsx` uses `tracking-tight` and `command.tsx` uses `tracking-widest` — with the mapping gone from the new shared `base.css`, those utilities now resolve to Tailwind's built-in default tracking scale instead of soft-pop's, a visible letter-spacing regression for the soft-pop theme wherever those components render. Since the migration guide points consumers at these same playground CSS files as the token source for hand-rolled themes, a consumer copying soft-pop's tokens loses the same override.

**2. `CHANGELOG-v7.md:100`** — the existing v7 changelog entry for `@rjsf/shadcn`/Dev still describes `build-css.ts` and the `build:css` script, both of which this PR deletes, so the changelog now documents functionality that no longer exists in the release it describes.

A reader of `CHANGELOG-v7.md` sees "`@rjsf/shadcn`'s `build:css` script runs `build-css.ts` with `node` directly" right next to the new migration-guide entry (added by this same PR) saying `build-css.ts`/`src/css`/the compiled stylesheets are gone — the two release documents contradict each other for the same v7 release.

---

Otherwise this is a small, well-contained change (31 files, +220/-933). I traced the Vite `?url` import / CSS asset pipeline, `knip.jsonc` and `package.json` dependency changes, and the README/migration-guide docs, and diffed every theme CSS file's extracted `@theme inline` block against the new `base.css` looking for silent data loss during the consolidation — that check is what turned up finding 1.

jimmycallin:
Both fixed in `27583aca9`.

**1. soft-pop's tracking scale** — confirmed, and worth noting the exact shape of it: soft-pop was the only theme carrying the six `--tracking-*` entries in its `@theme inline` block, and the only one whose `--tracking-normal` is `normal` rather than `0em`. For the other nine the block would have been a no-op — `calc(0em ± 0.05em)` and friends reproduce Tailwind's built-in scale exactly — so consolidating lost nothing there, but soft-pop's `calc(normal ± …)` is invalid at computed-value time, which is what neutralized `tracking-tight`/`tracking-widest` for that theme. Dropping the block swapped that neutral behaviour for Tailwind's real scale.

Since the whole mapping is a `calc()` over a token every theme defines, it belongs in the shared block rather than in one theme's file. `base.css`'s `@theme inline` now carries it, and because `inline` resolves the `var()` at the use site, each theme's own `--tracking-normal` still drives it. Verified in the built output: `tracking-tight` emits `letter-spacing: calc(var(--tracking-normal) - .025em)` in every theme's stylesheet, so soft-pop is back to what it rendered before and the nine others are unchanged.

**2. `CHANGELOG-v7.md`** — the `tsx` bullet no longer explains itself through a script this PR deletes; it just says a TypeScript script runs under `node` directly, which is the actual reason the dependency went.

While there, the PR was missing its own changelog entries, which the repo requires — added a `@rjsf/shadcn` **BREAKING CHANGE** entry for dropping the compiled stylesheets and a Dev entry for the playground token files.

`knip` and the playground build pass.

heath-freenome (approved):

PR #5289, opened 2026-09-15
feat(core): add getFormData() and the FormHandle ref contract

jimmycallin:

First additive step of the form data ownership work in `rfcs/form-state-ownership-spec.md` (§7, PR 1, fork PR jimmycallin#62). No behavior changes; everything here merges independently of the rest of that plan and of #5279, #5280 and #5285 (it touches no `Form.tsx` range those PRs edit).

### Why this exists

`Form` is moving to strict data ownership: a form will be either **parent-owned** (`formData` prop, updates accepted through `onChange`) or **self-owned** (`initialFormData`, the form keeps the value). Today `state.formData` is a reconciled copy whose job is to arbitrate between those two owners on every update — `getSnapshotBeforeUpdate` compares props against it to guess whether new props were an echo of its own `onChange`. Once a form has exactly one owner there is no copy to read, and that reconciliation is deleted.

Consumers therefore need a read path that survives the switch, and self-owned forms need a read path at all: between `onChange` calls their value is unreachable without mirroring every keystroke into parent state, which defeats the point of letting the form own it (autosave, route guards, a submit button outside the form). Every store-owning form library ships this getter; Mantine added `form.getValues()` for exactly the uncontrolled-mode reason.

### What it replaces

| Old | New | Status |
| --- | --- | --- |
| `formRef.current.state.formData` | `formRef.current.getFormData()` | deprecated now, `state.formData` stops being the current value in the ownership PR |
| `formRef.current.state.errorSchema` / `.errors` | `event.errorSchema` / `event.errors` from `onChange`/`onSubmit` | deprecated now |
| ref exposing the whole class (`state`, `setState`, lifecycle) | `FormHandle`: `getFormData`, `submit`, `reset`, `setFieldValue`, `validateForm`, `validateFormWithFormData`, `validate`, `focusOnError` | `Form implements FormHandle`; everything else is visibly unsupported |

### What this PR does

- `getFormData(): T | undefined` on `Form`. Under the current hybrid it returns `this.state.formData` in both modes, per spec §7 PR 1 ("must not claim the future ownership contract yet"); the ownership PR changes its *source*, not its meaning.
- `FormHandle<T, S, F>` exported from `@rjsf/core`; the class `implements` it so the compiler enforces the contract.
- TSDoc on `getFormData()` states the committed-read timing (an edit or `setFieldValue()` in the same tick is not yet visible) — required by spec §3.5 because every other library's getter is read-your-writes.
- v7 upgrade guide: new deprecation section with the why and the migration diff; new feature section. `internals.md`: handle and getter sections.

### One deviation from the spec, and why

§3.6 asks for `ref?: Ref<FormHandle>`. That does not typecheck while `Form` is a class: TSX types a class element's `ref` by the instance (`IntrinsicClassAttributes<Form>`), so `useRef<FormHandle>()` is rejected on `<Form ref>` and `withTheme`'s forward breaks the same way. The prop stays `Ref<Form>` with a TSDoc explaining this, and the docs show the interim pattern — a class-typed ref narrowed to `FormHandle` at the use site. The prop type change becomes possible exactly when `Form` becomes a function component, which is the deferred milestone the handle exists to carry. No casts were used.

### Not included

- The `IChangeEvent`/`FormState` decoupling from spec PR 1 — it edits the same `Pick<FormState, …>` line #5280 rewrites, and is a ten-line change once that lands.
- A private mode resolver — one line with no consumer; it lands with the ownership switch.
- `CHANGELOG.md` — the `v7` branch records changes in the upgrade guide rather than per-PR changelog entries.

## Checklist

  - [x] I've added and/or updated tests (`packages/core/test/Form.handle.test.tsx`, runtime and `expectTypeOf` surface checks)
  - [x] I've run `pnpm run typecheck`, `pnpm run lint`, `pnpm run knip`, `pnpm run cs-check`, and the `@rjsf/core` suite (1990 tests)


packages/core/src/components/FormHandle.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Only the members listed here are supported. Everything else on the class instance may change without notice.
+ */
+export interface FormHandle<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any> {
+  /** Returns the form data the `Form` currently renders: the value the parent supplied for a parent-owned form, or the
+   * value committed internally for a self-owned form. It is the read path for a self-owned form, whose data is not
+   * otherwise reachable between `onChange` calls (autosave, route guards, a submit button outside the form).
+   *
+   * It reads committed data only. An edit or `setFieldValue()` in the same tick is not visible until React commits
+   * it, and a proposal a parent-owned form has emitted but the parent has not yet accepted is never returned. Treat the
```
heath-freenome: This claim doesn't hold today. `getFormData()` reads `this.state.formData`, and `processPendingChange()` commits the edit to state *before* calling `props.onChange` — so if the parent ignores the proposal, props stay deep-equal, `getSnapshotBeforeUpdate` returns `{ shouldUpdate: false }`, and the un-accepted value stays. Confirmed against this branch: a `<Form formData={{name:'parent'}} onChange={() => {}} />` typed into returns `{ name: 'typed' }`.

Line 19's "the value the parent supplied for a parent-owned form" is wrong for the same reason. Since the PR body is explicit that this PR must not claim the ownership contract yet, maybe describe it as just "the value the form currently renders" and add the pending-proposal guarantee when ownership lands.

packages/core/src/components/FormHandle.ts
```diff
@@ -0,0 +1,56 @@
...
+} from '@rjsf/utils';
+
+/** The imperative surface a `Form` exposes through its `ref`. It is the supported alternative to holding a ref to the
+ * `Form` class instance, whose `state` and lifecycle are internals rather than API. A `Form` will be either
+ * parent-owned (a `formData` prop, accepted through `onChange`) or self-owned (seeded by `initialFormData`); this
+ * handle is the same for both, and it is the contract a later function-component `Form` keeps.
+ *
+ * Only the members listed here are supported. Everything else on the class instance may change without notice.
+ */
+export interface FormHandle<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any> {
```
heath-freenome, line 18: `formElement` is a public class field and the v5 upgrade guide (line 144) actively documents `formRef.current.formElement.current.reset()`, but it's absent from `FormHandle` and the new text says "Nothing else on the instance is supported" — with no row for it in the "Stop using / Use instead" table. Worth either adding it to the handle or giving it a migration row (native reset → `reset()`; direct DOM access → ?), so people on it get a signal rather than a surprise.

packages/core/src/components/Form.tsx
```diff
@@ -1075,6 +1077,13 @@ export default class Form<
     return schemaUtils.omitExtraData(schema, formData);
   }
 
+  /** Returns the form data currently rendered, see `FormHandle.getFormData()`. Until strict ownership lands this is the
+   * reconciled internal value in both modes; afterwards it reads the owner directly.
+   */
+  getFormData(): T | undefined {
```
heath-freenome: This is an unbound prototype method while `submit`, `reset`, `setFieldValue` and `validateFormWithFormData` are all arrow class properties, so destructuring the handle breaks on exactly the new member:

```tsx
const { submit } = ref.current!;      // fine
const { getFormData } = ref.current!;
getFormData();                        // TypeError: Cannot read properties of undefined (reading 'state')
```

Suggest `getFormData = (): T | undefined => this.state.formData;` for consistency — a future function-component `FormHandle` will hand back bound closures, so leaving it unbound means this silently changes behavior later. (`validateForm` and `focusOnError` have the same shape; pre-existing, but they're now advertised as handle members too.)

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -339,6 +379,28 @@ Note that `@rjsf/core` applies `ui:emptyValue` for you when the input is cleared
...
+  return () => clearInterval(timer);
+}, []);
+
+<Form ref={formRef} schema={schema} validator={validator} initialFormData={record} />;
+```
+
+Two things to know:
+
+- **It reads committed data.** An edit or `setFieldValue()` in the same tick is not visible until React commits it, so read after the commit (in tests, after `await act(...)`).
+- **With `formData` it returns the value you passed.** A controlled form never has data you do not already hold; a pending `onChange` you have not accepted is not returned.
```
heath-freenome: This isn't true on this branch — see the `FormHandle.ts` comment. A controlled form whose parent rejects or filters the change keeps the proposal in `state.formData`, and `getFormData()` returns it. The consequence is concrete: an autosave or "unsaved changes" guard built on this will persist a value the consumer's own state never contained, which is exactly what this bullet promises can't happen.

The identical sentence is at `packages/docs/docs/advanced-customization/internals.md:102`.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -2,6 +2,46 @@
 
 ## New deprecations
 
+### Reading `Form` internals through a `ref`
```
heath-freenome, line 5: This table announces `.state.formData`, `.state.errorSchema`, `.state.errors`, `.state.schemaUtils` and `.state.registry` as deprecated now, but no `@deprecated` tags were added to those `FormState` members (the only one in `Form.tsx` is on `noValidate`). Without them, consumers get no editor strikethrough or build signal during the window the guide is trying to give them — the first notice is the behavior change when ownership lands.

packages/docs/docs/advanced-customization/internals.md
```diff
@@ -73,6 +73,34 @@ i.glyphicon {
...
+
+```tsx
+import { useRef } from 'react';
+import type { FormHandle } from '@rjsf/core';
+import Form from '@rjsf/core';
+import { RJSFSchema } from '@rjsf/utils';
+import validator from '@rjsf/validator-ajv8';
+
+const schema: RJSFSchema = { type: 'object', properties: { title: { type: 'string' } } };
+const formRef = useRef<Form>(null);
```
heath-freenome: `useRef` is called at module scope here, next to a bare `<Form … />;` expression statement — pasted as shown this throws "Invalid hook call". The neighbouring existing examples in this file use `createRef` for that reason. Same issue in the upgrade guide snippet, which calls `useEffect` at top level.

Also, line 88's `import { RJSFSchema } from '@rjsf/utils'` should be `import type` under the repo's `verbatimModuleSyntax` style.

heath-freenome (commented):
## Review: `getFormData()` and the `FormHandle` contract

Nice shape overall — carving the supported surface out as a named type before the ownership switch is the right sequencing, and the `expectTypeOf` test pinning what is *not* on the handle is a good touch.

Two findings I confirmed empirically against `baad1d9` with throwaway probe tests (since deleted); the rest are read off the diff. The `@rjsf/core` suite passes and oxlint is clean on the changed files.

### 1. The documented contract doesn't hold yet (inline on `FormHandle.ts`)

The PR body says this PR "must not claim the future ownership contract yet" — but the TSDoc and the docs do claim it. `getFormData()` returns `this.state.formData`, and `processPendingChange()` commits the edit to `state.formData` *before* calling `props.onChange`. If the parent ignores or rejects the change, props stay deep-equal, `getSnapshotBeforeUpdate` returns `{ shouldUpdate: false }`, and the local value persists:

```tsx
const ref = createRef<Form>();
render(<Form ref={ref} schema={schema} validator={validator} formData={{ name: 'parent' }} onChange={() => {}} />);
fireEvent.change(container.querySelector('input')!, { target: { value: 'typed' } });

ref.current!.getFormData(); // → { name: 'typed' }, not { name: 'parent' }
```

So "a proposal a parent-owned form has emitted but the parent has not yet accepted is never returned" (`FormHandle.ts:24`) and "**With `formData` it returns the value you passed.**" (upgrade guide + `internals.md:102`) are both false today. This is the finding with real user consequences: an autosave or route guard on a controlled form that filters changes will persist a value the consumer's own state never held. Suggest describing it as "the value the form currently renders" until the ownership PR lands.

### 2. `getFormData()` is unbound while the rest of the handle is bound (inline on `Form.tsx`)

`submit`, `reset`, `setFieldValue` and `validateFormWithFormData` are arrow class properties; `getFormData` is a prototype method. Destructuring — the normal way to consume something called "the imperative handle" — breaks on exactly the new member:

```tsx
const { getFormData } = ref.current!;
getFormData(); // TypeError: Cannot read properties of undefined (reading 'state')
```

### 3. Not anchorable inline — the guide contradicts itself

`packages/docs/docs/migration-guides/v7.x upgrade guide.md:230` (outside the diff) still says the migration path off `Form.omitExtraData()` is "the `schemaUtils` is available via `formRef.current.state.schemaUtils`" — the exact member the new table at the top of the same file now deprecates. Worth pointing only at `createSchemaUtils` / the `registry` prop.

Remaining smaller points are inline.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

packages/core/src/components/FormHandle.ts
jimmycallin: Fixed — the TSDoc now says `getFormData()` returns what the form renders, and spells out that a parent-owned form keeps a reconciled copy so a declined change is returned until the parent re-renders it away. The pending-proposal guarantee moves to the ownership PR, where the copy is gone.

packages/core/src/components/FormHandle.ts
jimmycallin: Added a `.formElement` row to the table: `reset()` for a form reset, no replacement for direct DOM access. Keeping it off `FormHandle` — a DOM ref is exactly what a function-component `Form` should not be obliged to hand back.

packages/core/src/components/Form.tsx
jimmycallin: Fixed: `getFormData`, `validate`, `validateForm` and `focusOnError` are all bound class properties now, so every advertised member survives destructuring. New test covers it.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Reworded here and in `internals.md:102` to match the corrected TSDoc.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Tried it, and it does not work yet — two blockers, both verified on this branch:

1. `IChangeEvent` is `Pick<FormState, ... | 'formData' | 'errors' | 'errorSchema'>`, so a tag on those `FormState` members propagates to `event.formData`/`event.errorSchema` — the replacements this very table points at.
2. `typescript/no-deprecated` is `error` for `.ts`/`.tsx`, and a tag on `FormState.formData` alone produced ~20 lint errors inside `Form.tsx` on its own reads.

Deprecating `Form.state` as a whole has the same lint problem. The distinction that matters (reading these through `Form.state` is deprecated; reading them off the change event is not) is not expressible while the two types share members. The guide now says outright that there is no compiler signal yet and that the tags follow once `IChangeEvent` stops deriving from `FormState` — the decoupling this PR deferred.

packages/docs/docs/advanced-customization/internals.md
jimmycallin: Fixed: the `internals.md` snippet uses `createRef` like its neighbours, `RJSFSchema` is an `import type`, and both upgrade-guide snippets are wrapped in a component so `useRef`/`useEffect` are called from a hook context.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -2,6 +2,51 @@
...
+
+The `Form` ref now has a documented contract: the `FormHandle` type, exported from `@rjsf/core`. Its members are unchanged and stay supported. Anything else on the instance is an implementation detail and will break in a later 7.x release, so stop using it now:
+
+| Stop using                              | Use instead                                                               |
+| --------------------------------------- | ------------------------------------------------------------------------- |
+| `.state.formData`                       | `.getFormData()`, new in this release (see New Features)                  |
+| `.state.errorSchema`, `.state.errors`   | `event.errorSchema`, `event.errors` in your `onChange`/`onSubmit` handler |
+| `.state.schemaUtils`, `.state.registry` | the `registry` prop your field, widget or template already receives       |
```
heath-freenome: This replacement doesn't work for this row's own audience. `registry` is only injected as a prop into components rendered *inside* the form (fields/widgets/templates) — someone holding a `ref` to `Form` from outside has no way to reach "the `registry` prop", since they're not one of those components. I grepped and confirmed `registry` lives only on `this.state.registry`; there's no other accessor on `Form`/`FormHandle` that surfaces it to a ref holder.

The equivalent migration note for `Form.omitExtraData()` further down this same file was fixed to add `createSchemaUtils(...)` as the standalone option ("build one with `createSchemaUtils`, or use the `registry.schemaUtils` your fields, widgets and templates already receive") — worth giving this row the same two-part answer, since a ref holder needs the `createSchemaUtils` half and a field/widget/template author needs the `registry` half.

heath-freenome (commented):
Follow-up on the fixes from the last round — the false-contract claim, the unbound `getFormData()`, the missing `formElement` migration row, and the invalid-hook-call snippets are all fixed and verified (ran the `@rjsf/core` suite: 2052 passed, oxlint clean on the changed files, build succeeds). One new item from this pass, inline below.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Right — a ref holder can't reach `registry`. Split the row in two and gave `schemaUtils` the same two-part answer as the `Form.omitExtraData()` note:

| Stop using | Use instead |
| --- | --- |
| `.state.schemaUtils` | build one with `createSchemaUtils(validator, schema)`, or use the `registry.schemaUtils` a field, widget or template receives |
| `.state.registry` | the `registry` prop a field, widget or template receives; a ref holder has no replacement |

Saying outright that there's no ref-holder replacement for `registry` rather than inventing one — it's the form's own render-time object, and nothing on `Form`/`FormHandle` surfaces it, as you found.

Checked the rest of the diff for the same mistake: the only other spot pointing at `registry` is the `Form.getRegistry` removal note, and that one already splits by audience (field/widget/template → the `registry` prop, tests → `getTestRegistry()`, dictionaries → `generateTheme()`). No other rows send an outside-the-form caller at an in-form accessor.

heath-freenome: Did we want to give the user access to the registry? Some people may be using it

jimmycallin: Access, yes — a member on `FormHandle`, I'd rather not, and I think we already have the access covered.

The thing that makes `registry` different from the other rows in this table is *who it's a contract with*. `getFormData()`, `submit()`, `reset()` are contracts with the application: someone outside the form calls them, so their shape is chosen for that audience and changes to them are deliberately breaking. `Registry` is a contract with the *render tree* — the plumbing `Form` hands down to fields, widgets and templates. Its shape is chosen for their convenience, and it grows whenever a new rendering concern needs threading through. In v7 alone it gained `globalFormOptions`, `globalUiOptions` and `uiSchemaDefinitions`.

That's fine while the only consumers are components we render, because they receive it as a prop and never construct one: we can add a field, or change what an existing field resolves to, and nothing outside breaks. The moment a ref holder can read it, that freedom is gone — any change to the registry is now potentially a breaking change to the imperative API, and every internal rendering refactor has to be argued as a public API change. Which is exactly the coupling this PR exists to remove for `state`.

There's a second, smaller reason: a registry read through a ref is a snapshot. It's rebuilt from props on every `getDerivedStateFromProps`, so whatever a caller stashed goes stale the next time props change, in a way that's quiet rather than loud.

**The good news is I don't think a ref holder is actually stuck.** `buildRegistry()` is a public export in v7 (it's the old `Form.getRegistry()` static — there's a row for it further down this guide), and `Form` itself does nothing more than call it:

```ts
// packages/core/src/components/Form.tsx
const newRegistry = buildRegistry(props, rootSchema, schemaUtils);
```

All three inputs are things the caller already owns, so outside the form it's:

```ts
const schemaUtils = createSchemaUtils(validator, schema, experimental_defaultFormStateBehavior, experimental_customMergeAllOf);
const registry = buildRegistry(props, schemaUtils.getRootSchema(), schemaUtils);
```

So I'll fix this row to say that instead of "a ref holder has no replacement" — same two-part shape as the `state.schemaUtils` row above it. That leaves `buildRegistry()` as the documented path and keeps the handle narrow.

And the asymmetry is in our favour: adding `getRegistry()` to `FormHandle` later is non-breaking, removing it isn't. So if someone comes back with a use case `buildRegistry()` genuinely can't cover, it's a few lines to add then, with the real use case to shape it.

That said — happy to just add it now if you think it's needed. Say the word and it goes in this PR.

heath-freenome: Hmmm, I'm fine with holding off for now. And if, like you say, someone pushes a valid use case, we will likely want to find a way to make it immutable.

Perhaps we can try using `deepfreeze` on it when we create it and see whether it has any negative effects? If it does not, then returning a frozen registry in the future would definitely prevent someone from tampering. And maybe another breaking change to make `registry` read-only? This would definitely flow downhill into the utils layer, making all places then need to be read-only too. It doesn't have to happen in this PR though. Perhaps a follow-up

heath-freenome (approved):
@jimmycallin I'll leave the PR for you to merge in case you want to try the frozen `registry` idea here. Otherwise, let's definitely try it in a new PR

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Agree on holding off.

On the follow-up, I'd steer away from a runtime `deepfreeze` and toward `readonly` on the `Registry` type instead. The registry is mostly a shell around objects the caller owns and passed in by reference, so freezing it would freeze theirs, not ours:

- `formContext` is the user's object; people keep stores and class instances in it, and in strict-mode ESM an assignment into a frozen object throws.
- `rootSchema` is the user's schema object by reference on the non-2020-12 path.
- `schemaUtils.validator` is the AJV instance, which keeps mutable compile caches on itself.
- `fields` / `widgets` / `templates` hold user components; `React.memo` and `forwardRef` return plain objects that tooling sets properties on.
- `buildRegistry` runs on every `getDerivedStateFromProps`, so a deep walk would run on every prop change.

"Try it and see if anything breaks" would be measured against our suite, which never mutates `formContext` or exercises AJV caching across renders, so green there wouldn't mean safe.

`readonly` members give the same tamper protection at compile time with no runtime cost and no reach into user objects (`globalFormOptions` already has it). The downhill effect is smaller than it sounds: TypeScript lets a readonly type flow into mutable-typed variables, so only code that *assigns into* the registry breaks, and only the two places that construct one (`buildRegistry`, `getTestRegistry`) need touching.

For the "does anything mutate it" question, I'd answer it in tests rather than production: shallow-freeze the registry `getTestRegistry()` returns, plus its three component maps. That catches internal mutation the same way utils already uses `deepFreeze` on its schema fixtures, with zero production risk.

I'll open an issue for both so it's tracked separately from this PR.

heath-freenome: Great plan. We can address them a bit later on in the v7 process. Do add them to the V7 project though... Or let me know their numbers and I will

jimmycallin: Opened #5296 with this rather than an issue, since it turned out to be a small change.

It does the `readonly` version plus the shallow freeze on `getTestRegistry()`, and the rationale for preferring that over `deepfreeze` is in the description. Worth noting it caught exactly one violation across the repo, a core test assigning `registry.globalUiOptions` after construction, which I think is the reassuring result: nothing else writes into a registry today.

PR #5296, opened 2026-09-18
refactor(utils)!: make the Registry read-only and freeze the test registry

jimmycallin:

Follow-up to a [review discussion on #5289](https://github.com/rjsf-team/react-jsonschema-form/pull/5289#discussion_r4043833376). We agreed to keep a `getRegistry()` off `FormHandle` for now, and the open question was how to keep a registry from being tampered with if a ref holder ever gets one.

The suggestion there was a runtime `deepfreeze` on the registry when it is created. I think that is the wrong tool, because the registry is mostly a shell around objects the caller owns by reference:

- `formContext` is the user's object. People keep stores and class instances in it, and in strict-mode ESM an assignment into a frozen object throws rather than being ignored.
- `rootSchema` is the user's schema object by reference on the non-2020-12 path (`createSchemaUtils`).
- `schemaUtils` holds the AJV instance, which keeps mutable compile caches on itself.
- `fields`, `templates` and `widgets` hold user components. `React.memo` and `forwardRef` return plain objects that tooling sets properties on.

`buildRegistry()` also runs on every `getDerivedStateFromProps`, so a deep walk would run on every prop change.

"Try it and see if anything breaks" would be measured against our own suite, which never mutates `formContext` and never exercises AJV caching across renders, so a green run would not mean it is safe.

`readonly` members give the same tamper protection at compile time. They cost nothing at runtime and never reach into caller-owned objects.

## Changes

- Every member of `Registry` in `@rjsf/utils` is now `readonly`. `globalFormOptions` already was.
- `getTestRegistry()` shallow-freezes what it returns, plus its `fields`, `templates` and `widgets` maps, so a test that mutates the registry fails loudly instead of quietly affecting later assertions. Shallow by design, for the same reasons the deep freeze is rejected above.
- Documented both in the 7.x upgrade guide under Breaking changes.

Readonly modifiers do not affect assignability, so only code that *writes into* a registry breaks, and a spread copy drops the modifiers if someone genuinely needs a modified registry. Across the whole repo the change caught exactly one violation: a core test assigning `registry.globalUiOptions` after construction, now building a copy instead. That is the useful signal here, that nothing else writes into a registry today.

The same argument applies one level out, so a second commit makes `schema`, `uiSchema` and `registry` `readonly` on `RJSFBaseProps`, which every field, widget and template props type extends, and on `SubmitButtonProps`, which declares its own copies. These are values the form builds and passes down, and writing back into them never reached the form. It also closes a gap the first commit leaves, since `props.registry = someOtherRegistry` still compiled.

Two things I checked before proposing that. `readonly` survives the `GenericObjectType` index signature that `FieldProps` and `WidgetProps` inherit, for both dot and bracket access. And object spread drops the modifiers, so a caller who wants a modified copy still writes one line.

That found one real mutation. `ArrayField` built `arrayProps` and `arrayAsMultiProps`, then patched `.schema` on both when a schema had no `items`. It now decides the schema first and builds each object with it, dropping the mutation and the duplicated assignment. Behavior is unchanged and the snapshot suite confirms it.

Deliberately out of scope: the component maps are not deeply read-only, so `registry.widgets.MyWidget = ...` still compiles. Those types double as the user-facing `FormProps.fields`/`widgets`/`templates`, so a readonly index signature would break anyone assembling a map before passing it. The runtime freeze covers our own code, and the guide documents the boundary.

## Checklist

  - [x] I've added and/or updated tests. `packages/core/test/testing.test.ts` covers both halves: mutating the test registry or its component maps throws, and a caller's `formContext` stays unfrozen and passes through by reference.
  - [ ] I've updated the changelog. The 7.x work records breaking changes in the upgrade guide, which has no changelog section yet, matching #5289.
  - [x] I've verified the change: `lint`, `knip`, `build`, `typecheck` and `test` all pass across every package.
- [ ] I'm changing dependencies


heath-freenome:
In `packages/core/src/testing.ts` (around line 34), `getTestRegistry()`'s runtime freeze only deep-freezes the nested `ButtonTemplates` map, not other nested template groups that `TemplatesType`'s catch-all type (`& Record<string, ComponentType<any> | Record<string, ComponentType<any>> | undefined>`) permits.

If a theme or test adds a second nested template group (e.g. `templates.CustomButtons = {...}`), `Object.freeze(registry.templates)` only protects top-level keys, so `registry.templates.CustomButtons.Foo = () => null` would silently succeed instead of throwing — contradicting the code comment's stated goal of making a mutating test "fail loudly". Currently latent since no in-repo theme defines such a second group, but worth a recursive/deep freeze if the guarantee should hold for any valid `TemplatesType`.

heath-freenome (approved):
One small change proposed, but not required to be accepted to merge

jimmycallin:
Good catch — fixed in `cc892a1d2`. The freeze now walks `Object.values(registry.templates)` and freezes every nested group rather than naming `ButtonTemplates`, so any future template group is covered without another edit. The freeze still stops there: caller-owned `formContext`/`rootSchema` and the validator caches stay untouched, and the comment says so.

PR #5297, opened 2026-09-18
refactor(core)!: shallow comparison only, with structural sharing instead of deep compare (#5217 part 2)

jimmycallin:
Closes #5217.

Second half of #5217. Part 1 (#5280) made field identity primitive so that reference comparison *could* work. This PR removes the deep comparison that compensated for unstable references, and does the reference-stability work that makes shallow comparison sufficient. Base: `v7`.

## 1. The problem, from first principles

React decides whether to re-render a memoized component by comparing its props. There are two ways to make that decision:

- **Compare by reference** (`memo`, `PureComponent`): O(number of props), and it only works if unchanged data arrives as the *same object* it was last time.
- **Compare by value** (a custom `memo` comparator running `deepEquals`): O(size of the data), and it works even if unchanged data arrives as a *new object* that happens to be equal.

RJSF has used the second, at two places: `SchemaField`'s memo comparator, and `Form.shouldComponentUpdate`, both running `deepEquals` over **all** props on every render, behind `experimental_componentUpdateStrategy`. It was chosen because RJSF's object props were rebuilt on every change: `Form` handed every `onChange` a `structuredClone` of the whole formData tree, validation rebuilt `errorSchema` from scratch, `retrieveSchema()` returned a fresh object per call, and a prop update rebuilt the entire state. Deep-equal-but-new references everywhere, which reference comparison cannot see through.

Compare-by-value has two defects that are matters of fact, not preference:

1. **It is incorrect for functions.** `deepEquals` treats every function as equal to every other. A replaced `validator`, `customValidate`, `transformErrors`, template, widget or field was silently ignored by the update gate. This is a bug class the strategy option could not fix, only widen (`'always'`).
2. **It does not compose.** A deep compare at one component cannot stop its children from receiving fresh references. Every level of the tree pays its own full walk, so the cost is `deepEquals(props)` × fields × renders, which is [the pattern the React docs warn against](https://react.dev/reference/react/memo#specifying-a-custom-comparison-function).

Compare-by-reference has one requirement: producers must keep the reference of anything that did not change. Meet it once, at the producers, and every consumer in the tree gets an O(props) decision for free. That is the design this PR moves to, and it is the design `memo`, `PureComponent`, Redux, Immer and TanStack Query all assume.

## 2. What changes

**Shallow comparison becomes the only behavior.** `SchemaField` is `memo(SchemaFieldRender)` with memo's default comparison. `Form` extends `PureComponent`. `experimental_componentUpdateStrategy`, `shouldRender()` and `shallowEquals()` are gone. Shallow-equal implies deep-equal, so relative to `customDeep` this can produce *more* renders but never a missed update.

**Unchanged data keeps its references**, using structural sharing, the operation TanStack Query's [`replaceEqualDeep`](https://github.com/TanStack/query/blob/main/packages/query-core/src/utils.ts) performs on every fetch result:

- **`replaceEqualDeep(prev, next)`**, new in `@rjsf/utils`: returns `next` with every subtree deeply equal to its counterpart in `prev` replaced by the `prev` instance, and `prev` itself when nothing changed. Plain objects, arrays, equal-valued `Date`s and React elements of the same type, key and props are shared; other object types are opaque. Functions compare by identity. Non-mutating, one O(n) walk, cycle-safe.
- **Every `setState`** in `Form` goes through one `setSharedState`, which runs the new state through `replaceEqualDeep` against the current state inside the updater. One walk per keystroke at the root replaces one `deepEquals` per field per render.
- **The prop path.** `getSnapshotBeforeUpdate` shares the incoming `formData` prop against the state's own, so a controlled parent echoing the form's value back costs one identity check, and shares the derived state against the previous state, so `state.schema`, `state.uiSchema` and the registry keep their references whether or not the parent's objects did. `shouldUpdate` and the "did formData change" decision become identity checks.
- **The prop boundary.** Props whose values may hold functions (`IDENTITY_PROP_KEYS`: schema, uiSchema, validator, the callbacks, fields, widgets, templates, formContext, translateString, nameGenerator) are compared with `replaceEqualDeep`, functions by identity. A changed template or validator now takes effect, and a controlled form with live validation now re-validates when `validator`, `customValidate` or `transformErrors` changes. Any other prop change still re-derives state through `deepEquals`, preserving the controlled snap-back (see tradeoffs).
- **`retrieveSchema()` structural sharing.** Each `SchemaUtils` instance keeps a `WeakMap` from input schema to its last result, one per `resolveAnyOfOrOneOfRefs` value. Every call resolves afresh and shares every unchanged subtree with that result, so nothing is served by identity and an in-place mutation of the data is still seen. Instance-scoped and GC-managed.
- **`SchemaField` stops minting objects.** The resolved `uiSchema` and the stripped `fieldUiSchema` are memoized. The fresh `fieldUiSchema` was the last leak: it re-created `getAvailableKey` → `handleKeyRename` → every property's `onKeyRename` per keystroke, bailing every sibling's memo. `LayoutGridField` memoizes its cell `uiSchema` the same way.

**The contract is pinned by a test.** `renderStability.test.tsx` counts renders per field through a counting `FieldTemplate` and asserts sibling fields render **exactly zero** additional times across ten scenarios: typing at the top level, in a nested object, in an array item and in a layout grid cell; a controlled parent doing `setFormData(event.formData)`, which also asserts each proposal shares its unchanged subtrees with the value it was applied to; live validation on change with a sibling's unchanged error; live validation on blur; a submit that changes no errors; a parent re-render with rebuilt but equal props including inline JSX in the `uiSchema`, an inline `formContext` and an inline callback; and a replaced `forwardRef` widget taking effect. Any future reference leak fails it loudly.

All core tests and every theme's snapshot suite pass with zero snapshot updates. One existing test changed: `doesn't cause a race condition` asserted `customDeep` render *timing*; its assertion is now order-independent and still fails if either of two near-simultaneous changes is lost.

## 3. Why this is the path to a function-component `Form`

A class `Form` with `shouldComponentUpdate` running `deepEquals` has no function-component equivalent except a custom `memo` comparator, which is the thing being removed. `PureComponent` *is* `memo` with the default comparator; the two are the same decision expressed in two component models. After this PR every render decision in RJSF is one React already knows how to make for a function component, and every piece of new logic is a pure function (`replaceEqualDeep`, the key list, `setSharedState`'s updater) that moves into `useReducer`/`useMemo` unchanged.

What still blocks the conversion is not in this PR: `Form` derives state in `getSnapshotBeforeUpdate` and commits it in `componentDidUpdate`, a reconciler between two owners of the form data. That is the subject of the [form data ownership RFC](https://github.com/jimmycallin/react-jsonschema-form/pull/62) ([spec](https://github.com/jimmycallin/react-jsonschema-form/blob/rfc/form-state-ownership/rfcs/form-state-ownership-spec.md)), whose PR 3 removes the reconciler and whose deferred-work table lists the function-component conversion as the step after that. This PR adds nothing class-only to the reconciler, keeps its `Form.tsx` footprint small on purpose so the RFC's rebase cost stays low, and the RFC's shared change pipeline already ends with `replaceEqualDeep`, as introduced here.

## 4. Breaking changes and why each is justified

1. **`experimental_componentUpdateStrategy` is removed.** Shallow comparison is the only behavior. *Why:* the option existed to choose between an incorrect gate (`customDeep`) and one that was ineffective without reference stability (`shallow`). With stability in place the choice has no remaining purpose, and an option nobody should have to reason about is better deleted than documented. The `'always'` escape hatch goes with it; it exists in React as "don't wrap in memo".
2. **`shouldRender()` and `shallowEquals()` are removed from `@rjsf/utils`**, with the `ComponentUpdateStrategy` type. *Why:* `shouldRender` is `PureComponent`'s comparison as a function; a class component extends `PureComponent` instead. `shallowEquals` duplicated `shallowEqual` from `fast-equals`, which `deepEquals` already depends on; import that.
3. **`onChange` formData shares references with the current value.** v6's `structuredClone` handed consumers a fresh tree on every change; unchanged subtrees now keep the identity they had in the value the change was applied to, which for a controlled form is the parent's own object. Code that **mutates the formData it receives** would now corrupt the parent's data: treat received formData as immutable, as with any reducer. *Why:* this sharing is the mechanism that makes sibling fields skip. It also fixes an edge: `structuredClone` throws on non-cloneable values such as functions; retained references do not, and class instances keep their prototype where `structuredClone` flattened them.
4. **`retrieveSchema()` may return the same instance across calls** when the result is unchanged. *Why:* fields compare `schema` by reference now, so the resolver has to hand back the same object for the same input. RJSF already treats retrieved schemas as immutable.
5. **Props that affect state must be stable references.** A `customValidate`, `transformErrors`, `translateString`, template, widget or field recreated on every render now re-derives state on every render, where v6 ignored function changes entirely. Wrap them in `useCallback` or hoist them. *Why:* this is the correctness fix. Ignoring function changes was the bug; the cost of honoring them lands on exactly the code that was relying on the bug, and it is the same contract `memo` and `PureComponent` impose everywhere in React. Deep-equal objects such as an inline `uiSchema` are still interned and cost one comparison.

The changelog and the v7 migration guide carry all five.

## 5. Tradeoffs in this PR, and where each is resolved

Every item below is either scheduled in the ownership RFC, which now records these decisions in its spec, or is a bounded cost accepted on purpose.

| Tradeoff | Status |
| --- | --- |
| **The per-update walk is O(state) where producers rebuild whole trees.** `replaceEqualDeep` returns on the first identity hit, so untouched subtrees cost one comparison. The cost concentrates in `getDefaultFormState`, which rebuilds formData per change, and the validator's fresh `errorSchema`. | RFC PR 2 adds reference-keyed last-call caches on those two steps, keyed by each step's explicit input set. Not done here because those inputs only exist once PR 2 extracts the steps out of `getStateFromProps`; a cache written inside that function would be rewritten by the extraction. |
| **Two comparison semantics on props.** `IDENTITY_PROP_KEYS` compares functions by identity; the `deepEquals(props)` fallback for everything else still treats them as equal. A new function-carrying prop missing from the list inherits the old behavior. | The fallback exists only to preserve the controlled snap-back and is deleted with the reconciler in RFC PR 3. PR 2 turns the key list into per-step cache keys; the replaced-callback test covers omissions. A compile-time check on the list was considered and dropped because it would fence code restructured one PR later. |
| **A new immutability contract on formData** (breaking change 3). Inherent to structural sharing: you cannot share references and also protect against mutation. | RFC PR 3 adds a development-only deep freeze of the formData handed to `onChange` and committed to uncontrolled state, plain objects and arrays only, never schemas, turning silent corruption into a throw at the mutation site. It lands there, where ownership becomes explicit, so the diagnostic ships with the contract it guards. |
| **`Form` as `PureComponent` re-renders on every parent render that passes an inline callback**, where `customDeep` skipped it. | Accepted. The cost is `Form.render` plus root reconciliation; the tree below bails on stable references, which the parent re-render scenario in `renderStability` asserts. Any fix would reintroduce function-blind comparison at the root. |
| **`retrieveSchema` keeps the last result per input schema** for the schema's lifetime. | Accepted. Two results per input schema object, GC'd with it. |
| **Non-plain objects are opaque to sharing.** A subtree containing a class instance or `File` is never shared, so nearby fields re-render on sibling changes. | Accepted ceiling. Deep-compare skipping could not handle these either. |

## 6. Review guide

Inline comments on the diff mark the hunks that are hardest to read cold: the `getSnapshotBeforeUpdate` rewrite, the `IDENTITY_PROP_KEYS` gate, the React element and symbol-key handling in `replaceEqualDeep`, and the `retrieveSchema` sharing base. Suggested order: `replaceEqualDeep.ts` and its tests, then `Form.tsx`, then `SchemaField.tsx`, then `renderStability.test.tsx`.


packages/core/src/components/Form.tsx
```diff
@@ -337,9 +326,29 @@ interface PendingChange<T> {
   id?: string;
 }
 
+/** The props state derives from whose values may hold functions or class instances, which `deepEquals()` treats as
+ * equal. These are compared with functions by identity, so a changed callback or template re-derives state and one
+ * recreated on every render re-derives it on every render; wrap those in `useCallback` or hoist them. Data-only props
+ * are caught by the deep comparison of all props.
+ */
+const IDENTITY_PROP_KEYS = [
```
jimmycallin, line 321: **Why a key list instead of comparing every prop the same way.** `deepEquals` treats all functions as equal, so it cannot be the gate for anything that may carry one. But comparing *every* prop with functions by identity would make an inline `onChange={e => ...}` in the parent re-derive state on every parent render, which for a live-validating form means a validation pass per render. So only the props that feed state derivation and may hold a function or class instance are listed here. Everything else, including the event callbacks, falls through to the `deepEquals` of all props below, which is cheap because interned references short-circuit it.

The list survives only until the ownership RFC's PR 2, where each extracted derivation step gets its own reference-keyed cache and this becomes those steps' key sets. PR 3 then deletes the `deepEquals` fallback with the reconciler.

packages/core/src/components/Form.tsx
```diff
@@ -351,6 +360,16 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
    */
   pendingChanges: PendingChange<T>[] = [];
 
+  /** `setState` sharing every unchanged subtree of `state` with the current state, so fields' memo boundaries hold
+   * across the update. The updater form keeps it correct under batching.
+   */
+  private setSharedState<K extends keyof FormState<T, S, F>>(
```
jimmycallin, line 353: **Why the updater form, and why against `prevState`.** `setState` calls can be batched, so the state we share against must be whatever React has committed by the time the update applies, not `this.state` at the call site. Inside the updater `prevState` is exactly that.

`replaceEqualDeep` only walks the keys present in `state`, so a partial update such as `{ errors, errorSchema }` never touches `formData` or the registry. When nothing in the partial changed, every value comes back as the `prevState` instance and React's own `Object.is` check on the merged state sees no change for those keys.

packages/core/src/components/Form.tsx
```diff
@@ -403,11 +422,13 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
 
     const { formData: propsFormData, initialFormData, onChange } = props;
     const formData = propsFormData ?? initialFormData;
+    const initialState = this.getStateFromProps(props, formData, undefined, undefined, undefined, true);
     this.state = {
-      ...this.getStateFromProps(props, formData, undefined, undefined, undefined, true),
+      ...initialState,
+      formData: replaceEqualDeep(formData, initialState.formData),
```
jimmycallin, line 415: The constructor shares the derived formData against the incoming one so that the `onChange` guard on the next line can be an identity check. `getStateFromProps` runs `getDefaultFormState`, which rebuilds the tree, so without this pass a seed with no missing defaults would come back as a fresh but equal object and the old `!deepEquals(...)` would have been needed to avoid a spurious mount `onChange`.

packages/core/src/components/Form.tsx
```diff
@@ -435,37 +451,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
-      const isSchemaChanged = !deepEquals(prevProps.schema, this.props.schema);
-      // When formData is not an object, getChangedFields returns an empty array.
-      // In this case, deepEquals is most needed to check again.
-      const isFormDataChanged =
-        formDataChangedFields.length > 0 || !deepEquals(prevProps.formData, this.props.formData);
-      const isStateDataChanged =
-        stateDataChangedFields.length > 0 || !deepEquals(this.state.formData, this.props.formData);
-      const nextState = this.getStateFromProps(
+    // A state-only update hands over the same props object
+    if (this.props === prevProps) {
```
jimmycallin, line 442: A state-only update reaches `getSnapshotBeforeUpdate` with the same props object, so this line replaces a full `deepEquals(this.props, prevProps)` walk with one identity check for the common case: every keystroke.

packages/core/src/components/Form.tsx
```diff
@@ -435,37 +451,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
+      return { shouldUpdate: false };
+    }
+    // `replaceEqualDeep()` hands back `prev` exactly when the values are deep-equal with functions by identity
+    let isIdentityPropChanged = false;
+    let isSchemaChanged = false;
+    for (const key of IDENTITY_PROP_KEYS) {
```
jimmycallin, line 448: `replaceEqualDeep(prev, next) !== prev` is used here as an equality test, not for its result: it returns `prev` exactly when the two values are deep-equal *with functions compared by identity*. That is the comparison `deepEquals` cannot do. `isSchemaChanged` is tracked separately because it decides whether the retrieved schema in state can be kept.

packages/core/src/components/Form.tsx
```diff
@@ -435,37 +451,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
+      if (replaceEqualDeep(prevProps[key], this.props[key]) !== prevProps[key]) {
+        isIdentityPropChanged = true;
+        isSchemaChanged ||= key === 'schema';
+      }
+    }
+    // Any other prop change still re-derives state, which is what snaps a controlled form back to its `formData` prop
+    if (!isIdentityPropChanged && deepEquals(this.props, prevProps)) {
```
jimmycallin, line 455: The `deepEquals` fallback is kept on purpose. Any prop change, including one this PR does not care about, still re-derives state, and that re-derivation is what snaps a controlled form back to its `formData` prop today. Removing it here would change ownership semantics, which is the RFC's job. It is cheap in practice: the props interned above hit the identity fast path inside `deepEquals`.

packages/core/src/components/Form.tsx
```diff
@@ -435,37 +451,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
+      return { shouldUpdate: false };
+    }
+    // Shared against the state, so a prop echoing the state's own formData is that formData
+    const formData = replaceEqualDeep(this.state.formData, this.props.formData);
```
jimmycallin, line 459: **The controlled-parent case.** A parent doing `setFormData(event.formData)` hands back a value whose unchanged subtrees are already the state's own instances, because `setSharedState` produced them. Sharing the prop against `state.formData` therefore returns `state.formData` itself, and `isStateDataChanged` is a single `!==`. Before this, `getChangedFields` plus `deepEquals` walked the whole tree to reach the same answer.

packages/core/src/components/Form.tsx
```diff
@@ -435,37 +451,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
+    const isStateDataChanged = formData !== this.state.formData;
+    // An accepting parent hands the proposal back as a prop; without sharing, the rebuilt state would re-render every
+    // field
+    // Spread over `prevState` so the keys it alone carries, such as `prevExtraErrors`, do not stop `replaceEqualDeep()`
+    // from returning `prevState` itself when nothing changed
+    const nextState = replaceEqualDeep(prevState, {
```
jimmycallin, line 465: Two things worth knowing here. The spread over `prevState` is needed because `prevState` carries keys that `getStateFromProps` never sets, such as `prevExtraErrors`; without it the outer object could never come back as `prevState` and `shouldUpdate` would always be true. And `shouldUpdate` is now `nextState !== prevState`, which is exact: `replaceEqualDeep` returns `prevState` if and only if the derived state is deep-equal with functions by identity.

packages/core/src/components/Form.tsx
```diff
@@ -435,37 +451,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
-        // Skip live validation for this request if no form data has changed from the last state
-        !isStateDataChanged,
-      );
-      const shouldUpdate = !deepEquals(nextState, prevState);
-      return { nextState, shouldUpdate };
-    }
-    return { shouldUpdate: false };
+        () => getChangedFields(formData, prevProps.formData, true),
+        // Live validation is skipped only when neither the data nor anything that could take part in it changed
+        !isStateDataChanged && !isIdentityPropChanged,
```
jimmycallin: Behavior fix, not just a refactor: live validation used to be skipped whenever the *data* was unchanged, so replacing `validator`, `customValidate` or `transformErrors` on a controlled form did not re-validate. It now runs when any identity prop changed as well.

packages/core/src/components/Form.tsx
```diff
@@ -1147,17 +1145,16 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
-      const hasChanges = Object.keys(state)
-        // Filter out `schemaValidationErrors` and `schemaValidationErrorSchema` since they aren't IChangeEvent props
-        .filter((key) => !key.startsWith('schemaValidation'))
-        .some((key) => {
-          const oldData = this.state[key as keyof FormState<T, S, F>];
-          const newData = state[key as keyof FormState<T, S, F>];
-          return !deepEquals(oldData, newData);
-        });
-      this.setState(state as FormState<T, S, F>, () => {
+      const committed = this.state;
```
jimmycallin, line 1186: `hasChanges` is computed in the `setState` callback by comparing the committed state against a snapshot taken before, key by key, by identity. This is equivalent to the old `!deepEquals(old, new)` filter because `setSharedState` returns the previous instance for every key whose value is deep-equal, and it avoids walking the values a second time.

packages/utils/src/replaceEqualDeep.ts
```diff
@@ -0,0 +1,125 @@
+import isPlainObject from './isPlainObject.ts';
+
+function isReactElement(thing: unknown): thing is { $$typeof: unknown; type: unknown; key: unknown; props: unknown } {
```
jimmycallin, line 3: `$$typeof` alone is not enough to identify an element: `forwardRef()`, `memo()`, `lazy()`, contexts and portals carry one too and have no `type`/`props` pair. Matching on `$$typeof` only would make every `memo` widget equal to every other, which is exactly the swallowed-replacement bug this PR fixes. The `forwardRef` scenario in `renderStability.test.tsx` pins it.

packages/utils/src/replaceEqualDeep.ts
```diff
@@ -0,0 +1,125 @@
...
+  // `$$typeof` alone is not enough: `forwardRef()`, `memo()`, `lazy()`, contexts and portals carry one too and have
+  // no `type`/`props` pair, so comparing those three would report every one of them as equal to every other
+  return isPlainObject(thing) && '$$typeof' in thing && 'type' in thing && 'props' in thing;
+}
+
+/** Whether `prev` and `next` carry the same own enumerable symbol keys with identical values. RJSF marks schemas with
+ * symbols (`ADDITIONAL_PROPERTY_FLAG`, `RJSF_REF_KEY`, `RJSF_REF_CYCLE_KEY`), which `Object.keys()` cannot see, so
+ * without this two schemas differing only in a marker would be treated as the same one.
+ */
+function sameOwnSymbols(prev: object, next: object): boolean {
```
jimmycallin, line 13: `Object.keys()` cannot see symbol keys, and RJSF marks schemas with symbols (`ADDITIONAL_PROPERTY_FLAG`, `RJSF_REF_KEY`, `RJSF_REF_CYCLE_KEY`). Without this check two schemas that differ only in a marker would be treated as the same one and the marker would be lost when the previous instance is returned.

packages/utils/src/replaceEqualDeep.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/** The container pairs the walk is currently inside, as a flat `prev, next, prev, next, ...` stack. A `formContext`
+ * may hold a graph that refers back to itself, and `deepEquals()` survives one, so this walk has to as well. A cycle can
+ * only recur through an ancestor on the current path, so the path is all that needs checking, and one array per call
+ * keeps the cycle safety from costing an allocation per container in the plain data that never has one. Pairs are
+ * popped on the way out, so a value reachable by more than one path is still shared as long as it does not contain
+ * itself.
+ */
+type WalkedPairs = object[];
```
jimmycallin, line 30: Cycle safety as a path stack rather than a visited set. A cycle can only recur through an ancestor on the current path, so only the pairs currently being descended need checking, and popping on the way out keeps a value reachable by more than one path shareable. Flat `prev, next, prev, next` layout means one array per top-level call and no allocation per container, which matters because this runs on every keystroke over data that almost never contains a cycle.

packages/utils/src/replaceEqualDeep.ts
```diff
@@ -0,0 +1,125 @@
...
+    walked.push(prev, next);
+    const nextKeys = Object.keys(next);
+    let sameAsPrev = true;
+    let copy: Record<string, unknown> | undefined;
+    for (const key of nextKeys) {
+      const value = shareUnchanged(prev[key], next[key], walked);
+      sameAsPrev &&= Object.hasOwn(prev, key) && Object.is(value, prev[key]);
+      if (!Object.is(value, next[key])) {
+        // Spread keeps a JSON-sourced own `__proto__` key an own key, so the assignment below never reaches the setter
+        copy ??= { ...next };
```
jimmycallin, line 91: Spread rather than `Object.assign` or index assignment onto a fresh `{}`: a JSON-sourced own `__proto__` key stays an own key under spread, so the `copy[key] = value` below never reaches the prototype setter. Line 103 then restores a null prototype only when the source had one, since reassigning a prototype deoptimizes property access on the object.

packages/utils/src/createSchemaUtils.ts
```diff
@@ -314,14 +322,32 @@ class SchemaUtils<
...
-    return retrieveSchema<T, S, F>(
-      this.validator,
-      schema,
-      this.rootSchema,
-      rawFormData,
-      this.customMergeAllOf,
-      resolveAnyOfOrOneOfRefs,
+    // `Form` renders an error for a non-object schema instead of throwing, so one reaches here despite the type, and a
+    // `WeakMap` refuses it as a key
+    const cacheKey = schema !== null && typeof schema === 'object' ? schema : undefined;
```
jimmycallin, line 345: `WeakMap` refuses a primitive key, and `Form` deliberately renders an error for a non-object `schema` instead of throwing, so a non-object can reach here despite the type. Those calls simply bypass the cache.

packages/utils/src/createSchemaUtils.ts
```diff
@@ -314,14 +322,32 @@ class SchemaUtils<
...
+    const cached = cacheKey && this.lastRetrievedSchemas.get(cacheKey);
+    if (
+      cached &&
+      Object.is(cached.rawFormData, rawFormData) &&
+      cached.resolveAnyOfOrOneOfRefs === resolveAnyOfOrOneOfRefs
+    ) {
+      return cached.result;
+    }
+    const result = replaceEqualDeep(
```
jimmycallin, line 354: Two levels of reuse. The early return above handles an identical call. This line handles a *changed* call, sharing the fresh resolution against the previous result so that fields whose resolved subschema did not change keep their `schema` reference even when, say, a `dependencies` branch elsewhere in the schema flipped.

packages/core/src/components/fields/SchemaField.tsx
```diff
@@ -351,22 +355,11 @@ function SchemaFieldRender<T = any, S extends StrictRJSFSchema = RJSFSchema, F e
...
-  if (experimental_componentUpdateStrategy === 'always') {
-    return false; // always re-render — never consider props equal
-  }
-  if (experimental_componentUpdateStrategy === 'shallow') {
-    return shallowEquals(prevProps, nextProps);
-  }
-  // default: 'customDeep'
-  return deepEquals(prevProps, nextProps);
-}) as unknown as typeof SchemaFieldRender;
+const SchemaField = memo(SchemaFieldRender) as typeof SchemaFieldRender;
```
jimmycallin, line 398: `memo()` erases the generic parameters, so the cast keeps `<T, S, F>` for consumers. It is a single cast now because the comparator, which needed `unknown` in between, is gone.

jimmycallin (commented):
Self-review: inline notes on the hunks that are hardest to read cold, as promised in section 6 of the description.

packages/core/test/Form.behaviors.test.tsx
```diff
@@ -1080,8 +1080,12 @@ describe('Calling onChange right after updating a Form with props formData', ()
 
     await user.click(node.querySelector('.rjsf-array-item-add button')!);
 
-    expect(node.querySelector('#root_0')).toBeInTheDocument();
-    expect(node.querySelector('#root_1')).toHaveAttribute('value', 'test');
+    // Both near-simultaneous changes must survive; which lands first depends on render timing, so assert order-free
```
jimmycallin, line 1182: The old assertion encoded `customDeep` render timing: which of two near-simultaneous changes (an add-item click racing a typed value) landed at index 0 depended on when the deep compare let the second render through. Under shallow comparison the order differs, but the contract being tested is that neither change is lost, so the assertion is now order-independent and still fails if one is dropped.

packages/core/src/components/Form.tsx
```diff
@@ -422,37 +438,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
-        // Skip live validation for this request if no form data has changed from the last state
-        !isStateDataChanged,
-      );
-      const shouldUpdate = !deepEquals(nextState, prevState);
-      return { nextState, shouldUpdate };
-    }
-    return { shouldUpdate: false };
+        () => getChangedFields(formData, prevProps.formData, true),
+        // Live validation is skipped only when neither the data nor anything that could take part in it changed
+        !isStateDataChanged && !isIdentityPropChanged,
```
heath-freenome: **`liveValidate: 'onBlur'` now validates on every keystroke in a common controlled setup.**

`getStateFromProps` sets `mustValidate` whenever `liveValidate` is truthy, and `'onBlur'` counts as truthy. Before this change, the skip flag was `!isStateDataChanged`. That was `true` when the `formData` prop just echoed the form's own state, so validation was skipped until blur.

Adding `&& !isIdentityPropChanged` breaks that for a controlled form with an inline callback:

```tsx
<Form
  liveValidate='onBlur'
  formData={fd}
  onChange={(e) => setFd(e.formData)}
  transformErrors={(errs) => errs /* or an inline customValidate */}
/>
```

Each keystroke re-renders the parent and creates a new arrow, so `isIdentityPropChanged` is true. Live validation then runs, and errors show up before the field is blurred. In `'onChange'` mode, the same setup validates twice per keystroke.

packages/core/src/components/Form.tsx
```diff
@@ -422,37 +438,47 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
+    // Shared against the state, so a prop echoing the state's own formData is that formData
+    const formData = replaceEqualDeep(this.state.formData, this.props.formData);
+    const isStateDataChanged = formData !== this.state.formData;
+    // An accepting parent hands the proposal back as a prop; without sharing, the rebuilt state would re-render every
+    // field
+    // Spread over `prevState` so the keys it alone carries, such as `prevExtraErrors`, do not stop `replaceEqualDeep()`
+    // from returning `prevState` itself when nothing changed
+    const nextState = replaceEqualDeep(prevState, {
+      ...prevState,
+      ...this.getStateFromProps(
```
heath-freenome, line 467: **Spreading `prevState` can restore stale `customErrors` / `prevExtraErrors`.**

`getStateFromProps` never returns `customErrors` or `prevExtraErrors`, so `{ ...prevState, ...getStateFromProps(...) }` fills them in from `prevState`, the state from before this update. `componentDidUpdate` then calls `setState(nextState)`, which writes those old values back over anything newer. The old code applied only the derived keys, so it didn't have this problem.

Scenarios:
1. The parent changes `extraErrors` and another prop in the same render. `getDerivedStateFromProps` stores the new `prevExtraErrors`, and `setState(nextState)` restores the old one. That forces an extra `getDerivedStateFromProps` pass and render.
2. `reset()` (which sets `customErrors: undefined`) is batched with a parent prop change. The custom errors raised from `onChange` come back after the reset.

A possible fix is to spread only the derived keys and still let `replaceEqualDeep` keep `prevState`'s identity, e.g. compare against a `prevState` that has these keys picked out, rather than writing them back.

packages/utils/src/createSchemaUtils.ts
```diff
@@ -325,14 +333,32 @@ class SchemaUtils<
...
+    // `Form` renders an error for a non-object schema instead of throwing, so one reaches here despite the type, and a
+    // `WeakMap` refuses it as a key
+    const cacheKey = schema !== null && typeof schema === 'object' ? schema : undefined;
+    const cached = cacheKey && this.lastRetrievedSchemas.get(cacheKey);
+    if (
+      cached &&
+      Object.is(cached.rawFormData, rawFormData) &&
+      cached.resolveAnyOfOrOneOfRefs === resolveAnyOfOrOneOfRefs
+    ) {
+      return cached.result;
```
heath-freenome, line 352: **The identity-keyed cache returns a stale schema when `formData` is changed in place, and `Form.processPendingChange` does exactly that.**

In `processPendingChange`, `getStateFromProps` calls `retrieveSchema(rootSchema, formData)`, and the result is cached against that `formData` object. Right after, `setByPath(formData, path, undefined)` clears the leaf in the same object (the #5125 re-clear after defaults are merged, around `Form.tsx:1001`). For a flat object of primitives, `replaceEqualDeep` returns that changed object as the new `state.formData`. The root `SchemaField`'s `retrieveSchema(schema, formData)` then hits this cache and gets the schema resolved against the data from *before* the clear.

Repro shape:

```json
{
  "type": "object",
  "properties": { "a": { "type": "string", "default": "x" } },
  "if": { "required": ["a"] },
  "then": { "properties": { "b": { "type": "string" } } }
}
```

Clearing `a` keeps showing `b`. Before this PR, the schema was recomputed from the cleared data.

Either `processPendingChange` should stop changing `formData` in place after `retrieveSchema` runs, or the cache key needs more than object identity.

heath-freenome (commented):
Three issues found reading the code at c92882a (not run locally). Two change what users see: `liveValidate: 'onBlur'` validating on every keystroke, and a stale `retrieveSchema` cache hit after `processPendingChange` clears a field in place. The third is stale `customErrors`/`prevExtraErrors` coming back through the `prevState` spread. Details are inline.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed, and fixed in `4e4fe7541`. A changed identity prop now forces a re-validation only in `onChange` mode:

```ts
!isStateDataChanged && !(isIdentityPropChanged && this.props.liveValidate === 'onChange'),
```

`onBlur` owes its errors to the blur, so a parent handing over a fresh callback is not a reason to produce them early. The original intent of the widening — a swapped `validator`, `customValidate` or `transformErrors` re-validating a controlled form — is unaffected in `onChange` mode, which is the mode that revalidates on data changes anyway.

Regression test in `Form.behaviors.test.tsx`, under the existing `Live validation onBlur` describe: a controlled parent that recreates an inline `transformErrors` on every render. Typing produces no `.error-detail` entries; blurring produces the `minLength` one. It fails on the previous line (`expected 0, got 1` after the first keystroke) and passes with the fix.

The double validation you noted for `onChange` mode stays: an inline callback changes identity every render, which is exactly the case the `IDENTITY_PROP_KEYS` comment tells callers to wrap in `useCallback`. That one is a cost, not a behavior change, so I left it where the documentation already points.

packages/core/src/components/Form.tsx
jimmycallin: You're right that the spread writes back keys the derived state doesn't own, and that's fixed in `4e4fe7541`: `componentDidUpdate` now commits through an updater that keeps whatever React currently holds for the two keys `getStateFromProps()` never sets.

```ts
this.setState((current) => ({
  ...nextState,
  customErrors: current.customErrors,
  prevExtraErrors: current.prevExtraErrors,
}));
```

Being straight about the evidence, though: I could not turn either scenario into a failing test, and I tried both.

- **`customErrors`.** `processPendingChange` calls `props.onChange` from inside the `setSharedState` callback, so the custom errors are already committed before the parent re-renders. By the time the prop-driven update reaches `getSnapshotBeforeUpdate`, `prevState` carries them. I built a controlled parent with a field that raises a `newErrorSchema` per keystroke and typed into two fields in turn; the first field's custom error survives with and without the fix.
- **`prevExtraErrors`.** Restoring the old reference does make the next `getDerivedStateFromProps` pass re-merge, but I measured the render count with a counting `FieldTemplate` on a parent that changes `extraErrors` and `formContext` together: 1 → 3 renders both with and without the fix. The re-merge produces the same values, so React bails out and it costs nothing observable.

So no regression test with this one — a test that passes either way is noise. The guard is in anyway, because the correctness of the spread currently rests on that `setState`-callback ordering, and that isn't something this code should be depending on.

packages/utils/src/createSchemaUtils.ts
jimmycallin: The in-place clear is gone in `4e4fe7541` — it now lands on a copy, so the object `getStateFromProps()` resolved the schema against is never mutated afterwards:

```ts
formData = setByPath((Array.isArray(formData) ? [...formData] : { ...formData }) as T, path, undefined);
```

But I have to report that your repro does not fail on the current head. I wrote it as a test (`ifthenelse.test.tsx`, `re-resolves the conditional branch when a field with a default is cleared`, using exactly your schema) and it passes with the fix and without it.

Instrumenting the cache-hit branch shows why. Clearing `a` produces these two hits at the root schema:

```
CACHE HIT {"a":"x"} -> ["a","b"]
CACHE HIT {}        -> ["a"]
```

Between them, `retrieveSchema` is called with a *different* `formData` object — `getDefaultFormState()` builds a fresh tree on the follow-up pass — which misses the cache, re-resolves against the cleared data and overwrites the entry. The render's own call then hits the refreshed entry, so the `else` branch is what `SchemaField` sees and `b` is gone. Same result through a controlled parent.

I've kept both the fix and the test: the test pins the user-visible behavior your example is about, and the copy makes the invariant the cache needs true by construction rather than by that overwrite happening to land first. It also stops the clear mutating an object `replaceEqualDeep()` may hand straight back as `state.formData`.

If you have a shape where you saw the stale branch actually render, I'll take it and turn it into a failing test.

jimmycallin:
Two follow-ups, pushed as `06fc82126` and `2f2d8e78e`, from taking your `liveValidate: 'onBlur'` finding as a class rather than a single line.

### `liveValidate` was read as a boolean in two more places

The fix in `4e4fe7541` stopped a changed identity prop forcing a re-validation under `'onBlur'`, but the mode confusion it exposed sits deeper, and both remaining sites are on `v7` today rather than introduced here:

- `getStateFromProps()`: `mustValidate = edit && !noValidate && liveValidate`. Any truthy value counted, so deriving state under `'onBlur'` ran a full validation pass. Replacing the form data from outside produced the errors immediately, with no field ever having been left. `'onBlur'` owns its pass in `onBlur()`; only `'onChange'` belongs here.
- `getCurrentErrors()`: `if (!props.liveValidate)`, the same conflation from the other side. The base for that branch has to be the validator's own result, because `extraErrors` and `customErrors` are merged in right after. `state.errors` already carries them, and is only safe to reuse on the `'onChange'` pass that asked to be skipped. With `mustValidate` corrected, `'onBlur'` started reaching this branch and merged every `extraErrors` entry in twice — I caught it on the way in rather than shipping it.

Two regression tests in `Form.behaviors.test.tsx`, under `Live validation onBlur` → `when the data is replaced from outside the form`: the errors still wait for a blur, and each `extraError` is carried over exactly once. Both fail on the mode-blind version (`expected 0, got 1` and a duplicated `.name` entry). `@rjsf/core` is 2141 passing; `typecheck`, uncached `lint`, `knip`, `cs-check` and `pnpm test` across all packages are green. Changelog entry added, since this one is a behavior fix against `v7` rather than a regression internal to this PR.

### The mutation audit the cache invited

Since the stale-cache thread is really about an identity-keyed cache being only as correct as its callers' discipline, I swept every in-place mutation that can reach data this PR now shares by reference — `setByPath`/`unsetByPath` in `Form.tsx`, `ObjectField`, `LayoutMultiSchemaField`, `ArrayField`.

Result: the site fixed in `4e4fe7541` was the only one that broke it. Everything else operates on a fresh `structuredClone`, a spread copy or a reduce accumulator, and `LayoutMultiSchemaField`'s `setByPath` is unreachable in the one case where `sanitizeDataForNewSchema()` could hand back the caller's own object. The `ErrorSchemaBuilder` mutations are a mutable-by-design class instance, which `replaceEqualDeep()` treats as opaque, so they are outside the invariant.

What was missing was the invariant being written down anywhere, so it is now on `lastRetrievedSchemas`: both keys are matched by identity, which is correct only while callers treat the data they pass as frozen from that point on, and structural sharing makes breaking that easy by accident, because `replaceEqualDeep()` returns the value it was given when nothing could be shared back.

jimmycallin:
One more from the same family, found while checking whether the `liveValidate` mode confusion also exists on `main`. Pushed as `7fb039357`.

**`getStateFromProps()` dropped every `extraErrors` entry from the live validation it runs.** It passes `undefined` where `liveValidate()` takes the `extraErrors` to merge:

```ts
const liveValidation = this.liveValidate(
  rootSchema,
  schemaUtils,
  state.errorSchema,
  formData,
  undefined,          // <- props.extraErrors
  state.customErrors,
```

So the pass that runs when props change produced only the validator's own errors, and a controlled parent replacing the form data made every server-supplied error disappear:

```
initial=["from the server"]
after=["must NOT have fewer than 8 characters"]
```

`processPendingChange` passes `extraErrors` at that position, so this looks like an oversight rather than a decision. With `props.extraErrors` passed, both errors survive the replacement. Regression test alongside the other two, in `Form.behaviors.test.tsx`; it fails on the old line. Changelog entry added.

Worth noting that this one was still live here after the earlier fixes: `'onBlur'` only escaped it because it now takes the other branch. `'onChange'` went straight through it.

### On `main`

For the record, since it affects what gets backported: all three are reproducible on `main` at `3002e4791`, unfixed. `mustValidate` (`main:585`) and `getCurrentErrors()` (`main:677`) are the same two lines, and the dropped `extraErrors` hits both modes there. If you want them on the v6 line I'll open a separate PR against `main` — the predicate there also has to accept the deprecated boolean, i.e. `liveValidate === true || liveValidate === 'onChange'`, the form `processPendingChange` (`main:1067`) already uses inline.

CI green here at `7fb039357`.

packages/core/src/components/Form.tsx
```diff
@@ -422,37 +438,49 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
-      );
-      const shouldUpdate = !deepEquals(nextState, prevState);
-      return { nextState, shouldUpdate };
-    }
-    return { shouldUpdate: false };
+        () => getChangedFields(formData, prevProps.formData, true),
+        // Live validation is skipped only when neither the data nor anything that could take part in it changed.
+        // A changed identity prop counts only in `onChange` mode: `onBlur` owes its errors to the blur, not to the
+        // parent handing over a fresh callback
+        !isStateDataChanged && !(isIdentityPropChanged && this.props.liveValidate === 'onChange'),
```
heath-freenome: **`extraErrors` that the parent later drops never clear from the field.**

Widening live validation to any identity-prop change runs it with `mergeIntoOriginalErrorSchema=true` against `state.errorSchema`, which already has `extraErrors` and `customErrors` merged in. That merged result is then stored as `schemaValidationErrorSchema`.

Repro on `7fb039357`: `liveValidate='onChange'`, controlled `formData={{a:'x'}}`, `extraErrors={{a:{__errors:['EXTRA']}}}`.
1. The parent re-renders with a changed `uiSchema` (identity prop, data unchanged), so live validation runs against the already-merged errors.
2. The parent then drops `extraErrors`.
3. `state.errors` becomes `[]`, but `errorSchema` keeps `a.__errors=['EXTRA']` and the DOM still renders one `.text-danger`.

Base renders none. A server-side error sticks until the data changes. Restricting the gate to `'onChange'` in 4e4fe7541 doesn't cover this, because this is the `'onChange'` path.

heath-freenome: **In `'onChange'` mode, any parent re-render carrying a fresh function-bearing prop now validates the whole untouched form.** This is the part of my earlier "double validation" point that is still open.

Measured on the PR head: a controlled form with an inline `transformErrors` and `liveValidate='onChange'` makes 6 validator calls for 3 keystrokes, against 3 on base. The change's own validation runs, then the parent's re-render with a new `transformErrors` identity runs it again.

There's also a visible effect beyond cost. A form mounted with no errors starts showing e.g. `minLength` errors on fields the user never touched once the parent re-renders with a new `customValidate`, inline `transformErrors`, `widgets` or `formContext`. Base ignored function-identity changes.

Documenting "props holding functions must be stable references" doesn't remove this. I'd rather force re-validation only when a prop that actually feeds validation changed (`validator`, `customValidate`, `transformErrors`, `schema`, `extraErrors`), and not for `widgets`/`formContext`/`uiSchema`.

packages/utils/src/replaceEqualDeep.ts
```diff
@@ -0,0 +1,125 @@
...
+  if (isPlainObject(prev) && isPlainObject(next)) {
+    if (isWalking(walked, prev, next)) {
+      return next;
+    }
+    walked.push(prev, next);
+    const nextKeys = Object.keys(next);
+    let sameAsPrev = true;
+    let copy: Record<string, unknown> | undefined;
+    for (const key of nextKeys) {
+      const value = shareUnchanged(prev[key], next[key], walked);
```
heath-freenome: **`prev[key]` is read even when `prev` has no own `key`, so an inherited member leaks in as the previous value.**

Repro: `replaceEqualDeep({a: 2}, JSON.parse('{"a":1,"__proto__":{}}'))` returns an object whose own `__proto__` value is `=== Object.prototype`. `isPlainObject(Object.prototype)` is true because its own prototype is `null`, so the empty `{}` gets "shared" back to the global. `constructor: {}` is compared against `Object.prototype.constructor` in the same way.

`Form` runs this over `formData` on every state update. JSON-sourced data, or `additionalProperties` keys named `__proto__`, can therefore end up holding a live reference to `Object.prototype`. A later in-place write (`setByPath` in `LayoutMultiSchemaField`, or a consumer mutating `event.formData`) would then pollute the global prototype.

The `sameAsPrev` line already uses `Object.hasOwn(prev, key)`, so the fix is to read `prev[key]` only when `Object.hasOwn(prev, key)`, and pass `undefined` otherwise. A test with a `__proto__` key would pin it.

packages/core/src/components/Form.tsx
```diff
@@ -422,37 +438,49 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
+      ...this.getStateFromProps(
         this.props,
-        this.props.formData,
-        // If the `schema` has changed, we need to update the retrieved schema.
-        // Or if the `formData` changes, for example in the case of a schema with dependencies that need to
-        //  match one of the subSchemas, the retrieved schema must be updated.
-        isSchemaChanged || isFormDataChanged ? undefined : this.state.retrievedSchema,
+        formData,
+        // The retrieved schema in state matches the state's data, so it only holds while both are unchanged
+        isSchemaChanged || isStateDataChanged ? undefined : this.state.retrievedSchema,
```
heath-freenome, line 484: **`isSchemaChanged` tracks only `schema`, so a changed `validator`, `customMergeAllOf` or `defaultFormStateBehavior` reuses the old `state.retrievedSchema`.**

Repro with `customMergeAllOf`: a schema with `allOf: [{properties:{a:{title:'A1'}}}, {properties:{a:{title:'A2'}}}]` and unchanged `formData`. The parent swaps `customMergeAllOf` from first-wins to last-wins. `getStateFromProps` builds a new `schemaUtils`, but the `retrievedSchema` passed in here is reused. `state.retrievedSchema.properties.a.title` stays `'A1'` while the rendered label is `'A2'`. The stale `retrievedSchema` then goes to live validation and to sanitization.

The PR description and changelog say a changed validator or `customMergeAllOf` "takes effect at once", and that doesn't hold for the state-derived part. The simplest fix is to ignore the passed `retrievedSchema` inside `getStateFromProps` whenever `createSchemaUtils` just ran.

packages/core/src/components/Form.tsx
```diff
@@ -611,7 +645,10 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
       if (props.noValidate || isSchemaChanged) {
         return { errors: [], errorSchema: {} };
       }
-      if (!props.liveValidate) {
+      // `extraErrors` and `customErrors` are merged in below, so the base has to be the validator's own result.
+      // `state.errors` already carries them, and is only safe to reuse when live validation has just produced it,
+      // which is the `'onChange'` pass that asked to be skipped
+      if (props.liveValidate !== 'onChange') {
```
heath-freenome: **For `liveValidate='onChange'`, `getCurrentErrors` returns `state.errors` (which already carries `extraErrors`/`customErrors`), and the caller then merges `props.extraErrors`/`customErrors` in again, so errors duplicate whenever live validation is skipped.** That contradicts the comment above, which says the reuse is safe.

Repro: `liveValidate='onChange'`, `extraErrors={{name:{__errors:['from the server']}}}`, `formData={{name:'short'}}`.
- Re-rendering with only `className='x'` changes `state.errors` from `['from the server']` to `['from the server','from the server']`.
- Changing `extraErrors` to `'second'` gives `['second','second']`.

This also happens on base `f73249536`, so it predates the PR. But this function is the one 06fc82126 rewrote to fix the same conflation. The `onBlur` path was fixed by routing it to `schemaValidationErrors`, and `'onChange'` still takes the merged list. Using `schemaValidationErrors` here as well would let the caller's merge stay the only merge.

packages/utils/src/createSchemaUtils.ts
```diff
@@ -325,14 +340,32 @@ class SchemaUtils<
...
+    }
```
heath-freenome: **The early return keyed on `rawFormData` identity returns a stale schema after any in-place mutation of an already-resolved object.** The `processPendingChange` fix copies only the root object.

Repro at the cache level: `retrieveSchema(condSchema, d); d.k = 'a'; retrieveSchema(condSchema, d)` returns the identical stale result, and the `then` properties never appear. In `Form.tsx`, `{...formData}` followed by `setByPath(copy, path, undefined)` still mutates nested objects shared with the object `getStateFromProps` resolved against. It only works if no nested `retrieveSchema` entry exists yet. The constructor and `getSnapshotBeforeUpdate` can also make `state.formData` the caller's own props object, so a parent that mutates its `formData` in place is now silently ignored. Base had no cache.

The identity shortcut is only an optimisation. `replaceEqualDeep(cached?.result, fresh)` on the next lines already gives stable result identities. Dropping the `Object.is(cached.rawFormData, rawFormData)` return removes the invalidation hazard for the cost base already paid.

packages/utils/src/createSchemaUtils.ts
```diff
@@ -325,14 +340,32 @@ class SchemaUtils<
...
-      this.validator,
-      schema,
-      this.rootSchema,
-      rawFormData,
-      this.customMergeAllOf,
-      resolveAnyOfOrOneOfRefs,
+    // `Form` renders an error for a non-object schema instead of throwing, so one reaches here despite the type, and a
+    // `WeakMap` refuses it as a key
+    const cacheKey = schema !== null && typeof schema === 'object' ? schema : undefined;
+    const cached = cacheKey && this.lastRetrievedSchemas.get(cacheKey);
```
heath-freenome: **The cache holds one slot per schema object, keyed on `(rawFormData, resolveAnyOfOrOneOfRefs)`, so callers that share a schema object evict each other.** The fallback then shares against another caller's result.

Repro on a `oneOf`-of-`$ref`s schema: `retrieveSchema(s, fd)`, `(s, fd, true)`, `(s, fd)`, `(s, fd, true)` returns the flagged result with a new identity on every call. `SchemaField` and `LayoutGridField` (flag `undefined`) collide with `ObjectField` (flag `true`). Every `ArrayField` item also shares `schema.items` while passing different `itemData`.

Each render misses, recomputes, and runs `replaceEqualDeep` against a neighbour's result. That produces fresh subschema identities, which defeat the memo boundaries this PR exists to provide, and adds a full-schema walk per field per render. Results stay deep-equal, so this is about performance and stability. Keying the slot on the flag as well, or on `rawFormData`, would stop the eviction.

packages/core/test/Form.behaviors.test.tsx
```diff
@@ -1080,8 +1179,12 @@ describe('Calling onChange right after updating a Form with props formData', ()
...
+    const values = [
```
heath-freenome: **The "doesn't cause a race condition" assertion was loosened, but the PR head still satisfies the original positional one.**

I restored the original assertions (`#root_0` present, `#root_1` value `'test'`) on `7fb039357` and they passed 6 out of 6 runs. The order-free `[...].sort()` `toEqual(['', 'test'])` also accepts `['test', '']`, so it no longer catches the new item landing at the wrong index. That is the failure this test exists for.

Please revert to the original assertions. The PR description says the timing changed, and that isn't borne out here.

packages/core/src/components/Form.tsx
```diff
@@ -486,11 +514,18 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
...
+      // `getStateFromProps()` sets neither `customErrors` nor `prevExtraErrors`, so `nextState` carries the values
+      // `prevState` held before this update; `getDerivedStateFromProps()` or a batched `setState()` may already have
+      // replaced them, and writing the old ones back would undo that
       // oxlint-disable-next-line react/no-did-update-set-state -- guarded to prevent infinite loop
-      this.setState(nextState);
+      this.setState((current) => ({
+        ...nextState,
+        customErrors: current.customErrors,
+        prevExtraErrors: current.prevExtraErrors,
+      }));
```
heath-freenome: **Non-blocking.** The updater still commits the whole snapshot over `current` and protects only `customErrors` and `prevExtraErrors`, and neither guard has a failing test.

`getSnapshotBeforeUpdate` derives `nextState` from `this.state` before commit. If a batched `setSharedState` (from `processPendingChange`/`onBlur`) or a `getDerivedStateFromProps` result changes `errors`, `errorSchema`, `schemaValidation*` or `formData` in between, the snapshot values overwrite them. I couldn't build a concrete interleaving, and `setState(nextState)` overwrote the same way before the PR, so this isn't a regression.

The guard is harmless, but it's a per-key whitelist. Deriving only the keys `getStateFromProps` owns, or asserting the intended behaviour with a test for each of the two scenarios, would make it less fragile.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -66,7 +66,7 @@ The compiled `dist/<theme>.css` files (`default`, `amethyst-haze`, `caffeine`, `
 `@rjsf/shadcn` is now styled the way shadcn components are, by the application's Tailwind build. Add the package to your Tailwind sources:
 
 ```css
-@import 'tailwindcss';
+@import "tailwindcss";
```
heath-freenome: Unrelated formatter churn: `@import 'tailwindcss';` became `@import "tailwindcss";`. It departs from the repo's single-quote style and adds a merge-conflict surface in a heavily edited v7 doc. The same goes for `_current_` becoming `*current*` at line 361. Please revert both hunks.

Separately, the new "props holding functions must be stable references" contract is documented only in this guide. It should also sit next to `customValidate` and `transformErrors` in `form-props.md`, where people look up those props.

heath-freenome (commented):
Review of the latest round (head `7fb039357`).

**Last round:**
- `liveValidate: 'onBlur'` no longer validates on every keystroke. Fixed.
- The `prevState` spread restoring stale `customErrors`/`prevExtraErrors` is guarded, though neither scenario has a failing test.
- The `retrieveSchema` cache clear is only partly addressed. See the `createSchemaUtils.ts` comment.
- The extraErrors fix in `7fb039357` works.

**Still open, worth fixing before merge:**
- `extraErrors` never clear once the parent drops them, under `liveValidate='onChange'` (`Form.tsx:479`).
- Whole-form validation on any parent re-render with a fresh function prop (`Form.tsx:479`).
- The `Object.prototype` leak in `replaceEqualDeep`.
- Stale `retrievedSchema` when `customMergeAllOf`/`validator` changes.
- The identity-keyed `retrieveSchema` cache.

The rest are smaller: duplicate errors in `getCurrentErrors`, cache-slot eviction, the loosened race-condition test, and docs churn. The `componentDidUpdate` comment is non-blocking.

Core suite on the PR head passes (2142 tests).

packages/core/src/components/Form.tsx
jimmycallin: Fixed in `49ce6a4c0`. The merge into `state.errorSchema` is gone: `liveValidate()` no longer takes `originalErrorSchema`/`mergeIntoOriginalErrorSchema` at all, since `getStateFromProps()` was the only caller passing `true` and the other two never used the parameter. A re-validation now stores the validator's own result and merges `extraErrors`/`customErrors` on top once.

Regression test `clears extraErrors the parent drops after a re-validation of unchanged data` in `Form.behaviors.test.tsx`, under the new `Deriving state from changed props` describe. It follows your repro as three steps (swap a validation prop, then drop `extraErrors`) because both in one render never showed the bug: `getDerivedStateFromProps()` had already rebuilt `state.errorSchema` from the clean base before the snapshot read it. Fails on `7fb039357` with `['must NOT have fewer than 8 characters', 'from the server']`.

jimmycallin: Done in `49ce6a4c0`, as you suggested: a `VALIDATION_PROP_KEYS` subset (`schema`, `validator`, `customMergeAllOf`, `customValidate`, `transformErrors`) is what forces the re-validation now. `fields`, `templates`, `widgets`, `formContext`, `uiSchema`, `translateString` and `nameGenerator` still re-derive state when they change, but never validate unchanged data.

One deliberate omission to flag: `uiSchema` and `formContext` do feed `ui:required`, so a flipped `ui:required` on untouched data shows on the next data change, blur or submit rather than at once. That is what v6 did for every one of these props, and it seemed the right trade against validating the whole form on any parent render carrying an inline `ui:widget`.

Tests: `does not validate untouched data when a prop that takes no part in validation changes` (new `formContext`, no errors appear) and `re-validates unchanged data when a prop that takes part in validation changes` (new `transformErrors`, the `minLength` error appears). The first fails on `7fb039357`.

The inline-`transformErrors` double validation you measured remains by construction, since `transformErrors` is in the set; `form-props.md` now says so next to `customValidate` and `transformErrors`.

packages/utils/src/replaceEqualDeep.ts
jimmycallin: Fixed in `8ce059bfd`: `prev[key]` is read only when `Object.hasOwn(prev, key)`, `undefined` otherwise. Test `never reads an inherited member of prev as a previous value` uses your `JSON.parse('{"a":1,"__proto__":{}}')` shape plus a `constructor: {}` key and checks that neither `Object.prototype` nor its constructor ends up in the result; it fails on `7fb039357`.

I looked for the same pattern elsewhere in `@rjsf/utils`. `mergeObjects()` reads `obj1[key]` for every key of `obj2` the same way, but its recursion copies with a spread and never returns the inherited object itself, so there is no shared reference to pollute. Left alone; it predates this PR.

packages/core/src/components/Form.tsx
jimmycallin: Fixed in `49ce6a4c0`. `getStateFromProps()` now drops the passed `retrievedSchema` whenever it rebuilds `schemaUtils`, so the schema is re-resolved by the new validator/`customMergeAllOf`/`defaultFormStateBehavior` and that result is what live validation and sanitization see.

Test `re-resolves the retrieved schema when customMergeAllOf changes` with your `allOf` shape: after the swap the label reads `A2` and `state.retrievedSchema.properties.a.title` is `A2`. On `7fb039357` the label changes and the state keeps `A1`, as you described.

packages/core/src/components/Form.tsx
jimmycallin: Fixed in `49ce6a4c0`: `getCurrentErrors()` returns `schemaValidationErrors`/`schemaValidationErrorSchema` in both modes, so the caller's merge is the only merge. `newState.errors` from `processPendingChange()`'s skip pass is never read (it commits its own `liveValidation` or leaves errors out), so nothing depended on the merged list.

Test `carries each extraError over exactly once when a non-validation prop changes`: a `className`-only re-render under `'onChange'` keeps exactly one `from the server`. Fails on `7fb039357` with two. Changelog entry added since, as you say, the duplication predates the PR.

packages/utils/src/createSchemaUtils.ts
jimmycallin: Dropped in `28fa21fed`. The `Object.is(cached.rawFormData, rawFormData)` return is gone; every call resolves and shares against the last result, so an in-place mutation of an already-resolved object is picked up. The invariant comment on the cache went with it, since the cache is now only a `replaceEqualDeep()` base and nothing is served from it.

Test `re-resolves against form data that was mutated in place` is your repro at the cache level (`retrieveSchema(s, d); d.k = 'a'; retrieveSchema(s, d)` gains the dependent property); it fails on `7fb039357`. The copy in `processPendingChange` stays, now for the reason you gave in the other thread: the object may be the committed state's own.

packages/utils/src/createSchemaUtils.ts
jimmycallin: Done in `28fa21fed`: one `WeakMap` per flag value, so `SchemaField`/`LayoutGridField` and `ObjectField` no longer evict each other. Test `keeps a stable result per resolveAnyOfOrOneOfRefs value when callers alternate` alternates the two calls on a `oneOf`-of-`$ref`s schema and checks both identities hold; on `7fb039357` the flagged result is a new object every time.

Not keyed on `rawFormData`: `ArrayField` items still share the `schema.items` slot. When the per-item resolutions are deep-equal, sharing against a sibling's result returns that same instance anyway, so the slot only churns for an `items` schema whose conditional actually resolves differently per item. That felt like the wrong place to hold a per-data-object map.

packages/core/test/Form.behaviors.test.tsx
jimmycallin: Reverted to the original assertions in `49ce6a4c0`; they pass here too. The description's claim about the timing was wrong and I've stopped making it.

packages/core/src/components/Form.tsx
jimmycallin: Taken in `49ce6a4c0`: `getSnapshotBeforeUpdate` now shares only what `getStateFromProps()` returns against `prevState`, decides `shouldUpdate` by comparing those keys by identity, and `componentDidUpdate` commits that same partial through `setSharedState`. `customErrors` and `prevExtraErrors` are never in the object that gets committed, so the per-key list is gone rather than extended. Still no failing test for either scenario, for the reasons in the earlier thread; the change is that the code no longer has a place where the guard could be forgotten.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Both hunks restored in `e19e51213`, by running `oxfmt` on the file. For the record, since I went looking: the branch's copy of the guide was written before #5289 fixed those two lines on `v7`, and the rebase conflict in this file was resolved with the branch's copy, which carried the old spelling back. Nothing flagged it because `cs-check` only covers `src/**/*.{ts,tsx}` in the docs package.

The stable-reference note now sits under `customValidate` and `transformErrors` in `form-props.md`, and the upgrade-guide bullet names the props whose change re-validates under `'onChange'`.

jimmycallin:
Round 3 pushed, head `89d6a8724`. Every open thread has an inline reply naming the commit and the regression test; each new test fails on `7fb039357`.

- `8ce059bfd` `replaceEqualDeep` reads only own members of `prev`.
- `28fa21fed` `retrieveSchema` serves nothing by identity; one sharing slot per flag value.
- `49ce6a4c0` re-validation only for `schema`/`validator`/`customMergeAllOf`/`customValidate`/`transformErrors`; `liveValidate()` no longer merges into the previous `errorSchema`; `getCurrentErrors()` returns the validator's result in both modes; a rebuilt `schemaUtils` re-resolves the retrieved schema; `componentDidUpdate` commits only derived keys; race test restored.
- `e19e51213` docs and changelog; the guide's formatting churn was a rebase-conflict artifact and is undone.
- `89d6a8724` the same `__proto__`/`constructor` hole in `mergeObjects()`, found while checking whether the `replaceEqualDeep` pattern existed elsewhere. There it also swapped the result's prototype through the `__proto__` setter. Predates this PR, but it's the same class of bug, so it's here.

The description's `retrieveSchema` paragraphs are updated to the new semantics. The v6 backport of the three live-validation fixes plus the `getCurrentErrors` duplication is #5306.

`typecheck`, uncached `lint`, `knip`, `cs-check`: green. `@rjsf/core` 2147 passing, `@rjsf/utils` 1676 at 100% coverage.

packages/utils/src/mergeObjects.ts
```diff
@@ -31,10 +35,10 @@ export default function mergeObjects(
             return result;
           }, []);
         }
-        acc[key] = left.concat(toMerge);
-      } else {
-        acc[key] = right;
+        merged = left.concat(toMerge);
       }
+      // A plain assignment to `__proto__` would reach the setter and swap the result's prototype for `merged`
+      Object.defineProperty(acc, key, { value: merged, enumerable: true, writable: true, configurable: true });
```
heath-freenome: **`Object.defineProperty` on every key makes `mergeObjects` about 2.6x slower, and only `__proto__` needs it.**

`mergeObjects` sits under `retrieveSchema`, the defaults pass and error merging, so it runs on every keystroke. Here every merged key goes through `defineProperty` where it used to be a plain assignment.

Measured against the previous `mergeObjects.ts` (`7fb039357`), with 300 merges of two 20-key, 3-level trees on Node 24:

```
old  50.8 ms   47.6 ms
new 129.8 ms  128.9 ms
```

That is a regression on a PR whose purpose is to cut per-keystroke work.

Only the `__proto__` key reaches the setter, so the slow path can be limited to it:

```ts
if (key === '__proto__') {
  Object.defineProperty(acc, key, { value: merged, enumerable: true, writable: true, configurable: true });
} else {
  acc[key] = merged;
}
```

The `Object.hasOwn(obj1, key)` read above is cheap and can stay. `constructor` is safe under plain assignment because the data property just shadows the inherited one, and the existing test already pins both keys.

Separately, this is unrelated to the shallow-comparison change. It is a public `@rjsf/utils` behaviour fix with its own changelog entry, so it could be its own PR. I'll leave that to you.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -476,6 +476,19 @@ The modifier is shallow: it stops `event.formData = ...`, not a write inside `fo
...
+
+`Form` and `SchemaField` now compare their props shallowly, as `React.memo` does, and keep the references of unchanged data across changes so that comparison is enough. The `experimental_componentUpdateStrategy` prop is removed, and with it `shouldRender()`, `shallowEquals()` and the `ComponentUpdateStrategy` type from `@rjsf/utils`.
+
+What to change:
+
+- Remove the `experimental_componentUpdateStrategy` prop from `<Form>` and from any `globalFormOptions` read of it.
+- Give `Form` stable references for props that hold functions: `customValidate`, `transformErrors`, `translateString`, `nameGenerator`, `customMergeAllOf`, `validator`, `fields`, `widgets`, `templates`, `formContext`, and a `uiSchema` carrying components such as `ui:widget`. A value recreated on every render re-derives the form's state on every render, so wrap it in `useCallback` or `useMemo`, or define it outside the render function. A changed one now takes effect at once, where v6 ignored it until another prop changed; with `liveValidate: 'onChange'`, a changed `validator`, `customValidate`, `transformErrors`, `customMergeAllOf` or `schema` also re-validates the current data.
```
heath-freenome: **Doc gap: a changed `uiSchema` or `formContext` does not re-validate, though `validate()` reads both.**

`Form.validate()` passes `uiSchema` to `validateFormData()`. `getUiRequiredErrorSchema()` also reads `uiSchema` and `this.props.formContext`, for `ui:required`. Neither prop is in `VALIDATION_PROP_KEYS` (`Form.tsx:340`), so with `liveValidate: 'onChange'` a change to either does not re-validate unchanged data. That was a deliberate choice in `49ce6a4c0`, and it matches the base branch, so I'm not asking for a behaviour change.

I checked the visible effect with a controlled `liveValidate='onChange'` form and a parent button that flips `uiSchema.other['ui:required']` from `false` to `true` on an empty field. No required error appears after the flip, and it only shows up on the next data change, blur or submit.

The sentence "A changed one now takes effect at once" reads as covering every prop in that list, including `uiSchema` and `formContext`. For validation it doesn't, so a `ui:required` driven by parent state can look stale. Something like this would close the gap:

> `uiSchema` and `formContext` take effect in rendering at once, but a change to either re-validates only on the next data change, blur or submit. This matters if `ui:required` depends on state outside the form.

`form-props.md` could carry the same sentence under `uiSchema`, where people look up `ui:required`.

heath-freenome:
Also, will either of my 2 main PRs, once merged and rebased in v7 cause issues with this PR?

heath-freenome (approved):
@jimmycallin 2 comments remaining, one a doc fix, the other something you could fix in a separate PR. Approving

packages/utils/src/mergeObjects.ts
jimmycallin: Good catch, fixed in ee1ffac9f — `defineProperty` now only guards `__proto__`, everything else is a plain assignment again. Reproduced the gap locally (300 merges of two 20-key/3-level trees, Node 24): ~4.4s defining every key vs ~1.2s narrow. `constructor` stays safe because `Object.prototype.constructor` is a writable data property, so the assignment shadows it as own data; the existing test pins both keys and still passes.

On splitting it out: leaving it here. It is six lines plus its changelog entry, and the whole prototype-chain pass (`replaceEqualDeep` + `mergeObjects`) came out of the same audit as this PR — separating half of it would make both harder to read.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Confirmed — `VALIDATION_PROP_KEYS` holds neither, so the sentence overpromised. Fixed in `e0c661610`, close to your wording:

> `uiSchema` and `formContext` take effect in rendering at once, but a change to either re-validates only on the next data change, blur or submit, which matters if `ui:required` depends on state outside the form.

Put the second copy in `uiSchema.md` under `ui:required` rather than `form-props.md`, next to the existing line about the submit and `liveValidate` paths — that is the page that documents the directive, so it is where the caveat is in view when someone reads how enforcement works.

jimmycallin:
> will either of my 2 main PRs, once merged and rebased in v7 cause issues with this PR?

Checked all three of your open ones against this branch. Short answer: the two v7 feature PRs are fine, #5305 needs one line of care.

**#5303 (numeric `inputMode`) and #5301 (optgroups)** — no behavioural interaction. The overlap is textual only:

- `packages/utils/src/types.ts` — you add `inputMode`/`pattern` to `InputPropsType` and the `optgroups` option plus the grouped-option types; I remove `ComponentUpdateStrategy` and the `experimental_componentUpdateStrategy` prop. Different regions.
- `packages/utils/src/index.ts` (#5301) — you add three exports, I remove `shouldRender`/`shallowEquals`. Different regions.
- `packages/core/test/Form.behaviors.test.tsx` (#5303) — you touch ~1008 and ~1110, I touch ~1083. Close enough to conflict on rebase, trivial to resolve.
- `CHANGELOG-v7.md` and the v7 upgrade guide — the usual.

`groupEnumOptions()` runs inside each theme's `SelectWidget`, so the fresh array it builds never becomes a prop of a memoized component. No effect under shallow comparison.

**#5305 (typeless const `oneOf`/`anyOf`)** — one real interaction, though it targets `main`. `inferSelectType()` is called as

```ts
const schema = inferSelectType<S>(schemaUtils.retrieveSchema(_schema, formData));
```

and returns `{ ...schema, type }` for exactly the schemas it targets. This PR makes `retrieveSchema` hand back a stable instance across renders and `SchemaField` a plain `memo`, so wrapping the result in a fresh object per render means those fields re-render on every parent render and the stability never reaches the widget. The other branches return `schema` untouched, so it only bites typeless const selects. Cheapest fix once both are in v7: do the inference behind the same cache, so the inferred schema is shared like the retrieved one. Nothing needs changing in #5305 before then.

PR #5298, opened 2026-09-18
refactor(core)!: make IChangeEvent a standalone readonly interface

jimmycallin:
Lane C1 of the [form data ownership RFC](https://github.com/jimmycallin/react-jsonschema-form/pull/62) ([spec, section 7](https://github.com/jimmycallin/react-jsonschema-form/blob/rfc/form-state-ownership/rfcs/form-state-ownership-spec.md#7-implementation-and-review-sequence)). It depends on nothing that is open: not #5297, not #5296. Base: `v7`.

## 1. The problem, from first principles

A public type can be declared in two ways. It can be written out, so that its shape is whatever the declaration says. Or it can be derived from another type, so that its shape is whatever that other type happens to be.

`IChangeEvent`, the event `onChange` and `onSubmit` receive, was derived: a `Pick<FormState, ...>` of seven keys from the `Form` class's internal state. That reads as a convenience, and for a stable class it is one. It has a consequence that only shows up when the state needs to change: **every reshaping of `FormState` is silently a change to the public event.** Rename a state member, narrow its type, split it in two, and the event follows without anyone having written a line in the event's file, without a changelog entry, and without a type error anywhere in the repository, because the only code that builds an event is `toIChangeEvent()` and it reads from the same state.

The ownership RFC needs to reshape `FormState`. Its lane A extracts render-context derivation and the change pipeline out of `getStateFromProps`, and its ownership switch removes controlled `state.formData` altogether. With a derived event type, each of those steps would be a public API change in disguise. So the event has to become independent of the state before the state can move. That is the whole of this PR.

## 2. What changes

- **`IChangeEvent` is declared on its own** in `packages/core/src/components/IChangeEvent.ts`, with the same seven members and the same types they had through the `Pick`: `schema`, `uiSchema`, `schemaUtils`, `formData`, `edit`, `errors`, `errorSchema`, plus `status?: 'submitted'`. It extends nothing.
- **`toIChangeEvent()` stays in `Form.tsx`** as the only place an event is built. It is the single point where state becomes event, so `FormState` can change shape as long as that one function still compiles.
- **Every member is `readonly`.** See section 3.
- **A type test pins the equivalence.** `IChangeEvent.test.ts` asserts each member is mutually assignable with the corresponding `FormState` member today, that the event no longer extends the state, and that assignment into the event is a compile error. From now on the two types can only drift if someone edits the test, which turns an accident into a decision with a migration note.
- **`Form.tsx` changes in two hunks**: the deleted declaration and one added `import type`. Nothing lane A rewrites is touched, so this PR imposes no rebase on any `Form.tsx` work.

## 3. Why `readonly`, and why it is shallow

#5296 makes every `Registry` member and the `schema`, `uiSchema` and `registry` props `readonly`, and its argument applies unchanged here. Readonly modifiers cost nothing at runtime, never reach into caller-owned objects, and do not affect assignability, so the only code that stops compiling is code that *writes into* the value. That is the code we want to hear from.

For the event there is a second reason. #5297 makes the `formData` in an event share its unchanged subtrees with the value the change was applied to, which for a controlled form is the parent's own object. Writing into the event therefore writes into the parent's state. #5297 documents that as an immutability contract and the RFC adds a development-only freeze to enforce it at runtime. `readonly` on the event members is the compile-time half of the same contract: `event.formData = x` is now an error rather than a silent write into shared data.

It is shallow by nature. `readonly formData?: T` stops reassigning the member, not a write inside `formData`. A deep readonly type would have to be applied to `T`, the consumer's own data type, which would change the type of `formData` every consumer sees. That is the same boundary #5296 draws at runtime when it declines to freeze caller-owned objects inside the registry. The development-only freeze in the RFC covers the inside of `formData`; this covers the surface.

## 4. Breaking change

**`IChangeEvent` members are read-only.** Members and types are unchanged, so a handler that reads the event, destructures it, or passes it on compiles exactly as before. The only code affected assigns into the event:

```tsx
// Before: compiled, and wrote into data the form shares with its parent
onChange={(event) => { event.formData = normalize(event.formData); setData(event.formData); }}

// After
onChange={(event) => setData(normalize(event.formData))}
```

*Why it is justified:* that pattern was already a bug under #5297, because it mutated the parent's data. Making it a compile error is the cheapest possible way to find every instance. The migration guide and the v7 changelog carry the entry.

## 5. What this does not do

No `edit` redefinition, no new members, no runtime change, no `Form` behavior change. All 2055 core tests pass unchanged; the only new test is the type test. Root typecheck, lint, knip and format check pass.

## 6. Review guide

Inline comments mark the three places worth a second look: the `readonly` members and their TSDoc, the type test's two-directional `toExtend` and why it is not `toEqualTypeOf`, and the two-hunk footprint in `Form.tsx`.


packages/core/src/components/IChangeEvent.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/** The event handed to `onChange` and `onSubmit`. It is declared on its own rather than as a `Pick` of `FormState`
+ * because it is the public contract while `FormState` is an implementation detail: the state's layout is free to change
+ * as long as `toIChangeEvent()` in `Form.tsx`, the only place an event is built, still produces this shape.
+ *
+ * Every member is `readonly`. The event's `formData` shares its unchanged subtrees with the value the change was
+ * applied to, which for a parent-owned form is the parent's own object, so writing into the event writes into that
+ * data. Copy what you need out of the event instead.
+ */
+export interface IChangeEvent<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any> {
```
jimmycallin, line 19: **Why this extends nothing.** The old declaration was `extends Pick<FormState, ...>`, which made the event's shape a function of the state's shape. With the members written out, the state is free to change as long as `toIChangeEvent()` in `Form.tsx` still compiles. The member TSDoc is copied from `FormState` verbatim, so the docs site and IDE hovers read the same as before.

jimmycallin (commented):
Self-review: notes on the hunks that are easiest to misread, as promised in section 6 of the description.

packages/core/src/components/IChangeEvent.ts
```diff
@@ -0,0 +1,36 @@
...
+  /** The JSON schema object for the form */
+  readonly schema: S;
```
jimmycallin, line 21: **Why `readonly`, and why only this deep.** The modifier stops `event.schema = ...`, not a write inside the value. It costs nothing at runtime and does not affect assignability, so a `FormState` still satisfies this interface and `toIChangeEvent()` needs no change. Deep readonly was rejected for the same reasons #5296 rejected it for the registry: it does not survive generics and index signatures without contorting every consumer type. The runtime freeze in the ownership RFC covers the inside of `formData`.

packages/core/src/components/Form.tsx
```diff
@@ -55,6 +55,7 @@ import {
 import { buildRegistry } from '../Theme.ts';
 import { ADDITIONAL_PROPERTY_KEY_REMOVE, IS_RESET } from './constants.ts';
 import type { FormHandle } from './FormHandle.ts';
+import type { IChangeEvent } from './IChangeEvent.ts';
```
jimmycallin, line 58: The whole `Form.tsx` footprint of this PR is this import and the deleted declaration below. Nothing in `getStateFromProps`, the lifecycle methods or the `toIChangeEvent` call sites changes, which is what lets this land alongside any `Form.tsx` work in flight without a rebase.

packages/core/test/IChangeEvent.test.ts
```diff
@@ -0,0 +1,33 @@
+import { expectTypeOf } from 'vitest';
+
+import type { FormState, IChangeEvent } from '../src/index.ts';
+
+type EventKey = Exclude<keyof IChangeEvent, 'status'>;
+type Mutable<O> = { -readonly [K in keyof O]: O[K] };
```
jimmycallin, line 6: `Mutable` strips the new `readonly` so the two shapes can be compared on members and types alone. Without it the second `toExtend` below would fail for the wrong reason: a mutable `FormState` member is assignable to a readonly one, but the reverse direction would be reporting readonly-ness, not a type difference.

packages/core/test/IChangeEvent.test.ts
```diff
@@ -0,0 +1,33 @@
...
+/** The event is declared on its own, so nothing ties it to `FormState` any more. These assertions are what does: the
+ * day the two drift apart on a shared member, this file stops compiling, and the drift becomes a deliberate decision
+ * with a migration note instead of an accident of reshaping the state.
+ */
+describe('IChangeEvent', () => {
+  it('carries the same members as FormState, with the same types', () => {
+    expectTypeOf<EventKey>().toEqualTypeOf<
+      'schema' | 'uiSchema' | 'schemaUtils' | 'formData' | 'edit' | 'errors' | 'errorSchema'
+    >();
+    expectTypeOf<Mutable<Pick<IChangeEvent, EventKey>>>().toExtend<Pick<FormState, EventKey>>();
```
jimmycallin, line 17: **Why two `toExtend` calls instead of one `toEqualTypeOf`.** `toEqualTypeOf` compares by a deep structural brand, and `RJSFSchema` and `ErrorSchema` are recursive types; the brand collapses them to `never[]` and the comparison fails even though the types are identical. Mutual assignability, one `toExtend` each way, is exactly what the RFC asks for and is robust to recursion. If a member's type ever diverges in either direction, one of these two lines stops compiling.

packages/core/test/IChangeEvent.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    expectTypeOf<Pick<FormState, EventKey>>().toExtend<Mutable<Pick<IChangeEvent, EventKey>>>();
+    expectTypeOf<IChangeEvent['status']>().toEqualTypeOf<'submitted' | undefined>();
+  });
+
+  it('is no longer a view of FormState', () => {
+    expectTypeOf<IChangeEvent>().not.toExtend<FormState>();
```
jimmycallin: This is the assertion that the decoupling actually happened. The event lacks the state's private members, so it cannot extend `FormState`; had the `Pick` been left in place this would still hold, which is why the equivalence checks above are the load-bearing ones and this is the guard against someone reintroducing `extends`.

packages/core/test/IChangeEvent.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('cannot be written into', () => {
+    const event = {} as IChangeEvent;
+    // @ts-expect-error every member is readonly
```
jimmycallin, line 24: A `@ts-expect-error` is a stronger assertion than `not.toHaveProperty`: it fails to compile if the line ever becomes legal, which is exactly what happens if a future edit drops `readonly` from a member.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -426,6 +426,22 @@ The id helpers `titleId()`, `descriptionId()`, `helpId()`, `errorId()`, `example
 
 The `useDeepCompareMemo()` hook is removed. It existed to hand out a stable `fieldPathId` object across renders; with two primitive strings there is nothing to stabilize. A custom field or widget that used it for its own props can use `useMemo()`, or compare the values it cares about directly.
 
+### `IChangeEvent` members are read-only BREAKING CHANGE
```
jimmycallin, line 457: Placed as its own sub-heading rather than an edit to a shared paragraph, per the RFC's file-ownership rule for the migration guide: every open PR appends here, and a sub-heading per PR keeps conflicts to an append-order fix.

heath-freenome:
Reviewed this refactor — it's a clean, correct decoupling of `IChangeEvent` from `FormState`. Verified `tsc --noEmit` passes on `packages/core`, and grepped the repo for any mutation into `IChangeEvent`/event params that the new `readonly` constraint would break (found none besides the `@ts-expect-error`-guarded assertions in the new test file).

One minor test-quality nit:

In `packages/core/test/IChangeEvent.test.ts:22`, the assertion `expectTypeOf<IChangeEvent>().not.toExtend<FormState>()` is meant to guard against `IChangeEvent` becoming re-coupled to `FormState` again, but it's vacuously true — even the *old* `Pick<FormState, ...>`-based type already fails `toExtend<FormState>()`, since `FormState` has additional required private members (`schemaValidationErrors`, `retrievedSchema`, `registry`, etc.) that the `Pick`-derived type never had. So if a future PR reverts this refactor back to a `Pick<FormState, ...>`, this specific assertion would still pass, giving false confidence that the decoupling is still in place. The real protection against shape drift comes from the other two `toExtend` assertions on `EventKey` earlier in the same test.

Not a blocker — just worth tightening if you want the test to actually catch that regression.

heath-freenome (approved):
One small test improvement suggested. Approved assuming that the suggestion is accessed and accepted or rejected.

jimmycallin:
Agreed, and thanks for checking why it is vacuous rather than just flagging it — the `Pick<FormState, ...>` form never extended `FormState` either, so the assertion guarded nothing.

Removed in `0f6c08388` rather than tightened. There is nothing meaningful left for it to assert: a revert to `Pick<FormState, ...>` is caught by the `cannot be written into` test (a `Pick` is mutable, so the two `@ts-expect-error` lines would stop erroring), and a revert to `Readonly<Pick<FormState, ...>>` is structurally identical to the standalone interface, so there is no observable difference to test for. Shape drift stays covered by the `EventKey` assertions, as you noted.

PR #5299, opened 2026-09-18
refactor(core): keep only row keys in ArrayField state; read items from props (ownership RFC, lane C2)

jimmycallin:
Lane C2 of the [form data ownership RFC](https://github.com/jimmycallin/react-jsonschema-form/pull/62) ([spec, section 7 lane C2 and section 5](https://github.com/jimmycallin/react-jsonschema-form/blob/rfc/form-state-ownership/rfcs/form-state-ownership-spec.md)). Behavior-preserving preparation in the child fields so that the ownership switch later touches less. Base: `v7`. Does not depend on #5296. Never edits `Form.tsx`.

**#5297 has merged**, and this branch is rebased onto `v7` on top of it. The structural sharing it added is what keeps this PR from being a per-keystroke regression for every form with an array field, as discussed in the thread on `ArrayField.tsx`, and its render-stability gate, `renderStability.test.tsx`, is now in `v7`; see section 5.

## 1. What an optimistic local copy is, and why the fields have them

A field component receives the current value of its slice of the form data through props. When the user acts, the field computes a new value and calls `onChange` with it. That call is a *proposal*: the field is asking the form to make this the new value.

Today the class `Form` always accepts the proposal. It writes the value into its own state in the same React batch as the field's handler, so by the time anything re-renders, the field's props already carry the value it proposed. A field that also keeps a *local copy* of the value it just proposed is therefore never contradicted: props and copy always agree. That is why such copies have accumulated without anyone noticing them. They looked like a convenience, and under an always-accepting form they were harmless.

Under the ownership RFC they stop being harmless. Section 3 of the spec gives the form data exactly one owner. For a controlled form that owner is the parent, and a parent may reject or transform a proposal, exactly as a controlled `<input>` parent may. When that happens the *props* carry the authoritative value and the *local copy* carries a value nobody accepted. A field that renders from the copy shows a value that does not exist. Section 5 of the spec names this directly: rejected controlled add/remove/reorder proposals must not leave optimistic item values visible.

The fix is not clever: a field must render from props, always. The only thing a field may keep locally is UI metadata that has no counterpart in the data, such as which oneOf branch is selected or which React key belongs to which array row.

## 2. What changes: `ArrayField`

`ArrayField` kept the most complete copy of the three. Its `useKeyedFormData` hook held, in `useState`, the array items themselves paired with their React keys, plus a hash of the last `formData` prop it had seen. On every render it re-hashed the incoming `formData` and, if the hash differed from the stored one, decided that "something outside changed the array" and rebuilt the pairs from props. Its handlers wrote the proposed items straight into that state before calling `onChange`.

Read that flow once more from the RFC's point of view. The items in state were the proposal. The hash comparison was the mechanism that let a rejected proposal *eventually* be replaced by the real value on the next render, but only after one render had already shown the rejected items. And the hash walk over the whole array ran on every render of every array field.

After this PR the hook holds only the keys:

- **Items come from props on every render.** The rows always show the value the form holds. There is nothing to reconcile, so the hash and its per-render walk are gone, and with them the `hashObject` import.
- **Keys stay in state as UI metadata.** React needs a stable key per row so that an accepted add, remove or reorder reuses the existing row instance and keeps its DOM focus. Keys have no counterpart in the data, so they are exactly the kind of thing section 5 allows a field to keep.
- **A length mismatch regenerates every key.** When the array's length changes outside the handlers, an external replacement of the value or a proposal the form transformed, for example padding to `minItems`, there is no way to know which rows survived. The old code regenerated on a length mismatch inside its hash-mismatch branch; the new code regenerates on the length mismatch alone, which is the same outcome by a shorter route.
- **The handlers are untouched.** `updateKeyedFormData` keeps its signature. It now records the new keys and returns the plain items for `onChange`, instead of also storing the items.

### Why this preserves behavior today, traced

The question a reviewer should ask is: is there a render in which the old code showed something different from the new code?

- **Accepted proposal, the only case the current `Form` produces.** A handler calls `updateKeyedFormData` then `onChange`. With an idle pending queue, `Form.onChange` runs `setState` inside the same event handler, so React batches the field's `setKeys` with the form's state update into one render. In that render the keys are the proposed ones and the props are the proposed items. Old code: proposed items from state. New code: proposed items from props. Identical.
- **External replacement with the same length.** Old code: hash differs, length matches, keys mapped onto the new items by index. New code: length matches, keys kept, items from props. Identical.
- **External replacement with a different length.** Both regenerate every key. Identical.
- **Form transforms a proposal.** Same length: both keep keys and show the transformed items (old via hash mismatch, new via props). Different length: both regenerate. Identical.
- **Two handler calls before a re-render**, for example a double-click on add. Both compute from a ref that still holds the pre-render value, so the second call's proposal wins in both. Same limitation, unchanged.
- **A proposal that queues behind another.** `Form.onChange` processes a change synchronously only when its pending queue is empty; a second change arriving in the same flush, for instance from two effects, is processed in the first one's `setState` callback, a later commit. In that interim render the field's keys already reflect the add or remove but the props do not. Old code: the stored hash differs from the props hash and the lengths differ, so every key is regenerated and the optimistic items are shown for one frame. New code: the lengths differ, so every key is regenerated and the props items are shown for one frame. The regeneration condition is the same in both, because a length mismatch always implies a hash mismatch; the only difference is which items fill that one frame, and the next commit corrects both. Rows remount in that case today as well; this PR does not make that better or worse.
- **`formData` that is not an array.** Both treat it as empty.

The old `keyedFormDataRef`, which lets the four mutation handlers stay reference-stable without `keyedFormData` in their dependency arrays, is kept. It is a stability device, not a copy of the data: it is overwritten from the current render on every render and never written by a handler.

## 3. What was audited and deliberately left alone

**`ObjectField.handleKeyRename` and its `formDataRef`** (introduced in #5031). The ref has two halves. The *read* half lets the memoized rename callback see the latest `formData` without listing it as a dependency, which keeps every property's `onKeyRename` prop stable so the property memo boundaries hold. That half stays regardless. The *write* half, `formDataRef.current = renamedObj`, is a genuine optimistic copy: it exists so that a second rename in the same batch composes onto the first rename's result rather than onto stale props. Removing it today breaks `ObjectField.test.tsx`'s "two keys renamed in quick succession", because the second rename would read the pre-rename props and drop the first. The closure *is* stale in that case; the ref is doing real work. The correct replacement is for the form's change pipeline to compose queued operations against its committed value, which is what section 4 of the spec specifies for `applyChange` and the serial queue, and that is lane A work. Deferred to A4 with the rest of the rename audit section 5 asks for.

**`MultiSchemaField`.** Its `selectedOption` state is a branch choice, which section 5 explicitly allows to remain local because data can match several branches. `prevFormDataRef` and `prevFieldIdRef` mirror `componentDidUpdate` and hold no data the field renders from. `skipNextOptionRecalculation` is a timing flag that guards against the branch match being recomputed over a user's explicit choice after defaults are populated; section 5 says to remove timing flags only when tests establish they are unnecessary, and no test does today. Nothing here is an optimistic copy of the value. No change.

**Path propagation.** Every handler in the three files already passes its `fieldPath` explicitly to `onChange`; `ArrayField.handleChange` forwards the child's path untouched. There was nothing indirect to make explicit.

## 4. What this buys the RFC

When A4 flips ownership, a rejected array proposal is simply not in props, so the rows never show it, without A4 touching `ArrayField`. Section 5's array requirement is met structurally rather than by a fix in the switch PR. The `ObjectField` rename remains on A4's list, with the analysis above as its starting point.

## 5. Verification

On this branch, all from the repository root: `typecheck`, `lint` uncached, `knip`, `cs-check`, and `pnpm test` across all 15 packages. All pass. `@rjsf/core`: 2150 passing. Zero snapshot updates in any theme suite. One pre-existing oxlint warning in `ArrayField.tsx` about a template literal, present on `v7` before this change.

`renderStability.test.tsx` passes on this branch (221 tests with `ArrayField.test.tsx`), so the sibling rows the gate watches do not re-render. In-repo coverage of the rewrite, runnable against `v7` here, in `ArrayField.test.tsx` under `External formData replacement`: an external replacement of the same length keeps every row key and renders the new items from props; a shorter replacement regenerates every key; a longer one regenerates every key too. Disabling the length-mismatch regeneration fails the last two.

Render-stability gate: these commits cherry-picked cleanly onto the current head of #5297 (`4e4fe7541`) in a scratch worktree. `renderStability.test.tsx` and `ArrayField.test.tsx` there: 221 tests, all passing, every sibling render count still zero. The whole `@rjsf/core` suite on that combination: 2142 passing.


packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -816,49 +801,30 @@ interface KeyedFormDataState<T = any> {
...
 }
 
-/** Type used for the state of the `ArrayField` component */
-interface ArrayFieldState<T> {
-  /** The hash of the last formData passed in */
-  formDataHash: string;
-  /** The keyed form data elements */
-  keyedFormData: KeyedFormDataType<T>[];
-}
+const NO_ITEMS: never[] = [];
```
jimmycallin, line 791: A module-level empty array rather than a fresh `[]` per render: `items` feeds the `useMemo` below, and a new empty array on every render of an empty or undefined array field would rebuild `keyedFormData` each time. With a shared constant, an empty field's memo is stable.

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -816,49 +801,30 @@ interface KeyedFormDataState<T = any> {
...
+ * replacement or a proposal the form transformed, there is no way to tell which rows survived, so every key is
+ * regenerated, as the previous hash-based detection also did.
  */
-function useKeyedFormData<T = any>(formData: T[] = []): KeyedFormDataState<T> {
-  const newHash = useMemo(() => hashObject(formData), [formData]);
-  const [state, setState] = useState<ArrayFieldState<T>>(() => ({
-    formDataHash: newHash,
-    keyedFormData: generateKeyedFormData<T>(formData),
-  }));
+function useKeyedFormData<T = any>(formData: T[] = NO_ITEMS): KeyedFormDataState<T> {
```
jimmycallin, line 800: **What is no longer here.** The previous hook kept the array *items* in state next to the keys and a hash of the last `formData` prop, re-hashing on every render to detect outside changes. The items in state were an optimistic copy of the field's own proposal. Under the current `Form` that copy was always equal to the next props, so it was invisible; under the ownership RFC a rejected proposal must not render, so the copy has to go. Only the keys remain, and only because React needs them to keep row instances and focus across accepted operations.

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -816,49 +801,30 @@ interface KeyedFormDataState<T = any> {
...
-      nextFormData.length === previousKeyedFormData.length
-        ? previousKeyedFormData.map((previousKeyedFormDatum, index) => ({
-            key: previousKeyedFormDatum.key,
-            item: nextFormData[index],
-          }))
-        : generateKeyedFormData<T>(nextFormData);
-    formDataHash = newHash;
-    setState({ formDataHash, keyedFormData });
+  let itemKeys = keys;
+  if (itemKeys.length !== items.length) {
```
jimmycallin, line 806: **Why length alone is enough.** The old code compared hashes and then, inside the mismatch branch, regenerated keys only when the length differed; on a same-length change it mapped the old keys onto the new items by index. Since items are now read from props every render, a same-length change needs nothing at all, and the only situation that needs handling is a length change from outside the handlers: an external replacement, or the form padding or trimming a proposal. In that case there is no way to know which rows survived, so every key is regenerated, exactly as before. Calling `setKeys` during render is the same pattern the previous hook used; React re-renders the component once with the new state before committing.

jimmycallin (commented):
Self-review: notes on the hook rewrite, so the difference between the removed data copy and the retained key state is clear at each line.

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -816,49 +801,30 @@ interface KeyedFormDataState<T = any> {
...
+    itemKeys = items.map(generateRowId);
+    setKeys(itemKeys);
   }
 
+  const keyedFormData = useMemo(() => itemKeys.map((key, index) => ({ key, item: items[index] })), [itemKeys, items]);
```
jimmycallin, line 811: The pairs are rebuilt when either input changes. `ArrayFieldItem` receives `itemKey` and `itemData` as separate props, so the identity of a pair object never reaches a memo boundary; what matters for render stability is that `items[index]` keeps its reference for unchanged rows, which the `formData` prop guarantees once #5297 lands. The render-stability gate against #5297's head confirms sibling rows do not re-render.

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -816,49 +801,30 @@ interface KeyedFormDataState<T = any> {
...
+
   const updateKeyedFormData = useCallback((newData: KeyedFormDataType<T>[]) => {
```
jimmycallin, line 813: Same signature as before, so none of the four mutation handlers changed. It now stores only the keys of the proposed rows and hands the plain items back for `onChange`. The proposal itself lives nowhere in this component after this call: it reaches the rows only when the form commits it and it comes back down as props.

**On the `keyedFormDataRef` a few lines below**, which is unchanged and therefore outside this diff: it is kept on purpose and is a different thing from the copy removed here. It is overwritten from the current render on every render and never written by a handler, so it always reflects props. Its only job is to let the four mutation handlers read the latest rows without listing `keyedFormData` as a dependency, which keeps their references stable and the row memo boundaries intact. A stability device, not a mirror of the data.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome, line 811: **Efficiency regression:** removing the old `hashObject` deep-equality gate means `keyedFormData` (and everything downstream) is recomputed on every render where the `formData` prop reference changes, even when the content is unchanged.

`Form.processPendingChange` in `Form.tsx` calls `structuredClone(oldFormData)` on every field edit anywhere in the form, so every keystroke gives every `ArrayField` a new-but-deep-equal `formData` reference. The old code's `newHash !== formDataHash` check caught this and fully bailed out (state untouched). The new code only gates on `itemKeys.length !== items.length`, so it always reallocates the keyed array/objects here, cascading into `NormalArray`'s `formData` and `canAdd` `useMemo`s and the per-item props/JSX rebuild — wasted work on every unrelated keystroke in forms with array fields, worse as array size or field count grows.

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -816,49 +801,30 @@ interface KeyedFormDataState<T = any> {
...
- * `keyedFormData` this hook also returns an `updateKeyedFormData()` function for use by the `ArrayField`. The detection
- * of external `formData` are handled by storing the hash of that `formData` along with the `keyedFormData` associated
- * with it. The `updateKeyedFormData()` will update that hash whenever the `keyedFormData` is modified and as well as
- * returning the plain `formData` from the `keyedFormData`.
+/** Pairs each item of the `formData` prop with a stable React key. Only the keys live in state: the items are read
+ * from props on every render, so the rows always show the value the form actually holds, never a proposal the field
+ * made and the form did not commit. The keys are UI metadata, kept so that React reuses row instances (and their DOM
+ * focus) across accepted adds, removes and reorders. When the array's length changes outside the handlers, an external
+ * replacement or a proposal the form transformed, there is no way to tell which rows survived, so every key is
+ * regenerated, as the previous hash-based detection also did.
```
heath-freenome: This JSDoc's closing clause, "...every key is regenerated, as the previous hash-based detection also did," is diff-style before/after commentary comparing this code to the implementation being removed. Per this repo's CLAUDE.md: "Don't leave 'before/after' or diff-style commentary in code comments... That belongs in commit messages and PR descriptions, not in code that has to keep reading true after the change ships." Once this merges, the comment references code a future reader will never see.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome, line 800: This is a core state-management rewrite of `ArrayField` shipping with zero test file changes in this PR. The described render-stability gate (`renderStability.test.tsx`, 216 passing tests) only exists on the unmerged dependent branch #5297, not in this repo/diff. Per CLAUDE.md: "Snapshot tests in `@rjsf/snapshot-tests` are shared across theme packages — run `test:update` there when changing core rendering." As submitted against `v7`, there's no in-repo test evidence for the refactor's claimed behavior-equivalence for a reviewer to run in CI here.

packages/core/src/components/fields/ArrayField.tsx
```diff
@@ -816,49 +801,30 @@ interface KeyedFormDataState<T = any> {
...
-        ? previousKeyedFormData.map((previousKeyedFormDatum, index) => ({
-            key: previousKeyedFormDatum.key,
-            item: nextFormData[index],
-          }))
-        : generateKeyedFormData<T>(nextFormData);
-    formDataHash = newHash;
-    setState({ formDataHash, keyedFormData });
+  let itemKeys = keys;
+  if (itemKeys.length !== items.length) {
+    itemKeys = items.map(generateRowId);
```
heath-freenome: Minor: `items.map(generateRowId)` is duplicated verbatim here and in the `useState` initializer above (`items.map(generateRowId)`). If row-key generation logic ever changes, both call sites need updating in lockstep, or initial mount and later length-mismatch regeneration will diverge.

heath-freenome (commented):
Automated code review — 4 findings (1 efficiency regression, 1 test-coverage gap, 1 CLAUDE.md comment-style violation, 1 minor duplication). See inline comments.

packages/core/src/components/fields/ArrayField.tsx
jimmycallin: The efficiency claim holds only against `v7` alone, and this PR is gated on #5297 (stated in the description and in the merge condition). #5297 replaces `Form`s `structuredClone` with structural sharing, so an edit to one field hands every untouched `ArrayField` the *same* `formData` reference it already had. `items` is then reference-stable, the `useMemo` on `[itemKeys, items]` does not recompute, and nothing downstream reallocates — which is exactly what the old `hashObject` gate was paying a per-render deep walk to discover. Its render-stability suite is what measures this: on #5297s head with this commit cherry-picked, every sibling render count is still zero.

So the deep-equality gate is not lost, it moves one level up and stops being per-array-field work. Landing this before #5297 would indeed be a regression; that is why the merge condition is there.

packages/core/src/components/fields/ArrayField.tsx
jimmycallin: Right, that clause stops reading true once this merges. Dropped in `d3def9c0a`; the sentence now just states the rule (a length change outside the handlers regenerates every key, because there is no way to tell which rows survived). The comparison against the hash-based implementation lives in the PR description, where it belongs.

packages/core/src/components/fields/ArrayField.tsx
jimmycallin: Fair — the render-stability gate is on #5297, but that is no reason for this PR to carry no in-repo evidence of its own. Added two tests to `ArrayField.test.tsx` in `d3def9c0a`, both runnable against `v7` here, covering the paths the rewrite actually changed:

- an external `formData` replacement of the same length keeps every row key and renders the new items from props;
- a replacement of a different length regenerates every row key.

Add/remove/reorder key stability was already covered by the existing `ArrayKeyDataAttr` tests. Full `packages/core` suite: 2054 passing, zero snapshot updates in any theme suite.

packages/core/src/components/fields/ArrayField.tsx
jimmycallin: Fixed in `d3def9c0a` — a local `freshKeys()` now backs both the `useState` lazy initializer and the length-mismatch branch, so mount and regeneration cannot drift.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: Thanks, that reasoning holds. With #5297's structural sharing, `items` keeps its reference for untouched arrays, so the `[itemKeys, items]` memo doesn't recompute. The deep-equality work moves up to one walk at the root. I'm fine with dropping the per-array `hashObject` gate on that basis.

My concern is that this PR is only correct because of #5297, but nothing in the repo enforces that. The base is still `v7`, #5297 is still open, and the description states the condition only in prose. If this merges first, we ship the regression to every form with array fields (a new `keyedFormData` and downstream rebuild on every keystroke anywhere in the form), and I don't think a note in the description is enough to prevent that. Could we make the ordering explicit? Any one of these works for me:

1. Retarget this PR onto #5297's branch so it can't land before it.
2. Convert it to a draft (or add a blocking label) until #5297 merges, then rebase onto `v7` and mark it ready.
3. If neither suits, add a test in this PR that fails on `v7` alone when a new-but-deep-equal `formData` reference triggers a recompute, so the dependency is caught in CI.

Separately, the description cites #5297 at `8d84950`, but its head is now `c92882a`. Could you re-run the render-stability gate against the current head and update the description with the result, before this merges?

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: Confirmed. The JSDoc now states only the rule and its rationale, and the comparison to the old implementation is gone. Resolving from my side.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: Thanks, this addresses it. I ran the `@rjsf/core` suite at `b7aaa31` and it passes. Both new tests also pass against the `v7` implementation of `ArrayField`, which supports the behavior-preserving claim. With the length-mismatch regeneration disabled, the different-length test fails, so it does catch a regression there.

Non-blocking: the different-length test only goes from 2 items to 1, so it checks a single key. A growth case (1 to 2, asserting both keys differ from the start) would cover the other direction, but I'm fine with this as is.

packages/core/src/components/fields/ArrayField.tsx
heath-freenome: Confirmed. `freshKeys()` now backs both the `useState` initializer and the length-mismatch branch, so mount and regeneration can't drift. Resolving from my side.

packages/core/src/components/fields/ArrayField.tsx
jimmycallin: Taken your option 2, with a note on why option 1 isn't available: this PR targets `rjsf-team:v7`, and a base branch has to live in `rjsf-team`, while `poc/shallow-comparator` only exists on my fork. So there is no way to stack it on GitHub. It's now a draft, and I'll rebase onto `v7` and mark it ready once #5297 merges. The merge condition in the description says that outright rather than leaving it as prose about the dependency.

On the gate: re-run against #5297's current head, `4e4fe7541` (it moved again since `c92882a` — three review fixes on that PR, see its threads). These commits cherry-pick cleanly onto it; `renderStability.test.tsx` + `ArrayField.test.tsx` there are 221 passing with every sibling render count still zero, and the whole `@rjsf/core` suite on that combination is 2142 passing. Section 5 of the description now records that.

packages/core/src/components/fields/ArrayField.tsx
jimmycallin: Added the growth case in `17dc59585`: 1 item to 2, asserting that neither end key is the one the field started with.

```ts
it('should regenerate every row key when the replacement is longer', () => {
  const { node, rerender } = createFormComponent({ schema, formData: ['foo'], templates });
  const startKeys = rowKeys(node);

  rerender({ schema, formData: ['foo', 'bar'], templates });

  const endKeys = rowKeys(node);
  expect(endKeys).toHaveLength(2);
  expect(endKeys).not.toContain(startKeys[0]);
});
```

Checked it earns its place the same way you checked the shrink case: with the length-mismatch regeneration disabled it fails, and it fails on the length assertion rather than the key one, because the stale key list is then shorter than the items.

heath-freenome (approved):
Approved pending merge of 5297

jimmycallin:
#5297 is in, so this is rebased onto `v7` at `0d8cb675f` and out of draft. Head `52c7d1aea`, a clean rebase with no conflicts.

The merge condition in the description is replaced by what actually happened. Verified on the rebased branch: root `typecheck` green, `@rjsf/core` 2150 passing, and `renderStability.test.tsx` — the gate that only existed on #5297's branch when this was written — now runs here and passes, so sibling rows still do not re-render.

PR #5300, opened 2026-09-18
test(core): controlled parent harnesses and seeded-fixture classification (ownership RFC, lane B)

jimmycallin:
Lane B of the [form data ownership RFC](https://github.com/jimmycallin/react-jsonschema-form/pull/62) ([spec, section 7](https://github.com/jimmycallin/react-jsonschema-form/blob/rfc/form-state-ownership/rfcs/form-state-ownership-spec.md)). Tests only: no file under `src/` changes. It does not depend on #5297 or #5296 and conflicts with neither.

## 1. What a controlled parent is, and what a spy is not

In React, a controlled component renders a value its parent owns. The component never changes that value itself. It *proposes* a change through a callback, and the parent decides what to do with the proposal: store it as is, refuse it, or store something derived from it. Whatever the parent decides becomes the next value the component renders. An `<input value={x} onChange={...}>` works exactly this way, and so does `<Form formData={x} onChange={...}>`.

The loop closes only if the parent hands a value back. A test that passes `formData` together with `onChange={vi.fn()}` records the proposal and stops there. The spy never becomes the next `formData` prop, so the form keeps rendering whatever it decided on its own. Such a test exercises the form's internal state and tells us nothing about controlled composition: whether two edits compose through the parent, whether a refused edit disappears, whether a normalized value comes back to the field. Today's `Form` happens to pass these tests because it keeps a private copy of the data and edits that copy regardless of what the parent does. That private copy is the hybrid behavior the RFC removes, and when it goes, every test built on a spy will be testing something that no longer exists.

## 2. What this PR adds

**Three parents that hand values back**, in `testUtils.tsx`, each with a stated policy:

- `AcceptingParent` stores each proposal with an ordinary `useState` update, exactly `setData(event.formData)`. This is the pattern the migration guide will recommend.
- `RejectingParent` keeps rendering its initial value and records the proposals. This is the shape of a parent whose own validation refused the change.
- `TransformingParent` stores a caller-supplied transform of each proposal. This is the shape of a parent that normalizes what it is handed.

Each accepts the ordinary `Form` props apart from `validator`, `formData` and `onChange`, and an optional log the test can read for the parent's current value and the sequence of proposals it received.

**A test file that proves each parent against today's `Form`**, `Form.controlledParents.test.tsx`. Two of its assertions characterize current behavior rather than the target contract, and say so in a comment:

- With a rejecting parent, today's form renders the refused edit. The ownership switch (lane A, step A4) flips that assertion: a controlled form will render the parent's value.
- The dependent-field clear from #3367, where a second field clears itself from an effect when its sibling changes, survives today when the parent accepts. That is a contract the switch must keep, so the assertion stays as it is.

It also carries the near-simultaneous-change scenario from the spec, two widgets each changing a field from an effect on mount, through a real accepting parent. The existing version in `Form.handlers.test.tsx` merges into an external variable, which proves the form emitted both changes but not that they compose through a parent; it is kept unchanged for its domain coverage.

**A classification of every seeded fixture**, and a migration of the ones that describe a seeded, self-owned form.

## 3. The classification

Every test in the six `Form.*.test.tsx` suites that passes a `formData` prop was read and sorted into one of three categories:

- **(a) Editable seeded.** The test seeds a value and then edits the form, with no parent handing the edit back. It is describing a form that owns its data and started from a seed, which is what `initialFormData` means. Switching the prop preserves behavior exactly, and every assertion is unchanged.
- **(b) Accepting controlled.** A genuine parent stores `event.formData`. The five error-clearing tests in `Form.errors.test.tsx` had an inline `Controlled` component each; they now use `AcceptingParent`, which deletes five copies of the same twelve lines.
- **(c) Intentionally fixed, or prop-driven.** The test renders or submits a fixed value without editing, or drives the form by re-rendering with new props, or asserts the legacy hybrid behavior on purpose. Left untouched; the ones that assert hybrid behavior are the ownership switch's to change.

| File | Test | Category | Action |
| --- | --- | --- | --- |
| `Form.behaviors` | omitExtraData / liveOmit onBlur, 3 tests | (a) | `initialFormData` |
| `Form.behaviors` | omitExtraData / liveOmit onChange, 9 tests (call/not call, omit/not omit, additionalProperties, nested rename, oneOf, anyOf, nested if/then) | (a) | `initialFormData` |
| `Form.behaviors` | prunes empty optional object on change / on blur | (a) | `initialFormData` |
| `Form.behaviors` | keeps a root-level property named the empty string | (a) | `initialFormData` |
| `Form.behaviors` | submits valid formData after a nested string field is cleared | (a) | `initialFormData` |
| `Form.behaviors` | preserves an empty array when enabling the dependency branch; removes enum values when dependency switches; array item defaults switch with mergeExtraDefaults | (a) | `initialFormData` |
| `Form.behaviors` | field literally named "constructor"; omitExtraData on submit, 4 tests; bracket-notation names, 2 tests; no oneOf branch matches on submit | (c) render or submit only | left |
| `Form.behaviors` | Should update state when data updated from invalid to valid | (c) prop-driven `rerender` | left |
| `Form.behaviors` | Form resets without initial data when it is controlled | (c) asserts hybrid snap-back on purpose | left for A4 |
| `Form.errors` | clean contextualized errors when fixed; denote error in nested item; only show errors for selected branch; no errors when branch empty; sanitize stale enum data, 3 tests; customValidate Start/End | (a) | `initialFormData` |
| `Form.errors` | error clearing with a controlled parent, 5 tests (#5197 and dotted names) | (b) | inline parent replaced by `AcceptingParent` |
| `Form.errors` | contextualize the error for nested array indices | (c) submit only | left |
| `Form.handlers` | change handler with schema and uiSchema; blur handler; focus handler | (a) | `initialFormData` |
| `Form.handlers` | submit handler with form state; not called on validation errors; error handler | (c) submit only | left |
| `Form.handlers` | should call last provided change handler | (c) prop-driven `rerender` | left |
| `Form.handlers` | two near simultaneous changes | (c) external-variable merge, kept for domain coverage | left; real-parent version added in the new file |
| `Form.handlers` | switching to null option in oneOf | (c) asserts controlled prop echo on purpose | left for A4 |
| `Form.defaults` | should not set default when a text field is cleared | (a) | `initialFormData` |
| `Form.defaults` | referenced definitions for array items; null / non-object additionalProperties, 2 tests | (c) render only | left |
| `Form.state` | all 9 seeded tests | (c) submit only, or prop-driven `rerender` | left |
| `Form.props` | replace state when props remove / change formData keys | (c) asserts prop replacement on purpose | left for A4 |
| `Form.props` | disabled / readonly / submit button, 10 tests; portal submit; duplicate enum in dependencies | (c) render or submit only | left |
| `Form.rendering` | onChange called / not called on mount | (c) asserts mount notification | left |

Thirty-three fixtures moved to `initialFormData`, five inline parents replaced, everything else untouched.

## 4. Why this is behavior-preserving

Every migrated test passes with its assertions unchanged. `initialFormData` seeds the form's own state exactly as a `formData` prop with no accepting parent does today, and the form then edits that state either way. The five harness migrations replace one accepting parent with another that does the same `setState`. No assertion was weakened, loosened or reordered.

The core suite: 39 files, 2057 tests, all passing. Root `typecheck`, `lint`, `knip` and `cs-check` pass.

## 5. Why now

The ownership switch cannot be reviewed against fixtures that pass for the wrong reason. Once the private copy is gone, a spy-based test either fails or silently passes because the form no longer edits at all. Sorting the fixtures now, while behavior is unchanged, means the switch's diff shows only the tests whose expectations genuinely change, each of which is listed above as "left for A4".


packages/core/test/testUtils.tsx
```diff
@@ -26,6 +27,77 @@ export interface ConsoleSuppressionResult {
   readonly consoleSpy: MockInstance;
 }
 
+/** What a controlled parent has done so far: the value it currently renders and every proposal the form sent it.
+ * A `vi.fn()` `onChange` is not a controlled parent: it records the proposal but never hands it back as the next
+ * `formData` prop, so a test built on one exercises the form's own state and nothing about controlled composition.
+ * These parents do hand values back, each with a different policy, so a test can state which policy it relies on.
+ */
+export interface ControlledParentLog<T> {
```
jimmycallin, line 35: **Why a log object rather than a spy.** A `vi.fn()` gives you the sequence of proposals but not what the parent decided to render, and the decision is the whole point of a controlled parent. `value` is what the parent currently renders, `proposals` is what the form asked for. The gap between the two is what a rejecting or transforming parent is for. It is a plain mutable object because the parent has to write to it from render, and a test reads it after `act` has settled.

packages/core/test/testUtils.tsx
```diff
@@ -26,6 +27,77 @@ export interface ConsoleSuppressionResult {
...
+  return { value: undefined, proposals: [] };
+}
+
+export type ControlledParentProps<T> = Omit<FormProps<T>, 'validator' | 'formData' | 'onChange'> & {
+  initialValue?: T;
+  log?: ControlledParentLog<T>;
+};
+
+/** Stores every proposal as its next value, exactly `setData(event.formData)`, the ordinary React controlled pattern */
+export function AcceptingParent<T>({ initialValue, log, ...formProps }: ControlledParentProps<T>) {
```
jimmycallin, line 50: **Why `setValue(event.formData)` and not an updater.** This is the accepting pattern in its simplest form, the same shape as `onChange={(e) => setX(e.target.value)}` on a DOM input, and it is what the migration guide will recommend. The spec (section 5) asks for exactly `setData(event.formData)` so the near-simultaneous scenario tests composition through a real state update, not through a reducer that merges.

jimmycallin (commented):
Self-review: notes on the harness design and on the two assertions whose fate under the ownership switch differs, as promised in the description.

packages/core/test/testUtils.tsx
```diff
@@ -26,6 +27,77 @@ export interface ConsoleSuppressionResult {
...
+  const [value, setValue] = useState(initialValue);
+  Object.assign(log ?? {}, { value });
```
jimmycallin, line 52: `Object.assign` rather than `log.value = value` only to satisfy `no-param-reassign`; the log exists to be written to. Writing during render is deliberate: it keeps `log.value` equal to the value the form is rendering in this very commit, with no effect-timing gap for a test to trip over.

packages/core/test/testUtils.tsx
```diff
@@ -26,6 +27,77 @@ export interface ConsoleSuppressionResult {
...
+      onChange={(event) => {
+        log?.proposals.push(event.formData);
+        setValue(event.formData);
+      }}
+    />
+  );
+}
+
+/** Keeps rendering its initial value whatever the form proposes, the shape of a parent whose validation refused */
+export function RejectingParent<T>({ initialValue, log, ...formProps }: ControlledParentProps<T>) {
```
jimmycallin, line 67: **Why there is no state here.** A rejecting parent is one whose value never moves. Modelling that with `useState` and an updater that returns the old value would re-render the parent for no reason and hide the point: the `formData` prop the form receives is the same object on every render, so anything the form shows beyond it is the form's own doing.

packages/core/test/testUtils.tsx
```diff
@@ -26,6 +27,77 @@ export interface ConsoleSuppressionResult {
...
+      {...formProps}
+      validator={validator}
+      formData={initialValue}
+      onChange={(event) => log?.proposals.push(event.formData)}
+    />
+  );
+}
+
+/** Stores a transformed version of each proposal, the shape of a parent that normalizes what it is handed */
+export function TransformingParent<T>({
```
jimmycallin, line 80: The transform runs on the proposal, not on the field's DOM value, so a multi-character `user.type` produces a chain: the form proposes `a`, the parent commits `A`, the field now shows `A`, the next keystroke proposes `Ab`, the parent commits `AB`. The test in the new file asserts that exact chain, which is how you know the form really renders the parent's value and not its own.

packages/core/test/Form.controlledParents.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const log = createParentLog<Data>();
+    const { container } = render(<RejectingParent<Data> schema={schema} initialValue={{ name: 'a' }} log={log} />);
+
+    await user.type(input(container, 'root_name'), 'b');
+
+    expect(log.value).toEqual({ name: 'a' });
+    expect(log.proposals).toEqual([{ name: 'ab' }]);
+    // Today's Form keeps its own copy of the data and renders the edit even though the parent refused it. Under the
+    // ownership contract (RFC, section 3.2) a controlled form renders the parent's value, so A4 flips this to 'a'.
+    expect(input(container, 'root_name')).toHaveValue('ab');
```
jimmycallin, line 51: **This assertion characterizes today's behavior and the ownership switch flips it.** The parent refused the edit and still renders `{ name: 'a' }`, yet the input shows `ab`, because today's `Form` keeps a private copy of the data and edited that. Under the RFC's section 3.2 a controlled form renders the parent's value, so lane A step A4 changes this to `'a'`. It is asserted rather than skipped so the switch's diff shows the change explicitly.

packages/core/test/Form.controlledParents.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    );
+
+    await user.type(input(container, 'root_name'), 'ab');
+
+    expect(log.proposals).toEqual([{ name: 'a' }, { name: 'Ab' }]);
+    expect(log.value).toEqual({ name: 'AB' });
+    expect(input(container, 'root_name')).toHaveValue('AB');
+  });
+
+  it('two near-simultaneous changes both reach an accepting parent', async () => {
```
jimmycallin, line 68: Same scenario as `should call change handler with proper data after two near simultaneous changes` in `Form.handlers.test.tsx`, but the parent is real. The existing test merges into an external variable, which proves the form emitted both changes; it cannot prove that the second composes onto the first after a round trip through the parent's state, which is what a controlled consumer actually experiences.

packages/core/test/Form.controlledParents.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+          initialValue={{ name: 'a', other: 'b' }}
+          log={log}
+        />,
+      );
+    });
+
+    await waitFor(() => expect(log.value).toEqual({ name: 'a2', other: 'b2' }));
+  });
+
+  it('a dependent field clearing itself from an effect when its sibling changes (#3367)', async () => {
```
jimmycallin, line 103: **Why the harness is not used here.** The second field has to observe the first, and the only channel a field has to a sibling is `formContext`. That means `formContext` must derive from the parent's own state on every render, which the harness cannot do with static props. So this parent is written out. It still does exactly `setData(event.formData)`.

packages/core/test/Form.controlledParents.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+          formContext={{ name: data?.name }}
+          onChange={(event) => setData(event.formData)}
+        />
+      );
+    }
+    const { container } = render(<Parent />);
+
+    await user.type(input(container, 'root_name'), 'x');
+
+    await waitFor(() => expect(committed.at(-1)).toEqual({ name: 'x', other: null }));
```
jimmycallin, line 142: Today the clear survives: the parent's last committed value has both the typed `name` and `other: null`. That is not the failure #3367 reported, which was under the older reconciler, so this test is a contract the ownership switch must keep rather than an assertion it flips. If A4 ever loses the clear, this is the test that fails.

packages/core/test/Form.errors.test.tsx
```diff
@@ -94,20 +93,13 @@ describeRepeated('Form common: error contextualization', (createFormComponent) =
...
-                formData={formData}
-                noHtml5Validate
-                onChange={(e) => setFormData(e.formData)}
-              />
-            );
-          }
-
-          const { container } = render(<Controlled />);
+          const { container } = render(
+            <AcceptingParent<Record<string, unknown>>
```
jimmycallin, line 97: Before: twelve lines of inline `Controlled` component, repeated five times in this file with only the initial value and an optional `ref` differing. After: one line naming the policy. `ref` passes straight through because the harness forwards every `Form` prop except the three it owns, and the type parameter is spelled out because nothing else in the call pins `T`.

packages/core/test/Form.behaviors.test.tsx
```diff
@@ -151,7 +151,7 @@ describe('omitExtraData and live omit onBlur', () => {
     const onBlur = vi.fn();
     const { node, onChange } = createFormComponent({
       schema,
-      formData: formData1,
+      initialFormData: formData1,
```
jimmycallin, line 154: **Why `initialFormData` is the faithful spelling of this fixture.** The test seeds a value, then edits the field, then asserts what `onChange` reported. No parent hands anything back, so the form is editing its own state either way. `formData` here only ever worked because of the hybrid private copy; `initialFormData` says what the test means and keeps working after the copy is gone. Every assertion in the migrated tests is byte-for-byte unchanged.

packages/core/test/Form.controlledParents.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    expect(log.proposals.at(-1)).toEqual({ name: 'abc' });
+  });
+
+  it('a rejecting parent records proposals while its value stays put', async () => {
+    const log = createParentLog<Data>();
+    const { container } = render(<RejectingParent<Data> schema={schema} initialValue={{ name: 'a' }} log={log} />);
+
+    await user.type(input(container, 'root_name'), 'b');
+
+    expect(log.value).toEqual({ name: 'a' });
```
heath-freenome: This assertion can't fail: `RejectingParent` (in `testUtils.tsx`) sets `log.value` to `initialValue` on every render, so it's comparing `initialValue` with itself. If `RejectingParent` were later changed to store the proposal, this would still pass. The rejection is already covered by `log.proposals` and the rendered input value. Suggest either removing this line or having `log.value` record what the parent actually renders.

heath-freenome (approved):
Approved pending one proposed, potential test fix

packages/core/test/Form.controlledParents.test.tsx
jimmycallin: Correct — `RejectingParent` assigns `log.value = initialValue` on every render, so that line compared `initialValue` with itself and would survive the harness being changed to store the proposal. Removed in `1eb2ff728`; `log.proposals` and the rendered input value already carry the real assertion, and the input-value line is the one that flips under A4.

Issue #1813, opened 2020-06-14
Modify JSON Validation for Optgroup support


Enola-guy:
Hello, I've successfully created the widget to support optgroup select (i.e. categories inside a select). I'm now seeking help for the validation. 

In order to generate the categories I give a dict instead of a list to my "enum" field. How can I modify the JSON validation to avoid the error ```.enum should be array```

**Here is the codepen of my implementation:**  
https://codepen.io/enola-guy/pen/ZEQOVEN

If you can explain me the steps missing, I would gladly open a PR in order to make this feature in the main branch :) 

![image](https://user-images.githubusercontent.com/17969523/84604418-425dea80-ae96-11ea-8d58-8cbcd82b03f9.png)

epicfaace:
@Enola-guy thanks for your custom widget! I don't think we can merge this into the main library, though, until we have a JSON Schema-compliant approach (and in JSON Schema, `enum` must be an array, not a dict). Can you think of a JSON-schema-compliant way of using optgroups? For example, this might involve customizing the optgroups in uiSchema using something like this:


```json
{
    "ui:groups": [
         { "title": "Theropods", "items": ["Tyrannosaurus", "Velociraptor"] },
         ...
     ]
}
```

This is just an idea -- let me know if you have any other thoughts / ways of doing this.

Enola-guy:
Thanks for the hint, I'm gonna try this, unfortunately it's not going to be DRY, but that's something. Will update the code today.

Enola-guy:
Hey late update, thank you so much it worked `ui:groups` did the work ! @epicfaace

epicfaace:
Would you be interested in contributing your widget to this library as a PR?

Enola-guy:
I would be very interested !

VladislavArkavenko:
@Enola-guy Can you please share your working code here?

nagaozen:
I made a PR implementing this feature:

![image](https://github.com/user-attachments/assets/156b331f-de99-4dda-ac2f-6ce460f88f0e)

It extends the current SelectWidget to support `optgroups` at `ui:options`. e.g.:

```json
  "selectWidgetOptions3": {
    "ui:options": {
      "optgroups": {
        "lipsum": [
          "foo"
        ],
        "dolorem": [
          "bar"
        ]
      }
    }
  }
```

Issue #580, opened 2017-05-17
Adding option group support for select


vamsiampolu:
Is it possible to support option groups for select using `react-jsonschema-form` configuration, I do understand that I can add non standard properties to 'ui:options' but is there any way to do it without having to go for that.

It would also be great if we could automatically use the `maxItems` and `minItems` validation could be added to the multiple Select widget instead of having to recieve it from the schema and use it ourselves.

glasserc:
I'm sorry, I don't really understand. Is this two issues? I feel like option groups should be one issue (to which I think the correct answer is to add something to `uiSchema`). I don't see how it relates to `maxItems` and `minItems` in the MultiSelect case.

epicfaace:
Closing this issue due to inactivity. @vamsiampolu if you want to clarify your request, please do.

nagaozen:
Implemented as mentioned at https://github.com/rjsf-team/react-jsonschema-form/issues/1813#issuecomment-2467444265

PR #5301, opened 2026-09-18
feat: add optgroups support to SelectWidget across all themes

heath-freenome:

Groups a select-backed widget's enum options into labeled sections via ui:options.optgroups, closing #580 (and #1813). Reimplements the approach from the stale v6 PR #5000 against the current v7 SelectWidget architecture, extending it to every theme instead of just core and react-bootstrap.

The grouping/partitioning logic lives once in @rjsf/utils as groupEnumOptions()/isEnumOptionsGroup(); each theme's SelectWidget only supplies the rendering primitive its own UI library uses for a group (native <optgroup>, MUI's ListSubheader, antd's nested options, Mantine's data groups, Chakra's Select.ItemGroup, Fluent's OptionGroup, a custom header row for daisyUI, and grouped CommandGroups for shadcn).

Adds unit tests for the new utils functions and per-theme SelectWidget tests, two shared snapshot-tests cases regenerated across all themes, a playground example, uiSchema.md docs, and v7 upgrade guide/changelog entries.

### Checklist


packages/mui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -115,19 +137,14 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
...
-                disabled={isDisabled}
-              >
-                {enumLabel}
-              </MenuItem>
-            );
-          })}
+        {groupEnumOptions<S>(enumOptions, optgroups, enumDisabled).flatMap((item) =>
+          isEnumOptionsGroup<S>(item)
+            ? [
+                <ListSubheader key={`optgroup-${item.label}`}>{item.label}</ListSubheader>,
```
jimmycallin: MUI's `SelectInput` clones **every** direct child of the `Select` with `role: 'option'`, `aria-selected` and an `onClick` handler — it has no special case for `ListSubheader` (v9 `Select/SelectInput.js`, the `childrenArray.map(child => …React.cloneElement(child, {role: 'option', …}))` block), and the `role` from `cloneElement` wins over any role set here.

Verified against the installed `@mui/material@9.1.2` by rendering this widget with `optgroups: { 'Group A': ['foo','bar'] }` and dumping the open listbox:

```html
<li class="MuiListSubheader-root …" aria-selected="false" role="option">Group A</li>
```

`getAllByRole('option')` returns 6 for a 4-option enum. So a screen-reader user hears each group heading announced as a selectable option, and the heading is not exposed as a group at all — the opposite of the native `<optgroup>` semantics the other themes get. (Clicking it is harmless: `handleItemClick` bails on the missing `tabindex`.)

`aria-hidden` is not overwritten by the clone, so it survives:

```suggestion
                <ListSubheader key={`optgroup-${item.label}`} aria-hidden>
                  {item.label}
                </ListSubheader>,
```

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
```diff
@@ -146,39 +190,20 @@ export default function SelectWidget<
...
-                </div>
-              </li>
-            );
-          })}
+          {groupEnumOptions<S>(optionsList, optgroups).flatMap((item) =>
+            isEnumOptionsGroup<S>(item)
+              ? [
+                  <li
+                    key={`optgroup-${item.label}`}
+                    role='presentation'
```
jimmycallin: `role='presentation'` strips the element's semantics entirely, so assistive tech sees no grouping at all here — worse than the bare text, the label becomes an unlabelled orphan inside `role='listbox'`. A native `<optgroup>`, which the other themes render and which this row is standing in for, maps to `role="group"` with the label as its accessible name.

Concretely: with `optgroups: { 'Group A': ['foo','bar'], 'Group B': ['baz'] }`, a screen-reader user navigating this dropdown hears `foo`, `bar`, `baz` with no indication that they belong to different groups, while the same form under `@rjsf/core`/`react-bootstrap` announces the group.

Wrapping each group's options in a `<ul role='group' aria-label={item.label}>` (with the visible header inside it, `aria-hidden`) would match `<optgroup>` semantics. If the flat `<li>` structure has to stay, at minimum `role='presentation'` should be dropped so the text is still exposed.

packages/utils/src/groupEnumOptions.ts
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      byValue.set(option.value, [option]);
+    }
+  });
+
+  const claimedIndices = new Set<number>();
+  const groups: EnumOptionsGroupType<S>[] = Object.entries(optgroups)
+    .map(([label, values]) => {
+      const options = values.reduce<IndexedEnumOptionType<S>[]>((acc, value) => {
+        const option = byValue.get(value)?.find((candidate) => !claimedIndices.has(candidate.index));
```
jimmycallin: Group membership is resolved by strict `Map` key equality on the raw enum value, which is stricter than the sibling `enumOrder` feature this sits next to: `applyEnumOrder` in `optionsList.ts` keys its map by `String(opt.value)` and looks up `String(entry)`.

Two consequences, both silent (the group just vanishes, since empty groups are filtered out):

- `enum: [1, 2, 3]` with `'ui:options': { optgroups: { Low: ['1', '2'] } }` produces no grouping, while `'ui:enumOrder': ['1','2','*']` on the same schema works. uiSchemas authored as JSON / by form builders routinely stringify these.
- Object- or array-valued enums (the `enumObjects` playground sample's own shape) can *never* be grouped, because `Map` keys compare by reference and the `optgroups` literal is a different object than the one in `enumOptions`.

Matching on `String(value)` here would make `optgroups` behave like `enumOrder`; if the strictness is deliberate, the `optgroups` docs in `uiSchema.md` should say that values must match the enum entries by identity.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -472,3 +489,5 @@ The following new types were added to `@rjsf/utils`:
...
 - **`useTimeWidgetProps(props)`**: hook that combines the three functions above into a `TimeWidget`'s offset-free display value and a `computeTimeValue()` transform for its `onChange` handler. Shared by the `TimeWidget`s in `@rjsf/core`, `@rjsf/daisyui`, and `@rjsf/mantine`.
 - **`getDateTimeLocalValue(schema, value)`**: reports whether `schema.format` is `iso-date-time` and, if so, strips a timezone offset from the value for display. Shared by the `DateTimeWidget`s in `@rjsf/core`, `@rjsf/antd`, `@rjsf/daisyui`, and `@rjsf/mantine`.
+- **`groupEnumOptions(enumOptions, optgroups?, enumDisabled?)`**: groups `enumOptions` according to `ui:options.optgroups`, tagging every option with its original array index (needed for index-based value encoding) and its disabled status. Powers the `optgroups` support in every theme's `SelectWidget`, described above.
+- **`isEnumOptionsGroup(item)`**: type guard distinguishing a group from a standalone option in the list `groupEnumOptions()` returns.
```
jimmycallin, line 568: `flattenGroupedOptions()` is also a new public export of `@rjsf/utils` (used by the chakra-ui and shadcn `SelectWidget`s), but it's missing from this list and from the `## @rjsf/utils` entry in `CHANGELOG-v7.md`, which mention only `groupEnumOptions()` and `isEnumOptionsGroup()`. Consumers writing a custom themed `SelectWidget` need it for exactly the reason its own JSDoc gives.

packages/chakra-ui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -61,23 +64,28 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
...
-        return {
-          label: enumLabel,
-          value: enumOptionValueEncoder(enumValue, index, optionValueFormat),
-          disabled: Array.isArray(enumDisabled) && enumDisabled.includes(enumValue),
-        };
-      });
-      if (showPlaceholderOption) {
-        (computedOptions as any[]).unshift({ value: '', label: placeholder || '' });
-      }
+    const computedOptions: OptionsOrGroups<any, any> = flattenGroupedOptions<S>(groupedOptions).map(toItem);
```
jimmycallin: (Anchored here; the code in question is `createListCollection({ items: displayEnumOptions.filter((item) => item.value) })` a few lines below, at line 100-102.)

The collection and the rendered items can now disagree. Before this PR the `<Select.Item>`s were rendered from `selectOptions.items`, so an option dropped by this `.filter((item) => item.value)` simply wasn't rendered. Now rendering comes from `groupedOptions` (below), while the collection is still built from the filtered list.

Concrete case: `ui:options: { optionValueFormat: 'realValue' }` with an enum containing `''` or `null`. `enumOptionValueEncoder` returns `''` for those, the filter drops them from the `createListCollection` items, but `renderItem` still emits a `Select.Item` for them — Ark then has an item in the DOM that its collection doesn't know about, so it can't be highlighted or selected (and `collection.getItemDisabled`/`getItemValue` lookups miss).

Either build the collection from the same `displayEnumOptions` that drives rendering (dropping the filter, keeping only the placeholder exclusion), or filter `groupedOptions` with the same predicate before rendering.

packages/mui/src/SelectWidget/SelectWidget.tsx
heath-freenome: Good catch, thanks. Applied your `aria-hidden` suggestion, so the heading is no longer announced as a selectable option (a30db7413). MUI's `Select` clones every child with `role="option"` and has no group primitive, so this is the closest I can get: the heading is hidden from assistive tech rather than exposed as a group. Added a test that asserts the heading is `aria-hidden` and doesn't show up as an option.

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
heath-freenome: Agreed, dropped `role='presentation'`. Each group is now `<li role='presentation'><ul role='group' aria-label={label}>…</ul></li>`, with the visible header `aria-hidden` inside it (a30db7413). The `li` wrapper keeps the nested list valid HTML. Added a test that finds each group by role and name and checks it owns its options, and regenerated the two optgroup snapshots.

packages/utils/src/groupEnumOptions.ts
heath-freenome: You're right that it was stricter than `enumOrder`. It now matches like `enumOrder` does: an exact lookup first, then a string-form fallback for primitives, so `enum: [1, 2, 3]` with `optgroups: { Low: ['1', '2'] }` groups correctly (09fdc297c). An exact match wins over a string-form one, so `'1'` prefers the string `'1'` over the number `1`.

For object and array enums I didn't use `String()`, since every object would stringify to `[object Object]` and collide. They only match the very same object, so a JSON-authored uiSchema can't group them. That matches `enumOrder`, which can't either, and I documented it in `uiSchema.md`. I also made `enumDisabled` in the same function follow the same rules, so the two options agree.

While in there I replaced the linear scans with indexed lookups, so it stays near-linear for large enums (2db817ff3).

packages/docs/docs/migration-guides/v7.x upgrade guide.md
heath-freenome: Added to the utility functions list in the upgrade guide and to the `@rjsf/utils` entry in `CHANGELOG-v7.md` (a30db7413).

packages/chakra-ui/src/SelectWidget/SelectWidget.tsx
heath-freenome: Fixed. The collection is now built from the same items that get rendered (a30db7413), and I dropped the placeholder code, which never reached the collection anyway.

Your `''`/`null` case turned out to be deeper than the mismatch. In `realValue` mode those encode to `''`, which `enumOptionValueDecoder` deliberately reads as "no selection", so such an option can never be picked and would always round-trip to `emptyValue`. So I skip those options from both the collection and the rendered items, and drop any group left empty (df0c6bd7a). That restores the pre-PR behavior. Added tests for both cases.

Related cleanup: that removed the last use of `chakra-react-select` in the package, so I dropped it from `peerDependencies` as a v7 breaking change, with changelog, upgrade guide and README updates (3e4010dcf).

packages/utils/src/groupEnumOptions.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+  });
+
+  const claimedIndices = new Set<number>();
+  const findUnclaimed = (candidates: IndexedEnumOptionType<S>[] | undefined) =>
+    candidates?.find((candidate) => !claimedIndices.has(candidate.index));
+  const findOption = (value: unknown) =>
+    findUnclaimed(byValue.get(value)) ?? (isPrimitive(value) ? findUnclaimed(byString.get(String(value))) : undefined);
+
+  const groups: EnumOptionsGroupType<S>[] = Object.entries(optgroups)
```
jimmycallin, line 83: **Group order is not "key order" when a label looks like an integer.**

The doc comment and `uiSchema.md` both promise one group "per non-empty key, in key order", but `Object.entries()` returns integer-like keys first, in ascending numeric order, ahead of every string key regardless of insertion order.

Concrete case:

```ts
'ui:options': {
  optgroups: {
    Newest: ['a', 'b'],
    '2024': ['c'],
    '2023': ['d'],
  },
}
```

renders as `2023`, `2024`, `Newest` — not the authored order. Year/number-style group labels are a pretty natural thing for users to write.

If ordering by author intent matters, the mapping needs an array form (or at least the docs should say numeric-looking labels are hoisted and sorted).

packages/chakra-ui/src/NativeSelectWidget/NativeSelectWidget.tsx
```diff
@@ -62,46 +63,25 @@ export default function NativeSelectWidget<
...
-    items: displayEnumOptions.filter((item) => item.value),
-  });
+  const selectedIndex = enumOptionsIndexForValue<S>(value, enumOptions, false);
+  const formValue = typeof selectedIndex !== 'undefined' ? selectedIndex.toString() : '';
 
   const chakraProps = getChakra({ uiSchema });
 
+  function renderOption(option: IndexedEnumOptionType<S>) {
+    return (
+      <option key={option.index} value={String(option.index)}>
```
jimmycallin, line 79: `option.disabled` is computed here (the widget threads `enumDisabled` into `groupEnumOptions()` just above) but never applied to the rendered `<option>`, so `ui:enumDisabled` still has no effect in chakra's native select — every "disabled" option stays selectable.

```suggestion
      <option key={option.index} value={String(option.index)} disabled={option.disabled}>
        {option.label}
      </option>
```

(The same gap exists in `packages/daisyui/.../SelectWidget.tsx`, which doesn't even destructure `enumDisabled` and so passes `undefined` to `groupEnumOptions()` — every other theme now honours it.)

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
```diff
@@ -117,11 +126,46 @@ export default function SelectWidget<
     .flat()
     .filter((v) => v !== '');
 
-  const optionsList =
+  const optionsList: EnumOptionsType<S>[] =
     enumOptions ||
-    (Array.isArray(schema.examples) ? schema.examples.map((example) => ({ value: example, label: example })) : []);
+    (Array.isArray(schema.examples)
+      ? schema.examples.map((example) => ({ value: example, label: String(example) }))
```
jimmycallin, line 132: `String(example)` regresses the `schema.examples` fallback for non-string examples.

Previously `label` held the raw example and `getDisplayValue(label)` (still called in `renderOption`) unwrapped objects via `.name`, `.label`, or `JSON.stringify`. Now the object is stringified to `"[object Object]"` *before* `getDisplayValue` ever sees it, so for e.g.

```json
{ "type": "object", "examples": [{ "name": "Alpha" }, { "name": "Beta" }] }
```

with no `enum`, the dropdown used to show `Alpha` / `Beta` and now shows `[object Object]` twice.

If the cast is only here to satisfy `EnumOptionsType['label']: string`, `getDisplayValue(example)` would keep the old rendering and still produce a string.

packages/utils/src/groupEnumOptions.ts
```diff
@@ -0,0 +1,98 @@
...
+  optgroups?: Record<string, EnumValue[]>,
+  enumDisabled?: EnumValue[],
+): GroupedEnumOptionsType<S>[] {
+  if (!Array.isArray(enumOptions)) {
+    return [];
+  }
+  const disabledValues: unknown[] = Array.isArray(enumDisabled) ? enumDisabled : [];
+  const disabledStrings = new Set(disabledValues.filter(isPrimitive).map(String));
+  const isDisabled = (value: unknown) =>
+    isPrimitive(value) ? disabledStrings.has(String(value)) : disabledValues.includes(value);
```
jimmycallin, line 49: This widens `ui:enumDisabled` for every theme, not just the new `optgroups` path: the old code was `enumDisabled.includes(value)` (strict), and now any primitive matches by its string form.

For a mixed-type enum this changes existing behaviour silently:

```ts
schema:   { enum: [1, '1'] }
uiSchema: { 'ui:enumDisabled': [1] }
```

previously disabled only the numeric `1`; now both options are disabled (and symmetrically for `[true]` vs `'true'`, `[0]` vs `'0'`).

The string-form fallback is documented for `optgroups` matching, but the `ui:enumDisabled` change isn't mentioned in `uiSchema.md` or the v7 upgrade guide. Either scope the coercion to group matching only, or call the `enumDisabled` change out in the upgrade guide.

packages/shadcn/src/components/ui/fancy-multi-select.tsx
```diff
@@ -24,6 +24,17 @@ export interface FancySelectItem {
   disabled?: boolean;
 }
 
+/**
+ * Represents a labeled section (an `<optgroup>` equivalent) of items in the dropdown. A section without a `label`
+ * renders as a plain, unheaded group, used for options that aren't part of any `ui:options.optgroups` group.
+ */
+export interface FancySelectSection {
```
jimmycallin, line 31: `FancySelectSection` is now declared identically in both `fancy-select.tsx` and `fancy-multi-select.tsx`, and `SelectWidget.tsx` imports only the `fancy-select.tsx` copy while passing the value to *both* components. It typechecks today purely by structural compatibility, so adding a required field to one copy (or renaming `items`) breaks the other silently at the call site.

One declaration (exported from `fancy-select.tsx` and imported by `fancy-multi-select.tsx`) would keep them in sync. Same applies to the pre-existing duplicated `FancySelectItem`.

packages/utils/src/groupEnumOptions.ts
heath-freenome: Good catch. In 1f90127f1 I documented the rule instead of changing the shape: the `groupEnumOptions` JSDoc and the `optgroups` section of `uiSchema.md` now say groups follow the object's property order, that integer-like labels always come first in ascending numeric order, and that a non-integer label (e.g. `'Year 2024'`) keeps the written order. Your example is in the docs, and a test pins the ordering.

I kept the object form because it matches the shape proposed in #580/#1813 and reads naturally in JSON. Since `optgroups` hasn't shipped yet, we could still switch to an ordered array form (`[{ label, values }]`) if you think numeric labels will be common enough to justify the extra verbosity. Happy to do that if you feel strongly.

packages/chakra-ui/src/NativeSelectWidget/NativeSelectWidget.tsx
heath-freenome: Fixed both in 1f90127f1:
- chakra `NativeSelectWidget` sets `disabled={option.disabled}` again (this was a regression on this branch), with a test covering grouped and ungrouped disabled options.
- daisyUI now destructures `enumDisabled` and passes it to `groupEnumOptions()`. Disabled rows get `aria-disabled` and `tabIndex={-1}`, and clicks and Enter/Space on them are ignored. `main` never supported `enumDisabled` here, so this is a new behavior with its own changelog entry; the three daisyUI `enumDisabled` snapshots changed only on the disabled rows.

packages/daisyui/src/widgets/SelectWidget/SelectWidget.tsx
heath-freenome: Fixed in 1f90127f1: the `examples` fallback now labels options with `getDisplayValue(example)`, so `{ name: 'Alpha' }` shows "Alpha" again, in both the list and the button text.

While fixing it I also changed `getDisplayValue` to blank only `null`/`undefined` instead of every falsy value, so examples like `0` and `false` render as `0`/`false` rather than an empty row (that was already blank in the list on `main`). Tests cover both object and falsy primitive examples.

packages/utils/src/groupEnumOptions.ts
heath-freenome: Agreed, that change wasn't intended. In 1f90127f1 `enumDisabled` is back to the strict `includes` check, the same as every other enum widget, so `enum: [1, '1']` with `ui:enumDisabled: [1]` disables only the number again. The string-form fallback now applies to `optgroups` matching only. I replaced the string-form `enumDisabled` tests with a mixed-type test and a by-reference test, and the JSDoc now states that `enumDisabled` matches strictly. With no behavior change left, the upgrade guide didn't need an entry.

packages/shadcn/src/components/ui/fancy-multi-select.tsx
heath-freenome: Done in 1f90127f1: `FancySelectItem` and `FancySelectSection` are now declared only in `fancy-select.tsx`, and `fancy-multi-select.tsx` imports both as types. Neither is exported from the package, so this doesn't change the public API.

packages/utils/src/groupEnumOptions.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const claimedIndices = new Set<number>();
+  const findUnclaimed = (candidates: IndexedEnumOptionType<S>[] | undefined) =>
+    candidates?.find((candidate) => !claimedIndices.has(candidate.index));
+  const findOption = (value: unknown) =>
+    findUnclaimed(byValue.get(value)) ?? (isPrimitive(value) ? findUnclaimed(byString.get(String(value))) : undefined);
+
+  const groups: EnumOptionsGroupType<S>[] = Object.entries(optgroups)
+    .map(([label, values]) => {
+      const options = values.flatMap((value) => {
```
jimmycallin: `values.flatMap(...)` assumes every `optgroups` entry is an array, but nothing checks it — unlike `enumOptions` (`Array.isArray`), `optgroups` itself (`typeof !== 'object'`) and `enumDisabled` (`Array.isArray`), which are all guarded a few lines up.

A hand-authored uiSchema that writes a single value instead of a list, e.g.

```js
'ui:options': { optgroups: { 'Group A': 'foo' } }
```

throws `TypeError: values.flatMap is not a function` during render, which takes down the whole form rather than just ignoring the malformed group. Same for `null`/a number as the value. Confirmed by running it against this branch.

```suggestion
  const groups: EnumOptionsGroupType<S>[] = Object.entries(optgroups)
    .map(([label, values]) => {
      const options = (Array.isArray(values) ? values : []).flatMap((value) => {
```

packages/mui/src/SelectWidget/SelectWidget.tsx
```diff
@@ -115,19 +137,16 @@ export default function SelectWidget<T = any, S extends StrictRJSFSchema = RJSFS
...
-                disabled={isDisabled}
-              >
-                {enumLabel}
-              </MenuItem>
-            );
-          })}
+        {groupEnumOptions<S>(enumOptions, optgroups, enumDisabled).flatMap((item) =>
+          isEnumOptionsGroup<S>(item)
+            ? [
+                <ListSubheader key={`optgroup-${item.label}`} aria-hidden>
```
jimmycallin: `aria-hidden` removes the group label from the accessibility tree entirely, so a screen-reader user hears the options but never the heading they are grouped under — the opposite of what a native `<optgroup label>` does (and of what every other theme here does: `<optgroup>` in core/react-bootstrap/chakra-native, `OptionGroup` in fluentui, `Select.ItemGroupLabel` in chakra, `aria-label` on `role="group"` in daisyui). Concretely, with `optgroups: { Latin: [...], Greek: [...] }` and an enum where the same label appears in two groups, an AT user cannot tell the two apart.

MUI's own grouped-`Select` guidance uses a bare `ListSubheader`; it is worth dropping the `aria-hidden` (MUI clones it with `role="option"` either way, but at least the text is announced) or wrapping the group so the label is exposed rather than hidden.

packages/utils/src/types.ts
```diff
@@ -1242,6 +1267,10 @@ type UIOptionsBaseType<T = any, S extends StrictRJSFSchema = RJSFSchema, F exten
      * Supports a `'*'` wildcard to represent all remaining values in their original schema order.
      */
     enumOrder?: EnumValue[];
+    /** Groups enum options into labeled `<optgroup>`-like sections. Keys are group labels, values are arrays of enum
+     * values belonging to that group. Enum values not listed in any group are rendered ungrouped after the groups.
+     */
+    optgroups?: Record<string, EnumValue[]>;
```
jimmycallin, line 1275: `optgroups` was added to `UIOptionsBaseType` but not to `CoreUiOptionsChecks`, so the new option is unusable for anyone who opted into the type-safe `UiSchema` this release also introduces. With

```ts
const ui: UiSchema<{ pet: string }, RJSFSchema, any, CoreUiOptionsChecks> = {
  pet: { 'ui:options': { optgroups: { A: ['a'] } } },
};
```

tsc reports `TS2353: Object literal may only specify known properties, and 'optgroups' does not exist in type ...` (verified against this branch), while the sibling `enumOrder` on the same line compiles. `optionValueFormat` escapes this only because it lives on `GlobalUISchemaOptions`, which `CommonUiOptions` folds in; `optgroups` has no such path.

It needs adding next to `enumOrder` in all four enum-capable branches (string, number, boolean, array) — this file's own doc comment says it covers "the options that already exist on `UIOptionsBaseType`".

packages/utils/src/groupEnumOptions.ts
heath-freenome: Good catch, fixed in 23d2f5054 with your suggestion. A group whose value isn't an array now contributes no options, so the existing empty-group filter drops it, the same as a group whose values match nothing — malformed input is ignored instead of taking down the render. The JSDoc says so, and a test covers `{ 'Group A': 'foo', 'Group B': null, 'Group C': 3 }`.

packages/mui/src/SelectWidget/SelectWidget.tsx
heath-freenome: Fair, and I agree the hidden heading is the worse of the two. I went one step past dropping it, in 23d2f5054: the `ListSubheader` is now `aria-disabled` instead of `aria-hidden`, so the label is announced but as unavailable rather than as a choice. Clicking it was already a no-op — `handleItemClick` returns early on the missing `tabindex` (`SelectInput.js:452` in the installed 9.1.2) — so `aria-disabled` describes what actually happens.

I looked for a way to expose a real group and couldn't find one: the `cloneElement(child, { …, role: 'option' })` map is unconditional over direct children, and wrapping a group in a container would take its `MenuItem`s out of that map, which is what makes them selectable and value-matched in the first place. So it's "announced as a disabled option" or "not announced", and I took the former.

Tests now assert the heading is `aria-disabled`, is exposed as an option, and fires no `onChange` when clicked. I also added a note to the `optgroups` section of `uiSchema.md` calling out that mui's semantics differ from the other themes here, since it's a user-visible difference rather than an internal detail.

Happy to drop to a bare `ListSubheader` (what MUI's own grouping example uses) if you'd rather not have the heading read as disabled.

packages/utils/src/types.ts
heath-freenome: You're right — the type-safe `UiSchema` and `optgroups` land in the same release, so the new option was unusable to anyone who opted in. Added `optgroups?: Record<string, EnumValue[]>` next to `enumOrder` in all four enum-capable branches (string, number, boolean, array) in 23d2f5054, plus a test that it compiles on a field of each of those types and is rejected on an object field.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
```diff
@@ -199,6 +199,10 @@ This theme has been removed as it does not support React 19 and has become unmai
 Version 7 is dropping official support for `antd` version 5. You must upgrade to `antd` 6.3.6 or later.
 The `6.3.6` floor isn't arbitrary: `antd` 6.0.0-6.3.5 crash `@rjsf/antd` forms on first render, fixed upstream in [ant-design/ant-design#57636](https://github.com/ant-design/ant-design/pull/57636)
 
+#### @rjsf/chakra-ui
```
jimmycallin: Small placement nit: this section sits under **`### Theme version support changes`**, but it isn't one — it's a peer dependency *removal*. Its three siblings (`#### @rjsf/antd`, `#### @rjsf/mui`, `#### @rjsf/mantine`) are all "dropped support for vN, the range is now ^M" bumps, so a reader scanning that section for "which theme versions do I now need" gets one entry that answers a different question.

The alphabetical slot is right and it's clearly under `## Breaking changes` either way, so this is cosmetic. Renaming the parent to something like "Theme dependency changes", or lifting this to its own `### @rjsf/chakra-ui no longer requires chakra-react-select BREAKING CHANGE`, would both read truer. Your call — happy to leave it.

jimmycallin (approved):
Approving. The shared `groupEnumOptions()`/`isEnumOptionsGroup()`/`flattenGroupedOptions()` split is the right shape for this — the grouping and partitioning logic lives once in `@rjsf/utils` and each theme only supplies its own rendering primitive, which is what keeps nine `SelectWidget`s from drifting. Thanks for chasing down the accessibility side across MUI and daisyUI rather than settling for the visual grouping, and for the indexed lookups and `enumOrder`-matching semantics.

The diff is large but it reads honestly: of ~7.2k added lines, roughly 6k are regenerated theme snapshots and ~1.4k are new per-theme tests, leaving about 1k of actual source.

Breaking-change documentation checks out. The `chakra-react-select` peer dependency removal is flagged **BREAKING CHANGE** in `CHANGELOG-v7.md`, has its own block in the v7 upgrade guide, and the README prerequisites and install line are both updated. `optgroups` is documented in `uiSchema.md`, added to `CoreUiOptionsChecks`, and the `../api-reference/uiSchema.md#optgroups` anchor in the guide resolves.

One cosmetic comment left on the upgrade guide section placement — non-blocking. Good to merge once that's resolved to your satisfaction.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
heath-freenome: Agreed, and I took the second option: in 95e35298a the peer dependency removal is now its own `### `@rjsf/chakra-ui` no longer requires `chakra-react-select` BREAKING CHANGE` section, sitting between "Theme version support changes" and "Node support", so the three entries left under that heading are all version bumps and answer the one question a reader scans it for.

I preferred that to renaming the parent because the removal is a real BREAKING CHANGE and now says so in its own heading, the way the other breaking entries in that section do. While moving it I also expanded the body to say explicitly that no code change is required, and what to do if your own code uses `chakra-react-select`. Nothing linked to the old `#### @rjsf/chakra-ui` anchor, so no cross-references needed updating.

heath-freenome:
Thanks for the review, and for the approval. The upgrade guide nit is [addressed](https://github.com/rjsf-team/react-jsonschema-form/pull/5301#discussion_r4067308220) — the `chakra-react-select` removal is its own section now.

Two commits landed after your approval, both fixing the same class of bug you were circling in the ordering discussion, so flagging them rather than letting them slip in post-approval:

- **95e35298a** — shadcn's multi-select was reporting its values in *display* order. `items` was built from the grouped options, and `groupEnumOptions()` returns grouped ahead of ungrouped, so with enum `['a','b','c','d']` and `optgroups: { G: ['c','d'] }`, picking a, c, b produced `['c','a','b']` where the same picks without `optgroups` produced `['a','c','b']`. `items` now keeps the enum order and grouping applies only to `sections`, which is what the popup actually renders.
- **55c883178** — the same bug in `@rjsf/core` and `@rjsf/react-bootstrap`, which is worse there: a native `<select multiple>` reports its selection in *document* order only, with no notion of pick order at all. So `optgroups` reordering the DOM reordered the submitted array outright — enum `['a','b','c','d']` with `optgroups: { Zed: ['d','c'] }`, picking a then c, submitted `['c','a']`. Both widgets now map DOM position back to enum index before the values reach form data. Each fix has a regression test that fails without it.

I audited the remaining themes while I was there: mui, daisyui, antd, mantine, chakra and fluentui-rc all track their own selection and never read DOM order, so core, react-bootstrap and shadcn were the only three affected. `uiSchema.md` now states the guarantee and scopes it honestly — grouping never changes the submitted order, but the underlying order is the theme's own: the native `<select>` themes report enum order, while the custom-dropdown themes order it their own way.

Also tightened `isEnumOptionsGroup()` in the same commit: it keyed only on `Array.isArray(item.options)`, so a caller-supplied `enumOptions` entry carrying an array-valued `options` (react-select's own group shape) would silently render as an empty group in place of the option. It now also requires the absence of the numeric `index` tag that `groupEnumOptions()` puts on every standalone option and never on a group.

PR #5304, opened 2026-09-18
Fix #5294: apply schema changes that add a defaulted property after onChange

### LINKED ISSUE #5294: Schema update after `onChange` is dropped when the new schema adds a property with a default
### Prerequisites

### What theme are you using?

core

### Version

6.10.1

### Current Behavior

When the parent component derives the `schema` from the current `formData`, the schema update that follows an `onChange` is silently dropped. The new field never renders, and there is no error or warning.

This only happens when the property that the new schema adds has a `default`. Remove the `default` and the same update is applied correctly. Any field type is affected.

### Expected Behavior

A `schema` prop change should always be applied, whether or not the new schema adds a property with a `default`.

### Steps To Reproduce

Open the [sandbox](https://codesandbox.io/p/sandbox/2g44jl), or run this component:

```jsx
import { useState } from 'react';
import Form from '@rjsf/core';

const getSchema = (formData) => ({
  type: 'object',
  properties: {
    trigger: { type: 'string', title: 'Trigger' },
    ...(formData?.trigger === 'show'
      ? {
          extra: {
            type: 'string',
            title: 'Extra',
            default: 'preset', // remove this line and the bug goes away
          },
        }
      : {}),
  },
});

export default function App() {
  const [formData, setFormData] = useState({});

  return (
    <Form
      schema={getSchema(formData)}
      validator={validator}
      formData={formData}
      onChange={(event) => setFormData(event.formData)}
    />
  );
}
```

1. Type `show` into the `Trigger` field.
2. Expected: a second field, `Extra`, appears, pre-filled with `preset`.
3. Actual: nothing appears. Only `Trigger` is rendered.
4. Delete the `default: 'preset'` line and repeat. `Extra` now appears as expected.

`liveValidate` makes no difference; I tested with and without it.

### Environment

```markdown
- OS: macOS 27.0
- Node: 24.18.1
- npm: 11.16.0
- React: 19.2.8
```

### Anything else?

The cause looks like a flag that is set but not always cleared. `processPendingChange()` sets `isProcessingUserChange = true`. The `setState` it makes commits with unchanged props, so `getSnapshotBeforeUpdate` returns `{ shouldUpdate: false }` and `componentDidUpdate` never reaches the line that clears the flag, because that line sits inside the `if (snapshot.shouldUpdate)` block. The parent's reply to `onChange` then arrives with the stale flag still set. The `default` in the new schema is what makes `nextState.formData` differ from `props.formData`, so the guard fires and the computed state is thrown away

heath-freenome:

Fixes #5294

`componentDidUpdate` dropped the recomputed state whenever the `isProcessingUserChange` flag was set and the result differed from `props.formData`. The flag spans the round trip to the parent so a oneOf/anyOf switch to a `null` option isn't reverted by the root `default` (#4921), but the guard ignored what actually changed:

- A controlled parent returning the same `formData` with a new schema had that schema discarded whenever the schema added a property with a `default`, since the default made the recomputed state differ.
- An uncontrolled form never gets that echo, so the flag stayed set after the first edit and any later schema change adding a default was silently dropped.
- Whenever the guard fired, it threw away the whole of `nextState`, so `uiSchema`, `registry` (templates, widgets, `formContext`) and a new `schemaUtils` from the same update were lost too.

This replaces the instance flag with an `isEchoOfState` check computed in `getSnapshotBeforeUpdate`. The guard applies only when all of these hold:

- The form is re-deriving its own data. That covers an uncontrolled form, which `getStateFromProps` re-derives from its own state, and a parent handing back what `onChange` emitted.
- `schema` and `experimental_defaultFormStateBehavior` are unchanged. `validator` and `experimental_customMergeAllOf` are left out because they're compared by reference, and parents commonly create them inline.

The parent's reply is recognized by value: the form remembers the `formData` it hands to `onChange` and compares what the parent passes back against it. Parents routinely respell an absent value on the round trip, by spreading the emitted `undefined` into `{}`, coalescing it to `null` or round-tripping through JSON, so the comparison ignores how an absent value is spelled. It does that only until the reply arrives, though. Once it has, the remembered report is replaced by the spelling the parent chose, so the reply keeps counting as the form's own data however it was reshaped, while anything the parent passes afterwards is data the form never reported and is applied. Otherwise a parent dropping a key so its `default` comes back, or emptying the data, would look like the reply all over again.

Every `onChange` call site reports through a single `reportChange` method, so a parent's reply to a reset, a blur or the defaults derived at construction is recognized the same way as a reply to an edit.

When the guard applies, the rest of `nextState` is still applied. Only `formData` and `retrievedSchema` are kept. They come from `this.state` rather than `prevState`, so a state update batched into the same render, like `setFieldValue()`, isn't undone. When `nextState` may have live-validated the re-derived data (`liveValidate` on, `noValidate` off), the kept data's errors are kept too.

**Known limitation:** a switch to a `null` option at the root of a schema with a root `default` still reverts if the parent changes `schema` in the same reply. The switch emits `undefined`, which `getDefaultFormState` fills with the default. A proper fix is tracked in #5307: `MultiSchemaField` should emit `null`, and `getDefaultFormState` should treat an explicit `null` as a value.

### Checklist


packages/core/src/components/Form.tsx
```diff
@@ -507,7 +504,10 @@ export default class Form<
         !isStateDataChanged,
       );
       const shouldUpdate = !deepEquals(nextState, prevState);
-      return { nextState, shouldUpdate };
+      // The parent handing back the `formData` the form just emitted through `onChange`, with nothing that feeds the
+      // defaults changed; any change to the schema, validator or defaults behavior rebuilds `schemaUtils`
+      const isEchoOfState = !isStateDataChanged && nextState.schemaUtils === this.state.schemaUtils;
```
jimmycallin, line 509: `nextState.schemaUtils === this.state.schemaUtils` is a weaker proxy than "nothing that feeds the defaults changed": `doesSchemaUtilsDiffer` compares `validator` and `experimental_customMergeAllOf` **by reference**, not with `deepEquals`. So a parent that creates either inline in JSX — `validator={customizeValidator()}` or `experimental_customMergeAllOf={(s) => ...}`, both common — rebuilds `schemaUtils` on *every* update, `isEchoOfState` is then always `false`, and the #4921 guard is gone: the oneOf/anyOf switch to the `null` option is reverted by the root `default` on the parent's echo.

Verified against this branch with the PR's own oneOf fixture: the same test passes on `main` (flag-based guard) and fails here.

```tsx
const { node, rerender } = createFormComponent(
  { schema, formData: { types: 'advanced', content: 'placeholder' },
    experimental_defaultFormStateBehavior: behavior,
    onChange: (e) => { current = e.formData; } },
  customizeValidator(),
);
await user.selectOptions(node.querySelector('#root__oneof_select')!, '1');
rerender({ schema, formData: current, experimental_defaultFormStateBehavior: behavior }, customizeValidator());
// PR: value is '0' (reverted). main: '1'
expect(node.querySelector('#root__oneof_select')).toHaveValue('1');
```

The same probe with a fresh `experimental_customMergeAllOf` identity on rerender fails identically.

Consider deriving the flag from what actually feeds the defaults instead of from `schemaUtils` identity — e.g. `isSchemaChanged` plus `deepEquals` on `experimental_defaultFormStateBehavior`, and treating a mere identity change of `validator`/`customMergeAllOf` as "unchanged".

(Related, and inherent to this approach: when the parent legitimately changes the schema in the same round trip that echoes the change — the dynamic-schema pattern from the new test #1 — the guard is also disabled, so a oneOf-to-`null` switch in such a form regresses to #4921. Worth calling out in the changelog if it is the accepted tradeoff.)

heath-freenome: Thanks, good catch on both points. I was able to reproduce each of them. Fixed in ebbbf6754.

**1. `validator` / `customMergeAllOf` being recreated.** You're right, `schemaUtils` identity was the wrong thing to key on. The echo check now uses what actually feeds the defaults:

```ts
const isEchoOfState =
  !isStateDataChanged &&
  !isSchemaChanged &&
  deepEquals(prevProps.experimental_defaultFormStateBehavior, this.props.experimental_defaultFormStateBehavior);
```

A new `validator` or `experimental_customMergeAllOf` reference no longer switches off the guard. I added your probe as a test in `Form.handlers.test.tsx`, plus a variant with a recreated `customMergeAllOf`. Both fail on the previous commit and pass now. The downside: if a parent really swaps the validator in the same render that echoes the form's data, the form keeps its state until the next change. `main` already behaves that way, because it drops the whole state in that case.

**2. Schema change in the same round trip.** I couldn't find a way to fix this inside `Form`:
- Picking the root `null` option makes the form emit `formData` as `undefined`, not `null`.
- With a root `default`, rjsf reads `undefined` as "use the default", and a schema change has to recompute defaults.
- So from the data alone, the switch looks exactly like a hole the defaults should fill. I tried a guard that stops defaults from overwriting values the form emitted, but `undefined` never counts as an overwrite.

`main` only kept the switch in this case because it also dropped the schema change, which is #5294 itself. So I took your suggestion and noted it in the changelog as a known limitation.

The real fix probably belongs in a follow-up: have `MultiSchemaField` emit `null` for a `type: 'null'` option, and have `getDefaultFormState` treat an explicit `null` as a value rather than as empty data. That changes behavior in `@rjsf/utils`, so I'd rather not ship it in this PR. I'm happy to open an issue for it if you agree.

packages/core/src/components/Form.tsx
```diff
@@ -507,7 +504,14 @@ export default class Form<
         !isStateDataChanged,
       );
       const shouldUpdate = !deepEquals(nextState, prevState);
-      return { nextState, shouldUpdate };
+      // The parent handing back the `formData` the form just emitted through `onChange`, with nothing that feeds the
+      // defaults changed. `validator` and `experimental_customMergeAllOf` are left out because they are compared by
+      // reference, and parents commonly create them inline on every render
+      const isEchoOfState =
+        !isStateDataChanged &&
```
jimmycallin: `isEchoOfState` can never be true for an uncontrolled form that holds any data: `isStateDataChanged` compares `this.props.formData` (always `undefined` when uncontrolled) against `this.state.formData`, so as soon as the form holds an object the guard is permanently off, and `nextStateDiffersFromProps` (`nextState.formData` vs `undefined`) is permanently true.

That is fine for the #5294 fix, but it also removes the #4921 protection from every uncontrolled form with data — a protection the old `isProcessingUserChange` flag did give them (uncontrolled forms get no prop echo, so the flag was never cleared and every later prop update was skipped).

Concrete scenario: uncontrolled form, `{ type: 'object', properties: { other: {...}, cfg: { oneOf: [ {...}, { type: 'null' } ], default: {...} } } }`. The user switches `cfg` to the null option, so `state.formData` is `{ other: 'x', cfg: null }`. The parent then changes an unrelated prop (`uiSchema`, `disabled`, …): the guard cannot fire, `getStateFromProps` re-derives defaults from `state.formData`, and `cfg` is replaced by its `default` again.

The two new uncontrolled tests dodge this because the root-level null switch leaves `state.formData` as `undefined`, which happens to deep-equal the undefined `props.formData` and re-enables the guard. Worth adding a nested variant with sibling data to see whether this really reverts.

packages/core/src/components/Form.tsx
```diff
@@ -507,7 +504,14 @@ export default class Form<
...
+        !isSchemaChanged &&
+        deepEquals(prevProps.experimental_defaultFormStateBehavior, this.props.experimental_defaultFormStateBehavior);
```
jimmycallin, line 598: The echo check requires the parent to hand back exactly what the form emitted. Any parent that normalizes the value on the round trip breaks it: `null` -> `undefined` (or vice versa), a JSON/structuredClone round trip that drops `undefined`-valued keys, a redux/immer store that fills in missing keys. In those cases `isStateDataChanged` is true, the guard is off, and the very case the guard exists for — the switch to a `null` oneOf/anyOf option under a root `default` — reverts on the parent's reply. Note the existing #4921 test relies on the emitted value being the identical `undefined`/`null`, so this is not covered.

packages/core/src/components/Form.tsx
```diff
@@ -524,22 +528,14 @@ export default class Form<
...
       const nextStateDiffersFromProps = !deepEquals(nextState.formData, this.props.formData);
-      const wasProcessingUserChange = this.isProcessingUserChange;
-      this.isProcessingUserChange = false;
 
-      if (wasProcessingUserChange && nextStateDiffersFromProps) {
-        // Skip - the user's option switch is already applied via processPendingChange
+      if (isEchoOfState && nextStateDiffersFromProps) {
+        // Re-deriving defaults from echoed data can undo what `processPendingChange` resolved, such as a switch to a
+        // oneOf/anyOf option whose data the root `default` would otherwise replace, so keep the state as it is
         return;
```
jimmycallin, line 650: The early `return` drops the whole of `nextState`, not just its `formData` — `schemaUtils`, `retrievedSchema`, `errors`/`errorSchema` are all discarded. With the old instance flag that was a one-shot right after a user change; the new condition is a property of the current data, so once the form is in the state the guard protects (data that re-deriving defaults would change, e.g. the `null` oneOf option under a root `default`), *every* subsequent prop update takes this branch for as long as the data stays that way.

Concrete scenario: parent swaps `validator` (different custom formats / ajv options) or `experimental_customMergeAllOf` while echoing the same `formData`. `getStateFromProps` builds fresh `schemaUtils`, the guard throws it away, and the form keeps validating and retrieving schemas through the old validator indefinitely. Consider keeping the state update and only preserving `formData`, rather than skipping the update entirely.

packages/core/src/components/Form.tsx
```diff
@@ -524,22 +528,14 @@ export default class Form<
...
       }
```
jimmycallin, line 651: **Once this guard engages it drops every later prop update, not just one render.**

`isEchoOfState` is a steady state, not a one-shot flag. For a controlled form parked where re-deriving defaults is not idempotent — the #4921 case: root `oneOf` with a root `default`, sitting on the `null` option — `props.formData` keeps deep-equalling `state.formData` and the schema keeps being unchanged, so *every* subsequent prop change takes this `return`. And because the early return skips `this.setState(nextState)` wholesale, everything else `getStateFromProps` just recomputed is thrown away with the formData: `uiSchema`, `registry` (hence `templates`, `widgets`, `fields`, `formContext`, `globalUiOptions`), `schemaUtils` and `retrievedSchema`. The old `isProcessingUserChange` was cleared on the first `componentDidUpdate`, so at most one render was ever skipped.

Verified on this branch (`packages/core`, jsdom): controlled form on that schema, user selects the `null` option, parent echoes `formData`, then the parent passes a custom `MultiSchemaFieldTemplate`, then a new `formContext`.

- base `main`: `select=1 custom=false | select=0 custom=false | select=0 custom=true` — the option wrongly reverts (the bug being fixed here) but the template does land.
- this branch: `select=1 custom=false | select=1 custom=false | select=1 custom=false` — the option is kept, but the custom template (and `formContext`, `uiSchema`, a swapped `validator`'s `schemaUtils`, …) never arrives. It only recovers if `formData` or `schema` later changes.

Keeping just the resolved `formData`/`retrievedSchema` and applying the rest of `nextState` fixes both: in the same probe the option stays on `1` *and* the template lands. All 2011 `packages/core` tests pass with the change below.

```suggestion
      if (isEchoOfState && nextStateDiffersFromProps) {
        // Re-deriving defaults from echoed data can undo what `processPendingChange` resolved, such as a switch to a
        // oneOf/anyOf option whose data the root `default` would otherwise replace, so that data is kept while the
        // rest of the recomputed state (`uiSchema`, `registry`, `schemaUtils`) is still applied
        // oxlint-disable-next-line react/no-did-update-set-state -- guarded to prevent infinite loop
        this.setState({ ...nextState, formData: prevState.formData, retrievedSchema: prevState.retrievedSchema });
        return;
      }
```

packages/core/test/Form.handlers.test.tsx
```diff
@@ -466,6 +467,141 @@ describeRepeated('Form common: event handlers', (createFormComponent) => {
...
+        default: { types: 'advanced', content: 'placeholder' },
+      };
+      const experimental_defaultFormStateBehavior = { emptyObjectFields: 'populateAllDefaults' } as const;
+      const { node, rerender } = createFormComponent({ schema, experimental_defaultFormStateBehavior });
+
+      await user.selectOptions(node.querySelector<HTMLSelectElement>('#root__oneof_select')!, '1');
+      rerender({ schema, experimental_defaultFormStateBehavior, uiSchema: { 'ui:title': 'Changed' } });
+
+      expect(node.querySelector<HTMLSelectElement>('#root__oneof_select')).toHaveValue('1');
+      expect(node.querySelector('#root_content')).not.toBeInTheDocument();
```
jimmycallin: The "unrelated prop change" this test rerenders with is itself swallowed by the new guard, so the test cannot show what its name claims. `getSnapshotBeforeUpdate` sees `props.formData` (`undefined`) equal to `state.formData` and an unchanged schema, so `isEchoOfState` is true and `componentDidUpdate` returns before `setState`, which means the new `uiSchema` never reaches `state.uiSchema` — `ui:title` simply isn't rendered on the `null` branch, so neither assertion notices.

An observable unrelated prop makes the gap visible: rerender with `templates: { MultiSchemaFieldTemplate: Custom }` (or a new `formContext`) and assert the custom template renders — on this branch it never does, on `main` it does one render later. See the companion comment on `Form.tsx`.

packages/core/src/components/Form.tsx
heath-freenome: Thanks, all five were real. Fixed in f4d6de02b. I probed each one against this branch and `main` first. Here's what I found and what changed.

**The guard dropping the rest of `nextState` (this comment and r4054131244).** Confirmed: the custom `MultiSchemaFieldTemplate` never arrived. I took your suggestion with one change: the kept data comes from `this.state`, not `prevState`.

`prevState` is the state from before this commit. When a parent calls `setFieldValue()` or `reset()` and changes another prop in the same batch, the guard wrote the old value back. This was easy to hit once uncontrolled forms take the guard (below). I added a test for it.

A second problem was the errors. For the kept data to have matching errors, the guard also keeps `this.state`'s error fields, but only when `liveValidate` is on and `noValidate` is off. That's the only case where `nextState` could have validated the re-derived data. Keeping them always broke turning on `noValidate` on an uncontrolled form (the errors never cleared), and there's a test for that too.

```ts
const { formData, retrievedSchema, errors, errorSchema, schemaValidationErrors, schemaValidationErrorSchema } =
  this.state;
const mayHaveLiveValidated = this.props.liveValidate && !this.props.noValidate;
this.setState({
  ...nextState,
  formData,
  retrievedSchema,
  ...(mayHaveLiveValidated && { errors, errorSchema, schemaValidationErrors, schemaValidationErrorSchema }),
});
```

One thing about your probe: `select=1 custom=false` stays that way after the template-only rerender and flips to `custom=true` on the `formContext` one. `deepEquals` ignores functions, so a props change that only swaps a template doesn't count as a change. `main` has the same one-render delay. That's out of scope here.

**Uncontrolled forms (r4054131075).** Your exact scenario doesn't revert. When the user picks the `null` option, `cfg` is stored as `undefined`, and the defaults merge leaves an existing `undefined` key alone. It does revert with `omitExtraData` + `liveOmit`, because the key gets stripped and the default fills it back in.

Since `getStateFromProps` re-derives an uncontrolled form from its own state anyway, `props.formData === undefined` now counts as the form's own data. The nested-with-sibling-data test you suggested is added. It fails on the previous commit only in the `omitExtraData`/`liveOmit` variant, and on `main` in all four, because `main` also drops the `uiSchema` change.

**A parent changing the value on the round trip (r4054131172).** I probed `undefined` → `null` and a JSON round trip, each at the root and nested. Only one case reverts: the parent storing the root `undefined` as `null`. Nested cases are fine, and a JSON round trip leaves the root `undefined` as is.

`props.formData === null` while `state.formData === undefined` now also counts as the form's own data. There are tests for it, with and without `liveValidate`. Other normalizations can still turn the guard off, but I couldn't find one that reverts a switch. The real fix is still the follow-up where the switch emits `null`.

**The test that couldn't fail (r4054177687).** Agreed. It's replaced by a group of four tests (uncontrolled, controlled, `null`-normalized, nested with sibling data) that rerender with `ui:disabled` and assert the select is on option `1` and is disabled. 13 of its 16 runs fail on the previous commit and 12 fail on `main`.

The core suite passes (2035 tests), and lint, `cs-check` and typecheck are clean. The changelog entry now says the guard keeps only `formData`.

packages/core/src/components/Form.tsx
heath-freenome: Fixed in f4d6de02b. It only reverted with `omitExtraData` + `liveOmit`. Details in the reply on `componentDidUpdate`.

packages/core/src/components/Form.tsx
heath-freenome: Fixed in f4d6de02b for the case that reverts: the root `undefined` stored as `null`. Details in the reply on `componentDidUpdate`.

packages/core/src/components/Form.tsx
heath-freenome: Fixed in f4d6de02b. Only `formData`/`retrievedSchema` are kept, taken from `this.state`. Details in the reply on `componentDidUpdate`.

packages/core/test/Form.handlers.test.tsx
heath-freenome: Replaced in f4d6de02b with `ui:disabled`-based tests, which fail on the previous commit. Details in the reply on `componentDidUpdate`.

packages/core/src/components/Form.tsx
```diff
@@ -507,7 +504,20 @@ export default class Form<
         !isStateDataChanged,
       );
       const shouldUpdate = !deepEquals(nextState, prevState);
-      return { nextState, shouldUpdate };
+      // `getStateFromProps` re-derives an uncontrolled form from its own state, and a parent may store the `undefined`
+      // that a switch to a root `null` option emits as `null`, so both count as handing back the form's own data
+      const isOwnFormData =
+        !isStateDataChanged ||
+        this.props.formData === undefined ||
+        (this.props.formData === null && this.state.formData === undefined);
```
jimmycallin: `isOwnFormData` only recognises the reply as the form's own data when it deep-equals `this.state.formData`, is `undefined`, or is the `null`-for-`undefined` special case. The old `isProcessingUserChange` flag protected the switch for *any* reply that followed a user change, so parents that normalise the emitted value now lose the #4921 protection again.

Concrete, and verified against this branch vs. its merge base:

```jsx
// a very common parent shape — `{...undefined}` is `{}`
onChange={(e) => setData({ ...e.formData })}
```

With the root-`default` + `type: 'null'` oneOf schema from the tests: select "No Configuration", the parent echoes `{}`, so `isStateDataChanged` is true and `props.formData` is neither `undefined` nor `null` → `isEchoOfState` is false → `getDefaultFormState` refills the root `default` and the select snaps back to option `0`. On `main` it stays on `1`. Same result for a parent that serialises its reply (`JSON.parse(JSON.stringify(...))` drops the `cfg: undefined` key in the nested variant), which then also re-emits the reverted data through `onChange`.

Either the `null` special case wants to be a general "the reply is the emitted data modulo `undefined` normalisation" comparison, or the pending-change signal needs to stay as an additional way to enter the guard (scoped so it clears after the change is processed, which is what #5294 was really about).

jimmycallin (approved):
fix last issue, then you can merge

packages/core/src/components/Form.tsx
heath-freenome: Both spellings revert, confirmed. Fixed in ae9080348. I took the first option, the general comparison.

The form now remembers the `formData` it hands to `onChange` and compares the parent's reply against it with every spelling of an absent value normalized away: `null` becomes `undefined`, keys holding an absent value are dropped, and an object whose values are all absent becomes `undefined` too. So `{ ...e.formData }` giving `{}`, `e.formData ?? null`, and a JSON round trip that drops the `cfg: undefined` key all compare equal to the emitted `undefined`. The `null`-for-`undefined` special case is gone, subsumed by it.

I went with the comparison rather than keeping the pending-change signal because a boolean "a reply is outstanding" can't be matched to a specific reply. It has to be cleared on some heuristic, and the two obvious ones are each wrong in one direction: clear it on the next props change and a parent that defers its reply past an unrelated prop change loses the guard, don't clear it and a parent that ignores `onChange` keeps the guard forever, so its next genuine `formData` replacement gets eaten. I wrote both of those bugs before throwing the flag away. Comparing values has neither, because it answers the question the guard is actually asking.

Being value based also fixes the deferred reply, which neither this branch nor `main` handled: select "No Configuration", rerender with an unrelated prop, then let the parent reply. There's a test for it now, and it fails on both.

The comparison does need one piece of bookkeeping: the remembered report is dropped as soon as the parent supplies data of its own. Without that, a parent that ignores `onChange` and later happens to pass a value equal to something the form emitted earlier has it swallowed as an echo. That one is covered by `should apply a formData prop change matching data reported before the parent took over`, which passes on `main` and fails on the version of this commit without the clearing.

All five `onChange` call sites now report through a single `reportChange` method. The constructor, `componentDidUpdate`, `reset()` and `onBlur` handed data to the parent without recording it, so a reshaping parent replying to any of those lost the guard.

Core passes 2051 tests, the full monorepo suite is green, and lint, `cs-check` and typecheck are clean.

heath-freenome: Correction to the above: two of the things I said that comparison buys are only true of a narrower version of it, which is now in 795bb644c.

Normalizing every spelling of absence at every depth makes too much compare equal. Three prop changes that `main` applies were swallowed as echoes:

- a parent dropping a key so its `default` comes back, `{ a: undefined, b: 'y' }` replied to with `{ b: 'y' }`
- a parent emptying the data, `{ name: undefined }` replied to with `{}`
- a parent filling a subsection the form had reported as absent, `{ other: 'A' }` followed by `{ other: 'A', cfg: {} }`

Two changes. `null` and an empty object now only count as absent at the root, which is where a parent replying to an emitted `undefined` spreads it into an object or coalesces it to `null`. Nested, an empty object or a `null` is a value the parent is asking for. Keys holding `undefined` are still dropped at any depth, since a JSON round trip drops them anyway and they say nothing an absent key doesn't.

And the tolerance for respelling now lasts only until the reply arrives. Once it has, the remembered report is replaced by the spelling the parent chose, so the reply keeps counting as the form's own data however it was reshaped, while anything the parent passes after that is data the form never reported. That is what fixes the first two cases: the reply is recognized, and the key removal or the empty object that follows it is not.

So the rule is narrower than what I described: the reply may be respelled, but only once, and from then on only the parent's own spelling is the form's data. The deferred reply and the reshaping parents still work, including a JSON round tripping parent that keeps re-rendering after its reply, which the exact comparison alone would lose on the second render.

Tests: `re-applies the default of a key the parent removed after replying` and `re-applies the defaults when the parent empties the data after replying` in `Form.behaviors.test.tsx`, plus `should apply a formData prop change that fills a subsection the form reported as absent` in `Form.handlers.test.tsx`. The first two are outside `describeRepeated` deliberately: the report only carries the cleared key with an `undefined` value when `omitExtraData` is off, so two of the four variants can't express the scenario at all. All three fail on ae9080348 and pass on `main`.

2057 core tests, the full monorepo suite, lint, `cs-check` and typecheck are green.

Issue #4408, opened 2024-12-04
Changes to `extraErrors` prop are not reflected in `Form` state.


studiosciences:
### Prerequisites

- [X] I have searched the existing issues
- [X] I understand that providing a [SSCCE](http://sscce.org/) example is tremendously useful to the maintainers.
- [X] I have read the [documentation](https://rjsf-team.github.io/react-jsonschema-form/docs)
- [X] Ideally, I'm providing a [sample JSFiddle](https://jsfiddle.net), [Codesandbox.io](https://codesandbox.io) or preferably a [shared playground link](https://rjsf-team.github.io/react-jsonschema-form/) demonstrating the issue.

### What theme are you using?

core

### Version

5.x

### Current Behavior

When liveValidation is enabled, the `extraErrors` prop can only add errors, not clear or change them. It only clears previous `extraErrors` from state `onSubmit()`.

### Expected Behavior

Changing `extraErrors` prop immediately updates the UI.

This is necessary for cases such as:
- live validation of unique key requirements, such as username availability.
- RJSF is a part of larger form (onSubmit is not called within RJSF's Form component).

### Steps To Reproduce

Repo:
1. Go to Sandbox: https://codesandbox.io/p/sandbox/new-sky-82rvdl
1. Click "Set Extra Error"
1. Click "Clear Extra Error"

Observe: Extra error remains displayed.
Expected: Extra error disappears.

Also:
1. Click "Set Extra Error"
1. Click "Clear Extra Error"
2. Repeat

Observe: Extra error is duplicated in each cycle
Expected: Extra error disappears.

### Anything else?

When deriving state from props, `extraErrors` is merged into the existing state that already includes the extra error.
https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/core/src/components/Form.tsx#L462-L466

I'm open to making a PR, but I'm not clear on what other errors RJSF may have set in the errorSchema (required fields?) or how to separate the two.

studiosciences:
@heath-freenome It's still a bug, so it's up to you.

heath-freenome:
@studiosciences I believe we may have fixed this in v6, can you upgrade and let us know?

PR #5306, opened 2026-09-19
fix(core): decide live validation by mode and keep extraErrors when deriving state

jimmycallin:

Backport to the v6 line of three live-validation fixes made on `v7` in #5297, all in the state derivation that runs when `Form` receives new props. Each reproduces on `main` at `3002e4791` with a controlled form:

1. **`liveValidate: 'onBlur'` validates when the data is replaced from outside.** `mustValidate` treated any `liveValidate` value as "validate now", so `'onBlur'` ran a full pass and showed the errors before any field was left.
2. **Every `extraErrors` entry disappears when the data is replaced.** `getStateFromProps()` passed `undefined` where `liveValidate()` takes the `extraErrors` to merge, so only the validator's own errors survived a controlled parent replacing the data.
3. **`extraErrors` duplicate on any non-data prop change.** `getCurrentErrors()` reused `state.errors`, which already carries `extraErrors` and `customErrors`, and the caller then merged both in again. A re-render changing only `className` under `liveValidate` turned `['from the server']` into `['from the server', 'from the server']`. This is the duplication half of #4408, whose reporter pointed at this merge; v6's `getDerivedStateFromProps` fixed the clearing half and the issue was closed on that, but the duplicate on any other prop change stayed. #5041 was a different path to the same symptom, through array mutations, and #5044 closed that one.

## Changes

- `mustValidate` requires `liveValidate === true || liveValidate === 'onChange'`, the predicate `processPendingChange()` already uses.
- `getCurrentErrors()` returns `schemaValidationErrors`/`schemaValidationErrorSchema` in every mode, so the caller's merge is the only merge.
- `getStateFromProps()` passes `props.extraErrors` to `liveValidate()`.
- Four regression tests in `Form.behaviors.test.tsx` under `Live validation onBlur` → `when the data is replaced from outside the form`; all four fail on `main`.
- Changelog entries under `6.10.2` → `@rjsf/core`.

## Checklist

  - [x] I've updated the changelog with a description of the PR


packages/core/src/components/Form.tsx
```diff
@@ -696,7 +695,7 @@ export default class Form<
         schemaUtils,
         state.errorSchema,
         formData,
-        undefined,
+        props.extraErrors,
```
heath-freenome, line 875: Passing `props.extraErrors` here duplicates them in `errorSchema` on the `mergeIntoOriginalErrorSchema` path. The original error schema is `state.errorSchema` (line 696), which already contains the merged `extraErrors`/`customErrors`.

Repro: `liveValidate='onChange'` with `extraErrors={{ name: { __errors: ['from the server'] } }}` and a parent whose `formData` does not follow `onChange`. Type in the field so `state.formData` differs from `props.formData`, then change a non-data prop (`className`, or `extraErrors` arriving late). `errorSchema.name.__errors` becomes `['must NOT have fewer than 8 characters', 'from the server', 'from the server']` and the field renders three `<li>`, while `state.errors` has only two entries, so the two disagree. Before this PR `extraErrors` was `undefined` on this path and the server error showed once.

`schemaValidationErrorSchema` is also polluted with the server error, so when `extraErrors` is later cleared, `getDerivedStateFromProps` rebuilds from the polluted base and the server error stays stuck on the field.

The new `className` test doesn't catch it because its `formData` equals `state.formData`, which takes the `skipLiveValidate` path. Passing `state.schemaValidationErrorSchema` as the original error schema, as `processPendingChange` does with `mergeBaseErrorSchema`, should fix it.

packages/core/src/components/Form.tsx
```diff
@@ -674,15 +677,11 @@ export default class Form<
...
-          errors: state.schemaValidationErrors || [],
-          errorSchema: state.schemaValidationErrorSchema || {},
-        };
-      }
+      // `extraErrors` and `customErrors` are merged in below, so the base has to be the validator's own result;
+      // `state.errors` already carries them and would merge each in a second time
       return {
-        errors: state.errors || [],
-        errorSchema: state.errorSchema || {},
+        errors: state.schemaValidationErrors || [],
```
heath-freenome, line 857: `getCurrentErrors()` now always reads `state.schemaValidationErrors`/`schemaValidationErrorSchema`, but `nextState` keeps the old values on a schema change. Under `'onBlur'`, which no longer revalidates on schema change, errors from the old schema come back on the next derivation.

Repro: controlled form, `liveValidate='onBlur'`, `name.minLength: 8`, data `{ name: 'short' }`. Blur the field and the error shows. Change the schema to one without `minLength`, so errors reset to `[]` and the DOM is empty. Then change only `className`: the `must NOT have fewer than 8 characters` error reappears in `state.errors` and the DOM, although the schema no longer has that rule. Before this PR `'onBlur'` revalidated on schema change, so the stored validation errors were fresh.

Resetting `schemaValidationErrors` and `schemaValidationErrorSchema` in `nextState` when `isSchemaChanged` or `noValidate` is true would fix it.

packages/core/src/components/Form.tsx
```diff
@@ -674,15 +677,11 @@ export default class Form<
...
+        errorSchema: state.schemaValidationErrorSchema || {},
```
heath-freenome, line 858: Basing on `schemaValidationErrorSchema` in every mode loses custom errors under `'onBlur'`. A `newErrorSchema` that `processPendingChange` applies at a path that already has a validator error is written only to `errorSchema`/`errors`, never to `schemaValidationErrorSchema`.

Repro: `liveValidate='onBlur'`, controlled parent. Blur `name` so it has a validator error. A custom widget then calls `onChange('abc', ['name'], { __errors: ['custom:abc'] })`. After the parent echoes `formData`, the base is `schemaValidationErrorSchema`, which never held the custom error, and the changed-field clearing wipes that path. `errorSchema` becomes `{}` and the custom error disappears. Typing another character shows it again through `customErrors`, so it flickers per keystroke. Before this PR the base was `state.errorSchema`, which kept the error consistently.

The same loss already exists for `liveValidate` false or true; this PR extends it to `'onBlur'`.

packages/core/src/components/Form.tsx
```diff
@@ -582,7 +582,10 @@ export default class Form<
     const edit = typeof inputFormData !== 'undefined';
     const liveValidate = 'liveValidate' in props ? props.liveValidate : this.props.liveValidate;
     // oxlint-disable-next-line typescript/no-deprecated
-    const mustValidate = edit && !props.noValidate && liveValidate;
+    // `'onBlur'` owns its validation pass in `onBlur()`; deriving state must not run one for it, or the errors show
+    // up before the field the user is editing has been left
+    // oxlint-disable-next-line typescript/no-deprecated
+    const mustValidate = edit && !props.noValidate && (liveValidate === true || liveValidate === 'onChange');
```
heath-freenome: With `'onBlur'` no longer setting `mustValidate`, it now takes the `!mustValidate` branch in `getSnapshotBeforeUpdate` (around line 714), and the changed-field clearing runs for it. That clearing prunes only `errorSchema`/`schemaValidationErrorSchema`, never `errors` or `schemaValidationErrors`, so state diverges and controlled and uncontrolled forms behave differently.

Repro: controlled parent that echoes `onChange` into `formData`, `liveValidate='onBlur'`, `name` too short. Blur the field so the error shows in the field and in the `ErrorList`. Type one character. `errorSchema` becomes `{}` and the inline error vanishes, but `state.errors` and `schemaValidationErrors` keep the stale message, so the `ErrorList` and the `onChange`/`onError` payload still list it. An uncontrolled form keeps the error until the next blur. Likewise, if the parent replaces the data with a now-valid value, `state.errors` still holds the old error for the valid field until the next blur or submit. Before this PR everything stayed consistent until blur.

Two things to consider: pruning `errors` and `schemaValidationErrors` for the changed paths too, or skipping the clearing for `'onBlur'`. Either way the changelog should mention the behavior change.

Minor: under `'onBlur'` the lazily evaluated `getFormDataChangedFields` walk now also runs on every props-differ update (the old truthy `mustValidate` skipped it), even when `isFormDataChanged` is false and the result is always empty. It short-circuits on identical references, so it's cheap, but skipping the callback when `isFormDataChanged` is false would avoid it.

heath-freenome: The predicate `liveValidate === true || liveValidate === 'onChange'` is now duplicated here and in `processPendingChange` (line 1066), and the two `'onBlur'` checks around lines 1200 and 1210 are its complement. The bug this PR fixes came from those sites disagreeing. A single module-level helper (for example `isLiveOnChange(liveValidate)`) would remove the drift risk.

packages/core/src/components/Form.tsx
```diff
@@ -582,7 +582,10 @@ export default class Form<
     const edit = typeof inputFormData !== 'undefined';
     const liveValidate = 'liveValidate' in props ? props.liveValidate : this.props.liveValidate;
     // oxlint-disable-next-line typescript/no-deprecated
```
heath-freenome: The original `// oxlint-disable-next-line typescript/no-deprecated` was left above the two new comment lines, and a second copy was added above `const mustValidate`. This first one now targets a comment line and suppresses nothing (`oxlint --report-unused-disable-directives` flags it as unused). It can be deleted.

packages/core/test/Form.behaviors.test.tsx
```diff
@@ -134,6 +135,105 @@ describe('Live validation onBlur', () => {
...
+
+      await user.click(container.querySelector('button')!);
+
+      expect(formRef.current!.state.errors).toEqual([
+        expect.objectContaining({ property: '.name', message: 'must NOT have fewer than 8 characters' }),
+        expect.objectContaining({ property: '.name', message: 'from the server' }),
+      ]);
+    });
+
+    it('carries each extraError over exactly once when a prop other than the data changes', async () => {
```
heath-freenome, line 231: The new `extraErrors` tests assert only `state.errors` (the flat list), and this `className` test uses a `formData` equal to `state.formData`, so it never exercises the `mergeIntoOriginalErrorSchema` path. The duplicated-in-`errorSchema` regression (see my comment on `Form.tsx` line 698) passes all four new tests because none asserts `errorSchema` or the field-level DOM errors with `state.formData` diverging from `props.formData`.

Smaller things: 'carries each extraError over exactly once' (onBlur) and 'keeps the extraErrors alongside the schema errors under onChange' differ only in mode and expected list, so they're candidates for `it.each`. The identical `extraErrors` literal is repeated three times. `ClassNameParent` is defined inline while `ReplacingParent` is hoisted.

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ should change the heading of the (upcoming) version to include a major version b
 
 # 6.10.2
 
+## @rjsf/core
+
+- Fixed `liveValidate: 'onBlur'` validating whenever the form derived state from new props, rather than only when a field was left, so replacing the form data from outside showed the errors before any blur
+- Fixed live validation dropping every `extraErrors` entry when the form derived state from new props, so a server-supplied error disappeared as soon as a controlled parent replaced the form data
```
heath-freenome: This bullet says the duplication half of #4408 is fixed, but `extraErrors` still duplicate on the `mergeIntoOriginalErrorSchema` path (my comment on `Form.tsx` line 698) and in the `processPendingChange` `newErrorSchema` branch (see the review summary). The `'onBlur'` clearing change (comment on line 588) isn't mentioned at all.

Also, none of the three bullets carries a PR link; the `CHANGELOG.md` header asks for `- [Description] ([Link to PR])`.

heath-freenome (commented):
The core suite passes on this head (1,991 tests), but several of the issues below are not covered by the new tests. Most were reproduced with throwaway tests against both this head and the pre-PR `Form.tsx`. The main one is the `extraErrors` duplication in `errorSchema` (comment on `Form.tsx` line 698).

One finding is in unchanged code, so it can't be anchored inline: `processPendingChange`'s `else if (!noValidate && newErrorSchema)` branch (around line 1125) merges `extraErrors` into `{ errorSchema: mergeBaseErrorSchema, errors }`, where `errors` is `this.state.errors`, the already-merged list. With `liveValidate` off, a validator error, and `extraErrors` on another field, a field calling `onChange(value, path, newErrorSchema)` makes `validationDataMerge` concatenate `state.errors` (which already holds the server error) with `toErrorList(extraErrors)`, so the server error appears twice in `state.errors`. It's the same bug class this PR fixes and could reasonably be fixed here by using `this.state.schemaValidationErrors` as the base list. The comment near line 982 claims the function avoids this duplication, which holds only for `errorSchema`.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed and fixed in `e4a3edada`. The merge base is now the validator's own result, the way `processPendingChange` uses `mergeBaseErrorSchema`:

```ts
(state.schemaValidationErrorSchema || {}) as ErrorSchema<S>,
```

You were right about the stuck-error half too: the doubled list was being stored as `schemaValidationErrorSchema`, so it survived `extraErrors` being cleared.

Regression test: `does not duplicate an extraError once the form data has diverged from the prop`. It uses a parent that never follows `onChange`, types one character to put `state.formData` ahead of the prop, then restyles, and asserts the DOM `<li>`s, `state.errorSchema.name.__errors` and `state.errors` all read `['must NOT have fewer than 8 characters', 'from the server']`. It fails on the reviewed head with the triple you described.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed and fixed in `e4a3edada`. `getCurrentErrors()` was the only thing being reset; the stored pair kept the old schema's result and the next derivation read it back. The reset now reaches it:

```ts
const currentErrors = getCurrentErrors();
errors = currentErrors.errors;
errorSchema = currentErrors.errorSchema;
// The stored validator result is what the next derivation builds on, so a reset has to reach it as well.
schemaValidationErrors = currentErrors.errors;
schemaValidationErrorSchema = currentErrors.errorSchema;
```

In every case other than the reset, those are the values `getCurrentErrors()` just read, so this is a no-op there rather than a second rule to keep in sync.

Regression test: `drops the stored validator errors when the schema stops producing them under onBlur`, following your repro exactly — blur, relax the schema, then change only `className`.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed, and fixed at the source rather than by special-casing `'onBlur'` in the derivation. The reason the base never had the custom error is that `processPendingChange` writes a `newErrorSchema` landing on a path that already has a validator error into `mergeBaseErrorSchema` and then never stores it, so `schemaValidationErrorSchema` in state is the pre-`newErrorSchema` one. It is stored now:

```ts
state = {
  ...state,
  formData: newFormData,
  ...mergedErrors,
  schemaValidationErrorSchema: mergeBaseErrorSchema,
  customErrors,
};
```

That also closes the pre-existing loss you noted for `liveValidate` false and true, since it is the same write.

Regression test: `keeps a custom error raised on a path that already carries a validator error` — blur `name` for the validator error, then a widget raising `{ __errors: ['custom:…'] }` on every keystroke, with a parent echoing `onChange`. Fails on the reviewed head.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed. Took your second option: the clearing stays off for `'onBlur'`, so it runs exactly where it ran before this PR.

```ts
if (!mustValidate && liveValidate !== 'onBlur') {
```

Teaching the clearing to prune `errors` and `schemaValidationErrors` as well is the larger fix, but it would be a new behaviour for `'onBlur'` on a PR whose point is that `'onBlur'` waits for the blur — and it leaves the same divergence in place for every other mode that reaches the branch. Keeping the branch as-is and not extending its reach means this PR adds no new inconsistency, and pruning both halves stays a separate, wider change.

The changelog now says what that means for users: an `'onBlur'` form holds a changed field's error until the field is left. Regression test: `keeps the inline errors and the error list in agreement while typing under onBlur`.

On the minor point: `getFormDataChangedFields` no longer runs when the data did not change. The callback short-circuits at the call site now, so the lazy walk is not even asked for:

```ts
() => (isFormDataChanged ? getChangedFields(this.props.formData, prevProps.formData, true) : []),
```

jimmycallin: Agreed — that is the drift this PR exists because of. `isLiveOnChange()` is now a module-level helper in `Form.tsx` and both sites call it.

I left the two `liveValidate === 'onBlur'` checks in `onBlur()` alone. They are not the complement: `false` and `undefined` are neither `'onChange'` nor `'onBlur'`, and those two sites genuinely mean "the blur owns this pass", not "not onChange". Folding them into the helper would make the third state read as `'onBlur'`.

packages/core/src/components/Form.tsx
jimmycallin: Deleted. The remaining one sits directly above `const mustValidate`, where `props.noValidate` is; `oxlint --report-unused-disable-directives` over `src` and `test` reports nothing for `Form.tsx` now.

packages/core/test/Form.behaviors.test.tsx
jimmycallin: All four done.

- The `errorSchema` and DOM gap is covered by the new `does not duplicate an extraError once the form data has diverged from the prop`, in a new `Error state consistency when deriving from new props` block. It types a character against a parent that ignores `onChange`, so `state.formData` is ahead of the prop and the derivation takes the `mergeIntoOriginalErrorSchema` path, then asserts the `<li>`s, `state.errorSchema.name.__errors` and `state.errors` together.
- The two mode-only variants are one `it.each` over `[['onBlur', [...]], ['onChange', [...]]]`.
- The `extraErrors` literal is hoisted as `serverErrors` next to `objectSchema`.
- `ClassNameParent` is hoisted as `RestylingParent`, alongside `ReplacingParent`.

Four more regressions came out of your other comments; each one fails on the reviewed head, checked individually by restoring `Form.tsx` and running the block.

CHANGELOG.md
jimmycallin: Rewritten. The duplication bullet now names all three paths it actually covers, since the two you found are fixed in this PR as well: the no-re-validation derivation, the `mergeIntoOriginalErrorSchema` derivation, and `processPendingChange`'s `newErrorSchema` branch. Two new bullets cover the stale stored validation result and the lost custom error, and the `'onBlur'` bullet now states the clearing change in user terms — an `'onBlur'` form keeps a changed field's error until the field is left.

Every bullet carries `([#5306](…/pull/5306))`.

On the summary finding: fixed too. `processPendingChange` now reads `schemaValidationErrors` and merges against that instead of `state.errors`, and the comment above `mergeBaseErrorSchema` says the list needs the same treatment as the schema rather than implying it is already handled. Test: `lists a server error once when a field raises a custom error with live validation off`, on an uncontrolled form so nothing re-derives afterwards and the merge is the last word.

jimmycallin:
Rebased onto `main` at `4d1cb92af` (only the changelog conflicted) and the review round is in `e4a3edada`. Every thread has a reply naming the change and its test.

All five findings were real, and all five are fixed rather than argued with:

| Finding | Fix |
| --- | --- |
| `extraErrors` duplicated on the `mergeIntoOriginalErrorSchema` path | merge base is `state.schemaValidationErrorSchema`, as `processPendingChange` already does |
| stale stored validation result after a schema change under `'onBlur'` | the `getCurrentErrors()` reset now reaches `schemaValidationErrors`/`schemaValidationErrorSchema` |
| custom error lost on a path that already had a validator error | `processPendingChange` stores `mergeBaseErrorSchema` as the new base; this also closes the pre-existing loss for `liveValidate` false and true |
| changed-field clearing now running for `'onBlur'` | it stays off for `'onBlur'`, so it runs exactly where it ran before |
| `state.errors` as the merge base in the `newErrorSchema` branch (your summary) | reads `schemaValidationErrors` instead |

Plus `isLiveOnChange()` for the duplicated predicate, the dead `oxlint-disable` deleted, `getFormDataChangedFields` skipped when the data did not change, and the test restructuring you asked for (`it.each`, hoisted literal, hoisted parent).

Five regression tests in a new `Error state consistency when deriving from new props` block. Each was checked individually against the reviewed `Form.tsx`: four fail there outright, and the `newErrorSchema` one needed an uncontrolled form to fail, since a controlled parent's echo re-derives the state and hides the duplicate before the assertion.

Verification from the repo root: `typecheck`, `cs-check`, `knip` and `lint` clean, `oxlint --report-unused-disable-directives` reports nothing for `Form.tsx`, and `nx run-many --target=test` passes for all 17 projects. `@rjsf/core` is at 2,070.

**One thing for v7.** Three of these survive #5297 on `v7` today, since they are in code that PR did not touch: `getCurrentErrors()`'s reset still does not reach the stored pair (`Form.tsx:670`), the changed-field clearing still runs for everything that is not `'onChange'` (`:690`), and `processPendingChange` still merges against `state.errors` (`:1067`). The duplication one does not apply, because v7's `liveValidate()` has no `originalErrorSchema` parameter. Happy to forward-port the three as a separate PR against `v7` once this one settles.

jimmycallin:
> One thing for v7. Three of these survive https://github.com/rjsf-team/react-jsonschema-form/pull/5297 on v7 today, since they are in code that PR did not touch: getCurrentErrors()'s reset still does not reach the stored pair (Form.tsx:670), the changed-field clearing still runs for everything that is not 'onChange' (:690), and processPendingChange still merges against state.errors (:1067). The duplication one does not apply, because v7's liveValidate() has no originalErrorSchema parameter. Happy to forward-port the three as a separate PR against v7 once this one settles.

I can also close this PR, and fix them in v7 branch if you'd rather focus on v7 at this point.

heath-freenome (commented):
Three findings from a correctness pass over the diff. All three are in the part of the second commit that has no v7 counterpart (I diffed `v7:packages/core/src/components/Form.tsx`: `getCurrentErrors()`, the `props.extraErrors` pass-through and the `mustValidate` predicate match v7 exactly; the `'onBlur'` clearing exclusion, the `schemaValidationErrors/Schema` reset in the `else` branch and the `schemaValidationErrorSchema` store in `processPendingChange()` are new here).

Everything else checked out: full `@rjsf/core` suite (2070 pass), `@rjsf/utils` at 100% coverage, root `lint` and root `typecheck` all clean, and the five new tests do each fail against commit 1's `Form.tsx`. I also cleared the `isFormDataChanged ? … : []` short-circuit (deep-equal data yields no changed fields, so it is a pure optimization), the `errors` -> `schemaValidationErrors` swap at line 1263 (every writer keeps `schemaValidationErrors` as the validator-only subset, so nothing is lost), and `props.extraErrors`/`props.noValidate` being read off `props` directly (all four call sites pass `this.props`).

One aside: the bug behind `lists a server error once when a field raises a custom error with live validation off` reproduces on `v7` too, so that half probably wants forward-porting.

packages/core/src/components/Form.tsx
```diff
@@ -1226,13 +1256,24 @@ export default class Form<
...
       );
-      state = { ...state, formData: newFormData, ...mergedErrors, customErrors };
+      // A `newErrorSchema` landing on a path that already had a validator error was written into
+      // `mergeBaseErrorSchema` rather than into `customErrors`, so the stored base has to carry it too. Otherwise the
+      // next derivation rebuilds from a base that never saw it and the field loses the error until the next keystroke
+      state = {
+        ...state,
+        formData: newFormData,
+        ...mergedErrors,
+        schemaValidationErrorSchema: mergeBaseErrorSchema,
```
heath-freenome: This store is unconditional, but on the root path `mergeBaseErrorSchema` is not "the base with one path replaced" -- line 1225 sets it to `newErrorSchema` alone. Storing that replaces the whole validator base, so every *other* field's validator error is permanently dropped from it.

Repro, verified on this branch (and the old behaviour verified against `4d1cb92af`): schema `{ type: 'object', properties: { name: { type: 'string', minLength: 8 } } }`, `initialFormData={{ name: 'short' }}`, live validation off.

1. Submit -> `schemaValidationErrorSchema === {name:{__errors:['must NOT have fewer than 8 characters']}}`, inline error shown.
2. A root-path raise -- `onChange(data, [], { __errors: ['root level problem'] })`, which a custom root `Field` does, and which a root-level `ArrayField` reorder/remove/copy also takes, since `isRootPath` is true for `path === ['root']` -> `schemaValidationErrorSchema === {__errors:['root level problem']}`.
3. Any later derivation (the parent re-rendering with a new `className` is enough) rebuilds from that base: `.name` never shows its inline error again, while `state.errors` still lists it. So the `ErrorList` / `onError` / `onChange` payload claims `.name must NOT have fewer than 8 characters` against a field that shows nothing, permanently.

On `main` step 3 restored the `.name` error, because the base was left intact.

Suggested fix: store the base only in the `!isRootPath` branch, where `setByPath` produced a targeted clone, or merge rather than replace at the root -- i.e. keep what gets stored separate from the display base instead of reusing `mergeBaseErrorSchema` for both.

packages/core/src/components/Form.tsx
```diff
@@ -1226,13 +1256,24 @@ export default class Form<
...
-        { errorSchema: mergeBaseErrorSchema, errors },
+        { errorSchema: mergeBaseErrorSchema, errors: schemaValidationErrors },
         extraErrors,
         customErrors,
       );
-      state = { ...state, formData: newFormData, ...mergedErrors, customErrors };
+      // A `newErrorSchema` landing on a path that already had a validator error was written into
+      // `mergeBaseErrorSchema` rather than into `customErrors`, so the stored base has to carry it too. Otherwise the
+      // next derivation rebuilds from a base that never saw it and the field loses the error until the next keystroke
+      state = {
```
heath-freenome: Related to the comment below, but a separate consequence: this updates only the schema half of the stored validator pair. `schemaValidationErrors` keeps its old value while `schemaValidationErrorSchema` gains the field-raised entry and loses the validator entry that `setByPath` overwrote at that path.

That pair is the base every later derivation and `getDerivedStateFromProps()` rebuilds from, so the mismatch is now persistent rather than transient. In the scenario of the new `keeps a custom error raised on a path that already carries a validator error` test, the field shows `custom:shorty` while `state.errors` -- the `ErrorList` and the `onChange`/`onError` payload -- shows `must NOT have fewer than 8 characters`, and that disagreement survives every subsequent re-render instead of being corrected by the next rebuild from the base.

If the raised error is meant to live in the base, the matching list needs the same treatment (recompute it with `toErrorList()` over the new base, or keep the raise in `customErrors` so `mergeErrors()` puts it into both halves).

packages/core/src/components/Form.tsx
```diff
@@ -818,8 +832,16 @@ export default class Form<
...
+      // The stored validator result is what the next derivation builds on, so a reset has to reach it as well.
+      // Without this, a `'onBlur'` form, which no longer re-validates on a schema change, brings the old schema's
+      // errors back on the next props change. In every other case these are the values it just read.
+      schemaValidationErrors = currentErrors.errors;
+      schemaValidationErrorSchema = currentErrors.errorSchema;
+      // We only update the error schema for changed fields if mustValidate is false. `'onBlur'` is excluded
+      // although its `mustValidate` is now false: the clearing prunes `errorSchema` but not `errors`, so running it
+      // here would drop a field's inline error on the first keystroke while the `ErrorList` and the `onChange`
+      // payload kept it. An `'onBlur'` form holds its errors until the field is left, as it did before.
+      if (!mustValidate && liveValidate !== 'onBlur') {
```
heath-freenome: `liveValidate !== 'onBlur'` turns the changed-field clearing off for `'onBlur'`, which the v7 implementation this backports does not do -- v7 runs this block for every mode that is not `'onChange'`. Combined with `'onBlur'` no longer re-validating on derivation, a controlled `'onBlur'` form now never refreshes errors for data it did not see typed.

Verified regression against `4d1cb92af`: parent holds `{name:'short'}`, the user blurs `name` -> `must NOT have fewer than 8 characters`; the parent then replaces the data with `{name:'longenoughvalue'}`. On `main` the error clears. On this branch it stays in the inline error, in the `ErrorList` and in `state.errors`, and the only ways out are blurring that specific field or submitting -- so a parent that swaps to a different record displays the previous record's errors against the new data, and a parent that fixes the data programmatically leaves an error on a field the user may never touch.

The comment's rationale is that the clearing prunes `errorSchema` but not `errors`, but that asymmetry already exists for the live-validation-off mode, so it argues for pruning both rather than for withholding the clearing from one mode. Either run it for `'onBlur'` too, or give the mode some way to drop errors for replaced data.

packages/core/src/components/Form.tsx
jimmycallin: Fixed in `9b04bc2a5`. The root replacement now applies to what is displayed only: `mergeBaseErrorSchema`/`mergeBaseErrors` carry the raised schema and its `toErrorList()` for the merge, while the stored pair is written only in the `!isRootPath` branch, where `setByPath()` produced a targeted clone. Your repro is the test `keeps every other field's validator error when a root-path raise replaces the root errors` (submit, root raise via `formRef.current.onChange(data, [], { __errors })`, restyle): `#root_name__error` keeps `must NOT have fewer than 8 characters`. Fails on `e4a3edada`.

packages/core/src/components/Form.tsx
jimmycallin: Fixed in `9b04bc2a5`. The list is rewritten alongside the schema: the validator entries at or below the raise's path are filtered out (comparing `toPath(error.property)` segment-wise against the path) and `toErrorList(newErrorSchema, path)` is appended, and both halves are stored together as `storedValidation`. I kept the raise in the base rather than routing it through `customErrors`: the existing reorder test (`should keep the errors that a reorder of the array items left standing`) relies on the changed-field clearing reaching the remapped item errors an `ArrayField` raises, which it cannot do for `customErrors`. `toErrorList()` losing `name`/`params` is inherent to a raise, since an `ErrorSchema` never carried them.

The existing `keeps a custom error raised on a path that already carries a validator error` test now asserts `state.errors` and the inline list both read `['custom:shorty']` after a re-derivation. It is an uncontrolled form restyled without a click now, because with the clearing running for `'onBlur'` (thread below) an echoing parent's re-derivation treats the raise's own path as changed data and drops it, as it already did with live validation off. A raise arriving with the change it describes surviving that clearing needs the derivation to know about the raise, which is a separate change.

packages/core/src/components/Form.tsx
jimmycallin: Fixed in `9b04bc2a5`, the way you argued it: the clearing runs for `'onBlur'` again, and it prunes `errors`/`schemaValidationErrors` alongside the schema, for every mode that reaches the branch. An entry goes when its `toPath(error.property)` is at or below a changed field's path, or is the path of a container holding it (the containers whose own `__errors` the schema clearing already dropped); root-level entries stay, matching the `for (let i = 1; …)` loop. Tests: `drops the errors of data the parent replaced under onBlur` (your `{name:'short'}` → `{name:'longenoughvalue'}` swap: field, list and `state.errors` all empty) and `clears the errors of a changed field in both the field and the error list while typing under onBlur`, which replaces the earlier `keeps the inline errors and the error list in agreement while typing under onBlur` since the agreement now comes from both halves clearing. Both fail on `e4a3edada`. The changelog bullet no longer claims an `'onBlur'` form holds a changed field's error until it is left.

jimmycallin:
Round two is in `9b04bc2a5`; each of the three threads has a reply naming the change and its test. Four regression tests were added to the `Error state consistency when deriving from new props` block and each fails on `e4a3edada`. From the repo root: `typecheck`, `lint`, `knip` and `cs-check` clean; `@rjsf/core` at 2,072 passing.

Two of the three carry over to `v7`: `processPendingChange` there still merges against a list that does not follow the raise, and the root replacement is still stored unless #5312's extraction changed it. The `'onBlur'` clearing exclusion was only ever on this branch. The forward-port offer above stands.

packages/core/src/components/Form.tsx
```diff
@@ -1185,14 +1245,27 @@ export default class Form<
...
+          // the user's error at this path without mutating shared state. The list says the same as the schema: the
+          // errors at and below the path are the raised ones now. Both halves are stored, so the next derivation
+          // rebuilds from a base that carries the raise instead of losing it until the next keystroke
           mergeBaseErrorSchema = structuredClone(schemaValidationErrorSchema);
           // An `ErrorSchema` nests plain objects even at numeric segments, so never auto-vivify arrays
           setByPath(mergeBaseErrorSchema, path, newErrorSchema, true);
+          mergeBaseErrors = schemaValidationErrors
+            .filter((error) => !isPathPrefix(path, errorPath(error)))
+            .concat(toErrorList(newErrorSchema, path.map(String)));
+          storedValidation = {
```
heath-freenome: Storing the raise here writes `extraErrors` into the stored validator result when the raise is an `ArrayField` reorder of an array that isn't at the root. `ArrayField` builds that raise by remapping its `errorSchema` prop, which is what's displayed, so it already carries `extraErrors`/`customErrors`. The remapped copy then gets stored as `schemaValidationErrorSchema`/`schemaValidationErrors` and survives `extraErrors` being cleared.

Repro (live validation off or `'onBlur'`, parent that doesn't echo `onChange`):
- schema `{ arr: { type: 'array', items: { type: 'string', minLength: 3 } } }`, `formData={{ arr: ['a', 'bbbb', 'cccc'] }}`, `extraErrors={{ arr: { 2: { __errors: ['server'] } } }}`
- submit, then move item 1 down
- rerender with `extraErrors={undefined}`

| Commit | After the reorder | After clearing `extraErrors` |
| --- | --- | --- |
| `4d1cb92af` (main) | `server` listed twice | cleared |
| `e4a3edada` | `server` shown twice | still on the field at `arr.1`, not in the list |
| `9b04bc2a5` | `.arr.1` and `.arr.2` in field and list | stuck at `.arr.1` in field and list, and `schemaValidationErrorSchema.arr.1.__errors` is `['server']` |

The duplication after the reorder already happens on main, but the stuck error is new in this PR. It's the same pollution the first review round fixed, now coming in through the raise. A parent that echoes `onChange` mostly hides it, because the next derivation clears the changed indexes.

Two options: don't store a raise that is an `ArrayField` remap (the reorder test only needs it displayed), or remove the `extraErrors`/`customErrors` entries under `path` from `newErrorSchema` before it goes into `storedValidation`.

CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ should change the heading of the (upcoming) version to include a major version b
...
+- Fixed a stored validation result outliving the schema that produced it under `liveValidate: 'onBlur'`, which no longer re-validates on a schema change: the next change to any other prop brought back an error the current schema cannot raise ([#5306](https://github.com/rjsf-team/react-jsonschema-form/pull/5306))
+- Fixed an `ErrorSchema` a field raises through `onChange`, on a path that already carried a validator error, being lost as soon as the form derived state again, so a controlled form flickered the error once per keystroke; the stored error list now carries the raise too, so the `ErrorList` says what the field does, and a raise at the root, which replaces every displayed error, no longer replaces the stored validator result, which used to drop every other field's error for good ([#5306](https://github.com/rjsf-team/react-jsonschema-form/pull/5306))
```
heath-freenome: This bullet claims more than the code does. "so a controlled form flickered the error once per keystroke" reads as if the controlled case is fixed. But per your reply on the custom-error thread, a parent that echoes `onChange` still drops the raise, because the re-derivation treats that field as changed data and clears it. That's why the test became an uncontrolled form.

As it stands, the raise survives re-derivations that don't touch that field (a `className` change, a new `extraErrors` object), not the echo of the keystroke that raised it. Could the bullet say that, or drop the flicker claim?

heath-freenome (commented):
Round two looks right: the clearing now reaches the error list, the root raise no longer replaces the stored result, and the replies match the code. Two things left, inline. One is a new stuck-`extraErrors` path through `ArrayField` reorders of an array below the root. The other is a changelog bullet that claims more than the code does.

Also, the next two issues already happen on main, so they're not blockers for this PR:
- **Array at the root:** a reorder still lists an `extraErrors` entry twice, because `toErrorList(newErrorSchema)` already includes it. Since the root raise isn't stored, any new `extraErrors` object (for example an inline literal) resets the display to the pre-reorder indexes.
- **Under `'onBlur'`:** a controlled form now clears a field's error on the first keystroke, but an uncontrolled form keeps it until the field is left. Live validation off already behaves this way. It might be worth a clause in the changelog.

packages/core/src/components/Form.tsx
jimmycallin: Reproduced your table on `9b04bc2a5` (`server` stuck at `.arr.1` in field, list and `schemaValidationErrorSchema` after clearing `extraErrors`) and fixed in `8044a21c8` with the first option: a raise for an array value is merged for display only, like a root raise, and not stored. `ArrayField`'s reorder/remove/copy raises are the only raises that arrive with an array value from the library itself, and they are always a remap of the displayed `errorSchema`, so `extraErrors`/`customErrors` never enter the stored result through them; a custom field raising on an array-valued path is displayed until the next derivation, as on `main`. The reorder test still passes, since it only needs the remap displayed. Regression test `lets the parent clear the extraErrors an ArrayField reorder remapped`: your schema and data, submit, move item 1 down, clear `extraErrors`; asserts the field and `state.errors` both read only `must NOT have fewer than 3 characters`. Fails on `9b04bc2a5` with `server` still present.

CHANGELOG.md
jimmycallin: Rewritten in `8044a21c8`. The bullet now says the raise survives derivations that do not touch its field (a `className` change, a new `extraErrors` object) but not a controlled parent echoing the change that raised it, which the changed-field clearing still treats as new data for that field; the flicker claim is gone. It also carries the two clauses from your summary: the `ArrayField` remap is displayed but not stored, and under `'onBlur'` a controlled form clears a field's error on the first keystroke while an uncontrolled form keeps it until the field is left, as with live validation off.

jimmycallin:
Round three is in `8044a21c8`: the `ArrayField` remap is no longer stored (your first option), with a regression test that fails on `9b04bc2a5`, and the changelog bullet is rewritten to say what the code does, including the `'onBlur'` controlled/uncontrolled clause. The two `main` issues you listed (the root-array double listing and the reset to pre-reorder indexes) are left alone here. `@rjsf/core` at 2,073 passing; root `typecheck` and `lint` clean. The branch is on `main` at `4d1cb92af`, which has not moved.

packages/core/src/components/Form.tsx
```diff
@@ -1185,14 +1245,31 @@ export default class Form<
...
+          // rebuilds from a base that carries the raise instead of losing it until the next keystroke
           mergeBaseErrorSchema = structuredClone(schemaValidationErrorSchema);
           // An `ErrorSchema` nests plain objects even at numeric segments, so never auto-vivify arrays
           setByPath(mergeBaseErrorSchema, path, newErrorSchema, true);
+          mergeBaseErrors = schemaValidationErrors
+            .filter((error) => !isPathPrefix(path, errorPath(error)))
+            .concat(toErrorList(newErrorSchema, path.map(String)));
+          // An `ArrayField` raises the item errors it remapped after a reorder, remove or copy, taken from the displayed
+          // `errorSchema`, so they carry `extraErrors`/`customErrors` that must not outlive the props supplying them
+          if (!Array.isArray(newValue)) {
```
heath-freenome, line 1304: `ArrayField` isn't the only core field that sends back its displayed `errorSchema`, so an array check misses other ways in. These also pass their own `errorSchema` prop (which already carries `extraErrors`/`customErrors`) to `onChange`, with a non-array value:

- `OptionalDataControlsField.tsx:59` / `:69`: Remove (`undefined`) and Add (an object)
- `FallbackField.tsx:103`: a type change (usually not an array)

If the path already has a validator error, the raise gets past this check and is stored, `extraErrors` included.

Repro (live validation off or `'onBlur'`, a parent that doesn't echo `onChange`):
- `uiSchema={{ 'ui:globalOptions': { enableOptionalDataFieldForType: ['object'] } }}`
- schema `{ type: 'object', properties: { addr: { type: 'object', properties: { street: { type: 'string', minLength: 3 } } } } }`, `formData={{ addr: { street: 'a' } }}`, `extraErrors={{ addr: { street: { __errors: ['server'] } } }}`
- submit, click Remove on `addr`, rerender with `extraErrors={undefined}`, then change `className`

| Commit | `schemaValidationErrorSchema` after Remove | After clearing `extraErrors` + restyle |
| --- | --- | --- |
| `4d1cb92af` (main) | validator error only | `server` gone |
| `8044a21c8` | `addr.street.__errors`: `['must NOT have fewer than 3 characters', 'server']` | `server` still in `state.errors` and the stored result |

I checked `FallbackField` by reading the code, not with a repro. It's the same pattern.

Since the value type can't tell a real raise from a field sending back what's already on screen, I'd strip the `extraErrors`/`customErrors` entries from the raise before it goes into `storedValidation`, so the stored result only ever holds validator errors. That covers the `ArrayField` remap too, and third-party fields that copied this pattern. Changing the two core fields alone (for example Remove passing `undefined`) would fix core but not custom fields.

heath-freenome (commented):
Round three fixes both earlier threads, and the changelog now matches the code. One more issue, inline: the `Array.isArray(newValue)` check misses other core fields that send back their displayed `errorSchema`, so the stuck-`extraErrors` bug still reproduces through `OptionalDataControlsField`'s Remove button.

packages/core/src/components/Form.tsx
jimmycallin: Reproduced on `8044a21c8` and fixed in `3f4f0fb0f` your way. Before the raise is merged into the base, `withoutSuppliedErrors()` drops every message that `extraErrors` or `customErrors` holds at the same position, along with any `__errors` left empty. That covers `OptionalDataControlsField` Add/Remove, `FallbackField` and third-party fields that hand back their displayed `errorSchema`. It also stops the displayed `extraErrors` being doubled on that path, since `mergeErrors()` adds them back once. The `Array.isArray(newValue)` check stays, because an `ArrayField` remap moves the supplied messages to indexes where they no longer line up. Known limit: a real raise whose text is identical to an `extraErrors` message at the same node is displayed but not stored. Regression test: `lets the parent clear the extraErrors an optional object Remove sent back`, which is your repro (submit, Remove, clear `extraErrors`, restyle). It asserts `state.errors` and `schemaValidationErrorSchema` hold only the validator error, and fails on `8044a21c8` with `server` still there. The changelog bullet now describes this.

packages/core/src/components/Form.tsx
```diff
@@ -1185,14 +1272,35 @@ export default class Form<
...
+          }
         } else {
+          // A root raise is the whole error schema remapped, the way `ArrayField` reshuffles the item errors after a
+          // reorder, so it replaces what is displayed. It does not replace the stored base: that holds every field's
+          // own validator errors, which the next derivation brings back
           mergeBaseErrorSchema = newErrorSchema;
+          mergeBaseErrors = toErrorList(newErrorSchema);
         }
       } else {
         if (!customErrors) {
```
heath-freenome, line 1318: `withoutSuppliedErrors()` runs on only one of the two branches that store a raise. When the raised path has no validator error yet, the raise lands here (`setByPath(customErrors.ErrorSchema, path, newErrorSchema, true)`, line 1317) verbatim, `extraErrors` messages included, and `customErrors` is form state that nothing clears — so the server error outlives the prop supplying it, exactly what the commit message says it must not do.

Repro is your new test with `minLength: 3` dropped from `street`, so nothing but `extraErrors` reports on that path:

- `uiSchema={{ 'ui:globalOptions': { enableOptionalDataFieldForType: ['object'] } }}`
- schema `{ type: 'object', properties: { addr: { type: 'object', properties: { street: { type: 'string' } } } } }`, `formData={{ addr: { street: 'a' } }}`, `extraErrors={{ addr: { street: { __errors: ['server'] } } }}`
- submit, click Remove on `addr`, rerender with `extraErrors={undefined}`, then change `className`

On `4d1cb92af`, `8044a21c8` and `3f4f0fb0f` alike: after Remove, `customErrors.ErrorSchema` is `{addr:{street:{__errors:['server']}}}`; after clearing `extraErrors` and restyling, `server` is still in `state.errors` and `state.errorSchema`.

It isn't about the form being otherwise error-free either — with `other: { type: 'string', minLength: 3 }` failing alongside, the final `state.errors` is `['must NOT have fewer than 3 characters', 'server']`.

This one is unchanged from `main`, so it's not a regression from this PR. I'm raising it because the changelog bullet now says a raise that hands back the displayed `errorSchema`, "as the optional-data Add and Remove buttons and the fallback field's type change do, is stored without the `extraErrors` and `customErrors` it carries" — and that holds for only one of the two places such a raise is stored. The optional-data Remove reaches both, depending on whether the path already had a validator error.

The helper covers this branch too, with one caveat: strip against `extraErrors` only, not `customErrors`. `setByPath` replaces the node at `path`, so a field re-raising an error `customErrors` already holds would strip itself to `{}` and lose it.

```ts
setByPath(customErrors.ErrorSchema, path, withoutSuppliedErrors(newErrorSchema, [getByPath(extraErrors, path)]), true);
```

I ran that: it clears the repro and the core suite stays green (2074 tests).

packages/core/src/components/Form.tsx
```diff
@@ -404,6 +404,61 @@ function withoutAbsentValues(data: unknown): unknown {
...
+ * @param supplied - The `ErrorSchema`s, aligned with `raised`, whose messages are left out
+ * @returns - `raised` with the supplied messages and any emptied `__errors` removed
+ */
+function withoutSuppliedErrors<T>(raised: ErrorSchema<T>, supplied: unknown[]): ErrorSchema<T> {
+  const at = (node: unknown, key: string): unknown =>
+    isObject(node) && Object.hasOwn(node, key) ? node[key] : undefined;
+  return Object.entries(raised).reduce((acc: GenericObjectType, [key, value]) => {
+    const suppliedHere = supplied.map((node) => at(node, key));
+    if (key === ERRORS_KEY && Array.isArray(value)) {
+      const kept = value.filter((message) =>
```
heath-freenome: `filter` drops **every** copy of a message `extraErrors` holds at that node, the validator's own included. `mergeErrors()` merges `extraErrors` with `concatArrays: true`, not `'preventDuplicates'`, so when the server reports the same text the validator does, the displayed `errorSchema` the field hands back carries it twice and both copies go.

You call this a known limit, but it's the normal shape of a server that validates with the same JSON Schema — the messages are identical by construction — and it's a regression against `main`, not a pre-existing limit.

Your new test with only `addrServerErrors` changed to the message AJV produces:

```ts
const addrServerErrors = { addr: { street: { __errors: ['must NOT have fewer than 3 characters'] } } };
```

| Commit | final `state.errors` | final `schemaValidationErrorSchema` |
| --- | --- | --- |
| `4d1cb92af` (main) | `['must NOT have fewer than 3 characters']` | `{addr:{street:{__errors:['must NOT have fewer than 3 characters']}}}` |
| `8044a21c8` | the message twice | the message twice |
| `3f4f0fb0f` | `[]` | `{addr:{street:{}}}` |

Both of the test's assertions fail on `3f4f0fb0f` and the field is left with no error at all.

Since the raise carries one copy per source, removing one occurrence per supplied occurrence rather than filtering all matches keeps the validator's own:

```ts
const remaining = suppliedHere.flatMap((messages) => (Array.isArray(messages) ? messages : []));
const kept = value.filter((message) => {
  const found = remaining.indexOf(message);
  if (found === -1) {
    return true;
  }
  remaining.splice(found, 1);
  return false;
});
```

That's exact for `extraErrors`. `customErrors` merges with `'preventDuplicates'`, so a collision there still leaves one copy and it's still the validator's that goes — but `customErrors` is state that is re-merged on every derivation, so the message stays displayed.

Minor, same helper: an emptied `__errors` is dropped but the object that held it isn't, so the stored schema keeps phantom branches — `{addr:{street:{}}}` in the table above. `Object.keys()` on one is non-zero, so the `oldValidationError` test at line 1274 reads a parent of that path as still carrying a validator error and routes the next raise into the store branch instead of `customErrors`. I noticed that reading the code, not from a repro.

heath-freenome (commented):
Verified `3f4f0fb0f`: it does fix the optional-data Remove repro from the last round, and lint, typecheck and the full core suite (2074 tests) are green. Two things on the new `withoutSuppliedErrors()`, inline.

The first is the same bug through the sibling branch — a raise on a path with no validator error yet goes into `customErrors` unstripped, so the server error still outlives `extraErrors`. That one is unchanged from `main`, but it's what the new changelog bullet promises.

The second is a regression: stripping by message identity removes the validator's own error when `extraErrors` reports the same text, which is what a server validating against the same schema produces. Your own new test fails with one word changed, and the field is left showing nothing.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed: with `minLength` dropped from `street`, `server` was still in `state.errors` after Remove, clearing `extraErrors` and restyling. Fixed in aa17758d0 as you suggested: the `customErrors` branch now stores `withoutSuppliedErrors(newErrorSchema, [getByPath(extraErrors, path)])`, stripping against `extraErrors` only for the reason you gave. The Remove test is now an `it.each` and includes this case as "the only error on the path".

packages/core/src/components/Form.tsx
jimmycallin: Confirmed. With the server sending the AJV message, the final `state.errors` was `[]`. Fixed in aa17758d0 with your approach: each supplied copy removes one occurrence, so the validator's own copy stays. The Remove `it.each` covers this as "the message the validator reports". Your `customErrors`/`'preventDuplicates'` caveat still applies, and the message stays displayed.

I fixed the phantom-branch point in the same commit: a child object the helper empties is now dropped, so `{addr:{street:{}}}` isn't stored any more and can't fool the `oldValidationError` check.

packages/core/src/components/Form.tsx
```diff
@@ -1206,7 +1326,14 @@ export default class Form<
           }
         } else {
           // An `ErrorSchema` nests plain objects even at numeric segments, so never auto-vivify arrays
-          setByPath(customErrors.ErrorSchema, path, newErrorSchema, true);
+          // Only `extraErrors` is left out: `setByPath` replaces the node, so leaving out what `customErrors` holds
+          // there would drop an error a field re-raises
+          setByPath(
+            customErrors.ErrorSchema,
+            path,
+            withoutSuppliedErrors(newErrorSchema, [getByPath(extraErrors, path)]),
```
heath-freenome: `withoutSuppliedErrors()` now covers this branch and the store branch above it, but not the third place a raise is stored: the `isRootPath` side of this same `else`, `customErrors.setErrors(pathErrors)` at line 1325. A root raise still goes into `customErrors` with whatever `extraErrors` supplied at the root, and `customErrors` is state nothing clears.

Repro (live validation off, a parent that does not echo `onChange`):
- schema `{ type: 'object', properties: { street: { type: 'string' } } }`, `formData={{ street: 'a' }}`, `extraErrors={{ __errors: ['server'] }}` -- a form-level server error, nothing the validator reports
- submit -> `state.errors` is `['server']`, displayed root `errorSchema` is `{__errors:['server']}`
- the root field hands back what it displays: `formRef.current.onChange({ street: 'b' }, [], state.errorSchema)`, which is the idiom your own `keeps every other field's validator error when a root-path raise replaces the root errors` test calls "what a custom root `Field`, or a root-level `ArrayField` reorder, hands to `onChange`"
- rerender with `extraErrors={undefined}`, then change `className`

`customErrors.ErrorSchema` is `{__errors:['server']}` right after the raise, and the final `state.errors` and `state.errorSchema` still read `server`. Identical on `4d1cb92af`, so this is not a regression from the PR -- I am raising it because the changelog bullet says a raise that hands back the displayed `errorSchema` "is stored without the `extraErrors` and `customErrors` it carries", and that now holds for two of the three places a raise is stored.

`setErrors()` takes a list rather than a node, so the fix is not the same call shape, but it is the same idea:

```ts
const pathErrors = withoutSuppliedErrors(newErrorSchema, [extraErrors])[ERRORS_KEY];
```

Small related note while you are in here: the root branch stores only `newErrorSchema[ERRORS_KEY]`, so everything nested in a root raise is dropped. That is also unchanged from `main` and I am not asking for it in this PR.

packages/core/src/components/Form.tsx
```diff
@@ -1185,14 +1284,35 @@ export default class Form<
...
+          // errors at and below the path are the raised ones now. Both halves are stored, so the next derivation
+          // rebuilds from a base that carries the raise instead of losing it until the next keystroke
           mergeBaseErrorSchema = structuredClone(schemaValidationErrorSchema);
+          const raisedErrorSchema = withoutSuppliedErrors(newErrorSchema, [
+            getByPath(extraErrors, path),
+            getByPath(customErrors?.ErrorSchema, path),
+          ]);
           // An `ErrorSchema` nests plain objects even at numeric segments, so never auto-vivify arrays
-          setByPath(mergeBaseErrorSchema, path, newErrorSchema, true);
+          setByPath(mergeBaseErrorSchema, path, raisedErrorSchema, true);
```
heath-freenome: The emptied-branch drop you added covers a child of the raise, but not the raise's own node: the recursion returns `{}` and `setByPath` writes it here at `path`. So `{addr:{street:{}}}` is still stored, and it still reads as a validator error to the `oldValidationError` test one level up.

Repro (live validation off, schema `{ type: 'object', properties: { addr: { type: 'object', properties: { street: { type: 'string', minLength: 3 } } } } }`, `formData={{ addr: { street: 'a' } }}`):

1. a field raises the validator's own text as its own error before anything has validated -- `formRef.current.onChange('a', ['addr', 'street'], { __errors: ['must NOT have fewer than 3 characters'] })`. No validator error on the path yet, so it lands in `customErrors`.
2. submit. `schemaValidationErrorSchema` is `{addr:{street:{__errors:['must NOT have fewer than 3 characters']}}}` and the field shows one copy, since `customErrors` merges with `'preventDuplicates'`.
3. the field hands back what it displays -- `formRef.current.onChange('a', ['addr', 'street'], getByPath(state.errorSchema, ['addr', 'street']))`. This is the store branch, `suppliedHere` carries the `customErrors` copy, and the one copy the display had is removed.

| Commit | `schemaValidationErrorSchema` | `schemaValidationErrors` |
| --- | --- | --- |
| `4d1cb92af` (main) | `{addr:{street:{__errors:['must NOT have fewer than 3 characters']}}}` | the error (displayed twice -- the duplication this PR fixes) |
| `aa17758d0` | `{addr:{street:{}}}` | `[]` |

The field still shows the message, because `customErrors` re-merges it, which is the caveat you already called out. The two parts that are new: the stored pair has lost a validator error that is still real, and the phantom does what the drop was meant to prevent. A raise on `['addr']` next -- `onChange({ street: 'a' }, ['addr'], { __errors: ['addr-level custom'] })` -- sees `Object.keys({street:{}}).length === 1`, takes the store branch instead of `customErrors`, and replaces `schemaValidationErrorSchema.addr` wholesale. On `main` that raise is a custom error that survives; stored, it goes at the next validation pass or changed-field clearing.

Treating an emptied raise as no raise at all covers both branches and reads as what it means -- the field had nothing of its own to say here:

```ts
const raisedErrorSchema = withoutSuppliedErrors(newErrorSchema, [...]);
if (Object.keys(raisedErrorSchema).length > 0) {
  // setByPath + the list rewrite + storedValidation
}
```

That leaves the validator's entry standing, which is `main`'s result without the duplication. The `customErrors` branch at line 1334 writes the same `{}` through `setByPath`; there it clears the node, which is right, but it leaves the empty object behind rather than removing it.

CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ should change the heading of the (upcoming) version to include a major version b
...
...(a `className` change, a new `extraErrors` object); the raise now survives those derivations, though not a controlled parent echoing the change that raised it, which the changed-field clearing still treats as new data for that field. The stored error list carries the raise too, so the `ErrorList` says what the field does; a raise at the root, which replaces every displayed error, no longer replaces the stored validator result, which used to drop every other field's error for good; and a raise that hands back the displayed `errorSchema`, as the optional-data Add and Remove buttons and the fallback field's type change do, is stored without the `extraErrors` and `customErrors` it carries, which would otherwise stay after the parent cleared them; the item errors an `ArrayField` remaps after a reorder, remove or copy are displayed but not stored, since at their new indexes they no longer line up with the `extraErrors` to leave out. Under `liveValidate: 'onBlur'` a controlled form now clears a field's error on the first keystroke while an uncontrolled form keeps it until the field is left, as with live validation off ([#5306](https://github.com/rjsf-team/react-jsonschema-form/pull/5306))
```
heath-freenome: "clears a field's error on the first keystroke" does not cover everything the clearing now takes from an `'onBlur'` form. A reorder is not a keystroke and the item's value does not change, but the parent's echo makes both indexes changed data, and the clearing drops the remapped errors `ArrayField` just went to the trouble of moving.

Verified, controlled parent echoing `onChange`, `liveValidate='onBlur'`, schema `{ arr: { type: 'array', items: { type: 'string', minLength: 3 } } }`, `formData={{ arr: ['a', 'bbbb', 'cccc'] }}`: blur item 0 so its error shows, then move it down.

| Commit | After the reorder |
| --- | --- |
| `4d1cb92af` (main) | the error follows the item to `arr.1`, in the field and the list |
| `aa17758d0` | gone from the field, the `ErrorList` and `state.errors`; only blurring that item or submitting brings it back |

An uncontrolled `'onBlur'` form on this branch keeps it, and live validation off loses it on both commits -- `should keep the errors that a reorder of the array items left standing` pins that down -- so this is `'onBlur'` lining up with `'off'` rather than a defect. It is still a user-visible change from `main` that the bullet does not describe. Something like "when the parent hands back changed data for that field, whether a keystroke or a reorder, remove or copy" would say it.

heath-freenome (commented):
Follow-up on the four commits since the last round. Each fix does what its reply says -- I ran every commit's new tests against its own parent's `src` and all of them fail without their fix (8 for `9b04bc2a5`, 4 for `8044a21c8`, 3 for `3f4f0fb0f`, 2 for `aa17758d0`), and head is green on the core suite (2076), the full monorepo, `lint`, `cs-check` and the root `typecheck`. Three things left, none of them blocking: a third place a raise is stored that the helper does not reach, the emptied-raise node that is still written, and a changelog wording point about what the clearing takes from an `'onBlur'` form.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed: after the root raise, `customErrors.ErrorSchema` held `{__errors:['server']}` and `server` stayed after clearing `extraErrors` and restyling. Fixed in 0f57cb58a as you suggested: the root branch now stores `withoutSuppliedErrors(newErrorSchema, [extraErrors])[ERRORS_KEY]`. So all three places a raise is stored now leave out the supplied errors, as the changelog bullet says. Test: "lets the parent clear the root extraErrors a root-path raise sent back". It fails on aa17758d0 with `['server']`. I left the nested part of a root raise being dropped alone, as you said.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed: with your repro, `schemaValidationErrorSchema` ended up as `{addr:{street:{}}}`. Fixed in 0f57cb58a your way. In the store branch, a raise that is empty once the supplied errors are removed now counts as no raise, so the validator's entry stays and nothing gets written or stored. In the `customErrors` branch, an emptied raise now removes the node with `unsetByPath` instead of leaving `{}`. Test: "keeps the validator error when a raise hands back only errors supplied elsewhere". It asserts that the stored schema and list still hold the validator error, and it fails on aa17758d0 with `{addr:{street:{}}}`.

One note on the follow-up in your repro: with the validator's `street` error still in place, a raise on `['addr']` now takes the store branch, because a real validator error sits below that path. That is the same routing `main` would use for it, so I didn't add a test for that step.

CHANGELOG.md
jimmycallin: Agreed. 0f57cb58a rewords the bullet: under `'onBlur'`, a controlled form now clears a field's error whenever the parent hands back changed data for that field. That covers a keystroke and also an `ArrayField` reorder, remove or copy, which drops the item errors the array remapped. The same bullet now also says that a raise left with only supplied errors leaves the validator's error standing.

heath-freenome:
Oh interesting... I totally missed that we kept this on the main branch. How much of this PR survives if we move fixes onto v7?

jimmycallin:
About a third of it. It's #5297's v6 backport, and #5297 plus #5312 already brought most of it to `v7`. To measure it rather than guess, I ran this PR's 17 new tests against `v7` at `377683219`. **11 pass and 6 fail**, and all 6 fail on assertions, not on v6/v7 API differences.

**Already on `v7`, nothing to move (11 tests):**
- `'onBlur'` doesn't validate when the data is replaced from outside (the mode predicate, from #5297).
- Live validation on a derivation keeps `extraErrors`.
- `extraErrors`/`customErrors` are no longer merged twice, on all three paths (#5297 and #5312's #5041 follow-up).
- A schema change drops the stored validator results (#5312's `reconcileErrors()`).
- Most of the supplied-errors cases (ArrayField remap, optional-data Remove over a validator error, the validator's own copy, the emptied raise). These pass on `v7` because a raise on a path with a validator error is only applied to a throwaway merge base and never stored, so there is nothing to strip. That same shortcut is why the first item below is lost on the next derivation.

**Still missing on `v7` (6 tests), in two groups:**
1. **Where a raise is stored** (`9b04bc2a5`, `aa17758d0`, `0f57cb58a`):
   - A custom error raised on a path that already carries a validator error is lost on the next derivation.
   - A root-path raise replaces every other field's validator error.
   - A raise that lands in `customErrors` keeps the `extraErrors` it handed back, at both a nested path and the root, so a server error outlives its prop.
2. **The `'onBlur'` changed-field clearing only reaches the field, not the list** (`e4a3edada`, `9b04bc2a5`): it clears `errorSchema` but leaves `errors`, so the `ErrorList` keeps an error the field no longer shows. That's the same list/field split the #5312 code review flagged, and it happens on `v7` for live validation off too.

As I wrote https://github.com/rjsf-team/react-jsonschema-form/pull/5306#issuecomment-5748565625 I'm happy to close this one.

heath-freenome:
I'd say close this and make sure the missing tests (and any fixes) end up in v7 somehow.

jimmycallin:
Closing in favour of landing on v7 only: the missing regression tests, and the fixes five of them needed there, moved to #5341.

PR #5310, opened 2026-09-20
test(core): assert Form errors through the DOM and callbacks instead of the instance state

jimmycallin:
Test-only, three commits. Base: `v7`. No source change, no snapshot change. Prepares the `Form` test suite for the function-component conversion the [form data ownership RFC](https://github.com/jimmycallin/react-jsonschema-form/blob/rfc/form-state-ownership/rfcs/form-state-ownership-spec.md) schedules after the ownership switch; it does not depend on any other open PR and can merge first.

## 1. Why the tests have to change before the component does

A function component has no `this.state`. Whatever `Form` keeps internally after the conversion is held in hooks and is not reachable from a `ref`; the ref exposes `FormHandle` and nothing else. That is already the documented contract: the v7 upgrade guide's "Reading `Form` internals through a `ref`" table tells consumers to stop reading `.state.errors`, `.state.errorSchema` and `.state.schemaUtils` and to use the `onChange`/`onSubmit` event or `getFormData()` instead.

The test suite did not follow its own advice. About 40 assertions in `Form.behaviors.test.tsx` and `Form.errors.test.tsx` read `formRef.current!.state.*`, and three tests `vi.spyOn(formRef.current!.state.schemaUtils, 'omitExtraData')`. Every one of them would break in the conversion PR, and a conversion PR is the worst place to be rewriting assertions: when a test fails there, the reader has to decide whether the component or the assertion is wrong. Moving the assertions to observers that survive the conversion, while the class is still in place and every test is known-good, separates the two questions. This PR is that move, and it is green against the class.

## 2. What replaces the state reads

The witnesses a function component leaves available are the DOM, the callbacks and `FormHandle`. Two small helpers in `testUtils.tsx` make the DOM one readable:

- **`fieldErrorsById(node)`** returns the messages each `FieldErrorTemplate` renders, keyed by the id of the field the list sits under (`{ root_foo: ["must have required property 'foo'"] }`). It is the observable form of `state.errorSchema`: the same tree, projected onto the fields that display it.
- **`errorListMessages(node)`** returns what the top or bottom `ErrorList` renders, in order. It is the observable form of `state.errors`. Note it returns each error's `stack` (`.name from the server`), which is what the template prints, not the bare `message`; the assertions say so explicitly rather than hiding it in the helper.

Per assertion, the mapping is mechanical:

| Was | Now |
| --- | --- |
| `state.errorSchema` equals a tree | `fieldErrorsById(node)` equals the per-field messages |
| `state.errors` equals a list / has length N | `errorListMessages(node)` equals the rendered list / has length N |
| `state.errors` after a failed submit, full object shape | `onError` was last called with that exact list (the existing mock from `createFormComponent`) |
| `state.retrievedSchema.properties.a.title === 'A2'` | dropped: the same test already asserts the rendered label reads `A2` |
| `state.retrievedSchema` lacks `water` after a dependency switch | `#root_water` is not rendered and `#root_food` offers only the new branch's options |

Where a test asserted state and DOM side by side (`.error-detail` count plus `state.errorSchema`), the DOM assertion stays and the state one goes.

## 3. Three tests deleted, not rewritten

The three `vi.spyOn(state.schemaUtils, 'omitExtraData')` tests asserted that the omit function was or was not *called*. Each has an effect-based twin a few lines down that asserts the omitted data itself: "should omit data on change with omitExtraData=true and liveOmit=true", "should not omit data on change with omitExtraData=true and liveOmit unset", and "Should call validateFormWithFormData with a new formData with only used fields if omitExtraData is true". A spy on an internal object proves less than the data it produces, and the twin already proves that, so the spies are deleted rather than re-pointed. Test count goes from 2150 to 2147 for that reason alone.

## 4. Two tests keep reading `state`, on purpose

`should clear the error of a field whose name contains a dot` and `should only clear the error of the field that changed under a name containing a dot` (both in `Form.errors.test.tsx`) are unchanged, with a comment saying why. They pin that `toErrorSchema` nests a property named `foo.bar` at `foo → bar`, and that the changed-fields clearing pass reaches the same place. Rewriting them surfaced that nothing but `state` can see this:

- The entry they watch is cleared by the *prop update's* changed-fields pass, after an accepting parent hands the data back, so no `onChange` carries the cleared `errorSchema`.
- The field cannot display it either: a field named `foo.bar` looks its errors up by name (`errorSchema['foo.bar']`) while `toErrorSchema` has nested them by path, so the message never reaches the DOM. That is a pre-existing mismatch, out of scope here, and worth its own issue.

Until that lookup is fixed or the clearing pass gets a unit seam of its own, the state read is the only honest witness, and these two are the whole remaining inventory. The conversion PR will have to deal with exactly them and nothing else.

## 5. Commit 2: cover the `FormHandle` surface the conversion keeps

With the state reads gone, the question was what a function-component `Form` *does* expose and whether that surface is tested. `FormHandle` is the contract, and `Form.tsx` had no test at all for `submit()`, none for `focusOnError()`'s fallback paths, and none for `setFieldValue()` clearing an array item. Those are pinned now, through the handle, in `Form.handle.test.tsx`:

- **`submit()`** reaches `onSubmit` with `status: 'submitted'` for valid data and `onError` for invalid data. The invalid case sets `noHtml5Validate`, because `requestSubmit()` runs the browser's constraint validation before any submit event fires, so a `required` input that is empty never reaches the form otherwise. A form with `tagName='div'` cannot be submitted programmatically, which is the documented cost of a non-form `tagName`, and an unmounted form ignores the call.
- **`focusOnError()`** focuses the input the error names; falls back to the first control whose id starts with the field id when the widget has no named input, which is the shadcn button-group case the source comments describe; treats a missing `property` as the root field; and does nothing for an unmatched property or after unmount.
- **`setFieldValue()`** clears an array item to `null` when addressed by its numeric index, which is what the method's TSDoc promises, and clears a leaf inside an item of a root-level array without touching its siblings.
- **`showErrorList: 'bottom'`** renders the error list after the fields (`validate.test.tsx`).

`createComponent()` in `testUtils.tsx` now also returns `unmount`, for the two unmounted-handle cases.

Coverage of `Form.tsx` (`vitest --coverage --coverage.include=src/components/Form.tsx`):

| | Before | After |
| --- | --- | --- |
| Lines | 97.1% | 100% |
| Functions | 97.4% | 100% |
| Branches | 95.3% | 99.3% |

The two branches left are a defensive `errorSchema || {}` in `renderErrors()` and the array half of the copy in the cleared-leaf path, which `findFieldInSchema()` never reaches through a numeric index. Neither is worth a contorted test, and both are lines the RFC's later PRs rewrite anyway.

## 6. Commit 3: the two remaining class-shaped dependencies

Two more things in the suite were about the class rather than the component, and both would have failed in a conversion PR for reasons unrelated to behavior:

- **`createRef<Form>()`, 36 sites in seven files.** `Form` in type position is the class's instance type. A function component has none, so every one of these lines would have needed the same edit. They call `createFormRef()` from `testUtils.tsx` now, the single place that names the type; the conversion changes that one line to `createRef<FormHandle>()`.
- **`vi.spyOn(formRef.current!, 'validateFormWithFormData')`, four tests.** They spied on the instance method to see what data a submit or `validateForm()` validated. A hook-built handle is an object the component *returns*, not the function it *calls*, so after the conversion the spy is never hit and the assertion is silently vacuous. The submit pair now asserts the data `onSubmit` receives with and without `omitExtraData`. The `validateForm()` pair uses `additionalProperties: false`, so whether the extra field was validated shows in the boolean result and in `onError`.

## 7. What a conversion PR will still have to touch in the tests

For the record, so the conversion's author knows the inventory is complete:

| Site | Why | Size |
| --- | --- | --- |
| `createFormRef()` in `testUtils.tsx` | `RefObject<Form>` becomes `RefObject<FormHandle>` | 1 line |
| `expectTypeOf<Form>().toExtend<FormHandle>()` in `Form.handle.test.tsx` | the type contract test names the class on purpose; it becomes `ComponentRef<typeof Form>` | 1 line |
| The two dotted-name tests in `Form.errors.test.tsx` | the only `.state` reads left; see section 4 | 2 tests, or a fix to dotted-name error lookup first |
| `IChangeEvent.test.ts` | asserts the event's keys against `FormState`; survives if the state type is kept as the pipeline's type | maybe 0 |

Everything else asserts through the DOM, the callbacks or `FormHandle`, none of which the conversion changes.

## 8. Verification

- `pnpm --filter @rjsf/core test`: 44 files, 2159 tests, green against the current class `Form` (2150 before: three spy tests removed, twelve handle and rendering tests added).
- Root `lint`, `typecheck` and `cs-check` pass; no source file changed, so `build`, `knip` and the snapshot suites are unaffected.


heath-freenome:
## Review

Scope: `git diff v7...pr5310` (merge base `40a9e8c8`), 9 test files, no source changes. Checked out the branch and ran the full `@rjsf/core` suite (44 files / 2159 tests green), trial-merged current `v7` (2202 green), and ran `oxlint --type-aware`, `tsc --build packages/core/test`, and `oxfmt --check` on the changed files — all clean. Then mutation-tested the rewritten assertions to check whether they still have teeth.

### Verified

- The only remaining `.state` reads in `packages/core/test` are the two documented dotted-name tests in `Form.errors.test.tsx` (lines 160/167 and 202/212) — the inventory in the description is complete.
- The three deleted `omitExtraData` spy tests have live twins: mutating `omitExtraData === true && liveOmit === 'onChange'` to `omitExtraData === true` fails 3 tests.
- The `validateForm()` rewrite genuinely covers what it replaced: deleting the omit branch in `validateForm` fails _"validates only the fields the schema describes if omitExtraData is true"_.
- The `toEqual([])` / `toEqual({})` assertions in the `Async errors` resets are not vacuous — adding `expect(errorListMessages(node)).not.toEqual([])` before each rerender passes, so the lists really are populated first.
- Dropping `expect(node.querySelector('.errors')).not.toBeInTheDocument()` in _"Clear errors"_ loses nothing: forcing `renderErrors()` to always render fails 149 snapshot tests.

### Findings

**1. `packages/core/test/Form.behaviors.test.tsx:624` — medium — the rewritten `omitExtraData on submit` pair drops coverage the deleted spy tests had.**

The old tests asserted *which* data was handed to `validateFormWithFormData`; the new ones only assert what `onSubmit` receives. The schema here (`{ properties: { foo } }`, no `additionalProperties: false`) validates `{foo:'bar',baz:'baz'}` and `{foo:'bar'}` identically, so the submit-time validation input is now unobserved.

Concrete check: changing `Form.onSubmit` line 1223 to `this.validateFormWithFormData(this.state.formData)` — i.e. validating the un-omitted data while still submitting the omitted data — leaves the entire core suite green (2202 passed), whereas the deleted _"Should call validateFormWithFormData with a new formData with only used fields if omitExtraData is true"_ fails on it.

Suggested fix: give this `describe` the same `additionalProperties: false` schema the rewritten `validateForm()` pair uses (~line 1300), so validating the extra field surfaces as an error / `onError` and the assertion regains teeth.

**2. `packages/core/test/Form.behaviors.test.tsx:1141` — low — dead `formRef`.**

In _"re-resolves the retrieved schema when customMergeAllOf changes"_, the only reader of `formRef` was the deleted `state.retrievedSchema` assertion. `const formRef = createFormRef()` and `ref={formRef}` are now dead setup — nothing fails, but it leaves a `createFormRef()` site that a later sweep has to re-examine for no reason. Delete the ref and the prop.

**3. `packages/core/test/Form.errors.test.tsx:1058` — low — same leftover.**

In _"should sanitize stale enum data and persist the retrieved dependency schema"_, `formRef` is created and passed as `ref` but never read now that the `retrievedSchema` assertions go through the DOM. Delete it.

**4. `packages/core/test/testUtils.tsx:173` — low — `errorListMessages()` silently returns `[]` for a non-default error list.**

It hard-codes the Bootstrap-3 `ErrorList` markup (`.panel.errors li`) and returns `[]` for any other template, with no signal that it found no list at all. The suite already has a custom `ErrorListTemplate` test (`validate.test.tsx` _"should use CustomErrorList"_), so a future test combining a custom error list with `expect(errorListMessages(node)).toEqual([])` would pass for the wrong reason. Worth a line in the TSDoc stating the helper assumes the default template, or keying it off the template's container rather than the theme classes. `fieldErrorsById()` has the same coupling to `ul.error-detail`, but its assertions are all positive so it fails loudly.

jimmycallin:
Rebased onto `v7` at `d296f94f1` (one conflict in "Clear errors": kept v7's `'5'` input for the numeric `pattern` alongside the `errorListMessages(node)` assertion), then addressed the four findings in `61cb98ddf`.

1. **`omitExtraData on submit`.** The pair now uses the same `additionalProperties: false` schema as the `validateForm()` pair. With `omitExtraData: false` the extra field is validated, so `onError` receives an `additionalProperties` error and `onSubmit` is not called; with `true`, `onError` is not called and `{ foo: 'bar' }` is submitted. Re-ran your mutation (`this.validateFormWithFormData(this.state.formData)` in `Form.onSubmit`): the `true` case fails on it now, where the suite was green before.
2. **Dead `formRef` in the `customMergeAllOf` test.** Removed along with the `ref` prop.
3. **Dead `formRef` in "should sanitize stale enum data".** Same.
4. **`errorListMessages()`.** Its TSDoc now says it assumes the default Bootstrap 3 `ErrorList` markup and returns `[]` under a custom `ErrorListTemplate`.

`@rjsf/core` 2266 tests pass; root `typecheck`, `lint`, `knip` and `cs-check` clean.

heath-freenome:
## Follow-up review of `9fa44e58e`

Thanks, all four findings are addressed. I checked out the rebased branch (on `87e9bf279`) and verified each fix:

- **`omitExtraData on submit`.** I broke `Form.onSubmit` three ways, and a test now fails for each one:
  - validating `this.state.formData` instead of the omitted data fails the `true` test;
  - always omitting, even when `omitExtraData` is `false`, fails the `false` test;
  - passing the un-omitted data to `onSubmit` fails the `true` test.
- **Rebase conflict in "Clear errors".** The resolution looks right. If `reset()` keeps the old errors instead of clearing them, the test fails.
- **Dead `formRef`s.** Both are removed.
- **`errorListMessages()` TSDoc.** The new note covers it.

All 2266 `@rjsf/core` tests pass, and `lint`, `typecheck` and `cs-check` are clean.

**Optional nit:** "Clear errors" checks only that one error appears (`errorListMessages(node)).toHaveLength(1)`), not what it says. Asserting on the `minimum` message would prove the error comes from RJSF's validation of `minimum: 100` and not from something else. `v7` also checked only the length, so this is not a regression.

Nothing blocking from my side.

heath-freenome (approved):

PR #5312, opened 2026-09-20
refactor(core): derive Form state with pure functions (ownership RFC, A2 + A3)

jimmycallin:
Lanes A2 and A3 of the [form data ownership RFC](https://github.com/jimmycallin/react-jsonschema-form/blob/rfc/form-state-ownership/rfcs/form-state-ownership-spec.md). Base `v7`, after #5297 (A1), #5298, #5299 and #5300. The ownership switch (A4) is the next PR and depends on this one.

## Why this change exists

`Form` holds two kinds of state, and today it computes both in the same place.

**Derived state** is a pure function of the props and the data: the `schemaUtils` built from the validator and schema, the registry built from the theme props, the schema resolved for the current data, the defaults merged into it, the validation errors for it. Same inputs, same value. It is a cache of a computation.

**Owned state** is the value nobody else holds: the data of a self-owned form, the custom errors fields have raised, the queue of pending changes.

The RFC gives the form data exactly one owner. For a parent-owned (controlled) form that owner is the parent, which means a change becomes a *proposal* the parent may accept, transform or reject — and the same computation has to serve both modes:

```ts
const next = applyChange(current, change, props);
if (controlled) notify(next);        // a proposal; the parent decides
else commitAndNotify(next);          // the form's own value
```

A method that reads `this.state` and calls `setState` cannot be that computation. For a parent-owned form there is no `this.state.formData` to read — the prop *is* the current value — and nothing to `setState`, because the form does not decide. So the computation must first become a function of explicit arguments.

That is the whole of this PR. Today it is `getStateFromProps(props, inputFormData, …six positional flags)` reading `this.state` for each previous value, and `processPendingChange()` writing a field's value, re-deriving, committing and notifying in one method body that also reads the queue. Afterwards those are module-level functions that take everything they need and return the next state. Nothing about ownership changes yet; A4 then reads as a removal plus a switch rather than a refactor.

## What the code looks like afterwards

```ts
resolveSchemaUtils(props, prev)                             → { schemaUtils, hasNestedConditionalSchema }
deriveRenderContext(props, retrievedSchema, prev, resolved) → RenderContext
deriveFormData(current, inputFormData, options, props)      → { formData, context, … }
deriveFormState(current, inputFormData, options, props)     → FormState
applyChange(current, change, props, deferLiveValidate)      → FormState
applyReset / applyBlur / applySubmit / applyValidation      → FormState
```

`RenderContext` is the derived state rendering needs — `schemaUtils`, the root `schema`, the `uiSchema`, the `retrievedSchema`, `hasNestedConditionalSchema`, `registry` — typed as a `Pick` of `FormState`, so the state's shape is unchanged.

Two design points are worth the reviewer's attention, because both replace an invariant that used to live in a comment:

- **`deriveFormData` returns the context, not its parts.** A resolved schema is only meaningful next to the utilities that resolved it. Returning them together means state can never hold a `retrievedSchema` resolved by utilities it does not also hold; the type enforces what four call sites would otherwise each have to remember.
- **`validateFormData`/`runLiveValidation` take that context**, rather than a loose `schemaUtils` + `schema` + `registry` trio. Those three are always fields of one context, and passing them separately is what allowed a mismatched combination.

Each handler is now commit-plus-callback around a pure function — which is exactly the shape a hook body needs, so the later function-component conversion reroutes handlers instead of re-deriving them.

**Custom errors become copy-on-write.** `applyChange` used to edit the committed `ErrorSchemaBuilder` in place. A pure function may not mutate its input, and in A4 a rejected proposal must leave committed errors untouched. It now edits a copy.

**No memoization.** An earlier draft added per-step caches keyed by each step's inputs. They are gone and the spec is updated: under the reconciler they almost never hit, the cost they insure against is unmeasured until A4, and each key was a hand-maintained input list that goes stale silently. Once `Form` is a function component every one of them is a `useMemo` the hooks lint rule checks.

## Breaking changes

Two, both in public API surface. Each is in `CHANGELOG-v7.md` and has its own section in the v7 upgrade guide.

**1. Clearing a custom error now also removes it from the top-level error list.**
When a field raised a custom error through `onChange(value, path, errorSchema)` and later cleared it, the field-level message disappeared but `ErrorListTemplate` kept showing it until the next validation. That was an aliasing accident: `validationDataMerge` returns the additional schema *by reference* when the base is empty, so committed `state.errorSchema` *was* the builder's object, `clearErrors()` mutated it in place, and `state.errors` was never rebuilt. With the builder copied, both the field and the list drop the entry. Anyone asserting the stale list entry will see a test fail.

**2. `noValidate` now also turns off live validation on blur.**
`liveValidate='onBlur'` validated and displayed errors regardless of `noValidate`, where `liveValidate='onChange'`, the derivation and submission all honoured it. A form relying on that combination to validate will stop doing so.

**Not breaking:** `getStateFromProps()` is gone from the `Form` instance. It was undocumented, never part of `FormHandle`, and the upgrade guide already states that anything on the instance outside `FormHandle` is an implementation detail. Noted in the changelog, not in the guide.

## Other behavior changes

All three are errors that should not have been on display:

- **`extraErrors`/`customErrors` no longer accumulate.** `applyChange`'s custom-error branch re-merged them onto `current.errors`, which already carries them, while starting the `errorSchema` side from `schemaValidationErrorSchema`. With validation errors on display, every change raising a custom error appended another copy of each — four copies of one message after a single keystroke in the added regression test. This is the other half of the asymmetry #5041 fixed.
- **A schema change clears the validator's own results too.** They were carried forward, so a later `extraErrors` change rebuilt the display from the *previous* schema's errors through `getDerivedStateFromProps()`.
- **Live validation on a prop-driven change validates against the root schema.** A resolved schema has had the root's `if`, `dependencies` and `$ref`s folded into it, so validating against it drops the errors those keywords report. It may stand in for the root only while it still describes the data — which the schema the committed state is already rendering does, and one this pass resolved does not, since the prop change may be the schema itself.

Not changed: `onChange` timing and arguments, `reset()`, submit, the controlled snap-back, `edit`, `initialDefaultsGenerated`, and the shape of `FormState`.

## Not in this PR

- The mode resolver, the reconciler removal, the controlled queue checkpoint and the development diagnostics: A4.
- Moving the pure functions into their own module — it would turn a reviewable extract-in-place into a delete-and-add, and A4 edits them wherever they live.
- Two `structuredClone` calls on the change path. Both are the clone-then-`setByPath` shape the new `copyAlongPath()` replaces locally. The right fix is an immutable `setByPath` variant in `@rjsf/utils/pathUtils` retiring all three sites, which is a public API addition under that package's 100% coverage bar, and whose correctness becomes conditional on every downstream consumer treating the data as read-only rather than unconditional as a clone is. Its own PR; the cost is unchanged from `v7` today.

## Verification

Root `lint`, `knip`, `build`, `typecheck`, `cs-check` and `pnpm test` (all 15 packages) pass, and **every one of the 9 commits typechecks on its own**. `@rjsf/core`: 2156 tests across 45 files. Zero snapshot updates in any theme suite — the working tree is clean after a full run. `renderStability.test.tsx` counts are unchanged, with no loosening.

New tests: `Form.derivation.test.tsx` (a replaced `customValidate` takes effect on the next change; a changed `idPrefix` rebuilds the registry and the ids), plus regressions for breaking change 2 and for the `extraErrors` accumulation — both verified to fail when their fix is reverted.

## Review guide

`Form.tsx` is the whole diff apart from the changelog, two upgrade-guide sections and the tests. Read it top to bottom: `RenderContext`, `resolveSchemaUtils`, `deriveRenderContext`, `mergeErrors`, `validateFormData`, `runLiveValidation`, `deriveFormData`, `deriveFormState`, `copyAlongPath`, `applyChange`, the four handler functions, then the class, where the diff is deletions and call-site rewiring.

The commits are ordered to be read in sequence: three extraction commits, then six small self-contained fixes for things the extraction surfaced. `applyChange` is best read beside `v7`'s `processPendingChange` — the logic is a move, and the real diff is the copy-on-write builder.


packages/core/src/components/Form.tsx
```diff
@@ -345,10 +346,761 @@ const VALIDATION_PROP_KEYS: ReadonlySet<(typeof IDENTITY_PROP_KEYS)[number]> = n
...
+  options: DeriveOptions,
+  props: FormProps<T, S, F>,
+): FormState<T, S, F> {
+  const { isSchemaChanged = false, getFormDataChangedFields, skipLiveValidate = false } = options;
+  const { liveValidate } = props;
+  const edit = inputFormData !== undefined;
+  // `'onBlur'` owns its validation pass in `onBlur()`; deriving state must not run one for it, or the errors show
+  // up before the field the user is editing has been left. Skipping means the state already holds this pass's errors
+  // oxlint-disable-next-line typescript/no-deprecated
+  const mustValidate = edit && !props.noValidate && liveValidate === 'onChange' && !skipLiveValidate;
```
heath-freenome: Folding `skipLiveValidate` into `mustValidate` changes what the changed-field clearing is gated on, and a controlled parent that echoes `formData` back now loses the inline error.

`reconcileErrors()` runs the changed-field clearing whenever `mustValidate` is false. `getStateFromProps()` gated that clearing on `!mustValidate` computed *without* `skipLiveValidate`, so the "live validation already ran for this data" case skipped it. Here the two are the same flag, so `skipLiveValidate === true` both suppresses the validation pass and enables the clearing.

Repro (fails on this branch, passes on `f2657ee2a`): controlled form whose parent accepts and echoes `formData`, `liveValidate='onChange'`, schema `{ a: { type: 'string', minLength: 5 } }`. Type `ab`:

1. `applyChange` validates and commits the error.
2. The parent echoes the value back, so `isStateDataChanged` is false and `skipLiveValidate` is true.
3. `getChangedFields(formData, prevProps.formData)` returns `['a']`, and the entry is deleted from `schemaValidationErrorSchema`.

Result: `ErrorListTemplate` shows "must NOT have fewer than 5 characters" while the field shows nothing — `.error-detail li` count 0, was 1.

Passing `getFormDataChangedFields` only when `!(edit && !noValidate && liveValidate === 'onChange')` — i.e. keeping the old gate separate from the validate-or-not decision — makes the repro pass again.

packages/core/src/components/Form.tsx
```diff
@@ -345,10 +346,761 @@ const VALIDATION_PROP_KEYS: ReadonlySet<(typeof IDENTITY_PROP_KEYS)[number]> = n
...
+  const { schema, schemaUtils, customErrors, retrievedSchema } = current;
+  const formData =
+    omitExtraData === true && liveOmit === 'onBlur'
+      ? schemaUtils.omitExtraData(schema, current.formData)
+      : current.formData;
+  const validation =
+    liveValidate === 'onBlur' && !noValidate
+      ? runLiveValidation(props, current, formData, customErrors, retrievedSchema)
+      : undefined;
+  return { ...current, formData, ...validation };
```
heath-freenome, line 1033: Returning `{ ...current, ... }` and committing the whole thing clobbers any state update queued between the `this.state` read and the `setSharedState` call.

`current` is the snapshot taken when the handler started, and `setSharedState` commits every key of it. The old handlers committed only the keys they owned, so React's partial merge preserved a concurrent update. This applies to all four — `applyChange`, `applyReset`, `applySubmit`, `applyValidation` — and the call site at line 1404 is where it bites hardest.

Repro (fails on this branch, passes on `f2657ee2a`):

```tsx
<Form omitExtraData liveOmit='onBlur' onBlur={() => formRef.current.validateForm()} />
```

`Form.onBlur` calls the user's `onBlur` first, so `validateForm()` queues its error state; then `applyBlur(committed, props)` — where `committed` predates that queue entry — is committed in full and the errors disappear. `.panel-danger.errors` is absent; `v7` renders it. The old code committed just `{ formData }` here.

The pure-function shape itself is fine; it's the commit that needs to stay narrow — either return only the keys that changed, or diff against `this.state` at commit time rather than against the captured `current`.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
heath-freenome, line 1: Stale type reference further down this file, outside the diff: the "`getFormData()` and the `FormHandle` type" section (around line 548) still says

> `FormHandle<T, S, F>` names every supported ref member

but this PR drops the `S` and `F` parameters, and the new text at line 28 correctly says the type is now `FormHandle<T>`. A reader who copies the later form into their own ref type gets a TS error.

heath-freenome (commented):
Reviewed the PR's own commits (`f2657ee2a..9912ba498`, excluding the stacked #5310 test commits). The extraction reads well and the commit ordering makes it followable — two behavioral regressions to sort out, both reproduced against this branch and confirmed absent on the base in the same environment, plus one stale doc reference. Details inline.

1. **`Form.tsx:755`** — folding `skipLiveValidate` into `mustValidate` makes a controlled parent's echo clear the field-level live-validation error while the error list keeps showing it.
2. **`Form.tsx:1031` / `:1404`** — the `apply*` handlers commit a whole-`FormState` snapshot taken at call time, clobbering a state update queued in between (`omitExtraData` + `liveOmit='onBlur'` + an `onBlur` that calls `validateForm()`).
3. **v7 upgrade guide** — a later section still documents `FormHandle<T, S, F>`.

Checked and clean, not filed:

- `copyAlongPath()` and `ErrorSchemaBuilder`'s `structuredClone` constructor do make the copy-on-write claims hold.
- The three breaking changes (custom-error clearing, `noValidate` + `liveValidate='onBlur'`, the narrowed `FormHandle.validate()`) are in `CHANGELOG-v7.md` and the upgrade guide, and no in-repo caller passes the removed `validate()` arguments.
- One small doc/behavior mismatch worth a sentence rather than a change: `reconcileErrors()` also resets `schemaValidationErrors`/`schemaValidationErrorSchema` when `props.noValidate` is set, not only on a schema change as the changelog describes. Displayed errors were already cleared in that case on `v7`, so only the private fields differ.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed and fixed in `cbcca3ca4`. `deriveFormState()` now computes `isLiveValidated` (`edit && !noValidate && liveValidate === 'onChange'`) separately: `mustValidate` is that flag minus `skipLiveValidate`, and `getFormDataChangedFields` is handed to `reconcileErrors()` only when the form is not live-validated, so a skipped pass leaves the committed errors alone as `getStateFromProps()` did.

Regression test: `keeps the inline error of the field being typed in when the parent echoes the data back` (controlled echoing parent, `liveValidate='onChange'`, `minLength: 8`, one keystroke); it asserts both the inline `.error-detail` entry and the `ErrorList` message, and fails on `9912ba498` with the inline list empty.

packages/core/src/components/Form.tsx
jimmycallin: Confirmed and fixed in `cbcca3ca4`. `setSharedState()` now takes the state the handler started from alongside its result and commits only the keys where the two differ, inside the `setState` updater, so React's partial merge keeps anything queued in between. The `apply*` functions keep their whole-state shape; the narrowing happens once at the commit rather than in each of them. All seven call sites pass their starting state (`this.state`, the captured `committed` in `onBlur()`, `prevState` in `componentDidUpdate()`).

Regression test: `keeps the errors that validateForm() queued from inside onBlur while liveOmit runs on blur`, your repro (`omitExtraData`, `liveOmit='onBlur'`, `onBlur` calling `validateForm()`); it asserts the `ErrorList` and the inline error and fails on `9912ba498` with both empty.

packages/docs/docs/migration-guides/v7.x upgrade guide.md
jimmycallin: Fixed in `cbcca3ca4`: the later section now reads `FormHandle<T>`.

jimmycallin:
On the `reconcileErrors()` aside: the `CHANGELOG-v7.md` bullet now says `noValidate` clears `schemaValidationErrors`/`schemaValidationErrorSchema` the same way a schema change does, where it already cleared the displayed errors. The two regressions and the stale `FormHandle<T, S, F>` are fixed in `cbcca3ca4`, with a regression test for each; details in the threads.


Re-stacked on #5310 at `61cb98ddf` (its rebase onto `v7` plus the review round there); the two commits below it are the same as before.

packages/core/src/components/Form.tsx
```diff
@@ -533,259 +1286,15 @@ export default class Form<T = any, S extends StrictRJSFSchema = RJSFSchema, F ex
         this.props.onChange(toIChangeEvent(nextState));
       }
       // oxlint-disable-next-line react/no-did-update-set-state -- guarded to prevent infinite loop
-      this.setSharedState(nextState);
+      this.setSharedState(prevState, nextState);
```
heath-freenome: **Medium: this diffs the derived state against `prevState`, not against the state it was derived from.**

`getSnapshotBeforeUpdate` derives `nextState` from `this.state`, but `setSharedState(prevState, nextState)` commits only the keys that differ from `prevState`. When a state update and a prop change land in the same render, a key whose derived value matches `prevState` but not `this.state` is never committed, so the form keeps only part of the update.

Repro on this head:

```tsx
const schema = { type: 'object', properties: { a: { type: 'string', minLength: 5 } } };

function Parent() {
  const [formData, setFormData] = useState({});
  return (
    <Form
      schema={schema}
      validator={validator}
      liveValidate='onBlur'
      formData={formData}
      onChange={(e) => setFormData(e.formData)}
      onBlur={() => setFormData({ a: 'hello' })}
    />
  );
}
// type "ab", then blur
```

Result: the field shows `hello`, but the error list still says `.a must NOT have fewer than 5 characters`.

What happens: the blur validation queues `errors: [minLength]`, and the parent's `setFormData` changes the prop in the same render. The derivation moves `formData` to `hello`, and the changed-field clearing resets `errors` to `[]`. That value is deep-equal to `prevState.errors`, so `replaceEqualDeep` hands back `prevState`'s reference and the key is skipped. `formData` is committed, but the error reset is dropped.

Before this commit (and on the `v7` base) the same steps end with `ab` and its error. The parent's value is lost there, but the state at least matches itself.

Suggested fix: `this.setSharedState(this.state, nextState);`. `this.state` here is still the state `getSnapshotBeforeUpdate` derived from, which is what the `setSharedState` doc comment says `from` should be. With that change the repro ends in a consistent state again, and the core suite passes (2275 tests). A test along the lines of the repro would pin it.

heath-freenome (commented):
Thanks for the fixes. I re-reviewed 5c504a4d9. The rest of the branch is unchanged apart from the rebase onto `v7`.

All three findings from last round are resolved:
- The controlled echo no longer wipes the inline error: the clearing now depends on `isLiveValidated`, and there's a new test.
- The `apply*` handlers no longer overwrite queued updates: `setSharedState(from, next)` commits only the keys that changed, and there's a new test for the `validateForm()`-inside-`onBlur` case.
- The upgrade guide now says `FormHandle<T>`.

One new issue, from the key diff: the `componentDidUpdate` call site compares against `prevState` rather than the state it derived from. Details and a repro are inline. The other five call sites pass the state they derived from, so they're fine.

packages/core/src/components/Form.tsx
jimmycallin: Fixed in c9a3a9bf7: `componentDidUpdate()` now calls `this.setSharedState(this.state, nextState)`. There's a test for the repro in `Form.behaviors.test.tsx` ("renders the data its parent holds when a prop change lands in the same render as a blur validation"). It fails on 5c504a4d9 and passes now. The core suite passes (2275 tests).

The mechanism is a little different from the one described, though the fix is the same. The key that got dropped on the head was `formData`, one render later than the one you traced. The blur's `setState` callback fires after `componentDidUpdate` with `this.state` still at `ab`, so the parent accepts `{ name: 'ab' }`. On the next pass the derived `ab` is deep-equal to `prevState.formData`, so it takes `prevState`'s reference and gets skipped, even though `this.state.formData` is `hello`. So the form showed `hello` while its controlled parent held `ab`. The test pins that the rendered value matches the parent's.

The changed-field clearing never resets `errors`, only `errorSchema`, and that is also true on `v7`. So after the fix the list still shows the `ab` error while the inline error is cleared. That split is the clearing's existing behavior, not part of this fix. The parent's `hello` being overwritten by the blur's own `onChange` also happens on `v7`.

heath-freenome:
## Code review

Reviewed the whole `Form.tsx` rewrite, `FormHandle.ts`, `constants.ts`, the changelog, the upgrade guide and the tests, plus all four existing review threads.

Verification on the PR head: `@rjsf/core` **2275 tests / 45 files pass**; root `lint`, `knip` and `typecheck` all clean.

### Existing feedback — all four threads verified as handled

| Thread | Fix | Verdict |
|---|---|---|
| `skipLiveValidate` folded into `mustValidate`, so an echoing controlled parent lost the inline error | `5c504a4d9` splits out `isLiveValidated` and passes `getFormDataChangedFields` only when `!isLiveValidated` | Correct — this exactly reproduces the old gate (`getStateFromProps` ran the clearing iff `!mustValidate`, computed without `skipLiveValidate`). Regression test present. |
| `setSharedState` committing the whole snapshot clobbered updates queued in between | `5c504a4d9` narrows the commit to keys differing from `from`; all 7 call sites pass their starting state | Correct. The inner `replaceEqualDeep(prevState, changed)` also neutralises deep-equal-but-new references, so the narrowing costs nothing. `onBlur` reads `committed = this.state` *after* the user's `onBlur`, which is what makes the `validateForm()`-from-`onBlur` case work. Regression test present. |
| Stale `FormHandle<T, S, F>` later in the upgrade guide | `cbcca3ca4` | Fixed (`FormHandle<T>` at guide line 697). |
| `componentDidUpdate` diffing against `prevState` instead of `this.state` | `c9a3a9bf7` | Fixed at the commit — but the sibling gate was not; see finding 2. |

### Findings

**1. `CHANGELOG-v7.md:75` — four duplicated bullets for one change, three of them contradicting the PR on whether it is breaking.**

Lines 75, 76, 77 and 80 are four near-identical restatements of the same "`Form` derives its state through module-level pure functions / `getStateFromProps()` is gone" entry (apparently one per rewrite of the commit message), and line 78 repeats the custom-error-clearing note already contained in 75/76/77. Lines 75–77 label the `getStateFromProps()` removal **BREAKING CHANGE**, while line 80 and the PR description both state explicitly that it is *not* breaking ("undocumented, never part of `FormHandle`… Noted in the changelog, not in the guide").

Concrete effect: the published v7 changelog lists one change four times and tells readers two opposite things about it. Suggest keeping line 80 plus the standalone entries at 78 and 79, and dropping 75–77.

**2. `packages/core/src/components/Form.tsx:1248` — `shouldUpdate` is still computed against `prevState` while the commit now diffs against `this.state`, so `c9a3a9bf7`'s fix is incomplete.**

`getSnapshotBeforeUpdate` derives `nextState` from `this.state`, shares it against `prevState`, and gates on `Object.entries(nextState).some(([k, v]) => v !== prevState[k])`. When a handler's `setState` and a prop change land in the same render, a derivation whose every key matches `prevState` — but not `this.state` — is discarded before `setSharedState(this.state, nextState)` at line 1289 ever runs.

Repro (fails on this head):

```tsx
function Parent() {
  const [disabled, setDisabled] = useState(false);
  const formData = useMemo(() => ({ name: 'ab', extra: 'x' }), []);
  return <Form ref={formRef} schema={{type:'object',properties:{name:{type:'string'}}}}
    validator={validator} formData={formData}
    omitExtraData liveOmit='onBlur' disabled={disabled}
    onBlur={() => setDisabled(true)} />;
}
// click the input, then tab away
```

`applyBlur` commits `{name:'ab'}` into `this.state`; the parent's `setDisabled` lands in the same batch. The derivation pulls `formData` back to the prop value, `replaceEqualDeep(prevState, …)` hands back `prevState.formData` (`{name:'ab',extra:'x'}`), every derived key then equals `prevState`, `shouldUpdate` is `false`, and `getFormData()` stays `{name:'ab'}` while the parent holds `{name:'ab',extra:'x'}`. That is the same "renders data its parent no longer holds" class the fix targeted.

`bcebb15c6` behaves identically, so this is an incomplete fix rather than a regression introduced here. Note when fixing that `shouldUpdate` also gates the `onChange` call at line 1285, so widening it to `v !== this.state[k] || v !== prevState[k]` rather than swapping the base keeps the callback timing unchanged.

**3. `packages/core/src/components/Form.tsx:468` — `validateFormData()` mixes `props.uiSchema` with the *committed* registry's `globalUiOptions`/`formContext`.**

`getUiRequiredErrorSchema()` is now passed `uiSchema` from `props` but `registry.globalUiOptions` and `registry.formContext` from `context`. For `applyValidation()` and the `validate()` handle, `context` is `this.state`, whose registry is only refreshed by `componentDidUpdate`'s `setState`. So a `validateForm()`/`validate()` call made after a `uiSchema` prop change but before that commit flushes sees the new `uiSchema` alongside the previous `ui:globalOptions` and `formContext` — a combination the current `v7` could not produce, since it read all three off `this.props`.

The window is narrow (I could not construct a repro through normal React scheduling), but the mismatch is real, and the doc comment's claim that these "cannot be mismatched by a caller" only holds for a freshly derived context.

### Checked and found correct (no finding)

- **Copy-on-write custom errors**: `ErrorSchemaBuilder`'s constructor `structuredClone`s, and `isPlainObject` rejects class instances so `replaceEqualDeep` never shares an old builder back — the committed builder really is left untouched, and a new one always commits.
- **`copyAlongPath()`**: correctly copies the root plus every container along `path.slice(0, -1)` and stops where the path leaves existing containers; this genuinely fixes the root-only shallow copy for nested cleared leaves.
- **`applyReset()`** reproduces the `IS_RESET` sentinel's semantics exactly (uncontrolled + no `initialFormData` ⇒ `defaultsFormData === undefined`), and correctly leaves `edit` alone.
- **`reconcileErrors()` / `deriveFormState()`** reproduce `getStateFromProps()`'s error paths key-for-key apart from the three documented behavior changes; `schemaValidationErrors` clearing on `noValidate`/schema change, the #5041 merge base, and the resolved-schema live validation are all intentional, changelogged, guide-documented and covered by new tests.
- **`FormHandle.validate()` narrowing**: no remaining caller in the repo (playground, themes, docs, tests) passes the removed parameters; `IS_RESET` and `getStateFromProps` have no live references outside stale `lib/` build output.

jimmycallin:
Thanks. All three findings reproduced and fixed. The branch is also rebased onto the current `v7` (head `68a3c2536`).

**1. Duplicated changelog bullets.** Confirmed. `300a1b788` drops the three duplicates, including the two that called the `getStateFromProps()` removal breaking. What's left is the non-breaking entry, plus the separate entries for custom-error clearing and `noValidate` on blur.

**2. `shouldUpdate` still gated on `prevState`.** Confirmed with your repro: `getFormData()` returned `{ name: 'ab' }` while the parent held `{ name: 'ab', extra: 'x' }`. Fixed in `06bef0203` the way you suggested. The gate is now `value !== prevState[k] || value !== this.state[k]`, and the base and the `onChange` check are unchanged. New test in `Form.behaviors.test.tsx`: "holds the data its parent holds when a prop change lands in the same render as a blur that omits extra data".

**3. `validateFormData()` mixing `props.uiSchema` with the committed registry.** Confirmed, and it can be reproduced. A parent `useLayoutEffect` runs after the form's `componentDidUpdate` but before the `setState` queued there commits. If that effect calls `validateForm()` right after a `formContext` change, the function-form `uiSchema.items` still sees the old `formContext`. The test changes `formContext` to one whose `items` marks a field `ui:required`, and the call returned `true`. Fixed in `68a3c2536`: `formContext` and `ui:globalOptions` now come from the same props as `uiSchema`, normalized the way `buildRegistry()` does it (`formContext ?? {}`). The doc comment's "cannot be mismatched" claim is gone. Test: "reads ui:required with the formContext of the props it validates with, before the form commits them".

Both new tests fail before their fix. On the rebased head, `@rjsf/core` passes 2277 tests. Root `lint`, `typecheck`, `knip`, `cs-check` and `pnpm dedupe --check` are clean.

heath-freenome:
## Code review — `68a3c2536`

Re-reviewed the three fix commits (`300a1b788`, `06bef0203`, `68a3c2536`) and the rebase, then did a fresh pass over `Form.tsx` against both `origin/v7` and `origin/main` (v6), plus the test files and the two docs.

**Verification on this head:** `@rjsf/core` **2277 tests / 45 files pass**; root `lint`, `typecheck`, `knip` and `cs-check` clean; CI green (`build 22.x/24.x/26.x`, `Analyze`, `size`).

### Previous findings — all three resolved

| Finding | Fix | Verdict |
| --- | --- | --- |
| 1. Four duplicated changelog bullets, three contradicting the PR on whether the `getStateFromProps()` removal is breaking | `300a1b788` | Fixed. `CHANGELOG-v7.md:75-81` is six distinct bullets, appended last under `## @rjsf/core`, no contradiction left. |
| 2. `shouldUpdate` gated on `prevState` while the commit diffs against `this.state` | `06bef0203` | Correct, and the narrow form. Widening the gate rather than swapping its base means the newly-admitted cases are exactly those where every derived key equals `prevState` — so `nextState.formData === prevState.formData` and the `onChange` gate at `Form.tsx:1281` stays false. Callback timing is unchanged, as intended. I also checked this is not redundant with `4defd5edc`: reverting only the `\|\| value !== this.state[…]` term fails `06bef0203`'s test alone, reverting only `setSharedState(this.state, nextState)` fails both. Both are load-bearing. |
| 3. `validateFormData()` mixing `props.uiSchema` with the committed registry's `globalUiOptions`/`formContext` | `68a3c2536` | Correct, and it restores `origin/v7` exactly: the same `uiSchema?.[UI_GLOBAL_OPTIONS_KEY]` and `props.formContext ?? ({} as F)` expressions, matching `buildRegistry()` (`Theme.ts:84,87`) key for key. The overstated doc comment is gone with it. |

**No new implementation defects.** I went looking: `resolveSchemaUtils`/`deriveRenderContext`/`deriveFormData` reproduce `getStateFromProps()`'s reuse rules; `applyBlur` matches v7's `onBlur` apart from the changelogged `noValidate` gate; `applyChange`'s `!noValidate && newErrorSchema` branch and the `deferLiveValidate` skip are v7's; the registry is rebuilt exactly as often as before; `IS_RESET` has no reference left anywhere outside stale `lib/` output and `constants.ts` was never re-exported; no caller in the repo passes `validate()`'s removed parameters or spells `FormHandle<T, S, F>`.

Checked and deliberately **not** filed: `VALIDATION_PROP_KEYS` omits `uiSchema` and `formContext` even though both feed `getUiRequiredErrorSchema()`, so changing either does not re-validate under `liveValidate: 'onChange'`. That set is byte-identical to `origin/v7` and the upgrade guide already documents the consequence at line 548.

The findings below are test coverage and documentation.

### Test coverage

**1. `copyAlongPath()`'s fix (`5b3bc37ed`) is untested.** I reverted both halves at this head — the scalar guard, and `break` back to `return data` — and the full core suite stays green at **2277/2277**.

That fix is what stops `setByPath()` from auto-vivifying the remainder of a path directly into an object `replaceEqualDeep()` handed back, which is the committed state's own object, or for a controlled form the parent's. It is the same hazard this PR's own upgrade-guide bullet warns callers about ("Treat the `formData` you receive … as immutable … so mutating it in place corrupts your own state"), so it is worth a regression test rather than trust. Its only call site is `Form.tsx:911` (the `plainLeafWasCleared` re-clear), so a test wants a cleared plain leaf under a path whose intermediate container is absent, asserting the caller's `formData` object was not mutated.

**2. The `noValidate` half of `reconcileErrors()`'s error reset is untested.** Changing `Form.tsx:676` from `props.noValidate || isSchemaChanged` to `isSchemaChanged` alone also leaves the suite at **2277/2277**. The `isSchemaChanged` half is covered (setting the expression to `false` fails a test); the `noValidate` half — changelogged as "`noValidate` clears the two the same way, where it already cleared the displayed errors" — is not. No test in `packages/core/test` references `schemaValidationErrors` at all.

**3. `Form.behaviors.test.tsx:1349` asserts agreement rather than the behavior in its title.**

```tsx
expect(container.querySelector('input')).toHaveValue(parentData.name);
```

`parentData` is a module-level mirror of the parent's own state, so this passes whenever the DOM and the parent agree, never saying which value won. I deleted the `onBlur={() => setFormData({ name: 'hello' })}` prop — removing the prop-change-during-blur the title is about — and the test still passes. It does fail pre-fix, so it pins *something*, but asserting the literal `'ab'` and that the blur handler ran would pin the case it names. The `liveOmit` sibling at line 1374 is well specified by contrast: it fails when its prop change is removed.

### Documentation

**4. `v7.x upgrade guide.md:32` — the `validate()` replacement snippet is lossy and disagrees with the prose above it.**

```diff
-const { errors } = formRef.current.validate(data, otherSchema, otherSchemaUtils);
+const { errors } = customValidator.validateFormData(data, otherSchema);
```

`customValidator` is never introduced; the prose one line up says to call `schemaUtils.getValidator().validateFormData(...)` and the `-` line uses `otherSchemaUtils`, so the `+` line matches neither. More importantly it drops behavior: `Form.validate()` is `validateFormData(this.props, this.state, formData)` (`Form.tsx:1299`), which hands `customValidate`, `transformErrors` and `uiSchema` to the validator and then merges `getUiRequiredErrorSchema()` on top (`Form.tsx:459-482`). A bare `validateFormData(data, otherSchema)` does none of that, so a reader following the diff silently loses their `customValidate`, `transformErrors` and every `ui:required` error.

**5. `CHANGELOG-v7.md:81` labels the `FormHandle` narrowing BREAKING, but the guide documents it only under `## New deprecations`.**

It sits at guide line 29, inside `### Reading Form internals through a ref`. Every other breaking change in that file has its own `### … BREAKING CHANGE` heading under `## Breaking changes` — thirty-odd, including this PR's own two at lines 646 and 660. Someone scanning the breaking-changes section for what will break their build won't find the one that changes a type signature they may have hand-written.

**6. `Form.tsx:1249` — the widened `shouldUpdate` is a v6→v7 behavior change with no changelog bullet.**

`06bef0203` is the right fix and I'm not asking for it back. But it isn't only a repair of something this PR introduced: `origin/main` (v6) gates on `const shouldUpdate = !deepEquals(nextState, prevState);` (`Form.tsx:587`) and the `v7` base has the same shape, so both drop a derivation matching `prevState` but not the state it would replace. The user-visible result is what this PR's two new tests pin — a controlled form that used to keep rendering data its parent no longer held now snaps to the parent's value. That deserves a line under `## @rjsf/core` the way the three "errors that should not have been on display" got one.

**7. `CHANGELOG-v7.md:77` misattributes what the new functions do.** The bullet says `deriveRenderContext()` "resolves the schema utilities, retrieved schema and registry for the current props and data". It resolves neither of the first two — both arrive as arguments (`deriveRenderContext(props, retrievedSchema, prev, resolved)`, `Form.tsx:390-395`), produced by `resolveSchemaUtils()` (`Form.tsx:361-377`) and by `deriveFormData()`'s `retrieveSchema` loop (`Form.tsx:592-595`); it takes no form data either. Building the registry is the only accurate part. Weaker, same bullet: "`deriveFormState()` adds the render context … on top" — the context comes back out of `deriveFormData()` (`Form.tsx:627-628`, consumed at `:755`); what `deriveFormState()` adds on its own is the error reconciliation.

**8. `v7.x upgrade guide.md:548` — "whenever the schema, validator and merge settings are unchanged" omits `defaultFormStateBehavior`.** The reuse test is `doesSchemaUtilsDiffer(validator, schema, defaultFormStateBehavior, customMergeAllOf)` (`Form.tsx:366`), and `createSchemaUtils.ts:120-125` compares it. A `formData` replaced in the same update as a changed `defaultFormStateBehavior` rebuilds the utilities, so `areSchemaUtilsReused` is false (`Form.tsx:564`) and validation falls back to the root schema, which the sentence promises it won't. `CHANGELOG-v7.md:80`'s "whenever the schema utilities are unchanged" is exact — the guide could borrow it.

### PR description

**9.** Five stale points, worth a pass before merge since this is what the release notes get written from:

- The **"Stacked on #5310"** banner is obsolete — #5310 merged 2026-09-22, the merge base is `da6f38cce` on `v7`, and the diff is this PR's own commits over 9 files.
- "the **eleven** commits of this PR start at…" and "every one of the **9** commits typechecks on its own" — there are 17.
- "**Breaking changes.** Two, both in public API surface" — three, counting `d316b4655` (`refactor(core)!: narrow FormHandle.validate()`), which the changelog does label BREAKING. The same sentence's "each … has its own section in the v7 upgrade guide" is finding 5.
- "`@rjsf/core`: **2156 tests** across 45 files" — 2277 now.
- The **"Validation schema for a derivation"** paragraph sits *below* the generated-with footer, reading as a mis-paste; it belongs in the "Other behavior changes" list it restates.

### Checked and found correct (no finding)

- The widened gate cannot loop: a state-only re-render hands over the same props object and short-circuits at `Form.tsx:1204`.
- `setSharedState`'s inner `replaceEqualDeep(prevState, changed)` neutralises the deep-equal-but-new references the widening lets through, so nothing commits a spurious identity change.
- Every other new test in the PR does fail without its fix — I checked each against its commit's parent, including the two from the last round and the `#5041` and `noValidate`-on-blur regressions.
- `uiSchema.test.tsx`'s 2-line change is comment text only, and `StringField.test.tsx` is purely additive; neither weakens an existing expectation.
- The changelog convention holds — all six new bullets are appended last under `## @rjsf/core`, immediately before `## @rjsf/daisyui`.

heath-freenome (approved):
A bunch of documentation and other non-code fixes are pending. Approving, assuming you properly resolve the feedback I left and run a few claude code reviews

jimmycallin:
Thanks. Addressed at `983525507`, rebased onto `v7` at `a1a74e2e2` (#5321, #5323, #5324; changelog and upgrade-guide conflicts only).

**1. `copyAlongPath()` untested: not added, because I couldn't reach the branch through the form.** I tried five scenarios: a nested leaf under an absent container (`a.b` and `a.b.c` on `{ a: {} }`), `skipEmptyDefaults`, and two root `if`/`then` schemas that drop `b` on sanitization. Each time `applyChange`'s first `setByPath()` builds the missing containers on the `structuredClone`d data, and `deriveFormData()` keeps them. So the derived data always contains the full path, `replaceEqualDeep()` never hands back a committed container that is missing a child on that path, and the early stop never runs. The fix stays in as a guard, since it costs nothing and the hazard is real if the derivation ever starts dropping containers. I didn't export `copyAlongPath()` to test it directly, because everything under `src` can be deep-imported through `./lib/*.js`. If you'd rather have that export or a scenario I missed, say so.

**2. `noValidate` half of the reset: fixed in `9d95461c8`.** New test in `Form.derivation.test.tsx`: "turning noValidate on drops the validator results, so turning it back off does not bring them back". It fails when the expression is reduced to `isSchemaChanged` alone.

**3. Agreement-only assertion: fixed in `9d95461c8`.** The test now asserts the literal `'ab'` for both the parent and the input, and that the blur handler ran once. It still fails with `componentDidUpdate` diffing against `prevState`.

**4. `validate()` snippet: fixed in `ab0f1e617`.** It now calls `otherSchemaUtils.getValidator().validateFormData(data, otherSchema, customValidate, transformErrors, uiSchema)`, and a sentence says how to get the `ui:required` errors back (`getUiRequiredErrorSchema()` + `validationDataMerge()`).

**5. Moved to its own section: fixed in `ab0f1e617`.** `### FormHandle.validate() takes only the data BREAKING CHANGE` now sits under `## Breaking changes`, after the other two sections from this PR. The deprecations section links to it.

**6. Changelog bullet for the widened gate: moot, because I reverted the widening (`983525507`).** A `/code-review high` pass on the head found that the `|| value !== this.state[key]` term from `06bef0203` reverts a controlled edit. It happens when the parent re-renders in the same event as the edit, before the edit's `onChange` reaches it. A wrapper `<div onChange>` that bumps a `className` is enough, and the input renders `'' → 'x' → '' → 'x'` where `v7` renders `'' → 'x'`. Your omit-on-blur repro is the same shape: a prop change in the same batch as the form's own commit, before the parent has been told. The difference is only that that parent never accepts. Snapping early fixes the parent that never accepts but breaks every parent that does, so the gate is back to `prevState` alone, as on `v7`. That case goes back to `v7`'s behavior of snapping on the parent's next render. The A4 ownership switch is where "render what the parent holds" becomes exact. The `06bef0203` test is replaced by "keeps an edit when the parent re-renders in the same event, before it has been told of the edit", which fails with the widening restored. `4defd5edc` (`setSharedState(this.state, …)`) stays, and its test still passes.

**7. Misattributed bullet: fixed in `ab0f1e617`.** It now says: `resolveSchemaUtils()` reuses or rebuilds the utilities, `deriveFormData()` resolves the schema and hands back the context whose registry `deriveRenderContext()` builds, and `deriveFormState()` reconciles the errors on top.

**8. Missing `defaultFormStateBehavior`: fixed in `ab0f1e617`.** The guide now says "whenever the schema utilities are unchanged (the same `schema`, `validator`, `defaultFormStateBehavior` and `customMergeAllOf`)".

**9. PR description: updated.** I removed the stacking banner and the per-commit typecheck claim, and corrected the commit and test counts. It now lists three breaking changes. The stray paragraph at the end became the corrected resolved-schema bullet under *Other behavior changes*, 

**Rest of the code review.** I also fixed two diff-style comments it flagged (`Form.tsx` `applyChange`, `uiSchema.test.tsx`). Left as they are, all present on `v7` too: the falsy `formData ?` guard in `validateFormData()` (a root `0`/`false`/`''` reaches the validator as `undefined`), the changed-field clearing leaving `errors` while it clears `errorSchema`, and `VALIDATION_PROP_KEYS` omitting `uiSchema`/`formContext`. The first two are worth their own issues. The third is documented in the guide.

Gates at `983525507`: uncached `lint`, `knip`, `build`, `typecheck`, `cs-check`, `pnpm dedupe --check`, and `pnpm test` for all 15 projects (`@rjsf/core` 2614).

